PR #64, opened 2017-08-24
Add code coverage from system tests

vazexqi:
### What does this PR do?

* Instruments the code prior to running the system tests 
* Runs the system tests with the instrumented code
* Remaps the coverage info to account for JS <-> TS.

W-4254205

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -37,7 +37,8 @@ build_script:
 test_script:
   - node --version
   - npm --version
-  - npm run test
+  - npm run test:without-system-tests
+  - npm run coverage:system-tests
 
 on_finish:
   - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,8 @@ script:
   - npm run lint
   - |
     if [[ $TRAVIS_OS_NAME == "linux" ]]; then
-      npm run test
+      npm run test:without-system-tests
+      npm run coverage:system-tests
     fi
   - |
     if [[ $TRAVIS_OS_NAME == "osx" ]]; then
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -141,3 +141,41 @@ the system tests.
 1. `DEBUG_SPECTRON=1 npm run test:system-tests`.
 1. Use the "Attach to Process (Inspector Protocol)" debug configuration to
    attach to the process (uses port 9229).
+
+## Coverage numbers from System Tests
+
+System tests exercise a wide range of modules and, thus, are well-suited for
+providing an upper bound on our coverage numbers. However, they are harder to
+instrument since they are not isolated to one module; **all** modules need to be
+instrumented.
+
+The dynamic loading technique that we use with `testrunner.ts` is not sufficient
+since it _decaches_ the require cache and forces it to load dynamically
+instrumented files. That is only sufficient for that particular module. When you
+need to instrument all modules, you need a parent _module_ that can load
+everything. Such a module is not impossible but might deviate from usual user
+interactions, i.e., there is too much magic going on that we might not be able
+to trust the coverage numbers.
+
+Thus, instead of dynamic instrumentation, we opt for static instrumentation.
+
+1. Build all the modules, like normal, and write their JS files and source maps
+   to disk.
+1. Use `istanbul` (version 1.0 and above) that supports source maps to
+   instrument the newly build modules from step 1 and write it out to disk.
+1. `istanbul` works by source instrumentation and maintaining a global map of
+   files to covered locations called `__coverage__`.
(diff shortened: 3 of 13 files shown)


.appveyor.yml
```diff
@@ -37,7 +37,8 @@ build_script:
 test_script:
   - node --version
   - npm --version
-  - npm run test
+  - npm run test:without-system-tests
```
vazexqi: Running the system tests with coverage requires a different routine than normal testing. But since we don't want to run them twice, we split it up here. So we run the non-system tests and then the system tests, with the coverage mechanism.

.travis.yml
```diff
@@ -67,7 +67,8 @@ script:
   - npm run lint
   - |
     if [[ $TRAVIS_OS_NAME == "linux" ]]; then
-      npm run test
+      npm run test:without-system-tests
```
vazexqi: Ditto.

docs/tests.md
```diff
@@ -141,3 +141,41 @@ the system tests.
 1. `DEBUG_SPECTRON=1 npm run test:system-tests`.
 1. Use the "Attach to Process (Inspector Protocol)" debug configuration to
    attach to the process (uses port 9229).
+
+## Coverage numbers from System Tests
+
+System tests exercise a wide range of modules and, thus, are well-suited for
+providing an upper bound on our coverage numbers. However, they are harder to
```
vazexqi: "Upper bound" is the key here. Just loading the instrumented modules brings in a lot of line coverage, even if they are not tested explicitly.

docs/tests.md
```diff
@@ -141,3 +141,41 @@ the system tests.
...
+   instrument the newly build modules from step 1 and write it out to disk.
+1. `istanbul` works by source instrumentation and maintaining a global map of
+   files to covered locations called `__coverage__`.
+1. At the end of our extension deactivation, we grab the value of the
+   `__coverage__` variable and write it out into coverage/coverage.json
+1. The `coverage/coverage.json` file still uses the JS line numbers. We then use
+   `remap-istanbul` to remap the line numbers to their TypeScript counterparts.
+1. Then we replace the original `coverage/coverage.json` with this new version
+   and generate the corresponding coverage.lcov.
+1. `coverage/coverage.json` and `lcov.info` are uploaded to codecov.io.
```
vazexqi: I have found that codecov.io seems to be able to handle lcov.info much better but I'm sending both to it per the way the other tests work.

packages/salesforcedx-vscode-apex/coverconfig.json
```diff
@@ -4,6 +4,6 @@
   "relativeCoverageDir": "../../coverage",
   "ignorePatterns": ["**/node_modules/**"],
   "includePid": false,
-  "reports": ["json", "html", "lcov"],
+  "reports": ["json", "lcov"],
```
vazexqi: I've removed the HTML coverage since they don't seem useful. They are also not used by codecov.

packages/system-tests/package.json
```diff
@@ -25,12 +17,29 @@
...
+  },
+  "scripts": {
+    "compile": "tsc -p ./",
+    "lint": "tslint --project .",
+    "watch": "tsc -watch -p .",
+    "clean": "shx rm -rf .vscode-test && shx rm -rf node_modules && shx rm -rf out",
+    "postinstall": "node ./node_modules/vscode/bin/install",
+    "pretest": "npm run compile && node ../../scripts/download-vscode-for-system-tests",
+    "test": "node out/src/main.js",
+    "coverage": "npm run pretest && node ../../scripts/instrument-salesforcedx-vscode-extensions && cross-env COLLECT_COVERAGE=1 npm run test && node ../../scripts/remap-coverage"
```
vazexqi: @Sweetman This is how you set the environment variable local to the next executing command. You should use `cross-env` since that sets it in a platform-agnostic way.

scripts/instrument-salesforcedx-vscode-extensions.js
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env node
+
+const shell = require('shelljs');
+shell.set('-e');
+shell.set('+v');
+
+const path = require('path');
+
+const extensionsToInstrument = [
```
vazexqi: If we add a new module this will need to be updated.
I don't want to just scan the modules under /packages automatically since they could be packages that we don't care about in terms of code coverage.

vazexqi:
Travis has gotten into one of those weird states where it just refuses to build this properly.
Timing out, erroring out with lerna ERR! execute Xlib:  extension "RANDR" missing on display ":99.0".

I've run this locally without any issues so I'm going to merge this in. There is no change in functionality except to report code coverage numbers from the runs.

PR #69, opened 2017-08-25
Added missing engines definition / removed main key

### LINKED ISSUE #68: Missing 'engines' key in package.json of salesforce-utils-vscode in develop branch
### Missing 'engines' key in package.json of salesforce-utils-vscode

The develop branch cannot be used for launching the extensions. An error is raised that points to a missing engines key in the package.json of salesforce-utils-vscode.

Adding the key raises another error:
/.../salesforcedx-vscode/packages/salesforcedx-utils-vscode]: properties `activationEvents` and `main` must both be specified or must both be omitted

Removing the main key from the package.json solved that issue.

### Steps To Reproduce:
* Clone the repo, make sure that you're in the develop branch.
* In VS Code, open the debug view and from the launch configuration dropdown, pick "Launch Extensions".
* The new VS Code windows opens with an error message.
* Add engines key to package.json:
* Re-launch from debug view - new error message appears.

### Expected result

Launching the extensions without an error message.

muenzpraeger:
### What does this PR do?

This PR adds the 'engines' key to the package.json of salesforcedx-utils-vscode. It also removes the 'main' key in the package.json.

Fixes #68

### DIFF
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -7,7 +7,7 @@
   "license": "BSD-3-Clause",
   "categories": [
     "Other"
-  ],
+    ],
   "dependencies": {
     "rxjs": "^5.4.1",
     "tree-kill": "^1.1.0"
@@ -29,6 +29,9 @@
     "source-map-support": "^0.4.15",
     "typescript": "2.4.0"
   },
+  "engines": {
+    "vscode": "^1.15.0"
+  },
   "scripts": {
     "compile": "tsc -p ./",
     "lint": "tslint --project .",
@@ -42,6 +45,5 @@
       "text-summary",
       "lcov"
     ]
-  },
-  "main": "./out/src/"
+  }
 }
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -7,7 +7,7 @@
   "license": "BSD-3-Clause",
   "categories": [
     "Other"
-    ],
+  ],
   "dependencies": {
     "rxjs": "^5.4.1",
     "tree-kill": "^1.1.0"
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -30,7 +30,7 @@
     "typescript": "2.4.0"
   },
   "engines": {
-    "vscode": "^1.15.0"
+    "vscode": "^1.13.0"
   },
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-utils-vscode/package.json
```diff
@@ -29,6 +29,9 @@
     "source-map-support": "^0.4.15",
     "typescript": "2.4.0"
   },
+  "engines": {
+    "vscode": "^1.15.0"
```
vazexqi: @muenzpraeger - Can we peg this one at ^1.13 for now like the rest of the extensions (with the exception of the debugger that does need 1.15 features but is not shipping yet).

Also, the formatting seems a bit off. Did you apply the prettier formatter?

muenzpraeger:
@vazexqi I've changed the vscode engine version and fixed the formatting.

Prettier is installed, but for some weird reason it wrongly intended several parts of the file. I had to switch to "Text" (instead of JSON) and to manually re-format it so that formatting wouldn't cause to many file changes.

vazexqi:
LGTM. Merging this in.

There are some weird issues with AppVeyor today where it doesn't see the %SFDX_CI_DEVHUB_USERNAME% variable. But that is more of a CI issue than this commit itself.

sfdx force:auth:jwt:grant --clientid %SFDX_CI_DEVHUB_CLIENTID% --username %SFDX_CI_DEVHUB_USERNAME% --jwtkeyfile devhub.key --setdefaultdevhubusername --setalias devhub

packages/salesforcedx-utils-vscode/package.json
muenzpraeger: @vazexqi I'm wondering where these formatting changes (see keys "categories" and "activationEvents" are coming from.

https://github.com/muenzpraeger/salesforcedx-vscode/commit/53761a4308edf25df231815cab8d90ea67e00ab0#diff-d79693d63f96a4709135983f4a6f9e12R24

Prettier 0.22.1 is installed and I haven't applied any custom formatting for it. Would be nice to know where the changes are coming from.

vazexqi: This, annoyingly, comes from the publish process. See https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/scripts/publish.js#L51 

I don't know why when we use lerna to do the publishing, it cannot just reformat the lines (the dependencies) that it needs to update. Instead, it formats the file.

If someone could investigate why this comes to be, I would be very grateful. This is likely an issue with lerna itself.

PR #70, opened 2017-08-26
James/lightning cmp create

Sweetman:
### What does this PR do?
Add lightning component create to the command palette
@W-4197626@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceLightningComponentCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceLightningComponentCreate.ts
@@ -30,7 +30,7 @@ import {
 
 const LIGHTNING_CMP_EXTENSION = '.cmp';
 
-class ForceLightningCmpCreateExecutor extends SfdxCommandletExecutor<
+class ForceLightningComponentCreateExecutor extends SfdxCommandletExecutor<
   DirFileNameSelection
 > {
   public build(data: DirFileNameSelection): Command {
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/commands/forceLightningComponentCreate.ts
```diff
@@ -0,0 +1,97 @@
...
+          .openTextDocument(
+            path.join(
+              vscode.workspace.rootPath,
+              response.data.outputdir,
+              // fileName is also used to create a subdirectory for the component in the aura directory
+              response.data.fileName,
+              response.data.fileName + LIGHTNING_CMP_EXTENSION
+            )
+          )
+          .then(document => vscode.window.showTextDocument(document));
```
vazexqi: @Sweetman - Let's alter this later when you have put all your commands in.
I am trying to avoid using .. then() to be more consistent. I just want to use await.
We should lift this up into a promise and then await it.
So

```
const document = await vscode.workspace(...);
vscode.window.showTextDocument(document);
```

packages/system-tests/scenarios/scaffolding.test.ts
```diff
@@ -154,4 +154,36 @@ describe('Scaffolding commands', () => {
...
+    const lightningComponentTab = await common.getTab(`${fileName}.cmp`);
+    expect(lightningComponentTab).to.be.not.undefined;
+    if (lightningComponentTab) {
+      await common.closeTab();
+    }
+
+    await app.command('workbench.action.quickOpen');
+    await common.type(fileName);
+    const elCount = await common.getQuickOpenElements();
+    expect(elCount).to.equal(6);
```
vazexqi: Change the magic number 6 to something more descriptive.

I think the number 6 is also shared with the .app test as well.

Sweetman:
Included in PR #76

mshanemc:
Thanks @Sweetman @vazexqi you're making my life better.  :)

Issue #65, opened 2017-08-25
Support SOQL queries via SFDX


muenzpraeger:
### Functionality
A common use case when developing with the Platform is to query existing data. For that the sfdx command to query data via SOQL and output that data should be implemented.

### Implementation
* A new menu binding "SFDX: Execute SOQL query" will be implemented.
* User enters the wanted SOQL query into a VScode inputBox.
* SOQL query will be run against the project scratch org, output in JSON format.
* Rendering of the output JSON as HTML in preview side pane (maybe using mustache/handlebars).

muenzpraeger:
Initial implementation

https://github.com/muenzpraeger/salesforcedx-vscode/tree/sfdx-soql-query/packages/salesforcedx-vscode-apex

muenzpraeger:
@vazexqi I'm wondering what in your opinion the preferred option would be to fetch the json output of a sfdx command from the ChannelService. Any thoughts on that before I add something would be welcome.

muenzpraeger:
I played a bit around with that over the weekend. Generating a simple HTML doc works just fine. I used a custom URI scheme (sfdx://...) for that.

vazexqi:
> I'm wondering what in your opinion the preferred option would be to fetch the json output of a sfdx command from the ChannelService. Any thoughts on that before I add something would be welcome.

We have an example of this in https://github.com/forcedotcom/salesforcedx-vscode/blob/9e44a54603d454c4e8e44809d591586c5d487f83/packages/salesforcedx-apex-debugger/src/core/sessionService.ts#L63 that @DatGuyJonathan added.

For the display of the of the SOQL results, I was thinking of something like https://marketplace.visualstudio.com/items?itemName=ms-mssql.mssql

I think as a first pass we can display the raw STDOUT results (not the JSON). I think the STDOUT results is actually more human readable than the JSON output. JSON is meant for processing by a downstream consumer.

In the long term, I am thinking of something like the table/grid display of https://marketplace.visualstudio.com/items?itemName=ms-mssql.mssql

muenzpraeger:
Agreeing on the long term.

Having said that - once develop has been merged I'm ready to rebase and push the soql query action. ;-)

PR #72, opened 2017-08-28
Adding command for force:apex:execute

### LINKED ISSUE #66: Support execution of anonymous Apex via SFDX
### Functionality
To simplify executing anonymous apex (not using sfdx from the built-in VScode terminal window) a custom command should be implemented.

### Implementation

* A new menu binding "SFDX: Execute Anonymous Apex" will be implemented. Also a menu addition (in "editor/context" for .cls and .trigger files).
* Input will be saved in temp file with random file name.
* Anonymous Apex command gets executed and temp file deleted. Output gets send to standard "Output" console window.

muenzpraeger:
### What does this PR do?
This PR adds the command force:apex:execute.

Based on the selected text in the active document a temp file gets created, the command executed and then the file cleaned up.

This fixes #66


packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -120,6 +121,49 @@ export type DirFileNameSelection = {
   outputdir: string;
 };
 
+export type TempFile = {
+  fileName: string;
+};
+
+export class CreateApexTempFile
```
vazexqi: This could probably be moved into the forceApexExecute.ts file.
I don't think there is much reuse for this outside for execute anonymous.
If we do find that we have reuse we can move it back here.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -120,6 +121,49 @@ export type DirFileNameSelection = {
...
+        'tools',
+        'tempApex.input'
+      );
+      const editor = await vscode.window.activeTextEditor;
+
+      if (!editor || editor.selection.isEmpty) {
+        return { type: 'CANCEL' };
+      }
+      const document = editor.document;
+      const writeFile = await writeFileAsync(fileName, document.getText());
```
vazexqi: Maybe I'm reading https://code.visualstudio.com/docs/extensionAPI/vscode-api#TextDocument wrongly but won't document.getText() get the _whole_ text of the current active editor?

Based on what you had on the line to check for editor.selection I thought you wanted to have the selection piped to a new file?

vazexqi:
@muenzpraeger - Also, did you want to add a right-click menu to the editor when there is selected text and enable the SFDX: Execute anonymous Apex command there?

packages/salesforcedx-vscode-core/src/commands/commands.ts
muenzpraeger: I'll move the whole TempFile/CreateApexTempFile to the forceApexExecute.ts file. Was just following the existing pattern. ;-)

muenzpraeger:
@vazexqi I was considering about adding a context-menu, but that may likely only make sense for .cls and .trigger files. Also spinning up a new ("Untitled") doc to shoot some Apex code is IMHO also needed. So we would have to check for 3 specific document types.

Are there already specific checks for this in the extension or planned? If not I would for now just leave it as it is, as that would a permanent context menu which may often not be used. Less is more. ;-)

packages/salesforcedx-vscode-core/src/commands/commands.ts
muenzpraeger: Good catch as I forgot to pass the selection as parameter. Done.

vazexqi:
Here's my thought process on this.

1. It's unlikely that you would want to type the execute anonymous into an input box since it's too complicated. I think we agree on that front and we would rather use a selection or the current text document.
1. I think we should use the contents of the *any* currently active editor. I don't think it matters what the extension of the editor is. Any editor will do. My reasoning is that sometimes a snippet of text to use for execute anonymous might not come from a .cls or .trigger file. It might just be a random file (or temp file) or an "Untitled" file like you noted.
    * (If there is an active editor AND something is selected, see [editorHasSelection](https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts)): Enable the command "SFDX: Execute Anonymous Apex with Current Selection".
    * (If there is an active editor), Enable the command "SFDX: Execute Anonymous with Current Editor"
1. I am also thinking that we might want to enable both commands only when there is a default scratch org set. If you don't have a default scratch org set, then the commands will error out immediately since they don't have an org to work with. _I don't know of an easy way to detect if there is a default scratch org_, short of checking the hidden `.sfdx/sfdx-config.json` file in the current project. This will need to be pulled out into a when clause
1. [Addition] This is where we can add the right-click menu if we want. We can use https://code.visualstudio.com/docs/extensionAPI/extension-points#_example without specifying a resourceId so that it appears in all text editor, based on the criteria above.

Thoughts? Also, just thinking out loud, I would prefer to get the execute anonymous feature in first before the SOQL one. I want to think more about the SOQL one. There are overlapping pieces for both (possibly, the selection since you might not want to just type out a long SOQL in an input box).

muenzpraeger:
I'm agreeing with point 1 and 2. The latter sounds good as it kind of "imitates" the current functionality of the dev console.

I'm not sure about the check about the default org. The same precondition check would IMHO apply for example to force:source:push and other commands.

Adding "active text" to #65 makes sense. But I'd keep the input box for when no text is selected.

muenzpraeger:
Hit enter before finishing the though process. ;-)

In terms of consistency the input dialog doesn't make sense for #65 - same behavior for Apex and SOQL.

guw:
AFAIK the default scratch org is not visualized yet. We used a label decorator in Eclipse. Thus, it might be intransparent to a user why something is enabled/disabled. I would go with the error message for now.

guw:
BTW, the way we did detect the default workspace org in Eclipse was by parsing/reading `.sfdx/sfdx-config.json`. That was determined to be reasonable.

vazexqi:
Agreed. We should do the same checks for the SFDX: Push/pull commands. 

Let's **not** do 3 and 4 in this PR. I will create a story to improve those separately.

@muenzpraeger: Could you do 1 and 2 and then update this PR?

Sweetman:
hey @muenzpraeger I'd like to help out with moving this and the soql query PR you have to being merged in so let me know if there's any work you would want to share with me. I started seeing how to go about testing this and could contribute it to this PR if you want to add me to your fork. If you already have it in the works though, I can continue working on other parts.

muenzpraeger:
I rebased from the current develop and added the changed (two Apex execute commands).

@Sweetman Is there already some test framework in place that does e2e with scratch org involved? I tested with an initial UI implementation like your scaffolding, but that's IMHO not e2e.

vazexqi:
Closed via PR #152

PR #80, opened 2017-08-29
Reformat with prettier after lerna publish (release/v40.8.0 branch) 

vazexqi:
### What does this PR do?

When we use lerna --publish, it rewrites package.json and destroys the formatting from prettier. This restores the formatting.

I'm going to have to make a similar change in the develop branch to keep all the formatting consistent.

None.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,8 @@
 {
   "devDependencies": {
-    "lerna": "2.0.0",
     "istanbul": "1.1.0-alpha.1",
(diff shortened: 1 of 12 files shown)


vazexqi:
CC: @muenzpraeger per your comment on why formatting is messed up.

muenzpraeger:
@vazexqi Good to see that it's not me. ;-)

PR #82, opened 2017-08-29
Fix check version error

vazexqi:
### What does this PR do?

Need to use parseInt before doing a comparison with <


scripts/publish.js
```diff
@@ -16,7 +16,7 @@ shell.set('+v');
 const [version, major, minor, patch] = process.version.match(
   /^v(\d+)\.?(\d+)\.?(\*|\d+)$/
 );
-if (major !== 7 || minor < 9) {
+if (parseInt(major) !== 7 || parseInt(minor) < 9) {
```
vazexqi: @chowwinston - Made an error earlier with checking.

PR #86, opened 2017-08-29
Refactor commandOutput into common class in utils

chowwinston:
### What does this PR do?
Refactor commandOutput to a common class in saalesforcedx-utils-vscode/src/cli


chowwinston:
Can you look through this first to see if this makes sense before I continue cleaning up and fixing the tests?

packages/salesforcedx-apex-debugger/src/core/breakpointService.ts
```diff
@@ -104,18 +104,25 @@ export class BreakpointService {
...
+    const result = await cmdOutput.getCmdResult(execution);
+    try {
+      const breakpointId = JSON.parse(result).result.id as string;
+      if (this.isApexDebuggerBreakpointId(breakpointId)) {
+        return Promise.resolve(breakpointId);
+      } else {
+        return Promise.reject(result);
+      }
+    } catch (e) {
+      return Promise.reject(e);
```
DatGuyJonathan: It looks you want this to catch problems with the JSON.parse and throw the "result" which came from stdout. Is that what we have in "e"?

chowwinston: e is a SyntaxError. The change to streamingService.test.ts shows what the format is. Do you prefer having the raw json so you can see what was attempted to be parsed?

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
```diff
@@ -75,19 +75,26 @@ export class SessionService {
         .build(),
       { cwd: this.project }
     ).execute();
-    const result = await this.getIdFromCommandResult(execution);
```
DatGuyJonathan: getIdFromCommandResult is private and isn't used anymore, so you'll have some tslint errors. Could you run the tslint command on apex-debugger package?

packages/salesforcedx-apex-debugger/src/core/breakpointService.ts
DatGuyJonathan: Yea, "Unexpected token n in JSON at position 2" is not very useful.

PR #90, opened 2017-09-01
Add step in-out-over commands to debugger

guw:
@W-4178710@


guw:
@DatGuyJonathan This is a first draft.

DatGuyJonathan:
Looks good so far, @guw . Just need to fix the couple of test failures. We should also add tests for the step[In|Out|Over]Request in apexDebug.ts to verify they're calling the right Step[In|Out|Over]Command.

guw:
FYI: I'm looking into the CodeCov report. The delta/diff stats seems to be off. It contains changes for files that weren't modified as part of this PR.

PR #93, opened 2017-09-01
Debugger stack trace

DatGuyJonathan:
### What does this PR do?
Show stacktrace of Apex Debugger stopped requests (up to 2).

samplestacktrace

@W-4178724@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -5,12 +5,15 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import { basename } from 'path';
 import {
   ContinuedEvent,
   DebugSession,
   Event,
   InitializedEvent,
   OutputEvent,
+  Source,
+  StackFrame,
   StoppedEvent,
   TerminatedEvent,
   Thread,
@@ -22,12 +25,11 @@ import {
   LineBreakpointsInTyperef
 } from '../breakpoints/lineBreakpoint';
 import {
+  DebuggerResponse,
   ForceOrgDisplay,
   OrgInfo,
   RunCommand,
-  StepIntoCommand,
-  StepOutCommand,
-  StepOverCommand
+  StateCommand
 } from '../commands';
 import {
   GET_LINE_BREAKPOINT_INFO_EVENT,
@@ -237,6 +239,7 @@ export class ApexDebug extends DebugSession {
     args: DebugProtocol.ContinueArguments
   ): Promise<void> {
     response.success = false;
+    response.body = { allThreadsContinued: false };
     if (args.threadId >= 0 && args.threadId < this.requestThreads.length) {
       const requestId = this.requestThreads[args.threadId];
       try {
@@ -319,11 +322,75 @@ export class ApexDebug extends DebugSession {
   protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
     const debuggedThreads: Thread[] = [];
     for (let threadId = 0; threadId < this.requestThreads.length; threadId++) {
-      debuggedThreads.push(new Thread(threadId, this.requestThreads[threadId]));
+      debuggedThreads.push(
+        new Thread(threadId, `Request ID: ${this.requestThreads[threadId]}`)
+      );
     }
     response.success = true;
     response.body = { threads: debuggedThreads };
     this.sendResponse(response);
+  }
+
+  protected async stackTraceRequest(
+    response: DebugProtocol.StackTraceResponse,
+    args: DebugProtocol.StackTraceArguments
+  ): Promise<void> {
+    response.success = false;
+    if (args.threadId < 0 || args.threadId >= this.requestThreads.length) {
+      return this.sendResponse(response);
+    }
+
+    const requestId = this.requestThreads[args.threadId];
+    try {
+      const stateResponse = await new StateCommand(
+        this.orgInfo.instanceUrl,
+        this.orgInfo.accessToken,
+        requestId
+      ).execute();
+      const stateRespObj: DebuggerResponse = JSON.parse(stateResponse);
+      const clientFrames: StackFrame[] = [];
+      if (this.hasStackFrames(stateRespObj)) {
+        const serverFrames = stateRespObj.stateResponse.state.stack.stackFrame;
+        for (let i = 0; i < serverFrames.length; i++) {
+          const sourcePath = this.myBreakpointService.getSourcePathFromTyperef(
+            serverFrames[i].typeRef
+          );
+          clientFrames.push(
+            new StackFrame(
+              i,
+              serverFrames[i].fullName,
+              sourcePath
+                ? new Source(
+                    basename(sourcePath),
+                    this.convertDebuggerPathToClient(sourcePath)
+                  )
+                : undefined,
+              this.convertDebuggerLineToClient(serverFrames[i].lineNumber),
+              0
+            )
+          );
+        }
+      }
+      response.body = { stackFrames: clientFrames };
+      response.success = true;
+    } catch (error) {
+      response.message = error;
+    }
+    this.sendResponse(response);
+  }
+
+  private hasStackFrames(response: DebuggerResponse): boolean {
+    if (
+      response &&
+      response.stateResponse &&
+      response.stateResponse.state &&
+      response.stateResponse.state.stack &&
+      response.stateResponse.state.stack.stackFrame &&
+      response.stateResponse.state.stack.stackFrame.length > 0
+    ) {
+      return true;
+    }
+    return false;
   }
 
   protected customRequest(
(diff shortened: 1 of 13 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -243,13 +252,77 @@ export class ApexDebug extends DebugSession {
   protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
     const debuggedThreads: Thread[] = [];
     for (let threadId = 0; threadId < this.requestThreads.length; threadId++) {
-      debuggedThreads.push(new Thread(threadId, this.requestThreads[threadId]));
+      debuggedThreads.push(
+        new Thread(threadId, `Request ID: ${this.requestThreads[threadId]}`)
```
vazexqi: What did we use to display in IDE 1? Could we mimic that here?
I remember that sometimes it can be a URL to an VF page, etc.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -438,6 +522,13 @@ export class ApexDebug extends DebugSession {
         this.logEvent(message);
         this.requestThreads.splice(threadId, 1);
         this.sendEvent(new ThreadEvent('exited', threadId));
```
DatGuyJonathan: When user clicks 'continue', the debug adapter sends a run command to server. That request will either finish or hit another breakpoint. If it finishes, we're supposed to send a ThreadEvent('exited') to VS Code so it knows to stop tracking that thread. For some reason, VS Code loses track of all stopped threads (apex debugger can have up to 2). The current workaround is to send stopped events for threads that are still stopped so VS Code can continue to track it. The downside is it makes VS Code ask the debug adapter for stack trace of those threads, which is wasteful because they haven't changed.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: Here's what it looks like in IDE1:

image

That URL is stored in ApexDebuggerRequest.Description, so we'd have to fetch it when we handle RequestStarted. I didn't want to do it unless we really need it because calling out to the CLI is slow. If I were to add that, I would explore just using the CLI for access token, then use request-light for everything else.

vazexqi: Could you keep a list of things that would be more snappy if we could save a call to the CLI?
I want to surface this to our CLI team and discuss it with them on whether we should use their npm module directly or use something different like request-light.

By the way, I forgot to ask this before, why did we choose to use request-light instead of reusing jsforce? jsforce should be able to send http requests as well to the different endpoints.

CC: @chowwinston since he is also doing work with integration tests and calls to jsforce. This is not a simple jsforce vs request-light, but I want to understand if we can just use one to keep the code base more consistent.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -272,14 +346,21 @@ export class ApexDebug extends DebugSession {
...
+            typerefMapping.set(info.typeref, info.uri);
           }
-          this.myBreakpointService.setValidLines(lineNumberMapping);
+          this.myBreakpointService.setValidLines(
+            lineNumberMapping,
+            typerefMapping
+          );
         }
         this.sendResponse({
           request_seq: 1,
```
vazexqi: What do the numbers request_seq (1) and seq (0) denote?
Is there a way to pull these magic numbers out into constants so it's clearer what is going on?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -371,20 +452,22 @@ export class ApexDebug extends DebugSession {
       projectPath,
       instanceUrl,
       accessToken,
-      clientInfos
+      clientInfos[0],
```
vazexqi: Similar with this one. It's not immediately clear what I'm accessing from the clientInfos[0] and clientInfos[1]

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
vazexqi: Could you include a reference to the issue so that I can keep track?
You can just paste it as plain text so that GitHub won't auto-link the two.

vazexqi:
General comment: The apexDebug.ts file is getting quite big. Would it make sense to split it out so that different commands are handled by different delegates? Or would it better to keep everything in one file for now? I can see the following delegates:

1) Handling interactions (step in, step out, step over)
2) Handling variables display 

This doesn't need to be done now. It's just something to consider as we handle more functionality

packages/salesforcedx-apex-debugger/src/commands/protocol.ts
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export interface DebuggerResponse {
```
vazexqi: Was this hand generated? Or do we have a way to auto-generate this?
If we do have some fully/semi auto way to do this, could we document it somewhere in case we need to do this again?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: Okay, I'll note what we use the CLI for and their alternatives.

It looks like RequestStarted event has that url in its description field, so I can get that into the view without fetching ApexDebuggerRequest. I'll play around with it to see if it's too crowded in the callstack view. At the very least, it'll be logged in the debug console.

I picked request-light because I don't need all the other stuff that come with jsforce. If we want to just use jsforce, I think Winston would have to move the jsforce dependency to salesforcedx-utils-vscode so both salesforcedx-apex-debugger & salesforcedx-sobjects can use it.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: `https://github.com/Microsoft/vscode-debugadapter-node/issues/130`

packages/salesforcedx-apex-debugger/src/commands/protocol.ts
DatGuyJonathan: It was semi-automated. I documented it in an internal doc, but I can also summarize it the source.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: Here's what it looks like with the request's description in the callstack view:

callstacknarrow

callstackwide

The callstack view is restricted to the left hand side, so the description is not as useful. I'm going to stick with just showing the request ID and outputting the request description in debug console for now. The latter is taken care of in https://github.com/forcedotcom/salesforcedx-vscode/pull/94.

vazexqi: @chowwinston's jsforce dependency is currently in his new salesforcedx-sobject module. See #96. 

I think we can decide on whether to consolidate in a different PR.
For now, you can continue to use request-light first. 
Don't let this issue block you for now.

DatGuyJonathan:
Regarding the size of apexDebug.ts, the general rule is to separate logic that doesn't require vscode-debugadapter to their own classes. Other VS Code debug adapters follow the same approach and their debug adapter files are also quite big. Currently, we have separate classes that work with Apex Debugger sobjects, Streaming API, and run & step commands. We could have a separate class that handles debugger events which would need a reference to the DebugSession object so it can send events to VS Code, and that would eliminate a couple hundreds lines from apexDebug.ts. We can definitely look into this soon in another PR.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -249,13 +259,77 @@ export class ApexDebug extends DebugSession {
...
+  }
+
+  private hasStackFrames(response: DebuggerResponse): boolean {
+    if (
+      response &&
+      response.stateResponse &&
+      response.stateResponse.state &&
+      response.stateResponse.state.stack &&
+      response.stateResponse.state.stack.stackFrame &&
+      response.stateResponse.state.stack.stackFrame.length > 0
```
guw: Sight, I wish TS would move a bit faster (https://github.com/Microsoft/TypeScript/issues/16).

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: @DatGuyJonathan Did you see the last reply? They closed https://github.com/Microsoft/vscode/issues/33839  as "as-designed". Is the additional stop event still necessary with `allThreadsContinued: false`?

DatGuyJonathan: Yes, I already have that in my handling of continuedRequest. I'll keep the additional stop event for now and continue the conversation with them.

PR #96, opened 2017-09-05
Winston/sobject fetch

chowwinston:
### What does this PR do?
Adds sobject lists and describe calls to prepare for generating faux apex classes


packages/salesforcedx-sobjects/test/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// tslint:disable-next-line:no-var-requires
+const testRunner = require('@salesforce/salesforcedx-utils-vscode/out/src/test/testrunner');
```
chowwinston: Will update this to use mocha

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -2,8 +2,7 @@
   "preview": true,
   "name": "salesforcedx-vscode-apex",
   "displayName": "Apex Code Editor for Visual Studio Code",
-  "description":
```
chowwinston: tslint has the one-line rule set to true so I think someone's linter isn't set up properly.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -28,6 +29,7 @@ async function createServer(
 ): Promise<Executable> {
   try {
     deleteDbIfExists();
+    generateFauxClasses();
```
chowwinston: Placeholder for now to be able to invoke the generation until we figure out where to put it.

packages/salesforcedx-sobjects/README.md
```diff
@@ -0,0 +1,5 @@
+# salesforcedx-sobjects
```
vazexqi: Probably want to rename this to salesforcedx-sobjects-faux-generator to be more descriptive.

packages/salesforcedx-sobjects/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "Fetches sobjects and generates their faux apex class to be used for Apex Language Server",
+  "version": "40.8.0",
+  "publisher": "salesforce",
+  "license": "BSD-3-Clause",
+  "engines": {
+    "vscode": "^1.15.0"
+  },
+  "dependencies": {
+    "@salesforce/salesforcedx-utils-vscode": "40.8.0",
+    "jsforce": "1.7.1"
```
vazexqi: Ensure that this is the same version used by the CLI. And then also add @types/jsforce to the devDependencies section.

packages/salesforcedx-sobjects/src/describe/sobjectDescribe.ts
```diff
@@ -0,0 +1,210 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import {
+  CliCommandExecutor,
+  CommandOutput,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+
+export interface SObject {
```
vazexqi: How did you arrive at this interface? Is it complete? Is it partially generated?
How would we update this in the future?

packages/salesforcedx-sobjects/src/describe/sobjectDescribe.ts
```diff
@@ -0,0 +1,210 @@
...
+  quickActions: string;
+  layouts: string;
+  sobject: string;
+}
+
+export interface DescribeSObjectResult {
+  result: SObject;
+}
+
+export class SObjectDescribe {
```
vazexqi: Just to confirm, the goal is to isolated the SObjectDescribe functionality to this class, right?
If we decide to to do a call to the jsforce module directly or use the NPM CLI directly, we can make this change here and all consumers are not affected?

The contract is `describeSObject(string, string, username): Promise <SObject>`?

packages/salesforcedx-sobjects/src/describe/sobjectDescribe.ts
```diff
@@ -0,0 +1,210 @@
...
+  public async describeSObject(
+    projectPath: string,
+    type: string,
```
vazexqi: Could you use string enums (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html) instead of string? I think the type is bounded to just 'standard' and 'custom'?

packages/salesforcedx-sobjects/src/generator/fauxClassGenerator.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as fs from 'fs';
+import * as path from 'path';
+import { SObjectDescribe } from '../describe';
+
+export class FauxClassGenerator {
```
vazexqi: This is also another contract point for faux class generation?

packages/salesforcedx-sobjects/test/integration/config/project-scratch-def.json
```diff
@@ -0,0 +1,8 @@
+{
+  "orgName": "Your Company",
+  "edition": "Developer",
+  "orgPreferences": {
+    "enabled": ["Translation"],
```
vazexqi: Why did you choose to enabled the "Translation" preference, in particular?

packages/salesforcedx-sobjects/test/integration/sobjectDescribeTest.ts
```diff
@@ -0,0 +1,91 @@
...
+import { expect } from 'chai';
+import * as jsforce from 'jsforce';
+import { SObjectDescribe } from '../../src/describe/sObjectDescribe';
+
+describe('Fetch sObjects', function() {
+  // tslint:disable-next-line:no-invalid-this
+  this.timeout(10000);
+  let username: string;
+  const sobjectdescribe = new SObjectDescribe();
+  const scratchDefFilePath =
```
vazexqi: Per our conversation offline, take a look at using `path.join` and also `__dirname`.

packages/salesforcedx-sobjects/test/integration/sobjectDescribeTest.ts
```diff
@@ -0,0 +1,91 @@
...
+    '/test/integration/config/project-scratch-def.json';
+  before(async function() {
+    const execution = new CliCommandExecutor(
+      new SfdxCommandBuilder()
+        .withArg('force:org:create')
```
vazexqi: How does this test pass locally? I think this can pass on the autobuilds because we created a default deb hub. See https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/.appveyor.yml#L29 and https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/.travis.yml#L57

But how do we run this test locally?

packages/salesforcedx-sobjects/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "sourceMap": true,
+    "declaration": true,
+    "moduleResolution": "node",
+    "noImplicitAny": true,
+    "rootDir": ".",
+    "outDir": "out",
+    "preserveConstEnums": true,
+    "strict": true
+  },
+  "exclude": ["node_modules", ".vscode-test", "out"]
```
vazexqi: Since you are not a vscode module, you won't have a .vscode-test folder. So you can take that out.

packages/salesforcedx-vscode-apex/package.json
vazexqi: What does prettier formatter say? I don't think ts-lint should be running on the .json file. It is not running for me.

packages/salesforcedx-vscode-apex/src/languageServer.ts
vazexqi: Please remove this line. We cannot have it accidentally try to generate the faux classes on start up now. That will be too slow. We will think of a different gating mechanism later.

Also, you cannot just call the generateFauxClasses here since it's possible that you don't even have a scratch org set up yet. So the call would fail.

packages/salesforcedx-sobjects/src/describe/sobjectDescribe.ts
chowwinston: I used json2ts with the results of the sobjectDescribe call. It's the complete json output, but with result renamed to SObject. We only need the name and Field[] values for the purposes of generation.

vazexqi: @DatGuyJonathan - Is json2ts the one that you used too for your debugger? Sorry, I know I asked a similar question earlier but the search feature in the PR in GitHub didn't allow me to find it.

packages/salesforcedx-sobjects/src/describe/sobjectDescribe.ts
chowwinston: Correct

packages/salesforcedx-sobjects/src/generator/fauxClassGenerator.ts
chowwinston: Yes

packages/salesforcedx-sobjects/test/integration/config/project-scratch-def.json
chowwinston: I just copied it from one of my existing scratch-def files. I can remove it.

packages/salesforcedx-sobjects/test/integration/sobjectDescribeTest.ts
chowwinston: I have a default dev hub enabled locally. I think we should make that a requirement since it's hard to automate the authorization of a new dev hub.

packages/salesforcedx-sobjects/src/generator/fauxClassGenerator.ts
chowwinston: Yes

packages/salesforcedx-sobjects/test/integration/sobjectDescribeTest.ts
vazexqi: Correct. In that case, do you want to add it to the doc/testing.md? Since if your changes go in, it would no longer possible to just use `npm run test` if you don't have a dev hub set up.

packages/salesforcedx-vscode-apex/src/languageServer.ts
chowwinston: I have it as an async function so it's not blocking startup. Would you be ok with it behind an environment variable? Or do you want it gone completely until we figure out how to expose it?

packages/salesforcedx-sobjects-faux-generator/README.md
```diff
@@ -0,0 +1,5 @@
+# salesforcedx-sobjects
```
vazexqi: Minor. Rename this title as well.

docs/tests.md
```diff
@@ -6,6 +6,9 @@ describes them and give pointers on how to run/debug them.
 These tests are described in the order in which we prefer to write them. Always
 prefer unit tests, integration tests, and system tests, in that order.
 
+Ensure that you have set a default dev hub, through one of the `sfdx force:auth:*` commands,
```
vazexqi: @GaryFNC, @guw, @Sweetman, @DatGuyJonathan - Take note of this. This means that you need to have a default scratch org set authenticated against (doesn't need to be production) if you want to run *all* tests. We are starting with the sobject tests, but eventually as more integration tests are written, those will be affected as well.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  'objects'
+);
+
+//const util = new TestUtil();
+const sobjectdescribe = new SObjectDescribe();
+
+// tslint:disable:no-unused-expression
+describe('Fetch sObjects', function() {
+  // tslint:disable-next-line:no-invalid-this
+  this.timeout(40000);
```
chowwinston: I'll take this out with next commit

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    expect(cmdOutput.fields.length).to.be.greaterThan(0);
+    const customField = cmdOutput.fields[cmdOutput.fields.length - 1];
+    expect(customField.custom).to.be.true;
+    expect(customField.precision).to.be.equal(18);
+    expect(customField.scale).to.be.equal(0);
+    expect(customField.name).to.be.equal('MyCustomField__c');
+  });
+
+  it('Should be able to call describeSObject on standard object', async function() {
+    // const cmdOutput = await sobjectdescribe.describeSObject(
```
chowwinston: The command output needs a buffer for standard objects. The amount of json is too large and gets truncated. I'll do this as part of adding explicit refresh command.

packages/salesforcedx-sobjects-faux-generator/test/integration/integrationTestUtil.ts
```diff
@@ -0,0 +1,146 @@
...
+    { cwd: path.join(process.cwd(), projectName) }
+  ).execute();
+  const cmdOutput = new CommandOutput();
+  const result = await cmdOutput.getCmdResult(execution);
+  const source = JSON.parse(result).result.pushedSource;
+  return Promise.resolve(source);
+}
+
+export async function createPermissionSet(
+  permissionSetName: string,
```
chowwinston: Still have FLS issues. Using source:push didn't work.

chowwinston:
@guw the tests are failing in appveyor with 
`Error: the string "{\"message\":\"ENOENT: no such file or directory, open 'C:\\\\projects\\\\salesforcedx-vscode\\\\packages\\\\salesforcedx-sobjects-faux-generator\\\\project_1504830925511\\\\devhub.key'\",\"status\":1,\"name\":\"Error\",\"warnings\":[]}\n" was thrown, throw an Error :)`

Also some parts have 
`lerna ERR! execute npm ERR! Please try running this command again as root/Administrator.`

docs/tests.md
vazexqi: @chowwinston -  Can you actually put the actual command and parameters that you use here? That would help standardize what we use. And we can also compare this to what the appveyor and travis files are doing.

packages/salesforcedx-sobjects-faux-generator/test/integration/integrationTestUtil.ts
```diff
@@ -11,9 +11,25 @@ import {
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 import childProcess = require('child_process');
+import * as fs from 'fs';
 import * as path from 'path';
 import * as util from 'util';
 
+// Used only for CI purposes. Must call delete if you call create
+export function createCIKey(keyLocation: string) {
+  const SERVER_KEY = process.env.SFDX_CI_DEVHUB_JWTKEY;
```
chowwinston: For local development/testing, this environment variable is not necessary as long as you have a default devhub set.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -15,6 +15,7 @@ import {
 import * as util from './integrationTestUtil';
 
 const PROJECT_NAME = `project_${new Date().getTime()}`;
+const keyLocation = path.join(process.cwd(), PROJECT_NAME, 'server.key');
```
chowwinston: After using jwt, further org commands expect a file called server.key in the directory. We can obfuscate the filename if we change the oAuth options, but we delete the key in the after() anyways.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -25,40 +26,48 @@ const SIMPLE_OBJECT_DIR = path.join(
...
       CUSTOM_OBJECT_NAME,
       CUSTOM_FIELD_FULLNAME,
       username
     );
+
     await util.assignPermissionSet(permSetName, username);
   });
 
   after(function() {
+    util.deleteCIKey(keyLocation);
```
chowwinston: Not strictly required since the key will also get deleted through the call below.

packages/salesforcedx-utils-vscode/src/cli/commandOutput.ts
```diff
@@ -9,13 +9,22 @@ import { CommandExecution } from './commandExecutor';
 
 export class CommandOutput {
   public async getCmdResult(execution: CommandExecution): Promise<string> {
+    let buffer = '';
     return new Promise<
       string
     >((resolve: (result: any) => void, reject: (reason: string) => void) => {
       execution.processExitSubject.subscribe(data => {
         if (data != undefined && data.toString() === '0') {
```
vazexqi: @GaryFNC @chowwinston - I _think_ I know why we are seeing the hangs or incomplete data. If you do it this way, I suspect that you are only draining data from `execution.stdoutSubject.subscribe` at the end after you have received an `execution.processExitSubject`. However, by that time, the stdout buffer has already filed up and the process is stuck from not draining it.

I know I had suggested serializing the result so that we only get them on processExit but this probably prevents the  data from being drained.

vazexqi: As a quick sanity check, https://nodejs.org/api/child_process.html#child_process_event_exit says:

>The 'exit' event is emitted after the child process ends. If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null. **Note that when the 'exit' event is triggered, child process stdio streams might still be open.**

And, https://nodejs.org/api/child_process.html#child_process_event_close says:

>**The 'close' event is emitted when the stdio streams of a child process have been closed.** This is distinct from the 'exit' event, since multiple processes might share the same stdio streams.

So, instead of listening on the exit event, we might want to listen to the _close_ event when data from the streams are important.

So, things to try:

1. Don't serialize the promise to only read from stdout and stderr if the exit signal has been received. Instead, continue reading and buffer the chunks.
1. Instead of listening to _exit_, listen to _close_ to know when all data from stdio have been received, then you can proceed to parse the JSON, once and for all (you no longer need the weird trick of parsing each time).

vazexqi: Actually a bit of spelunking shows that the serialization technique was done earlier as part of https://github.com/forcedotcom/salesforcedx-vscode/commit/8c3cebda40148e30f451aa471309a9a5cca6b971#diff-222256c5627f520a77b31ab383f6ffa1. When you have small responses, this doesn't present a problem, but when you have larger responses, this becomes an issue.

PR #98, opened 2017-09-06
Fix error messages with starting & stopping debugger session

DatGuyJonathan:
### What does this PR do?
When an Apex Debugger session is terminated outside of VS Code, we handle a SessionTerminated event by printing the event description to debug console. Now, we also show an error top-level notification:
sessionterminatednotification

When the scratch org or user does not have Debug Apex perm, we show an error top-level notification that suggests they need both org and user perms. We cannot use "sfdx force:schema:sobject" commands to individually detect for org & user perm. The command sobject:list will return ApexDebuggerSession with & without Debug Apex org perm because its org access is only restricted to AuthorApex. The command sobject:describe on ApexDebuggerSession will return the same thing with & without Debug Apex org perm.
noaccesstoapexdebuggersessionobject

@W-4241196, W-4241210@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -462,14 +462,15 @@ export class ApexDebug extends DebugSession {
       response.success = false;
       const errorObj = JSON.parse(error);
       if (errorObj && errorObj.message) {
+        const errorMessage: string = errorObj.message;
         if (
-          errorObj.message.indexOf(
+          errorMessage.includes(
             'entity type cannot be inserted: Apex Debugger Session'
-          ) !== -1
+          )
         ) {
           response.message = nls.localize('session_no_entity_access_text');
         } else {
-          response.message = errorObj.message;
+          response.message = errorMessage;
         }
         if (errorObj.action) {
           this.errorToDebugConsole(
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -462,7 +462,15 @@ export class ApexDebug extends DebugSession {
       response.success = false;
       const errorObj = JSON.parse(error);
       if (errorObj && errorObj.message) {
-        response.message = errorObj.message;
+        if (
+          errorObj.message.indexOf(
```
vazexqi: Let's use the new [`string.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). Seems cleaner than checking for -1.

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
```diff
@@ -19,6 +19,8 @@ export const messages = {
   command_error_help_text: 'Command returned the following error',
   session_language_server_error_text:
     'Apex language server could not provide information about valid breakpoints.',
+  session_no_entity_access_text:
+    "Either your org or user doesn't have the permission to debug Apex.",
```
vazexqi: I think this comes directly from the Force.com IDE? Should be kosher then.

vazexqi:
LGTM. Thanks for doing this. This makes the errors much easier to understand.

PR #99, opened 2017-09-07
Handle events Debug, SystemGack, ApexException, & a few more.

DatGuyJonathan:
### What does this PR do?
Handle some Apex Debugger events that print to VS Code debug console.

@W-4179020, W-4179047@

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -13,8 +13,8 @@
     "@salesforce/salesforcedx-utils-vscode": "40.9.0",
     "faye": "1.1.2",
     "request-light": "0.2.1",
-    "vscode-debugadapter": "1.22.0",
-    "vscode-debugprotocol": "1.22.0"
+    "vscode-debugadapter": "1.23.0",
+    "vscode-debugprotocol": "1.23.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
@@ -29,7 +29,7 @@
     "sinon": "^2.3.6",
     "typescript": "2.4.0",
     "vscode": "1.1.4",
-    "vscode-debugadapter-testsupport": "1.22.0"
+    "vscode-debugadapter-testsupport": "1.23.0"
   },
   "scripts": {
     "vscode:package": "npm prune --production",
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-apex-debugger/package.json
```diff
@@ -13,8 +13,8 @@
     "@salesforce/salesforcedx-utils-vscode": "40.8.0",
     "faye": "1.1.2",
     "request-light": "0.2.1",
-    "vscode-debugadapter": "1.22.0",
-    "vscode-debugprotocol": "1.22.0"
+    "vscode-debugadapter": "1.23.0",
```
DatGuyJonathan: Upgraded so I can use source location in a VS Code debug adapter Output event.
`https://github.com/Microsoft/vscode-debugadapter-node/issues/94`

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -588,9 +616,42 @@ export class ApexDebug extends DebugSession {
...
+      const regExp: RegExp = /^(.*)\[(\d+)\]\|/;
+      const matches = message.sobject.Description.match(regExp);
+      if (matches && matches.length === 3) {
+        const possibleClassName = matches[1];
+        const possibleClassLine = parseInt(matches[2]);
+        const possibleSourcePath = this.myBreakpointService.getSourcePathFromPartialTyperef(
+          possibleClassName
+        );
+        if (possibleSourcePath) {
+          eventDescriptionSourceFile = new Source(
```
DatGuyJonathan: Apex Debugger event type 'Debug' has a description in the form of '<classname>[line number]|<actual message>'. The debug protocol has recently supported including a source location in an output event, so I'm parsing that stuff here. The idea is to link a debug console message so user can quickly navigate to the Apex source. The VS Code UI doesn't actually support it yet, but including it now is harmless because it's ignored. The frontend support seems to be planned for this month: `https://github.com/Microsoft/vscode/issues/24490`

packages/salesforcedx-apex-debugger/src/core/breakpointService.ts
```diff
@@ -62,6 +62,16 @@ export class BreakpointService {
     return this.typerefMapping.get(typeref);
   }
 
+  public getSourcePathFromPartialTyperef(
```
guw: This method (and associated logic) looks like potential candidate for future refactoring/higher re-use outside of breakpoint service.

PR #101, opened 2017-09-07
Add SLDS Linter

engai:
### What does this PR do?
Adds SLDS Linter feature to Lightning Package

@W-4267770@


packages/salesforcedx-vscode-lightning/README.md
```diff
@@ -1,5 +1,5 @@
 # salesforcedx-vscode-lightning
```
vazexqi: @ruthemmanuelle - Could you review this README for @ayeshakmaz and @engai?

vazexqi:
@engai @ayeshakmaz Could you take a look at your test to see why it's failing on windows? Could it be you are looking for a particular path and that you are using `/` instead of `\`?

engai:
@vazexqi Yes. It's trying to open `./assets/sfdx-simple`. How would you suggest fixing this to work on Mac, Linux & Windows?

vazexqi:
Try using https://nodejs.org/api/path.html#path_path_joint paths. You might want to also check to see that you are not using / in your code and not just the tests.

engai:
Paths look correct at the moment (some are `glob` patterns). The extension and tests run when tested in vscode on windows. However, they aren't running with the test script on Windows. Any ideas on what might be happening here?

vazexqi:
@engai - Could you get a Windows VM and run the tests locally or add more logging for when the test fails?

The error is 

```
lerna ERR! execute   1) SLDS Deprecated Class Name Should create deprecatedClassName command:
lerna ERR! execute      AssertionError: expected [ Array(693) ] to include 'deprecatedClassName'
lerna ERR! execute       at Object.<anonymous> (C:\projects\salesforcedx-vscode\packages\salesforcedx-slds-linter\test\deprecatedClassName.test.ts:27:28)
lerna ERR! execute       at Generator.next (<anonymous>)
lerna ERR! execute       at fulfilled (C:\projects\salesforcedx-vscode\packages\salesforcedx-slds-linter\out\test\deprecatedClassName.test.js:10:58)
lerna ERR! execute       at <anonymous>
lerna ERR! execute 
lerna ERR! execute 

``` but it's not clear why it can't find what it's looking for.

**Comment by engai** [comment]
@vazexqi the test is essentially looking for our command within the workspace command list. This will only be there when the extension runs. Not sure what else to log in this case. The extension functions and tests pass when ran in VSCode on our VM, but the extension doesn't seem to run when ran from the test script. 

The correct workspace is opened, but no syntax highlighting or linting appears.

File: packages/salesforcedx-slds-linter/test/deprecatedClassName.test.ts [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+      res = await vscode.workspace.findFiles(path.join('**', '*.cmp'));
+      await vscode.workspace.openTextDocument(res[0]).then(
+        document => vscode.window.showTextDocument(document)
+      );
+    }
+  });
+
+  it('Should create deprecatedClassName command', async () => {
+    const commandList = await vscode.commands.getCommands(true);
+    expect(commandList).to.include('deprecatedClassName');
```
vazexqi: @engai I guess you can log here to see what is the list of commandList on the autobuild and how that differs from the one you have locally. I am not yet sure why the test will fail since this seems to be a simple test.

engai:
screen shot 2017-09-11 at 2 38 54 pm

When working on mac and within vscode on windows, it's logging 45 more commands. These include ours and commands from git and sfdx:force. When running on the test script however, it's not logging these commands.

vazexqi:
@engai That is helpful. So we can at least narrow it down to the command not loading.

How are you running the tests locally? How does it differ from the appveyor.yml and travis.yml files? Could you be running it from a different directory?

On that note, I just noticed that your command doesn't follow the convention. Could you rename your command to `sfdx.force.lightning.slds.deprecated.class.fix` (or something similar)? We need to have some form of a namespace to ensure that command names do not collide and just `deprecatedClassName` is too generic.

engai:
@vazexqi To run via test script (Does not pass) - Following these steps in the `salesforce-vscode` directory:

1. `npm run bootstrap`
2. `npm run compile`
3. `npm run test` (Sometimes ran from `/packages/salesforce-slds-linter` to ignore the other packages' tests)

Running in VSCode (Passes):

1. `Run Task > Bootstrap` if changing anything in dependencies
2. `Launch Salesforce DX VS Code SLDS Linter Tests` to run tests in ext. dev host

On a side note, currently unable to run tests via the script (was working yesterday). Seems like the script can't find a VSCode executable:

screen shot 2017-09-11 at 4 11 18 pm

Have you run into this problem? Looks similar to https://github.com/Microsoft/vscode-extension-vscode/issues/30

As for the command name, thanks for the heads up. I'll go ahead and change that.

vazexqi:
I just ran the tests locally and I'm able to produce the issue.

1. `git clone git://github.com/forcedotcom/salesforcedx-vscode.git`
1. `cd salesforcedx-vscode`
1. `npm install`
1. `npm run compile`
1. `npm run test:without-system-tests` (we don't need to run the UI tests to repro the issue)

packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
```diff
@@ -13,12 +13,13 @@ describe('SLDS Deprecated Class Name', () => {
   let res: vscode.Uri[];
 
   before(async () => {
-
     if (vscode.workspace.rootPath) {
-      res = await vscode.workspace.findFiles(path.join('**', '*.cmp'));
-      await vscode.workspace.openTextDocument(res[0]).then(
-        document => vscode.window.showTextDocument(document)
+      res = await vscode.workspace.findFiles(
+        path.join('**', 'DemoComponent.cmp')
```
vazexqi: You should be precise and look exactly for the .cmp you know has the deprecatedclassnames.

vazexqi:
@engai - Take a look at my latest commit where I moved your tests to the salesforcedx-vscode-lightning package. The tests need to be there since it depends on an actual activation event.

It runs when you do it from VS Code itself since you are loading all the packages (see .vscode/launch.json). But, when you run the tests from the command line, the tests only load the extensions for that package.

engai:
Ah that makes sense. The tests aren't passing within VSCode, but i'll work on that within the lightning tests

.vscode/launch.json
```diff
@@ -124,7 +124,26 @@
...
       "stopOnEntry": false,
       "sourceMaps": true,
-      "outFiles": ["${workspaceRoot}/packages/*/out/**/*.js"],
+      "outFiles": [
+        "${workspaceRoot}/packages/*/out/**/*.js"
+      ],
+      "preLaunchTask": "Compile"
+    },
+    {
+      "name": "Launch Salesforce DX VS Code SLDS Linter Tests",
```
vazexqi: @engai - This one probably needs to be updated since I removed the ${workspaceRoot}/packages/salesforcedx-slds-linter/assets/sfdx-simple

packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+describe('SLDS Deprecated Class Name', () => {
+  let res: vscode.Uri[];
+
+  before(async () => {
+    if (vscode.workspace.rootPath) {
+      res = await vscode.workspace.findFiles(
+        path.join('**', 'DemoComponent.cmp')
+      );
+      await vscode.workspace
```
vazexqi: @engai Since you open the file, you should probably have an after that closes it too (just to be tidy). So far it's not causing problems but as we have more tests in the future, I think it's good for each test to clean up after itself.

engai:
@vazexqi Wasn't sure if it would affect the end-to-end tests, i'll remove mine and add the .cmp file to system-test. 

Though .DS_Store was in the .git_ignore, i'll remove those as well.

vazexqi:
In one of my previous commits I already added the DemoComponent to the system-tests/assets folder. You should be able to see it.

ayeshakmaz:
@vazexqi I'm looking in system-tests/assets, but not seeing that moved DemoComponent

vazexqi:
@ayeshakmaz - Crap. I must have forgotten to commit it. One moment.

vazexqi:
@ayeshakmaz and @engai - I've added the DemoComponent.cmp now.

packages/salesforcedx-slds-linter/src/messages/i18n.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Conventions:
+ * _message: is for unformatted text that will be shown as-is to
+ * the user.
+ * _text: is for text that will appear in the UI, possibly with
+ * decorations, e.g., $(x) uses the https://octicons.github.com/ and should not
+ * be localized
+ *
+ * If ommitted, we will assume _message.
+ */
+export const messages = {
```
vazexqi: @ruthemmanuelle - Could you take a look at these messages? There is a screenshot that shows where there are used in this PR.

vazexqi:
@Sweetman  - Here's another case where the travis builds just hangs.

packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
engai: Added an after to close the editor. Looks like closeFile is deprecated.

engai:
@vazexqi I've added 3 tests to our slds-linter package. Let us know if there is anything else we should add. Appveyor build looks good as well.

packages/salesforcedx-slds-linter/test/server.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  const connection = {
+    sendDiagnostics: (obj: any) => {
+      args = obj;
+    }
+  };
+
+  it('Should send correct string', () => {
+
+    validateTextDocument('\n class"slds-button--brand" \n', 'uri', connection);
+    if (args) {
```
vazexqi: @engai I think you should also assert that `args` is not null in the tests.
Because if it turns out that `args` is null, then if will skip the if(args) {...} and then the test will pass, right?

packages/salesforcedx-vscode-lightning/README.md
```diff
@@ -28,10 +28,10 @@ Currently, Visual Studio Code extensions are not signed or verified on the Micro
...
 * Salesforce Lightning Design System Linter
-   * Detects deprecated BEM syntax `'--'` for Salesforce Lightning Design System class names in html files
+   * Detects deprecated BEM syntax (`--`) for Salesforce Lightning Design System class names in HTML files
    * Warning message displays on hover
-   * Code actions available on click
-   * Linter won't run if SLDS is included as a static resource in your project
+   * Code actions are available on click
+   Note: The linter won't run if SLDS is included as a static resource in your project.
 
    ![SLDS Linter detecting deprecated '--' class name syntax](https://raw.githubusercontent.com/forcedotcom/salesforcedx-vscode/ayesha/add-slds-linter/packages/salesforcedx-vscode-lightning/images/lightning_slds.png)
```
ruthemmanuelle: We should take a new screenshot that includes the updated messages. We'll also need to update this path to use /develop/ instead of /ayesha/add-slds-linter/, but if we do that before merging in this PR the image won't display. It would be good if the resolution and text size of the screenshot approximately matched the other screenshots in this README.

packages/salesforcedx-slds-linter/src/messages/i18n.ts
```diff
@@ -17,15 +17,15 @@
...
-  `Fix: %s`,
+  `Fix this problem: %s`,
   fix_same:
-  `Fix all: %s`,
+  `Fix all instances of: %s`,
   fix_all:
   'Fix all auto-fixable problems',
   fix_error:
-  'Failed to apply SLDS Validator fixes to the document',
+  'Failed to apply SLDS Validator fixes to the document.',
```
ruthemmanuelle: Do we have any idea what would have caused this problem? Can we suggest a solution? Also, I added a period, but if it displays in the same format as the other messages in this section we can remove the period. VS Code seems to use sentence-case capitalization and no end punctuation for these contextual messages.

engai: It will only fail if the edit function doesn't successfully run. This wouldn't be a problem that the end user can fix. One thing we can suggest is filing an issue on Github with us.

engai: It doesn't display along with the other messages, so the period should be okay

packages/salesforcedx-slds-linter/test/server.test.ts
engai: Right. Pushed a fix to assert when null

engai:
@ruthemmanuelle - I've added a new screenshot and changed the image path to point to develop. Let me know if you need me to make any other changes

vazexqi:
@engai @ayeshakmaz Looks good to me. Can you rebase onto develop and then we will do one final pass?

packages/salesforcedx-slds-linter/src/server/index.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  return result;
+});
+
+function allCodeActions(result: Command[], uri: string) {
+  const fixAllEdits: TextEdit[] = [];
+
+  if (activeDiagnostics.length > 1) {
+    for (const codeAction of activeDiagnostics) {
+      const codeStr = <string>codeAction.code;
```
vazexqi: Cosmetics: Please use `codeAction.code as string` for casting. See http://acdcjunior.github.io/typescript-cast-object-to-other-type-or-instanceof.html for why we want to avoid using `<>`

packages/salesforcedx-slds-linter/src/server/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+}
+
+function sameCodeActions(result: Command[], uri: string, problem: string) {
+  const fixSameEdits: TextEdit[] = [];
+  let codeMessage;
+
+  if (activeDiagnostics.length > 1) {
+    for (const codeAction of activeDiagnostics) {
+      const code = <string>codeAction.code;
```
vazexqi: Same here.

packages/salesforcedx-slds-linter/src/server/index.ts
```diff
@@ -0,0 +1,207 @@
...
+      if (code[0] === problem) {
+        const codeStr = <string>codeAction.code;
```
vazexqi: And here.

packages/salesforcedx-slds-linter/src/server/index.ts
```diff
@@ -0,0 +1,207 @@
...
+          newText: replacementStr
+        });
+
+        switch (problem) {
+          case '0': {
+            codeMessage = nls.localize('general_deprecated_class_name');
+            break;
+          }
+          default: {
+            codeMessage = 'same problems';
```
vazexqi: Is there a reason this one is not done using `nls.localize`?

packages/salesforcedx-vscode-lightning/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import * as languageServer from '@salesforce/salesforcedx-slds-linter/out/src/client';
+import * as vscode from 'vscode';
+
+export function activate(context: vscode.ExtensionContext) {
+  console.log('SFDX SLDS Linter Extension Activated');
+  vscode.workspace.findFiles('**/staticresources/*.resource').then(
+    // all good
+    (result: vscode.Uri[]) => {
+      for (let i = 0; i < result.length; i++) {
```
vazexqi: Cosmetic: Is there a reason we use `for(let i = 0...)` instead of the for(const a of blah)?

packages/salesforcedx-slds-linter/src/client/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      module: serverModule,
+      transport: TransportKind.ipc,
+      options: debugOptions
+    }
+  };
+
+  // Options to control the language client
+  const clientOptions: LanguageClientOptions = {
+    // Register the server for plain text documents
+    documentSelector: ['plaintext', 'html'],
```
vazexqi: @ayeshakmaz - I think we only want HTML, right? No need for it to scan plaintext?

ayeshakmaz: Yeah, we can take out the plain text option

vazexqi:
Merging this in! Thanks @engai and @ayeshakmaz

PR #102, opened 2017-09-07
James/local and source changes

Sweetman:
### What does this PR do?
Add commands to the palette for local source and remote source changes
@W-4197703@, @W-4197715@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -2,18 +2,24 @@
   "running_tasks_title_text": "Running Tasks",
 
   "force_auth_web_login_authorize_dev_hub_text": "SFDX: Authorize a Dev Hub",
-  "force_org_create_default_scratch_org_text": "SFDX: Create a Default Scratch Org...",
+  "force_org_create_default_scratch_org_text":
+    "SFDX: Create a Default Scratch Org...",
   "force_org_open_default_scratch_org_text": "SFDX: Open Default Scratch Org",
-  "force_source_pull_default_scratch_org_text": "SFDX: Pull Source from Default Scratch Org",
-  "force_source_push_default_scratch_org_text": "SFDX: Push Source to Default Scratch Org",
-  "force_source_status_text": "SFDX: View All Changes (Local and in Default Scratch Org)",
+  "force_source_pull_default_scratch_org_text":
+    "SFDX: Pull Source from Default Scratch Org",
+  "force_source_push_default_scratch_org_text":
+    "SFDX: Push Source to Default Scratch Org",
+  "force_source_status_text":
+    "SFDX: View All Changes (Local and in Default Scratch Org)",
   "force_apex_test_run_text": "SFDX: Invoke Apex Tests...",
   "cancel_sfdx_command_text": "SFDX: Cancel Active Command",
   "force_apex_class_create_text": "SFDX: Create Apex Class",
-  "force_visualforce_component_create_text": "SFDX: Create Visualforce Component",
+  "force_visualforce_component_create_text":
+    "SFDX: Create Visualforce Component",
   "force_visualforce_page_create_text": "SFDX: Create Visualforce Page",
   "force_lightning_app_create_text": "SFDX: Create Lightning App",
   "force_lightning_component_create_text": "SFDX: Create Lightning Component",
   "force_lightning_event_create_text": "SFDX: Create Lightning Event",
-  "force_lightning_interface_create_text": "SFDX: Create Lightning Interface"
+  "force_lightning_interface_create_text": "SFDX: Create Lightning Interface",
+  "force_source_status_local_text": "SFDX: View Local Changes"
 }
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -21,5 +21,6 @@
   "force_lightning_component_create_text": "SFDX: Create Lightning Component",
   "force_lightning_event_create_text": "SFDX: Create Lightning Event",
   "force_lightning_interface_create_text": "SFDX: Create Lightning Interface",
-  "force_source_status_local_text": "SFDX: View Local Changes"
+  "force_source_status_local_text": "SFDX: View Local Changes",
+  "force_source_status_remote_text": "SFDX: View Changes in Default Scratch Org"
 }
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -71,6 +72,11 @@ export class ChannelService {
       } else {
         this.channel.appendLine(nls.localize('channel_end'));
       }
+      this.channel.appendLine('');
     });
   }
+
+  public showChannelOutput() {
```
Sweetman: Exposed this so the commands can open the channel output

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -207,6 +207,7 @@ export abstract class SfdxCommandletExecutor<T>
     }).execute(cancellationToken);
 
     channelService.streamCommandOutput(execution);
+    channelService.showChannelOutput();
```
Sweetman: This change will make all the commands open the channel output. I think it would be nice to show this on top of the notification that shows at the top for all commands but I can change this to only affect these commands for now.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -54,6 +54,16 @@ function registerCommands(): vscode.Disposable {
     'sfdx.force.source.status',
     forceSourceStatus
   );
+  const forceSourceStatusLocalCmd = vscode.commands.registerCommand(
+    'sfdx.force.source.status.local',
+    forceSourceStatus,
+    { flag: '--local' }
```
Sweetman: registerCommand takes a third parameter and binds it to the `this` context of forceSourceStatus.

packages/salesforcedx-vscode-core/src/commands/commands.ts
Sweetman: @vazexqi and I also discussed letting the CommandExecutor take a parameter for a notifier. Some commands will have different actions to show that they've executed (i.e. apex:class:create opens the file and org:open opens a browser) so it would be good to let the commands choose how they want to do that without overriding the execute method and repeating a lot of that method's code for a slight change. This will be done in a later PR.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -2,18 +2,25 @@
   "running_tasks_title_text": "Running Tasks",
 
   "force_auth_web_login_authorize_dev_hub_text": "SFDX: Authorize a Dev Hub",
-  "force_org_create_default_scratch_org_text": "SFDX: Create a Default Scratch Org...",
+  "force_org_create_default_scratch_org_text":
```
vazexqi: @ruthemmanuelle - Have these messages been approved? They come from the CLI in Palette spreadsheet but I'm not sure if these have been approved yet.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -53,6 +53,7 @@ export class ChannelService {
       } else {
         this.channel.appendLine(nls.localize('channel_end'));
       }
+      this.channel.appendLine('');
```
vazexqi: Is this just to add an extra newline for visual aesthetics or is there a functional reason for this?

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -18,23 +18,41 @@ import {
 } from './commands';
 
 class ForceSourceStatusExecutor extends SfdxCommandletExecutor<{}> {
+  private flag: string | undefined;
```
vazexqi: I think this should be a String enum given that we have only two choices: --local or --remote. This helps prevent failures in the future by enumerating the choice of everything that is possible.

Either use that or the "String Literal Types". See https://www.typescriptlang.org/docs/handbook/advanced-types.html

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -70,5 +70,7 @@ export const messages = {
   force_lightning_app_create_text: 'SFDX: Create Lightning App',
   force_lightning_component_create_text: 'SFDX: Create Lightning Component',
   force_lightning_event_create_text: 'SFDX: Create Lightning Event',
-  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface'
+  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface',
```
vazexqi: @ruthemmanuelle: These ones as well.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Yes, the messages in my column of the spreadsheet are fine to use. I haven't been running VS Code messages by the UI Text Reviews team, since their primary focus is on UI text and error messages for the core app. Just having me review these messages is fine.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Yes, these are fine. See my previous comment.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
Sweetman: It's just for visual aesthetic. Without this the first line of the output with the command title would not have a line separating it from the previous output.

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -17,20 +17,25 @@ import {
...
+export enum SourceStatusFlags {
+  Local = '--local',
+  Remote = '--remote'
+}
+
 class ForceSourceStatusExecutor extends SfdxCommandletExecutor<{}> {
-  private flag: string | undefined;
-  public constructor(flag?: string) {
+  private flag: SourceStatusFlags | undefined;
+  public constructor(flag?: SourceStatusFlags) {
```
vazexqi: Add a newline.

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -17,20 +17,25 @@ import {
...
     super();
     this.flag = flag;
   }
   public build(data: {}): Command {
```
vazexqi: Add a newline.

PR #103, opened 2017-09-07
Log Apex Debugger event SystemInfo

DatGuyJonathan:
### What does this PR do?
Apex Debugger event SystemInfo only has a value in its description field. It tells the client whether a request matched their session's userId/request/entryPoint filter. It can get pretty noisy so I'm just logging it in the debug console instead of also showing a top-level notification.

systeminfoevent

@W-4179008@

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -100,7 +100,7 @@ See this
 [guide](https://www.atlassian.com/git/tutorials/comparing-workflows#gitflow-workflow)
 from Atlassian on the flow. These steps are manual because you might encounter merge conflicts.
 
-1. `git push tags` to push the tags to GitHub.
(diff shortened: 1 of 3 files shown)


docs/publishing.md
```diff
@@ -100,7 +100,7 @@ See this
 [guide](https://www.atlassian.com/git/tutorials/comparing-workflows#gitflow-workflow)
 from Atlassian on the flow. These steps are manual because you might encounter merge conflicts.
 
-1. `git push tags` to push the tags to GitHub.
+1. `git push --tags` to push the tags to GitHub.
```
guw: I think this should be a separate commit (directly on develop is ok).

PR #105, opened 2017-09-08
Fix debugger thread ID

DatGuyJonathan:
### What does this PR do?
Use one-based thread ID that is unique per debugger session.

@W-4297137@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -65,7 +65,8 @@ export class ApexDebug extends DebugSession {
   protected myStreamingService = StreamingService.getInstance();
   protected sfdxProject: string;
   protected orgInfo: OrgInfo;
-  protected requestThreads: string[];
+  protected requestThreads: Map<number, string>;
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -65,7 +65,8 @@ export class ApexDebug extends DebugSession {
   protected myStreamingService = StreamingService.getInstance();
   protected sfdxProject: string;
   protected orgInfo: OrgInfo;
-  protected requestThreads: string[];
+  protected requestThreads: Map<number, string>;
```
guw: @DatGuyJonathan There is a `Handles` structure in the VS Code debugger package. I've seen it being used in the mock debugger. However, I don't know if it's zero-based or one-based. But it might simplify some of the logic.

DatGuyJonathan: Looks like a map with a customizeable counter that defaults to 1000. However, it can only get by key, but I need to be able to iterate through its values as well when handling Apex Debugger events.

PR #106, opened 2017-09-11
Prompt user to overwrite existing files

Sweetman:
### What does this PR do?
Addresses bug where files will be overwritten without a prompt. Add postcondition checker
@W-4252622@


packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -24,6 +24,70 @@ export interface PreconditionChecker {
   check(): boolean;
 }
 
+export interface PostconditionChecker<T> {
```
Sweetman: Added a PostconditionChecker interface that will check inputs gathered by ParameterGatherers. If it passes then the inputs will just be passed along, otherwise it will be replaced by a CancelResponse.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -70,5 +70,9 @@ export const messages = {
   force_lightning_app_create_text: 'SFDX: Create Lightning App',
   force_lightning_component_create_text: 'SFDX: Create Lightning Component',
   force_lightning_event_create_text: 'SFDX: Create Lightning Event',
-  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface'
+  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface',
+  warning_prompt_file_override:
+    'File(s) with specified path already exist in workspace. Would you like to override?',
```
ruthemmanuelle: Should "override" be "overwrite"?

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -24,6 +24,70 @@ export interface PreconditionChecker {
...
+  check(
+    inputs: ContinueResponse<T> | CancelResponse
+  ): Promise<ContinueResponse<T> | CancelResponse>;
+}
+
+export class FilePathExistsChecker
```
Sweetman: This will check if the file path given by the user already contains the file name they specified. If it does, then a warning prompt will show asking to cancel will show. Yes will pass the ContinueResponse along, No will replace with CancelResponse

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -24,6 +24,70 @@ export interface PreconditionChecker {
...
+        if (override === 'Yes') {
+          return inputs;
+        }
+      }
+    }
+    return { type: 'CANCEL' };
+  }
+}
+
+export class EmptyPostChecker implements PostconditionChecker<any> {
```
Sweetman: Added an EmptyPostChecker that will be the default for already existing commands.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -219,24 +283,27 @@ export abstract class SfdxCommandletExecutor<T>
...
+  private readonly postchecker: PostconditionChecker<T>;
   private readonly gatherer: ParametersGatherer<T>;
   private readonly executor: CommandletExecutor<T>;
 
   constructor(
     checker: PreconditionChecker,
     gatherer: ParametersGatherer<T>,
-    executor: CommandletExecutor<T>
+    executor: CommandletExecutor<T>,
+    postchecker = new EmptyPostChecker()
```
Sweetman: Make PostconditionChecker optional by specifying the default EmptyPostChecker.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -70,5 +70,9 @@ export const messages = {
   force_lightning_app_create_text: 'SFDX: Create Lightning App',
   force_lightning_component_create_text: 'SFDX: Create Lightning Component',
   force_lightning_event_create_text: 'SFDX: Create Lightning Event',
-  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface'
+  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface',
+  warning_prompt_file_override:
```
Sweetman: @ruthemmanuelle Could you take a look at this message? The context is what to prompt when the user tries to create an Apex/Lightning/VF file that already exists. I made it 'file(s)' because in Lightning's case, multiple files will be overwritten.

Sweetman:
@vazexqi I would like some feedback on this before I add it to the other create commands. We briefly talked about adding a PostconditionChecker when we talked about the Commandlet framework so I want to make sure this is the direction you had in mind.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
Sweetman: Right, good catch. Just started to notice that I've been mentally using them interchangeably which is incorrect. I'll update.

packages/salesforcedx-vscode-core/src/commands/commands.ts
vazexqi: It seems like it would better to create two postcondition checkers: one for the general case and one for the Lightning case. That way, it's easier to understand what is going on instead of having to read this `FilePathExistsChecker` and figure out why there is an 'if' statement case.

Then for the Lightning commands, those will specifically use the Lightning-specific one.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -24,6 +24,70 @@ export interface PreconditionChecker {
...
+      // If file does not exist then create it, otherwise prompt user to override the file
+      if (files.length === 0) {
+        return inputs;
+      } else {
+        const override = await notificationService.showWarningMessage(
+          nls.localize('warning_prompt_file_override'),
+          nls.localize('warning_prompt_yes'),
+          nls.localize('warning_prompt_no')
+        );
+        if (override === 'Yes') {
```
vazexqi: I'm not sure if this will always be 'Yes'. 
If you call nls.localize on the message before showing it, won't it be whatever the localized message is?

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -24,6 +24,70 @@ export interface PreconditionChecker {
...
+}
+
+export class FilePathExistsChecker
+  implements PostconditionChecker<DirFileNameSelection> {
+  private fileExtension: string;
+
+  public constructor(fileExtension: string) {
+    this.fileExtension = fileExtension;
+  }
+  public getGlobPattern(
```
vazexqi: Formatting. Add a newline between methods.

Sweetman: Method was deleted when separated Lightning into a separate PostconditionChecker

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -70,5 +70,11 @@ export const messages = {
   force_lightning_app_create_text: 'SFDX: Create Lightning App',
   force_lightning_component_create_text: 'SFDX: Create Lightning Component',
   force_lightning_event_create_text: 'SFDX: Create Lightning Event',
-  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface'
+  force_lightning_interface_create_text: 'SFDX: Create Lightning Interface',
+  warning_prompt_file_overwrite:
+    'File(s) with specified path already exists in workspace. Would you like to overwrite?',
+  warning_prompt_lightning_bundle_overwrite:
```
Sweetman: @ruthemmanuelle Could you check this new message? Context is that Lightning is a little different than the other create commands since it will place the files into a new directory with the name specified by the user.

We want to warn the user of 2 things with the message:
1) This directory is already in use by a lightning bundle (event, cmp, app, interface)
2) Depending on the bundle being created, some files may be overwritten since there are overlap in the files being created. See attached screenshot for an example.

image

/cc @vazexqi

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -72,5 +72,11 @@ export const messages = {
   force_lightning_event_create_text: 'SFDX: Create Lightning Event',
   force_lightning_interface_create_text: 'SFDX: Create Lightning Interface',
   force_source_status_local_text: 'SFDX: View Local Changes',
-  force_source_status_remote_text: 'SFDX: View Changes in Default Scratch Org'
+  force_source_status_remote_text: 'SFDX: View Changes in Default Scratch Org',
+  warning_prompt_file_overwrite:
+    'File(s) with specified path already exists in workspace. Would you like to overwrite?',
```
ruthemmanuelle: 'One or more files with the specified path already exist in your workspace. Do you want to overwrite them?'

Sweetman:
@vazexqi I have added the integration tests for the PostconditionCheckers and removed the (potentially flaky) spectron tests. I had to add a directory into the system-tests assets folder containing a lightning bundle for the integration tests.

One of the previous tests that verifies a glob expression depends on the number of directories in the assets folder so I am considering making the core package have it's own workspace. Seems somewhat odd for a test in core to be dependent on the spectron tests assets. Thoughts?

Other than that, I think it's ready to be merged.

PR #108, opened 2017-09-12
Debugger stop cmd

DatGuyJonathan:
### What does this PR do?
Query for active apex debugger session and update its status to detach. 

@W-4288245@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceDebuggerStop.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDebuggerStop.ts
@@ -1,0 +1,132 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/commands/forceDebuggerStop.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export class DebuggerSessionDetachExecutor extends SfdxCommandletExecutor<
+  IdSelection
+> {
+  public build(data: IdSelection): Command {
+    return new SfdxCommandBuilder()
+      .withArg('force:data:record:update')
+      .withFlag('--sobjecttype', 'ApexDebuggerSession')
+      .withFlag('--sobjectid', data ? data.id : '')
+      .withFlag('--values', 'Status="Detach"')
```
DatGuyJonathan: IDE1 updated its status to 'Kill' with a status message that says one of the org's admins killed the session because people were sharing development in a sandbox org. Scratch orgs are supposed to be per developer so I'm just using detach.

PR #109, opened 2017-09-12
James/travis timeout

Sweetman:
### What does this PR do?
Fix autobuild dependency issue
@W-4268060@

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,11 @@ os:
   - linux
   - osx
 
+addons:
+  apt:
+    packages:
+      - libsecret-1-dev
+
 before_install:
   - |
     if [[ $TRAVIS_OS_NAME == "linux" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,8 +68,7 @@ script:
   - npm run lint
   - |
     if [[ $TRAVIS_OS_NAME == "linux" ]]; then
-      travis_wait npm run test:without-system-tests
-      travis_wait npm run coverage:system-tests
+      travis_wait npm run test:without-apex-debugger-tests
     fi
   - |
     if [[ $TRAVIS_OS_NAME == "osx" ]]; then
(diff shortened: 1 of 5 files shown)


Sweetman:
@vazexqi I reverted the explicit xvfb dependency since it didn't fix the spectron application starting within the 36000ms period.

However, on a related note, I believe this error used to be `Check dependency list! Synchronous require cannot resolve module 'electron'. This is the first mention of this module!` until I added `electron` as a dev dependency. A git issue I followed for that error in VSCode seems to have been fixed.
https://github.com/Microsoft/vscode/issues/31106#issuecomment-329241638 

I have removed the electron dependency and reflected their fix in the most recent commit and think we should merge this in.

.travis.yml
```diff
@@ -28,6 +28,11 @@ os:
   - linux
   - osx
 
+addons:
+  apt:
+    packages:
+      - libsecret-1-dev
```
vazexqi: Do we care about which version of libsecret that we depend on? Also, is the package really called libsecret-1-dev, with the 1 in the middle?

packages/system-tests/src/spectron/application.ts
```diff
@@ -61,7 +61,8 @@ export class SpectronApplication {
       path: electronPath,
       args: args,
       chromeDriverArgs: chromeDriverArgs,
-      startTimeout: 10000
+      startTimeout: 10000,
+      requireName: 'nodeRequire'
```
vazexqi: Is this the change from the VS Code team?

Sweetman: Yes they added this to their spectron framework.

.travis.yml
Sweetman: I believe that is what the package is called. I am reflecting what VSCode uses in their .travis.yml

https://github.com/Microsoft/vscode/blob/master/.travis.yml

vazexqi:
LGTM. Let's make sure that all CI pass (since the purpose of this PR is to fix the PR) before merging it in :-)

Sweetman:
Yup will do. I forgot that adding the `electron` dependency wasn't only for seeing if it would change the error I mentioned before. It also has the latest declaration file for it's types. Adding it back in.

PR #110, opened 2017-09-12
Apex Debugger file change warning

DatGuyJonathan:
### What does this PR do?
Log warning message to debug console if class/trigger file is changed.

hotswapwarn

@W-4255647@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/messages/i18n.ts
+++ b/packages/salesforcedx-apex-debugger/src/messages/i18n.ts
@@ -30,5 +30,5 @@ export const messages = {
   streaming_handshake_timeout_text:
     'Streaming API handshake failed due to a socket timeout. Check your connection and try again.',
   hotswap_warn_text:
-    'Apex classes & triggers should not be modified during an active Apex Debugger session.'
+    "You can't modify Apex classes or triggers during an Apex Debugger session. Save your changes after you’re done debugging."
 };
--- a/packages/salesforcedx-apex-debugger/src/messages/i18n.ts
+++ b/packages/salesforcedx-apex-debugger/src/messages/i18n.ts
@@ -28,5 +28,7 @@ export const messages = {
   streaming_handshake_error_text:
     'Streaming API handshake failed with the following error',
   streaming_handshake_timeout_text:
-    'Streaming API handshake failed due to a socket timeout. Check your connection and try again.'
+    'Streaming API handshake failed due to a socket timeout. Check your connection and try again.',
+  hotswap_warn_text:
+    'Apex classes & triggers should not be modified during an active Apex Debugger session.'
 };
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-apex-debugger/src/messages/i18n.ts
```diff
@@ -28,5 +28,7 @@ export const messages = {
   streaming_handshake_error_text:
     'Streaming API handshake failed with the following error',
   streaming_handshake_timeout_text:
-    'Streaming API handshake failed due to a socket timeout. Check your connection and try again.'
+    'Streaming API handshake failed due to a socket timeout. Check your connection and try again.',
+  hotswap_warn_text:
+    'Apex classes & triggers should not be modified during an active Apex Debugger session.'
```
DatGuyJonathan: @ruthemmanuelle Apex Debugger doesn't allow changes to Apex source code while there's an active debugger session. The server side would actually terminate the session if the user pushes Apex source. If they don't push and keep the changes locally, it could confuse their breakpoints, stacktrace, and variables. This PR is to watch .cls & .trigger files in their SFDX project, and warn them that they shouldn't change Apex source during an active debugger session. Let me know how to best convey that message in this text.

guw: I think there should be two message:
* a warning that notifies people that changes to class and trigger will not be honored (this can be tight to the file watch)
* a general notice that hot-swap is not support and any push will terminate the debug session (possibly at every begin of debug session and/or readme)

DatGuyJonathan: @guw Regarding the second bullet point, is this sufficient? https://developer.salesforce.com/docs/atlas.en-us.eclipse.meta/eclipse/debugger_considerations.htm

ruthemmanuelle: @DatGuyJonathan - How's this? 
'You can’t modify Apex classes or triggers during an Apex Debugger session. Save your changes after you’re done debugging.'

DatGuyJonathan: Thanks Ruth, I'll update the text to that.

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
```diff
@@ -30,5 +30,5 @@ export const messages = {
   streaming_handshake_timeout_text:
     'Streaming API handshake failed due to a socket timeout. Check your connection and try again.',
   hotswap_warn_text:
-    'Apex classes & triggers should not be modified during an active Apex Debugger session.'
+    "You can't modify Apex classes or triggers during an Apex Debugger session. Save your changes after you’re done debugging."
```
ruthemmanuelle: There's one curly apostrophe and one straight apostrophe in this message. I'd prefer that we use curly apostrophes, but if we need to use straight ones we should do it consistently.

ruthemmanuelle: @DatGuyJonathan ^^^

DatGuyJonathan: Good catch! There's [this](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/messages/i18n.ts#L46) other usage of straight apostrophe, so I'll stick with that.

PR #111, opened 2017-09-13
[DO NOT MERGE] Add a simple SFDX: SObject Describe Account Command for testing

vazexqi:
### What does this PR do?

Adds a simple top-level command that is used to fetch the result of a describe call.

This is meant to illustrate the problem that for large data from the CLI, the results could be truncated. Now, **before anyone is alarmed**, this only happens under very special circumstances and doesn't affect any of the commands that we have in the released version of our extensions:

stdout-truncated

1. This seems to be on OS X.
1. This only seems to happen when you use `child_process.spawn` and are using the `process.stdout.on('data',....)` events.

This might be related to 
* https://github.com/nodejs/node/issues/6379
* https://github.com/nodejs/node/issues/2360 

where we might need to change the SFDX CLI to call `process.stdout._handle.setBlocking(true)`

However, having an interactive top-level command that reproduces this issue makes it easier for us to debug this on different platforms.

This PR is meant to serve as an easy way for us to debug this and discuss solutions with traceability.

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { ChildProcess, ExecOptions, spawn } from 'child_process';
+import { ChildProcess, spawn, SpawnOptions } from 'child_process';
 import 'rxjs/add/observable/fromEvent';
 import 'rxjs/add/observable/interval';
 import { Observable } from 'rxjs/Observable';
@@ -22,9 +22,9 @@ export interface CancellationToken {
 
 export class CliCommandExecutor {
   private readonly command: Command;
-  private readonly options: ExecOptions;
+  private readonly options: SpawnOptions;
 
-  public constructor(command: Command, options: ExecOptions) {
+  public constructor(command: Command, options: SpawnOptions) {
     this.command = command;
     this.options = options;
   }
@@ -64,6 +64,7 @@ export class CommandExecution {
     let timerSubscriber: Subscription | null;
 
     // Process
+
     this.processExitSubject = Observable.fromEvent(
       childProcess,
       'exit'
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -22,9 +22,9 @@ export interface CancellationToken {
 
 export class CliCommandExecutor {
   private readonly command: Command;
-  private readonly options: ExecOptions;
+  private readonly options: SpawnOptions;
```
vazexqi: This is a legit typing issue that we need to fix. So far it has not affected us since we don't use any of the other options.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -77,12 +78,27 @@ export class CommandExecution {
       childProcess,
       'error'
     ) as Observable<Error | undefined>;
-    this.processExitSubject.subscribe(next => {
+    this.processErrorSubject.subscribe(next => {
```
vazexqi: This is another error that should be fixed.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -77,12 +78,27 @@ export class CommandExecution {
...
       if (timerSubscriber) {
         timerSubscriber.unsubscribe();
       }
     });
 
+    childProcess.stdout.setEncoding('utf8');
+
+    childProcess.stdout.on('error', err => {
+      console.log(err);
```
vazexqi: Just logging to see if we have any errors on stdout and we don't.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -77,12 +78,27 @@ export class CommandExecution {
...
+    });
+
+    childProcess.stdout.on('end', () => {
+      console.log('end');
```
vazexqi: Just logging to see if we have any errors on stdout and we don't.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -203,7 +203,12 @@ export abstract class SfdxCommandletExecutor<T>
     const cancellationToken = cancellationTokenSource.token;
 
     const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: vscode.workspace.rootPath
+      cwd: vscode.workspace.rootPath,
+      stdio: [
+        'pipe',
+        'pipe', // fs.openSync(path.join(vscode.workspace.rootPath!, 'stdout.txt'), 'w'),
```
vazexqi: A possible, but annoying workaround, is to pipe the result to another stream.
Using a temp file seems to gather all the data. 
However, with that, you lose the ability to stream data as it arrives.

vazexqi:
The CI tests will fail since I'm adding a lot of logging. Like I mentioned, this is not for merging but for others to take out this branch, try it out and have a discussion on the PR.

vazexqi:
process-exit-truncates-stdout

Confirming that https://github.com/nodejs/node/issues/2360 still happens on Node v7.10.0 on OS X.

vazexqi:
The next version of the CLI, aka V6, should explicitly flush the buffer for us. See https://github.com/heroku/cli-engine/blob/master/src/cli.js#L118-L123

Closing this for now, will reopen if the issue reappears in V6.

vazexqi:
I just tested this with V6 of the CLI and it still exhibits this truncated issue.

screen shot 2017-09-28 at 10 44 25 am

PR #47, opened 2017-08-10
James/create apex class

Sweetman:
### What does this PR do?
Command palette action for creating an apex class with a specified name and directory
### What issues does this PR fix or reference?
@W-4197613@

PR #116, opened 2017-09-15
James/list config alias details

Sweetman:
### What does this PR do?
Adds alias:list, config:list, org:display to palette
@W-4285402, W-4285397, W-4285377@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -264,6 +264,14 @@
       {
         "command": "sfdx.force.config.list",
         "title": "%force_config_list_text%"
+      },
+      {
+        "command": "sfdx.force.alias.list",
+        "title": "%force_alias_list_text%"
+      },
+      {
+        "command": "sfdx.force.org.display",
+        "title": "%force_org_display_text%"
       }
     ]
   }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -181,6 +181,10 @@
         {
           "command": "sfdx.force.debugger.stop",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.config.list",
+          "when": "sfdx:project_opened"
         }
       ]
     },
@@ -256,6 +260,10 @@
       {
         "command": "sfdx.force.debugger.stop",
         "title": "%force_debugger_stop_text%"
+      },
+      {
+        "command": "sfdx.force.config.list",
+        "title": "%force_config_list_text%"
       }
     ]
   }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -272,10 +272,6 @@
       {
         "command": "sfdx.force.org.display.default",
         "title": "%force_org_display_default_text%"
-      },
-      {
-        "command": "sfdx.force.org.display.username",
-        "title": "%force_org_display_username_text%"
       }
     ]
   }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -270,8 +270,12 @@
         "title": "%force_alias_list_text%"
       },
       {
-        "command": "sfdx.force.org.display",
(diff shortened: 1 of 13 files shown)


DatGuyJonathan:
Let's add a couple quick unit tests for alias:list and config:list.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -22,6 +22,11 @@
   "force_lightning_event_create_text": "SFDX: Create Lightning Event",
   "force_lightning_interface_create_text": "SFDX: Create Lightning Interface",
   "force_source_status_local_text": "SFDX: View Local Changes",
-  "force_source_status_remote_text": "SFDX: View Changes in Default Scratch Org",
-  "force_debugger_stop_text": "SFDX: Stop Apex Debugger Session"
+  "force_source_status_remote_text":
+    "SFDX: View Changes in Default Scratch Org",
+  "force_debugger_stop_text": "SFDX: Stop Apex Debugger Session",
```
Sweetman: @ruthemmanuelle could you verify that these messages are ok? The last one has ellipses to indicate that there will be input required to indicate which org. This seems to be a common trend in other VSCode commands/extensions (i.e. `Files: Open...`, `Git: Push to...`, etc)

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -44,6 +44,7 @@ export const messages = {
   parameter_gatherer_enter_file_name: 'Enter desired filename',
   parameter_gatherer_enter_dir_name:
     "Enter desired directory (Press 'Enter' to confirm or 'Escape' to cancel)",
+  parameter_gatherer_enter_username_name: 'Enter target username',
```
Sweetman: @ruthemmanuelle This is the message that will show in the input box for the Display Org command where the user indicates which username. Target username is the name of the flag so I went with that.

Sweetman: Just noticed that the message above this one includes directions about confirming/canceling. Do we want that to be uniform across all parameter gathering messages?

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -22,6 +22,11 @@
...
+  "force_config_list_text": "SFDX: List All Config Variables",
+  "force_alias_list_text": "SFDX: List All Aliases",
+  "force_org_display_default_text": "SFDX: Display Org Details",
```
ruthemmanuelle: What about this? "SFDX: Display Org Details for Default Scratch Org"

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Yeah, it looks like you added the message above this one as part of PR #47. I think it was while I was on PTO. Do you remember why you included the text in parentheses? Is it standard in VS Code? It would be good to be consistent.

Sweetman: I think I added it after discussing with Nick. I've seen it in other extension commands like `Git: Create Branch...` but it doesn't seem to be a standard amongst commands.
image

ruthemmanuelle: Thanks for the info. Might as well go ahead and add it to the other prompts, and to this new one—it might be useful, and I don't think it will hurt.

Per Salesforce's style guidelines, Escape should be Esc, and there shouldn't be any quotation marks around Enter or Esc.

Sweetman:
Separated out org:display with target username for now

packages/salesforcedx-vscode-core/package.json
```diff
@@ -181,6 +181,10 @@
         {
           "command": "sfdx.force.debugger.stop",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.config.list",
+          "when": "sfdx:project_opened"
         }
```
DatGuyJonathan: We should add alias:list and org:display here as well. Otherwise, those commands can be executed without an sfdx folder and it'll just error out.

PR #119, opened 2017-09-18
Glob should list paths directly to folders with specified key word

Sweetman:
### What does this PR do?
Changes glob pattern from listing all paths with keyword to paths with keyword as last fragment
@W-4271155@


packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -276,7 +276,7 @@ export class SelectPrioritizedDirPath extends SelectDirPath {
 export class SelectStrictDirPath extends SelectDirPath {
   public globDirs(srcPath: string, priorityKeyword?: string): string[] {
     const globPattern = priorityKeyword
-      ? path.join(srcPath, '**/', priorityKeyword + '/**/')
+      ? path.join(srcPath, '**/', priorityKeyword + '/')
```
Sweetman: Removing the "**" will make the glob return only paths ending with the keyword appear.

PR #120, opened 2017-09-18
Add the time to the start of each test command

Sweetman:
### What does this PR do?
Adds timestamps to the commands ran on the travis build
Will help record how long commands are taking and if there is any opportunity for performance improvement


chowwinston:
@Sweetman was this created when you were debugging build issues in travis? Do you still want this in?

PR #121, opened 2017-09-18
Gary/genfauxclass

GaryFNC:
Add the capability and command to process SObjects from a scratch org and generate faux classes for both Standard and Custom SObjects.

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -15,7 +15,7 @@
   "devDependencies": {
     "@types/chai": "^4.0.0",
     "@types/jsforce": "^1.7.1",
-    "@types/mkdirp": "0.3.29",
+    "@types/shelljs": "0.7.4",
     "@types/mocha": "2.2.38",
     "@types/node": "^6.0.40",
     "@types/rimraf": "^2.0.2",
@@ -24,7 +24,7 @@
     "request-light": "0.2.1",
     "cross-env": "5.0.4",
     "jsforce": "1.7.1",
-    "mkdirp": "0.5.1",
+    "shelljs": "0.7.8",
     "mocha": "3.2.0",
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -46,7 +46,7 @@
   },
   "contributes": {
     "configuration": {
-      "title": "Salesforce DX Gating",
+      "title": "Salesforce DX",
       "properties": {
         "sfdx.refreshSObjects.enabled": {
           "type": "boolean",
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -41,5 +41,18 @@
   },
   "nyc": {
     "reporter": ["text-summary", "lcov"]
+  },
+  "contributes": {
+    "configuration": {
+      "title": "Salesforce DX Gating",
+      "properties": {
+        "sfdx.refreshSObjects.enabled": {
+          "type": "boolean",
+          "default": false,
+          "description":
+            "Enables refresh of SObject definitions from a scratch org"
+        }
+      }
+    }
   }
 }
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -20,6 +20,7 @@
     "@types/rimraf": "^2.0.2",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
+    "request-light": "0.2.1",
     "cross-env": "5.0.4",
     "jsforce": "1.7.1",
     "mocha": "3.2.0",
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -15,6 +15,7 @@
   "devDependencies": {
     "@types/chai": "^4.0.0",
     "@types/jsforce": "^1.7.1",
+    "@types/mkdirp": "0.3.29",
     "@types/mocha": "2.2.38",
     "@types/node": "^6.0.40",
     "@types/rimraf": "^2.0.2",
@@ -23,6 +24,7 @@
     "request-light": "0.2.1",
     "cross-env": "5.0.4",
     "jsforce": "1.7.1",
+    "mkdirp": "0.5.1",
     "mocha": "3.2.0",
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -172,7 +172,7 @@ type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
 export class SObjectDescribe {
   private accessToken: string;
   private instanceUrl: string;
-  // TODO should get the proper version from ??
+  // TODO should get the proper version from the project config info
   private readonly servicesPath: string = 'services/data';
   private readonly targetVersion = '40.0';
   private readonly versionPrefix = 'v' + this.targetVersion;
@@ -198,12 +198,13 @@ export class SObjectDescribe {
       this.instanceUrl = orgInfo.instanceUrl;
     }
   }
+
   public async describeSObject(
     projectPath: string,
     type: string,
     username?: string
   ): Promise<SObject> {
-    await this.setupConnection(projectPath);
+    await this.setupConnection(projectPath, username);
 
     const urlElements = [
       this.instanceUrl,
@@ -269,18 +270,18 @@ export class SObjectDescribe {
   public async describeSObjectBatch(
     projectPath: string,
     types: string[],
-    lastProcessed: number,
+    nextToProcess: number,
     username?: string
   ): Promise<SObject[]> {
     const batchSize = 25;
 
-    await this.setupConnection(projectPath);
+    await this.setupConnection(projectPath, username);
 
     const batchRequest: BatchRequest = { batchRequests: [] };
 
     for (
-      let i = lastProcessed + 1;
-      i <= lastProcessed + batchSize && i < types.length;
+      let i = nextToProcess;
+      i < nextToProcess + batchSize && i < types.length;
       i++
     ) {
       const urlElements = [this.sobjectsPart, types[i], 'describe'];
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -164,29 +164,24 @@ export enum SObjectCategory {
   CUSTOM = 'CUSTOM'
 }
 
-/*export interface Command {
-  readonly command: string;
-  readonly description?: string;
-  readonly args: string[];
+type SubRequest = { method: string; url: string };
+type BatchRequest = { batchRequests: SubRequest[] };
 
-  toString(): string;
-  toCommand(): string;
-}
+type SubResponse = { statusCode: number; result: SObject };
 
-export interface Options {
-  // Supports "string", "buffer", "array", "uint8array" or "object".
-  encoding?: string;
-}*/
+type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
 
 export class SObjectDescribe {
-  private conn: jsforce.Connection;
   private accessToken: string;
   private instanceUrl: string;
-  private readonly servicesPath: string = 'services/data/v40.0/sobjects';
+  // TODO should get the proper version from ??
+  private readonly servicesPath: string = 'services/data';
+  private readonly sobjectsPart: string = 'v40.0/sobjects';
+  private readonly batchPart: string = 'v40.0/composite/batch';
 
   // get the token and url by calling the org - short term, should be able to get it from the sfdx project
   private async setupConnection(projectPath: string) {
-    if (!this.conn) {
+    if (!this.accessToken) {
       let orgInfo: any;
       const execution = new CliCommandExecutor(
         new SfdxCommandBuilder()
@@ -200,11 +195,6 @@ export class SObjectDescribe {
       orgInfo = JSON.parse(result).result;
       this.accessToken = orgInfo.accessToken;
       this.instanceUrl = orgInfo.instanceUrl;
-
-      this.conn = new jsforce.Connection({
-        accessToken: this.accessToken,
-        instanceUrl: this.instanceUrl
-      });
     }
   }
   public async describeSObject(
@@ -212,30 +202,35 @@ export class SObjectDescribe {
     type: string,
     username?: string
   ): Promise<SObject> {
-    console.log('SObjectDescribe - ' + type);
-    const builder = new SfdxCommandBuilder()
-      .withArg('force:schema:sobject:describe')
-      .withFlag('--sobjecttype', type);
-    if (username) {
-      builder.args.push('--targetusername', username);
-    }
-    const command = builder.withJson().build();
-    const execution = new CliCommandExecutor(command, {
-      cwd: projectPath
-    }).execute();
+    await this.setupConnection(projectPath);
 
-    const cmdOutput = new CommandOutput();
-    let result;
+    const urlElements = [
+      this.instanceUrl,
+      this.servicesPath,
+      this.sobjectsPart,
+      type,
+      'describe'
+    ];
+
+    const requestUrl = urlElements.join('/');
+
+    const options: XHROptions = {
+      type: 'GET',
+      url: requestUrl,
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+        Authorization: `OAuth ${this.accessToken}`
+      }
+    };
+
     try {
-      result = await cmdOutput.getCmdResult(execution);
-    } catch (e) {
-      return Promise.reject(e);
-    }
-    try {
-      const sobject = JSON.parse(result).result as SObject;
+      const response: XHRResponse = await xhr(options);
+      const sobject = JSON.parse(response.responseText) as SObject;
       return Promise.resolve(sobject);
-    } catch (e) {
-      return Promise.reject(result);
+    } catch (error) {
+      const xhrResponse: XHRResponse = error;
+      return Promise.reject(xhrResponse.responseText);
     }
   }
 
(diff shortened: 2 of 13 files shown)


packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -15,13 +15,16 @@
   "devDependencies": {
     "@types/chai": "^4.0.0",
     "@types/jsforce": "^1.7.1",
+    "@types/mkdirp": "0.3.29",
```
vazexqi: Don't use this. Just use shelljs.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -15,13 +15,16 @@
...
     "@types/mocha": "2.2.38",
     "@types/node": "^6.0.40",
     "@types/rimraf": "^2.0.2",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
+    "request-light": "0.2.1",
     "cross-env": "5.0.4",
     "jsforce": "1.7.1",
+    "mkdirp": "0.5.1",
```
vazexqi: Don't use this.

packages/salesforcedx-sobjects-faux-generator/package.json
GaryFNC: OK will do

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -161,30 +162,76 @@ export enum SObjectCategory {
...
+
+type SubResponse = { statusCode: number; result: SObject };
+
+type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
+
 export class SObjectDescribe {
+  private accessToken: string;
+  private instanceUrl: string;
+  // TODO should get the proper version from ??
+  private readonly servicesPath: string = 'services/data';
```
vazexqi: If you use a readonly field, there is a slot created for SObjectDescribe instance. This seems redundant since all of them are the same.
Does it make more sense to pull this out of the class and use a `const` instead?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -161,30 +162,76 @@ export enum SObjectCategory {
...
+        cwd: projectPath
+      }).execute();
+      const cmdOutput = new CommandOutput();
+      const result = await cmdOutput.getCmdResult(execution);
+      orgInfo = JSON.parse(result).result;
+      this.accessToken = orgInfo.accessToken;
+      this.instanceUrl = orgInfo.instanceUrl;
+    }
+  }
   public async describeSObject(
```
vazexqi: Newline between methods.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -161,30 +162,76 @@ export enum SObjectCategory {
...
+    ];
+
+    const requestUrl = urlElements.join('/');
+
+    const options: XHROptions = {
+      type: 'GET',
+      url: requestUrl,
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
```
vazexqi: 'Content-Type' in quotes but not 'Accept'? Probably should try to be consistent throughout unless you have a particular reason for that.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
+
+  private createIfNeededOutputFolder(folderPath: string): boolean {
+    if (!fs.existsSync(folderPath)) {
+      mkdirp.sync(folderPath);
+      return fs.existsSync(folderPath);
+    }
+    return true;
+  }
+
+  public async generate(
```
vazexqi: This method seems too long. Please refactor.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
+    }
+
+    return '';
   }
 
-  private async generateFauxClass(
-    projectPath: string,
+  private generateChildRelationship(rel: ChildRelationship): string {
+    if (rel.relationshipName) {
+      return 'List<' + rel.childSObject + '> ' + rel.relationshipName;
```
vazexqi: Use string templates instead of concatenation.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
+    } else {
+      // expect the name to end with Id, then strip off Id
+      if (rel.field.endsWith('Id')) {
+        return (
+          rel.childSObject + ' ' + rel.field.slice(0, rel.field.length - 2)
```
vazexqi: String templates instead of concatenation.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
vazexqi: It seems like most of it is long because of all the logging? Is the logging required? Does it provide us with useful information? Or is it just generating a bunch of noise for the user?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: wonder why lint didnt complain - should it have?  - no problem, will fix...

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
-import { SObjectCategory, SObjectDescribe } from '../describe';
+import {
+  ChildRelationship,
+  Field,
+  SObject,
+  SObjectCategory,
+  SObjectDescribe
+} from '../describe';
 
 export class FauxClassGenerator {
```
vazexqi: How do you feel about creating mini classes that generate different parts of the class stub? 
One part for the class itself.
One part for the fields. 
All the classes can be contained inside this single file.
But it might make it more clear and easier to test instead of just doing everything in one class.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
+    return declarations;
+  }
+
+  private generateFauxClassTextFromDecls(
+    className: string,
+    declarations: string[]
+  ): string {
+    // sort, but filter out duplicates
+    // which can happen due to childRelationships w/o a relationshipName
+    declarations.sort(function nameCompare(
```
vazexqi: Can we use the lambda syntax instead of declaring a function explicitly in there?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
+    ) {
+      return (
+        !index ||
+        FauxClassGenerator.fieldName(value) !==
+          FauxClassGenerator.fieldName(array[index - 1])
+      );
+    });
+
+    const indentAndModifier = '    global ';
+    const generatedClass: string =
```
vazexqi: I don't think we should hardcode '\n'. This is unlikely to work on Windows.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: IMHO it feels odd,  there really isn't anything in those classes except for a function, not what I think of when I think of a class...    I agree that methods/functions shouldn't be long at all - but I"m not so certain that the class size is fundamental.    Seems more important to have class coherence...      And yes, there are some tests still missing

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
```diff
@@ -0,0 +1,135 @@
+import * as chai from 'chai';
+import * as fs from 'fs';
+import { FauxClassGenerator } from '../src/generator/fauxClassGenerator';
+
+const expect = chai.expect;
+
+describe('generate fields set', function() {
```
vazexqi: Do you clean up the generated folders?

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -73,11 +74,20 @@ export class CommandExecution {
         timerSubscriber.unsubscribe();
       }
     });
+    this.processCloseSubject = Observable.fromEvent(
```
vazexqi: We probably don't need this. The issue seems to be a bit more subtle than just using Close vs Exit. See #111

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
GaryFNC: no, good catch - when we regenerate, I need to blow away the old files since I don't want to try to track exactly what has changed, etc....    that's a bug

packages/salesforcedx-vscode-core/package.json
```diff
@@ -181,6 +182,10 @@
         {
           "command": "sfdx.force.debugger.stop",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "refreshsobjects",
```
vazexqi: Put this under the sfdx.force.internal.refreshsobjects namespace for now.
Command ids are meant to be unique so it's best to qualify them properly with some prefix.

GaryFNC: OK, will do - I wasn't sure whether the pattern was to only use sfdx.force for items that were CLI related to sfdx....

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
vazexqi: Right now your tests verify both the generated text and also the creation of the folders.
Might be better to be able to verify them independently.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
GaryFNC: I had moved away from using the Executor as much as possible towards REST although there is more to do on describeGlobal.     I just updated Winston's last code to use Close instead of Exit as we had discussed...

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: Good point - I did see it used in other places (localization.ts and a couple of places in linter)...    I'll try to come up with a better way.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
vazexqi: It can't :-(

Or at least, I don't know how to enforce it with either tslint or prettier.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
vazexqi: OK. Just checking with you. I don't have a strong opinion of it now. We can revisit if the class's responsibilities get bigger.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: And I do agree that the method is too long  - will fix that ASAP...   I think that the logging is actually useful information. Except probably the one on the output folders - I'll get rid of that.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: You are right, it's a bit cleaner to pull it out.  I put it there because the other relevant strings were dynamic in some way tied to the api version which should really be found from the workspace.  Needs to be cleaned up properly once I have the code to obtain that info properly.  I"m going to just leave it until then.   On my list of things to figure out from force-com-toolbelt code.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: just a mistake - will fix

GaryFNC: Hmm, you are probably right.  I just copied this from BaseCommand.ts that Jonathan pointed me at - and that I was going to unify later on how we do REST calls....

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,249 @@
...
-      console.log(sobject.name);
-      this.generateFauxClass(projectPath, sobject);
+    const standardSObjects: SObject[] = [];
+    const customSObjects: SObject[] = [];
+    let fetchedSObjects: SObject[] = [];
+    let j = 0;
+    while (j < sobjects.length) {
+      try {
+        fetchedSObjects = fetchedSObjects.concat(
+          await describe.describeSObjectBatch(projectPath, sobjects, j - 1)
```
chowwinston: Why don't you just have this start at j = 0 rather than -1? It's weird to set lastProcessed to -1 then increment it during the batch call.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -0,0 +1,53 @@
...
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}> {
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_force_refresh_sobjects'))
```
chowwinston: An extra force here?

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -40,5 +43,18 @@
   },
   "nyc": {
     "reporter": ["text-summary", "lcov"]
+  },
+  "contributes": {
+    "configuration": {
+      "title": "Salesforce DX Gating",
+      "properties": {
+        "sfdx.refreshSObjects.enabled": {
```
vazexqi: The way it's written, I believe that this setting will be publicly visible when people go to Code > Settings. That's intentional? I think that should be fine since this feature is going to be stable enough soon and this setting can be removed.

There is also the way of doing a hidden setting. This is used by the tslint VS code extension for turning on extra logging. See https://github.com/Microsoft/vscode-tslint/blob/9477a7e48e0ed4029923aa51ebf7953ca0cfd8ea/tslint/README.md#faq

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -205,12 +252,70 @@ export class SObjectDescribe {
...
+        Authorization: `OAuth ${this.accessToken}`
+      },
+      data: JSON.stringify(batchRequest)
+    };
+
+    try {
+      const response: XHRResponse = await xhr(options);
+      const batchResponse = JSON.parse(response.responseText) as BatchResponse;
```
chowwinston: We had talked about storing the raw JSON for validation or future uses. Should we still do that?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: Well, it is because the arrays are 0 based, so lastProcessed needs to be lower than 0..  But, I agree it reads funny, so I may change it to "nextToProcess" and 0 base.    Old habit from days where you could overflow the integer range easily.

packages/salesforcedx-sobjects-faux-generator/package.json
GaryFNC: I will do hidden - I didn't quite realize how to do it - thanks

GaryFNC: I see tslint.trace.server - doesn't seem to be hidden.  And I tried the same scope: "window" that they used and I still see the setting..      I think I won't worry about it for now - it's temporary in any case.    If you see what I missed to make it hidden, let me know and I'll change it.

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
GaryFNC: fixed

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: yes, we should - I just didn't want to try to jam it in - hopefully get this PR in and done :-)...     But yes, we will need it for sure when we do the XML parse step with custom fields...

GaryFNC:
There is still an outstanding issue with the integration with the Apex LSP where it doesn't always pick up the changes when a Refresh SObjects Definition is done.   When the LSP/VSCode is restarted, it always gets the right info from any prior Refresh...   Under investigation in LSP.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
vazexqi: I know why this happens. It's because of the prettier formatter removing unnecessary quotes.

packages/salesforcedx-sobjects-faux-generator/package.json
vazexqi: screen shot 2017-09-21 at 8 45 49 pm

Yes, you're right. Apologies for leading you down a rabbit hole.

vazexqi:
@GaryFNC What did you do with the default header that's created for all PR?

Could you fill this in?

### What does this PR do?

### What is not covered in this PR?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,245 @@
...
-      console.log(sobject.name);
-      this.generateFauxClass(projectPath, sobject);
+    const standardSObjects: SObject[] = [];
+    const customSObjects: SObject[] = [];
+    let fetchedSObjects: SObject[] = [];
+    let j = 0;
+    while (j < sobjects.length) {
+      try {
+        fetchedSObjects = fetchedSObjects.concat(
+          await describe.describeSObjectBatch(projectPath, sobjects, j)
```
vazexqi: This might be a long-hanging fruit. We might be doing things too sequentially here.
See the section "Careful! Avoid going too sequential" at https://developers.google.com/web/fundamentals/getting-started/primers/async-functions

Doesn't need to be fixed in this PR but I want  you to be aware of it.

CC: @chowwinston @DatGuyJonathan @guw @Sweetman

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -1,33 +1,245 @@
...
 
 export class FauxClassGenerator {
-  public async generate(projectPath: string, type: SObjectCategory) {
+  private SFDX_DIR = '.sfdx';
+  private TOOLS_DIR = 'tools';
+  private SOBJECTS_DIR = 'sobjects';
+  private STANDARDOBJECTS_DIR = 'standardObjects';
+  private CUSTOMOBJECTS_DIR = 'customObjects';
+
+  public async generate(
```
vazexqi: @GaryFNC  - What is your strategy to ensure _completeness_ of what you are doing? Did you go through each case at https://help.salesforce.com/articleView?id=overview_of_custom_object_relationships.htm&type=0 and handle each relationship?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -222,7 +223,8 @@ export class SObjectDescribe {
       headers: {
         'Content-Type': 'application/json',
         Accept: 'application/json',
-        Authorization: `OAuth ${this.accessToken}`
+        Authorization: `OAuth ${this.accessToken}`,
+        'User-Agent': 'salesforcedx-extension'
```
vazexqi: Please remove this per our discussion. It's best to not simple introduce a random user-agent.

GaryFNC: It needs to remain until the 210/patch is released and verified that it works - otherwise we will get random failures.    It can be removed once the 210/patch is fully available.

vazexqi:
@GaryFNC Please do the following:

- [ ] Change the name of the preference top-level from "Salesforce DX Gating" to "Salesforce DX"
- [ ] Remove your changes to introduce in commandExecutor.ts and commandOutput.ts. We know that  using `processClose` doesn't make a difference so instead of introducing yet another one, let's just use  `processExit`.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: looking at that doc this morning - I have looked previously, but need to review it again

GaryFNC:
**What does this PR do?**
This PR provides the VSCode extension support for creating faux classes for all SObjects in the default scratch org.     This includes the command "Refresh SObjects Definition" and the gating of that command.    It generates the faux classes for both standard SObjects and custom SObjects by invoking the "describe" REST call to core.

**What issues does this PR fix or reference?**
W-4231782

**What is not covered in this PR?**
There are still tests that need to be added.   And a new bug with generated fields named null.   This PR does not address the changes on the Apex LS side for properly handling the refresh of the newly generated classes.   Those will be both be done as additional work items here and for the LS.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: I will make sure that the test plan covers all these situations and enter child bugs for anything not handled.   The primary cases are handled, but need to verify some of the more unusual cases

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: interesting - I can try doing the batch fetches in parallel and see if that helps - good spot

PR #124, opened 2017-09-19
Variables support for debugger

guw:
### What does this PR do?
This PR adds variables support to the debugger. Supported variable types are objects, collections and map. Whatever is returned from the Apex debugger will be shown.

#### Out of scope:
* Editing of variables (not supported)

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -919,6 +919,7 @@ export class ApexDebug extends LoggingDebugSession {
     response: DebugProtocol.VariablesResponse,
     args: DebugProtocol.VariablesArguments
   ): Promise<void> {
+    response.success = true;
     const variablesContainer = this.variableHandles.get(
       args.variablesReference
     );
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -65,13 +65,16 @@ import { VscodeDebuggerMessage, VscodeDebuggerMessageType } from '../index';
 import { nls } from '../messages';
 import os = require('os');
 
+const TRACE_ALL = 'all';
+const TRACE_CATEGORY_VARIABLES = 'variables';
+const TRACE_CATEGORY_LAUNCH = 'launch';
+const TRACE_CATEGORY_PROTOCOL = 'protocol';
+
+export type TraceCategory = 'all' | 'variables' | 'launch' | 'protocol';
+
 export interface LaunchRequestArguments
   extends DebugProtocol.LaunchRequestArguments {
-  /** comma separated list of trace selectors. Supported:
-	  * 'all': all
-	  * 'la': launch/attach
-	  * 'vh': variable handling
-	  * 'dap': debug adapter protocol
+  /** comma separated list of trace selectors (see TraceCategory)
 	  */
   trace?: boolean | string;
   userIdFilter?: string;
@@ -471,14 +474,14 @@ export class ApexDebug extends LoggingDebugSession {
     args: LaunchRequestArguments
   ): Promise<void> {
     if (typeof args.trace === 'boolean') {
-      this.trace = args.trace ? ['all'] : undefined;
+      this.trace = args.trace ? [TRACE_ALL] : undefined;
       this.traceAll = args.trace;
     } else if (typeof args.trace === 'string') {
       this.trace = args.trace.split(',');
-      this.traceAll = this.trace.indexOf('all') >= 0;
+      this.traceAll = this.trace.indexOf(TRACE_ALL) >= 0;
     }
-    if (this.trace && this.trace.indexOf('dap') >= 0) {
-      // only log debug adapter protocol if 'dap' tracing flag is set, ignore traceAll here
+    if (this.trace && this.trace.indexOf(TRACE_CATEGORY_PROTOCOL) >= 0) {
+      // only log debug adapter protocol if 'protocol' tracing flag is set, ignore traceAll here
       logger.setup(Logger.LogLevel.Verbose, /*logToFile=*/ false);
     } else {
       logger.setup(Logger.LogLevel.Stop, false);
@@ -752,7 +755,7 @@ export class ApexDebug extends LoggingDebugSession {
           if (i === 0 && stateRespObj.stateResponse.state) {
             // populate first stack frame with info from state response (saves a server round trip)
             this.log(
-              'va',
+              TRACE_CATEGORY_VARIABLES,
               'stackTraceRequest: state=' +
                 JSON.stringify(stateRespObj.stateResponse.state)
             );
@@ -887,7 +890,7 @@ export class ApexDebug extends LoggingDebugSession {
     const frameInfo = this.stackFrameInfos.get(args.frameId);
     if (!frameInfo) {
       this.log(
-        'va',
+        TRACE_CATEGORY_VARIABLES,
         `scopesRequest: no frame info found for stack frame ${args.frameId}`
       );
       response.body = { scopes: [] };
@@ -931,7 +934,7 @@ export class ApexDebug extends LoggingDebugSession {
     );
     if (!variablesContainer) {
       this.log(
-        'va',
+        TRACE_CATEGORY_VARIABLES,
         `variablesRequest: no container for variablesReference=${args.variablesReference}`
       );
       // no container found: return empty variables array
@@ -953,7 +956,7 @@ export class ApexDebug extends LoggingDebugSession {
       })
       .catch(err => {
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `variablesRequest: error reading variables ${err} ${err.stack}`
         );
         // in case of error return empty variables array
@@ -978,7 +981,7 @@ export class ApexDebug extends LoggingDebugSession {
       frameRespObj.frameResponse.frame
     ) {
       this.log(
-        'va',
+        TRACE_CATEGORY_VARIABLES,
         `fetchFrameVariables: frame ${frameInfo.frameNumber} frame=` +
           JSON.stringify(frameRespObj.frameResponse.frame)
       );
@@ -1032,7 +1035,7 @@ export class ApexDebug extends LoggingDebugSession {
           new ObjectReferenceContainer(reference, requestId)
         );
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `populateReferences: new object reference: ${variableReference} for ${reference.id} ${reference.nameForMessages}`
         );
       } else if (reference.type === 'list' || reference.type === 'set') {
@@ -1040,7 +1043,7 @@ export class ApexDebug extends LoggingDebugSession {
           new CollectionReferenceContainer(reference, requestId)
         );
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `populateReferences: new ${reference.type} reference: ${variableReference} for ${reference.id} ${reference.nameForMessages}`
         );
       } else if (reference.type === 'map') {
@@ -1055,13 +1058,13 @@ export class ApexDebug extends LoggingDebugSession {
         }
         variableReference = this.variableHandles.create(mapContainer);
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `populateReferences: new map reference: ${variableReference} for ${reference.id} ${reference.nameForMessages}`
         );
       } else {
         const referenceInfo = JSON.stringify(reference);
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `populateReferences: unhandled reference: ${referenceInfo}`
         );
         return;
@@ -1086,7 +1089,7 @@ export class ApexDebug extends LoggingDebugSession {
       await this.fetchReferences(requestId, apexId);
       if (!this.variableContainerReferenceByApexId.has(apexId)) {
         this.log(
-          'va',
+          TRACE_CATEGORY_VARIABLES,
           `resolveApexIdToVariableReference: no reference found for apexId ${apexId} (request ${requestId})`
         );
         return;
@@ -1096,7 +1099,7 @@ export class ApexDebug extends LoggingDebugSession {
       apexId
     );
     this.log(
-      'va',
+      TRACE_CATEGORY_VARIABLES,
       `resolveApexIdToVariableReference: resolved apexId=${apexId} to variableReference=${variableReference}`
     );
     return variableReference;
@@ -1107,7 +1110,7 @@ export class ApexDebug extends LoggingDebugSession {
     ...apexIds: number[]
   ): Promise<void> {
     this.log(
-      'va',
+      TRACE_CATEGORY_VARIABLES,
       `fetchReferences: fetching references with apexIds=${apexIds} (request ${requestId})`
     );
     const referencesResponse = await new ReferencesCommand(
@@ -1143,7 +1146,7 @@ export class ApexDebug extends LoggingDebugSession {
       return;
     }
     this.log(
-      'va',
+      TRACE_CATEGORY_VARIABLES,
       `fetchReferences: fetching references with apexIds=${apexIdsToFetch} (request ${requestId})`
     );
     await this.fetchReferences(requestId, ...apexIdsToFetch);
@@ -1180,7 +1183,7 @@ export class ApexDebug extends LoggingDebugSession {
     }
   }
 
-  public log(traceCategory: string, message: string) {
+  public log(traceCategory: TraceCategory, message: string) {
     if (
       this.trace &&
       (this.traceAll || this.trace.indexOf(traceCategory) >= 0)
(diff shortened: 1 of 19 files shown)


guw:
@DatGuyJonathan first PR draft - currently on rework to support multiple requests

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -361,9 +744,61 @@ export class ApexDebug extends DebugSession {
           const sourcePath = this.myBreakpointService.getSourcePathFromTyperef(
             serverFrames[i].typeRef
           );
+          const frameInfo = new ApexDebugStackFrameInfo(
+            requestId,
+            serverFrames[i].frameNumber
+          );
+          const frameId = this.stackFrameInfos.create(frameInfo);
+          if (i === 0 && stateRespObj.stateResponse.state) {
```
DatGuyJonathan: You changed StackFrame id below from i to frameId. Should this be frameId as well?

packages/salesforcedx-vscode-apex-debugger/package.nls.json
```diff
@@ -1,9 +1,15 @@
...
+    "A regular expression matching entry points whitelisted for debugging",
   "sfdx_project_text": "Location of SFDX project",
   "launch_snippet_label_text": "Apex Debugger: Launch",
-  "launch_snippet_description_text": "A new configuration for launching Apex Debugger",
-  "launch_snippet_name": "Launch Apex Debugger"
+  "launch_snippet_description_text":
+    "A new configuration for launching Apex Debugger",
+  "launch_snippet_name": "Launch Apex Debugger",
+  "trace_text":
+    "Produce diagnostic output. Instead of setting this to true you can list one or more selectors separated with commas. The 'all' selector enables very detailed output."
```
DatGuyJonathan: Include possible values so people know what to put in.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -445,14 +880,286 @@ export class ApexDebug extends DebugSession {
...
-  private logToDebugConsole(
+  protected async scopesRequest(
+    response: DebugProtocol.ScopesResponse,
+    args: DebugProtocol.ScopesArguments
+  ): Promise<void> {
+    const frameInfo = this.stackFrameInfos.get(args.frameId);
+    if (!frameInfo) {
+      this.sendErrorResponse(
+        response,
+        2020,
```
DatGuyJonathan: Did you find out what this does and if we need it?

packages/salesforcedx-apex-debugger/test/unit/commands/frameCommand.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    };
+
+    await frameCommand.execute();
+
+    expect(sendRequestSpy.calledOnce).to.equal(true);
+    expect(sendRequestSpy.getCall(0).args[0]).to.deep.equal(expectedOptions);
+  });
+
+  it('Should handle run command error', async () => {
```
DatGuyJonathan: Is this covered by baseCommand.test.ts? Same question with the negative test in references.test.ts

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: No. This needs to identify the first frame only. The state response includes variables information for the first frame (index == 0).

packages/salesforcedx-vscode-apex-debugger/package.nls.json
guw: @DatGuyJonathan Should we add more logging to existing areas and add them here as well?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: We can replace the code with a message object `codeOrMessage: number | DebugProtocol.Message`. Would you prefer a message object?

packages/salesforcedx-apex-debugger/test/unit/commands/frameCommand.test.ts
guw: Do you mean moving it into baseCommand tests?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: Actually, I'll replace with a log message and simply return an empty result here.

vazexqi: In general if we have a type of the form `TypeAOrTypeB` that means we should be explicit about it as being a _disjoint_ union rather than just a union. I would try to avoid them. But, if they are necessary, we should use an explicit `Either` like what Scala does. See https://alvinalexander.com/scala/scala-either-left-right-example-option-some-none-null

Is there a use case where we should introduce the generic Either<T> type into our code base? If there is, we should instead of propagating a bunch of `TypeAOrTypeB` types.

packages/salesforcedx-apex-debugger/src/commands/baseCommand.ts
```diff
@@ -14,19 +15,22 @@ export abstract class BaseCommand {
   private readonly accessToken: string;
   private readonly debuggedRequestId: string;
   private readonly debuggerApiPath = 'services/debug/v41.0';
```
vazexqi: I know this has existed before this CL, could the number 41.0 be pulled into a constants file somewhere?

packages/salesforcedx-apex-debugger/src/commands/baseCommand.ts
```diff
@@ -47,7 +51,8 @@ export abstract class BaseCommand {
         'Content-Type': 'application/json',
```
vazexqi: This one is also not part of this CL but since I just saw this the other day in @GaryFNC 's PR, I will ask again here. Is there a reason why Content-Type is quoted but Accept and Authorization are not? Should we be consistent and quote everything?

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -94,6 +94,11 @@
                 "type": "string",
                 "description": "%sfdx_project_text%",
                 "default": "${workspaceRoot}"
+              },
+              "trace": {
+                "type": ["boolean", "string"],
```
vazexqi: Could you elaborate on what `"type": ["boolean", "string"]` means? Why do we want to take two kinds of distinct types?

packages/salesforcedx-apex-debugger/src/commands/baseCommand.ts
guw: Is this really necessary? This basically *is* the single constant used for all debugger api calls. We can update it here when necessary.

Are you thinking of some global constant class/interface for all code?

packages/salesforcedx-apex-debugger/src/commands/baseCommand.ts
guw: @vazexqi This is a side effect of code formatting/cleanup. Saving the file removes unnecessary quotes.

packages/salesforcedx-vscode-apex-debugger/package.json
guw: you can set `trace: true`, which is equal/shortcut to `trace: 'all'`

packages/salesforcedx-apex-debugger/test/unit/commands/frameCommand.test.ts
DatGuyJonathan: I mean you already have a negative test like this in baseCommand.test.ts

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,11 +1450,27 @@ export class ApexDebug extends DebugSession {
   }
 
   private handleStopped(message: DebuggerMessage): void {
+    const reason = message.sobject.BreakpointId ? 'breakpoint' : 'step';
     const threadId = this.getThreadIdFromRequestId(message.sobject.RequestId);
+
+    this.log(
+      'la',
```
DatGuyJonathan: Should this be a different log category?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -54,13 +67,362 @@ import os = require('os');
 
 export interface LaunchRequestArguments
   extends DebugProtocol.LaunchRequestArguments {
+  /** comma separated list of trace selectors. Supported:
+	  * 'all': all
+	  * 'la': launch/attach
+	  * 'vh': variable handling
```
DatGuyJonathan: va or vh? I see va being used below. Can we use an enum instead so we don't have to type these as strings and prevent typos like this.

packages/salesforcedx-apex-debugger/src/commands/baseCommand.ts
vazexqi: Into a file called Constants.ts

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
vazexqi: Change this to use full words.
la and vh do not make any sense to me.

packages/salesforcedx-vscode-apex-debugger/package.nls.json
DatGuyJonathan: It looks like you mainly trace variable handling for now. You mentioned 'dap' somewhere in there; I think that could be used to trace the order of initializeRequest, attachRequest, launchRequest, setBreakPointsRequest, etc. We can also trace debugger events in handleEvent(). Sometimes we disconnect and reconnect from streaming api, so we can log that as well.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -664,7 +1379,7 @@ export class ApexDebug extends DebugSession {
       logMessage += ` |${os.EOL}${message.sobject.Stacktrace}`;
     }
 
-    this.logToDebugConsole(
+    this.printToDebugConsole(
```
vazexqi: Why the name change? I thought logToDebugConsole was OK.

packages/salesforcedx-vscode-apex-debugger/package.json
vazexqi: 👍 but let's change the type from 'va', 'dap' to full words.

packages/salesforcedx-vscode-apex-debugger/package.nls.json
guw: `'protocol'` (former `'dap'`) is used to enable tracing of the communication between VS Code and the debug adapter.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: I thought the name would make it more clear that this is not the preferred way of logging/tracing in the code but more a utility method for printing something to the DebugConsole.

packages/salesforcedx-vscode-apex-debugger/package.json
guw: done

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,11 +1458,30 @@ export class ApexDebug extends DebugSession {
   }
 
   private handleStopped(message: DebuggerMessage): void {
+    const reason = message.sobject.BreakpointId ? 'breakpoint' : 'step';
     const threadId = this.getThreadIdFromRequestId(message.sobject.RequestId);
+
+    this.log(
+      TRACE_CATEGORY_LAUNCH,
+      `handleStopped: got ${reason} event from server for thread ${threadId}`
```
DatGuyJonathan: This doesn't seem to fit the launch category.

packages/salesforcedx-vscode-apex-debugger/package.nls.json
```diff
@@ -1,9 +1,15 @@
...
+    "A regular expression matching entry points whitelisted for debugging",
   "sfdx_project_text": "Location of SFDX project",
   "launch_snippet_label_text": "Apex Debugger: Launch",
-  "launch_snippet_description_text": "A new configuration for launching Apex Debugger",
-  "launch_snippet_name": "Launch Apex Debugger"
+  "launch_snippet_description_text":
+    "A new configuration for launching Apex Debugger",
+  "launch_snippet_name": "Launch Apex Debugger",
+  "trace_text":
+    "Produce diagnostic output. Instead of setting this to true you can list one or more selectors separated with commas. The 'all' selector enables very detailed output. Other selectors are 'protocol', 'variables', 'launch'"
```
DatGuyJonathan: For the ones that are unused, do you plan on adding them in this PR or in another work item?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -695,6 +1415,17 @@ export class ApexDebug extends DebugSession {
       this.logEvent(message);
       this.requestThreads.delete(threadId);
       this.sendEvent(new ThreadEvent('exited', threadId));
+
+      // cleanup everything that's no longer necessary after all request finished
```
DatGuyJonathan: We should have tests that verify when we clear these handles.

packages/salesforcedx-apex-debugger/src/commands/requestService.ts
```diff
@@ -80,15 +80,28 @@ export class RequestService {
...
+          }
+        : {
+            type: 'POST',
+            url: requestUrl,
+            headers: {
+              'Content-Type': 'application/json',
+              Accept: 'application/json',
+              Authorization: `OAuth ${this.accessToken}`
+            },
+            data: JSON.stringify(command.getRequest())
```
DatGuyJonathan: Could we use if-else shorthand here?

DatGuyJonathan:
@guw , thanks for triaging the blitz issues. Let's start fixing the test failures.

packages/salesforcedx-apex-debugger/src/commands/requestService.ts
guw: I started from that. However, that puts an empty "data" into the options which would require rework of all existing tests. Not sure if it's worth it.

packages/salesforcedx-vscode-apex-debugger/package.nls.json
guw: They are all in use right now. New ones should be added as we go.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: Would you think that "lifecycle" is a better term than "launch"? The launch category is used for general debugger events - launched/started/connected/stopped

DatGuyJonathan: I think a category specifically for apex debugger events would fit better.

DatGuyJonathan:
@guw , are there test cases that can be copied from IDE1 to boost the coverage of apexDebug.ts? It went down by 30%.

https://codecov.io/gh/forcedotcom/salesforcedx-vscode/src/gunnar%2Fdebugger/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts

guw:
@DatGuyJonathan I've create more tests. Slowly getting back to 70% coverage in that file (which I consider the low water mark). I think a couple for for correct variable expansion and we should be good to go.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -59,15 +71,376 @@ import {
...
+        ? 'null'
+        : `null [${value.nameForMessages}]`;
+    }
+
+    if (ApexVariable.isString(value)) {
+      // We want to explicitly quote string values like in Java. This allows us to differentiate null from 'null'.
+      return `'${value.value}'`;
+    }
+
+    return `${value.value} [${value.nameForMessages}]`;
```
DatGuyJonathan: VS Code debugger UI gets crowded pretty quickly because we can't move them away from the left hand side. I'm more in favor of showing the type using https://github.com/Microsoft/vscode-debugadapter-node/blob/master/protocol/src/debugProtocol.ts#L1260.

DatGuyJonathan:
@guw, since you're out next week, please focus your efforts on Friday to get this checked in to develop branch. Create separate work items for any new bug you find. Document any gotchas so I can seamlessly pick up the bugs next week. Thanks!

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: Done.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -424,14 +872,279 @@ export class ApexDebug extends DebugSession {
...
+        'Static',
+        this.variableHandles.create(new ScopeContainer('static', frameInfo)),
+        false
+      )
+    );
+    scopes.push(
+      new Scope(
+        'Global',
+        this.variableHandles.create(new ScopeContainer('global', frameInfo)),
+        true
```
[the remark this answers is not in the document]
guw: Good catch. They should all be true because be potentially reach out to the server.

guw:
@DatGuyJonathan I'm waiting for a green build then this is ready to go in!

PR #131, opened 2017-09-21
James/display org username

Sweetman:
### What does this PR do?
Allow user to display org details for a specific username
@W-4285382@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -280,6 +280,10 @@
       {
         "command": "sfdx.force.org.display.default",
         "title": "%force_org_display_default_text%"
+      },
+      {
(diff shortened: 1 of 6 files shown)


Sweetman:
@DatGuyJonathan I spoke to Nick yesterday about the command and we will use the input box for now. If the quickpick is feasible, it will not be a breaking change but rather a UI enhancement.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -272,6 +272,10 @@
       {
         "command": "sfdx.force.org.display.default",
         "title": "%force_org_display_default_text%"
+      },
+      {
+        "command": "sfdx.force.org.display.username",
```
DatGuyJonathan: Does this work without SFDX folder?

PR #132, opened 2017-09-21
Validate tag and release version before publishing

chowwinston:
### What does this PR do?
Validates git tags and release version before publishing

### DIFF
--- a/scripts/publish.js
+++ b/scripts/publish.js
@@ -10,7 +10,11 @@ shell.set('+v');
  * 1. The script is running locally - it's not optimized for Travis workflow
  *    yet.
  * 2. The script is running in the right branch (e.g., release/vxx.y.z)
- * 3. You have set the next version SALESFORCEDX_VSCODE_VERSION as an environment variable
+ * 
+ * Instructions: 
+ * Run this script with SALESFORCEDX_VSCODE_VERSION as an environment variable
+ * i.e. SALESFORCE_VSCODE_VERSION=x.y.z ./scripts/publish.js
+ * 
  */
 
 // Checks that you are running this with Node v7.9.0 and above
@@ -64,7 +68,7 @@ if (!nextVersion) {
 const checkTags = shell.exec('git tag', { silent: true }).stdout;
 if (checkTags.includes(nextVersion)) {
   console.log(
-    `A git tag already exists. Try running \'git tag -d v${nextVersion}\'`
+    'There is a conflicting git tag. Reclone the repository and start fresh to avoid versioning problems.'
   );
   exit(-1);
 }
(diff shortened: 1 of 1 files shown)


scripts/publish.js
```diff
@@ -46,6 +47,28 @@ if (!publishers.includes('salesforce')) {
...
 
+// Checks that you have specified the next version as an environment variable, and that it's properly formatted.
+const nextVersion = process.env['SALESFORCEDX_VSCODE_VERSION'];
+if (!nextVersion) {
+  console.log(
+    'You must specify the next version of the extension by setting SALESFORCEDX_VSCODE_VERSION as an environment variable.'
+  );
+  exit(-1);
+} else {
+  const [version, major, minor, patch] = nextVersion.match(
```
chowwinston: Should we do a try catch here? If the version is malformed it fails with a cryptic error `TypeError: Cannot read property 'Symbol(Symbol.iterator)' of null`

scripts/publish.js
```diff
@@ -46,6 +47,28 @@ if (!publishers.includes('salesforce')) {
...
+    /^(\d+)\.(\d+)\.(\d+)$/
+  );
+}
+
+// Checks that a tag of the next version doesn't already exist
+const checkTags = shell.exec('git tag', { silent: true }).stdout;
+if (checkTags.includes(nextVersion)) {
+  console.log(
+    `A git tag already exists. Try running \'git tag -d v${nextVersion}\'`
```
vazexqi: I would be more precise here and say you probably should reclone the whole repository and start fresh to avoid any problems. That's how an autobuild will do it and that's how we should do it.

scripts/publish.js
```diff
@@ -46,6 +47,28 @@ if (!publishers.includes('salesforce')) {
   exit(-1);
 }
 
+// Checks that you have specified the next version as an environment variable, and that it's properly formatted.
+const nextVersion = process.env['SALESFORCEDX_VSCODE_VERSION'];
+if (!nextVersion) {
+  console.log(
+    'You must specify the next version of the extension by setting SALESFORCEDX_VSCODE_VERSION as an environment variable.'
```
vazexqi: Put at the top of this file to run with `SALESFORCE_VSCODE_VERSION=x.y.z ./scripts/publish.js`

That should be a good example.

scripts/publish.js
chowwinston: @vazexqi

vazexqi: No, it's OK. This is a script meant for a limited people to execute. It just needs to fail fast. Errors that are bit cryptic are OK.

vazexqi:
@chowwinston - Could you try this locally to make sure that it works (comment out the actual publishing step, of course :-)).

chowwinston:
I'll double check with the new changes. Already tested it before I created the PR

vazexqi:
@chowwinston - You've accidentally committed more files to this PR. You have bumped up all the version numbers. Please undo that and only check in the relevant changes.

vazexqi:
@chowwinston - I thought that we also wanted to ensure that you are on the release/v$[SALESFORCEDX_VSCODE_VERSION] git branch for even better safety?

PR #134, opened 2017-09-21
Use vscode's http proxy setting

DatGuyJonathan:
### What does this PR do?
There are three parts in the debugger that would concern a proxy: CLI calls for Apex Debugger objects (child_process lib), Streaming API (faye lib), and HTTP calls to Apex Debugger APIs (request-light lib). All those readily accept HTTP_PROXY and HTTPS_PROXY environment variables. VS Code doc suggests extensions could handle http proxy settings https://code.visualstudio.com/docs/setup/network, so this PR is for that.

@W-4315069@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -27,6 +27,7 @@ import {
   DebuggerResponse,
   ForceOrgDisplay,
   OrgInfo,
+  RequestService,
   RunCommand,
   StateCommand,
   StepIntoCommand,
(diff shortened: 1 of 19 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -86,6 +94,7 @@ export class ApexDebug extends DebugSession {
   ): void {
     this.myBreakpointService.clearSavedBreakpoints();
     this.initializedResponse = response;
+    this.sendEvent(new Event(GET_PROXY_SETTINGS_EVENT));
```
DatGuyJonathan: Requesting proxy settings from the vscode-apex-debugger module since it's accessible through vscode.workspace.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -438,6 +432,12 @@ export class ApexDebug extends DebugSession {
       case HOTSWAP_REQUEST:
         this.warnToDebugConsole(nls.localize('hotswap_warn_text'));
         break;
+      case PROXY_SETTINGS_REQUEST:
```
DatGuyJonathan: Proxy settings from vscode-apex-debugger. Used in requestService.ts, streamingClient.ts, sessionService.ts, & breakpointService.ts.

PR #140, opened 2017-09-26
Fix packaging issue

vazexqi:
### What does this PR do?

Puts secondary level packages as top-level dependencies.
Moves real dependencies from `devDepedencies` section to `dependencies` section.

@W-4345609@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -10,21 +10,21 @@
     "vscode": "^1.15.0"
   },
   "dependencies": {
-    "@salesforce/salesforcedx-utils-vscode": "40.11.0"
+    "@salesforce/salesforcedx-utils-vscode": "40.11.0",
+    "request-light": "0.2.1",
+    "rxjs": "^5.4.1",
+    "shelljs": "0.7.8",
+    "tree-kill": "^1.1.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
-    "@types/jsforce": "^1.7.1",
     "@types/shelljs": "0.7.4",
     "@types/mocha": "2.2.38",
     "@types/node": "^6.0.40",
     "@types/rimraf": "^2.0.2",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
-    "request-light": "0.2.1",
     "cross-env": "5.0.4",
-    "jsforce": "1.7.1",
-    "shelljs": "0.7.8",
     "mocha": "3.2.0",
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -10,21 +10,21 @@
     "vscode": "^1.15.0"
   },
   "dependencies": {
-    "@salesforce/salesforcedx-utils-vscode": "40.11.0"
+    "@salesforce/salesforcedx-utils-vscode": "40.11.0",
+    "request-light": "0.2.1",
```
vazexqi: Hoisting these up to the top-level will help with packaging albeit at the cost of some duplicated packages.

The crux is that A -> B -> C. This transitive dependency doesn't bring in all the dependencies when we are using lerna (since it's locally linked).

We can avoid this in the future if we publish everything to the public npm - that would resolve this issue but that would mean that we would need to publish publicly which puts another burden on our ops.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -10,21 +10,21 @@
...
+    "rxjs": "^5.4.1",
+    "shelljs": "0.7.8",
+    "tree-kill": "^1.1.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
-    "@types/jsforce": "^1.7.1",
```
vazexqi: This is no longer used.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -80,7 +80,6 @@
     ]
   },
   "dependencies": {
-    "@salesforce/salesforcedx-sobjects-faux-generator": "40.11.0",
```
vazexqi: This is no longer used.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -10,21 +10,21 @@
...
     "@types/shelljs": "0.7.4",
     "@types/mocha": "2.2.38",
     "@types/node": "^6.0.40",
     "@types/rimraf": "^2.0.2",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
-    "request-light": "0.2.1",
```
vazexqi: This is actually used outside of tests so it needs to be moved to the dependencies section.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -10,21 +10,21 @@
...
     "cross-env": "5.0.4",
-    "jsforce": "1.7.1",
-    "shelljs": "0.7.8",
```
vazexqi: This is actually used outside of tests so it needs to be moved to the dependencies section.

PR #142, opened 2017-09-27
Update breakpoint reconcile strategy

DatGuyJonathan:
### What does this PR do?
If you enable breakpoints on lines 1, 2, 3, VS Code sends three requests to the adapter. Request 1 wants a breakpoint on line 1. Request 2 wants breakpoints on lines 1 & 2. Request 3 wants breakpoints on lines 1, 2, & 3. 

Each breakpoint request is not handled individually from start to finish. Due to nodejs event loop, a breakpoint request will be put aside when it spawns an SFDX CLI process to add/delete breakpoint. The next breakpoint request will start getting handled. 

We also need to watch out for when we delete & add quickly, and vice versa. For example, if you click on line 1 to delete, then click it again to add, the first request will kick off a breakpoint delete with the CLI. The second request starts getting handled, but the adapter already has line 1 in its cache because the delete operation hasn't finished yet, so the adapter thought the add was a no-op.

That means the breakpoint service needs to remove duplicates, reconcile the client's list of breakpoints vs. cached breakpoints to know which to add and delete, and also reconcile with in-flight breakpoint operations.

@W-4336889@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -188,67 +188,34 @@ export class ApexDebug extends DebugSession {
     response: DebugProtocol.SetBreakpointsResponse,
     args: DebugProtocol.SetBreakpointsArguments
   ): Promise<void> {
-    if (args.source && args.source.path) {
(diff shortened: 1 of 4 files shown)


guw:
@DatGuyJonathan are you using the anonymous GitHub email? Can you associate that with the CLA bot so that we don't get the redundant message?

PR #144, opened 2017-09-27
Fix debugger capabilities and stop command

DatGuyJonathan:
### What does this PR do?
Disable debugger capabilities that we don't support. Remove Apex Debugger stop command from right-click menu in the project explorer. Clean up the notifications from the stop command.

@W-4337014, W-4331254@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -90,5 +90,8 @@ export const messages = {
   force_alias_list_text: 'SFDX: List All Aliases',
   force_org_display_default_text:
     'SFDX: Display Org Details for Default Scratch Org',
-  force_org_display_username_text: 'SFDX: Display Org Details...'
+  force_org_display_username_text: 'SFDX: Display Org Details...',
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -90,5 +90,8 @@ export const messages = {
   force_alias_list_text: 'SFDX: List All Aliases',
   force_org_display_default_text:
     'SFDX: Display Org Details for Default Scratch Org',
-  force_org_display_username_text: 'SFDX: Display Org Details...'
+  force_org_display_username_text: 'SFDX: Display Org Details...',
+  force_debugger_query_session_text: 'query for Apex Debugger session',
+  force_debugger_stop_text: 'SFDX: Stop Apex Debugger Session',
+  force_debugger_stop_none_found_text: 'No Apex Debugger session found.'
 };
```
DatGuyJonathan: The debugger session stop command in the command palette is a two-step process. First, we query for active session. If there is none, we show this message:

stopsessioncommandnoop

If there is an active session, we update its status to stop it.

stopsessioncommandsuccess

PR #146, opened 2017-09-27
Debugger whitelist

DatGuyJonathan:
### What does this PR do?
Change UserIdFilter into an array so user doesn't have to type commas. Change RequestTypeFilter into an array and suggest possible values. 

requesttypefiltersuggestion

Also, remove the stopped event reason so it defaults to just "Paused" in the callstack.

callstackpaused

@W-4314787, W-4314153@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -61,8 +61,8 @@ import os = require('os');
 
 export interface LaunchRequestArguments
   extends DebugProtocol.LaunchRequestArguments {
-  userIdFilter?: string;
-  requestTypeFilter?: string;
+  userIdFilter?: string[];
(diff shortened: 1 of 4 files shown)


vazexqi:
Thanks for doing the auto-completion for the Requests. It's must nicer from a usability standpoint!

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -744,6 +744,13 @@ export class ApexDebug extends DebugSession {
       );
     }
   }
+
+  private stringifyLaunchArg(arg?: string[]): string {
```
guw: I'd prefer a more descriptive name, eg. `toCommaSeparatedString`

guw:
FWIW, I believe to improve the diff hit metric you need to make the helper method public and add a test specific for it.

PR #148, opened 2017-09-28
Update travis.yml and appveyor.yml to use latest CLI

vazexqi:
### What does this PR do?

Changes travis.yml to use the latest version of the CLI

Part of the preparation for the V6 CLI.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,7 +22,8 @@ install:
   - 7z x sfdx-windows-amd64.tar.xz
   - 7z x sfdx-windows-amd64.tar
   - SET PATH=%APPVEYOR_BUILD_FOLDER%\sfdx\bin;%PATH%
-  - sfdx update
+  - sfdx update alpha
+  - sfdx plugins:install salesforcedx@41.2.0
   - npm install -g codecov 
   # the JWT key should be stored base 64 encoded in AppVeyor settings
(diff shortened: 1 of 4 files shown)


packages/system-tests/package.json
```diff
@@ -15,7 +15,7 @@
     "@types/node": "^6.0.40",
     "@types/rimraf": "0.0.28",
     "@types/shelljs": "^0.7.4",
-    "@types/webdriverio": "^4.6.1",
+    "@types/webdriverio": "4.6.1",
```
vazexqi: @chowwinston, @Sweetman, @GaryFNC, @DatGuyJonathan - Just FYI. When I compile this locally, I saw some type errors with TypeScript 2.4.0 if I let it use the next (that is what the ^ signifies) version of types. So, I'm locking it down to 4.6.1 for now.

.appveyor.yml
```diff
@@ -22,7 +22,7 @@ install:
   - 7z x sfdx-windows-amd64.tar.xz
   - 7z x sfdx-windows-amd64.tar
   - SET PATH=%APPVEYOR_BUILD_FOLDER%\sfdx\bin;%PATH%
-  - sfdx update
+  - sfdx update alpha
```
vazexqi: We will remove the alpha once R2 goes out since then it would be the default and no longer alpha. Ditto for the other CI files.

vazexqi:
Waiting for our test instance that is no NA50 to be updated this weekend.

vazexqi:
Travis on Mac failed for what seems like a slow run (I've tested locally).
I'm going to merge this in since we should be testing with the v6 version of our CLI.
Some changes might be necessary on the travis.yml and appveyor.yml when the cli is no longer on the alpha channel (we do a `sfdx update alpha`).

PR #149, opened 2017-09-28
Rene/sfdx soql query

### LINKED ISSUE #65: Support SOQL queries via SFDX

Sweetman:
### What does this PR do?

Implements UI for sfdx force:data:soql:query

Option 1:
User has no editor open or no text selected => a VS code input is used to gather user input

Option 2:
User has text selected => selected text is used as SOQL query

E2E tests are implemented.

In addition the utils class from @chowwinston's faux package has been moved to the utils-vscode package.

Fixes #65

### DIFF
--- a/packages/system-tests/scenarios/forceSoqlQueryUi.test.ts
+++ b/packages/system-tests/scenarios/forceSoqlQueryUi.test.ts
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// tslint:disable:no-unused-expression
+import * as util from '@salesforce/salesforcedx-utils-vscode/out/src/test/orgUtils';
+import { expect } from 'chai';
+import * as path from 'path';
+import { CommonActions } from '../src/areas/common';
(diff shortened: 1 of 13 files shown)


packages/system-tests/scenarios/forceSoqlQueryUi.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    for (let i = 0; i < consoleHtml.length; i++) {
+      if (consoleHtml[i].indexOf('xit&nbsp;code') > 0) {
+        expect(consoleHtml[i]).to.contain('exit&nbsp;code&nbsp;0');
+      }
+    }
+  });
+
+  it('Should execute SOQL query with current selection', async () => {
+    // Open new untitled file
+    await app.command('workbench.action.files.newUntitledFile');
```
Sweetman: @muenzpraeger The issue with the test was that on Windows, the app would open with no editor available like it does on Mac/Linux. This meant nothing could be highlighted and so the command wouldn't show

Sweetman:
@muenzpraeger Thanks for doing the bulk of the work on this 👍

PR #150, opened 2017-09-28
Fix null field showing up in generated SObject faux classes

GaryFNC:
### What does this PR do?
When fields are generated in the SObject faux class, there were situations
where the relationshipName was not always specified.  In that case, there is
a field/childRelationship name that usually ends in “Id”.
This fix uses that name if the relationshipName is missing and strips off the “Id”.
This is consistent with what Apex and IlluminatedCloud expect and do.

Resolves: @W-4344749@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -78,18 +78,25 @@ export class FauxClassGenerator {
     return Promise.resolve('');
   }
 
-  private generateChildRelationship(rel: ChildRelationship): string {
-    if (rel.relationshipName) {
-      return `List<${rel.childSObject}> ${rel.relationshipName}`;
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -78,30 +78,41 @@ export class FauxClassGenerator {
...
-        return `${rel.childSObject} ${nameWithoutId}`;
-      } else {
-        return '';
-      }
+      return name;
     }
   }
 
+  private getReferenceName(relationshipName: string, name: string): string {
+    if (relationshipName) {
```
vazexqi: `return relationshipName ? relationshipName : this.stripId(name);`

GaryFNC: yep, that is better - I always find that harder to read so I default to an "if" unless I think to do it

PR #151, opened 2017-09-29
First step at the Visualforce Language Server

vazexqi:
### What does this PR do?

This is the first step toward our own Visualforce language server.

This PR bundles the existing [HTML language server](https://github.com/Microsoft/vscode/tree/master/extensions/html) from VS Code. This mimics the behavior of the current Visualforce extension. The existing HTML extension actually uses a combination of HTML, JavaScript, and CSS language servers since you can have all three in a .html file (similarly you can do that too in a .page and .component file).

The old extension exists alongside this new extension so that we can easily have parity first before adding Visualforce specific features. This new extension is not published to the marketplace.

### What is out of the scope of this PR?

This PR doesn't make this the default language server for .page and .components yet. We will do that once we have done more extensive testing and also renaming of preferences (so that we don't share any with the html setting).

@W-4344488@


packages/salesforcedx-visualforce-language-server/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "vscode-nls": "2.0.2",
+    "vscode-uri": "1.0.1"
+  },
+  "devDependencies": {
+    "@types/mocha": "^2.2.38",
+    "@types/node": "^6.0.40",
+    "mocha": "3.2.0",
+    "shx": "0.2.2"
+  },
+  "scripts": {
```
vazexqi: @DatGuyJonathan - I removed the vscode:publish here so that should prevent it from being published to the marketplace for now. This is the main thing I want you to check (to see if I might have forgotten something).

vazexqi:
screen shot 2017-09-29 at 2 02 47 pm

Another issue to be aware of is that the current version of the HTML extension uses a proposed API for color provider (https://github.com/Microsoft/vscode/issues/34366). It looks like this proposed API might be stable in the next version. Without stabilization, the only lost of functionality is that we don't provide the color preview when you use a color name, hex, rgb in the .page and .component.

So if you run the extension in debug mode it will run.
But if you generate the .vsix and install it into Code, it will show that error.

Also, anything that uses `enableProposedApi` also cannot be published to the marketplace. Hence, why this won't be a replacement until later.

packages/salesforcedx-vscode-visualforce-next/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "compile": "tsc -p ./",
+    "lint": "tslint --project .",
+    "watch": "tsc -watch -p .",
+    "clean":
+      "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
+    "postinstall": "node ./node_modules/vscode/bin/install",
+    "test": "node ./node_modules/vscode/bin/test"
+  },
+  "activationEvents": ["workspaceContains:sfdx-project.json"],
+  "enableProposedApi": true,
```
vazexqi: This is where we enable the proposedApi

packages/salesforcedx-vscode-visualforce-next/typings/vscode.proposed.d.ts
```diff
@@ -0,0 +1,316 @@
...
+       * @return Disposable which unregisters this command on disposal.
+       */
+    export function registerDiffInformationCommand(
+      command: string,
+      callback: (diff: LineChange[], ...args: any[]) => any,
+      thisArg?: any
+    ): Disposable;
+  }
+
+  /**
```
vazexqi: This line and below are the proposed color API that we are using. The above are not being used.

This vscode.proposed.d.ts was copied from this particular version https://github.com/Microsoft/vscode/commit/a7ad71be3bf9c1db4ad247d9a1ac39738540c13a#diff-4878e8b25d41f2ab7c2715852b1b97aa

packages/salesforcedx-vscode-visualforce-next/typings/vscode.proposed.d.ts
```diff
@@ -0,0 +1,316 @@
+/*---------------------------------------------------------------------------------------------
```
vazexqi: The idea of embedding this file is from https://github.com/Microsoft/vscode/issues/25784

PR #153, opened 2017-10-02
Fix SObject field types that were not properly mapped to Apex

GaryFNC:
### What does this PR do?

Fix SObject field types that were not mapped properly to Apex  …
This is corrected in FauxClassGenerator and test added for all the different
field types and relationships
Also, added a test for the batch version of describeSObjectBatch
Moved the “contributes” for the SObject gate to vscode-core so that it  …
shows up when the .vsix is produced
Added tests for __mdt and __e objects

@W-4354915, W-4346259@


packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
```diff
@@ -26,17 +26,33 @@ describe('generate fields set', function() {
     expect(classText).to.include('String Foo;');
```
vazexqi: Please take a look at the other tests classes and

1. See how the file is named.
1. See how the it(....) clauses are named.

Let's be consistent.

vazexqi:
1. Please fix the way you are referencing issues. @chowwinston - please help gary with being consistent (and in this case, correct, or else the auto-linker won't pick up the issue)
1. Could you try the end-to-end generation of the .vsix to ensure that we have no issues with the configuration?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
...
+    ['url', 'String'],
+    ['id', 'Id'],
+    // note that the mappings below "id" only occur in standard SObjects
+    ['base64', 'Blob'],
+    ['address', 'Address'],
+    ['int', 'Integer'],
+    ['anyType', 'Object'],
+    ['combobox', 'String'],
+    ['time', 'Time'],
+    // TBD what are these mapped to and how to create them
```
vazexqi: What is the severity/frequency of these two types occurring?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -86,12 +119,17 @@ export class FauxClassGenerator {
     }
   }
 
+  private capitalize(input: string): string {
+    return input.charAt(0).toUpperCase() + input.slice(1);
+  }
+
+  private getGenType(describeType: string): string {
```
vazexqi: Do not use short names in method names. Expand `gen`

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -37,5 +37,8 @@
   "force_refresh_sobjects": "SFDX: Refresh SObject Definitions",
   "force_data_soql_query_input_text": "SFDX: Execute SOQL Query...",
   "force_data_soql_query_selection_text":
-    "SFDX: Execute SOQL Query with Currently Selected Text"
+    "SFDX: Execute SOQL Query with Currently Selected Text",
+  "gating_title": "Salesforce Feature Previews",
```
vazexqi: Please choose a better name than gating_title.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -93,6 +101,26 @@ describe('Fetch sObjects', function() {
     expect(customField.name).to.be.equal('MyCustomField__c');
   });
 
+  it('Should be able to call describeSObjectBatch on custom objects', async function() {
+    const cmdOutput = await sobjectdescribe.describeSObjectBatch(
+      process.cwd(),
+      [CUSTOM_OBJECT_NAME, CUSTOM_OBJECT2, CUSTOM_OBJECT3],
```
vazexqi: I think we need better names than CustomObject1, CustomObject2,...
Anyone reading this code would not have an idea what distinguishes Object1 from Object2 and how to maintain modify in the future.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
```diff
@@ -93,6 +101,26 @@ describe('Fetch sObjects', function() {
...
+      0,
+      username
+    );
+    expect(cmdOutput[0].name).to.be.equal(CUSTOM_OBJECT_NAME);
+    expect(cmdOutput[0].custom).to.be.true;
+    expect(cmdOutput[0].fields.length).to.be.least(9);
```
vazexqi: No magic numbers.

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
```diff
@@ -26,17 +26,33 @@ describe('generate fields set', function() {
...
+      '{"name": "Foo12", "type" : "textarea", "referenceTo": []}',
+      '{"name": "Foo13", "type" : "encryptedstring", "referenceTo": []}',
+      '{"name": "Foo14", "type" : "url", "referenceTo": []}',
+      '{"name": "Foo15", "type" : "id", "referenceTo": []}'
+    ];
+
+    const fieldsString = fields.join(',');
+    const sobject1 = `${fieldsHeader}${fieldsString}${closeHeader}`;
+
     const sobjectFolder = './';
```
vazexqi: Is this safe to use? This seems platform-specific. 
I am a bit surprised that the tests on Appveyor pass. Are you using the `./` in any meaningful way in the tests?

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
GaryFNC: will do - this one has been out there for a while, but I'll be glad to change it for consistency.

GaryFNC:
wrt referencing issues - I followed the pattern I saw in other PRs (ex #146).   Yes, let me know what I'm doing wrong - be glad to do it right...

I did the .vsix and verified that things worked properly even just with loading the .vsix for salesforce-vscode-core.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: 1. I have run all the standard SObjects that I have been able to (I need to try some other perms, etc. to see if I can get some others.   I don't think we still have access to the org that had everything that Josh knew about)....     I can't find these in any of those describes.
2. IF they show up, they will create a "type" for the field that is the name with a capital first.  It "probably" isn't correct, but at least it won't be "undefined" or null...      
3. Since I haven't been able to create these in the standard SObjects, I don't think that the risk is very big - although I still plan to circle back at some point and see if there is a way to create them.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: I'll follow what you want for a style.  I personally find the sometimes very long names to be unreadable in practice...

packages/salesforcedx-vscode-core/package.nls.json
GaryFNC: feature_previews_title ??    Unclear to me why this matters as it is strictly internal and describes what is happening accurately...   But I can change it...

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
GaryFNC: There isn't anything that distinguishes these except their name.   I can say CustomObject1Name if you'd like.    They are simply there to manifest enough objects to show whether a batch works.

packages/salesforcedx-sobjects-faux-generator/test/integration/sObjectDescribe.test.ts
GaryFNC: Are you referring to the 9 fields?   Winston put that in earlier and I was just following it.   Do you want it to be a constant instead with a name?

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
GaryFNC: Good catch - I copied it from somewhere else and didn't notice..   I'll change it to use process.cwd().

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
...
+  // usually multiple fields with specialized names
+  private static typeMapping: Map<string, string> = new Map([
+    ['string', 'String'],
+    ['double', 'Decimal'],
+    ['reference', ''],
+    ['boolean', 'Boolean'],
+    ['currency', 'Currency'],
+    ['date', 'Date'],
+    ['datetime', 'Datetime'],
+    ['email', 'Email'],
```
chowwinston: Email should be type String

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
...
+    ['id', 'Id'],
+    // note that the mappings below "id" only occur in standard SObjects
+    ['base64', 'Blob'],
+    ['address', 'Address'],
+    ['int', 'Integer'],
+    ['anyType', 'Object'],
+    ['combobox', 'String'],
+    ['time', 'Time'],
+    // TBD what are these mapped to and how to create them
+    //['calculated', 'xxx'],
```
chowwinston: Calculated is for formula fields, where the type can be vary depending on what kind of formula it is. If you just get the type of the field, it should be limited to checkbox, currency, date, datetime, number, percent, text, and time. These are already mapped above so this should be safe to ignore.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
...
   private CUSTOMOBJECTS_DIR = 'customObjects';
 
+  // the empty string is used to represent the need for a special case
+  // usually multiple fields with specialized names
+  private static typeMapping: Map<string, string> = new Map([
+    ['string', 'String'],
+    ['double', 'Decimal'],
+    ['reference', ''],
+    ['boolean', 'Boolean'],
+    ['currency', 'Currency'],
```
chowwinston: Currency should be decimal

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
   private STANDARDOBJECTS_DIR = 'standardObjects';
   private CUSTOMOBJECTS_DIR = 'customObjects';
 
+  // the empty string is used to represent the need for a special case
+  // usually multiple fields with specialized names
+  private static typeMapping: Map<string, string> = new Map([
+    ['string', 'String'],
+    ['double', 'Decimal'],
```
chowwinston: Double is a valid type. Why is it converted to decimal?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
...
+    ['reference', ''],
+    ['boolean', 'Boolean'],
+    ['currency', 'Currency'],
+    ['date', 'Date'],
+    ['datetime', 'Datetime'],
+    ['email', 'Email'],
+    ['location', 'Location'],
+    ['percent', 'Decimal'],
```
chowwinston: Percent should be double https://developer.salesforce.com/docs/atlas.en-us.208.0.object_reference.meta/object_reference/field_types.htm

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: This one is interesting..    Email is not a built-in type in Apex, but it is in the Standard Apex Library.  I was following the pattern used by Illuminated Cloud which maps this to a reference to the Email class....    Not totally sure which is the right approach...

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: agreed - I tried a number of different formula fields and they gave the appropriate results - not sure whether calculated was used at one time and not any more or whether there is something out there somewhere that still uses it - but I couldn't find anything

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: again, I was following the IlluminatedCloud pattern - so we need to decide whether IC is correct in using System.Currency or whether this should just be an Apex primitive type

packages/salesforcedx-sobjects-faux-generator/test/fauxgeneratortest.ts
```diff
@@ -176,4 +238,105 @@ describe('generate relationship tests', function() {
...
+    expect(classText).to.include('String ExtRef__c');
+  });
+
+  // Note - hierarchical relationship, many-to-many, and master-detail all look like lookup relationships in terms of
+  //  the relevant parts of describe
+  // For some of these, can create another test with extra info that is present, but it isn't paid attention to
+  // Might be good for completeness in case this other info eventually is relevant, but will  not be addressed currently
+});
+
+// Note, currently __x (ExternalObject) is not handled by describe (REST or Cli), but is handled by SDD
```
chowwinston: I'd just take this out. If externalObject isn't an sobject in terms of list/describe, we don't need to worry about it.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
chowwinston: This can be taken out. The types of the calculated field are known.

chowwinston:
Just remove the extra comments and this looks good.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: Yes, but, both of these exist as legitimate describe field types, but I haven't been able to figure out how to produce them and what they should map to (if they can be produced at all).   Feels wrong to just leave them out when we know that they are missing.   Next person will have no clue if they ever show up

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      classPath = '';
+    }
+  });
+
+  it('Should generate a faux class with all types of fields that can be in custom SObjects', async () => {
+    const fieldsHeader = '{ "name": "Custom__c", "fields": [ ';
+    const closeHeader = ' ], "childRelationships": [] }';
+
+    const fields: string[] = [
+      '{"name": "Foo1", "type": "string", "referenceTo": []}',
```
vazexqi: Maybe we should call these like `StringType` instead of `Foo1`. It's more explicit what we are testing. It's a small change but I think it adds clarity. I do not like names like Foo1, Foo2, etc.

GaryFNC: will do

GaryFNC:
@vazexqi are there still changes that your require?  I'd like to get this merged in before folks change the develop branch again.   Assuming it passes the testing after this latest rebase from develop, it passed the blitz - be good to get it in.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -25,6 +25,39 @@ export class FauxClassGenerator {
   private STANDARDOBJECTS_DIR = 'standardObjects';
   private CUSTOMOBJECTS_DIR = 'customObjects';
 
+  // the empty string is used to represent the need for a special case
+  // usually multiple fields with specialized names
+  private static typeMapping: Map<string, string> = new Map([
```
vazexqi: If it is a specialized field, we should designate a proper sentinel to mark it. Using an empty '' string is bad. Are you doing special checks somewhere to check for '' strings?

vazexqi: Is this only used at 

```
private getTargetType(describeType: string): string {
 +    const gentype = FauxClassGenerator.typeMapping.get(describeType) as string;
 +    return gentype ? gentype : this.capitalize(describeType);
 +  }
``

PR #157, opened 2017-10-03
Add alias prompt and unit tests

Sweetman:
### What does this PR do?
Add alias prompt for force:org:create. The prompt will set the alias inputted by the user with the --setalias flag on the command. If user hits Enter with no alias inputted, the alias will be set to "scratchOrg" by default. If user hits Escape to cancel out, the command should not execute.
@W-4345929@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -26,9 +26,9 @@ import {
 
 export const DEFAULT_ALIAS = 'scratchOrg';
 export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
-  FileSelection & Alias
+  AliasAndFileSelection
 > {
-  public build(data: FileSelection & Alias): Command {
+  public build(data: AliasAndFileSelection): Command {
     const selectionPath = path.relative(
       vscode.workspace.rootPath!, // this is safe because of workspaceChecker
       data.file
@@ -47,9 +47,9 @@ export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
 
 export class AliasGatherer implements ParametersGatherer<Alias> {
   public async gather(): Promise<CancelResponse | ContinueResponse<Alias>> {
-    const aliasInputOptions = <vscode.InputBoxOptions>{
+    const aliasInputOptions = {
       prompt: nls.localize('parameter_gatherer_enter_alias_name')
-    };
+    } as vscode.InputBoxOptions;
     const alias = await vscode.window.showInputBox(aliasInputOptions);
     // Hitting enter with no alias will default the alias to 'scratchOrg'
     if (alias === undefined) {
@@ -64,9 +64,11 @@ export interface Alias {
   alias: string;
 }
 
+export type AliasAndFileSelection = Alias & FileSelection;
+
 const workspaceChecker = new SfdxWorkspaceChecker();
 const parameterGatherer = new CompositeParametersGatherer<
-  FileSelection & Alias
+  AliasAndFileSelection
 >(new FileSelector('config/**/*-scratch-def.json'), new AliasGatherer());
 
 export function forceOrgCreate() {
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -13,15 +13,22 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
 import {
+  CancelResponse,
+  CompositeParametersGatherer,
+  ContinueResponse,
   FileSelection,
   FileSelector,
+  ParametersGatherer,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
-class ForceOrgCreateExecutor extends SfdxCommandletExecutor<FileSelection> {
-  public build(data: FileSelection): Command {
+export const DEFAULT_ALIAS = 'scratchOrg';
+export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
+  FileSelection & Alias
+> {
+  public build(data: FileSelection & Alias): Command {
     const selectionPath = path.relative(
       vscode.workspace.rootPath!, // this is safe because of workspaceChecker
       data.file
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -32,13 +39,35 @@ class ForceOrgCreateExecutor extends SfdxCommandletExecutor<FileSelection> {
...
+      : { type: 'CONTINUE', data: { alias } };
+  }
+}
+export interface Alias {
+  alias: string;
+}
+
 const workspaceChecker = new SfdxWorkspaceChecker();
-const parameterGatherer = new FileSelector('config/**/*-scratch-def.json');
+const parameterGatherer = new CompositeParametersGatherer<
```
Sweetman: CompositeParametersGatherer will aggregate all the inputs gathered the type of the Gatherer should be the intersection of the interfaces

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import { nls } from '../../src/messages';
+
+// tslint:disable:no-unused-expression
+describe('Force Org Create', () => {
+  describe('Alias Gatherer', () => {
+    const TEST_ALIAS = 'testAlias';
+    let inputBoxSpy: sinon.SinonStub;
+
+    before(() => {
+      inputBoxSpy = sinon.stub(vscode.window, 'showInputBox');
```
Sweetman: Create a stub for InputBox. I want to simulate the 3 possible responses of a user hitting Escape (returns undefined), a user hitting Enter, and a user typing in an alias

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+  });
+
+  describe('Org Create Builder', () => {
+    it('Should build the org create command', async () => {
+      const CONFIG_FILE = 'configFile.txt';
+      const TEST_ALIAS = 'testAlias';
+      const forceOrgCreateBuilder = new ForceOrgCreateExecutor();
+      const createCommand = forceOrgCreateBuilder.build({
+        file: path.join(vscode.workspace.rootPath!, CONFIG_FILE),
```
Sweetman: The Executor will make find a path relative to the rootPath. The tests are set up to always open with the workspace right now so I put the non-null assertion operator

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -45,6 +45,7 @@ export const messages = {
   parameter_gatherer_enter_dir_name:
     'Enter desired directory (Press Enter to confirm or Esc to cancel)',
   parameter_gatherer_enter_username_name: 'Enter target username',
+  parameter_gatherer_enter_alias_name: 'Enter an alias',
```
Sweetman: @ruthemmanuelle may you verify that this text is ok?

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -13,15 +13,22 @@ import * as path from 'path';
...
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
-class ForceOrgCreateExecutor extends SfdxCommandletExecutor<FileSelection> {
-  public build(data: FileSelection): Command {
+export const DEFAULT_ALIAS = 'scratchOrg';
+export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
+  FileSelection & Alias
```
vazexqi: Why is it the intersection rather than the union? Also, you should define a type alias instead of repeating `FileSelection & Alias` everywhere.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -32,13 +39,35 @@ class ForceOrgCreateExecutor extends SfdxCommandletExecutor<FileSelection> {
...
       .withFlag('-f', `${selectionPath}`)
+      .withFlag('--setalias', data.alias)
       .withArg('--setdefaultusername')
       .build();
   }
 }
 
+export class AliasGatherer implements ParametersGatherer<Alias> {
+  public async gather(): Promise<CancelResponse | ContinueResponse<Alias>> {
+    const aliasInputOptions = <vscode.InputBoxOptions>{
```
vazexqi: http://acdcjunior.github.io/typescript-cast-object-to-other-type-or-instanceof.html

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Maybe 'Enter a scratch org alias', just to be 100% clear?

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
Sweetman: Union will OR the interfaces together which is not correct. The ContinueResponse should always have the fields from both interfaces since both a config file and alias are required.

I will make the type alias

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
Sweetman: Ok, I will change the casting syntax. I think I put this syntax in a few other classes so I will change them in this PR as well.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -13,15 +13,22 @@ import * as path from 'path';
...
   FileSelector,
+  ParametersGatherer,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
-class ForceOrgCreateExecutor extends SfdxCommandletExecutor<FileSelection> {
-  public build(data: FileSelection): Command {
+export const DEFAULT_ALIAS = 'scratchOrg';
```
vazexqi: @ruthemmanuelle - Is this an acceptable default name for scratchOrg? This is going to show up when you do `force:org:list`. Since we are setting this as the default for you if you don't provide one, is this OK or should we go with something more explicit like `defaultScratchOrg`?

Sweetman: Thanks for catching that Nick, meant to Ruth ask about this. Josh wanted to stay away from using "default" (and the like?) because if would be confusing to see an org with that alias and then see another org such as myFoo set as `defaultusername`. Not sure if it applies to defaultScratchOrg but something to think about

ruthemmanuelle: I'd like to set an alias that's not likely to overwrite an alias the user has already specified. Would something like `vscodeScratchOrg` work?

Sweetman: That works for me

Sweetman:
@DatGuyJonathan do you have any other comments for the PR?

PR #159, opened 2017-10-05
Lock breakpoint request

DatGuyJonathan:
### What does this PR do?
Use a lock library to make sure we only process one setBreakpointsRequest for each file at a time.

@W-4336889@

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -11,10 +11,12 @@
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "40.13.0",
+    "@types/async-lock": "0.0.20",
     "faye": "1.1.2",
     "request-light": "0.2.1",
     "vscode-debugadapter": "1.24.0",
-    "vscode-debugprotocol": "1.24.0"
+    "vscode-debugprotocol": "1.24.0",
+    "async-lock": "1.0.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -6,7 +6,7 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.16.0"
+    "vscode": "^1.17.0"
   },
   "categories": ["Debuggers"],
   "dependencies": {
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -468,7 +468,6 @@ export class ApexDebug extends LoggingDebugSession {
   private traceAll = false;
 
   private lock = new AsyncLock({ timeout: DEFAULT_LOCK_TIMEOUT_MS });
-  protected breakpointLockTimeout = DEFAULT_LOCK_TIMEOUT_MS;
 
   constructor() {
     super('apex-debug-adapter.log');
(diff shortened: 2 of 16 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -588,33 +598,53 @@ export class ApexDebug extends LoggingDebugSession {
...
-          line: this.convertDebuggerLineToClient(validBreakpoint)
-        });
-      });
-      args.lines.forEach(lineArg => {
-        if (!validBreakpoints.has(lineArg)) {
-          processedBreakpoints.push({
-            verified: false,
-            source: args.source,
-            line: this.convertDebuggerLineToClient(lineArg)
+    response = await this.lock.acquire(
```
DatGuyJonathan: Lock and process the breakpoint request for one file.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -588,33 +598,53 @@ export class ApexDebug extends LoggingDebugSession {
...
+              ','
+            )} unverified=${unverifiedBreakpoints.join(',')}`
+          );
         }
-      });
-      response.body = { breakpoints: processedBreakpoints };
-    }
+        return Promise.resolve(response);
+      },
+      { timeout: this.breakpointLockTimeout }
```
DatGuyJonathan: This is either the default or configured in workspace settings.

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -58,6 +58,13 @@
           "minimum": 500,
           "maximum": 120000,
           "description": "%connection_timeout_ms_description%"
+        },
+        "salesforcedx-vscode-apex-debugger.setBreakpointTimeoutMs": {
+          "type": "number",
+          "default": 10000,
+          "minimum": 500,
+          "maximum": 120000,
+          "description": "%set_breakpoint_timeout_ms_description%"
```
DatGuyJonathan: Set-breakpoint timeout setting.

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -13,14 +13,21 @@
...
+    "vscode": "^1.17.0"
   },
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-apex-debugger": "40.13.0"
   },
   "devDependencies": {
-    "vscode": "1.1.5"
+    "vscode": "1.1.5",
+    "@salesforce/salesforcedx-utils-vscode": "40.13.0",
```
DatGuyJonathan: Dependency on testrunner

vazexqi:
Since this uses the async-lock package, I am less familiar with it semantics (unlike locks in Java) so I'm going to ask some high-level questions:

1. Why did you choose this particular package? Any advantages over any other?
1. How many _domains_ (the term from async-lock) are you using? I'm asking because if we are not careful, and we have made a circular dependency, we can get into a deadlock if no one is releasing the locks.
1. What is the max pending events that we can wait on? I'm asking because usually in Java when we acquire a lock, there is a nice queue of some size for threads trying to acquire the lock to wait on automatically. What is the size we are using here? And, more importantly, it is big enough for all intents and purposes or will we have more debugger BP requests?
1. What happens if you cannot acquire the lock within the timeout? Do we error out for that particular BP request or we error out for everything else that is currently in the queue? Trying to see the impact of failing to acquire the lock.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -460,6 +467,9 @@ export class ApexDebug extends LoggingDebugSession {
   private trace: string[] | undefined;
   private traceAll = false;
 
+  private lock = new AsyncLock({ timeout: DEFAULT_LOCK_TIMEOUT_MS });
```
vazexqi: This seems to imply that we have a lock for each MapReferenceContainer? Is the MapReferenceContainer a singleton? It doesn't have the word service or singleton in it so I'm not sure. 

I'm asking because I want to know how many locks are we creating.

If we don't have a shared lock, then we cannot actually _serialize_ the operations and we might still end up with a race condition.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
vazexqi: What happens if we have an error, does it automatically release the lock? Or do we need to have to do something similar to http://www.javapractices.com/topic/TopicAction.do?Id=253

packages/salesforcedx-vscode-apex-debugger/package.json
vazexqi: Hopefully people won't need to set this one. I'm actually less inclined with exposing this for now since this actually is a technical detail that we have. If we expose this, then we need to support this even if we come up with a better, faster scheme in the future for setting breakpoints.

Could we hide this setting? Or do you want to have this setting exposed?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
vazexqi: Also, just to be certain, all the requests for Create, Update, Delete, And Read (read is important as well or else we can have a race condition if on the data) go through this entry point?

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -6,15 +6,17 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.16.0"
+    "vscode": "^1.17.0"
```
vazexqi: 👍

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -75,13 +77,18 @@ const TRACE_ALL = 'all';
 const TRACE_CATEGORY_VARIABLES = 'variables';
 const TRACE_CATEGORY_LAUNCH = 'launch';
 const TRACE_CATEGORY_PROTOCOL = 'protocol';
+const TRACE_CATEGORY_BREAKPOINTS = 'breakpoints';
 
-export type TraceCategory = 'all' | 'variables' | 'launch' | 'protocol';
+export type TraceCategory =
```
vazexqi: 👍

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
vazexqi: What is the difference between breakpointLockTimeout and DEFAULT_LOCK_TIMEOUT_MS

packages/salesforcedx-apex-debugger/test/unit/adapter/apexDebugVariablesHandling.test.ts
```diff
@@ -621,6 +806,91 @@ describe('Debugger adapter variable handling - unit', () => {
...
+        1001
+      );
+
+      const localScope = new ScopeContainer('global', frameInfo);
+      const vars = await localScope.expand(adapter, 'all', 0, 0);
+      expect(vars.length).to.equal(1);
+      expect(
+        ApexVariable.compareVariables(expectedVariableObj, vars[0])
+      ).to.equal(0);
+    });
```
guw: @DatGuyJonathan Thanks for adding the tests! 👍

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: The lock will be released when there's an error in the locked request.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: I'll stick with default for now.

packages/salesforcedx-vscode-apex-debugger/package.json
DatGuyJonathan: That makes sense to me. I don't think there's a config property to hide it, so I'll just remove it for now. At least we know how to make this setting configurable.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: This lock belongs to the ApexDebug object. VS Code ensures there can only be one 'apex' debug type active at a time, so there will only be one lock. All breakpoint operations go through setBreakpointsRequest where the lock is used.

PR #160, opened 2017-10-06
Ensure that apiVersion of the cli and REST calls are the same

GaryFNC:
Fix the SObjectDescribe REST calls to use the same apiVersion as the
current cli.  Done by calling sfdx force --json and extracting the
apiVersion output and using that in the describe REST calls

@W-4369079@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -339,7 +339,14 @@ export class SObjectDescribe {
       const response: XHRResponse = await xhr(options);
       const batchResponse = JSON.parse(response.responseText) as BatchResponse;
       const fetchedObjects: SObject[] = [];
+      let i = nextToProcess;
       for (const sr of batchResponse.results) {
+        if (sr.result instanceof Array) {
+          if (sr.result[0].errorCode && sr.result[0].message) {
+            console.log(`Error: ${sr.result[0].message} - ${types[i]}`);
+          }
+        }
+        i++;
         fetchedObjects.push(sr.result);
       }
       return Promise.resolve(fetchedObjects);
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -173,14 +173,14 @@ export class SObjectDescribe {
   private accessToken: string;
   private instanceUrl: string;
   private readonly servicesPath: string = 'services/data';
-  // the targetVersion MUST be consistent with the cli that is being used
-  // at least until describeGlobal is converted to REST calls and even then it is safer
-  private readonly targetVersion = '40.0';
-  private readonly versionPrefix = 'v' + this.targetVersion;
-  private readonly sobjectsPart: string = this.versionPrefix + '/sobjects';
-  private readonly batchPart: string = this.versionPrefix + '/composite/batch';
+  // the targetVersion should be consistent with the Cli even if only using REST calls
+  private targetVersion = '';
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -198,6 +198,25 @@ export class SObjectDescribe {
       this.accessToken = orgInfo.accessToken;
       this.instanceUrl = orgInfo.instanceUrl;
     }
+    if (!this.targetVersion) {
+      this.targetVersion = await this.getTargetApiVersion(projectPath);
+    }
+  }
+
+  private async getTargetApiVersion(projectPath: string): Promise<string> {
```
vazexqi: Just wanted to let you know that this will add about 1.5-2 seconds to run time. This probably won't matter since the describe call dominates but I want you to be aware of this and if it is a worthwhile tradeoff.

vazexqi:
After talking more to @GaryFNC about the real problem, this fix is nice to have but won't necessarily fix the general problem. I think we need some way to actually detect if an individual describe fails and not output undefined.cls.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: yes, it does add it once to the total.  And it will be replaced as soon as I can get the right code from the BUT-T work that Thomas, etc. are doing.   He knows about the need and is intending to do that, not there yet.
I do think it's quite important that we are consistent with the cli view of the world.

GaryFNC:
Not going to merge this until I add better error reporting for the situation where there is an error fetching the detailed describe info.   That should "never" happen if the apiVersions match, but "never" is a strong word and far better to catch it if it occurs.

chowwinston:
@GaryFNC @vazexqi I just pushed a commit that should fix the root cause of the undefined faux class issue. For some reason the sobject list call will return some sobjects that are undefined. This is outside of our control so the best thing to do is just not attempt generate when name = undefined.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -180,13 +180,14 @@ export class FauxClassGenerator {
...
       fs.mkdirSync(folderPath);
     }
     const fauxClassPath = path.join(folderPath, sobject.name + '.cls');
-
-    fs.writeFileSync(fauxClassPath, this.generateFauxClassText(sobject));
+    if (sobject.name) {
+      fs.writeFileSync(fauxClassPath, this.generateFauxClassText(sobject));
+    }
 
     return fauxClassPath;
```
vazexqi: Why is this returning the "undefined.cls" then? What is the reasoning for that? Is that a potential bug?

chowwinston: I just deleted my org so I can't get the exact error anymore, but somehow some of the sobjects from sobject list return an error with "object not found" or something similar.

chowwinston: The generation assumes it's a valid sobject, so only line 187, it will just give the path to be ../undefined.cls when sobject.name is undefined.

chowwinston: Actually I'll move this check to a better place so it won't even come into this method if it's an invalid sobject

vazexqi:
@GaryFNC - What is blocking this from going in?

GaryFNC:
@vazexqi It was blocked on the Travis timeout problem.  Jonathan changed it so that Travis was no longer a blocker for merge.

PR #162, opened 2017-10-09
Feature parity with the HTML language server for Visualforce

vazexqi:
### What does this PR do?

We have started taking the VS Code HTML language server and shaping it as the basis for our Visualforce one. 

This PR ensures that we have feature parity with the HTML one in terms of

* formatting
* code-completion (this PR also removes code completion for Ionic and Angular tags, since those are not supported by default on our platform)

There is one feature that we don't have feature parity yet - the color indicator when you have an inline style in CSS. This is because it is using a new proposed API from VS Code that is only available to their built-in extensions and not available publicly yet. See https://github.com/Microsoft/vscode/issues/34366

feature-parity-with-html

@W-4344488@


packages/salesforcedx-vscode-visualforce-next/package.nls.json
```diff
@@ -0,0 +1,36 @@
+{
```
vazexqi: @ruthemmanuelle - These are the same text that is used for the built-in HTML language server from Microsoft. We had shared this text before. As we move forward, I'm specializing our Visualforce language server and we will have more control over the text that we use. You don't have to make the change here, you can change it in a later PR but this is just FYI.

Here's where the text appears:

screen shot 2017-10-09 at 11 46 15 am

PR #164, opened 2017-10-10
Update to use signed jar

chowwinston:
### What does this PR do?
Update jorje jar to include default sobject methods and properly set primary key and category of SobjectFieldInfo


chowwinston:
I wasn't able to sign with the mvn command so I used jarsigner directly, although this gave a warning. 

jarsigner -keystore sfdc.jks -storepass ***** -keypass ***** apex-jorje-lsp/target/apex-jorje-lsp-210.0-SNAPSHOT.jar sfdc

`This jar contains signatures that does not include a timestamp. Without a timestamp, users may not be able to validate this jar after the signer certificate's expiration date (2018-08-15) or after any future revocation date.`

PR #167, opened 2017-10-11
Notify the user of Refresh command status

GaryFNC:
Convert the ForceGenerateFauxClassesExecutor (command) and FauxClassGenerator to properly utilize the support in SfdxCommandLetExecutor to hook into notification, cancellation, and status bar.  This involves creating a LocalCommandExecution to feed the observables from an emitter (exit, error, stdout, stderr). 
Cancellation is checked before each batch of SObject describes is done - as these are the long operations in FauxClassGenerator.
Added fauxGenerateTest for integration tests of the generate() call.
More cleanup of the testing in general to ensure that scratch orgs are deleted, etc.

@W-4358436@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -30,7 +30,7 @@
     "nyc": "^11.0.2",
     "sinon": "^2.3.6",
     "typescript": "2.4.0",
-    "vscode": "1.1.4"
+    "vscode": "1.1.5"
   },
   "scripts": {
     "vscode:package": "npm prune --production",
--- a/packages/salesforcedx-sobjects-faux-generator/package.json
+++ b/packages/salesforcedx-sobjects-faux-generator/package.json
@@ -29,8 +29,7 @@
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
     "sinon": "^2.3.6",
-    "typescript": "2.4.0",
-    "vscode": "1.1.5"
+    "typescript": "2.4.0"
   },
   "scripts": {
     "vscode:package": "npm prune --production",
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -27,6 +27,7 @@ export interface CancellationToken {
 
 export const SUCCESS_CODE = '0';
 export const FAILURE_CODE = '1';
+const SFDX_PROJECT_FILE = 'sfdx-project.json';
 export class FauxClassGenerator {
   private emitter: EventEmitter;
   private cancellationToken: CancellationToken | undefined;
@@ -97,6 +98,14 @@ export class FauxClassGenerator {
       this.SOBJECTS_DIR
     );
 
+    if (
+      !fs.existsSync(projectPath) ||
+      !fs.existsSync(path.join(projectPath, SFDX_PROJECT_FILE))
+    ) {
+      return this.errorExit(
+        nls.localize('no_generate_if_not_in_project', sobjectsFolderPath)
+      );
+    }
     this.cleanupSObjectFolders(sobjectsFolderPath);
 
     const describe = new SObjectDescribe();
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -4,14 +4,13 @@
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 */
+import { SFDX_PROJECT_FILE } from '@salesforce/salesforcedx-utils-vscode/out/src';
 import { LocalCommandExecution } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
-
 import { EventEmitter } from 'events';
 import * as fs from 'fs';
 import { EOL } from 'os';
 import * as path from 'path';
 import { mkdir, rm } from 'shelljs';
-
 import {
   ChildRelationship,
   Field,
@@ -25,18 +24,15 @@ export interface CancellationToken {
   isCancellationRequested: boolean;
 }
 
-export const SUCCESS_CODE = '0';
-export const FAILURE_CODE = '1';
-const SFDX_PROJECT_FILE = 'sfdx-project.json';
+const SFDX_DIR = '.sfdx';
+const TOOLS_DIR = 'tools';
+const SOBJECTS_DIR = 'sobjects';
+const STANDARDOBJECTS_DIR = 'standardObjects';
+const CUSTOMOBJECTS_DIR = 'customObjects';
+
 export class FauxClassGenerator {
   private emitter: EventEmitter;
   private cancellationToken: CancellationToken | undefined;
-
-  private SFDX_DIR = '.sfdx';
-  private TOOLS_DIR = 'tools';
-  private SOBJECTS_DIR = 'sobjects';
-  private STANDARDOBJECTS_DIR = 'standardObjects';
-  private CUSTOMOBJECTS_DIR = 'customObjects';
 
   // the empty string is used to represent the need for a special case
   // usually multiple fields with specialized names
@@ -77,14 +73,26 @@ export class FauxClassGenerator {
   }
 
   private errorExit(errorMessage: string): Promise<string> {
-    this.emitter.emit(LocalCommandExecution.STDERR_EVENT, errorMessage);
-    this.emitter.emit(LocalCommandExecution.ERROR_EVENT, '1');
-    return Promise.reject(`${FAILURE_CODE} - ${errorMessage}`);
+    this.emitter.emit(LocalCommandExecution.STDERR_EVENT, `${errorMessage}\n`);
+    this.emitter.emit(
+      LocalCommandExecution.ERROR_EVENT,
+      new Error(errorMessage)
+    );
+    this.emitter.emit(
+      LocalCommandExecution.EXIT_EVENT,
+      LocalCommandExecution.FAILURE_CODE
+    );
+    return Promise.reject(
+      `${LocalCommandExecution.FAILURE_CODE.toString()} - ${errorMessage}`
+    );
   }
 
   private successExit(): Promise<string> {
-    this.emitter.emit(LocalCommandExecution.EXIT_EVENT, '0');
-    return Promise.resolve(SUCCESS_CODE);
+    this.emitter.emit(
+      LocalCommandExecution.EXIT_EVENT,
+      LocalCommandExecution.SUCCESS_CODE
+    );
+    return Promise.resolve(LocalCommandExecution.SUCCESS_CODE.toString());
   }
 
   public async generate(
@@ -93,9 +101,17 @@ export class FauxClassGenerator {
   ): Promise<string> {
     const sobjectsFolderPath = path.join(
       projectPath,
-      this.SFDX_DIR,
-      this.TOOLS_DIR,
-      this.SOBJECTS_DIR
+      SFDX_DIR,
+      TOOLS_DIR,
+      SOBJECTS_DIR
+    );
+    const standardSObjectsFolderPath = path.join(
+      sobjectsFolderPath,
+      STANDARDOBJECTS_DIR
+    );
+    const customSObjectsFolderPath = path.join(
+      sobjectsFolderPath,
+      CUSTOMOBJECTS_DIR
     );
 
     if (
@@ -151,19 +167,16 @@ export class FauxClassGenerator {
 
     this.logFetchedObjects(standardSObjects, customSObjects);
 
-    const standardResult = this.generateFauxClasses(
-      standardSObjects,
-      path.join(sobjectsFolderPath, this.STANDARDOBJECTS_DIR)
-    );
-    if (standardResult) {
-      return this.errorExit(standardResult);
+    try {
+      this.generateFauxClasses(standardSObjects, standardSObjectsFolderPath);
+    } catch (exc) {
+      return this.errorExit(exc);
     }
-    const customResult = this.generateFauxClasses(
-      customSObjects,
-      path.join(sobjectsFolderPath, this.CUSTOMOBJECTS_DIR)
-    );
-    if (customResult) {
-      return this.errorExit(customResult);
+
+    try {
+      this.generateFauxClasses(customSObjects, customSObjectsFolderPath);
+    } catch (exc) {
+      return this.errorExit(exc);
     }
 
     return this.successExit();
@@ -223,23 +236,19 @@ export class FauxClassGenerator {
     return decls;
   }
 
-  private static fieldName(decl: string) {
+  private static fieldName(decl: string): string {
     return decl.substr(decl.indexOf(' ') + 1);
   }
 
-  private generateFauxClasses(
-    sobjects: SObject[],
-    targetFolder: string
-  ): string {
+  private generateFauxClasses(sobjects: SObject[], targetFolder: string): void {
     if (!this.createIfNeededOutputFolder(targetFolder)) {
-      return nls.localize('no_sobject_output_folder_text', targetFolder);
+      throw nls.localize('no_sobject_output_folder_text', targetFolder);
     }
     for (const sobject of sobjects) {
       if (sobject.name) {
         this.generateFauxClass(targetFolder, sobject);
       }
     }
-    return '';
   }
 
   // VisibleForTesting
@@ -341,29 +350,20 @@ export class FauxClassGenerator {
     }
   }
 
+  private logSObjects(sobjectKind: string, fetchedLength: number) {
+    if (fetchedLength > 0) {
+      this.emitter.emit(
+        LocalCommandExecution.STDOUT_EVENT,
+        nls.localize('fetched_sobjects_length_text', fetchedLength, sobjectKind)
+      );
+    }
+  }
+
   private logFetchedObjects(
     standardSObjects: SObject[],
     customSObjects: SObject[]
   ) {
-    if (standardSObjects.length > 0) {
-      this.emitter.emit(
-        LocalCommandExecution.STDOUT_EVENT,
-        nls.localize(
-          'fetched_sobjects_length_text',
-          standardSObjects.length,
-          'Standard'
-        )
-      );
-    }
-    if (customSObjects.length > 0) {
-      this.emitter.emit(
-        LocalCommandExecution.STDOUT_EVENT,
-        nls.localize(
-          'fetched_sobjects_length_text',
-          customSObjects.length,
-          'Custom'
-        )
-      );
-    }
+    this.logSObjects('Standard', standardSObjects.length);
+    this.logSObjects('Custom', customSObjects.length);
   }
 }
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -19,6 +19,7 @@ import {
   SObjectCategory,
   SObjectDescribe
 } from '../describe';
+import { nls } from '../messages';
 
 export interface CancellationToken {
   isCancellationRequested: boolean;
@@ -71,6 +72,17 @@ export class FauxClassGenerator {
     this.cancellationToken = cancellationToken;
   }
 
+  private errorExit(errorMessage: string): Promise<string> {
+    this.emitter.emit(LocalCommandExecution.STDERR_EVENT, errorMessage);
+    this.emitter.emit(LocalCommandExecution.ERROR_EVENT, '1');
+    return Promise.reject(errorMessage);
+  }
+
+  private successExit(successMessage: string): Promise<string> {
+    this.emitter.emit(LocalCommandExecution.EXIT_EVENT, '0');
+    return Promise.resolve(successMessage);
+  }
+
   public async generate(
     projectPath: string,
     type: SObjectCategory
@@ -81,9 +93,6 @@ export class FauxClassGenerator {
       this.TOOLS_DIR,
       this.SOBJECTS_DIR
     );
-    if (!projectPath) {
-      return Promise.reject('foobar');
-    }
 
     this.cleanupSObjectFolders(sobjectsFolderPath);
 
@@ -95,7 +104,9 @@ export class FauxClassGenerator {
     try {
       sobjects = await describe.describeGlobal(projectPath, type);
     } catch (e) {
-      return Promise.reject('Failure fetching list of SObjects from org ' + e);
+      return this.errorExit(
+        nls.localize('failure_fetching_sobjects_list_text', e)
+      );
     }
     let j = 0;
     while (j < sobjects.length) {
@@ -104,14 +115,16 @@ export class FauxClassGenerator {
           this.cancellationToken &&
           this.cancellationToken.isCancellationRequested
         ) {
-          return Promise.reject('faux class generation cancelled');
+          return this.errorExit(nls.localize('faux_generation_cancelled_text'));
         }
         fetchedSObjects = fetchedSObjects.concat(
           await describe.describeSObjectBatch(projectPath, sobjects, j)
         );
         j = fetchedSObjects.length;
       } catch (e) {
-        return Promise.reject('Failure performing SObject describe ' + e);
+        return this.errorExit(
+          nls.localize('failure_in_sobject_describe_text', e)
+        );
       }
     }
 
@@ -131,17 +144,17 @@ export class FauxClassGenerator {
       path.join(sobjectsFolderPath, this.STANDARDOBJECTS_DIR)
     );
     if (standardResult) {
-      return Promise.reject(standardResult);
+      return this.errorExit(standardResult);
     }
     const customResult = this.generateFauxClasses(
       customSObjects,
       path.join(sobjectsFolderPath, this.CUSTOMOBJECTS_DIR)
     );
     if (customResult) {
-      return Promise.reject(customResult);
+      return this.errorExit(customResult);
     }
 
-    return Promise.resolve('');
+    return this.successExit('');
   }
 
   private stripId(name: string): string {
@@ -207,7 +220,7 @@ export class FauxClassGenerator {
     targetFolder: string
   ): string {
     if (!this.createIfNeededOutputFolder(targetFolder)) {
-      return 'no sobjects output folder ' + targetFolder;
+      return nls.localize('no_sobject_output_folder_text', targetFolder);
     }
     for (const sobject of sobjects) {
       if (sobject.name) {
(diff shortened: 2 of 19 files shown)


packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -30,7 +30,7 @@
     "nyc": "^11.0.2",
     "sinon": "^2.3.6",
     "typescript": "2.4.0",
-    "vscode": "1.1.4"
+    "vscode": "1.1.5"
```
vazexqi: Any reason for this bump? I do not oppose the bump but why are you bumping it only for yours?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -4,10 +4,12 @@
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 */
-import * as fs from 'fs';
-import * as path from 'path';
+import { LocalCommandExecution } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
```
vazexqi: Do *not* have a new line here. All imports need to be back-to-back for the TSLint ordered-import rules to work (https://palantir.github.io/tslint/rules/ordered-imports/).

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -4,10 +4,12 @@
...
 
+import { EventEmitter } from 'events';
+import * as fs from 'fs';
 import { EOL } from 'os';
+import * as path from 'path';
 import { mkdir, rm } from 'shelljs';
```
vazexqi: Ditto.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -17,8 +19,19 @@ import {
...
 } from '../describe';
+import { nls } from '../messages';
+
+export interface CancellationToken {
+  isCancellationRequested: boolean;
+}
 
+export const SUCCESS_CODE = '0';
+export const FAILURE_CODE = '1';
+const SFDX_PROJECT_FILE = 'sfdx-project.json';
```
vazexqi: Is there no where else where we define this?  This seems like a useful constant that can go into some shared library.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -17,8 +19,19 @@ import {
...
 export class FauxClassGenerator {
+  private emitter: EventEmitter;
+  private cancellationToken: CancellationToken | undefined;
+
   private SFDX_DIR = '.sfdx';
```
vazexqi: Why can't these be constants as well? I think I asked this during the previous code review.

packages/salesforcedx-utils-vscode/src/cli/localCommandExecutor.ts
```diff
@@ -0,0 +1,41 @@
...
+export class LocalCommandExecution implements CommandExecution {
+  public readonly command: Command;
+  public readonly cancellationToken?: CancellationToken;
+  public readonly processExitSubject: Observable<number | undefined>;
+  public readonly processErrorSubject: Observable<Error | undefined>;
+  public readonly stdoutSubject: Observable<Buffer | string>;
+  public readonly stderrSubject: Observable<Buffer | string>;
+
+  public cmdEmitter: EventEmitter = new EventEmitter();
+  public static readonly EXIT_EVENT = 'exitEvent';
```
vazexqi: Why not make these constants in the module/file instead of in the `LocalCommandExecution`? This is not Java. It's possible to have constants outside the file.

packages/salesforcedx-utils-vscode/src/cli/localCommandExecutor.ts
```diff
@@ -0,0 +1,41 @@
+import { EventEmitter } from 'events';
+import 'rxjs/add/observable/fromEvent';
+import { Command } from '.';
+import { CancellationToken, CommandExecution } from './commandExecutor';
+
```
vazexqi: Remove the new line per the reasons mentioned.

packages/salesforcedx-utils-vscode/src/constants.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export const SFDX_PROJECT_FILE = 'sfdx-project.json';
```
vazexqi: Oh so we do have a constant. We need to use it in fauxClassGenerator.ts

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -316,14 +317,11 @@ export interface CommandletExecutor<T> {
...
   implements CommandletExecutor<T> {
-  public execute(response: ContinueResponse<T>): void {
-    const cancellationTokenSource = new vscode.CancellationTokenSource();
-    const cancellationToken = cancellationTokenSource.token;
-
-    const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: vscode.workspace.rootPath
-    }).execute(cancellationToken);
-
+  protected attachExecution(
```
vazexqi: I think this needs a newline before here. I'll be honest and say that I don't like to leave such comments in code reviews. I'm open to spending time using a better code formatter. But as far as I know, many code formatters cannot insert new lines between fields, methods.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -25,16 +26,29 @@ class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}> {
...
 
   public async execute(response: ContinueResponse<{}>): Promise<void> {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const execution = new LocalCommandExecution(this.build(response.data));
+
+    this.attachExecution(execution, cancellationTokenSource, cancellationToken);
+
     const projectPath: string = <string>vscode.workspace.rootPath;
```
vazexqi: Use `as string` instead of <string>. http://acdcjunior.github.io/typescript-cast-object-to-other-type-or-instanceof.html. I mentioned this several times already but I'm not sure if you were one of them I mentioned it to. Again, if this is something that if we can enforce at code formatter level, that would be better.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -100,5 +100,6 @@ export const messages = {
   force_apex_execute_document_text:
     'SFDX: Execute Anonymous Apex with Editor Contents',
   force_apex_execute_selection_text:
-    'SFDX: Execute Anonymous Apex with Currently Selected Text'
+    'SFDX: Execute Anonymous Apex with Currently Selected Text',
+  force_refresh_sobjects: 'SFDX: Refresh SObject Definitions'
```
vazexqi: `force_sobjects_refresh`? Since that seems to follow more with how the CLI does it and how these other messages do it. `force_<category>_<action>`

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Conventions:
+ * _message: is for unformatted text that will be shown as-is to
+ * the user.
+ * _text: is for text that will appear in the UI, possibly with
+ * decorations, e.g., $(x) uses the https://octicons.github.com/ and should not
+ * be localized
+ *
+ * If ommitted, we will assume _message.
+ */
+export const messages = {
```
vazexqi: @ruthemmanuelle - Could you take a look at these?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -169,14 +227,17 @@ export class FauxClassGenerator {
...
+  ): string {
     if (!this.createIfNeededOutputFolder(targetFolder)) {
-      console.log('no sobjects output folder ' + targetFolder);
-      return;
+      return nls.localize('no_sobject_output_folder_text', targetFolder);
     }
     for (const sobject of sobjects) {
       this.generateFauxClass(targetFolder, sobject);
     }
+    return '';
```
vazexqi: Why are you returning an empty string to signify success? This doesn't seem like a good pattern.

Either move the check earlier `if (!this.createIfNeededOutputFolder(targetFolder)) {` or think of something else.

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -0,0 +1,200 @@
...
+// and multiple objects for testing describeGlobal
+const PROJECT_NAME = `project_${new Date().getTime()}`;
+const CUSTOM_OBJECT_NAME = 'MyCustomObject__c';
+const CUSTOM_OBJECT2_NAME = 'MyCustomObject2__c';
+const CUSTOM_OBJECT3_NAME = 'MyCustomObject3__c';
+const CUSTOM_FIELD_FULLNAME = '.MyCustomField__c';
+const SIMPLE_OBJECT_SOURCE_FOLDER = 'simpleObjectAndField';
+
+// tslint:disable:no-unused-expression
+describe('Generate faux classes for SObjects', function() {
```
vazexqi: Why function()? I thought we decided to use the `() = > {` syntax per other tests files?

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  let cancellationTokenSource: util.CancellationTokenSource;
+  let projectPath: string;
+  let emitter: EventEmitter;
+
+  function getGenerator(): FauxClassGenerator {
+    return new FauxClassGenerator(emitter, cancellationTokenSource.token);
+  }
+
+  before(async function() {
```
vazexqi: Same here as well. Why not `async () => {`

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      PROJECT_NAME,
+      SIMPLE_OBJECT_SOURCE_FOLDER,
+      customFields
+    );
+
+    projectPath = path.join(process.cwd(), PROJECT_NAME);
+    emitter = new EventEmitter();
+  });
+
+  after(async function() {
```
vazexqi: Here as well. And I see that you are using it probably as `async() => {` below.

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+    expect(result).to.equal(SUCCESS_CODE);
+    expect(stdoutInfo).to.contain(
+      nls.localize('fetched_sobjects_length_text', 3, 'Custom')
+    );
+  });
+});
+
+// easy way to force the generator to throw an error
+const DUMMY_PROJECT_FILE = `zzz${SFDX_PROJECT_FILE}`;
```
vazexqi: Don't do `zzz`. Either use fake, bogus, or some other adjective. It's very annoying to have to read a zzz, xyz, and not understand why someone named it that way. It's the same with CUSTOM_OBJECT_2, CUSTOM_OBJECT_3,

packages/salesforcedx-sobjects-faux-generator/test/integration/integrationTestUtil.ts
```diff
@@ -50,6 +52,21 @@ export async function createScratchOrg(projectName: string): Promise<string> {
   return Promise.resolve(username);
 }
 
+export async function deleteScratchOrg(userName: string): Promise<void> {
```
vazexqi: @chowwinston - Do we need this function? One of the new features of the CLI is the ability to create a scratch org with a "purpose". Could we change the scratch org creation to do a CLI creation purpose of CI, which would make it auto-delete in 3 days (I don't know the exact days).

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -58,6 +71,22 @@ export class FauxClassGenerator {
...
   ]);
 
+  constructor(emitter: EventEmitter, cancellationToken?: CancellationToken) {
+    this.emitter = emitter;
+    this.cancellationToken = cancellationToken;
+  }
+
+  private errorExit(errorMessage: string): Promise<string> {
+    this.emitter.emit(LocalCommandExecution.STDERR_EVENT, errorMessage);
+    this.emitter.emit(LocalCommandExecution.ERROR_EVENT, '1');
```
vazexqi: Do you properly conform to the events from Node.js? The error event is supposed to emit an `Error` object and not a string, like what you have done. See https://nodejs.org/api/child_process.html#child_process_event_error

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -58,6 +71,22 @@ export class FauxClassGenerator {
...
+    return Promise.reject(`${FAILURE_CODE} - ${errorMessage}`);
+  }
+
+  private successExit(): Promise<string> {
+    this.emitter.emit(LocalCommandExecution.EXIT_EVENT, '0');
```
vazexqi: The exit even is also supposed to emit something similar to https://nodejs.org/api/child_process.html#child_process_event_exit

Could you check that you are emitting the right signatures if you want to use this technique?

packages/salesforcedx-sobjects-faux-generator/package.json
GaryFNC: I should take it back out - it was when I was trying to use vscode - probably shouldn't depend on vscode at all for this non-vsix package.     I was just trying to make it consistent with the other ones -but best to remove it - I'll verify

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: will do - just slipped through

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: Yes, I moved it to a constants.ts in utils-vscode.  Missed updating this one - thanks

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: could be and should be - I'll fix

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
GaryFNC: will do - I wasn't sure which was the preferred form

packages/salesforcedx-vscode-core/src/messages/i18n.ts
GaryFNC: good suggestion - slightly more consistent

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: agreed - not a good pattern - I was trying to avoid duplicating code, but I see a better way to do it.

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
GaryFNC: just a mistake in copy/paste from a different test... oops

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
GaryFNC: As I noted before, CUSTOM_OBJECT_2, etc are because they are all the same and just there to deal with having multiple custom objects...    But I'll change zzz - I can understand that concern.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: I'll go take a look - good observation.  At some level, the emitter events are arbitrarily defined by the emitter - but best to conform to standard patterns.

packages/salesforcedx-sobjects-faux-generator/test/integration/integrationTestUtil.ts
GaryFNC: Depending on whether that can be controlled (asking platform-cli), even 3 days would be too long when testing code locally and repeatedly.  Pain in the butt to have to go to the devhub and manually remove these so frequently.

packages/salesforcedx-utils-vscode/src/cli/localCommandExecutor.ts
GaryFNC: wasn't sure what the preferred pattern was for Typescript.   And to some extent, I'd prefer that these be used clearly as tied to LocalCommandExecution rather than being an arbitrary constant.  At the point of use in tests, etc. it makes more sense to me to prefix them with LocalCommandExecution.   IMHO

packages/salesforcedx-sobjects-faux-generator/test/integration/integrationTestUtil.ts
vazexqi: Where are you running your tests? If you are doing this locally, you can bump up your allocation of scratch orgs. Our CI org has a high scratch org limit as well.

That being said, your point is understood. Let's just schedule it for cleaning to be good citizens.

packages/salesforcedx-utils-vscode/src/cli/localCommandExecutor.ts
vazexqi: @GaryFNC - Sounds reasonable. Let's keep it this way for this one. It does seem better to localize it within the class.

GaryFNC:
Modified LocalCommandExecution to issue an Error object for error event and a number for exit event.

@vazexqi Note minor change in ChannelService - since the processErrorSubject was taking in an Error object, it was outputting "ended with error Error: xxxxx".   Error.toString appends the Error: - where error.message just gives the output message

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -99,16 +167,19 @@ export class FauxClassGenerator {
...
-      standardSObjects,
-      path.join(sobjectsFolderPath, this.STANDARDOBJECTS_DIR)
-    );
-    this.generateFauxClasses(
-      customSObjects,
-      path.join(sobjectsFolderPath, this.CUSTOMOBJECTS_DIR)
-    );
+    try {
+      this.generateFauxClasses(standardSObjects, standardSObjectsFolderPath);
+    } catch (exc) {
```
vazexqi: I just found out that it's probably more idiomatic to use `e` rather than `exc`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -99,16 +167,19 @@ export class FauxClassGenerator {
...
+      return this.errorExit(exc);
+    }
 
-    return Promise.resolve('');
+    try {
+      this.generateFauxClasses(customSObjects, customSObjectsFolderPath);
+    } catch (exc) {
```
vazexqi: Change this too.

vazexqi:
1. What happens if I were to cancel while it is fetching the sobjects halfway? Does it properly cancel?
1. What happens when I invoke the command multiple times, waiting for it to finish each time? Does it probably delete the sobjects directory? Or does it just add new things to it (potentially not being able to remove old sobjects that I have deleted)?
1. What happens if I invoke the command in quick succession without waiting for it to finish? This is an edge case, but I'd like to know what happens.

vazexqi:
screen shot 2017-10-12 at 3 29 12 pm

Do we really need all that logging?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
GaryFNC: will do

GaryFNC:
logging was very useful at times to see what was generated and capture that.   I don't think it has any impact on the speed, but it does add a lot to the log - are you concerned with clutter or ?

Cancel does work properly in the sense that it stops the operation.   However, it will leave an empty sobjects folder (didnt seem worth trying to create a new folder, delete the old one, and rename - likely to send even a lot more events to the Apex LS for a rare situation).

The rapid fire is interesting..    It will delete each time, but it is working against the same scratch org.   So, the last generate of the classes will "win".   Even if the 2nd one is part way through generating, the later one will overwrite (need to confirm that).    The actual generation step is very fast - the batch describe fetch is the slow async step.   There is no delete that could come after the last generate set and lose files.

vazexqi:
The logging is too noisy. It floods the console. I think using console.debug is better since according to https://developer.mozilla.org/en-US/docs/Web/API/Console that only appaers when you select the verbose mode.

This also made me realize that we don't have a chosen logging library yet. The CLI is using bunyan. Maybe we should use that one.

vazexqi:
It will delete each time, but it is working against the same scratch org. So, the last generate of the classes will "win"

Does this work on Windows? I think sometimes Windows holds on the file resource for a while. If you are opening up multiple processes (the case of rapid fire) I am not sure if you will see a "resource is being used" error.

GaryFNC:
Need to verify on Windows.   We aren't really running multiple processes, just async inside VSCode.   But yes, Windows does do some odd things with Files... So, it's possible that it could fail to delete or block an overwrite or .....    

On the logging, I'll change to console.debug ASAP.

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  faux_generation_cancelled_text: 'Faux class generation cancelled',
```
ruthemmanuelle: Our style guidelines say to use only one 'l' in "canceled": https://confluence.internal.salesforce.com/display/public/DOCSTYLE/canceled%2C+canceling%2C+cancellation

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  failure_fetching_sobjects_list_text:
+    'Failure fetching list of SObjects from org %s',
```
ruthemmanuelle: Is there anything we can say about why the failure occurred, or about what customers can do to recover from it? Also, the 's' in "sObject" should be lowercase when the word is used in a sentence like this.

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  failure_in_sobject_describe_text: 'Failure performing SObject describe %s',
```
ruthemmanuelle: Is there anything we can say about why the failure occurred, or about what customers can do to recover from it?

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  no_sobject_output_folder_text: 'No output folder available %s',
```
ruthemmanuelle: Why is the folder unavailable? What should customers do next?

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  fetched_sobjects_length_text:
+    'Fetched %s %s SObjects from default scratch org\n',
```
ruthemmanuelle: The 's' in "sObject" should be lowercase when the word is used in a sentence like this.

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  no_generate_if_not_in_project:
+    'Unable to generate faux classes for SObjects when not in an SFDX project %s'
```
ruthemmanuelle: The 's' in "sObject" should be lowercase when the word is used in a sentence like this. Also, can we give the user more info on how to move forward? Or is `%s` already appending that?

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
GaryFNC: really just able to provide the error back - shouldn't occur and all they can really do is try again depending on the error.   That said, I don't know any easy way to reproduce this failure - just that we should catch it and respond to the user if it happens

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
GaryFNC: shouldnt happen - but could due to some form of permissions issue in the local file system.  I'll add a suggestion to create it.

packages/salesforcedx-sobjects-faux-generator/src/messages/i18n.ts
GaryFNC: I think this is about all that we can tell them.   It just means that somehow they aren't doing all this in an SFDX project - opened vscode in the wrong place.    Hopefully, we are checking this for most all operations in the CLI and VSCode, but I haven't confirmed that.

GaryFNC:
@vazexqi On the issue of issuing the Refresh command multiple times - it does work,but might be cleaner to just return an error when the user attempts it while the task is still running

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
vazexqi: @GaryFNC - So this has not been fixed yet. You are still using the old syntax.

PR #168, opened 2017-10-12
Lock stacktraceRequest. Fix default type of requestTypeFilter.

DatGuyJonathan:
### What does this PR do?
The adapter sends one stopped event to VS Code when we've hit a breakpoint or finished stepping. VS Code sends a stackTraceRequest as a result of that stopped event. Then, for some reason, their tree api refreshes the callstack view which again sends a stackTraceRequest. As a result, our adapter fires off two state requests. If the timing is very close, apex debugger engine could still be processing the first request and cannot evaluate the state again. So, we're locking the stackTraceRequest to process one at a time.

Also, make default type of requestTypeFilter from string to array.

@W-4380397, W-4380128@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -745,9 +745,16 @@ export class ApexDebug extends LoggingDebugSession {
 
     const requestId = this.requestThreads.get(args.threadId)!;
     try {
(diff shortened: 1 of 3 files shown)


DatGuyJonathan:
@guw There are two uses for the locks right now: breakpoint and stacktrace. The former uses the CLI and the latter uses a RequestService. I think we should leave things as-is for now. If we need other operations to be serial, like setting breakpoints with RequestService, then we can move the locking mechanism.

PR #173, opened 2017-10-13
Add debugger integration tests

DatGuyJonathan:
### What does this PR do?
Add apex debugger integration test that invokes SFDX CLI.

@W-4269831@


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -972,7 +974,7 @@ export class ApexDebug extends LoggingDebugSession {
     response: DebugProtocol.VariablesResponse,
     args: DebugProtocol.VariablesArguments
   ): Promise<void> {
-    response.success = true;
+    response.success = false;
```
guw: @DatGuyJonathan I've had it that way initially. However, in case of "errors" the variables request logs the error and returns an empty variables list. Thus, technically the response should be considered _successful_.

packages/salesforcedx-apex-debugger/test/integration/adapter.test.ts
```diff
@@ -5,32 +5,229 @@
...
+const LINE_BREAKPOINT_INFO: LineBreakpointInfo[] = [];
 
-    before(async () => {
-      dc = new DebugClient('node', './out/src/adapter/apexDebug.js', 'apex');
-      // Use dc.start(4711) to debug the adapter during
-      // tests (adapter needs to be launched in debug mode separately).
-      await dc.start();
+/**
+ * These integration tests assume the environment has authenticated to
+ * a Dev Hub and it is set as the default Dev Hub.
```
guw: If the tests require an authenticated DevHub it should be moved to system tests.

guw: On a side note, would it make sense to introduce a second kind of integration tests which deal with a "mocked" Apex server? But that might just be too much effort.

DatGuyJonathan: @guw We've discussed that these integration tests don't fit the description of system tests. So, I've added a config in launch.json and script in package.json to run debugger unit and integration tests separately for our devs. Is there a specific flow that you're looking for in the autobuild?

guw: Thanks @DatGuyJonathan! Were you able to sort out the coverage issue? Re: flow I think we should discuss whether we want to run them by default with "npm run tests" or create a separate goal for "tests-with-devhub" in all packages. They should run after unit tests.

DatGuyJonathan: @guw What do you have in mind from segmenting unit tests, integration tests with devhub, and UI tests in appveyor when currently we just run them all in appveyor.yml?

guw: I usually run the unit test locally before committing without a fully running server and authenticated DevHub available.

DatGuyJonathan: Okay, I believe you should be able to do that with the changes in debugger's launch.json and package.json. We'll address the same concern with sobject tests in W-4402142.

DatGuyJonathan: I experimented with making Appveyor run unit tests first, then integration tests, then system tests in branch jwidjaja/appveyor. It's not ready, but you check it out there. I'll merge this PR to develop for now.

PR #174, opened 2017-10-14
Set the first character of the faux class field names to lower case

GaryFNC:
Apex is case insensitive, but it is better to set the field names on
the faux classes to lower case first character (camelCase) to be
consistent with the general pattern in Apex classes

@W-4369226@


vazexqi:
I was mistaken on this request to lowercase the first character of the fields in SObjects. After some valuable feedback from the community, I realized that the naming convention should be based on what the developer dictates (if they decide to use UpperCase, we should leave it as-is). Also, our [documentation](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_account.htm) also uses the UpperCase convention.

So, I'm going to close this PR and leave the code as-is.

@GaryFNC - Apologies for making the request for this PR.

PR #176, opened 2017-10-16
Removes .project file

vazexqi:
### What does this PR do?

Removes the .project. This is an Eclipse-specific file that should not be part of this project.


guw:
@vazexqi I wanted to avoid the PR to save on CI cycle time. Can you merge it in right away?

PR #178, opened 2017-10-17
Allow osx build to fail

Sweetman:
### What does this PR do?
Allows OSX builds to fail without blocking a PR
Tests on mac builds have been flakey and don't usually catch errors that the linux build do not catch. Let's make them optional to pass so flakiness does not block a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ node_js:
 
 cache:
   directories:
-    - node_modules
     - $HOME/.npm
 
 notifications:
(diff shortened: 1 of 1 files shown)


vazexqi:
@Sweetman https://travis-ci.org/forcedotcom/salesforcedx-vscode/builds/288835728?utm_source=github_status&utm_medium=notification is that the intended interface? So, now i I don't even know if it has failed in my PR. See the checks above.

Is there a way to do the following:

1. During the PR, it will notify if the Mac build has failed, but it won't block merger.
1. On the front page where we have the badge, only use the Linux build?

guw:
FYI: https://github.com/travis-ci/travis-ci/issues/7304

PR #180, opened 2017-10-17
Provide completion for standard tags in Visualforce

vazexqi:
### What does this PR do?

This adds basic tag completion for Visualforce. The list of tags is from https://developer.salesforce.com/docs/atlas.en-us.pages.meta/pages/pages_compref_map.htm.

visualforce-completion

This achieves parity with what we have in the Setup UI on Salesforce and the Developer Console.

@W-4344488@


.codecov.yml
```diff
@@ -1,4 +1,5 @@
 coverage:
   ignore:
     - "packages/salesforcedx-visualforce-markup-language-server/src/beautify"  # This folder contains generated files that we have embedded
-    - "packages/salesforcedx-visualforce-markup-language-server/src/parser/htmlTags.ts"  # This file is a collection of all HTML tags
\ No newline at end of file
+    - "packages/salesforcedx-visualforce-markup-language-server/src/parser/htmlTags.ts"  # This file is a collection of all HTML tags
+    - "packages/salesforcedx-visualforce-markup-language-server/src/parser/visualforceTags.ts"  # This file is a collection of all Visualforce tags
```
vazexqi: Another file that is just full of tags, not executable code.

.vscode/launch.json
```diff
@@ -15,7 +15,7 @@
       "request": "attach",
       "protocol": "inspector",
       "name": "Attach to Process (Inspector Protocol)",
-      "port": 9229,
+      "port": 6004,
```
vazexqi: I'll change this. This was just because the VF language server runs on port 6004.

packages/salesforcedx-visualforce-markup-language-server/src/utils/strings.ts
```diff
@@ -5,50 +5,23 @@
...
     }
   }
 
   return true;
 }
 
-/**
- * Determines if haystack ends with needle.
- */
-export function endsWith(haystack: string, needle: string): boolean {
```
vazexqi: This method is not actually used.

packages/salesforcedx-visualforce-markup-language-server/src/utils/strings.ts
```diff
@@ -5,50 +5,23 @@
...
-    return haystack === needle;
-  } else {
-    return false;
-  }
-}
-
-/**
- * @returns the length of the common prefix of the two strings.
- */
-export function commonPrefixLength(a: string, b: string): number {
```
vazexqi: This method is not actually used.

packages/salesforcedx-visualforce-markup-language-server/test/completion.test.ts
```diff
@@ -564,4 +569,118 @@ describe('HTML Completion', () => {
...
+          { label: 'title' },
+          { label: 'wizard' },
+          { label: 'ion-checkbox', notAvailable: true },
+          { label: 'ng-model', notAvailable: true },
+          { label: 'aria-activedescendant', notAvailable: true }
+        ]
+      });
+    });
+
+    it('Should show true, false as attribute options automatically', () => {
```
vazexqi: For types that are boolean, we also provide completion of true|false.

vazexqi:
One caveat, that is apparent from the video, is that the tooltip contains `<p>...</p>` tags. HTML tags are not rendered in VS Code for its language server.

We could try to sanitize the output (the output comes directly from the component registry on the SFDC side) but we cannot change the output at the source since too many other consumers rely on it.

packages/salesforcedx-visualforce-markup-language-server/test/completion.test.ts
```diff
@@ -564,4 +569,118 @@ describe('HTML Completion', () => {
...
+    return list;
+  }
+
+  describe('Visualforce Completions', () => {
+    // This list is from https://developer.salesforce.com/docs/atlas.en-us.pages.meta/pages/pages_compref_map.htm
+    it('Should display basic tags', () => {
+      const expectedNamespaces = [
+        'analytics',
+        'apex',
+        'chatter',
```
vazexqi: One useful note: chatteranswer is not here. I think it's OK to remove it because it will be retired in Spring '18.

See https://help.salesforce.com/articleView?id=000240491&type=1. Useful to know this when doing QA.

vazexqi:
screen shot 2017-10-17 at 6 59 59 pm

I stripped the HTML tags so that it looks better. This is an indiscriminate removal so things like <ol> and <li> are also removed. But, generally, the text looks better now.

PR #181, opened 2017-10-18
Rename settings from 'Salesforce DX' to 'Salesforce'

vazexqi:
### What does this PR do?

Rename the configuration titles to use the more general "Salesforce" term instead of "Salesforce DX".

@W-4401094@

### DIFF
--- a/packages/salesforcedx-vscode-visualforce/package.nls.json
+++ b/packages/salesforcedx-vscode-visualforce/package.nls.json
@@ -1,4 +1,5 @@
 {
+  "configuration.title": "Salesforce Visualforce Configuration",
   "visualforce.format.enable.desc":
     "Enable/disable default Visualforce formatter",
   "visualforce.format.wrapLineLength.desc":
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-visualforce-next/package.json
```diff
@@ -73,7 +73,7 @@
     ],
     "configuration": {
       "type": "object",
-      "title": "Salesforce DX Visualforce Configuration",
+      "title": "%configuration.title%",
```
vazexqi: The VisualForce one is not released yet but this is just a preemptive change.

packages/salesforcedx-vscode-visualforce/package.nls.json
```diff
@@ -1,4 +1,5 @@
 {
+  "configuration.title": "Salesforce Visualforce Configuration",
```
vazexqi: This file follows a different convention since the original language server (the one from VS Code) uses the . convention instead of the _ convention.

PR #184, opened 2017-10-18
Remove "enableProposedApi": true

vazexqi:
### What does this PR do?

We are aware we are using a proposed API for the color preview but there is no loss of functionality for now.

#180

### DIFF
--- a/packages/salesforcedx-vscode-visualforce/package.json
+++ b/packages/salesforcedx-vscode-visualforce/package.json
@@ -1,6 +1,6 @@
 {
   "name": "salesforcedx-vscode-visualforce",
-  "displayName": "Next Visualforce Code Editor for Visual Studio Code",
+  "displayName": "Visualforce Code Editor for Visual Studio Code",
   "description": "Provides syntax highlighting for the Visualforce framework",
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-visualforce/package.json
```diff
@@ -53,7 +53,6 @@
     "test": "node ./node_modules/vscode/bin/test"
   },
   "activationEvents": ["workspaceContains:sfdx-project.json"],
-  "enableProposedApi": true,
```
vazexqi: I'm still able to compile and when I generate the .vsix, I don't see any loss in functionality.

packages/salesforcedx-vscode-visualforce/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "salesforcedx-vscode-visualforce",
-  "displayName": "Next Visualforce Code Editor for Visual Studio Code",
+  "displayName": "Visualforce Code Editor for Visual Studio Code",
```
vazexqi: There shouldn't be the word "Next" in there :-)

PR #187, opened 2017-10-20
Debugger requires 1.17 so the extension pack should also require 1.17

DatGuyJonathan:
### What does this PR do?
Once they install the pack, vscode prevents them from uninstalling the pack's dependencies, so they won't be able to remove the debugger extension. By bumping the pack's vscode version, the error message should be about the pack's required version instead of a specific dependency.

### DIFF
--- a/packages/salesforcedx-slds-linter/package.json
+++ b/packages/salesforcedx-slds-linter/package.json
@@ -7,7 +7,7 @@
   "license": "BSD-3-Clause",
   "categories": ["Other"],
   "engines": {
-    "vscode": "^1.13.0"
+    "vscode": "^1.17.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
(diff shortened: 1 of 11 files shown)


vazexqi:
@DatGuyJonathan - Good call. Let's make the extension pack require 1.17.

Should we also just bump up every other dependency to 1.17 to be consistent? No point having different versions around.

DatGuyJonathan:
I noticed the install numbers for individual extensions were higher than the extension pack, so I think that means some people install the specific extension they want. If we bump all to 1.17, we'd be forcing those people to update their vscode version when the extension doesn't actually need the latest vscode APIs. Is that alright?

guw:
I think we should be consistent across all modules.

vazexqi:
I understand the drawbacks of bumping the requirements up even if an older version could work. 
At the same time, it probably makes it easier to keep track of just one overall version. Our future plans will revolve around using more of the newer APIs from VS Code.

Do you see any risk in bumping up all the versions in this PR? Would you like to do this one first and then create a separate PR?

I guess it would help if we had some statistics on how many people are on different versions of code, but that information is only available to microsoft.

As for the installing individual modules, I do want to encourage people to install the extension pack. Sooner or later things would become quite dependent that it will be easier to just get the extension pack.

DatGuyJonathan:
I just wanted to be on the same page about the effect this bump would have after publishing. I've bumped all vscode reqs to 1.17.

PR #188, opened 2017-10-21
Remove non-public attributes

vazexqi:
### What does this PR do?

We were showing non-public attributes for Visualforce completion. This PR corrects that are puts it more inline with what we have at https://developer.salesforce.com/docs/atlas.en-us.pages.meta/pages/pages_compref.htm

@W-4404980@


vazexqi:
@Sweetman - Can you take a quick look at this? Should be a simple review since all that I am doing is removing some attributes.

PR #191, opened 2017-10-25
Winston/test release

chowwinston:
### What does this PR do?
Explicitly remove dev dependencies prior to packaging
@w-4414355@

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ script:
       npm run test:without-system-tests
     fi
   - codecov
+  - npm run vscode:package
 
 after_script:
-  - npm run vscode:package
   - rm $SFDX_CI_KEY_LOCATION
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,16 +73,17 @@ script:
   - npm run lint
   - |
     if [[ $TRAVIS_OS_NAME == "linux" ]]; then
-      npm run test:without-system-tests
-      npm run coverage:system-tests
+      #npm run test:without-system-tests
+      #npm run coverage:system-tests
     fi
   - |
     if [[ $TRAVIS_OS_NAME == "osx" ]]; then
       # Due to slowness of OS X on Travis
       # https://github.com/travis-ci/travis-ci/issues/6095
-      npm run test:without-system-tests
+      #npm run test:without-system-tests
     fi
-  - codecov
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -85,4 +85,5 @@ script:
   - codecov
 
 after_script:
+  - npm run vscode:package
```
vazexqi: I don't think this can be in the after_script.

See https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build 

    The exit code of after_success, after_failure, after_script and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

We need this to fail the build so it cannot be in the `after_script`.

I'm less with familiar with appveyor but I suspect that you shouldn't put it under the `on_finish` as well.

Was there a reason you chose to put it there?

package.json
```diff
@@ -24,7 +24,7 @@
     "coverage:system-tests":
       "lerna run coverage --scope system-tests --concurrency 1 --stream",
     "vscode:package":
-      "lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1",
+      "lerna exec -- find . -name @types -type d -exec shx rm -rf {} + && lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1",
```
vazexqi: Don't use find. It's behavior might not be standard across different OS.

Won't `lerna --exec shx rm -rf node_modules/@types` do the trick?

Or am I missing something?

.travis.yml
chowwinston: Didn't know failures in after_script don't affect build results. Just thought it'd be a good place for a final sanity check. I'll move both to the script areas.

package.json
chowwinston: That doesn't remove one in system-tests. Wanted to be thorough in case others that actually affect the packaging slip through in the future.

vazexqi: Ok does `find` work on both Linux and Mac? I don't think find works on windows. So you will fail the appveyor build immediately.

chowwinston: Yes works on both but I'll change to your suggestion. I think it's worth having this work on windows even if none of us package there. If people run into issues we can resurface this then.

PR #201, opened 2017-10-31
Remove keybinding fetching and use local bindings from test_data

Sweetman:
### What does this PR do?
Add platform keybinding json files to our test_data directory and read the bindings from them directly.
Autobuilds failing because keybinding retrieval is broken. Used to pull them from vscode-docs but they removed the files in https://github.com/Microsoft/vscode-docs/pull/1229

### DIFF
--- a/packages/system-tests/src/spectron/application.ts
+++ b/packages/system-tests/src/spectron/application.ts
@@ -111,9 +111,23 @@ export class SpectronApplication {
     );
   }
 
+  private getKeybindingPlatform(): string {
+    switch (process.platform) {
+      case 'darwin':
+        return 'osx';
+      case 'win32':
+        return 'win';
+      default:
+        return process.platform;
+    }
+  }
+
   private retrieveKeybindings() {
     fs.readFile(
-      path.join(process.cwd(), `test_data/keybindings.json`),
+      path.join(
+        process.cwd(),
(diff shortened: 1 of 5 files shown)


packages/system-tests/src/spectron/application.ts
```diff
@@ -111,9 +111,23 @@ export class SpectronApplication {
     );
   }
 
+  private getKeybindingPlatform(): string {
```
Sweetman: I followed VSCode's implementation for this - https://github.com/Microsoft/vscode/blob/master/test/smoke/src/main.ts#L118

packages/system-tests/src/spectron/application.ts
```diff
@@ -111,9 +111,23 @@ export class SpectronApplication {
...
+        return process.platform;
+    }
+  }
+
   private retrieveKeybindings() {
     fs.readFile(
-      path.join(process.cwd(), `test_data/keybindings.json`),
+      path.join(
+        process.cwd(),
+        `test_data/keybindings.${this.getKeybindingPlatform()}.json`
```
Sweetman: Read directly from the bindings stored in test_data. `getKeybindingPlatform` will retrieve the bindings for the correct OS.

vazexqi: @Sweetman Actually, I'm surprised that this works. You hardcoded the '/' in the second part.

PR #204, opened 2017-11-01
Fix debugger launch sequence

DatGuyJonathan:
### What does this PR do?
I reverted some changes made in https://github.com/forcedotcom/salesforcedx-vscode/pull/194. How it used to work before that PR:
1. VS Code asks the adapter to initialize.
1. The adapter asks apex extension for line breakpoint info.
1. When the adapter gets them (could take however long), it sends an initialized response to VS Code so it can continue the launch. 

https://github.com/forcedotcom/salesforcedx-vscode/pull/194 replaced that sequence with a setTimeout so after the default timeout, the adapter checks if it has received line breakpoint info. If yes, send an initialized response to continue the launch. If not, send error message. It ensured we error out with a message if it takes too long. However, in the case where line breakpoint info is received before the timeout, we don't check until that setTimeout is triggered. So, the happy path always takes however long the default timeout is.

I'm reverting back to the old sequence while keeping the setTimeout to just check if it does not already have line breakpoint info.

@W-4428718@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -463,6 +463,7 @@ export class ApexDebug extends LoggingDebugSession {
   protected variableContainerReferenceByApexId = new Map<number, number>();
 
   private static LINEBREAK = `${os.EOL}`;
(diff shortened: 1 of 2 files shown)


guw:
Makes sense!

PR #206, opened 2017-11-02
Run single apex test with synchronous option

DatGuyJonathan:
### What does this PR do?
This adds the ability to run force:apex:test:run command with a single Apex class. The command always executes with --synchronous option so the test is debuggable.

debugapextest

@W-4337479@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
@@ -9,6 +9,7 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
```diff
@@ -21,47 +22,86 @@ import {
...
-        description: file.fsPath
+        description: testSuite.fsPath,
+        type: TestType.Suite
       };
     });
 
+    const apexClasses = await vscode.workspace.findFiles('**/*.cls');
+    apexClasses.forEach(apexClass => {
+      const fileContent = fs.readFileSync(apexClass.fsPath).toString();
+      if (fileContent && fileContent.toLowerCase().includes('@istest')) {
```
DatGuyJonathan: Find files with .cls extension and filter the ones with "@istest" in the file.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
```diff
@@ -21,47 +22,86 @@ import {
...
+        .withFlag('--suitenames', `${data.label}`)
         .withFlag('--resultformat', 'human')
         .build();
+    } else if (data.type === TestType.Class) {
+      return new SfdxCommandBuilder()
+        .withDescription(nls.localize('force_apex_test_run_text'))
+        .withArg('force:apex:test:run')
+        .withFlag('--classnames', `${data.label}`)
+        .withFlag('--resultformat', 'human')
+        .withArg('--synchronous')
```
DatGuyJonathan: This runs the single test class with Tooling API's /runTestsSynchronous. Requires the user to have updated to the CLI plugin version with that feature.

vazexqi:
Looks good. In a future PR, I would like to make it so that there should be a light-build next to the class when you hover it, it also offers the option to run the test class. This would require modifying the Apex language server to support [code actions](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_codeAction).

Questions:

1. Do we also have the ability to run just a single test method?
2. Do we need to provide the facility to run a test class in async mode? Or if you select just a single class, we should just run it synchronously? I think that makes sense but I just want to consider the alternatives.

I think this is fine to go in for now, but we should probably hook this up into the apex language server as well to get the list of test classes in the future.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
vazexqi: This is where we could use the apex language server in the future.

DatGuyJonathan:
1. Tooling API supports running specific test methods, but the CLI does not.
2. I don't think there's any advantage for the user to run a single test class with the asynchronous endpoint. I'd rather have just one way for them to run a single test class that works for both debuggers and non-debuggers. More options could confuse users.

PR #209, opened 2017-11-08
Paginate collections

DatGuyJonathan:
### What does this PR do?
paginatedcollections

@W-4356830@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -121,15 +121,22 @@ export enum ApexVariableKind {
 export class ApexVariable extends Variable {
   public readonly declaredTypeRef: string;
   public readonly type: string;
+  public readonly indexedVariables?: number;
   private readonly slot: number;
   private readonly kind: ApexVariableKind;
 
   constructor(
     value: Value,
     kind: ApexVariableKind,
-    variableReference?: number
+    variableReference?: number,
+    numOfChildren?: number
   ) {
-    super(value.name, ApexVariable.valueAsString(value), variableReference);
+    super(
+      value.name,
+      ApexVariable.valueAsString(value),
+      variableReference,
+      numOfChildren
+    );
     this.declaredTypeRef = value.declaredTypeRef;
     this.kind = kind;
     this.type = value.nameForMessages;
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -121,15 +121,22 @@ export enum ApexVariableKind {
...
-    variableReference?: number
+    variableReference?: number,
+    numOfChildren?: number
   ) {
-    super(value.name, ApexVariable.valueAsString(value), variableReference);
+    super(
+      value.name,
+      ApexVariable.valueAsString(value),
+      variableReference,
+      numOfChildren
```
DatGuyJonathan: This ApexVariable class extends VS Code debug protocol's Variable which has these fields:

debugprotocolvariable

If VS Code receives a large indexedVariables number, it will automatically paginate with 100 in each page. So, the important thing for our debugger adapter is to save the collection's reference size and assign that to indexedVariables property, which is what we're doing here.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -304,38 +324,55 @@ export class ObjectReferenceContainer implements VariableContainer {
...
+    if (start === undefined) {
+      start = 0;
+    }
+    if (count === undefined) {
+      count = this.reference.value.length;
+    }
+    const apexVariables: ApexVariable[] = [];
+    for (
+      let i = start;
+      i < start + count && i < this.reference.value.length;
```
DatGuyJonathan: Previously, it would fetched every variable in the collection. Now, we fetch from the provided `start` until `start + count`. Those values are provided by variablesRequest() which is a request for a variable from VS Code to the adapter. VS Code knows whether or not a variable has paginated children, so it will automatically provide those start and count values depending on which page the user expands.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1142,6 +1195,17 @@ export class ApexDebug extends LoggingDebugSession {
     });
   }
 
+  public getNumOfChildren(
```
guw: I'd prefer the full name, i.e. `getNumberOfChildren`.

packages/salesforcedx-apex-debugger/test/unit/adapter/apexDebugVariablesHandling.test.ts
```diff
@@ -358,6 +385,8 @@ describe('Debugger adapter variable handling - unit', () => {
 
       expect(container).to.be.ok;
       expect(container).to.be.instanceOf(CollectionReferenceContainer);
+      expect(container!.getNumberOfChildren()).to.equal(1);
```
guw: Is the `'!'` (exclamation mark) really necessary given the assert in the previous line?

DatGuyJonathan: Yea, otherwise it complains container could be undefined.

PR #219, opened 2017-11-17
Terminate idle debugger session

DatGuyJonathan:
### What does this PR do?
Prompts the user with a non-modal dialog if their debugger session has been idle too long. 

idleprompt

If the user doesn't respond to the prompt, their debugger session is terminated.

idleterminated

@W-4179093@


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1034,6 +1039,23 @@ export class ApexDebug extends LoggingDebugSession {
           typerefs: Array.from(exceptionBreakpoints.keys())
         };
         break;
+      case TERMINATE_SESSION_REQUEST:
+        this.sendEvent(new TerminatedEvent());
```
DatGuyJonathan: A request from the extension to the adapter to terminate the session because the user responded with `No` to the question `Are you still debugging?`

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1034,6 +1039,23 @@ export class ApexDebug extends LoggingDebugSession {
...
+        break;
+      case IDLE_SESSION_REQUEST:
```
DatGuyJonathan: A request from the extension to the adapter to terminate the session because the user took too long to respond to the prompt.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1034,6 +1039,23 @@ export class ApexDebug extends LoggingDebugSession {
...
+        this.warnToDebugConsole(nls.localize('terminate_idle_session_text'));
+        this.sendEvent(new TerminatedEvent());
+        break;
+      case SEND_HEARTBEAT_REQUEST:
```
DatGuyJonathan: A request from the extension to the adapter to keep the debugger session alive.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1113,6 +1135,7 @@ export class ApexDebug extends LoggingDebugSession {
       );
     }
 
+    this.sendEvent(new Event(RESET_HEARTBEAT_EVENT));
```
DatGuyJonathan: This is in the method variablesRequest(). We use this to tell the extension to reset the heartbeat timer.

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
```diff
@@ -32,5 +32,7 @@ export const messages = {
   hotswap_warn_text:
     "You can't modify Apex classes or triggers during an Apex Debugger session. Save your changes after you're done debugging.",
   created_exception_breakpoint_text: 'Created exception breakpoint for %s.',
-  removed_exception_breakpoint_text: 'Removed exception breakpoint for %s.'
+  removed_exception_breakpoint_text: 'Removed exception breakpoint for %s.',
+  terminate_idle_session_text:
+    'Your debugger session is being terminated because it has been idle for a while.'
```
DatGuyJonathan: If the user doesn't respond to the prompt, we output this to the Debug Console.

idleterminated

packages/salesforcedx-vscode-apex-debugger/src/messages/i18n.ts
```diff
@@ -19,5 +19,9 @@ export const messages = {
   select_exception_text: 'Select an exception',
   select_break_option_text: 'Select break option',
   always_break_text: 'Always break',
-  never_break_text: 'Never break'
+  never_break_text: 'Never break',
+  are_you_still_debugging_text:
+    'We noticed that you have been idle for awhile. Are you still debugging?',
```
DatGuyJonathan: Shown at the top of VS Code when the idle-session timer is triggered:

idleprompt

ruthemmanuelle: "awhile" should be "a while" (http://www.quickanddirtytips.com/education/grammar/while-versus-awhile)

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
ruthemmanuelle: Can we say "it has been idle for an hour" instead of "it has been idle for a while"?

packages/salesforcedx-apex-debugger/src/constants.ts
```diff
@@ -9,6 +9,8 @@ export const DEFAULT_INITIALIZE_TIMEOUT_MS = 10000;
 export const DEFAULT_STREAMING_TIMEOUT_MS = 14400;
 export const DEFAULT_CONNECTION_TIMEOUT_MS = 20000;
 export const DEFAULT_LOCK_TIMEOUT_MS = 10000;
+export const DEFAULT_IDLE_TIMEOUT_MS = 1200000;
+export const DEFAULT_IDLE_QUESTION_TIMEOUT_MS = 10000;
```
DatGuyJonathan: 10 seconds for the user to respond to the prompt.

packages/salesforcedx-apex-debugger/src/constants.ts
```diff
@@ -9,6 +9,8 @@ export const DEFAULT_INITIALIZE_TIMEOUT_MS = 10000;
 export const DEFAULT_STREAMING_TIMEOUT_MS = 14400;
 export const DEFAULT_CONNECTION_TIMEOUT_MS = 20000;
 export const DEFAULT_LOCK_TIMEOUT_MS = 10000;
+export const DEFAULT_IDLE_TIMEOUT_MS = 1200000;
```
DatGuyJonathan: 20 minutes is the idle timeout.

ruthemmanuelle: Ah, ok. I assumed it was an hour because that was what we used in Eclipse. You can change "for an hour" to "for 20 minutes" in my suggested message.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: @DatGuyJonathan Why is this distinction necessary?

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -102,21 +119,54 @@ function registerCommands(): vscode.Disposable {
...
+    terminateSessionHandler
   );
 }
 
+export function areYouStillDebugging(
+  session: vscode.DebugSession
+): NodeJS.Timer | undefined {
+  return setInterval(async () => {
+    const innerTimer = setTimeout(() => {
+      session.customRequest(IDLE_SESSION_REQUEST);
```
guw: @DatGuyJonathan From a usability point this is worrisome. Microsoft argues a warning message must be confirmed by a user. That's why there is no programmatic way to close those. Thus, as long as the message is visible and the user has not made a decision yet we must keep the session alive. We are suggesting a continue is possible by presenting the option. Silently timing out is not good (IMO).

guw:
One more thought: I recall that one design goal is to terminate a connections as quickly as possible (due to resource reasons). In this case, we do not want to keep the session alive while the user is making a decision. So why should we pretend to users they have a choice if we hard kill their session anyway when not making a timely answer?

What about the following workflow:
* No debug activity for 10 minutes triggers warning first warning logged to Debug Console
* Still no activity after another 5 minutes (15 min total) triggers another warning logged to Debug Console, also Debug Console panel is given focus (`workbench.debug.action.focusRepl`)
* Still no activity after 4 minutes (19 min total), final warning that debug session will be terminated in about 1 minute, also Debug Console panel is given focus (`workbench.debug.action.focusRepl`)
* Still no activity after 1 minute, debug session is terminated

If - however - the goal to allow users to keep the debug session alive is more important that saving shared resources we could experiment with a `StatusBarItem` instead of a warning message. The `StatusBarItem` can have a countdown as well as an icon. On click could become the "keep alive" meaning and triggers a heartbeat as well as resetting the counter.

DatGuyJonathan:
Going to close this and create a new PR with @guw suggestions.

PR #221, opened 2017-11-21
Terminate idle debugger session

DatGuyJonathan:
### What does this PR do?
As suggested in https://github.com/forcedotcom/salesforcedx-vscode/pull/219, VS Code's warning message doesn't make much sense if we end up terminating the idle session anyways due to lack of response. Instead, we'll print warning messages to Debug Console at multiple intervals. The user can step, run, or inspect variables to reset the idle timer.

idleoutputs

The message would look something like that, except in minutes and they will be outputted at the 10th, 15th, 18th minute, 20th minute.

@W-4179093@


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -623,6 +682,7 @@ export class ApexDebug extends LoggingDebugSession {
           nls.localize('session_started_text', sessionId)
         );
         this.sendEvent(new InitializedEvent());
+        this.resetIdleTimer();
```
DatGuyJonathan: Start idle timers on session launch.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -662,6 +722,9 @@ export class ApexDebug extends LoggingDebugSession {
     } else {
       response.success = true;
     }
+    if (response.success) {
+      this.clearIdleTimers();
```
DatGuyJonathan: Clear idle timers on session disconnect.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -740,6 +803,7 @@ export class ApexDebug extends LoggingDebugSession {
         response.message = error;
       }
     }
+    this.resetIdleTimer();
```
DatGuyJonathan: Reset idle timer when user selects "Run". This is needed in case they run to the end of a debugged request. There's no request for variables, but we still need to reset the timer because they technically did something with the debugger.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1126,6 +1190,7 @@ export class ApexDebug extends LoggingDebugSession {
       );
       variables.sort(ApexVariable.compareVariables);
       response.body = { variables: variables };
+      this.resetIdleTimer();
```
DatGuyJonathan: Reset idle timers when a variable is inspected (either by expanding a variable or pausing the apex runtime).

packages/salesforcedx-apex-debugger/src/constants.ts
```diff
@@ -9,6 +9,10 @@ export const DEFAULT_INITIALIZE_TIMEOUT_MS = 10000;
 export const DEFAULT_STREAMING_TIMEOUT_MS = 14400;
 export const DEFAULT_CONNECTION_TIMEOUT_MS = 20000;
 export const DEFAULT_LOCK_TIMEOUT_MS = 10000;
+export const DEFAULT_IDLE_TIMEOUT_MS = 1200000;
+export const DEFAULT_IDLE_WARN1_MS = 600000;
+export const DEFAULT_IDLE_WARN2_MS = 900000;
+export const DEFAULT_IDLE_WARN3_MS = 1080000;
```
DatGuyJonathan: Warn at minutes 10, 15, 18. Terminate at minute 20.

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
```diff
@@ -32,5 +32,9 @@ export const messages = {
   hotswap_warn_text:
     "You can't modify Apex classes or triggers during an Apex Debugger session. Save your changes after you're done debugging.",
   created_exception_breakpoint_text: 'Created exception breakpoint for %s.',
-  removed_exception_breakpoint_text: 'Removed exception breakpoint for %s.'
+  removed_exception_breakpoint_text: 'Removed exception breakpoint for %s.',
+  idle_warn_text:
+    'You have been idle for %s minutes. Your debugger session will be terminated in %s minutes if it continues to be idle.',
+  idle_terminated_text:
+    'Your debugger session is being terminated because it has been idle for %s minutes.'
```
DatGuyJonathan: @ruthemmanuelle , see screenshot above for these messages. Is this enough to suggest that the user needs to do something with the debugger to not have their session terminated?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: I think this should be wrapped in a finally, i.e. we want to clear the timers in any case. Also, what happens is the response ain't successful? Shouldn't this be considered a dead session anyway and the timers cleared no matter what?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
guw: This is very visible in the actual debugger logic. I think it would be better to move the timeout lifecycle handling to a lower level. For example, what about overriding `dispatchRequest`? This would reset the timeout on every request from VS Code.

How is the timeout handle on the server? Would it also make sense to reset timers in `handleEvent`, i.e. whenever we receive an event from the server?

What about `handleSessionTerminated`? Shouldn't this clear timers when the server terminates a session?

DatGuyJonathan: @guw The server side inserts a heartbeat whenever it gets the state or a frame. A debugger client gets the state when we stop (because of either a breakpoint or step command). A debugger client gets the frame if the user selects a different frame to inspect. Either way, those actions involve getting variables which is why I put the reset in `variablesRequest`.

`handleSessionTerminated` sends a `TerminatedEvent` to VS Code, which in turn will send a `disconnectRequest` to the adapter which already clears the timer.

packages/salesforcedx-apex-debugger/src/messages/i18n.ts
ruthemmanuelle: Would something like this work for idle_warn_text?
'You have been idle for %s minutes. To prevent your debugger session from being terminated, run or step through code, or inspect your variables, in the next %s minutes.'

guw:
Thanks for clarification @DatGuyJonathan!

PR #224, opened 2017-11-29
James/create trigger

Sweetman:
### What does this PR do?
Adds create trigger command to vscode command palette
@W-4470502@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -106,6 +106,10 @@
         {
           "command": "sfdx.force.lightning.interface.create",
           "when": "explorerResourceIsFolder && sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.apex.trigger.create",
+          "when": "explorerResourceIsFolder && sfdx:project_opened"
         }
       ],
       "commandPalette": [
@@ -227,6 +231,10 @@
         },
         {
           "command": "sfdx.force.project.create"
+        },
+        {
+          "command": "sfdx.force.apex.class.create",
+          "when": "sfdx:project_opened"
         }
       ]
     },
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -227,6 +231,10 @@
         },
         {
           "command": "sfdx.force.project.create"
+        },
+        {
+          "command": "sfdx.force.apex.class.create",
```
DatGuyJonathan: This should be the trigger command ID.

packages/system-tests/scenarios/scaffolding.test.ts
```diff
@@ -248,7 +275,7 @@ describe('Empty VSCode workspace', () => {
     return await app.stop();
   });
 
-  it('Should not show any commands for non-SFDX workspace', async () => {
+  it('Should only show create project command for non-SFDX workspace', async () => {
```
ruthemmanuelle: I assume this is internal, but it would be clearer if you moved "only" before "for".

Sweetman: "Should show create project command **only** for non-SFDX workspace" is incorrect since it's designed to show up in SFDX workspaces and non-SFDX workspaces but it will be the only command to show up in a non-SFDX workspace. I wrote it like this to be a little bit more specific than "Should only show one command for non-SFDX workspace" but I can use that instead if that's more clear?

DatGuyJonathan:
@Sweetman , the test will not pass on the autobuild until your CLI feature is released, which is fine as long as you've verified the test locally with the updated CLI. Is the PR titled "DO NOT MERGE" because you still have other changes pending? Or do you want me to merge this in for you?

DatGuyWidjaja:
You could also have a unit test for now that verifies the command builder, then enable the system test after the CLI feature has been released. That option seems appealing so we maintain a green appveyor build.

Sweetman:
@DatGuyJonathan @DatGuyWidjaja I put "DO NOT MERGE" because I wanted to wait for the CLI command branch to be merged in first. I'll add the unit test, disable the system-test, and then merge in if appveyor passes

PR #225, opened 2017-11-29
Update Build Infrastructure

guw:
### What does this PR do?
Update our Travis CI Build Infra to use Ubuntu Trusty and XCode 9.1 and also reference VS Code 1.19 to work around an bug in 1.18.

Fixes the Linux build issues and improved the macOS build.


guw:
@vazexqi I wasn't able to get the tests running with the insiders build. It's currently not supported by VS Code. I've opened a feature request to get it added (Microsoft/vscode-extension-vscode#88). 

I'd still like to merge this PR because it contains a few good tweaks to our builds.

.travis.yml
```diff
@@ -15,7 +15,7 @@ notifications:
 
 matrix:
   allow_failures:
-    - os: osx
+    - os: linux
```
vazexqi: Can we specify it for both linux and osx? The way you have done it now makes is so that osx must pass, which we know is a losing battle given https://github.com/travis-ci/travis-ci/issues/7304

.appveyor.yml
```diff
@@ -70,5 +70,5 @@ test_script:
 
 on_finish:
   - del C:\projects\devhub.key
-  - codecov
+  - codecov --disable=gcov
```
vazexqi: How does this help with long path issues, and why does it only happen now?

vazexqi:
@guw This looks fine to me. Just a couple of questions. I think we do need to allow osx builds to fail, since they will due to the slow builds.

.travis.yml
guw: Done.

.appveyor.yml
guw: The codecov tool gcov module was scanning very aggressiv for coverage reports. It also scanned the node_modules folders and caused those errors to appear in the build log. Disabling this made those errors go away and our build pass. I wonder if this was the result of a resent update to the codecov tool - we install it at build time via NPM.

PR #227, opened 2017-12-03
Prevent repeats of Refresh SObject while running

GaryFNC:
The Refresh SObject Definitions command can take a few seconds.
If the user hits the command again while running, it will do the command
all over again when the first one ends - not intended behavior.
Add checks and a warning message to the command forceGenerateFauxClasses.ts
to disallow multiple invocations at the same time.

@W-4471720@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -107,5 +107,7 @@ export const messages = {
   force_sobjects_refresh: 'SFDX: Refresh SObject Definitions',
   force_project_create_text: 'SFDX: Create Project',
   force_project_create_open_dialog_create_label: 'Create Project',
-  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger'
+  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger',
+  no_refresh_if_currently_active:
+    'Refresh is already active.  Cancel the existing task if there is a need to restart'
 };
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -107,5 +107,7 @@ export const messages = {
   force_sobjects_refresh: 'SFDX: Refresh SObject Definitions',
   force_project_create_text: 'SFDX: Create Project',
   force_project_create_open_dialog_create_label: 'Create Project',
-  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger'
+  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger',
+  no_refresh_if_currently_active:
+    'Refresh is already active.  Cancel the existing task if there is a need to restart'
```
ruthemmanuelle: We use only one space after periods. We also have a "goal must precede action" style guideline, which basically means that "if xyz" should come before "then abc." Also, can we be more specific about what kind of refresh this is, or is this message likely to be used for different types of refreshes? Maybe something like this would work?

'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.'

chowwinston: Group this message along with the other sobject one, and make the name conform to the conventions.

GaryFNC: @ruthemmanuelle @chowwinston  done - please verify

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -108,6 +108,6 @@ export const messages = {
   force_project_create_text: 'SFDX: Create Project',
   force_project_create_open_dialog_create_label: 'Create Project',
   force_apex_trigger_create_text: 'SFDX: Create Apex Trigger',
-  no_refresh_if_currently_active:
-    'Refresh is already active.  Cancel the existing task if there is a need to restart'
+  force_sobjects_no_refresh_if_already_active_error_text:
```
chowwinston: Move this under line 107 so it's next to the other sobject text. Feel free to merge after that.

PR #228, opened 2017-12-03
Add an explanatory comment to the generated SObject faux classes

GaryFNC:
…lasses

The generated faux classes are intended for use by the Apex Language Server
and for viewing by a developer.   They are not to be edited and an explanatory
comment about being generated is added.

The generation is also changed to make the generated files read-only.

@W-4477572@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
@@ -24,6 +24,43 @@ describe('SObject faux class generator', function() {
     }
   });
 
+  it('Should generate a faux class with a proper header comment', async () => {
+    const fieldsHeader = '{ "name": "Custom__c", "fields": [ ';
+    const closeHeader = ' ], "childRelationships": [] }';
+
+    const sobject1 = `${fieldsHeader}${closeHeader}`;
+
+    const sobjectFolder = process.cwd();
+    const gen = getGenerator();
+    classPath = await gen.generateFauxClass(
+      sobjectFolder,
+      JSON.parse(sobject1)
+    );
+    expect(fs.existsSync(classPath));
+    const classText = fs.readFileSync(classPath, 'utf8');
+    expect(classText).to.include(
+      '// This file is generated as an Apex representation'
+    );
+  });
+
+  it('Should generate a faux class as read-only', async () => {
+    const fieldsHeader = '{ "name": "Custom__c", "fields": [ ';
+    const closeHeader = ' ], "childRelationships": [] }';
+
+    const sobject1 = `${fieldsHeader}${closeHeader}`;
+
+    const sobjectFolder = process.cwd();
+    const gen = getGenerator();
+    classPath = await gen.generateFauxClass(
+      sobjectFolder,
+      JSON.parse(sobject1)
+    );
+    expect(fs.existsSync(classPath));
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
```diff
@@ -24,6 +24,43 @@ describe('SObject faux class generator', function() {
     }
   });
 
+  it('Should generate a faux class with a proper header comment', async () => {
+    const fieldsHeader = '{ "name": "Custom__c", "fields": [ ';
+    const closeHeader = ' ], "childRelationships": [] }';
+
+    const sobject1 = `${fieldsHeader}${closeHeader}`;
```
chowwinston: Why bother splitting this into 2 variables?

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
```diff
@@ -24,6 +24,43 @@ describe('SObject faux class generator', function() {
...
+    const sobjectFolder = process.cwd();
+    const gen = getGenerator();
+    classPath = await gen.generateFauxClass(
+      sobjectFolder,
+      JSON.parse(sobject1)
+    );
+    expect(fs.existsSync(classPath));
+    const classText = fs.readFileSync(classPath, 'utf8');
+    expect(classText).to.include(
+      '// This file is generated as an Apex representation'
```
chowwinston: You should reference the message variable's value directly rather than a random substring here.

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
```diff
@@ -24,6 +24,43 @@ describe('SObject faux class generator', function() {
...
+    );
+  });
+
+  it('Should generate a faux class as read-only', async () => {
+    const fieldsHeader = '{ "name": "Custom__c", "fields": [ ';
+    const closeHeader = ' ], "childRelationships": [] }';
+
+    const sobject1 = `${fieldsHeader}${closeHeader}`;
```
chowwinston: Just have 1 field called sobject

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
GaryFNC: because the other tests were already set up that way - they have extra fields that weren't necessary for this test.  Seemed simpler to keep the same pattern rather that do something special here.

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
GaryFNC: yes, probably cleaner - I was checking for just a subset, but safer to use what is in the messages file - especially now that we are talking about making it internationalizable.  Will do

GaryFNC:
@ruthemmanuelle @chowwinston  changes pushed - please verify

packages/salesforcedx-sobjects-faux-generator/test/fauxClassGenerator.test.ts
chowwinston: Having 3 separate variables for 1 JSON string is more convoluted than being consistent across various unit tests.

PR #235, opened 2017-12-08
[WIP] James/api logging

Sweetman:
### What does this PR do?
Add logging to API calls made from VS Code for commands palette commands and for commands executed through the integrated terminal
@W-4351074@, @W-4497742@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -11,6 +11,7 @@ import {
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 import { xhr, XHROptions, XHRResponse } from 'request-light';
+import { CLIENT_ID } from '../constants';
 
 export interface SObject {
   actionOverrides: any[];
@@ -244,7 +245,8 @@ export class SObjectDescribe {
         'Content-Type': 'application/json',
         Accept: 'application/json',
         Authorization: `OAuth ${this.accessToken}`,
-        'User-Agent': 'salesforcedx-extension'
+        'User-Agent': 'salesforcedx-extension',
+        'Sforce-Call-Options': `client=${CLIENT_ID}`
       }
     };
 
@@ -330,7 +332,8 @@ export class SObjectDescribe {
         'Content-Type': 'application/json',
         Accept: 'application/json',
         Authorization: `OAuth ${this.accessToken}`,
-        'User-Agent': 'salesforcedx-extension'
+        'User-Agent': 'salesforcedx-extension',
+        'Sforce-Call-Options': `client=${CLIENT_ID}`
       },
       data: JSON.stringify(batchRequest)
     };
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -6,6 +6,7 @@
  */
 
 import * as vscode from 'vscode';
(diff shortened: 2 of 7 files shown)


packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -244,7 +245,8 @@ export class SObjectDescribe {
         'Content-Type': 'application/json',
         Accept: 'application/json',
         Authorization: `OAuth ${this.accessToken}`,
-        'User-Agent': 'salesforcedx-extension'
+        'User-Agent': 'salesforcedx-extension',
+        'Sforce-Call-Options': `client=${CLIENT_ID}`
```
Sweetman: This will add the logging for direct REST calls. I followed Jonathan's example from the apex-debugger package. One thing I would like to note here is that I noticed that there are no outside references to describeSObject (this method) other than a test for the method. If we're not using it, should we remove it?

/cc @GaryFNC

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -227,6 +234,27 @@ export async function activate(context: vscode.ExtensionContext) {
     sfdxProjectOpened = files && files.length > 0;
   }
 
+  // Set environment variable to add logging for VSCode API calls
+  process.env.SFDX_SET_CLIENT_IDS = CLIENT_ID;
```
Sweetman: Process environment will add logging to CLI commands executed through the command palette.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -227,6 +234,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
+  const config = vscode.workspace.getConfiguration();
+  config.update(
+    TERMINAL_INTEGRATED_ENV_MAC,
+    { SFDX_SET_CLIENT_IDS: CLIENT_ID },
+    ConfigurationTarget.Global
```
Sweetman: This will add the environment variable to integrated terminals before they are started.

vazexqi: What does the user workspace.json look like when this is added? Can we also add some `// This was added for logging...` in the json file? I don't want to just surprise them.

Also, this is idempotent, right? It will only add it if it is not there. And subsequent runs will not re-populate?

Sweetman: It does not look like there is a way to add a comment through the update method since it can only update a known section key.

Thought about it further and this will not only update that field. Unfortunately, we cannot specify updating only SFDX_SET_CLIENT_IDS because this must be registered. I did not take into account that the user might have their own env vars set in this as well. I will need to get their current configuration and append our env var to it. If the user changed SFDX_SET_CLIENT_IDS to something else, do we respect that or do we override it with ours?

vazexqi: It's OK to override `SFDX_SET_CLIENT_IDS` with just ours. That is not a generally used variable and changing it doesn't have any effects on functionality.

Sweetman:
Marking as WIP, need to add tests on Monday.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
GaryFNC: I left describeSObject  there (it originally preceded the batch calls) because there is a good chance it will be used when processing the XML for custom objects.     But it could be removed and added back later if needed.  OK either way

PR #240, opened 2017-12-21
Add visual indication for exception being paused on

Sweetman:
### What does this PR do?
Add a visual indicator telling the user what exception the user is paused on
image

@W-4518683@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -1702,7 +1702,7 @@ export class ApexDebug extends LoggingDebugSession {
     this.sendEvent(new TerminatedEvent());
   }
 
-  private handleStopped(message: DebuggerMessage): void {
+  private async handleStopped(message: DebuggerMessage): Promise<void> {
     const threadId = this.getThreadIdFromRequestId(message.sobject.RequestId);
 
     if (threadId !== undefined) {
@@ -1720,8 +1720,25 @@ export class ApexDebug extends LoggingDebugSession {
 
       // log to console and notify client
       this.logEvent(message);
+      let reason = '';
+
+      // if breakpointid is found in exception breakpoint cache
+      // set the reason for stopped event to that reason
+      if (message.sobject.BreakpointId) {
+        const cache: Map<
+          string,
+          string
+        > = this.myBreakpointService.getExceptionBreakpointCache();
+        cache.forEach((value, key) => {
+          if (value === message.sobject.BreakpointId) {
+            if (key.indexOf('Exception') !== -1) {
+              reason = key.substring(key.lastIndexOf('/') + 1);
+            }
+          }
+        });
+      }
       const stoppedEvent: DebugProtocol.StoppedEvent = new StoppedEvent(
-        '',
+        reason,
         threadId
       );
       this.sendEvent(stoppedEvent);
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -63,7 +63,9 @@ import {
   HOTSWAP_REQUEST,
   LINE_BREAKPOINT_INFO_REQUEST,
   LIST_EXCEPTION_BREAKPOINTS_REQUEST,
+  SALESFORCE_EXCEPTION_PREFIX,
   SHOW_MESSAGE_EVENT,
+  TRIGGER_EXCEPTION_PREFIX,
   WORKSPACE_SETTINGS_REQUEST
 } from '../constants';
 import {
@@ -1734,8 +1736,8 @@ export class ApexDebug extends LoggingDebugSession {
             // typerefs for exceptions will change based on whether they are custom,
             // defined as an inner class, defined in a trigger, or in a namespaced org
             reason = key
-              .replace('com/salesforce/api/exception/', '')
-              .replace('__sfdc_trigger/', '')
+              .replace(SALESFORCE_EXCEPTION_PREFIX, '')
+              .replace(TRIGGER_EXCEPTION_PREFIX, '')
               .replace('$', '.')
               .replace('/', '.');
           }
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1720,8 +1720,25 @@ export class ApexDebug extends LoggingDebugSession {
 
       // log to console and notify client
       this.logEvent(message);
+      let reason = '';
+
+      // if breakpointid is found in exception breakpoint cache
+      // set the reason for stopped event to that reason
+      if (message.sobject.BreakpointId) {
+        const cache: Map<
```
Sweetman: @DatGuyJonathan not sure if this is what you had in mind for getting the exception type. The other way of easily getting it is by parsing `message.sobject.Description`. 

Another option would be to inverse the exception breakpoint cache map so that breakpoint id are the keys and the type is the value but this would require changing the exception breakpoint reconciliation.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1702,7 +1702,7 @@ export class ApexDebug extends LoggingDebugSession {
     this.sendEvent(new TerminatedEvent());
   }
 
-  private handleStopped(message: DebuggerMessage): void {
+  private async handleStopped(message: DebuggerMessage): Promise<void> {
```
DatGuyJonathan: You're not doing any awaits, so this shouldn't have to be async.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1720,8 +1720,25 @@ export class ApexDebug extends LoggingDebugSession {
...
+      // if breakpointid is found in exception breakpoint cache
+      // set the reason for stopped event to that reason
+      if (message.sobject.BreakpointId) {
+        const cache: Map<
+          string,
+          string
+        > = this.myBreakpointService.getExceptionBreakpointCache();
+        cache.forEach((value, key) => {
+          if (value === message.sobject.BreakpointId) {
+            if (key.indexOf('Exception') !== -1) {
```
DatGuyJonathan: Did you run into a scenario where the typeref didn't have the substring `Exception`?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1720,8 +1720,25 @@ export class ApexDebug extends LoggingDebugSession {
...
+              reason = key.substring(key.lastIndexOf('/') + 1);
```
DatGuyJonathan: What does this look like when the org is namespaced? 

How about an inner exception?
`public class MyClass {
  public class MyInnerException extends Exception {}
}`

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
Sweetman: No, think I forgot that the cache only holds the exception breakpoints, will remove the check.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
Sweetman: The listing does not provide namespaces for the inner exceptions. I created two classes MyClass and MyClass2 both with MyInnerException classes and the list shows the following:
image

DatGuyJonathan: There's a work item for the apex-jorje bug shown in the screenshot. The work item suggest apex-jorje should label it MyClass.MyInnerException (instead of just MyInnerException) because that's what the user would have to type if they were writing Apex code using an inner class. That bug shouldn't affect what you're doing here. The breakpoint cache is keyed by typeref, so for the case of an inner class, the typeref would be MyClass$MyInnerException. My suspicion is the visual indicator shows "Paused on MyClass$MyInnerException". If that's true, could you update the reason to use a period instead of a dollar sign?

Sweetman: Had a lot of issues getting this set up locally but finally able to confirm - it does show up as you suspected with "Paused on MyClass$MyInnerException". I'll replace the dollar sign with a period.

For a namespaced org, the type ref will be "namespace/MyClass$MyInnerException" but we will only show the exception name ("MyClass$MyInnerException). What should the behavior be for this? Should we show the full type ref or just the exception? Dollar signs will be changed to periods.

Sweetman: @DatGuyJonathan here's what I have them looking like right now.
image

image

DatGuyJonathan: We should show namespace.MyClass.MyInnerException. The use case would be an ISV debugging a namespaced subscriber org. The ISV's package would be namespaced so we should make it clear which namespace the exception class belongs to.

I suspect you're using `lastIndexOf('/')` because System exceptions are prefixed with `com/salesforce/api/exception/`. The alternative would be to just remove `com/salesforce/api/exception/` so then you can replace the remaining `/` with a period.

I believe you can also have an exception class inside a trigger. The typeref would be __sfdc_trigger/MyTrigger$MyException, so you'd have to remove `__sfdc_trigger/` as well.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1720,8 +1720,29 @@ export class ApexDebug extends LoggingDebugSession {
...
+          string,
+          string
+        > = this.myBreakpointService.getExceptionBreakpointCache();
+        cache.forEach((value, key) => {
+          if (value === message.sobject.BreakpointId) {
+            // typerefs for exceptions will change based on whether they are custom,
+            // defined as an inner class, defined in a trigger, or in a namespaced org
+            reason = key
+              .replace('com/salesforce/api/exception/', '')
+              .replace('__sfdc_trigger/', '')
```
DatGuyJonathan: Can we pull these two prefixes out to constants.ts so they can also be used in the tests?

PR #242, opened 2018-01-04
Upgrade to TypeScript 2.6.2

vazexqi:
### What does this PR do?

Switch salesforcedx-vscode over to use TypeScript 2.6 instead of TypeScript 2.4. 

TypeScript 2.6 was released in October so it’s stable now. See https://blogs.msdn.microsoft.com/typescript/2017/10/31/announcing-typescript-2-6/

None. This is an engineering commit.

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
@@ -99,7 +99,6 @@ describe('Generate faux classes for SObjects', function() {
   });
 
   it('Should emit an error event on failure', async () => {
-    let result = '';
     let errorMessage = '';
     let exitCode: number = LocalCommandExecution.SUCCESS_CODE;
     let rejectOutput = '';
@@ -112,7 +111,7 @@ describe('Generate faux classes for SObjects', function() {
     });
     invalidateProject(projectPath);
     try {
-      result = await generator.generate(projectPath, SObjectCategory.CUSTOM);
+      await generator.generate(projectPath, SObjectCategory.CUSTOM);
     } catch (e) {
       rejectOutput = e;
     } finally {
@@ -128,7 +127,6 @@ describe('Generate faux classes for SObjects', function() {
   });
 
   it('Should emit message to stderr on failure', async () => {
-    let result = '';
     let stderrInfo = '';
     let rejectOutput = '';
     const generator = getGenerator();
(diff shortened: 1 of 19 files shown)


packages/salesforcedx-visualforce-language-server/src/modes/javascriptMode.ts
```diff
@@ -133,7 +133,8 @@ export function getJavascriptMode(
       const offset = currentTextDocument.offsetAt(position);
       const completions = jsLanguageService.getCompletionsAtPosition(
         FILE_NAME,
-        offset
+        offset,
+        { includeExternalModuleExports: false }
```
vazexqi: We now need to explicitly fill in the arguments (can't call a method that expects x arguments with y arguments, JavaScript style). See TS2554 for the error.

packages/salesforcedx-vscode-apex/test/languageServer.test.ts
```diff
@@ -16,7 +16,10 @@ describe('Apex Language Server Client', () => {
     let originalPlatform: PropertyDescriptor;
 
     before(() => {
-      originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform');
+      originalPlatform = Object.getOwnPropertyDescriptor(
```
vazexqi: I think this is a combination of https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#weak-type-detection and because `PropertyDescriptor` has all optional fields and could return undefined.

vazexqi: I could also use the non-null assertion operator here instead of casting. See https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator

I think I will switch to that since the intent is clearer. Using then non-null assertion operator seems to be more intention-revealing here.

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -99,7 +99,6 @@ describe('Generate faux classes for SObjects', function() {
   });
 
   it('Should emit an error event on failure', async () => {
-    let result = '';
```
vazexqi: The new typescript + tslint analyzer is not very aggressive in pruning unread variables.

If you only write to it but never read from it, it's considered an unused variable.

vazexqi:
@DatGuyJonathan - Thanks. I'll take a look at the unknown dependency. It must be because this is a clean installation of VS code (vs what I have locally). And we are running into https://github.com/Microsoft/vscode-extension-vscode/issues/74

vazexqi:
Merging this in first and then fixing the `salesforcedx-vscode-lwc: Extension salesforce.salesforcedx-vscode-lwc failed to activate. Reason: unknown dependency dbaeumer.vscode-eslint` issue.

PR #241, opened 2018-01-04
Add LWC Extension

vazexqi:
### What does this PR do?

First version of our extension for LWC.

This moves the work that we have been doing in our internal branch to the public GitHub.

### What issues does this PR fix or reference?

@W-4473315@

PR #244, opened 2018-01-05
[WIP] Add script to enable testing with external dependencies

vazexqi:
### What does this PR do?

In #242 we discovered that due to #241 the tests were failing. This is because we were not able to run the tests due to the dependency on an external .vsix package. This PR follows the steps at https://github.com/Microsoft/vscode-extension-vscode/issues/74 to resolve it.


scripts/install-vsix-dependencies.js
```diff
@@ -0,0 +1,67 @@
...
+
+const version = process.env.CODE_VERSION || '*';
+
+const downloadPlatform =
+  process.platform === 'darwin'
+    ? 'darwin'
+    : process.platform === 'win32' ? 'win32-archive' : 'linux-x64';
+
+const windowsExecutable = path.join(testRunFolderAbsolute, 'Code');
+const darwinExecutable = path.join(
```
vazexqi: This is basically from https://github.com/Microsoft/vscode-extension-vscode/blob/master/bin/test but modified for OS X since we you cannot just invoke Electron. You need to find the path to the actual code binary to pass the `install-extensions` argument to it.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -48,8 +48,9 @@
     "clean":
       "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
+    "pretest": "node ../../scripts/download-vscode-for-system-tests",
     "test":
-      "cross-env CODE_TESTS_WORKSPACE='../system-tests/assets/sfdx-simple' node ./node_modules/vscode/bin/test"
+      "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && cross-env CODE_TESTS_WORKSPACE='../system-tests/assets/sfdx-simple' CODE_TESTS_PATH='./' node ./node_modules/vscode/bin/test"
```
vazexqi: There is a dependency from LWC to the external dbaeumer.vscode-eslint. Without having this dependency, the entire test will fail.

.gitignore
```diff
@@ -87,3 +87,6 @@ packages/system-tests/project_*
 
 # Node 8 files (not yet used)
 package-lock.json
+
+# External extension dependencies
```
vazexqi: We will install this at the top-level so we should definitely ignore it from being committed. We need it to be at the same level as the other extensions.

The downside of this is that now we are also loading all the other extensions from the /packages directory so there will be some additional overhead.

packages/salesforcedx-vscode-lwc/test/extension.test.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
+
+describe('LWC Extension Tests', () => {
```
vazexqi: This is just a sample test for now to appease the code coverage collection routine.

vazexqi:
The system tests are stuck on travis Linux builds.
They are passing on Windows and also the unit tests (which the LWC tests are), are passing on Mac as well.
I am merging this in because this seems to be more of a Travis Linux issue.

cc: @Sweetman

PR #247, opened 2018-01-05
Signed jar for rebase and goto for class declation

chowwinston:
### What does this PR do?
Allows goto definition of a class's parent and super interfaces from the class declaration
@W-4317370@


chowwinston:
Signed the jar but just waiting for @GaryFNC to do another round of QA after the rebase.

PR #248, opened 2018-01-08
Start & stop replay debugger

DatGuyJonathan:
### What does this PR do?
This is the Replay Debugger extension which can launch a session from a config, stop on the first line of a debug log, continue to finish the stopped thread, and stop the session. A sample config:

```
{
  "name": "Launch Apex Replay Debugger",
  "type": "apex-replay",
  "request": "launch",
  "logFile": "${workspaceFolder}/${command:AskForLogFileName}",
  "stopOnEntry": true,
  "traceLog": true
}
```

`LogFile` takes a command to trigger a prompt for the user to enter the path to a debug log file. They can also just hardcode the path in there. I allow the hardcoded option because it was a bit cumbersome for me to test having to copy-paste every time I start a session. Feel free to pull the branch and play around with it. Let me know if we want to hide the `LogFile` parameter so it always prompts; although I'm not sure yet where we can trigger the command so it's not too late.

`StopOnEntry` decides whether we stop on the first line of debug log file or on the first breakpoint hit. Currently, only the former works. When we implement breakpoints, the idea is to start a session and go through the debug log until it finds a statement that matches an enabled breakpoint.

`TraceLog` decides whether to output each debug log line as they are parsed so the user knows how the log is affecting the debugger. It's easy to implement, but it would fit better as part of the work for hitting a breakpoint so we can actually see a partial log file in the debug console. I'm open to renaming this parameter because we may use something like `Trace` in the future that enables internal logging so we can see what the debugger is doing.

replaydebugger

@W-4549178@


packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -0,0 +1,124 @@
...
+  "contributes": {
+    "breakpoints": [
+      {
+        "language": "apex"
+      }
+    ],
+    "debuggers": [
+      {
+        "type": "apex-replay",
+        "label": "Apex Replay Debugger",
```
DatGuyJonathan: This label is shown in the dropdown of debuggers when you try to create a new launch config.

image

packages/salesforcedx-vscode-replay-debugger/package.nls.json
```diff
@@ -0,0 +1,11 @@
...
+  "launch_snippet_label_text": "Apex Replay Debugger: Launch",
+  "launch_snippet_description_text":
+    "A new configuration for launching Apex Replay Debugger",
+  "launch_snippet_name": "Launch Apex Replay Debugger",
+  "logfile_text": "Path to an Apex Debug Log file in the workspace folder.",
+  "stop_on_entry_text":
+    "Stop the debugger after launching. If false, the debugger will run until it hits a breakpoint.",
+  "trace_log_text":
+    "Output debug log lines as they are processed by the debugger."
+}
```
DatGuyJonathan: These messages are used in the launch config (see .gif in the PR description).

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export interface LaunchRequestArguments
+  extends DebugProtocol.LaunchRequestArguments {
+  logFile: string;
+  stopOnEntry: boolean | true;
+  traceLog: boolean | true;
+}
+
+export class ApexReplayDebug extends DebugSession {
+  public static THREAD_ID = 1;
```
DatGuyJonathan: Unlike the Apex Interactive Debugger, we don't allow debugging multiple threads, so just hardcode this ID.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -0,0 +1,152 @@
...
+    super();
+    this.setDebuggerLinesStartAt1(true);
+    this.setDebuggerPathFormat('uri');
+  }
+
+  public initializeRequest(
+    response: DebugProtocol.InitializeResponse,
+    args: DebugProtocol.InitializeRequestArguments
+  ): void {
+    response.body = {
```
DatGuyJonathan: Disable optional features of the debugger for now. We'll enable them as we add support.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -0,0 +1,152 @@
...
+    args: LaunchRequestArguments
+  ): void {
+    this.logFile = new LogFile(args);
+    if (!this.logFile.hasLogLines()) {
+      response.success = false;
+      response.message = nls.localize('no_log_file_text');
+      this.sendResponse(response);
+      return;
+    }
+    if (args.stopOnEntry) {
```
DatGuyJonathan: Here's the part where it decides to stop on the first line of debug log or continue until the first breakpoint.

packages/salesforcedx-vscode-replay-debugger/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  config_name_text: 'Launch Apex Replay Debugger',
+  config_request_type_text: 'launch',
+  config_type_text: 'apex-replay',
+  session_started_text: 'Apex Replay Debugger session started for log file %s.',
+  session_terminated_text: 'Apex Replay Debugger session terminated.',
+  no_log_file_text:
+    'The log file is either missing or does not have any log lines in it.',
+  prompt_for_log_file_text:
+    'Please enter the path of a debug log file in the workspace folder'
+};
```
DatGuyJonathan: Some messages for asking the user for a debug log file, starting a session, stopping a session, and error when debug log file is invalid.

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -0,0 +1,124 @@
...
+    "nyc": "^11.0.2",
+    "cross-env": "5.0.4",
+    "mocha-junit-reporter": "^1.13.0",
+    "mocha-multi-reporters": "^1.1.4",
+    "typescript": "2.6.2",
+    "vscode": "1.1.10",
+    "vscode-debugadapter-testsupport": "1.25.0"
+  },
+  "extensionDependencies": ["salesforce.salesforcedx-vscode-core"],
+  "scripts": {
```
DatGuyJonathan: No publish script.

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -0,0 +1,124 @@
...
+    "@types/node": "^6.0.40",
+    "@types/sinon": "^2.3.2",
+    "chai": "^4.0.2",
+    "mocha": "3.2.0",
+    "sinon": "^2.3.6",
+    "nyc": "^11.0.2",
+    "cross-env": "5.0.4",
+    "mocha-junit-reporter": "^1.13.0",
+    "mocha-multi-reporters": "^1.1.4",
+    "typescript": "2.6.2",
```
DatGuyJonathan: Use the same typescript version as other packages.

packages/salesforcedx-vscode-replay-debugger/package.nls.json
```diff
@@ -0,0 +1,11 @@
+{
+  "launch_snippet_label_text": "Apex Replay Debugger: Launch",
+  "launch_snippet_description_text":
+    "A new configuration for launching Apex Replay Debugger",
+  "launch_snippet_name": "Launch Apex Replay Debugger",
+  "logfile_text": "Path to an Apex Debug Log file in the workspace folder.",
```
ruthemmanuelle: Do we want to say what the file extension should be?

packages/salesforcedx-vscode-replay-debugger/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+ * If ommitted, we will assume _message.
+ */
+export const messages = {
+  config_name_text: 'Launch Apex Replay Debugger',
+  config_request_type_text: 'launch',
+  config_type_text: 'apex-replay',
+  session_started_text: 'Apex Replay Debugger session started for log file %s.',
+  session_terminated_text: 'Apex Replay Debugger session terminated.',
+  no_log_file_text:
+    'The log file is either missing or does not have any log lines in it.',
```
ruthemmanuelle: "is" should be between "either" and "missing" (because the file either a. is missing or b. doesn't have log lines).

packages/salesforcedx-vscode-replay-debugger/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  prompt_for_log_file_text:
+    'Please enter the path of a debug log file in the workspace folder'
```
ruthemmanuelle: We use "please" only when asking the user to do something inconvenient. You can just start this with "Enter."

packages/salesforcedx-vscode-replay-debugger/package.json
Sweetman: Should we be more specific about the debuggers now and say "Apex Interactive Debugger"?

ruthemmanuelle: The official name of the existing debugger, even though it's been called various other things in blog posts and presentations, is "Apex Debugger." We'd asked about changing the name of the existing debugger, to prevent confusion with the Replay Debugger, but thus far that request hasn't gained traction.

.gitignore
```diff
@@ -8,7 +8,6 @@ out/
 
 # Logs
 logs
-*.log
```
vazexqi: @DatGuyJonathan - Could we be more specific and just ignore the *.log from certain folders? You can use nested .gitignore files. See https://stackoverflow.com/questions/3305869/are-multiple-gitignores-frowned-on

I worry that if we just whitelist *.log that someone will accidentally add and commit a *.log file.

.vscode/launch.json
```diff
@@ -307,12 +336,62 @@
...
       "preLaunchTask": "Compile"
+    },
+    {
+      "name": "Launch Replay Debugger adapter",
+      "type": "node",
+      "request": "launch",
+      "cwd": "${workspaceRoot}",
+      "program": "${workspaceRoot}/packages/salesforcedx-vscode-replay-debugger/out/src/adapter/apexReplayDebug.js",
+      "args": [
+        "--server=4711"
```
vazexqi: Is there a reason why we are choosing the same port, i.e, 4711 for both the Apex Debugger and the Apex Replay Debugger? This makes it harder to be able to debug both of them at the same time, right?

packages/salesforcedx-vscode-replay-debugger/package.nls.json
vazexqi: (*.log) would be good to include.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
vazexqi: I think this is true (or at least for all intents and purposes it might not matter much).

I'm curious when we have a situation like https://help.salesforce.com/articleView?id=code_debug_log_details.htm&type=5 (more description on the different code_units) are also there

```
EXECUTION_STARTED
CODE_UNIT_STARTED|[EXTERNAL]execute_anonymous_apex
CODE_UNIT_STARTED|[EXTERNAL]MyTrigger on Account trigger event BeforeInsert for [new]
CODE_UNIT_FINISHED <-- The trigger ends
CODE_UNIT_FINISHED <-- The executeAnonymous ends
EXECUTION_FINISHED
```

This seems to be under two different CODE_UNITS. Are we just going to lump them into one thread? AFAIC, lumping them under one thread should not cause any issues since the log has already serialized them linearly so we can just step through them (there is no weird interleaving).

packages/salesforcedx-vscode-replay-debugger/src/core/logFileUtil.ts
```diff
@@ -0,0 +1,36 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as fs from 'fs';
+import { EOL } from 'os';
+import { DebugLogEvent, LogEntry, NoOp } from '../events';
+
+export class LogFileUtil {
```
vazexqi: Something to keep in mind. Eventually, we want to be able to operate on logs that come back as a stream. So, let's not hardcode and assume that we have the entire contents in memory at once.

packages/salesforcedx-vscode-replay-debugger/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es6",
+    "lib": ["dom", "es2016"],
```
vazexqi: Do you need to use dom? I'm not sure if you need it since you are not using anything from the web UI.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+import { nls } from './messages';
+
+export function activate(context: vscode.ExtensionContext) {
+  console.log('Apex Replay Debugger Extension Activated');
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      'extension.replay-debugger.getLogFileName',
+      config => {
+        return vscode.window.showInputBox({
+          placeHolder: nls.localize('prompt_for_log_file_text')
```
Sweetman: @DatGuyJonathan I think it would be more convenient for the user if we prompt the user with showOpenDialog so that they interact with the native file explorer to find the logs. Would save the user some steps in getting the full file path since this doesn't provide autocomplete and if they do have it in their clipboard, it would be better to put it in the launch.json. Thoughts?

packages/salesforcedx-vscode-replay-debugger/test/unit/adapter/debugConfigurationProvider.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    expect(configs).to.deep.equal([expectedConfig]);
+  });
+
+  it('Should fill in empty attributes in the config', () => {
+    const expectedConfig = {
+      name: nls.localize('config_name_text'),
+      type: nls.localize('config_type_text'),
+      request: nls.localize('config_request_type_text'),
+      logFile: '${workspaceFolder}/${command:AskForLogFileName}',
```
vazexqi: I think being able to specify one is fine. We would need to do so for testing purposes.

vazexqi:
As a general design comment, take a look into the State Pattern (https://sourcemaking.com/design_patterns/state) and see how to design it this way. For each state, we should enumerate the proper actions/events that it can respond to. This will allow us to be more fault-tolerant. This is more inline with how we do hardware design where every possible transaction is mapped out and analyzed properly.

We need this in the replay debugger (not so much in the interactive one) because there could be cases of log truncation (even when we increase the log file or switch to streaming).

So, having a more robust and fault-tolerant design is better.

vazexqi:
> TraceLog decides whether to output each debug log line as they are parsed so the user knows how the log is affecting the debugger. It's easy to implement, but it would fit better as part of the work for hitting a breakpoint so we can actually see a partial log file in the debug console. I'm open to renaming this parameter because we may use something like Trace in the future that enables internal logging so we can see what the debugger is doing. 

If we are going to implement a Trace in the future, couldn't one of the levels be to just output what TraceLog currently does? Do we actually need two top-level entries?

I think the functionality of TraceLog could be subsumed as part of Trace, as part of some configuration array.

packages/salesforcedx-vscode-replay-debugger/test/integration/config/data/apexTest.log
```diff
@@ -0,0 +1,481 @@
+41.0 APEX_CODE,FINEST;APEX_PROFILING,FINEST;CALLOUT,FINEST;DB,FINEST;SYSTEM,FINE;VALIDATION,INFO;VISUALFORCE,FINER;WAVE,FINEST;WORKFLOW,FINER
```
vazexqi: @DatGuyJonathan - How did you generate this file? Is this really the trace from a single transaction? Or did you copy-and-paste several examples together?

packages/salesforcedx-vscode-replay-debugger/package.nls.json
DatGuyJonathan: I can add that to suggest it's a .log file, but the code is not enforcing that extension because the user can rename their log file to anything. Is that okay?

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
DatGuyJonathan: I haven't seen any interleaving, so I was going to treat CODE_UNIT_STARTED/FINISHED like entering and leaving a stackframe.

packages/salesforcedx-vscode-replay-debugger/tsconfig.json
DatGuyJonathan: I can remove; it's in all other packages.

packages/salesforcedx-vscode-replay-debugger/test/integration/config/data/apexTest.log
DatGuyJonathan: This is a real debug log from running a test class.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
DatGuyJonathan: It's an interesting idea. We're going to have a command to download debug logs, but it'll default to the hidden .sfdx folder. Could you walk me through some a flow that involves selecting a hidden file with showOpenDialog?

vazexqi: @DatGuyJonathan I had the chance to play with this quickly:

```
  context.subscriptions.push(
    vscode.commands.registerCommand(
      'extension.replay-debugger.getLogFileName',
      config => {
        return vscode.window.showOpenDialog({
          canSelectFiles: true,
          canSelectFolders: false,
          canSelectMany: false,
          openLabel: nls.localize('prompt_for_log_file_text') // This is a bit ugly in the GIF
        });
      }
    )
  );
```

opendialog

The API is at https://code.visualstudio.com/docs/extensionAPI/vscode-api#OpenDialogOptions

I think we can use the OpenDialog and default the folder to the .sfdx/logs folder. The OpenDialog is a bit more general and can allow the user to select a different log.

The downside is that the user might select a log file that doesn't have any source that belongs to the current workspace.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -7,16 +7,17 @@
...
 
 export function activate(context: vscode.ExtensionContext) {
   console.log('Apex Replay Debugger Extension Activated');
   context.subscriptions.push(
     vscode.commands.registerCommand(
       'extension.replay-debugger.getLogFileName',
       config => {
-        return vscode.window.showInputBox({
-          placeHolder: nls.localize('prompt_for_log_file_text')
+        return vscode.window.showOpenDialog({
```
vazexqi: We probably can default the defaultUri paramter to the root workspace.

PR #250, opened 2018-01-10
Update docs for v41.16.0

ruthemmanuelle:
### What does this PR do?
* salesforcedx-vscode-apex README:  
    * Adds classes and interfaces to list of Apex items whose definitions you can visit (https://github.com/forcedotcom/salesforcedx-vscode/pull/247)  
    * Alphabetizes items in lists of items whose definitions you can visit
* salesforcedx-vscode-apex-debugger README: Adds mention that call stack shows paused-on exception (https://github.com/forcedotcom/salesforcedx-vscode/pull/240)

@W-4605866@

### DIFF
--- a/packages/salesforcedx-vscode-apex/README.md
+++ b/packages/salesforcedx-vscode-apex/README.md
@@ -20,18 +20,18 @@ To see code-completion suggestions, press Ctrl+space when you’re working in a 
 ## View or Jump to Definitions
 You can preview, view, or go to definitions of:  
 * Apex  
+  * classes (from within extending classes)  
+  * constructors  
+  * interfaces (from within implementing classes)  
   * methods  
   * properties  
-  * constructors  
-  * local variables  
-  * class variables  
+  * variables (local and class variables)  
 * standard objects  
+  * fields (standard and custom fields)
   * object definitions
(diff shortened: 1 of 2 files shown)


ruthemmanuelle:
@GaryFNC - Let me know if we need more caveats for classes and interfaces in the salesforcedx-vscode-apex list. I wasn't sure if we should list that Go To Definition works but Preview and View don't, so I didn't add that caveat.

packages/salesforcedx-vscode-apex/README.md
```diff
@@ -20,18 +20,18 @@ To see code-completion suggestions, press Ctrl+space when you’re working in a
 ## View or Jump to Definitions
 You can preview, view, or go to definitions of:  
 * Apex  
+  * classes (from within extending classes)
```
GaryFNC: It only works for the usage site in "extends x" or "implements xx, yy".  It doesn't work (yet) for Parent foo; where you click on Parent.   That work is on future work items.   Not quite sure how to say this in the docs.   We decided it was better to release the functionality that we have rather than wait until all possible sites are working

vazexqi:
LGTM 👍

PR #255, opened 2018-01-12
Simplify checks for activating LWC

vazexqi:
### What does this PR do?

We moved the functionality into the lwc-language-server module so we can just reference it and do the checks from there.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/index.ts
+++ b/packages/salesforcedx-vscode-lwc/src/index.ts
@@ -5,7 +5,6 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -66,54 +77,3 @@ export async function activate(context: ExtensionContext) {
   // client can be deactivated on extension deactivation
   context.subscriptions.push(disposable);
 }
-
-function isLWCProject(): boolean {
```
vazexqi: These parts have been moved to the lwc-language-server so there is no need to duplicate the logic here.

vazexqi:
@esalman-sfdc - This is now merged into develop. Sorry for the confusion earlier.

PR #256, opened 2018-01-13
Make publishing compatible with Node 8

vazexqi:
### What does this PR do?

This changes our publishing script so that it works with Node 8. This helps us keep up to date with the Node ecosystem. Node 8 is now at 8.9.x and most bugs should have been ironed out.

I've only tested this with 8.9.4 but I'm guessing anything above that should work (and anything close to 8.9.0 should work).

Please try it out with the version of Node 8 that you have installed.

Note that we are still emitting es6 style JavaScript so this should not affect any code.

@W-4395117@


package.json
```diff
@@ -12,28 +12,23 @@
     "vsce": "1.29.0"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap",
-    "bootstrap": "lerna bootstrap",
+    "postinstall": "lerna bootstrap -- --no-package-lock",
```
vazexqi: With Node 8 (or Yarn), there is a lock file that is created after `npm install`. The problem with this lock file is that it won't contain packages that are symlinked locally by Lerna. Thus, when we run `npm --prune production` it will also use the lock file and prune the symlinked @salesforce packages even though they are required. Thus, we tell Node 8 to not produce lock files during installation.

However, running other commands that affect the node_modules directory, e.g., `npm prune` will also affect the lock file. Since, we don't commit the lock file, this is fine.

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -37,7 +37,6 @@
     "vscode-debugadapter-testsupport": "1.24.0"
   },
   "scripts": {
-    "vscode:package": "npm prune --production",
```
vazexqi: With Node 8 and the fixed npm, we no longer need to do the double pruning.

package.json
```diff
@@ -12,28 +12,23 @@
...
     "test:unit": "lerna run test:unit --concurrency 1 --stream",
     "test:integration": "lerna run test:integration --concurrency 1 --stream",
-    "test:without-system-tests":
-      "lerna run test --ignore system-tests --concurrency 1 --stream",
-    "test:system-tests":
-      "lerna run test --scope system-tests --concurrency 1 --stream",
-    "coverage:system-tests":
-      "lerna run coverage --scope system-tests --concurrency 1 --stream",
-    "vscode:package":
-      "lerna exec -- shx rm -rf node_modules/@types && lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1",
```
vazexqi: With Node 8 and the fixed npm, we no longer need to do the double pruning.

.appveyor.yml
```diff
@@ -16,7 +16,7 @@ environment:
   SFDX_DOMAIN_RETRY: 300
 
 install:
-  - ps: Install-Product node 7.4.0 x64
+  - ps: Install-Product node 8 x64
```
vazexqi: AppVeyor only seems to have 

`Installing node 8.9.3 (x64)..` for now

vazexqi:
Using zsh's time command, I also noticed that using Node 8 is a bit more CPU intensive and takes a few seconds longer.

### Node 8.9.4
`time ( npm install; npm run compile; npm run vscode:package; ) `
**real 196.94s**
user	498.43s
sys	225.06s

### Node 7.9.0
`time ( npm install; npm run compile; npm run vscode:package; ) `
**real 140.48s**
user	193.97s
sys	70.44s

I do notice that on Node 8, when you do `npm install` that it reformats package.json using lerna style. This is because it runs `lerna bootstrap -- --no-package-lock`. So, I've also changed that to reformat the package.json's after the command.

package.json
guw: @vazexqi should we disable the file completely?

Like this:
`echo 'package-lock=false' >> .npmrc`

vazexqi:
There is an on-going issue with Travis on

```
system-tests: Xlib:  extension "RANDR" missing on display ":99.0".
system-tests: Xlib:  extension "RANDR" missing on display ":99.0".
system-tests: [main 5:25:43 PM] Running extension tests from the command line is currently only supported if no other instance of Code is running.
```

It seems to think another instance of Code is running. This seems specific to Travis (or, Linux). We would need to try to repro this locally to see if anything in Code has changed w/r to Linux. Otherwise, the PR is passing on AppVeyor.

package.json
vazexqi: I fee like doing it explicitly through Lerna's `--no-package-lock` is a bit more explicit, for now. I might change my mind about this later if there package-lock.json annoy me.

PR #258, opened 2018-01-15
Fix Go To Definition for class/interface usage sites for inner classes

GaryFNC:
Checkin in the signed apex-jorje-lsp.jar for the type usage bug
fixes made in Apex LS

Added URI information on the source file being compiled to ensure that
the location can be found for inner classes and interfaces referenced
from the same file

Changed the matching algorithm on interfaces to handle type refs that
only give the stated name of an interface, not the full name (ex.
OuterClass.Interface1 while the type ref name is just Interface1).

Also this matching test must be case insensitive like Apex (user could type
interface1).

@W-4593425, W-4598183@


vazexqi:
@GaryFNC 

```
@W-4593425@
@W-4598183@
``` doesn't work.

It needs to be `@W-4593425, W-4598183@`

PR #261, opened 2018-01-17
Upgrade LWC to use 14.0.1

vazexqi:
### What does this PR do?

Bumps the dependency for LWC.


scripts/install-vsix-dependencies.js
```diff
@@ -28,27 +28,9 @@ const darwinExecutable = path.join(
 const linuxExecutable = path.join(
   testRunFolderAbsolute,
   'VSCode-linux-x64',
+  'bin',
   'code'
```
vazexqi: This is the actual location the executable that can understand `code --install-extension....`

PR #266, opened 2018-01-18
Update the way we publish

vazexqi:
### What does this PR do?

Node 8 makes it more difficult to `require` global libraries. So the previous way of trying to use the globally installed shelljs library doesn't work. See https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders for the full resolution order.

We could make the old way work but that would require us to either install it in the $HOME directory, or mess with NODE_PATH. Neither sounds desirable.

So, instead the main changes are:

1. You do a `npm install` at the top-level. This brings in the dependencies.
1. When we do a `git clean` we ignore the top-level node_module which only contains things like shelljs, lerna, etc. These are never bundled so they don't affect the generated vsix.

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -31,8 +31,7 @@ It is possible to run each step manually as illustrated below.
 
 ### Prerequisite
 
-* Lerna is properly installed (`npm install -g lerna@2.0.0`).
-* shelljs is properly installed (`npm install -g shelljs`).
+* Lerna is properly installed (`npm install -g lerna@2.4.0`).
 * You have a clean workspace. You can use `git clean -xfd` to get to a pristine
   stage.
 * All tests have been run prior to publishing. We don't run the tests during the
@@ -43,9 +42,9 @@ It is possible to run each step manually as illustrated below.
 
 1. `npm install` to install all the dependencies and to symlink interdependent local modules.
 1. `npm run compile` to compile all the TypeScript files.
-1. `lerna publish ...` will increment the version in the individual package.json
-   to prepare for publication. **This also commits the changes to git and adds a
-   tag.**
+1. `lerna publish ...` (see scripts/publish.js for the full command) will
+   increment the version in the individual package.json to prepare for
+   publication. **This also commits the changes to git and adds a tag.**
 1. `npm run vscode:package` packages _each_ extension as a .vsix.
 
 **At this stage, it is possible to share the .vsix directly for manual installation.**
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -25,6 +25,11 @@ The files under scripts use [shelljs/shx](https://github.com/shelljs/shx) and
 [shelljs/shelljs](https://github.com/shelljs/shelljs) to write scripts in a
 portable manner across platforms.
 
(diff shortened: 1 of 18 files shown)


docs/publishing.md
```diff
@@ -31,8 +31,7 @@ It is possible to run each step manually as illustrated below.
 
 ### Prerequisite
 
-* Lerna is properly installed (`npm install -g lerna@2.0.0`).
-* shelljs is properly installed (`npm install -g shelljs`).
```
vazexqi: You cannot require shelljs globally using 

```
const shell = require('shelljs');
```

so we remove this line.

scripts/publish.js
```diff
@@ -99,9 +99,6 @@ shell.exec(
   `lerna publish --force-publish --exact --repo-version ${nextVersion} --yes --skip-npm`
 );
 
-// Reformat with prettier since lerna changes the formatting in package.json
-shell.exec('./scripts/reformat-with-prettier.js');
```
vazexqi: We tie this now in the top-level package.json to the end of the vscode:package.

@DatGuyJonathan

scripts/publish.js
```diff
@@ -84,7 +84,7 @@ if (checkTags.includes(nextVersion)) {
 }
 
 // Real-clean
-shell.exec('git clean -xfd');
+shell.exec('git clean -xfd -e node_modules');
```
vazexqi: So when we clean we now leave the top-level node_modules (with shelljs) alone.

docs/publishing.md
```diff
@@ -25,16 +25,18 @@ The files under scripts use [shelljs/shx](https://github.com/shelljs/shx) and
 [shelljs/shelljs](https://github.com/shelljs/shelljs) to write scripts in a
 portable manner across platforms.
 
+1. `git checkout -t origin release/vxx.yy.zz`
```
vazexqi: I think Winston never updated this doc with the simplified steps.
But here it is.

scripts/publish.js
DatGuyJonathan: My top level package.json also got formatted during packaging, so I think the reformat script needs to be edited to include top level package.json.

docs/publishing.md
```diff
@@ -25,16 +25,18 @@ The files under scripts use [shelljs/shx](https://github.com/shelljs/shx) and
...
 It is possible to run each step manually as illustrated below.
 
 ## Packaging as .vsix
 
 ### Prerequisite
 
-* Lerna is properly installed (`npm install -g lerna@2.0.0`).
-* shelljs is properly installed (`npm install -g shelljs`).
-* You have a clean workspace. You can use `git clean -xfd` to get to a pristine
-  stage.
```
lcampos: Shouldn't we keep this but modify it to leave top-level node_modules (git clean -xfd -e node_modules) ?

scripts/publish.js
vazexqi: Interesting. Sure I can include that.

docs/publishing.md
vazexqi: The publish scripts actually runs `git clean -xfd -e node_modules` for you. That's why I avoided doing it here.

PR #267, opened 2018-01-19
Only activate for SFDX workspace

vazexqi:
### What does this PR do?

The LWC extension is activated for both local and external users. This limits it so that it only activates for SFDX users.


vazexqi:
I'll combine this into #270.

PR #268, opened 2018-01-19
Replay Debugger stops on line breakpoints

DatGuyJonathan:
### What does this PR do?
hitbreakpoint

- If the launch config has `stopOnEntry = false`, the debugger launches and tries to run to the next breakpoint.
  - If there is no breakpoint in the code path, the debugger will keep running until the end of the debugged thread.
- If the launch config has `stopOnEntry = true`, the debugger launches and stops in the debug log.
  - When the user clicks continue, the debugger tries to run to the next breakpoint or end of the debugged thread.
- If the launch config has `trace`, the debugger will output to the debug console depending on what the user wants to trace. For example, if the value contains `logfile`, we output each debug log line as they are processed. Other values are `all`, `protocol`, `launch`, and `breakpoints`.

**There are some scenarios that won't work yet because of known bugs in the debug log (e.g. constructors, triggers, apex tests, visualforce invocations)**

@W-4549440@


packages/salesforcedx-vscode-replay-debugger/package.nls.json
```diff
@@ -7,5 +7,6 @@
     "Path to an Apex Debug Log file (*.log) in the workspace folder.",
   "stop_on_entry_text":
     "Stop the debugger after launching. If false, the debugger will run until it hits a breakpoint.",
-  "trace_text": "Output debug log lines as they are processed by the debugger."
+  "trace_text":
+    "Produce diagnostic output. Instead of setting this to true you can list one or more selectors separated with commas. The 'all' selector enables very detailed output. Other selectors are 'protocol', 'logfile', 'launch', 'breakpoints'."
```
DatGuyJonathan: This text is similar to the one in Apex Interactive Debugger extension.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -67,27 +86,55 @@ export class ApexReplayDebug extends DebugSession {
...
     response.success = true;
     this.sendResponse(response);
+
+    if (args.stopOnEntry) {
+      // Stop in the debug log
+      this.logContext.updateFrames(this.getHandlerForDebugConsole());
+      this.sendEvent(new StoppedEvent('entry', ApexReplayDebug.THREAD_ID));
+    } else {
+      // Set breakpoints first, then try to continue to the next breakpoint
+      setTimeout(() => {
```
DatGuyJonathan: After `launchRequest`, we want VS Code to send breakpoints to the adapter first before we force a continue. This timeout is to delay the forced continue.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -101,9 +148,7 @@ export class ApexReplayDebug extends DebugSession {
 
   public threadsRequest(response: DebugProtocol.ThreadsResponse): void {
     response.body = {
-      threads: [
-        new Thread(ApexReplayDebug.THREAD_ID, this.logContext.getLogFileName())
-      ]
+      threads: [new Thread(ApexReplayDebug.THREAD_ID, '')]
```
DatGuyJonathan: Thread name is only useful if the debugger supports multiple debugged threads. If there's only one, VS Code doesn't show the name.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -122,9 +172,28 @@ export class ApexReplayDebug extends DebugSession {
     response: DebugProtocol.ContinueResponse,
     args: DebugProtocol.ContinueArguments
   ): void {
-    this.sendEvent(new ThreadEvent('exited', ApexReplayDebug.THREAD_ID));
     response.success = true;
     this.sendResponse(response);
+    while (this.logContext.hasLogLines()) {
```
DatGuyJonathan: Parse log lines until we find a STATEMENT_EXECUTE whose location matches a breakpoint.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
```diff
@@ -65,12 +135,32 @@ export class LogContext {
...
     }
+    if (logLine.startsWith(EVENT_EXECUTE_ANONYMOUS)) {
+      this.execAnonMapping.set(
+        this.execAnonMapping.size + 1,
+        this.logLinePosition + 1
+      );
+    }
     const fields = logLine.split('|');
     if (fields.length >= 3) {
       switch (fields[1]) {
```
DatGuyJonathan: Some new events that add a frame, remove a frame, or update the line value of a frame.

packages/salesforcedx-vscode-replay-debugger/package.nls.json
Sweetman: Should the message should mention that setting the trace to `true` defaults to all?

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
Sweetman: I reread the debug api doc and this approach seems kind of weird. Maybe I'm misreading the docs but it looks like this would go into configurationDoneRequest? Wouldn't that send all the breakpoint info to the adapter first and then we can send the continueRequest once the breakpoint configuration is done?
https://code.visualstudio.com/docs/extensionAPI/api-debugging

DatGuyJonathan: Thanks, I forgot about `configurationDoneRequest`. Will update.

packages/salesforcedx-vscode-replay-debugger/package.nls.json
DatGuyJonathan: I was thinking of just removing the boolean option so it only takes a string. `all` is basically true; nothing specified is basically false. @guw do you remember why the interactive debugger extension supports boolean and string?

DatGuyJonathan: From an old PR, we have these two types because "This is how the VS Code node debugger does it so I would prefer re-using existing concepts from VS Code for consistency."

For comparison purpose, VS Code's node debugger uses this description:
`Produce diagnostic output. Instead of setting this to true you can list one or more selectors separated with commas. The 'verbose' selector enables very detailled output.`

@ruthemmanuelle , do you think our text needs to be updated?

ruthemmanuelle: Yeah, this is kind of confusing. How about something like this?

"Produce diagnostic output. You can set this to true, false, or one or more selectors separated with commas. Selectors include 'all' (very detailed output; the equivalent of true), 'protocol', 'logfile', 'launch', 'breakpoints'."

packages/salesforcedx-vscode-core/test/commands/commands.test.ts
```diff
@@ -53,7 +53,7 @@ describe('Command Utilities', () => {
         new class implements ParametersGatherer<{}> {
           public async gather(): Promise<
             CancelResponse | ContinueResponse<{}>
-          > {
+            > {
```
vazexqi: Did something change with the prettier algorithm here? It looks to be mostly whitespace issues.

DatGuyJonathan: I accidentally prettier disabled the last time I made changes to this file, so I enabled it and fixed the files with bad formatting.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -26,21 +29,37 @@ import {
...
 
-export class ApexReplayDebug extends DebugSession {
+export class ApexReplayDebug extends LoggingDebugSession {
   public static THREAD_ID = 1;
   protected logContext: LogContext;
+  protected trace: string[] = [];
+  protected traceAll = false;
   private breakpointUtil: BreakpointUtil;
   private initializedResponse: DebugProtocol.InitializeResponse;
+  protected breakpoints: Map<string, number[]> = new Map();
```
vazexqi: Do we have any way to let the user know if the source and the debug logs have deviated?
With the Apex interactive debugger, we could detect if the user is trying to save to the server and terminate the debugger session.
With the Apex replay debugger, it's possible that an accidental <new line> could shift all the breakpoints such that nothing hits.
I don't have a solution off the top of my head but I'm wondering if there is anything that we can/should do here.

packages/salesforcedx-vscode-replay-debugger/src/states/frameExitState.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { LogContext } from '../core/logContext';
+import { DebugLogState } from './debugLogState';
+
+export class FrameExitState implements DebugLogState {
```
vazexqi: What happens if you don't see a corresponding METHOD_ENTRY and METHOD_EXIT (due to log truncation)?

METHOD_ENTRY method1()
METHOD_ENTRY method2()
...truncation... // exited method2
METHOD_EXIT method1() <- does this one know to pop off both method1 and method2

packages/salesforcedx-vscode-replay-debugger/test/unit/core/logContext.test.ts
```diff
@@ -109,41 +135,135 @@ describe('LogContext', () => {
...
 
-    context.updateFrames();
+    context.updateFrames(debugConsoleHandler);
 
     expect(context.getLogLinePosition()).to.equal(1);
+    expect(context.hasState()).to.be.true;
+    expect(context.getFrames()).to.be.empty;
   });
 
   describe('Log event parser', () => {
```
vazexqi: @DatGuyJonathan - Is it possible for a user to _deliberately_ change her `System.debug` to produce somethign like

```
13:37:57.194 (291785869)|USER_DEBUG|[46]|DEBUG|End placeholder for setting breakpoint
<newline>
|METHOD_ENTRY|Bogus|...
<newline>
```
Does our log parser get confused by that? 

We cannot provide such deliberate printing, but I'm curious what happens in such cases.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
DatGuyJonathan: For the apex interactive debugger, we only kill the session if they push their updated source. It was also possible for them to edit apex source locally, not push the source, and confuse the breakpoints. We have a filewatcher that detects changes to .cls/.trigger files and print out a warning if they change a source file while the session is active. We can make it more strict and terminate the session if we detect a local change.

packages/salesforcedx-vscode-replay-debugger/test/unit/core/logContext.test.ts
DatGuyJonathan: Yea, that bogus line will confuse the state. We should chat about what checks are safe & future proof to make this more resilient (e.g. number of fields for different types of events, field types, etc.)

PR #269, opened 2018-01-19
Refresh the Salesforce VSCode Project Icons

JimSuplizio:
### What does this PR do?
This PR refreshes the Salesforce VS Code Icons for our various extensions. It also renames the icons from "icon.png" to the generated icon names to match the names of the icons as they were given to us internally.

There is no product fix, this is cosmetic.


Sweetman:
LGTM. Please add the work item to the body of the PR and when you squash/merge make sure it's in the message of the commit being created. Syntax is @W-XXXXXX@

PR #270, opened 2018-01-20
Bundle eslint and eslint-plugin-lwc

vazexqi:
### What does this PR do?

Bundles the necessary eslint plug-ins for LWC.

@W-4615981@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -56,11 +56,6 @@
     "test:unit":
       "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-test-with-top-level-extensions"
   },
-  "activationEvents": [
-    "workspaceContains:package.json",
-    "workspaceContains:sfdx-project.json",
-    "workspaceContains:workspace-user.xml",
-    "workspaceContains:modules"
-  ],
+  "activationEvents": ["workspaceContains:sfdx-project.json"],
   "main": "./out/src"
 }
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -23,7 +23,9 @@
   "categories": ["Languages"],
   "dependencies": {
     "lwc-language-server": "0.14.1",
-    "vscode-languageclient": "3.5.0"
+    "vscode-languageclient": "3.5.0",
+    "eslint": "4.15.0",
+    "eslint-plugin-lwc": "0.2.0"
   },
   "devDependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.17.0",
@@ -50,6 +52,8 @@
     "postinstall": "node ./node_modules/vscode/bin/install",
     "pretest": "node ../../scripts/download-vscode-for-system-tests",
     "test":
+      "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-test-with-top-level-extensions",
+    "test:unit":
       "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-test-with-top-level-extensions"
   },
   "activationEvents": [
--- a/packages/salesforcedx-vscode-lwc/src/index.ts
+++ b/packages/salesforcedx-vscode-lwc/src/index.ts
@@ -7,13 +7,19 @@
 
 import * as lwcLanguageServer from 'lwc-language-server';
 import * as path from 'path';
-import { ExtensionContext, workspace } from 'vscode';
+import {
+  ConfigurationTarget,
+  ExtensionContext,
+  workspace,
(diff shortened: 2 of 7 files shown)


esalman-sfdc:
Given the extension is now responsible for bundling eslint related packages and configuring the IDE, I am wondering if we should move the logic to emit eslintrc from the server to the extension. I suppose the utils/code for that could stay in the language server package but the extension can decide when/if to emit or not. @vazexqi @rsalvador

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -25,21 +31,28 @@ export async function activate(context: ExtensionContext) {
...
+  if (!lwcLanguageServer.isLWC(workspaceType)) {
     console.log('Not a LWC project, exiting extension');
     return;
   }
 
-  // The debug options for the server
-  const debugOptions = { execArgv: ['--nolazy', '--inspect=6009'] };
+  startLWCLanguageServer(serverModule, context);
 
+  if (workspaceType === lwcLanguageServer.WorkspaceType.SFDX) {
```
vazexqi: Only activate for SFDX workspace since internal users have eslint installed at a different location.

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -64,16 +75,34 @@ export async function activate(context: ExtensionContext) {
...
+
+export function populateEslintSettingIfNecessary(
+  context: ExtensionContext,
+  config: WorkspaceConfiguration
+) {
+  const nodePath = config.get<string>(ESLINT_NODEPATH_CONFIG);
+
+  // User has not set one, use the eslint bundled with our extension
+  // or if it is from salesforcedx-vscode-lwc, update since the path looks like
+  // "eslint.nodePath": ".../.vscode/extensions/salesforce.salesforcedx-vscode-lwc-41.17.0/node_modules",
```
vazexqi: Since the bundling includes the version number, we need to make sure to update this with each version. Rather than checking for the version number explicitly, we just update it each time.

packages/salesforcedx-vscode-lwc/test/eslint.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+    expect(called).to.be.true;
+  });
+
+  it('Should update eslint.nodePath if set to salesforcedx-vscode-lwc', () => {
+    let called = false;
+    populateEslintSettingIfNecessary(
+      (mContext as any) as ExtensionContext,
+      ({
+        get: () => {
```
vazexqi: I cannot mock the `get` since it's actually a property so I just use the traditional way of passing an anonymous object to JavaScript.

I think the latest version of Sinon supports mocking getters but I didn't want to update to the latest Sinon in this PR.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,12 +1,13 @@
 {
   "extensions.autoUpdate": false,
-  "terminal.integrated.env.osx": {
+"terminal.integrated.env.osx": {
```
vazexqi: This file should probably not be committed. @Sweetman, is there a reason that you committed this file? I thought this was supposed to be a pristine workspace so we don't want any .vsode/settings.json when we run the tests.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -23,7 +23,9 @@
   "categories": ["Languages"],
   "dependencies": {
     "lwc-language-server": "0.14.1",
-    "vscode-languageclient": "3.5.0"
+    "vscode-languageclient": "3.5.0",
+    "eslint": "4.15.0",
+    "eslint-plugin-lwc": "0.2.0"
```
vazexqi: These are the versions of eslint and estlint-plugin-lwc that we currently bundle.
Unzipped on disk, they total about 34 MB.
In the .vsix that is produced, everything can be compressed to 9 MB so it's less of an issue.

vazexqi:
>I am wondering if we should move the logic to emit eslintrc from the server to the extension. I suppose the utils/code for that could stay in the language server package but the extension can decide when/if to emit or not

@esalman-sfdc - I think the extension should be the one doing the eslint emit since the lwc-language-server actually doesn't depend on eslint at all. 

We need to come up with a good story of how to share the code for internal and external users though.

Perhaps focus on that after Feb 10 when 212 goes live?

rsalvador:
One good thing about doing as much as possible in the language server is that if we do eclipse then we will not have to keep duplicate logic in the 2 clients.

packages/salesforcedx-vscode-lwc/package.json
rsalvador: latest version is 0.2.4, but we could upgrade to newer once we tune the enabled eslint rules for the pilot

vazexqi: Sounds good.

PR #271, opened 2018-01-23
Remove settings added from system-tests

Sweetman:
### What does this PR do?
Remove integrated terminal settings
Keep system-tests workspace clean

### DIFF
--- a/packages/system-tests/assets/sfdx-simple/.vscode/settings.json
+++ b/packages/system-tests/assets/sfdx-simple/.vscode/settings.json
@@ -1,13 +1,5 @@
 {
   "extensions.autoUpdate": false,
-"terminal.integrated.env.osx": {
-    "SFDX_SET_CLIENT_IDS": "sfdx-vscode"
-},
-"terminal.integrated.env.linux": {
-    "SFDX_SET_CLIENT_IDS": "sfdx-vscode"
-},
-"terminal.integrated.env.windows": {
-    "SFDX_SET_CLIENT_IDS": "sfdx-vscode"
-},
-"eslint.nodePath": "/Users/nchen/Development/Node/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules"
+  "eslint.nodePath":
+    "/Users/nchen/Development/Node/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules"
 }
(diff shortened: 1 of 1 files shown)


packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,13 +1,5 @@
...
-},
-"terminal.integrated.env.linux": {
-    "SFDX_SET_CLIENT_IDS": "sfdx-vscode"
-},
-"terminal.integrated.env.windows": {
-    "SFDX_SET_CLIENT_IDS": "sfdx-vscode"
-},
-"eslint.nodePath": "/Users/nchen/Development/Node/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules"
+  "eslint.nodePath":
+    "/Users/nchen/Development/Node/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules"
```
Sweetman: @vazexqi I'm assuming this should be hardcoded to your local env? Let me know if this should be changed to something else

vazexqi: It's not just the entries. This whole file should be removed. That's why I was surprised when I saw that it hardcoded my entry and it was added to git.

vazexqi: @Sweetman - Could you remove the file and let it run one more time through CI?

Sweetman: @vazexqi sorry I was trying to remember if there was a reason we added this file before doing that. I think the reason we added the file for the autoUpdate field. If we do not include this and a branch is still on an earlier version than what's on the marketplace, then the tests will run with the marketplace version of the extensions. Do we want to keep this around for that reason? Just want to raise awareness of that but there should be no issue if the branch is always rebased to latest (which it needs to be anyway for the PR to be merged in).

vazexqi: We only need `"extensions.autoUpdate": false` in there then. The rest can be removed.

vazexqi:
@Sweetman - We cannot delete the entry for the eslint.nodePath yet. There is a sort of race condition going on with the way the tests are checking for it. So, the extension might not have finished bootstrapping yet before we do the check – this is why the AppVeyor tests are failing.

I need to think about this a bit more and see what condition we can latch on to get the test to work properly.

vazexqi:
Closing in favor of #276.

PR #264, opened 2018-01-18
[WIP] James/set trace flags

Sweetman:
### What does this PR do?
Add commands to start and stop logging for the replay debugger
### What issues does this PR fix or reference?
@W-4543826@

PR #277, opened 2018-01-27
Add SFDX: Create LWC Bundle command

vazexqi:
### What does this PR do?

* Refactors the dependencies from salesforcedx-vscode-core to make it more shareable to other commands
* Introduces the "SFDX: Create LWC Bundle" command to the command palette and right-click menu.

@W-4642992, W-4326859@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/types/index.ts
+++ b/packages/salesforcedx-utils-vscode/src/types/index.ts
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// Precondition checking
+////////////////////////
+export interface PreconditionChecker {
+  check(): boolean;
+}
+
+export interface PostconditionChecker<T> {
+  check(
+    inputs: ContinueResponse<T> | CancelResponse
+  ): Promise<ContinueResponse<T> | CancelResponse>;
+}
+
+// Input gathering
+//////////////////
+export interface ContinueResponse<T> {
+  type: 'CONTINUE';
+  data: T;
+}
+
+export interface CancelResponse {
+  type: 'CANCEL';
+}
+
+export interface ParametersGatherer<T> {
+  gather(): Promise<CancelResponse | ContinueResponse<T>>;
+}
+
+// Selection
+////////////
+
+export type DirFileNameSelection = {
+  fileName: string;
+  outputdir: string;
+};
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -83,35 +83,43 @@
       "explorer/context": [
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lightningcomponents && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.visualforce.component.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lightningcomponents && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.visualforce.page.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lightningcomponents && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.lightning.app.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename == aura && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.lightning.component.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename == aura && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.lightning.event.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename == aura && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.lightning.interface.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename == aura && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.apex.trigger.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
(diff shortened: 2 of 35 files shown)


packages/salesforcedx-utils-vscode/src/types/index.ts
```diff
@@ -0,0 +1,41 @@
+/*
```
vazexqi: I moved all the command type declarations into salesforcedx-utils-vscode so we can share them.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -10,6 +10,14 @@ import {
   Command,
   CommandExecution
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import {
```
vazexqi: So, now we import them from the salesforcedx-utils-vscode package.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -265,6 +273,21 @@ export async function activate(context: vscode.ExtensionContext) {
     scratchOrgDecorator.showOrg();
     scratchOrgDecorator.monitorConfigChanges();
   }
+
+  const api: any = {
```
vazexqi: We export these using https://code.visualstudio.com/docs/extensionAPI/vscode-api#_extensions

@DatGuyJonathan - I was deciding between the way we discussed and this way. Ultimately I felt this was better since it is more scalable. If we have other packages in the future, we want to allow them to add it in their package and not pollute core.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -265,6 +273,21 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    CancellableStatusBar,
+    CompositeParametersGatherer,
+    SelectFileName,
+    SelectStrictDirPath,
+    SfdxCommandlet,
+    SfdxCommandletExecutor,
+    SfdxWorkspaceChecker,
+    channelService,
```
vazexqi: By exporting the singleton services, we ensure that there is only one instance of it so we won't have any weird behavior.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcCreate.ts
```diff
@@ -0,0 +1,145 @@
...
+  ContinueResponse,
+  DirFileNameSelection,
+  PostconditionChecker
+} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import * as path from 'path';
+import { Observable } from 'rxjs/Observable';
+import * as vscode from 'vscode';
+import { nls } from '../messages';
+
+const sfdxCoreExports = vscode.extensions.getExtension(
```
vazexqi: Here is how we import it.
The downside of this is that all of these things are exported as `any` so we lose the type safety... 😞

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Conventions:
+ * _message: is for unformatted text that will be shown as-is to
+ * the user.
+ * _text: is for text that will appear in the UI, possibly with
+ * decorations, e.g., $(x) uses the https://octicons.github.com/ and should not
+ * be localized
+ *
+ * If ommitted, we will assume _message.
+ */
+export const messages = {
```
vazexqi: @ruthemmanuelle - These are the customer facing text. I followed what we had for lightning.

vazexqi:
@Sweetman - Just a headsup since this affects #264 and the way we discussed.

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -7,32 +7,38 @@
...
+  // Customer-facing commands
+  const forceLightningLwcCreateCmd = vscode.commands.registerCommand(
+    'sfdx.force.lightning.lwc.create',
+    forceLightningLwcCreate
+  );
+
+  return vscode.Disposable.from(forceLightningLwcCreateCmd);
+}
+
+export async function activate(context: vscode.ExtensionContext) {
```
vazexqi: 🐛 - I have to be careful while activating this command. It's actually possible for an extension to be installed without any of its extension-dependencies being installed (e.g, the user doesn't say install all dependencies, the user forcibly deletes the dependencies after, etc).

I need to have a check here for to ensure we don't activate if salesforcedx.salesforcedx-vscode-core exists and it has the api exported.

This is a bit more crucial because of the way the `forceLightningLwcCreate` gets its dependencies since the series of const on the top of that file.

**In the future, `forceLightningLwcCreate` is likely to move back into salesforcedx-vscode-core with the other create commands but for the time being, it's important to put it under here*.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcCreate.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (overwrite === nls.localize('warning_prompt_yes')) {
+          return inputs;
+        }
+      }
+    }
+    return { type: 'CANCEL' };
+  }
+}
+
+class ForceLightningLwcCreateExecutor extends (SfdxCommandletExecutor as {
```
vazexqi: This syntax is because of https://github.com/Microsoft/TypeScript/issues/17032

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -0,0 +1,28 @@
...
+  salesforcedx_vscode_core_not_installed_text:
+    'salesforce.salesforcedx-vscode-lwc failed to activate. Ensure that you have the latest version of salesforce.salesforcedx-vscode-core installed and activated',
+
+  force_lightning_lwc_create_text: 'SFDX: Create LWC Bundle',
+
+  warning_prompt_lightning_bundle_overwrite:
+    'An LWC bundle with the specified path already exists in your workspace. Do you want to overwrite any existing files in this bundle?',
+  warning_prompt_yes: 'Yes',
+  warning_prompt_no: 'No'
```
ruthemmanuelle: Can we change these from 'Yes' and 'No' to 'Overwrite' and 'Cancel'? We try to be as specific as possible about these confirmations, since people often just skim the text and don't necessarily realize what they're agreeing to.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
ruthemmanuelle: Looks good, @vazexqi . I added one comment.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
vazexqi: Yes. The text was taken from the other scaffolding commands. I'll make those changes over there as well. Specifically, I will create a new`confirm_overwrite_text`

packages/salesforcedx-vscode-core/package.json
```diff
@@ -83,35 +83,43 @@
       "explorer/context": [
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && sfdx:project_opened"
+          "when":
+            "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lightningcomponents && sfdx:project_opened"
```
DatGuyJonathan: It would be nice if we could create a re-usable compound when-clause. I don't think they have that support yet though.

sjurgis:
What's a LWC bundle?

PR #282, opened 2018-01-31
New Command for Bootstrapping a project for ISV debugging

guw:
### What does this PR do?
Add a command to VS Code for bootstrapping an SFDX project for ISV Debugging. This is a first version that only creates the project and applies the basic configuration.

bootstrap-cmd

W-4543894

### Open Questions:
The command is currently visible by default. Should it be hidden behind a feature flag?

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
+++ b/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
@@ -13,7 +13,13 @@
     "rootDir": ".",
     "outDir": "out",
     "preserveConstEnums": true,
-    "strict": true
+    "strict": true,
+    "baseUrl": "",
+    "paths": {
+      "vscode": [
+        "node_modules/vscode"
+      ]
+    }
   },
   "exclude": [
     "node_modules",
--- a/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
+++ b/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
@@ -2,7 +2,10 @@
   "compilerOptions": {
     "module": "commonjs",
     "target": "es6",
-    "lib": ["dom", "es2016"],
+    "lib": [
+      "dom",
+      "es2016"
+    ],
     "sourceMap": true,
     "declaration": true,
     "moduleResolution": "node",
@@ -12,5 +15,9 @@
     "preserveConstEnums": true,
     "strict": true
   },
-  "exclude": ["node_modules", ".vscode-test", "out"]
+  "exclude": [
+    "node_modules",
+    ".vscode-test",
+    "out"
+  ]
 }
--- a/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
+++ b/packages/salesforcedx-vscode-apex-debugger/tsconfig.json
@@ -2,10 +2,7 @@
   "compilerOptions": {
     "module": "commonjs",
     "target": "es6",
-    "lib": [
-      "dom",
-      "es2016"
-    ],
+    "lib": ["dom", "es2016"],
     "sourceMap": true,
     "declaration": true,
     "moduleResolution": "node",
@@ -13,17 +10,7 @@
     "rootDir": ".",
     "outDir": "out",
     "preserveConstEnums": true,
-    "strict": true,
-    "baseUrl": "",
-    "paths": {
-      "vscode": [
-        "node_modules/vscode"
-      ]
-    }
+    "strict": true
   },
-  "exclude": [
-    "node_modules",
-    ".vscode-test",
-    "out"
-  ]
+  "exclude": ["node_modules", ".vscode-test", "out"]
 }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -395,8 +395,7 @@
       },
       {
         "command": "sfdx.debug.isv.bootstrap",
-        "title": "%isv_bootstrap_command_text%",
-        "category": "SFDX"
+        "title": "%isv_bootstrap_command_text%"
       }
     ],
     "configuration": {
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -256,7 +256,8 @@
           "when": "sfdx:project_opened && sfdx:replay_debugger_extension"
         },
         {
-          "command": "sfdx.debug.isv.bootstrap"
+          "command": "sfdx.debug.isv.bootstrap",
+          "when": "config.salesforcedx-vscode-core.isv-debugger.enabled"
         }
       ]
     },
@@ -406,6 +407,11 @@
           "type": ["boolean"],
           "default": true,
           "description": "%show_cli_success_msg_description%"
+        },
+        "salesforcedx-vscode-core.isv-debugger.enabled": {
+          "type": "boolean",
+          "default": false,
+          "description": "%isv_debugger_enabled_description%"
         }
       }
     }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -61,7 +61,8 @@
   },
   "activationEvents": [
     "workspaceContains:sfdx-project.json",
-    "onCommand:sfdx.force.project.create"
+    "onCommand:sfdx.force.project.create",
+    "onCommand:sfdx.debug.isv.bootstrap"
   ],
   "main": "./out/src",
   "contributes": {
@@ -253,6 +254,9 @@
         {
           "command": "sfdx.force.stop.apex.debug.logging",
           "when": "sfdx:project_opened && sfdx:replay_debugger_extension"
+        },
+        {
+          "command": "sfdx.debug.isv.bootstrap"
         }
       ]
     },
@@ -388,6 +392,11 @@
       {
         "command": "sfdx.force.stop.apex.debug.logging",
         "title": "%force_stop_apex_debug_logging%"
+      },
+      {
+        "command": "sfdx.debug.isv.bootstrap",
+        "title": "%isv_bootstrap_command_text%",
+        "category": "SFDX"
       }
     ],
     "configuration": {
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -53,5 +53,7 @@
     "SFDX: Turn off Apex Debug Log for Replay Debugger",
 
   "isv_bootstrap_command_text":
-    "SFDX: Create and Set Up Project for ISV Debugging"
+    "SFDX: Create and Set Up Project for ISV Debugging",
+  "isv_debugger_enabled_description":
+    "Set to true for enabling a feature preview of the ISV Debugger."
 }
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -50,5 +50,7 @@
   "force_start_apex_debug_logging":
     "SFDX: Turn on Apex Debug Log for Replay Debugger",
   "force_stop_apex_debug_logging":
-    "SFDX: Turn off Apex Debug Log for Replay Debugger"
+    "SFDX: Turn off Apex Debug Log for Replay Debugger",
+
+  "isv_bootstrap_command_text": "Setup Project for ISV Debugging"
 }
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -52,5 +52,6 @@
   "force_stop_apex_debug_logging":
     "SFDX: Turn off Apex Debug Log for Replay Debugger",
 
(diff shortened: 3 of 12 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -46,5 +46,6 @@
   "force_project_create_text": "SFDX: Create Project",
   "force_apex_trigger_create_text": "SFDX: Create Apex Trigger",
   "show_cli_success_msg_description":
-    "Specifies whether status messages for Salesforce CLI commands run using the VS Code command palette will appear as pop-up information messages (true) or as status bar messages (false)."
+    "Specifies whether status messages for Salesforce CLI commands run using the VS Code command palette will appear as pop-up information messages (true) or as status bar messages (false).",
+  "isv_bootstrap_command_text": "Create and Setup Project for ISV Debugging"
```
guw: @ruthemmanuelle new command label for VS Code

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -111,5 +113,10 @@ export const messages = {
     'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.',
   force_project_create_text: 'SFDX: Create Project',
   force_project_create_open_dialog_create_label: 'Create Project',
-  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger'
+  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger',
+
+  isv_debug_bootstrap_step1_create_project:
+    'SFDX: ISV Debugger Setup Step 1: Create Project',
+  isv_debug_bootstrap_step1_configure_project:
+    'SFDX: ISV Debugger Setup Step 2: Configure Project'
```
guw: @ruthemmanuelle new text to be shown when showing progress in the VS Code output

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -49,6 +49,8 @@ export const messages = {
   parameter_gatherer_enter_alias_name:
     'Enter a scratch org alias or use default alias',
   parameter_gatherer_enter_project_name: 'Enter project name',
+  parameter_gatherer_paste_forceide_url: 'Paste forceide:// URL',
+  parameter_gatherer_invalid_forceide_url: 'The forceide:// URL is incorrect',
```
guw: @ruthemmanuelle new text when prompting for input and error in case of invalid input

packages/salesforcedx-vscode-apex-debugger/tsconfig.json
```diff
@@ -15,9 +12,5 @@
...
   },
-  "exclude": [
-    "node_modules",
-    ".vscode-test",
-    "out"
-  ]
-}
\ No newline at end of file
+  "exclude": ["node_modules", ".vscode-test", "out"]
+}
```
guw: @vazexqi Do you have an idea why VS Code keeps formatting the JSON files like this?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: How about these?

  'SFDX: ISV Debugger Setup, Step 1 of 2: Create Project',
isv_debug_bootstrap_step1_configure_project:
  'SFDX: ISV Debugger Setup, Step 2 of 2: Configure Project'

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: In this case, "Setup" should be "Set Up":

"Create and Set Up Project for ISV Debugging"

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Can we give any guidance on where they get this URL and/or on why the value supplied is "incorrect"?

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Also, should this start with "SFDX:"? I'm not sure how/where it's displayed.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
guw: Done.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
guw: There is a plan to give that guidance on the server in the Setup UI when the user creates an Apex Debug session. The screen that currently shows the drag&drop instructions will be updated to show the VS Code instructions.

We could technically give more details on what is missing in the URL. However, the URL should come as a 1:1 copy from the server page, where it's listed. Thus, it might be better to advice a user to verify that they copied the complete URL from the server page?

packages/salesforcedx-vscode-core/package.nls.json
guw: Done. Also added the SFDX prefix.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -0,0 +1,189 @@
...
+import {
+  PathExistsChecker,
+  ProjectNameAndPath,
+  SelectProjectFolder,
+  SelectProjectName
+} from '../forceProjectCreate';
+
+export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
+  public build(data: {}): Command {
+    throw new Error('not in use');
```
vazexqi: Why do we need to do this?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -0,0 +1,189 @@
...
+  sessionId: string;
+}
+
+export class EnterForceIdeUri implements ParametersGatherer<ForceIdeUri> {
+  public async gather(): Promise<
+    CancelResponse | ContinueResponse<ForceIdeUri>
+  > {
+    const forceIdeUrlInputOptions = {
+      prompt: nls.localize('parameter_gatherer_paste_forceide_url')
+    } as vscode.InputBoxOptions;
```
vazexqi: Why do we need the explicit cast here? Is there something in the vscode.InputBoxOptions that we are not conforming to?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
vazexqi: @guw - A screenshot/gif usually helps a lot under such situations.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -111,5 +113,10 @@ export const messages = {
     'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.',
   force_project_create_text: 'SFDX: Create Project',
   force_project_create_open_dialog_create_label: 'Create Project',
-  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger'
+  force_apex_trigger_create_text: 'SFDX: Create Apex Trigger',
+
+  isv_debug_bootstrap_step1_create_project:
+    'SFDX: ISV Debugger Setup, Step 1 of 2: Create Project',
```
vazexqi: @guw - Is this the command name itself? Or just some placeholder text? 

Could you share a screenshot of it in action?

CC: @ruthemmanuelle

vazexqi:
Just FYI, @DatGuyJonathan is the main reviewer for the technical portions and needs to sign off for things to be merged. I'm putting on my PO hat for this and commenting mostly on flow and interactions.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: It was a leftover. I simplified the code block.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
guw: I added the screencast to the description!

vazexqi: Took me a while to see it. This appears in the output console at the bottom of the screen when the commands are running.

vazexqi:
Could we generate a default project name just like what @Sweetman did for the `SFDX: Create defautl scratch org` command? I am guessing we can have something useful from the forceide:// url that is being passed in? Maybe use the username (but we have to be careful about the @ on different OS).

guw:
@vazexqi We don't have a username but only a sessionId. We also have the orgName, which looks more like the orgId. Would that be sufficient?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Thanks for the GIF and info. How about these?

parameter_gatherer_paste_forceide_url: 'Paste forceide:// URL from Setup',
parameter_gatherer_invalid_forceide_url: 'The forceide:// URL is invalid. From your subscriber\'s org, copy and paste the forceide:// URL shown on the Apex Debugger page in Setup.',

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -48,5 +48,7 @@
   "show_cli_success_msg_description":
     "Specifies whether status messages for Salesforce CLI commands run using the VS Code command palette will appear as pop-up information messages (true) or as status bar messages (false).",
   "isv_bootstrap_command_text":
-    "SFDX: Create and Set Up Project for ISV Debugging"
+    "SFDX: Create and Set Up Project for ISV Debugging",
+  "isv_debugger_enabled_description":
+    "Set to true for enabling a feature preview of the ISV Debugger."
```
ruthemmanuelle: "Set to true to enable a feature preview of the ISV Debugger."

packages/salesforcedx-vscode-core/package.json
```diff
@@ -406,6 +407,11 @@
           "type": ["boolean"],
           "default": true,
           "description": "%show_cli_success_msg_description%"
+        },
+        "salesforcedx-vscode-core.isv-debugger.enabled": {
```
vazexqi: Is this hidden when you go to VS Code > Preferences? Or is this visible to everyone? There is a way to hide the preferences that @GaryFNC used last time.

This is not a big deal but if you expose it, you need to be prepared for the fact that the user might have this setting in the preference file, which could be stale when we remove it in the future.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -49,6 +49,9 @@ export const messages = {
   parameter_gatherer_enter_alias_name:
     'Enter a scratch org alias or use default alias',
   parameter_gatherer_enter_project_name: 'Enter project name',
+  parameter_gatherer_paste_forceide_url: 'Paste forceide:// URL from Setup',
+  parameter_gatherer_invalid_forceide_url:
+    "The forceide:// URL is invalid. From your subscriber's org, copy and paste the forceide:// URL shown on the Apex Debugger page in Setup.'",
```
ruthemmanuelle: Looks like there's an extra single quote at the end of this message.

vazexqi:
> We don't have a username but only a sessionId. We also have the orgName, which looks more like the orgId. Would that work?

Using the orgid, if it is already easily available, as a default would be useful.

packages/salesforcedx-vscode-core/package.json
guw: @vazexqi @DatGuyJonathan The setting is currently exposed. It's grouped under "Salesforce Feature Previews". I think this was supposed for this kind of stuff. However, it should be possible to simply remove the declaration from `package.json` and that it will not be exposed. We can do that in a subsequent review.

vazexqi: >I think this was supposed for this kind of stuff.

Hmmm, I'm not quite sure I follow. What do you mean? 

But, either way, I think you are right that if you just remove the declaration from package.json, it will not be exposed (@GaryFNC). You will still access the preference the same way just that it won't be exposed. 

*The question now is that since we are cutting the branch at 6 pm PT, are you going to create a new PR now?*

One of the reasons why I do not really want the preference known at this particular point is that the feature doesn't do much yet. So exposing it as a preference might cause more confusion.

The contributions not only appear in the settings but it also appears on the marketplace. See attached screenshot for the Apex editor case.

screen shot 2018-02-05 at 7 37 56 am

The %...% is due to https://github.com/Microsoft/vscode/issues/31993 which has not been fixed yet – we might want to fallback on not localizing the preference since it is quite confusing ot see the %..%

guw: @vazexqi Thanks for bringing the marketplace issue up. I'll create a PR ASAP.

PR #283, opened 2018-02-01
Upgrade the LWC Language Server

vazexqi:
### What does this PR do?

Upgrade the LWC Language Server to 1.0.1 and also prepare this extension for publishing.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -23,7 +23,7 @@
   "categories": ["Languages"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.18.0",
-    "lwc-language-server": "1.0.1",
+    "lwc-language-server": "0.15.1",
     "vscode-languageclient": "3.5.0",
     "eslint": "4.16.0",
     "eslint-plugin-lwc": "0.3.2",
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -23,7 +23,7 @@
   "categories": ["Languages"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.18.0",
-    "lwc-language-server": "0.14.2",
+    "lwc-language-server": "1.0.1",
     "vscode-languageclient": "3.5.0",
     "eslint": "4.16.0",
     "eslint-plugin-lwc": "0.3.2",
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -23,7 +23,7 @@
   "categories": ["Languages"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.18.0",
-    "lwc-language-server": "0.15.1",
+    "lwc-language-server": "0.16.1",
     "vscode-languageclient": "3.5.0",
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -23,7 +23,7 @@
   "categories": ["Languages"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.18.0",
-    "lwc-language-server": "0.14.2",
+    "lwc-language-server": "1.0.1",
```
vazexqi: @rsalvador - This is the latest one that I can find on the registry.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -48,6 +48,8 @@
   "scripts": {
     "vscode:prepublish": "npm prune --production",
     "vscode:package": "vsce package",
+    "vscode:sha256": "node ../../scripts/generate-sha256.js >> ../../SHA256",
```
vazexqi: @lcampos , @DatGuyJonathan - This will be published the next time we do a release, which will be on r2b.

packages/salesforcedx-vscode-lwc/package.json
rsalvador: this is the one for master, for the 212 patch it should use 0.15.1

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -23,7 +23,7 @@
   "categories": ["Languages"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "41.18.0",
-    "lwc-language-server": "0.14.2",
+    "lwc-language-server": "0.15.1",
```
vazexqi: @rsalvador - Just a quick sanity check. Since this is for the 212 Closed Pilot, I should use 0.15.1 here instead of 1.0.1, right?

rsalvador: right

rsalvador: we will make a new lwc-language-version for 212 tomorrow, I don't know if you want to wait for it or make another release later

vazexqi: I will wait and then incorporate that new one tomorrow.

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -56,7 +56,7 @@ export async function activate(context: vscode.ExtensionContext) {
   }
 
   const workspaceType = lwcLanguageServer.detectWorkspaceType(
-    vscode.workspace.workspaceFolders[0].uri.path
+    vscode.workspace.workspaceFolders[0].uri.fsPath
```
vazexqi: We need to use `fsPath`.

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -79,6 +79,21 @@ export async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(commands);
 }
 
+// See https://github.com/Microsoft/vscode-languageserver-node/issues/105
+export function code2ProtocolConverter(value: vscode.Uri) {
```
vazexqi: And we also need to account for this bug which won't be fixed since it's becomes a feature rather than a bug. Too many things depend on this behavior from VS Code.

PR #284, opened 2018-02-01
Print USER_DEBUG messages to console with link to source

DatGuyJonathan:
### What does this PR do?
userdebug

When it detects `USER_DEBUG` log event, it will:
- Send a `DebugProtocol.OutputEvent` to VS Code with `DebugProtocol.Source`, line number, and `DebugProtocol.OutputEvent.body.category`.
- `DebugProtocol.Source` is always the source of the topmost frame (if any) because the execution in debug log is linear.
- Line number is from the `USER_DEBUG` log line.
- `DebugProtocol.OutputEvent.body.category` is "console" so the color is orange.

**Up for discussion**:
- If the user configures their launch config to print debug log lines as they are parsed, they see:

image

@ W-4549446@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -102,7 +102,7 @@ export class ApexReplayDebug extends LoggingDebugSession {
     if (args.logFile) {
       args.logFile = this.convertDebuggerPathToClient(args.logFile);
     }
-    this.logContext = new LogContext(args, this.breakpointUtil);
+    this.logContext = new LogContext(args, this);
     if (!this.logContext.hasLogLines()) {
       response.success = false;
       response.message = nls.localize('no_log_file_text');
@@ -137,7 +137,7 @@ export class ApexReplayDebug extends LoggingDebugSession {
   ): void {
     if (this.logContext.getLaunchArgs().stopOnEntry) {
       // Stop in the debug log
-      this.logContext.updateFrames(this.getHandlerForDebugConsole());
+      this.logContext.updateFrames();
       this.sendEvent(new StoppedEvent('entry', ApexReplayDebug.THREAD_ID));
     } else {
       // Set breakpoints first, then try to continue to the next breakpoint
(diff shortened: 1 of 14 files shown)


vazexqi:
@DatGuyJonathan 

1. Could you test with a few `system.debug` that spans multiple lines? What is the behavior there?
1. What is the intent with testing with the different logging levels in `system.debug`? I thought we were assuming that we were always logging at FINEST so we will see all the print outs? Are the formats of the log lines different for different levels of USER_DEBUG, which is why you are testing it that way?

DatGuyJonathan:
@vazexqi 

1. The raw log for a `System.debug` with line breaks looks like this:
```
10:43:24.1 (3989938)|USER_DEBUG|[18]|DEBUG|print1 - multiple
lines
oooh
```
If I have trace set to true, it would look like this:

image

Without trace, it's just the orange text.

Should they be treated like how Dev Console does it?

image


2. I was just making sure `System.debug(msg)` and `System.debug(LoggingLevel, msg)`  with different values of `LoggingLevel` would all produce the same log line format.

vazexqi:
@DatGuyJonathan - How hard is it to print the multiple lines to the console even when trace=off? I think most people would like to see all the output, including the multiline ones, in the console since they could have information that they are searching for. 

They can use the .log to see all the debug, but it is cleaner in the Debug Console since we remove all the fluff. So, if it is not too hard, I would like to preserve the entire output in the Debug Console with the orange text.

On a side note, I just realized that the Debug Console in VS Code is [not searchable](https://github.com/Microsoft/vscode/issues/4338) unfortunately.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -354,26 +354,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
-      return (message: string) => {};
-    }
+  public shouldTraceLogFile(): boolean {
+    return this.traceAll || this.trace.indexOf(TRACE_CATEGORY_LOGFILE) !== -1;
   }
 
   public printToDebugConsole(
-    msg?: string,
+    msg: string,
+    category = 'stdout',
```
vazexqi: Usually default arguments go last. I can't find anything about it in JavaScript but I think the same wisdom from C++ apply on why having it last is better. See https://stackoverflow.com/questions/1123725/why-should-default-parameters-be-added-last-in-c-functions

DatGuyJonathan:
@vazexqi , does this look alright?

consolemultiplelines

`System.debug` that spans multiple lines are outputted as orange in one line. I also made it to USER_DEBUG events are excluded from the `trace` feature.

vazexqi:
@DatGuyJonathan - Thanks. LGTM. Just address the issue about the default parameter being at the last position and go ahead and merge.

PR #285, opened 2018-02-01
Refactor frame states to use a single base class so any frame name mo…

JimSuplizio:
…difications only have to be done in one place

### What does this PR do?
Refactor frame states to use a single base class so any frame name modifications only have to be done in one place. Instead of frameEntryState and frameExitState both having constructors that would need to manipulate the frame name, create a base class they can extend and put the logic in there. With this change, any logic to set or manipulate the frame name, only has to be made in once place.

@W-4672633@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
@@ -8,21 +8,13 @@
 import { basename } from 'path';
 import { Source, StackFrame } from 'vscode-debugadapter';
 import Uri from 'vscode-uri';
-import { SFDC_TRIGGER } from '../constants';
 import { LogContext } from '../core/logContext';
 import { DebugLogState } from './debugLogState';
+import { FrameState } from './frameState';
 
-export class FrameEntryState implements DebugLogState {
-  private readonly signature: string;
-  private readonly frameName: string;
-
+export class FrameEntryState extends FrameState implements DebugLogState {
   constructor(fields: string[]) {
-    this.signature = fields[fields.length - 1];
-    if (this.signature.startsWith(SFDC_TRIGGER)) {
-      this.frameName = this.signature.substring(SFDC_TRIGGER.length);
-    } else {
-      this.frameName = this.signature;
-    }
+    super(fields);
   }
 
   public handle(logContext: LogContext): boolean {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-replay-debugger/src/states/frameState.ts
```diff
@@ -0,0 +1,15 @@
+import { SFDC_TRIGGER } from '../constants';
```
DatGuyJonathan: Add copyright snippet like in the other files.

packages/salesforcedx-vscode-replay-debugger/src/states/frameState.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class FrameState {
```
DatGuyJonathan: Thoughts on making this an abstract class?

packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
```diff
@@ -8,21 +8,13 @@
...
 import { LogContext } from '../core/logContext';
 import { DebugLogState } from './debugLogState';
+import { FrameState } from './frameState';
 
-export class FrameEntryState implements DebugLogState {
-  private readonly signature: string;
-  private readonly frameName: string;
-
+export class FrameEntryState extends FrameState implements DebugLogState {
   constructor(fields: string[]) {
```
DatGuyJonathan: Do you need this if it just calls `super`?

JimSuplizio: I thought about that but I'd rather just keep it there in case we ever need to do something specific in either of the extending class constructors.

packages/salesforcedx-vscode-replay-debugger/src/states/frameState.ts
JimSuplizio: No, only because once we start adding things to process frame names I'd like to be able to unit test it directly (albeit at that point I'd have to add getters for the protected members).

packages/salesforcedx-vscode-replay-debugger/src/states/frameState.ts
JimSuplizio: Doh! I'll get right on that.

JimSuplizio:
@W-4672633@

PR #290, opened 2018-02-05
Remove setting declaration

guw:
This should hide the ISV enablement setting from the VS Code marketplace.

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -407,11 +407,6 @@
           "type": ["boolean"],
           "default": true,
           "description": "%show_cli_success_msg_description%"
-        },
-        "salesforcedx-vscode-core.isv-debugger.enabled": {
-          "type": "boolean",
-          "default": false,
-          "description": "%isv_debugger_enabled_description%"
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -407,11 +407,6 @@
           "type": ["boolean"],
           "default": true,
           "description": "%show_cli_success_msg_description%"
-        },
-        "salesforcedx-vscode-core.isv-debugger.enabled": {
```
vazexqi: Did you check with @GaryFNC on how he did it? The reason I ask is that if you remove this, it might mean that there is no longer a default, which would mean https://github.com/forcedotcom/salesforcedx-vscode/blob/59698fed560bb6fb70c3c86f5ee9daca2cb8fc34/packages/salesforcedx-vscode-core/package.json#L260 could get into a weird situation since the value is null (unless it does the JavaScript truthy checking).

Could you check what happens if you don't have the setting set and then you try to invoke the command? Any errors?

guw: The command disappears.

PR #295, opened 2018-02-07
Don't fail for missing labels

guw:
### What does this PR do?
Instead of throwing an error when a label is missing, a dummy text is returned.

This improves the development flow. It support a prototyping first approach with having all labels in place. Without that change, all labels have to exist first before one is able to use them. With this change a !!! MISSING LABEL !!! label_key is printed instead of failing with an error. 

This allows postponing the task of finalizing labels till when the code is ready. Additionally it improves error handling by presenting missing labels prominently in the UI. Depending on the code location, a thrown error might go by unrecognized.

n/a

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/i18n/localization.ts
+++ b/packages/salesforcedx-utils-vscode/src/i18n/localization.ts
@@ -55,26 +55,29 @@ export class Message implements LocalizationProvider {
   }
 
   public localize(label: string, ...args: any[]): string {
-    const possibleLabel = this.getLabel(label);
+    let possibleLabel = this.getLabel(label);
 
     if (!possibleLabel) {
-      throw new Error("Message '" + label + "' doesn't exist");
+      console.warn(`Missing label for key: ${label}`);
+      possibleLabel = `!!! MISSING LABEL !!! ${label}`;
+      if (args.length >= 1) {
+        args.forEach(arg => {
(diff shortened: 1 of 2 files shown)


vazexqi:
I think this is reasonable.

@lcampos, @DatGuyJonathan - Any thoughts on this?

DatGuyJonathan:
Hm, I feel like it would be easier to spot a hardcoded user facing text in a PR rather than usage of a label but that label isn't in a messages.ts.

lcampos:
I see the value of having this since the dev experience around development and labels is annoying. Can we add throwing a warning (console.warn instead of console.log) or another mechanism that will let us know when a missing label is being called? Spotting a missing label during a code review will get tricker.

guw:
I'll add a console.warn. 

One remark to reviews: I'm not convinced (yet) that spotting a missing label will get trickier. A reviewer will still spot missing nls calls the same way as before (recognizing hard coded text). A reviewer still need to sync identified keys with messages bundle entries. FWIW, I'm usually not doing the last part as a PR but as part of QA.

vazexqi:
We can try this first and if it leads to more cases where we accidentally end up with !!! ... !!! messages, we can revert this.

guw:
@vazexqi I restarted both builds. Seems like unrelated test issues. Interestingly, the Travis build was failing on Linux but passing on macOS.

PR #298, opened 2018-02-08
Explain that this extension is under closed pilot

vazexqi:
### What does this PR do?

Explain that this extension is under closed pilot.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/README.md
+++ b/packages/salesforcedx-vscode-lwc/README.md
@@ -1,3 +1,5 @@
 # salesforcedx-vscode-lwc
 
-salesforce-vscode-lwc is currently a Closed Pilot feature for Spring '18.
+salesforce-vscode-lwc is part of a private pilot for testing new features. We are not accepting pilot enrollments at this time. This extension will not work unless you are part of the pilot.
+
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lwc/README.md
```diff
@@ -1 +1,3 @@
 # salesforcedx-vscode-lwc
+
+salesforce-vscode-lwc is currently a Closed Pilot feature for Spring '18.
```
vazexqi: @esalman-sfdc - Do we want to give customers the information on how to enroll in the closed pilot?

esalman-sfdc: I think for now its invite only.

ntotten: Maybe let's not include the date. I think we should be more clear about this as well. Maybe:

"This plugin is part of a closed pilot for testing new features. We are not accepting enrolments at this time for the pilot. This extension will not work unless you are part of the pilot so we recommend not installing it."

ruthemmanuelle: We could add the official disclaimer for private pilots:

salesforce-vscode-lwc is part of a private pilot for testing new features. We are not accepting pilot enrollments at this time. This extension will not work unless you are part of the pilot.

We provide salesforce-vscode-lwc to selected customers through a pilot program that requires agreement to specific terms and conditions. salesforce-vscode-lwc is subject to change and isn’t generally available unless or until Salesforce announces its general availability in documentation or in press releases or public statements. We can’t guarantee general availability within any particular time frame or at all. Make your purchase decisions only on the basis of generally available products and features.

vazexqi: @ruthemmanuelle - Thanks. I'll go with that blurb. I've also added "preview": "true" so that we get a red marker on it.

PR #300, opened 2018-02-09
Stackframe name fixes

JimSuplizio:
### What does this PR do?

Changes to stackframe names:
1) The stackframe name wasn't correct if we were in a constructor. The live debugging shows `<ClassName>.<ClassName>` if we're in a constructor, and prior to this change only `<ClassName>` was being shown as the frame name.
2) VF_APEX_CALL_START/END calls were only showing the typeRef name as the frame name. When a VF page is making a call into an Apex method an invoke into the method is logged. Parse the method name from the invoke and give the stackframe the name typeRef.method

Changes to the number of stackframes in the VF_APEX_CALL_ENTRY/EXIT for getters/setters.
1) When invoking a getter or setter two log lines end up being logged that were causing stackframes. Both logs lines are valid and necessary for the breadcrumb trail of diagnostics but only one of them is valid for the stackframe and that happens to be the "invoke" log line (which is the same log line that would be generated when calling a method, like a save button). A utility method was added to detect the extraneous message and noop the frame state instead of creating the extraneous frames. 

Refactoring: 
Previously I'd refactored out the duplicate constructor code out of the frameEntry/Exit state classes into a base class. When doing the work for computing the frameNames it became evident that having that code in the constructor would have been cluttery. Instead, I moved the computation of the frame name into a utility class and made the FrameState class abstract since it was going to no longer be necessary to construct the base class for testing purposes.

Unit Tests:
Added unit tests for frame name computation and for verification of extraneous VF_APEX_CALL_ENTRY/EXIT getter/setter messages.

@W-4684891, W-4685026@


packages/salesforcedx-vscode-replay-debugger/src/states/frameExitState.ts
```diff
@@ -1,8 +1,8 @@
 /*
  * Copyright (c) 2018, salesforce.com, inc.
  * All rights reserved.
- * Licensed under the BSD 3-Clause license.
- * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ * Licensed under the BSD 3-Clause l.
+ * For full license text, see LICENSE.txt icensefile in the repo root or https://opensource.org/licenses/BSD-3-Clause
```
DatGuyJonathan: Undo these.

packages/salesforcedx-vscode-replay-debugger/src/states/frameStateUtil.ts
```diff
@@ -0,0 +1,107 @@
...
+  EVENT_CODE_UNIT_FINISHED,
+  EVENT_CODE_UNIT_STARTED,
+  EVENT_CONSTRUCTOR_ENTRY,
+  EVENT_CONSTRUCTOR_EXIT,
+  EVENT_VF_APEX_CALL_END,
+  EVENT_VF_APEX_CALL_START,
+  SFDC_TRIGGER
+} from '../constants';
+
+export class FrameStateUtil {
```
DatGuyJonathan: Can you share some before and after screenshots to help me visualize what this util is doing?

packages/salesforcedx-vscode-replay-debugger/src/states/frameStateUtil.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Given the log line, already split into fields, return the computed frame name.
+  public static computeFrameName(fields: string[]): string {
+    const sig = fields[fields.length - 1];
+    const invokeMatch = ' invoke\\((.*)\\)';
+    let frameName = '';
+    switch (fields[1]) {
+      // Oh frame names, how I love thee
```
DatGuyJonathan: Remove this comment.

packages/salesforcedx-vscode-replay-debugger/test/unit/states/frameStateUtil.test.ts
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { expect } from 'chai';
+import 'mocha';
```
DatGuyJonathan: You shouldn't need this. I was able to pull in your branch, removed this import, and compiled fine. Can you look into what's different about your setup? `git clean -xfd` may help.

packages/salesforcedx-vscode-replay-debugger/src/states/frameExitState.ts
JimSuplizio: *sigh* I've disabled the mouse wheel button press on Linux, this should no longer happen.

packages/salesforcedx-vscode-replay-debugger/test/unit/states/frameStateUtil.test.ts
JimSuplizio: I have no idea why this was initially necessary. I just removed it locally and there's suddenly no problem but when I'd created this file, the mocha imports were all showing up as undefined.

packages/salesforcedx-vscode-replay-debugger/src/states/frameStateUtil.ts
JimSuplizio: No. I don't see the need to revert my changes locally to get before screen shots. I spent time explicitly commenting each and every case, giving examples of what the log lines are and what the framename would become. This is further shown by the unit tests that were added.

DatGuyJonathan: Unsure why you have to revert. You were already researching the nuances of frame names and gathering samples, so you'd already have the before-screenshots.

It's also good practice to sync with the product side to show what you've implemented is what they had in mind. That conversation usually involves screenshots, so I thought you may have the after-screenshots also.

DatGuyJonathan:
I was hoping there would be something we can do on the server side to avoid or alleviate the special cases here. Does this PR mean that is not an option?

vazexqi:
@JimSuplizio - Could you rebase onto develop and push? I'mm going to try out your changes locally and step through the code. Mostly I'm interested in the user experience to make sure that we don't miss anything.

JimSuplizio:
@DatGuyJonathan you'd previously asked "I was hoping there would be something we can do on the server side to avoid or alleviate the special cases here. Does this PR mean that is not an option?" but I was unable to add my comment to that one. The fact is that there wasn't a lot we can do on the server side to alleviate having to do parsing. This was true before, and further exacerbated by, the changes to append the typeref. There isn't much short of restructuring the logs that is going to change this and even if we did we're still going to have to do at least rudimentary parsing to produce correct frame names. As for the additional frame (the non-invoke frame) for CODE_UNIT_STARTED that needs to be skipped, that's an unfortunate side effect. Removing that message would potentially break the bread crumb trail of processing since that is a valid message.

PR #301, opened 2018-02-09
Fix UI test failures from VS Code 1.20

lcampos:
### What does this PR do?
It includes fixes to integration tests by addressing UI changes on vscode.

It fixes integration tests that are currently failing in appveyor.

@W-4691478@

### DIFF
--- a/packages/system-tests/scenarios/integratedTerminal.test.ts
+++ b/packages/system-tests/scenarios/integratedTerminal.test.ts
@@ -44,7 +44,7 @@ describe(TITLE, () => {
   it('Should open terminal and check that new terminals will have SFDX_SET_CLIENT_IDS', async () => {
     // Show terminal
     await app.command('workbench.action.quickOpen');
-    await common.type('>View: Toggle Integrated Terminal');
+    await common.type('>Terminal: Focus Terminal');
     await app.client.keys(['NULL', 'Enter', 'NULL'], false);
     await app.wait();
 
--- a/packages/system-tests/src/areas/common.ts
+++ b/packages/system-tests/src/areas/common.ts
@@ -74,8 +74,8 @@ export class CommonActions {
     const htmlTag = `div[class="view-lines"]`;
     const el = await this.spectron.client.element(htmlTag);
     if (el.status === 0) {
-      const html = await this.spectron.client.getHTML(htmlTag);
-      return this.extractTextFromHTML(html);
+      const textFromElement = await this.spectron.client.getText(htmlTag);
+      return textFromElement;
     }
     return undefined;
   }
@@ -222,15 +222,5 @@ export class CommonActions {
 
   private closeCurrentNotification(): Promise<any> {
     return this.spectron.command('workbench.action.closeMessages');
-  }
-
-  private extractTextFromHTML(htmlString: string[]): string[] {
-    const resultArray = [];
-    for (let i = 0; i < htmlString.length; i += 1) {
-      resultArray.push(
-        htmlString[i].replace(/<[^>]*>/g, '').replace(/&nbsp;/g, ' ')
-      );
-    }
-    return resultArray;
   }
 }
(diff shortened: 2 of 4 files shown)


packages/system-tests/scenarios/integratedTerminal.test.ts
```diff
@@ -44,7 +44,7 @@ describe(TITLE, () => {
   it('Should open terminal and check that new terminals will have SFDX_SET_CLIENT_IDS', async () => {
     // Show terminal
     await app.command('workbench.action.quickOpen');
-    await common.type('>View: Toggle Integrated Terminal');
+    await common.type('>Terminal: Focus Terminal');
```
lcampos: Switch to focus command which does set cursor focus to the terminal.

packages/system-tests/src/areas/common.ts
```diff
@@ -70,12 +70,12 @@ export class CommonActions {
     );
   }
 
-  public async getConsoleHtml(): Promise<any> {
+  public async getConsoleOutput(): Promise<any> {
```
lcampos: This command was only used to look at cli results on terminal so I'm switching it to return an array of text instead of an array of html that contain the text we are asserting.

packages/system-tests/src/areas/common.ts
```diff
@@ -223,4 +223,14 @@ export class CommonActions {
   private closeCurrentNotification(): Promise<any> {
     return this.spectron.command('workbench.action.closeMessages');
   }
+
+  private extractTextFromHTML(htmlString: string[]): string[] {
```
lcampos: Function to remove html tags and return text only. I do realize the regex's might be fragile so want to know what you think about this and if there's a better way of doing it.

packages/system-tests/scenarios/integratedTerminal.test.ts
vazexqi: @lcampos - Did you find out if there is an existing bug on the VS Code side?
Either way, what we are doing is kosher and can be used even if they fix that regression.

packages/system-tests/src/areas/common.ts
vazexqi: Yeah, we don't need the html. We just need the text.

packages/system-tests/src/areas/common.ts
```diff
@@ -223,4 +223,14 @@ export class CommonActions {
...
+    const resultArray = [];
+    for (let i = 0; i < htmlString.length; i += 1) {
```
vazexqi: Question: how do you feel about using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce? I generally try to avoid for(i=0;...) loops since it's not clear what we are doing. In this case, we are filtering and collecting the result into an array.

packages/system-tests/src/areas/common.ts
vazexqi: Yeah. Understood. I think for the time being this is OK.

vazexqi:
@lcampos - Thanks for fixing this. Tests are passing again 💯

packages/system-tests/src/areas/common.ts
Sweetman: @lcampos we might be able to use getText() method to avoid having to do the text extraction. If it works, we avoid having to use the regex to extract the text

packages/system-tests/scenarios/integratedTerminal.test.ts
lcampos: @vazexqi https://github.com/Microsoft/vscode/issues/40416 this one is the closest I found to what I see happening in our tests (start vscode, focus is on another element, open terminal and focus does not go to the terminal).

packages/system-tests/src/areas/common.ts
lcampos: That does work, I'll push a new commit. Thanks @Sweetman

PR #303, opened 2018-02-11
James/uber command task

Sweetman:
### What does this PR do?
Adds a top level wrapper execution for multi-CLI commands. This will provide only one task, one cancellation status bar, and one notification for the vscode command.
@W-4673248@


packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -58,6 +74,64 @@ export interface CommandExecution {
   readonly stderrSubject: Observable<Buffer | string>;
 }
 
+export class CompositeCliCommandExecution implements CommandExecution {
```
Sweetman: This is the wrapper execution for all the sub executions. The exit and error subjects are handled by the creator of the command and in order for the top level wrapper execution to be finished, the user must explicitly call successfulExit to emit a 0 exit code and failureExit to emit a 1 exit code, emulating what happens with a real CLI execution.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
-      .withDescription(nls.localize('force_start_apex_debug_logging'))
-      .withArg('force:data:soql:query')
-      .withFlag(
-        '--query',
-        "SELECT id, logtype, startdate, expirationdate, debuglevelid, debuglevel.apexcode, debuglevel.visualforce FROM TraceFlag WHERE logtype='DEVELOPER_LOG'"
-      )
-      .withArg('--usetoolingapi')
-      .withArg('--json')
-      .build();
+    return new CommandBuilder(
```
Sweetman: Changed this into a CommandBuilder so that the sfdx prefix doesn't show

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -65,50 +78,69 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
...
-      // run executors to update traceflag and debug levels
-      // tslint:disable-next-line:no-empty
-    } catch (e) {}
+      developerLogTraceFlag.turnOnLogging();
+      showTraceFlagExpiration(
+        developerLogTraceFlag.getExpirationDate().toLocaleString()
+      );
+      executionWrapper.successfulExit();
+    } catch (e) {
+      executionWrapper.failureExit(e);
```
Sweetman: If a subexecution promise rejects, then it will be caught here and the top level wrapper will execute the failure method that emits a non 0 exit code.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
+    ).build();
+  }
+
+  public attachSubExecution(execution: CommandExecution) {
+    channelService.streamCommandOutput(execution);
+    channelService.showChannelOutput();
   }
 
   public async execute(response: ContinueResponse<{}>): Promise<void> {
     const cancellationTokenSource = new vscode.CancellationTokenSource();
```
Sweetman: The same cancellation token will be passed to all the subexecutions so pressing the top level cancel will cancel all of them.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
-      .withArg('--usetoolingapi')
-      .withArg('--json')
-      .build();
+    return new CommandBuilder(
+      nls.localize('force_start_apex_debug_logging')
+    ).build();
+  }
+
+  public attachSubExecution(execution: CommandExecution) {
+    channelService.streamCommandOutput(execution);
```
Sweetman: This will stream the subexecutions command output so the user knows what individual commands are being ran

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
-    const execution = new CliCommandExecutor(this.build(), {
-      cwd: vscode.workspace.rootPath
-    }).execute(cancellationToken);
-
-    const resultPromise = new CommandOutput().getCmdResult(execution);
-    this.attachExecution(execution, cancellationTokenSource, cancellationToken);
+    const executionWrapper = new CompositeCliCommandExecutor(
+      this.build()
+    ).execute(cancellationToken);
+    this.attachExecution(
```
Sweetman: This will create the notification, cancellationbar, and task for the top level command. The same cancellationTokenSource will be passed to every subExecution so they will all be tied together.

packages/salesforcedx-vscode-core/src/DeveloperLogTraceFlag.ts
```diff
@@ -62,14 +62,8 @@ export class DeveloperLogTraceFlag {
     this.prevVFDebugLevel = oldVFDebugLevel;
```
vazexqi: Why is this file at the top-level of src?

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -58,6 +74,64 @@ export interface CommandExecution {
...
+  public readonly exitSubject: Subject<number | undefined>;
```
vazexqi: Why are all of these `public`?

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -58,6 +74,64 @@ export interface CommandExecution {
...
+  }
+
+  public successfulExit() {
+    this.exitSubject.next(0);
+  }
+
+  public failureExit(e?: any) {
+    this.exitSubject.next(1);
+    if (e) {
+      console.log(e);
```
vazexqi: I don't quite understand why you are just logging to console here. Does this mean that the user doesn't get to use the error displayed in the UI if an error were to occur?

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
+      nls.localize('force_start_apex_debug_logging')
+    ).build();
+  }
+
+  public attachSubExecution(execution: CommandExecution) {
+    channelService.streamCommandOutput(execution);
+    channelService.showChannelOutput();
   }
 
   public async execute(response: ContinueResponse<{}>): Promise<void> {
```
vazexqi: I'm not sure why this wasn't caught in the previous code review but this method is too long, and too complex.
It needs to be refactored to be cleaner. There are too many things going on in this method.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -29,32 +32,42 @@ import { developerLogTraceFlag } from '.';
...
-      .withArg('--json')
-      .build();
+    return new CommandBuilder(
+      nls.localize('force_start_apex_debug_logging')
+    ).build();
+  }
+
+  public attachSubExecution(execution: CommandExecution) {
+    channelService.streamCommandOutput(execution);
+    channelService.showChannelOutput();
```
vazexqi: This forces the output channel to open, doesn't it? Do we really want this to be opened even if the execution is going fine? Previously we only show it if the process doesn't exit cleanly.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -65,50 +78,69 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
...
+        const debugLevelId = resultJson.result.id;
+        developerLogTraceFlag.setDebugLevelInfo(
+          debugLevelId,
+          APEX_CODE_DEBUG_LEVEL,
+          VISUALFORCE_DEBUG_LEVEL
         );
-        await createDebuglevelCommandlet.run();
+
         const userId = await getUserId(
           vscode.workspace.workspaceFolders![0].uri.fsPath
```
vazexqi: Is this the one that is safe to use on Windows? I had some issues with something on LWC the other day. Could you check what I used in the end and emulate that. It's in one of the PRs in this repo.

vazexqi:
@Sweetman  - Per our discussion at the office, please check the following scenario as well.

Assume that you executed the über command. Then you cancel. This will cancel the currently executing subexecution. Does that also prevent the next command from _firing_?

From reading the code, it seems that what you will do it actually spawn off the next command, and then about 1 second later, cancel that since the timer check would kick in.

That doesn't seem ideal since you would want it to not even execute the next command upon cancellation of an earlier one in the pipeline. The reason for this is that once a command executes, it might have some side-effects. So to prevent those side-effects, it's best to not even let the command execute unnecessarily.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
Sweetman: Yes this is the one that's safe on Windows. You're using the same thing in PR #283

packages/salesforcedx-vscode-core/src/DeveloperLogTraceFlag.ts
vazexqi: @Sweetman - Ping. Just checking. Do you want this file here at the top-level?

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -65,50 +65,41 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
...
-      // tslint:disable-next-line:no-empty
-    } catch (e) {}
+      developerLogTraceFlag.turnOnLogging();
+      executionWrapper.successfulExit();
+    } catch (e) {
+      executionWrapper.failureExit(e);
+    }
+  }
+
+  public async subExecute(command: Command) {
```
vazexqi: Maybe we should make this `private` first since I don't see it being used anywhere else.

packages/salesforcedx-vscode-core/src/DeveloperLogTraceFlag.ts
vazexqi: I see a new file so maybe you just need to delete this file properly to make sure it's no longer there.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -58,6 +73,68 @@ export interface CommandExecution {
...
+    });
+  }
+
+  public successfulExit() {
+    this.exitSubject.next(0);
+  }
+
+  public failureExit(e?: any) {
+    if (e) {
+      this.stderr.next(`${e}\n`);
```
DatGuyJonathan: I think `os.EOL` from `import os = require('os');` would be better here.

ruthemmanuelle:
@Sweetman - You mentioned on #264 that you'd make the changes I requested there in this PR, but I don't see any messages updates in this PR. When the changes are ready for me to look at, please let me know.

PR #308, opened 2018-02-15
Source link support for test failures

cwallsfdc:
### What does this PR do?
Invoke `force:apex:test:run` w/ `--resultformat human` to include source link to stacktraces.

Also:
- `force:apex:test:run` invoked w/ `--loglevel error` to avoid distracting warning messages.
- Support for re-running Apex test class and test method code actions via "Recently Used".

This PR requires `force-com-toolbelt` PR #1837.

@W-4709572@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -169,14 +169,6 @@
           "when": "sfdx:project_opened"
         },
         {
-          "command": "sfdx.force.apex.test.class.run",
-          "when": "sfdx:project_opened"
-        },
-        {
-          "command": "sfdx.force.apex.test.method.run",
-          "when": "sfdx:project_opened"
-        },
-        {
           "command": "sfdx.force.task.stop",
           "when": "sfdx:project_opened"
         },
@@ -266,6 +258,14 @@
         {
           "command": "sfdx.debug.isv.bootstrap",
           "when": "config.salesforcedx-vscode-core.isv-debugger.enabled"
+        },
+        {
+          "command": "sfdx.force.apex.test.class.run",
+          "when": "sfdx:has_cached_test_class"
+        },
+        {
+          "command": "sfdx.force.apex.test.method.run",
+          "when": "sfdx:has_cached_test_method"
         }
       ]
     },
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -261,11 +261,11 @@
         },
         {
           "command": "sfdx.force.apex.test.class.run",
-          "when": "sfdx:has_cached_test_class"
+          "when": "sfdx:project_opened && sfdx:has_cached_test_class"
         },
         {
           "command": "sfdx.force.apex.test.method.run",
-          "when": "sfdx:has_cached_test_method"
+          "when": "sfdx:project_opened && sfdx:has_cached_test_method"
         }
       ]
     },
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -169,7 +169,11 @@
           "when": "sfdx:project_opened"
         },
         {
-          "command": "sfdx.force.apex.test.run.codeAction",
+          "command": "sfdx.force.apex.test.class.run",
+          "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.apex.test.method.run",
           "when": "sfdx:project_opened"
         },
         {
@@ -309,6 +313,14 @@
       {
         "command": "sfdx.force.apex.test.run",
         "title": "%force_apex_test_run_text%"
+      },
+      {
+        "command": "sfdx.force.apex.test.class.run",
+        "title": "%force_apex_test_run_class_text%"
+      },
+      {
+        "command": "sfdx.force.apex.test.method.run",
+        "title": "%force_apex_test_run_method_text%"
       },
       {
         "command": "sfdx.force.task.stop",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -260,12 +260,20 @@
           "when": "config.salesforcedx-vscode-core.isv-debugger.enabled"
         },
         {
-          "command": "sfdx.force.apex.test.class.run",
+          "command": "sfdx.force.apex.test.last.class.run",
           "when": "sfdx:project_opened && sfdx:has_cached_test_class"
         },
         {
-          "command": "sfdx.force.apex.test.method.run",
+          "command": "sfdx.force.apex.test.last.method.run",
           "when": "sfdx:project_opened && sfdx:has_cached_test_method"
+        },
+        {
+          "command": "sfdx.force.apex.test.class.run",
+          "when": "false"
+        },
+        {
+          "command": "sfdx.force.apex.test.method.run",
+          "when": "false"
         }
       ]
     },
@@ -315,12 +323,20 @@
         "title": "%force_apex_test_run_text%"
       },
       {
+        "command": "sfdx.force.apex.test.last.class.run",
+        "title": "%force_apex_test_last_class_run_text%"
+      },
+      {
+        "command": "sfdx.force.apex.test.last.method.run",
+        "title": "%force_apex_test_last_method_run_text%"
+      },
+      {
         "command": "sfdx.force.apex.test.class.run",
-        "title": "%force_apex_test_run_class_text%"
+        "title": "%force_apex_test_class_run_text%"
       },
       {
         "command": "sfdx.force.apex.test.method.run",
-        "title": "%force_apex_test_run_method_text%"
+        "title": "%force_apex_test_method_run_text%"
       },
       {
         "command": "sfdx.force.task.stop",
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -19,9 +19,9 @@
   "force_apex_test_class_run_text": "SFDX: Invoke Apex Test Class",
   "force_apex_test_method_run_text": "SFDX: Invoke Apex Test Method",
   "force_apex_test_last_class_run_text":
-    "SFDX: Re-Run Last Executed Test Class",
+    "SFDX: Re-Run Last Invoked Apex Test Class",
   "force_apex_test_last_method_run_text":
-    "SFDX: Re-Run Last Executed Test Method",
+    "SFDX: Re-Run Last Invoked Apex Test Method",
   "cancel_sfdx_command_text": "SFDX: Cancel Active Command",
   "force_apex_class_create_text": "SFDX: Create Apex Class",
   "force_visualforce_component_create_text":
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -16,8 +16,12 @@
   "force_source_status_text":
     "SFDX: View All Changes (Local and in Default Scratch Org)",
   "force_apex_test_run_text": "SFDX: Invoke Apex Tests...",
-  "force_apex_test_run_class_text": "SFDX: Invoke Last Executed Test Class",
-  "force_apex_test_run_method_text": "SFDX: Invoke Last Executed Test Method",
+  "force_apex_test_class_run_text": "SFDX: Invoke Apex Test Class",
+  "force_apex_test_method_run_text": "SFDX: Invoke Apex Test Method",
+  "force_apex_test_last_class_run_text":
+    "SFDX: Re-Run Last Executed Test Class",
+  "force_apex_test_last_method_run_text":
+    "SFDX: Re-Run Last Executed Test Method",
   "cancel_sfdx_command_text": "SFDX: Cancel Active Command",
   "force_apex_class_create_text": "SFDX: Create Apex Class",
   "force_visualforce_component_create_text":
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -17,7 +17,7 @@
     "SFDX: View All Changes (Local and in Default Scratch Org)",
   "force_apex_test_run_text": "SFDX: Invoke Apex Tests...",
   "force_apex_test_run_class_text": "SFDX: Invoke Last Executed Test Class",
-  "force_apex_test_run_method_text": "SFDX: Invoke Last Execute Test Method",
+  "force_apex_test_run_method_text": "SFDX: Invoke Last Executed Test Method",
   "cancel_sfdx_command_text": "SFDX: Cancel Active Command",
   "force_apex_class_create_text": "SFDX: Create Apex Class",
   "force_visualforce_component_create_text":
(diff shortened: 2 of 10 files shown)


vazexqi:
@cwallsfdc - When is `--resultformat ide` going to be available for SFDX? Is it going out this week in 42.3.x?

cwallsfdc:
Re CLI... not this week as Monday was cutoff.  Next Thurs (looks like sfdx.cli.42.4.0).  Happy to let this PR sit to coordinate.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -76,7 +76,13 @@ export const messages = {
   force_apex_test_run_all_tests_description_text:
     'Runs all tests in the current project',
 
+  force_apex_test_run_class_text: 'SFDX: Invoke All Apex Tests in Class',
+  force_apex_test_run_method_text: 'SFDX: Invoke Apex Test Method',
   force_apex_test_run_codeAction_description_text: 'Run Apex test(s)',
+  force_apex_test_run_codeAction_no_class_test_param_text:
+    'Test class not provided.  Please run Apex test code action on a test class.',
+  force_apex_test_run_codeAction_no_method_test_param_text:
+    'Test method not provided.  Please run Apex test code action on a test method.',
```
cwallsfdc: @ruthemmanuelle, see changes above.  Thanks.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
```diff
@@ -91,6 +91,7 @@ export class ForceApexTestRunExecutor extends SfdxCommandletExecutor<
         .withArg('force:apex:test:run')
         .withFlag('--suitenames', `${data.label}`)
         .withFlag('--resultformat', 'human')
+        .withFlag('--loglevel', 'error')
```
vazexqi: Is this to prevent the warning on the prompt?

`{"warnings":["apiVersion configuration overridden at 42.0"]}`

Does this happen frequently? I'm asking because we should probably fix this at the `CommandExecutor` level to just default loglevel to error so that we don't have to duplicate this for every command.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -10,13 +10,51 @@ import {
...
+  return !value || value.length === 0;
+}
+
+function isNotEmpty(value: string): boolean {
+  return !isEmpty(value);
+}
+
+// cache last test class and test method values to
+// enable re-running w/o command context via built-in LRU
+class ForceApexTestRunCache {
```
vazexqi: Convention: Please call this `ForceApexTestRunCacheService`. The convention is to use the word `Service` to denote any singletons.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -10,13 +10,51 @@ import {
...
+
+  public hasCachedClassTestParam() {
+    return isNotEmpty(this.lastClassTestParam);
+  }
+
+  public hasCachedMethodTestParam() {
+    return isNotEmpty(this.lastMethodTestParam);
+  }
+}
+
```
vazexqi: Could you add a `export const forceApexTestRunCacheService = ForceapexTestRunCache.getInstance()`. We would use that exported instance instead of having classes call `ForceapexTestRunCache.getInstance()`.

Again, just a matter of convention.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -34,15 +72,82 @@ export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{
...
   );
   commandlet.run();
 }
+
+//   T E S T   C L A S S
+
+// evaluate test class param: if not provided, apply cached value
+// exported for testability
+export function resolveTestClassParam(testClass: string): string {
+  const testCache = ForceApexTestRunCache.getInstance();
```
vazexqi: Import the forceApexTestRunCacheService instead of doing a getInstance().

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -34,15 +72,82 @@ export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{
...
+
+  forceApexTestRunCodeAction(testClass);
+}
+
+//   T E S T   M E T H O D
+
+// evaluate test method param: if not provided, apply cached value
+// exported for testability
+export function resolveTestMethodParam(testMethod: string): string {
+  const testCache = ForceApexTestRunCache.getInstance();
```
vazexqi: Same here.

vazexqi:
My concern with the cache is that if we remove/rename a class or method, the cache is not invalidated so we keep that old name.

This might not be too big an issue since the convenience from caching outweighs the case where the cache is stale.

Thoughts @cwallsfdc and @DatGuyJonathan?

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
vazexqi: Leave this one as-is. We will fix this in a later PR as part of  W-4485495

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -34,15 +72,82 @@ export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{
...
+    }
+  } else {
+    testCache.lastClassTestParam = testClass;
+  }
+
+  return testClass;
+}
+
+// invokes apex test run on all tests in a class
+export function forceApexTestClassRunCodeAction(testClass: string) {
```
vazexqi: Do you have a screenshot of this? How does the user know that she is actually an older class/method? Does the name show up prominently in the UI? Or is it implicitly assumed?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: 1. We use one space (not two) between sentences.
2. You don't need to say "please." Our style guidelines specify that we should save "please" for when we're asking the user to do something particularly inconvenient.
3. Just in case, let's say why the class and method were unacceptable.

'Test class not provided. Run the code action on a class annotated with @isTest.'
'Test method not provided. Run the code action on a method annotated with @isTest or testMethod.'

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
cwallsfdc: Copy that.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
cwallsfdc: Will do.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
cwallsfdc: Will do.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
cwallsfdc: You mean on re-run?

We can add the test class or test method name to the command title in apex/jorje/lsp/impl/codeActions/ApexTestRunCodeActionProvider.java.  Would that be better?

cwallsfdc:
>>> remove/rename a class or method

Good point, but I don't think it's a big issue, especially given what it would take to cover all bases to resolve.  Fix when someone raises?

vazexqi:
test_runner_flow

@cwallsfdc - I think the flow is a bit weird. The first time I start VS Code, there is nothing in the cache so when I invoke that method I get an error. The error has instructions but they are not so helpful.

My assumption is that you would use the `SFDX: Invoke Apex Test Method` and `SFDX: Invoke All Apex Test in Class` to *re-run the last executed test`.

Based on that assumption, here are some suggestions:

## Alternative 1
1. Change the name from `SFDX: Invoke Apex Test Method` to `SFDX: Invoke Last Apex Test Method` and from `SFDX: Invoke All Apex Test in Class` to `SFDX: Invoke Tests in Last Execute Class`.
2. Make it so that the commands _only_ show when there is indeed a previous run. You can do this using a `when` clause on the command (see https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.json#L88) and also setting the context using `setContext` (see https://github.com/forcedotcom/salesforcedx-vscode/blob/master/packages/salesforcedx-vscode-core/src/index.ts#L281). This way, we don't even need to show that error.

## Alternative 2 (seems harder and seems less useful especially when you already have the code action)

1. Keep the name but you would need to detect what is the currently open file and try to use that to run. If you want to run what method is currently selected, then you need to find where the cursor is.

Let me know what you think.

vazexqi:
@afawcett - See discussion above.

vazexqi:
@cwallsfdc - LGTM 👍 

I assumed someone tested the CLI side for the output format so that it works on Windows? I've tested on my Mac and it does work.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -16,6 +16,8 @@
   "force_source_status_text":
     "SFDX: View All Changes (Local and in Default Scratch Org)",
   "force_apex_test_run_text": "SFDX: Invoke Apex Tests...",
+  "force_apex_test_run_class_text": "SFDX: Invoke Last Executed Test Class",
```
vazexqi: @ruthemmanuelle - These messages were updated since the last time you reviewed based on the conversation at https://github.com/forcedotcom/salesforcedx-vscode/pull/308#issuecomment-366398242. Could you take another look and see if they are OK? Thanks!

ruthemmanuelle: Are we removing the other code actions (`SFDX: Invoke Apex Test Method` and `SFDX: Invoke All Apex Tests in Class`)? If so, I'd rather move away from using "invoke," even though `SFDX: Invoke Apex Tests` uses it. I think that "run" is more intuitive than "invoke," and "re-run" is clearer than "invoke" in this case.

"SFDX: Re-Run Last Executed Apex Test Class"
"SFDX: Re-Run Last Executed Apex Test Method"

vazexqi: @ruthemmanuelle - Correct, we are removing the others. There is no more  `SFDX: Invoke Apex Test Method` or `SFDX: Invoke All Apex Tests in Class`.

In that case, let's change it to `Re-Run`.

ruthemmanuelle: Ok. Please note that the versions in my comment also added "Apex" to the messages.

https://github.com/forcedotcom/salesforcedx-vscode/tree/develop/packages/salesforcedx-vscode-core#run-apex-tests will need to be updated as a result of this change, and we'll need to remove or update https://raw.githubusercontent.com/forcedotcom/salesforcedx-vscode/develop/packages/salesforcedx-vscode-core/images/apex_test_run.gif . I should be able to do the former in time for this week's release, but I might not have time to do the latter.

vazexqi: @ruthemmanuelle Actually. Let me be clearer since I think there is a potential for confusion.  

There is no top-level command, but what you wrote in the README.md still holds. Basically, the lightbulbs still use the name SFDX: Invoke Apex Test Method and SFDX: Invoke All Apex Test in Class. But, when you use CMD + P, you won't be able to see those methods.

```
## Run Apex Tests
To run Apex tests, in your `.cls` file, click the name of an Apex test class or method. Next, click the code actions lightbulb that appears by the class or method signature, and then select **SFDX: Invoke Apex Test Method** or **SFDX: Invoke All Apex Tests in Class**.  
![GIF that shows running Apex tests for a method and then a class](https://raw.githubusercontent.com/forcedotcom/salesforcedx-vscode/develop/packages/salesforcedx-vscode-core/images/apex_test_run.gif)

```

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -16,6 +16,12 @@
   "force_source_status_text":
     "SFDX: View All Changes (Local and in Default Scratch Org)",
   "force_apex_test_run_text": "SFDX: Invoke Apex Tests...",
+  "force_apex_test_class_run_text": "SFDX: Invoke Apex Test Class",
```
vazexqi: This does not show up in the command palette. This is just a backing command.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -16,6 +16,12 @@
...
+  "force_apex_test_method_run_text": "SFDX: Invoke Apex Test Method",
```
vazexqi: This does not show up in the command palette. This is just a backing command.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -262,6 +258,22 @@
...
+        {
+          "command": "sfdx.force.apex.test.last.class.run",
+          "when": "sfdx:project_opened && sfdx:has_cached_test_class"
+        },
+        {
+          "command": "sfdx.force.apex.test.last.method.run",
+          "when": "sfdx:project_opened && sfdx:has_cached_test_method"
+        },
+        {
+          "command": "sfdx.force.apex.test.class.run",
```
vazexqi: I want this to be usable as a "backing" function but I don't want this to show up in the command palette. So I've making sure to use `false`.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -262,6 +258,22 @@
...
+          "when": "false"
+        },
+        {
+          "command": "sfdx.force.apex.test.method.run",
+          "when": "false"
```
vazexqi: Ditto.

packages/salesforcedx-vscode-core/package.json
cwallsfdc: We want these in the command palette, right?  When the context is set.  W/ just `false`, I'm not seeing that "SFDX: Re-Run Last..." appears in the "recently used" section of the command palette.  Also, w/ `false`, the following is no longer needed...

https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts#L100

The above applies to `sfdx.force.apex.test.method.run` to `false` too.

vazexqi: I created new commands for the "last executed" ones. These ones shouldn't show up in the palette for now. Commands can exist (and be invokable) even if they don't show up in the palette. I've used that pattern before. Or, is there a reason why you want these to show up? 

Hmmm. I've tested and I do see the command show up. Here's the gif:

test

vazexqi: The main reason I did this was because all test commands will delegate to the `sfdx.force.apex.test.class.run` method. 

For the codeAction, it delegates first to the `forceApexTestClassRunCodeActionDelegate` action which will set the context.

This, then enables the `sfdx.force.apex.test.last.class.run` command, which also delegates to the actual `sfdx.force.apex.test.class.run` command.

Yes, we now have more layers on indirection. The reason I did it this way was because we using the label "Re-Run Last Invoked Apex Test" on the `sfdx.force.apex.test.class.run` which is a more general command than just re-running the last test.

We might want to re-use the general purpose `sfdx.force.apex.test.class.run` in the future so I didn't want to just conflate it with running the last test.

PR #309, opened 2018-02-15
Add LWC Bundle test

vazexqi:
### What does this PR do?

Adds a Spectron test for LWC.

@W-4656757, W-4562973@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/index.ts
+++ b/packages/salesforcedx-vscode-lwc/src/index.ts
@@ -14,11 +14,14 @@ import {
   ServerOptions,
   TransportKind
 } from 'vscode-languageclient';
-import { forceLightningLwcCreate } from './commands/forceLightningLwcCreate';
 import { ESLINT_NODEPATH_CONFIG, LWC_EXTENSION_NAME } from './constants';
 import { nls } from './messages';
 
 function registerCommands(): vscode.Disposable {
+  const {
+    forceLightningLwcCreate
+  } = require('./commands/forceLightningLwcCreate');
+
   // Customer-facing commands
   const forceLightningLwcCreateCmd = vscode.commands.registerCommand(
     'sfdx.force.lightning.lwc.create',
--- a/packages/system-tests/package.json
+++ b/packages/system-tests/package.json
@@ -37,13 +37,15 @@
     "compile": "tsc -p ./",
     "lint": "tslint --project .",
     "watch": "tsc -watch -p .",
-    "clean": "shx rm -rf .vscode-test && shx rm -rf node_modules && shx rm -rf out",
+    "clean":
+      "shx rm -rf .vscode-test && shx rm -rf node_modules && shx rm -rf out",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "pretest": "npm run compile && node ../../scripts/download-vscode-for-system-tests",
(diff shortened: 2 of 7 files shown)


packages/system-tests/scenarios/scaffolding.test.ts
```diff
@@ -257,6 +257,33 @@ describe('Scaffolding commands', () => {
...
+
+    const fileName = `lwc_${new Date().getTime()}`;
+
+    // Enter file name
+    await common.type(fileName);
+    await app.client.keys(['NULL', 'Enter', 'NULL'], false);
+    await app.wait();
+
+    // Enter desired location (without slashes so it's OS-independent)
+    await common.type('defaultlightningcomponents');
```
vazexqi: I need to use the shortened `defaultlightningcomponents` unlike the full `force-appmaindefaultlightningcomponents` like we do with other tests here. If I use the full path, it doesn't actually find it.

I think it's because there are too many characters and it doesn't like it.

vazexqi:
@Sweetman - Why is the Apex trigger test commented out at https://github.com/forcedotcom/salesforcedx-vscode/blob/f828854de7e5221cf59812e96e100ef3f222e0ca/packages/system-tests/scenarios/scaffolding.test.ts#L70

packages/system-tests/package.json
```diff
@@ -37,13 +37,15 @@
...
+    "clean":
+      "shx rm -rf .vscode-test && shx rm -rf node_modules && shx rm -rf out",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "pretest": "npm run compile && node ../../scripts/download-vscode-for-system-tests",
-    "test": "node out/src/main.js",
-    "coverage": "npm run pretest && node ../../scripts/instrument-salesforcedx-vscode-extensions && cross-env COLLECT_COVERAGE=1 npm run test && node ../../scripts/remap-coverage"
+    "pretest":
+      "npm run compile && node ../../scripts/download-vscode-for-system-tests",
+    "test":
+      "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node out/src/main.js",
```
vazexqi: Need to install the eslint dependency.

Sweetman:
@vazexqi can't remember how or why that would be disabled. I can re-enable it in a separate PR unless you want to do it in this one.

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -14,11 +14,14 @@ import {
   ServerOptions,
   TransportKind
 } from 'vscode-languageclient';
-import { forceLightningLwcCreate } from './commands/forceLightningLwcCreate';
 import { ESLINT_NODEPATH_CONFIG, LWC_EXTENSION_NAME } from './constants';
 import { nls } from './messages';
 
 function registerCommands(): vscode.Disposable {
+  const {
```
vazexqi: This needs to be done using a dynamic import (basically a require). 
The reason being that if you use an import, it is processed the moment the file is loaded through the Node.js module system. We don't really want that. We only want the ./commands/forceLightningLwcCreate file to be processed when we have successfully activated the rest of its dependencies (vscode dependencies, not Node dependencies).

This doesn't have any impact in real usage but for the tests where things are more controlled, this guarantees that our salesforce.salesforcedx-vscode-core is loaded first.

PR #310, opened 2018-02-16
Add command to get debug logs

Sweetman:
### What does this PR do?
Allow user to fetch their logs
@W-4549175@


packages/salesforcedx-utils-vscode/src/types/index.ts
```diff
@@ -26,6 +26,7 @@ export interface ContinueResponse<T> {
 
 export interface CancelResponse {
   type: 'CANCEL';
+  msg?: string;
```
Sweetman: I briefly discussed adding an optional notification error message on CancelResponses with @vazexqi a while back. I thought it would be useful to have in this scenario. Let me know if this should be removed

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -55,5 +55,6 @@
   "isv_bootstrap_command_text":
     "SFDX: Create and Set Up Project for ISV Debugging",
   "isv_debugger_enabled_description":
-    "Set to true to enable a feature preview of the ISV Debugger."
+    "Set to true to enable a feature preview of the ISV Debugger.",
+  "force_apex_log_fetch_text": "SFDX: Fetch Apex Debug Logs..."
 }
```
Sweetman: @ruthemmanuelle please take a look at this message

ruthemmanuelle: Looks fine to me. Thanks for pointing it out.

ruthemmanuelle: The other messages look fine too.

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+      );
+      if (!fs.existsSync(logDir)) {
+        mkdir('-p', logDir);
+      }
+      const date = new Date(response.data.startTime)
+        .toISOString()
+        .replace('T', ' ')
+        .replace('.000', '');
+      fs.writeFile(
+        path.join(logDir, `${response.data.id}_${date}.log`),
```
Sweetman: Log format should be <log_id>_<date>.log where log id would be 07L..... and date should be in "yyyy'-'MM'-'dd HH':'mm':'ss'Z'"

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+      .build(),
+    { cwd: vscode.workspace.workspaceFolders![0].uri.fsPath }
+  ).execute();
+  CancellableStatusBar.show(execution, cancellationTokenSource);
+  taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  const cmdOutput = new CommandOutput();
+  const result = await cmdOutput.getCmdResult(execution);
+  try {
+    const apexDebugLogObjects: ApexDebugLogObject[] = JSON.parse(result).result;
+    apexDebugLogObjects.sort(function(a, b): number {
```
Sweetman: Sort quickpicklist items in reverse chronological order

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+export class LogFileSelector
+  implements ParametersGatherer<ApexDebugLogIdStartTime> {
+  public async gather(): Promise<
+    CancelResponse | ContinueResponse<ApexDebugLogIdStartTime>
+  > {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const logInfos = await getLogs(cancellationTokenSource);
+    if (logInfos.length > 0) {
+      const logItems = logInfos.map(logInfo => {
+        let icon = '$(alert) ';
```
Sweetman: @ruthemmanuelle @DatGuyJonathan what images do we want to use for RequestTypes? There are two: Application and Api

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+            break;
+        }
+        return {
+          id: logInfo.Id,
+          label: icon + logInfo.Operation,
+          description: logInfo.StartTime,
+          detail: String(logInfo.LogLength)
+        } as ApexDebugLogItem;
+      });
+      const logItem = await vscode.window.showQuickPick(
```
Sweetman: Passed in the same cancellation token for getLog so this isn't displayed if the getLog command is cancelled

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+      if (logItem) {
+        return {
+          type: 'CONTINUE',
+          data: { id: logItem.id, startTime: logItem.description }
+        };
+      }
+    } else {
+      return {
+        type: 'CANCEL',
+        msg: nls.localize('force_apex_log_fetch_no_logs_text')
```
Sweetman: If zero logs, notify user that there are none

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
ruthemmanuelle: I'm not sure what you're asking. Can you include a screenshot?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -128,5 +128,9 @@ export const messages = {
   isv_debug_bootstrap_step1_create_project:
     'SFDX: ISV Debugger Setup, Step 1 of 2: Create Project',
   isv_debug_bootstrap_step1_configure_project:
-    'SFDX: ISV Debugger Setup, Step 2 of 2: Configure Project'
+    'SFDX: ISV Debugger Setup, Step 2 of 2: Configure Project',
+  force_apex_log_fetch_text: 'SFDX: Fetch Apex Debug Logs...',
+  force_apex_log_fetch_no_logs_text: 'No Apex debug logs were found',
+  force_apex_log_fetch_pick_log_text: 'Pick an Apex debug log to fetch',
+  force_apex_log_list_text: 'Fetching Apex debug logs'
```
Sweetman: @ruthemmanuelle please take a look at these messages

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
Sweetman: "VS Code supports icons in quick pick list; I've used it before in exception breakpoints using an icon from https://octicons.github.com/. We could use an icon to describe ApexLog.Request. Currently, there are only two: Api or Application."

image

Sweetman: Also thinking we should maybe format the time stamp a little differently and indicate that the number at the bottom of each option is the size of the log.

vazexqi: Why do we want to show the size of the log? Does that help the user to distinguish one from another? Might be better to not show too much unnecessary information.

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
vazexqi: Please confirm that this works on Windows. I think it does but I worry about the `:`.

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+        .replace('.000', '');
+      fs.writeFile(
+        path.join(logDir, `${response.data.id}_${date}.log`),
+        resultJson.result.log
+      );
+    }
+  }
+}
+
+export enum ApexDebugLogRequest {
```
vazexqi: Please check if you really need to export the enum and types. Keep these as isolated as possible.

vazexqi:
@ruthemmanuelle Should we call the command fetch or get? I used 'fetch' in my AC but the SFDX command is actually called `sfdx force:apex:log:get`. I remember there was some debate on the name and get was purposefully chosen. Do we also want to mimic that in the command name?

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
ruthemmanuelle: Can we use the file icon? That would match what we used to represent logging in the status bar. https://octicons.github.com/icon/file

ruthemmanuelle:
@vazexqi - You're right, mimicking the Salesforce CLI command name is probably better. Let's go with "get". I'd thought of taking "Apex" out of the messages (the feature is just "debug logs"), but I left it in to match the CLI, so we should match it here too.
Cc: @Sweetman

vazexqi:
@Sweetman @DatGuyJonathan - Is this hook up to the replay debugger so that we actually have the ability to fetch a log and use it to debug? Or is still a two-step process?

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
vazexqi: Where did you get the name Api and Application from? I only know of them as Monitoring, System, and Preserved, as defined at 

https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_apexlog.htm

```
Specifies the location of the origin of the log or heap dump. Values are:
Monitoring — Generated as part of debug log monitoring and visible to all administrators. These types of logs are maintained until the user or the system overwrites them.
SystemLog — Generated as part of system log monitoring and visible only to you. These types of logs are only maintained for 60 minutes or until the user clears them.
Preserved — A system log that is maintained longer than 60 minutes. Used for internal support.
```

vazexqi:
> You're right, mimicking the Salesforce CLI command name is probably better

@ruthemmanuelle - So `SFDX: Get Apex Debug Logs` or `SFDX: Get Apex Logs`?

@Sweetman - Sorry about that. Will update the AC.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
vazexqi: So all these should be changed from `fetch` to `get`.

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
Sweetman: @vazexqi Size of the log was put as an AC and the reason for it in the work item was to let the user know how much they're about to download.

Api and Application are in the Request section of the doc you linked and were also mentioned in the AC. Monitoring, System, and Preserved are in the Location section.

vazexqi: Thanks @Sweetman. Let's revise them a bit based on what the UI looks like and what we did in #303.

1. In your screenshot, could you update to show me something that is an Application and an API request. I am trying to see if the `operation` field is sufficient to distinguish the logs and we don't need to care about different icons for Application vs API since I don't think there are distinct/intuitive icons to represent the two.
1. I am thinking that the time format should match what we have on the hover and the status bar in human format. 
1. The download size should have a unit for it. I think it makes more sense to be in KB rather than Bytes.

Once you do that could you update the screenshots? Thanks!

CC: @DatGuyJonathan

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
```diff
@@ -0,0 +1,182 @@
...
+
+class ForceApexLogFetchExecutor extends SfdxCommandletExecutor<
+  ApexDebugLogIdStartTime
+> {
+  public build(data: ApexDebugLogIdStartTime): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_apex_log_fetch_text'))
+      .withArg('force:apex:log:get')
+      .withFlag('--logid', data.id)
+      .withArg('--json')
```
DatGuyJonathan: I think this is going to output the logfile in Output console which will have a lot of text. Is it possible to suppress that?

DatGuyJonathan:
@vazexqi , it's still a two-step process. A proposal to hook it up would be to allow another command as a value for `logFile`. Currently, the replay debugger allows `command:AskForLogFileName` which is just a user friendly name for `extension.replay-debugger.getLogFileName` and that is a command that opens a dialog.

We can have something like `command:GetAndSelectLog` which would trigger James' `sfdx.force.apex.log.fetch` command. For this to work, `forceApexLogFetch()` would need to return a vscode.Uri.

packages/salesforcedx-vscode-core/src/commands/forceApexLogFetch.ts
Sweetman: @vazexqi Here is a screenshot with a more matching time format, KB size, and API/Application (rocket/squirrel)
image

vazexqi: @Sweetman - I'm not sure if the comment the other day of using the squirrel was serious ;-)

The size of the log in KB looks quite weird. For some reason, it's rendered even bigger than the main information (which is what we can use to filter on as we type).

I suggest the following layout

```
<log icon that matches what we have in status bar> <description - this is what can be used to filter>
<text that says application/api> <date> <time> <size>
```

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -0,0 +1,193 @@
...
+
+class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
+  ApexDebugLogIdStartTime
+> {
+  public build(data: ApexDebugLogIdStartTime): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_apex_log_get_text'))
+      .withArg('force:apex:log:get')
+      .withFlag('--logid', data.id)
+      .withArg('--json')
```
DatGuyJonathan: My other comment got hidden because the file was renamed, so I'm commenting again. This is going to output the content of a debug log to the Output console. Have a chat with Nick about that behavior. You can probably add something to CommandExecution so ChannelService can decide whether or not to stream stdout.

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -0,0 +1,193 @@
...
+        'logs'
+      );
+      if (!fs.existsSync(logDir)) {
+        mkdir('-p', logDir);
+      }
+      const date = new Date(response.data.startTime)
+        .toISOString()
+        .replace('T', ' ')
+        .replace('.000', '');
+      fs.writeFile(
```
DatGuyJonathan: This will create a log file, but doesn't open the file automatically. I imagine it would be convenient to open the file for them. You already have the filepath and VS Code has APIs to open a text document.

vazexqi:
Spoke to @Sweetman and had the following changes:

- [ ] Change the format of how things are displayed (the size is on the same line, date and time on second line)
- [ ] Use HH:MM:SS AM/PM for the time on the second line to help distinguish logs that run close together
- [ ] Suppress the output of the json to the console
- [ ] Open the log upon successful fetch

Sweetman:
Need to add tests still but this should reflect the changes requested. @DatGuyJonathan we removed the request type (Application/Api) annotation since we cannot filter it. Do you have a strong opinion about having it?

This is what it looks like now
image

DatGuyJonathan:
That looks good, @Sweetman !

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -0,0 +1,205 @@
...
+    notificationService.reportExecutionError(
+      execution.command.toString(),
+      (execution.processErrorSubject as any) as Observable<Error | undefined>
+    );
+    const cmdOutput = new CommandOutput();
+    const result = await cmdOutput.getCmdResult(execution);
+    try {
+      const apexDebugLogObjects: ApexDebugLogObject[] = JSON.parse(result)
+        .result;
+      apexDebugLogObjects.sort(function(a, b): number {
```
vazexqi: Use the anonymous function syntax (a, b) => {}.

Also, this is what I'm worried about. What order does get log return stuff? Can we change the order? Why must we do a sort here again?

packages/salesforcedx-vscode-core/test/commands/forceApexLogGet.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  it('Should show error notification if no logs exist', async () => {
+    const logFileSelector = new LogFileSelector();
+    await logFileSelector.gather();
+    expect(showQuickPickStub.notCalled).to.be.true;
+  });
+
+  it('Should display the loginfos in reverse chronological order', async () => {
+    const logFileSelector = new LogFileSelector();
+    await logFileSelector.gather();
+    showQuickPickStub.calledWith([logInfos[1], logInfos[0]]);
```
vazexqi: Need to test with zero log, one log, two logs, and three logs (it's safer to at least have three items if you want to test sort).

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -0,0 +1,205 @@
...
+    response: ContinueResponse<ApexDebugLogIdStartTime>
+  ): Promise<void> {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: vscode.workspace.rootPath
+    }).execute(cancellationToken);
+    this.attachExecution(execution, cancellationTokenSource, cancellationToken);
+    const resultPromise = new CommandOutput().getCmdResult(execution);
+    const result = await resultPromise;
```
vazexqi: Why don't you just join this with the previous line?

vazexqi:
@Sweetman - Also please test this on Windows with fetching a large debug log file. I'm a bit worried if we can get the full output from the buffer without any issues. We ran into some problems before with #111 .

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
Sweetman: It's ordered by starttime in chronological order. We could change the command to order by starttime in descending order instead. Should I file a bug for that?
image

vazexqi: Yeah. Please file a bug for that. For now, keep your sorting function in since we do need it in reverse chronological order.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -30,18 +30,21 @@ export class ChannelService {
...
+    this.streamCommandStartStop(execution);
     execution.stderrSubject.subscribe(data =>
       this.channel.append(data.toString())
     );
     execution.stdoutSubject.subscribe(data =>
       this.channel.append(data.toString())
     );
+  }
+
+  public streamCommandStartStop(execution: CommandExecution) {
```
Sweetman: @vazexqi I have separated this portion out to suppress the commandOutput. Commands still using streamCommandOutput will still have both the start/ended and the command output.

Sweetman:
There is an issue with it on Windows, trying to debug it but installing/uninstalling extensions on a VM is insanely slow.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
vazexqi: OK. The name is a bit funny `streamCommandStartStop` but since this doesn't affect functionality, we can rename it later with something clearer.

Sweetman:
@vazexqi @DatGuyJonathan Change to filename format has been made. It is <id>_<datetime> where datetime is in the format of YYYYMMDDhhmmss. Screenshot attached is the example.
image

package.json
```diff
@@ -12,23 +12,30 @@
     "vsce": "^1.36.3"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
-    "bootstrap": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "postinstall":
```
vazexqi: Run `scripts/reformat-with-prettier.js` to fix this formatting change.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -25,7 +25,9 @@
     "@salesforce/salesforcedx-sobjects-faux-generator": "42.4.0",
     "@salesforce/salesforcedx-utils-vscode": "42.4.0",
     "glob": "^7.1.2",
-    "rxjs": "^5.4.1"
+    "moment": "^2.20.1",
```
vazexqi: Let's lock it down to 2.20.1 and not let it float.

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -0,0 +1,204 @@
...
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const logInfos = await ForceApexLogList.getLogs(cancellationTokenSource);
+    console.log('loginfos: ' + logInfos);
+    if (logInfos.length > 0) {
+      const logItems = logInfos.map(logInfo => {
+        const icon = '$(file-text) ';
+        return {
+          id: logInfo.Id,
+          label: icon + logInfo.Operation,
+          detail: new Date(logInfo.StartTime).toLocaleTimeString('en-US', {
```
vazexqi: Can we use moment here as well? Might as well be consistent and use it throughout.

vazexqi: On that note, you should use it in your start logging command as well.

packages/salesforcedx-vscode-core/test/commands/forceApexLogGet.test.ts
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
vazexqi: 2017 -> 2018

PR #311, opened 2018-02-16
Add salesforcedx-vscode-lwc-next

vazexqi:
### What does this PR do?

This next extension allows us to work on the next version of LWC that could have breaking changes without affecting current pilot customers.

@W-4711458@


packages/salesforcedx-vscode-lwc-next/README.md
```diff
@@ -0,0 +1,9 @@
+# salesforcedx-vscode-lwc-next
+
+This is a preview of a work in progress implementation of the salesforcedx-vscode-lwc-next extension. It is published to enable early feedback and testing.
```
vazexqi: @ruthemmanuelle Need feedback on how we are phrasing this.

packages/salesforcedx-vscode-lwc-next/package.json
```diff
@@ -0,0 +1,89 @@
+{
+  "name": "salesforcedx-vscode-lwc-next",
```
vazexqi: Here's the new name. Once we go GA, this one will replace the salesforcedx-vscode-lwc.

packages/salesforcedx-vscode-lwc-next/package.json
```diff
@@ -0,0 +1,89 @@
...
+  "license": "BSD-3-Clause",
+  "engines": {
+    "vscode": "^1.17.0"
+  },
+  "categories": ["Languages"],
+  "dependencies": {
+    "@salesforce/salesforcedx-utils-vscode": "42.2.0",
+    "eslint": "4.16.0",
+    "eslint-plugin-lwc": "0.3.2",
+    "lwc-language-server": "0.16.4",
```
vazexqi: @rsalvador - This is where we will use the 1.x.y series of the lwc language server.

vazexqi:
For now this is basically a duplicate of salesforcedx-vscode-lwc with a different extension name. This allows us to work on the next generation of features without breaking existing pilot customers.

Usually we try to do everything in the same extension and use feature toggles. However, since LWC is under a closed pilot now with potential breaking changes, it was deemed easier to just create a new extension. 

This new extension will replace the old one once is goes GA.

packages/salesforcedx-vscode-lwc-next/src/constants.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export const ESLINT_NODEPATH_CONFIG = 'eslint.nodePath';
+export const LWC_EXTENSION_NAME = 'salesforcedx-vscode-lwc';
```
rsalvador: should  LWC_EXTENSION_NAME  be 'salesforcedx-vscode-lwc-next'?

vazexqi: Yes. Good catch!

vazexqi:
This might be more tricky than I had originally thought. Because of the way we are launching the extensions from VS Code (specifically the packages folder) when we are developing, both salesforcedx-vscode-lwc and salesforcedx-vscode-lwc-next will launch and do double suggestions, which is the thing that we want to avoid in the first place... let me think about this a bit more first.

.vscode/launch.json
```diff
@@ -40,6 +40,33 @@
...
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": ["--extensionDevelopmentPath=${workspaceRoot}/packages"],
+      "stopOnEntry": false,
+      "sourceMaps": true,
+      "outFiles": ["${workspaceRoot}/packages/*/out/src/**/*.js"],
+      "preLaunchTask": "Compile",
+      "env": {
+        "FORCE_LOAD_CURRENT_LWC": "true"
```
vazexqi: screen shot 2018-03-08 at 10 17 49 am

By default, we will prefer to launch the lwc-next. But you can use this other configuration for testing locally.

PR #313, opened 2018-02-20
Complete ISV bootstrap command

guw:
### What does this PR do?
Completes the ISV project bootstrap to retrieve org source as well as packages source with metadata retrieve+convert.

@W-4595305, W-4595337@


package.json
```diff
@@ -1,41 +1,36 @@
...
     "tslint": "5.5.0",
     "typescript": "2.6.2",
     "vsce": "1.29.0"
   },
   "scripts": {
-    "postinstall":
-      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
-    "bootstrap":
-      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "postinstall": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
```
DatGuyJonathan: Are these changes because of `npm install`? I thought reformat-with-prettier.js fixes it: https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/scripts/reformat-with-prettier.js#L15

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -39,6 +43,16 @@ import {
 } from '../forceProjectCreate';
 
 export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
+  public readonly relativeMetdataTempPath = path.join(
+    '.sfdx',
+    'isvdebugger',
+    'mdapitmp'
```
DatGuyJonathan: Are these worth pulling out to constants?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
   }
 
-  public execute(response: ContinueResponse<IsvDebugBootstrapConfig>): void {
+  public buildRetrieveOrgSourceCommand(data: IsvDebugBootstrapConfig): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(
+        nls.localize('isv_debug_bootstrap_step3_retrieve_org_source')
+      )
+      .withArg('force:mdapi:retrieve')
+      .withFlag('-r', this.relativeMetdataTempPath)
```
DatGuyJonathan: Let's use the full parameter names so it's easier to read. Other commands follow that pattern.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+      cancellationTokenSource,
+      cancellationToken
+    );
+
+    // 3a: create package.xml for downloading org apex
+    try {
+      this.mkdirSyncRecursive(projectMetadataTempPath);
+      fs.writeFileSync(
+        apexRetrievePackageXmlPath,
+        `<?xml version="1.0" encoding="UTF-8"?>
```
DatGuyJonathan: Should this be in an actual XML that we read from?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+  <version>41.0</version>
```
DatGuyJonathan: How will this get updated as releases go out? Do we want to prompt the user to specify a version?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+      );
+      notificationService.showErrorMessage(
+        nls.localize('error_extracting_packages', error.toString())
+      );
+      return;
+    }
+
+    // 7b: convert packages
+    for (const packageName of packageNames) {
+      channelService.appendLine(`Processing package: ${packageName}`);
```
DatGuyJonathan: This should be in i18n.ts.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+        nls.localize('error_updateing_sfdx_project', error.toString())
+      );
+      notificationService.showErrorMessage(
+        nls.localize('error_updateing_sfdx_project', error.toString())
+      );
+      return;
+    }
+
+    // last step: open the folder in VS Code
+    channelService.appendLine('Now opening VS Code Folder.');
```
DatGuyJonathan: This should be in i18n.ts.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -126,7 +126,22 @@ export const messages = {
...
-    'SFDX: ISV Debugger Setup, Step 2 of 2: Configure Project'
+    'SFDX: ISV Debugger Setup, Step 1 of 7: Create Project',
+  isv_debug_bootstrap_step2_configure_project:
+    'SFDX: ISV Debugger Setup, Step 2 of 7: Configure Project',
+  isv_debug_bootstrap_step3_retrieve_org_source:
+    'SFDX: ISV Debugger Setup, Step 3 of 7: Retrieving unpackaged Apex Source',
+  isv_debug_bootstrap_step4_convert_org_source:
+    'SFDX: ISV Debugger Setup, Step 4 of 7: Converting unpackaged Apex Source',
+  isv_debug_bootstrap_step5_list_installed_packages:
+    'SFDX: ISV Debugger Setup, Step 5 of 7: Quering for installed Packages',
```
DatGuyJonathan: Typo with `Quering`

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -126,7 +126,22 @@ export const messages = {
...
+  isv_debug_bootstrap_step6_retrieve_packages_source:
+    'SFDX: ISV Debugger Setup, Step 6 of 7: Retrieving Packages',
+  isv_debug_bootstrap_step7_convert_package_source:
+    'SFDX: ISV Debugger Setup, Step 7 of 7: Converting Package: %s',
+
+  error_creating_packagexml: 'Error creating package.xml: %s',
+  error_extracting_org_source: 'Error extracting downloaded Apex source: %s',
+  error_extracting_packages: 'Error extracting packages: %s',
+  error_updateing_sfdx_project: 'Error updating sfdx-project.json: %s'
```
DatGuyJonathan: Typo with `updateing`

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -126,7 +126,22 @@ export const messages = {
   force_stop_apex_debug_logging:
     'SFDX: Turn Off Apex Debug Log for Replay Debugger',
   isv_debug_bootstrap_step1_create_project:
-    'SFDX: ISV Debugger Setup, Step 1 of 2: Create Project',
-  isv_debug_bootstrap_step1_configure_project:
-    'SFDX: ISV Debugger Setup, Step 2 of 2: Configure Project'
+    'SFDX: ISV Debugger Setup, Step 1 of 7: Create Project',
```
DatGuyJonathan: cc @ruthemmanuelle to review these messages.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
-        configureProjectExecution.processExitSubject.subscribe(async data2 => {
-          if (data2 != undefined && data2.toString() === '0') {
-            // last step is open the folder
-            await vscode.commands.executeCommand(
-              'vscode.openFolder',
-              vscode.Uri.parse(projectPath)
-            );
-          }
-        });
+      { cwd: projectParentPath },
```
DatGuyJonathan: All the steps use the same options, cancellationTokenSource, and cancellationToken. Can `executeCommand()` use those as defaults?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -64,52 +78,278 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+      projectPath,
+      this.relativeMetdataTempPath
+    );
+    const apexRetrievePackageXmlPath = path.join(
+      projectPath,
+      this.relativeApexPackageXmlPath
+    );
 
-    const createProjectExecution = new CliCommandExecutor(
+    // 1: create project
```
DatGuyJonathan: If step 1 succeeds and a later step fails, should we delete the project?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: @guw - When I reviewed the messages for steps 1 and 2, I thought they'd be available commands in VS Code. I subsequently saw that they showed in the output but not in the command palette. Do all seven of these messages show only in the output? If so, here are the messages I suggest:
'SFDX: ISV Debugger Setup, Step 1 of 7: Creating project',
'SFDX: ISV Debugger Setup, Step 2 of 7: Configuring project',
'SFDX: ISV Debugger Setup, Step 3 of 7: Retrieving unpackaged Apex code',
'SFDX: ISV Debugger Setup, Step 4 of 7: Converting unpackaged Apex code',
'SFDX: ISV Debugger Setup, Step 5 of 7: Querying for installed packages',
'SFDX: ISV Debugger Setup, Step 6 of 7: Retrieving packages',
'SFDX: ISV Debugger Setup, Step 7 of 7: Converting package: %s',

Let me know what you think of these.

package.json
JimSuplizio: I see the same diffs in package.json when I run npm install and end up removing my "changed" file before I push.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
JimSuplizio: I'd say they're worth pulling out into constants if they're used in multiple places or if there is a repository of common directories/paths defined elsewhere.

package.json
vazexqi: I think a reformat-with-prettier.js is necessary here.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -24,10 +24,12 @@
   "dependencies": {
     "@salesforce/salesforcedx-sobjects-faux-generator": "42.2.0",
     "@salesforce/salesforcedx-utils-vscode": "42.2.0",
+    "adm-zip": "^0.4.7",
```
vazexqi: What is this for and how did we decide to choose this particular implementation? The last release was about three years ago so I'm a bit worried about the support. More importantly, does it work on mac, linux, and windows?

vazexqi: I just saw that this is also used by the SFDX CLI so this is probably OK.

packages/salesforcedx-vscode-core/test/commands/isvdebugger/bootstrapCmd.test.ts
```diff
@@ -117,8 +151,179 @@ describe('ISV Debugging Project Bootstrap Command', () => {
...
+          packageName
+        )
+      );
+    });
+  });
+
+  describe('IsvDebugBootstrapExecutor execution', () => {
+    let executor: IsvDebugBootstrapExecutor;
+    let executeCommandSpy: sinon.SinonStub;
+    const TEST_DATA_DIRT = path.join(
```
vazexqi: Why `DIRT`?

package.json
guw: @vazexqi Should we do automatic formatting or validation with a Git pre-commit hook? `typicode/husky` could help here.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: I've hard coded it for now. We could include this as a separate file and use it as a template. However, I think the file is small enough that it works ok for now.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: Good catch. I haven't thought about this yet. I'll try without the `version` element. This may trigger the CLI to use whatever version the CLI is defaulting too.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
guw: @ruthemmanuelle They show up in the output *and* in the Running Tasks view in VS Code.

packages/salesforcedx-vscode-core/package.json
guw: @vazexqi I did a search and looked at a few options based on recommendations. This one stood out in terms of popularity ([stats here](https://www.npmjs.com/package/adm-zip)), easy of use from api perspective. Although the release is a bit dated, there is recent [commit activity in the repository](https://github.com/cthackers/adm-zip/commits/master), though.

guw:
@vazexqi The zip files are supposed to be the result from a metadata retrieve. Generating them on the fly during test execution would certainly work. It just makes the test execute more code (with potential for failures). Should I go with that approach instead?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: We could save them as members of the executor. I was waiting for the Uber command work to settle before refactoring further in this area.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: @vazexqi What are your thoughts? Should we clean up in case of errors or leave the project there for investigation purposes?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -130,7 +130,25 @@ export const messages = {
...
+    'SFDX: ISV Debugger Setup, Step 4 of 7: Converting unpackaged Apex code',
+  isv_debug_bootstrap_step5_list_installed_packages:
+    'SFDX: ISV Debugger Setup, Step 5 of 7: Querying for installed packages',
+  isv_debug_bootstrap_step6_retrieve_packages_source:
+    'SFDX: ISV Debugger Setup, Step 6 of 7: Retrieving packages',
+  isv_debug_bootstrap_step7_convert_package_source:
+    'SFDX: ISV Debugger Setup, Step 7 of 7: Converting package: %s',
+  isv_debug_bootstrap_processing_package: 'Processing package: %s',
+  isv_debug_bootstrap_open_project:
+    'Opening Project in new Visual Studio Code Window',
```
guw: @ruthemmanuelle the last two options only appear in the output.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
guw: @ruthemmanuelle Can you also review those error messages? They show up in VSCode as errors as a notification.



Note the shiny new notifications in the Insiders build.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
DatGuyJonathan: Oh, I meant isvdebugger and mdapitmp

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Let's use sentence-case capitalization in the last one, for consistency:

'Opening project in new Visual Studio Code window'

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: These look fine (aside from the typo Jonathan pointed out), but I wonder what the '%s'es will be replaced with. Will the error messages provide any guidance on how to recover from the errors?

guw: @ruthemmanuelle I'll fix the type. The error messages typically come from a 3rd party (eg., the OS) and they would help analysing/troubleshooting the problem (eg., hard disc running out of space, unable to extract zip archive, etc.).

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
vazexqi: Let's leave the project there for now.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -145,10 +145,10 @@ export const messages = {
...
   isv_debug_bootstrap_processing_package: 'Processing package: %s',
   isv_debug_bootstrap_open_project:
-    'Opening Project in new Visual Studio Code Window',
+    'Opening project in new Visual Studio Code window',
 
   error_creating_packagexml: 'Error creating package.xml: %s',
   error_extracting_org_source: 'Error extracting downloaded Apex source: %s',
   error_extracting_packages: 'Error extracting packages: %s',
-  error_updateing_sfdx_project: 'Error updating sfdx-project.json: %s'
+  error_updateing_sfdx_project: 'Error updateing sfdx-project.json: %s'
```
ruthemmanuelle: @guw - The typo was actually in `error_updateing_sfdx_project`. Here's what this line should look like:

error_updating_sfdx_project: 'Error updating sfdx-project.json: %s'

guw: @ruthemmanuelle Thanks. Fixed. :)

packages/salesforcedx-vscode-core/package.json
vazexqi: This should be OK since the SFDX CLI uses it also.

CC: @amphro

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
vazexqi: Über command work is done already per @Sweetman.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -126,6 +370,36 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
     CancellableStatusBar.show(execution, cancellationTokenSource);
     taskViewService.addCommandExecution(execution, cancellationTokenSource);
   }
+
+  public mkdirSyncRecursive(dirPath: string) {
```
vazexqi: Why can't we use `mkdir -p` from shelljs?

guw: I introduced this code before introducing `shelljs` in the tests. This would make `shelljs` a runtime dependency. Right now it's only a dev dependency and used in tests only.

You ok with adding `shelljs` as runtime dep?

DatGuyJonathan: `shelljs` is a runtime dep in sobjects-faux-generator package. @Sweetman is also adding it as a runtime dep in https://github.com/forcedotcom/salesforcedx-vscode/pull/310. There's a common need for it so I think it should be okay.

vazexqi: Yeah. We already use it in https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-sobjects-faux-generator/package.json#L16 as well for its `mkdir -p` functionality. I would rather rely on it then write our own function which requires testing on all OS.

guw: I'll update the code then to use `shelljs`. Thanks!

PR #319, opened 2018-02-22
Access preference using scoped resource style

vazexqi:
### What does this PR do?

Removes error of the form 

```
[Extension Host] [salesforce.salesforcedx-vscode-lwc] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'eslint.nodePath';, provide the URI of a resource or 'null' for any resource. (at e._validateConfigurationAccess (/Applications/Visual Studio Code - Insiders.app/Contents/Resources/app/out/vs/workbench/node/extensionHostProcess.js:511:635))
```

@W-4650780@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/index.ts
+++ b/packages/salesforcedx-vscode-lwc/src/index.ts
@@ -73,7 +73,10 @@ export async function activate(context: vscode.ExtensionContext) {
   if (workspaceType === lwcLanguageServer.WorkspaceType.SFDX) {
     await populateEslintSettingIfNecessary(
       context,
-      vscode.workspace.getConfiguration()
+      vscode.workspace.getConfiguration(
+        '',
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -73,7 +73,10 @@ export async function activate(context: vscode.ExtensionContext) {
   if (workspaceType === lwcLanguageServer.WorkspaceType.SFDX) {
     await populateEslintSettingIfNecessary(
       context,
-      vscode.workspace.getConfiguration()
+      vscode.workspace.getConfiguration(
+        '',
+        vscode.workspace.workspaceFolders[0].uri
```
vazexqi: Usually on Windows we use `uri.fsPath` for the LSP but here we just use `uri` since it's only affecting VS Code and not any LSP.

PR #321, opened 2018-02-24
Handle spaces in file path

DatGuyJonathan:
### What does this PR do?
spaceinwindowspath

@W-4728610@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/core/logContextUtil.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/core/logContextUtil.ts
@@ -10,7 +10,7 @@ import * as fs from 'fs';
 export class LogContextUtil {
   public readLogFile(logFilePath: string): string[] {
     try {
-      return fs.readFileSync(logFilePath, 'utf-8').split('\n');
+      return fs.readFileSync(logFilePath, 'utf-8').split(/\r?\n/);
     } catch (e) {
       return [];
     }
--- a/packages/salesforcedx-vscode-replay-debugger/src/core/logContextUtil.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/core/logContextUtil.ts
@@ -6,13 +6,11 @@
  */
 
 import * as fs from 'fs';
-import { EOL } from 'os';
 
 export class LogContextUtil {
   public readLogFile(logFilePath: string): string[] {
     try {
-      const fileContent = fs.readFileSync(logFilePath).toString('utf-8');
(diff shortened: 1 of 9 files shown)


packages/salesforcedx-vscode-replay-debugger/src/core/logContextUtil.ts
```diff
@@ -6,13 +6,11 @@
...
 
 import * as fs from 'fs';
-import { EOL } from 'os';
 
 export class LogContextUtil {
   public readLogFile(logFilePath: string): string[] {
     try {
-      const fileContent = fs.readFileSync(logFilePath).toString('utf-8');
-      return fileContent.split(EOL);
+      return fs.readFileSync(logFilePath, 'utf-8').split('\n');
```
DatGuyJonathan: When I downloaded a debug log to my Windows, the line breaks were actually `\n` instead of the expected `\r\n`. The latter is what is provided by `os.EOL`, but it wasn't working due to this anomaly. 

unexpectedeolonwindows

I think hardcoding `\n` here is safe for other platforms. The alternative is https://www.npmjs.com/package/eol and it would cost more in memory:

`import * as eol from 'eol';`
`eol.split(eol.auto(fs.readFileSync(logFilePath, 'utf-8')));`

vazexqi: I think that's expected since our servers are the ones that generate the log file and the servers are running on *nix. The OS of the user is not taken into account when we generate the logs.

However, how about using a safer regex to split like `"a\nb\r\nc".split(/\r?\n/)` from https://stackoverflow.com/a/21895354 

This will take care of both *nix and Windows. It doesn't take care of classic Mac which used only `\r` but I don't think that is relevant anymore.

Also, this handles the case in case someone were to edit the file by hand (on a Windows machine) which would actually put \r\n in one some lines.

packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
```diff
@@ -26,7 +26,7 @@ export class FrameEntryState extends FrameState implements DebugLogState {
           logContext.getFrames().length,
           this._frameName,
           sourceUri
-            ? new Source(basename(sourceUri), Uri.parse(sourceUri).path)
+            ? new Source(basename(sourceUri), Uri.parse(sourceUri).fsPath)
```
vazexqi: This is probably the key. Using `fsPath` will take care of the encoding for you from the VS Code side.

PR #322, opened 2018-02-25
Store .vsix as part of appveyor CI

vazexqi:
### What does this PR do?

Stores the .vsix as part of the artifact for debugging purposes.

@W-4700676@

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -66,7 +66,14 @@ test_script:
         exit $NpmTestsExitCode
       }
   - npm run coverage:system-tests
+
+after_test:
   - npm run vscode:package
+
+artifacts:
+  - path: '**\*.vsix'
+
+
 
 on_finish:
   - del C:\projects\devhub.key
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -73,8 +73,6 @@ after_test:
 artifacts:
   - path: '**\*.vsix'
 
-
-
 on_finish:
   - del C:\projects\devhub.key
   - codecov --disable=gcov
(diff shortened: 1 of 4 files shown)


vazexqi:
There seems to be a build issue that just surfaced. It looks to be related to https://github.com/npm/npm/issues/19877

vazexqi:
I followed what they have done at https://github.com/angular/angular-cli/pull/9713 to fix this issue temporarily - add the dependency to a specific version of `ajv` at the top-level.

vazexqi:
@guw - Artifacts are now stored and available at https://circleci.com/gh/forcedotcom/salesforcedx-vscode/218#artifacts/containers/0

@ruthemmanuelle - Just FYI, this is also a way for you to generate the .vsix yourself for installation. Just open a PR against any branch and wait for the Circle CI just to finish (no need to wait for Travis or AppVeyor) and you should be able to get the .vsix that you want.

vazexqi:
As part of this work item, I realized that we don't have a proper .vscodeignore for the salesforcedx-vscode-apex-debugger and salesforcedx-vscode-lightning. This is why the size of the .vsix are larger as seen in https://ci.appveyor.com/project/guw/salesforcedx-vscode/build/develop-2008/artifacts

This happens because as part of the CI, we run the tests which will bring .vscode-test which is a binary file. This doesn't happen when we do the publishing because we use a different run (more similar to what we do in CircleCI today) which doesn't bring in the .vscode-test.

However, we can avoid this size expansion issue altogether by providing a .vscodeignore. I'll do that as part of another PR.

PR #327, opened 2018-02-28
Support line breakpoints in Apex switch statements

vazexqi:
### What does this PR do?

Add initial support for Apex Switch Statements. We are starting off first by ensuring that line breakpoints work in the interactive debugger.

@W-4677434@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/test/integration/config/simpleObjectAndField/objects/MyCustomObject2__c/MyCustomObject2__c.object-meta.xml
+++ b/packages/salesforcedx-sobjects-faux-generator/test/integration/config/simpleObjectAndField/objects/MyCustomObject2__c/MyCustomObject2__c.object-meta.xml
@@ -52,7 +52,6 @@
     <enableSearch>false</enableSearch>
     <enableSharing>true</enableSharing>
     <enableStreamingApi>true</enableStreamingApi>
-    <externalSharingModel>ReadWrite</externalSharingModel>
     <label>MyCustomObject</label>
     <nameField>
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/syntaxes/apex.tmLanguage
```diff
@@ -685,7 +685,7 @@
                </dict>
                <dict>
                   <key>match</key>
-                  <string>\b(return|break|case|continue|default|do|while|for|switch|if|else)\b</string>
+                  <string>\b(return|break|case|continue|default|do|while|for|switch|when|if|else)\b</string>
```
vazexqi: The only syntax highlighting we need is to add the `when` clause.

packages/salesforcedx-sobjects-faux-generator/test/integration/config/simpleObjectAndField/objects/MyCustomObject2__c/MyCustomObject2__c.object-meta.xml
```diff
@@ -52,7 +52,6 @@
     <enableSearch>false</enableSearch>
     <enableSharing>true</enableSharing>
     <enableStreamingApi>true</enableStreamingApi>
-    <externalSharingModel>ReadWrite</externalSharingModel>
```
vazexqi: @GaryFNC - I'm removing this since the externalSharingModel is now being turned off for all scratch orgs. This means that we can't deploy a custom object if we have this set here.

PR #332, opened 2018-03-05
Peg language server versions

vazexqi:
### What does this PR do?
 
Pegs the server versions to avoid https://github.com/Microsoft/vscode-languageserver-node/issues/317

We have a special dependency on version 3.4.x for the visualforce branch because of our use of the proposed color provider. 

The rest have been updated to v3.5.x

### What issues does this PR fix or reference?

Fixes the build.

PR #333, opened 2018-03-06
[WIP] Do not bundle all the vscode-languageserver/clients/types/protocol

vazexqi:
### What does this PR do?

A fix has been made upstream from Microsoft that will rectify #332. We don't need to bundle all the versions of it, which would help make our .vsix leaner.

Build issues.


vazexqi:
Build looks good. I'm going to merge this in since it is a leaner fix than the change I made in #332 yesterday.

PR #334, opened 2018-03-06
Add checkpoint, client side only

JimSuplizio:
### What does this PR do?
This is Checkpoint, client side, work. This adds the ability to create checkpoints, piggybacking off of the conditional breakpoint events. When processing a breakpoint, if the conditional breakpoint's condition contains the word 'checkpoint' then create a Checkpoint. Note, the Checkpoint service cannot be directly accessed from within ApexReplayDebug and had to be scheduled.

The CheckpointService processes this event but creating a CheckpointNode that stores off the values that are not user editable or displayed and creates a CheckpointInfoNode's as children for the ones that are. The separate subclasses of CheckpointInfoNode were created for the fields that are going to be user editable and are require custom verification when changed (when that feature is done). 

Also added: When called from the main CheckpointNode createJsonStringForApexExecutionOverlayAction will create a JSON string of the arguments required to be passed over to create ApexExecutionOverlayActionObject.

@W-4719058@

This is what the Checkpoint will look like in the UI
checkpointssnapshot


JimSuplizio:
I'd had to do a second push. There were a couple of AppVeyor failures due to TSLint that required removal of a couple of imports in two files.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -43,6 +79,17 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
   public update(): void {
     this._onDidChangeTreeData.fire();
```
DatGuyJonathan: You can remove the `update` method if you're not using it. I originally put it there to give you a hint on how to trigger the UI update.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -281,6 +289,23 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
           );
         }
+        // If there is a condition and that condition is a checkpoint then
+        // sent CHECKPOINT_INFO_EVENT with the args.source, line and uri
+        if (args.breakpoints[i].condition) {
+          if (
+            args.breakpoints[i].condition!.toLowerCase().indexOf(CHECKPOINT) >=
+            0
+          ) {
+            // JRS
```
DatGuyJonathan: What's this comment for?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -53,13 +100,152 @@ export abstract class BaseNode extends TreeItem {
...
+  ) {
+    super(
+      sourceInput.name + ':' + lineInput,
+      TreeItemCollapsibleState.Expanded
+    );
+
+    this.line = lineInput;
+    this.source = sourceInput;
+    this.uri = uriInput;
+    const sourceNameSplit = this.source.name!.split('.');
```
DatGuyJonathan: I'm confused why you're using DebugProtocol.Source.Name for ExecutableEntityName. It's not a typeref which is why you have to prepend `__sfdc_trigger` for the trigger case. It also doesn't contain the namespace if the org is namespaced. The reason why we made ExecutableEntityName a typeref was because apexReplayDebug already has a mapping of URIs and typerefs, so you should use the info from that. That typeref is already in the correct format for ExecutableEntityName.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -53,13 +100,152 @@ export abstract class BaseNode extends TreeItem {
...
+      ActionScriptType.None
+    );
+    this.children.push(cpASTNode);
+    const cpScriptNode = new CheckpointInfoActionScriptNode('');
+    this.children.push(cpScriptNode);
+    const cpIterationNode = new CheckpointInfoIterationNode(1);
+    this.children.push(cpIterationNode);
+  }
+
+  public createJsonStringForApexExecutionOverlayAction(): string {
```
DatGuyJonathan: I'm thinking you can have a type for ApexExecutionOverlayAction with fields like ExecutableEntityName, ActionScript, etc. Then you can create an instance of that object with the values that you have. With that, you can do `JSON.stringify(theCheckpointObject)` to get the JSON string instead of this string builder.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -53,13 +100,152 @@ export abstract class BaseNode extends TreeItem {
...
+// changing the values for and are of different value types:
+// CheckpointInfoActionScriptNode: holds the action script (string)
+// CheckpointInfoActionScriptTypeNode: holds the type of action script (ActionScriptType enum)
+// CheckpointInfoIterationsNode: holds the number of iterations (integer)
+// The reason for creating these would be to add specific validations when
+// the fields are made editable.
+export class CheckpointInfoActionScriptNode extends CheckpointInfoNode {
+  public actionScript: string;
+  constructor(actionScriptInput: string) {
+    super('Script: ' + actionScriptInput);
```
DatGuyJonathan: Can these labels be externalized?

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -51,6 +55,21 @@ function registerDebugHandlers(): vscode.Disposable {
...
+            eventBody.source &&
+            eventBody.line &&
+            eventBody.uri
+          ) {
+            checkpointService.addCheckpointNode(
+              eventBody.source,
+              eventBody.line,
+              eventBody.uri
+            );
+            // JRS Log something here?
```
DatGuyJonathan: What logging did you have in mind?

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -258,15 +261,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
         `setBreakPointsRequest: path=${args.source
           .path} lines=${args.lines.join(',')}`
       );
       const uri = this.convertClientPathToDebugger(args.source.path);
       this.breakpoints.set(uri, []);
-      for (const lineArg of args.lines) {
+      // While processing the breakpoints, if there's a conditional breakpoint
+      // then create a checkpoint. This requires checking the breakpoint condition
+      // for the appropriate args.lines item being processed.
+      let i: number;
```
DatGuyJonathan: This can be included in the for statement: `for (let i = 0; ...)`

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -5,7 +5,43 @@
...
+  APEX_EXECUTION_OVERLAY_ACTION_EXECUTABLE_ENTITY_NAME,
+  APEX_EXECUTION_OVERLAY_ACTION_IS_DUMPING_HEAP,
+  APEX_EXECUTION_OVERLAY_ACTION_ITERATION,
+  APEX_EXECUTION_OVERLAY_ACTION_LINE,
+  SFDC_TRIGGER
+} from '../constants';
+
+import { CheckpointUtil } from './checkpointUtil';
+
+import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
```
DatGuyJonathan: The ordering of the imports seems off. Tslint usually would've complained to move `from 'vscode-debugprotocol/lib/debugProtocol'` to be below `from 'vscode'`. Is it because you have a new line in between imports? Other files don't do that.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointUtil.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export class CheckpointUtil {
+  // If the valueInput is a string then add quoutes to both inputs
+  // "<nameInput>": "<valueInput>"
+  // If the valueInput is a number or boolean then only add quotes to the name
+  // "<nameInput>": <valueInput>
+  public static formatNameValuePairForJSon(
```
DatGuyJonathan: This seems more complicated than necessary. See my earlier comment about using `JSON.stringify()` which should take care of the quotes for you.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done. It'll be gone when I push the update based upon feedback.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: Gone, it was just a marker for my searching.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Look at the source files on the disk for you project. These have to be in a specific format <ClassName>.cls for classes and <TriggerName>.trigger. These file names are the typeRef of the outermost type. source.name is the filename without the path, splitting off the extension leaves you with the typeRef name.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: These shouldn't be needed outside of checkpointService.ts, I can make them constants in there, which will at least make them defined in one place and easier to find.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Yeah, you're right, it was the newline.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
DatGuyJonathan: The typeref in apexReplayDebug is from the apex language server which is our source of truth for typerefs. Are you saying you want this other thing here to figure typerefs? How are you going to handle namespace?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: I can probably do that, but I'm still going to have to go through an piece together the object. The CheckpointNode contains the pieces that aren't editable by the user. The list of children are the CheckpointInfoNode variants which contain the editable pieces. I'm still gong to need the createJsonStringForApexExecutionOverlayAction or some function to call that'll do this.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Ignore my previous comment, I'm going to be using the typeref mapping in the breakpoint util.

JimSuplizio: Yes, I made a mistake here. I'm going to be using the typerefMapping on the breakpointUtils to get the top level typeref.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -5,7 +5,44 @@
...
 
-import { Event, EventEmitter, TreeDataProvider, TreeItem } from 'vscode';
+import {
+  Event,
+  EventEmitter,
+  TreeDataProvider,
+  TreeItem,
+  TreeItemCollapsibleState
+} from 'vscode';
+import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
```
DatGuyJonathan: This can be removed.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -281,6 +288,23 @@ export class ApexReplayDebug extends LoggingDebugSession {
             this.convertClientLineToDebugger(lineArg)
           );
         }
+        // If there is a condition and that condition is a checkpoint then
+        // sent CHECKPOINT_INFO_EVENT with the args.source, line and uri
+        if (args.breakpoints[i].condition) {
```
DatGuyJonathan: I just remember - search for `supportsConditionalBreakpoints` in this file and flip it to true.

DatGuyJonathan:
@JimSuplizio , in my sample https://github.com/forcedotcom/salesforcedx-vscode/commit/59a18b7c35587fff1a853746081c10c8c397d28d, the adapter tests show up as a move because they were deleted from `salesforcedx-vscode-replay-debugger/test` and added to `salesforcedx-vscode-replay-debugger/adaptertest`. Your commit only did the former, so your branch has both adapter and extension tests in `salesforcedx-vscode-replay-debugger/test`. Could you fix that?

PR #335, opened 2018-03-06
Add demo mode for VS Code

vazexqi:
### What does this PR do?

Supports demo mode from the CLI. If we detect that SFDX_ENV is set to DEMO we activate demo mode, which does two things:

1. Shows the status bar item at the bottom right corner.
2. If the user authenticates against a non-trial org, we warn the user and ask if she wants to proceed.

@W-4714447@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/test/testrunner.ts
+++ b/packages/salesforcedx-utils-vscode/src/test/testrunner.ts
@@ -78,7 +78,6 @@ function run(testsRoot: any, clb: any): any {
     try {
       // Fill into Mocha
       files.forEach((f): Mocha => {
-        console.log(`Adding test script: ${f}`);
         return mocha.addFile(paths.join(testsRoot, f));
       });
       // Run the tests
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -291,7 +291,7 @@ export class SelectUsername
 
 export class DemoModePromptGatherer implements ParametersGatherer<{}> {
   private readonly LOGOUT_RESPONSE = 'Cancel';
-  private readonly DO_NOT_LOGOUT_RESPONSE = 'Continue';
+  private readonly DO_NOT_LOGOUT_RESPONSE = 'Authorize Org';
   private readonly prompt: string;
 
   public constructor() {
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -289,6 +289,28 @@ export class SelectUsername
   }
 }
 
+export class DemoModePromptGatherer implements ParametersGatherer<{}> {
+  private readonly LOGOUT_RESPONSE = 'Cancel';
+  private readonly DO_NOT_LOGOUT_RESPONSE = 'Continue';
+  private readonly prompt: string;
+
+  public constructor() {
+    this.prompt = nls.localize('demo_mode_prompt');
+  }
+
+  public async gather(): Promise<CancelResponse | ContinueResponse<{}>> {
+    const response = await vscode.window.showInformationMessage(
+      this.prompt,
+      this.DO_NOT_LOGOUT_RESPONSE,
+      this.LOGOUT_RESPONSE
+    );
+
+    return response && response === this.LOGOUT_RESPONSE
+      ? { type: 'CONTINUE', data: {} }
+      : { type: 'CANCEL' };
+  }
+}
+
 // Command Execution
 ////////////////////
 export interface CommandletExecutor<T> {
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -317,13 +317,19 @@ export interface CommandletExecutor<T> {
 
 export abstract class SfdxCommandletExecutor<T>
   implements CommandletExecutor<T> {
+  protected showChannelOutput = true;
+
   protected attachExecution(
     execution: CommandExecution,
     cancellationTokenSource: vscode.CancellationTokenSource,
     cancellationToken: vscode.CancellationToken
   ) {
     channelService.streamCommandOutput(execution);
-    channelService.showChannelOutput();
+
+    if (this.showChannelOutput) {
+      channelService.showChannelOutput();
+    }
+
     notificationService.reportCommandExecutionStatus(
       execution,
       cancellationToken
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -292,11 +292,7 @@ export class SelectUsername
 export class DemoModePromptGatherer implements ParametersGatherer<{}> {
   private readonly LOGOUT_RESPONSE = 'Cancel';
   private readonly DO_NOT_LOGOUT_RESPONSE = 'Authorize Org';
-  private readonly prompt: string;
-
-  public constructor() {
-    this.prompt = nls.localize('demo_mode_prompt');
-  }
+  private readonly prompt = nls.localize('demo_mode_prompt');
 
   public async gather(): Promise<CancelResponse | ContinueResponse<{}>> {
     const response = await vscode.window.showInformationMessage(
--- a/packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
@@ -18,6 +18,12 @@ import {
 } from './commands';
 
 export class ForceAuthLogoutAll extends SfdxCommandletExecutor<{}> {
+  public static withoutShowingChannel(): ForceAuthLogoutAll {
+    const instance = new ForceAuthLogoutAll();
+    instance.showChannelOutput = false;
+    return instance;
+  }
+
   public build(data: {}): Command {
     return new SfdxCommandBuilder()
       .withDescription(nls.localize('force_auth_logout_all_text'))
--- a/packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import {
+  Command,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import { nls } from '../messages';
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+export class ForceAuthLogoutAll extends SfdxCommandletExecutor<{}> {
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_auth_logout_all_text'))
+      .withArg('force:auth:logout')
+      .withArg('--all')
+      .withArg('--noprompt')
+      .build();
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+const executor = new ForceAuthLogoutAll();
+const commandlet = new SfdxCommandlet(
+  workspaceChecker,
+  parameterGatherer,
+  executor
+);
+
+export function forceAuthLogoutAll() {
+  commandlet.run();
+}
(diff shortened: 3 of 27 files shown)


vazexqi:
Tests are coming in the next commit. Want to make sure the user-facing text is correct first.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -155,5 +155,13 @@ export const messages = {
...
-  error_updating_sfdx_project: 'Error updating sfdx-project.json: %s'
+  error_updating_sfdx_project: 'Error updating sfdx-project.json: %s',
+
+  demo_mode_status_text: `$(gist-secret) SFDX DEMO`,
+  demo_mode_status_tooltip:
+    'You are running the SFDX VS Code extensions in demo mode. You will be prompted if you are connecting to production orgs.',
+  demo_mode_prompt:
+    'Authenticating against a business or production org is not recommended on a demo or shared machine. Continue authenticating or cancel?',
+
+  force_auth_logout_all_text: 'SFDX: Log Out from All Authorized Orgs'
```
vazexqi: screen shot 2018-03-06 at 3 32 49 pm

untitled


CC: @ruthemmanuelle

packages/salesforcedx-vscode-core/src/decorators/index.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export { showDemoMode } from './demo-mode-decorator';
```
vazexqi: @Sweetman - I organized all the different decorators under the new decorators folder.

packages/salesforcedx-vscode-core/src/modes/demo-mode.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export function isDemoMode() {
+  return process.env.SFDX_ENV ? process.env.SFDX_ENV === 'DEMO' : false;
```
vazexqi: @amphro - I've verified this against the release branch of the SFDX CLI that `DEMO` is what we are checking for.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
vazexqi: By the way, this style of prompt is in VS Code Insiders. It's not in the normal release yet but will come out soon in early March. Just wanted to use this since it is what will be available to users soon.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -155,5 +155,13 @@ export const messages = {
   error_creating_packagexml: 'Error creating package.xml. %s',
   error_extracting_org_source: 'Error extracting downloaded Apex source. %s',
   error_extracting_packages: 'Error extracting packages: %s',
-  error_updating_sfdx_project: 'Error updating sfdx-project.json: %s'
+  error_updating_sfdx_project: 'Error updating sfdx-project.json: %s',
+
+  demo_mode_status_text: `$(gist-secret) SFDX DEMO`,
+  demo_mode_status_tooltip:
+    'You are running the SFDX VS Code extensions in demo mode. You will be prompted if you are connecting to production orgs.',
```
ruthemmanuelle: 'You are running Salesforce Extensions for VS Code in demo mode. You will be prompted for confirmation when connecting to production orgs.'

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -155,5 +155,13 @@ export const messages = {
...
+  demo_mode_prompt:
+    'Authenticating against a business or production org is not recommended on a demo or shared machine. Continue authenticating or cancel?',
```
ruthemmanuelle: Can we control the button labels? If so, can we use "Authorize Org" and "Cancel" instead of "Continue" and "Cancel"? We try to make the button labels as specific as we can, especially with dangerous actions, because many users won't read the confirmation message.

My suggestion for this message (unless there's a better way to recommend logging out):

'Authorizing a business or production org is not recommended on a demo or shared machine. If you continue with the authentication, be sure to run "sfdx force:auth:logout --targetusername <your username> --noprompt" when you're done using the org. Are you sure you want to authorize this org?'

vazexqi: @ruthemmanuelle Yes, I can control the labels. I can use what you suggested for the button labels.

The resultant message is a bit long but it should be fine since this is not a message that they see normally. So, fortunately, this new notification UI for VS Code should handle it. It would have been a big readability issue with the old notification scheme.

screen shot 2018-03-06 at 4 07 58 pm

packages/salesforcedx-vscode-core/src/messages/i18n.ts
vazexqi: @ruthemmanuelle - Updated text.

screen shot 2018-03-06 at 4 10 00 pm

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: After I posted my comment I saw the VS Code command that you added to log out of all orgs, so I edited my comment to use the VS Code version. You're too quick on the draw for me. : )

I think that with the more descriptive button label we can probably eliminate the last sentence, to shorten up the message. What do you think of this?

'Authorizing a business or production org is not recommended on a demo or shared machine. If you continue with the authentication, be sure to run "SFDX: Log Out from All Authorized Orgs" when you're done using this org.'

ruthemmanuelle: @vazexqi - @mentioning you to make sure you see this, since this comment doesn't show up in the main feed after your most recent commit.

vazexqi: screen shot 2018-03-06 at 4 22 10 pm

@ruthemmanuelle - Sounds good to me. I have to lift the 'SFDX: Log Out from All Authorized Orgs' up to a top-level command, but that should be fine. This command now shows up even if you are not in demo mode since I think it provides value for non-demo scenarios.

packages/salesforcedx-utils-vscode/src/test/testrunner.ts
```diff
@@ -78,7 +78,6 @@ function run(testsRoot: any, clb: any): any {
     try {
       // Fill into Mocha
       files.forEach((f): Mocha => {
-        console.log(`Adding test script: ${f}`);
```
vazexqi: Remove logging that @guw added. I don't think we need these and there were cluttering up the results when running locally.

packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
```diff
@@ -0,0 +1,42 @@
...
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import { nls } from '../messages';
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+export class ForceAuthLogoutAll extends SfdxCommandletExecutor<{}> {
```
vazexqi: Add a new "SFDX: Log Out from All Authorized Orgs" command.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -29,15 +41,68 @@ class ForceAuthWebLoginExecutor extends SfdxCommandletExecutor<{}> {
   }
 }
 
+export class ForceAuthWebDemoModeLoginExecutor extends SfdxCommandletExecutor<{}> {
```
vazexqi: This is the special one that we are using in demo mode. When we use this particular implementation, we will detect if the org that the user has authorized is a prod org and prompt the user appropriately.

packages/salesforcedx-vscode-core/test/commands/commands.test.ts
```diff
@@ -231,7 +232,6 @@ describe('Command Utilities', () => {
         const dirList: string[] = dirPathGatherer.globDirs(
           vscode.workspace.rootPath
         );
-        console.log(`Found dirs: ${dirList}`);
```
vazexqi: Remove logging that @guw added. I don't think we need these and there were cluttering up the results when running locally.

packages/salesforcedx-vscode-core/test/commands/commands.test.ts
```diff
@@ -505,4 +505,36 @@ describe('Command Utilities', () => {
       });
     });
   });
+
+  // Due to the way the prompt is phrased
+  // CONTINUE means that we will execute the forceLogoutAll command
+  // CANCEL means that we will not execute the forceLogoutAll command
+  describe('DemoModePrompGatherer', () => {
```
vazexqi: Care must be taken to interpret the prompt so I've added some commands in addition to the tests.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -1,3 +1,10 @@
+/*
```
vazexqi: Some files were missing copyrights so I've added them as well.

packages/salesforcedx-vscode-core/test/commands/forceProjectCreate.test.ts
```diff
@@ -1,3 +1,10 @@
+/*
```
vazexqi: Some files were missing copyrights so I've added them as well.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -29,14 +41,72 @@ class ForceAuthWebLoginExecutor extends SfdxCommandletExecutor<{}> {
...
+    );
+
+    channelService.streamCommandOutput(execution);
+    CancellableStatusBar.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+
+    const cmdOutput = new CommandOutput();
+    const result = await cmdOutput.getCmdResult(execution);
+
+    if (isTrialOrg(JSON.parse(result))) {
```
Sweetman: Reading this is kind of confusing since it reads like you're prompting for log out if it's a trial org but isTrialOrg returns false if it is a trial org.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -379,6 +385,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
 export function deactivate(): Promise<void> {
   console.log('SFDX CLI Extension Deactivated');
-  disposeTraceFlagExpiration();
+
+  decorators.disposeTraceFlagExpiration();
+
+  if (isDemoMode()) {
+    forceAuthLogoutAll();
```
Sweetman: I think you might need to put this and restoreDebugLevels in a Promise.all and return that promise. The deactivate method will wait 4 seconds or until the returned promise is completed so forceAuthLogoutAll might not be allowed to complete if restoreDebugLevels finishes before.

https://github.com/Microsoft/vscode/issues/11895

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
vazexqi: Yup, I was tripped out by it too. I renamed it to isProdOrg instead in my latest commit.

packages/salesforcedx-vscode-core/src/index.ts
vazexqi: Thanks. I'm thinking that It's likely that I am going to remove this. The 4 seconds is too finicky. 

With demo mode the user is also expected to use a guest account on the OS, which will wipe out all data after that. That is a safer guarantee. This 4 second wait is not very reliable.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -299,13 +317,19 @@ export interface CommandletExecutor<T> {
...
+
   protected attachExecution(
     execution: CommandExecution,
     cancellationTokenSource: vscode.CancellationTokenSource,
     cancellationToken: vscode.CancellationToken
   ) {
     channelService.streamCommandOutput(execution);
-    channelService.showChannelOutput();
+
+    if (this.showChannelOutput) {
```
vazexqi: @Sweetman - Per our discussion the other day, this is one way we can hide the output channel if it is not necessary. I've made the default to show it always, which is how it is today so there will be no observable change of behavior.

packages/salesforcedx-vscode-core/src/commands/forceAuthLogout.ts
```diff
@@ -0,0 +1,48 @@
...
+  public static withoutShowingChannel(): ForceAuthLogoutAll {
```
vazexqi: This is how I would instantiate a special command that doesn't show the output channel unnecessarily. I don't like this particular way of instantiating it. I would rather use some generic factory method but there seems to be issues with the type system of TypeScript to support that. I need to think about this  more but the issue is at https://github.com/Microsoft/TypeScript/issues/5863

PR #336, opened 2018-03-07
[WIP, DO-NOT-MERGE] Show how to access breakpoints before a debug session

vazexqi:
### What does this PR do?

Show how to access breakpoints before a session using the new debug API.

Nothing, not for merging. Just showing the use of the latest debug API in VS Code.

Based on https://github.com/Microsoft/vscode/issues/23188 and https://github.com/Microsoft/vscode/issues/43492

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-replay-debugger/package.json
@@ -21,6 +21,7 @@
     "vscode": "^1.19.0"
   },
   "categories": ["Debuggers"],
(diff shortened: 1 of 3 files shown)


vazexqi:
This shouldn't be necessary anymore. VS code https://code.visualstudio.com/updates/v1_21 is available and it should have the debug apis ready in the stable version.

If we bump up the dependency in package.json for that version, we should be able to use the latest api.

CC: @DatGuyJonathan, @JimSuplizio

PR #339, opened 2018-03-08
Run npm lint before push

vazexqi:
### What does this PR do?

I've noticed that a lot of times (myself included) that we have pushed to GitHub without first running `npm run lint`. Instead of trying to force everyone to remember to do so, I've automated it using [husky](https://www.npmjs.com/package/husky).

So, now when you do a `npm install` it will set up everything for you.

Now, when you do a `git push` it will run lint first for you.
If for some reason you want to bypass this, you can always do `git push --no-verify` to skip all the hooks.

Code health.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,9 @@
 {
   "devDependencies": {
+    "ajv": "6.1.1",
     "eslint": "4.13.1",
     "eslint-config-prettier": "2.9.0",
+    "husky": "^0.14.3",
     "istanbul": "1.1.0-alpha.1",
     "lerna": "2.4.0",
     "prettier": "1.6.0",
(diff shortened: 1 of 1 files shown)


vazexqi:
@ruthemmanuelle - I don't think this will affect your workflow but if it does let me know. You can also bypass this using `git push --no-verify`.

package.json
```diff
@@ -1,15 +1,16 @@
 {
   "devDependencies": {
+    "ajv": "6.1.1",
     "eslint": "4.13.1",
     "eslint-config-prettier": "2.9.0",
+    "husky": "^0.14.3",
```
lcampos: Whenever they release version 0.15 (release candidate right now) the way the script hooks are configured will change (https://github.com/typicode/husky#upgrading-from-014). Should we just explicitly stick to 0.14.3 ?

vazexqi: That is a good catch. I read about it but forgot. Yeah, let's lock this down to 0.14.3 for now and then when 0.15 is released we can use the new syntax. I actually prefer the new syntax since it doesn't clutter up the `scripts` section.

PR #340, opened 2018-03-08
Build vscode packages on CircleCI nightly regardless of commits

vazexqi:
### What does this PR do?

There were two occasions where an upstream changed not related to our changes broke the `vscode:package` step. This was in #322, #332, and #333. When it happened, we basically only found out about it a few days later and had to quickly fix it.

It would be good to have a nightly build on Circle CI that executes that step so that we can find out about any errors earlier.

I've chosen to use CircleCI since right now, it's very lean. It's only running the build of `vscode:package`. Also, CircleCI has been configured to e-mail me (anyone else can sign up) so that you get notifications when the build passes/fails.

Build stability.


.appveyor.yml
```diff
@@ -67,9 +67,6 @@ test_script:
       }
   - npm run coverage:system-tests
 
-after_test:
```
vazexqi: Let's save some cycles off AppVeyor since we are also doing this on CircleCI for now.

.circleci/config.yml
```diff
@@ -27,3 +27,14 @@ jobs:
...
           paths:
             - node_modules
+workflows:
+  version: 2
+  commit-workflow:
+    jobs:
+      - build
+  scheduled-workflow:
+    triggers:
+      - schedule:
```
lcampos: Should we add a filter to run on a specific branch or is that part of the config of the job already ?

vazexqi: @lcampos - Good question. It's for all branches now. The syntax allows us to limit it. See https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow

I'll limit it to develop (and master since we might need to do e-releases from there).

PR #343, opened 2018-03-13
Update .travis.yml

vazexqi:
### What does this PR do?

Use the new sfdx-cli installers for linux.


vazexqi:
We won't use this way anymore. We will use `npm install -g sfdx-cli` instead as done in https://github.com/forcedotcom/LightningTestingService/pull/79

PR #344, opened 2018-03-13
[Replay Debugger] Show variables w/o nesting

Sweetman:
### What does this PR do?
Add variables for primitives for the replay debugger
@W-4801201@


packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -59,6 +61,45 @@ export interface LaunchRequestArguments
   trace?: boolean | string;
 }
 
+export class ApexVariable extends Variable {
+  public type: string;
```
Sweetman: Needed to expose the type attribute to show type on hover

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -59,6 +61,45 @@ export interface LaunchRequestArguments
...
+  }
+}
+
+export enum SCOPE_TYPES {
+  LOCAL = 'local',
+  STATIC = 'static',
+  GLOBAL = 'global'
+}
+
+export class ScopeContainer {
```
Sweetman: This class is not necessarily needed right now but will be when I implement objects/collections. Wanted to just keep it so it wouldn't throw me off.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -178,6 +219,75 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+    args: DebugProtocol.ScopesArguments
+  ): Promise<void> {
+    response.success = true;
+    const frameInfo = this.logContext.getFrameHandler().get(args.frameId);
+    if (!frameInfo) {
+      response.body = { scopes: [] };
+      this.sendResponse(response);
+      return;
+    }
+    const scopes = new Array<Scope>();
```
Sweetman: Every scope we push needs a unique ID because when we enter scopesRequest this ID is what we will use to retrieve the ScopeContainer. Gunnar used the Handles class provided by the debugger api so I followed this as well.

packages/salesforcedx-vscode-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -178,6 +219,75 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+    response.body = { scopes: scopes };
+    this.sendResponse(response);
+  }
+
+  protected async variablesRequest(
+    response: DebugProtocol.VariablesResponse,
+    args: DebugProtocol.VariablesArguments
+  ): Promise<void> {
+    response.success = true;
+    const scopesContainer = this.logContext
```
Sweetman: Use the ID assigned in the scope creation to retrieve the ScopeContainer which holds the variables information

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
```diff
@@ -44,6 +51,13 @@ export class LogContext {
   private readonly launchArgs: LaunchRequestArguments;
   private readonly logLines: string[] = [];
   private state: DebugLogState | undefined;
+  private frameHandles = new Handles<ApexDebugStackFrameInfo>();
```
Sweetman: Each of the handles are a map from a unique ID (needed for when you expand elements) to an object. This is needed for the UI to correctly retrieve information about whatever frame, scope, or variable is clicked on.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
```diff
@@ -44,6 +51,13 @@ export class LogContext {
...
+  private scopeHandles = new Handles<ScopeContainer>();
+  private staticVariablesClassMap = new Map<
```
Sweetman: Static variables should be shared amongst frames of the same class. This map will store a variable->variableName map to every class name/namespaced used.

packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
```diff
@@ -19,11 +20,25 @@ export class FrameEntryState extends FrameState implements DebugLogState {
 
   public handle(logContext: LogContext): boolean {
     const sourceUri = logContext.getUriFromSignature(this._signature);
+    const frame = new ApexDebugStackFrameInfo(
```
Sweetman: Assigned it a unique id and pushed that into the frame itself. This id is used to retrieve the stack frame info object where it can access the variables.

packages/salesforcedx-vscode-replay-debugger/src/states/frameEntryState.ts
```diff
@@ -19,11 +20,25 @@ export class FrameEntryState extends FrameState implements DebugLogState {
...
+      logContext.getFrames().length,
+      this._signature
+    );
+    const id = logContext.getFrameHandler().create(frame);
+    const className =
+      this._signature.indexOf('.') > -1
+        ? this._signature.substring(0, this._signature.lastIndexOf('.'))
+        : this._signature;
+    if (logContext.getStaticVariablesClassMap().has(className)) {
```
Sweetman: We get the className (and namespace if it has one) and checks to see if there was a previous frame with it. If so then we assign it the same static variables map. If not we create a new mapping.

packages/salesforcedx-vscode-replay-debugger/src/states/variableBeginState.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  constructor(fields: string[]) {
+    this.fields = fields;
+  }
+  public handle(logContext: LogContext): boolean {
+    const currFrame = logContext.getTopFrame();
+    if (currFrame) {
+      const id = currFrame.id;
+      const frameInfo = logContext.getFrameHandler().get(id);
+      const name = this.fields[3];
```
Sweetman: In the case of a static variable this will be "ClassName.VariableName".
In the case of a local variable this will be "variableName"

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.fields = fields;
+  }
+  public handle(logContext: LogContext): boolean {
+    const currFrame = logContext.getTopFrame();
+    if (currFrame) {
+      const id = currFrame.id;
+      const frameInfo = logContext.getFrameHandler().get(id);
+      const name = this.fields[3];
+      const nameSplit = this.fields[3].split('.');
+      const varName =
```
Sweetman: On a begin scope for local variables, only the variable name is given. On an assignment, the classname and variable name are given. We stored the variable name as the key in the map so we need to retrieve just that piece to update the local variable.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -0,0 +1,43 @@
...
+          : name;
+      const varName =
+        nameSplit.length > 0 ? nameSplit[nameSplit.length - 1] : name;
+      const value = this.fields[4];
+      if (logContext.getStaticVariablesClassMap().has(className)) {
+        const statics = logContext.getStaticVariablesClassMap().get(className)!;
+        statics.get(name)!.value = value;
+      } else if (frameInfo.locals.has(varName)) {
+        const frameLocals = frameInfo.locals.get(varName) as ApexVariable;
+        frameLocals.value = value;
```
DatGuyJonathan: The interactive debugger needed a special handler so it can show the difference between:

String nullString;
String nullAsString = 'null';
String emptyString = '';

Is that also necessary here?

Sweetman: Doesn't look like it's necessary, I tested it out and here's what will be shown
image

PR #353, opened 2018-03-19
Add several fixes for Go To Definition & Find All References

GaryFNC:
For a class that implements its own inner interface, there was
a problem that caused an exception to be thrown and the class
was not fully processed showing up as failures to find references
or go to definition.

Also, a Go To Definition for an implicit constructor now goes to
the class definition instead of just showing no definition.

These are resolved by adding a new Apex Language Server jar file.


lcampos:
@GaryFNC I've triggered the appveyor tests once again since I wasn't able to find the root cause for the failures. Travis is currently failing cause system-tests are timing out.

GaryFNC:
fun with CI - thanks Luis

PR #359, opened 2018-03-30
Bump version for vscode debugger modules to use their uri to path fixes

DatGuyJonathan:
### What does this PR do?
This allows interactive debugger to work on Windows when there's spaces in the project's path

screen shot 2018-03-30 at 3 56 43 pm

@W-4709469@


vazexqi:
@DatGuyJonathan - LGTM. Could you see why all the CIs are failing though?

PR #365, opened 2018-04-05
Support addition/deletion of checkpoints

JimSuplizio:
### What does this PR do?
Checkpoints - part 3
- Moved all the checkpoint code out of the adapter and now everything is fully in the extension. The reason for this was the addition of vscode.debug.onDidChangeBreakpoints which allows processing of breakpoint events without an active debug session. This event is processed in the extension regardless of an active debug session.

- BreakpointUtil - Previously, the breakpoint info was only needed in the adapter, with the addition of breakpoint processing for checkpoints in the extension it needed a few changes. The class was turned into a singleton and line breakpoint info process was moved into the class, itself. The raw line breakpoint information is now stored on the class and the primary reason for this is because both the adapter and extension need it. The extension retrieves the line breakpoint information from the language server when it is activated, creates the mapping and stores the raw line breakpoint info. When the adapter has to setup its BreakpointUtil, it can get this from the extension instead of having to double tap the language server as getting the line breakpoint info could be potentially expensive for a large amount of Apex.

- BaseCommand and other commands under salesforcedx-apex-debugger - The CheckpointService is using the RequestService to manipulate ApexExecutionOverlayActionObjects on the AppServer but in order to do this, the BaseCommand had to be modified to remove the debugger specific items. The base command refactored into its basest form. The contructor only takes the optional queryString and the getCommandUrl and getRequest functions, both required by the RequestService are now abstract functions that need to be implemented by extending classes. A BaseDebuggerCommand was created that contains all the variables and provide the same functionality. The apex-debugger commands now all inherit from that.

- RequestService - The execute method was hard coded for POST only commands. This was changed to accept an enum for the REST command to execute. Post is still the default but if a RestHttpMethodEnum parameter is passed in it'll override it. Right now only DELETE and POST are implmented but this can be extended for other REST commands.

- The language client (salesforcedx-vscode-apex) - Index.ts was modified to export isLanguageClientReady. This was necessary because when BreakpointUtil, now being created in the extension process, can end up trying to call to get the line breakpoint information before it was ready. Added a handler for the languageClient.onReady event to set the languageClientReady boolean to true so we can ensure the language client is ready before requesting line breakpoint information.

- ApexExecutionOverlayActionCommand - command passed to the RequestService.execute for the addition or deletion of checkpoints on the AppServer.

- CheckpointService - When the extension's activate is called it makes a call to activateRequestService which grabs the org information for the projects and instantiates the RequestSerivce. This is how checkpoints are created/delete from the AppServer when they're created/delete from within VS Code. processBreakpointChangedForCheckpoints is the function setup to be the handler for vscode.debug.onDidChangeBreakpoints to deal with Checkpoints. Checkpoints are only added or removed through this function. 

- Additional tests for CheckpointService and new tests for ApexExecutionOverlayActionCommand. The commands going through the mocked RequestService were easier to test as commands than trying to test them as part of the CheckpointService. To this end the CheckpointService's add/remove checkpoint methods can be called with makeCommandCall set to false. Breaking these two apart for testing was far less convoluted than trying to do everything in just the CheckpointService.

- Additional updates par initial feedback. Added code to remove any existing checkpoints on the server during startup. 

- If the checkpoint fails to create when added, then give an error and remove it instead of leaving a defunct checkpoint in VS.

Here is a snapshot of what the error notification looks like in VS Code when the user tries to add more than 5 checkpoints.
5checkpoints

@W-4719064@ @W-4857885@ @W-4858717@


packages/salesforcedx-vscode-replay-debugger/adaptertest/unit/adapter/apexReplayDebug.test.ts
```diff
@@ -853,6 +775,7 @@ describe('Replay debugger adapter - unit', () => {
         setValidLines.restore();
       });
 
+      /*JRS-these can be removed
```
DatGuyJonathan: Why can these tests be removed? The code it was testing still exists in apexReplayDebug.ts. It was modified, so at the very least these tests should be modified. Otherwise, we’ve lost test coverage.

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -18,11 +18,13 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.19.0"
+    "vscode": "^1.21.0"
   },
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "42.10.0",
+    "@salesforce/salesforcedx-apex-debugger": "42.10.0",
```
DatGuyJonathan: Any estimate on how much work to move to utils and when? This dependency is **really** weird; I suppose I can look the other way for now if you really want to handle this in another PR…

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/breakpointUtil.ts
```diff
@@ -36,6 +50,36 @@ export class BreakpointUtil {
     );
   }
 
+  public getRawLineBPInfo(): any | undefined {
+    return this.rawLineBPInfo;
+  }
+
+  public createMappingsFromLineBreakpointInfo(lineBpInfo: any): void {
```
DatGuyJonathan: Are there tests for this?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/breakpointUtil.ts
```diff
@@ -36,6 +50,36 @@ export class BreakpointUtil {
...
+      }
+      this.lineNumberMapping.set(
+        info.uri,
+        this.lineNumberMapping.get(info.uri)!.concat(info.lines)
+      );
+      this.typerefMapping.set(info.typeref, info.uri);
+    }
+  }
+
+  public returnLinesForLoggingFromBreakpointArgs(
```
DatGuyJonathan: This iterates through each object, grabs lines numbers, and joins them all with commas. Can this be simplified to `bpArr.map(bp => bp.line).join(‘,’)`?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -41,13 +50,17 @@ export interface ApexExecutionOverlayAction {
...
   private static instance: CheckpointService;
   private checkpoints: CheckpointNode[];
   private _onDidChangeTreeData: EventEmitter<
     BaseNode | undefined
   > = new EventEmitter<BaseNode | undefined>();
+  private myRequestService = RequestService.getInstance();
+  private orgInfo: OrgInfo;
+  private sfdxProject: string | null = null;
+
+  //  private orgInfo: OrgInfo;
```
DatGuyJonathan: This can be removed.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -44,39 +45,27 @@ function registerDebugHandlers(): vscode.Disposable {
...
-            eventBody.line
-          ) {
-            checkpointService.addCheckpointNode(
-              eventBody.sourceFile,
-              eventBody.typeRef,
-              eventBody.line
+          // Load the breakpoint line info stuff right here. Need to await this, otherwise
+          // we can (and will if they're stored) get breakpoint events before the line
+          // breakpoint information is fully loaded.
+          if (!BreakpointUtil.getInstance().hasLineNumberMapping()) {
```
DatGuyJonathan: This looks like it’s here because retrieveLineBreakpointInfo could fail in activate, so you want to try again. Are you able to repro a scenario where this tries again and actually succeeds?

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +83,22 @@ export function activate(context: vscode.ExtensionContext) {
     debugConfigProvider,
     checkpointsView
   );
+
+  await checkpointService.activateRequestService();
```
DatGuyJonathan: I think you need error handling here. If there’s no org, this doesn’t look like it would exit cleanly and inform the user with what the error is and what they can do to resolve.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +83,22 @@ export function activate(context: vscode.ExtensionContext) {
...
+
+  // Load the breakpoint line info stuff right here. Need to await this, otherwise
+  // we can (and will if they're stored) get breakpoint events before the line
+  // breakpoint information is fully loaded.
+  if (!BreakpointUtil.getInstance().hasLineNumberMapping()) {
```
DatGuyJonathan: When will we not go into this if-block?

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +83,22 @@ export function activate(context: vscode.ExtensionContext) {
...
+    console.log('in activate, getting line breakpoint info');
+    await retrieveLineBreakpointInfo();
+  }
+
+  // This particular event handler will get called whether or not there
+  // is an active debug session. It is necessary for checkpoints to able to
+  // be created without an active debug session since the information from
+  // them will potentially be used for the reply-debugger session.
+  vscode.debug.onDidChangeBreakpoints(processBreakpointChangedForCheckpoints);
```
DatGuyJonathan: Can you make this a `Disposable`? See `context.subscriptions` above.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+        theNode.setActionCommandResult(duplicateId, undefined);
+        await this.executeRemoveApexExecutionOverlayActionCommand(theNode);
+        // clear out the action command result
+        theNode.setActionCommandResult(undefined, undefined);
+        await this.executeCreateApexExecutionOverlayActionCommand(
+          theNode,
+          false /* recreateIfDupe */
+        );
+      } else {
+        theNode.setActionCommandResult(undefined, result[0].message);
```
DatGuyJonathan: This experience is different than typical breakpoints. In VS Code, if a breakpoint is not valid, the color turns gray. Looking at this, if it fails to create a checkpoint on the server side, the conditional breakpoint and checkpoint on the client side still exist but will an error message. The conditional breakpoint is probably still red.

What can we do to make the experience more holistic? Should we show an error message to the user and delete the breakpoint & checkpoint? Should we show an error message to the user, delete the checkpoint, and mark the breakpoint as invalid?

DatGuyJonathan:
General comment: You have a lot of comments. A lot of these comments look like notes during your development, so if you’re done with them, please remove them. I understand it can be helpful to explain things that are not so obvious, but keep https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/docs/coding-guidelines.md#comments in mind.

packages/salesforcedx-vscode-replay-debugger/package.json
JimSuplizio: Yes, I want to handle this in another PR. There's enough churn with this PR where I really didn't want to add this to the mix.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
-    line: number
-  ): CheckpointNode {
-    const cpNode = new CheckpointNode(sourceFile, typeRef, line);
+  // There's a limit to the number of checkpoints for a given user and
+  // that number is 5.
+  public canAddCheckpointNote(): Boolean {
+    return this.getChildren().length < MAX_ALLOWED_CHECKPOINTS;
+  }
+
+  public async addCheckpointNode(
```
Sweetman: I think renaming this to getOrCreateCheckpointNode would be more succinct and let us remove the comment block on 117-119

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+      return cpTemp;
+    }
+    const cpNode = new CheckpointNode(
+      uriInput,
+      sourceFileInput,
+      typeRefInput,
+      lineInput
+    );
+    // This is here for testing purposes. The command is being tested separately
+    if (makeCommandCall === true) {
```
Sweetman: This makeCommandCall might not be necessary. I read over your tests that use this and I think you could stub out this method to achieve the same result without a boolean parameter

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+  private returnCheckpointNodeIfAlreadyExists(
+    uriInput: string,
+    lineInput: number
+  ): CheckpointNode | undefined {
+    for (const cp of this.checkpoints) {
+      if (cp.getCheckpointLineNumber() === lineInput) {
+        if (
+          cp.getCheckpointUri().toLocaleLowerCase() ===
+          uriInput.toLocaleLowerCase()
+        ) {
```
Sweetman: These 2 if checks can be combined into 1 check

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+    }
+    // If we didn't find anything then just return undefined
+    return undefined;
+  }
+
+  public async deleteCheckpointNode(
+    uriInput: string,
+    lineInput: number,
+    makeCommandCall = true
+  ): Promise<boolean> {
```
Sweetman: The boolean value doesn't seem to be used anywhere so it might be better to make this return a void promise.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -75,16 +101,172 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+    theNode: CheckpointNode,
+    recreateIfDupe = true
+  ) {
+    // create the overlay action
+    const overlayActionCommand = new ApexExecutionOverlayActionCommand(
+      theNode.createJSonStringForOverlayAction()
+    );
+    let errorString = undefined;
+    let returnString = undefined;
+    await this.myRequestService.execute(overlayActionCommand).then(
```
Sweetman: Using `await` and `then` together here is kind of weird. If the promise rejects, an error will be thrown and so you'll want to wrap this call in a try/catch block. If an error occurs with the request, catch it and do your error handling with the error string there

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -193,5 +426,112 @@ export class CheckpointInfoIterationNode extends CheckpointInfoNode {
...
+  }
+
+  // The remove always needs to call delete to get rid of the checkpoint.
+  for (const bp of breakpointsChangedEvent.removed) {
+    if (
+      bp.enabled &&
+      bp.condition &&
+      bp.condition!.toLowerCase().indexOf(CHECKPOINT) >= 0
+    ) {
+      if (bp instanceof vscode.SourceBreakpoint) {
```
Sweetman: I think deleting the checkpoints should happen before adding them since there is a limit of 5 checkpoints.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
Sweetman: My fault for not telling you about this Jim. I think when we talked I mentioned that you didn't need to push this as a disposable but it should be since we want this listener to be thrown away when the extension is deactivated.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
```diff
@@ -167,7 +168,7 @@ export class LogContext {
     const processedSignature = signature.endsWith(')')
       ? signature.substring(0, signature.lastIndexOf('.'))
       : signature;
-    const typerefMapping = this.session.getBreakpointUtil().getTyperefMapping();
+    const typerefMapping = BreakpointUtil.getInstance().getTyperefMapping();
```
vazexqi: What's the reasoning behind changing the call from using `this.session.getBreakpointUtil()` to `BreakpointUtil.getInstance()`.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
vazexqi: And also, for JavaScript and TypeScript, we don't check if something === true, we just do `if(makeCommandCall)`

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -4,35 +4,44 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
+import {
```
vazexqi: My gut feeling tells me that this file has too many lines of comments. 
I worry that these are going to go out of date and no one will maintain them.
We should evaluate this and see if any of the comments should be removed in lieu of extracting things into new methods with better names.

vazexqi:
@JimSuplizio - Is this behind some pref so that this feature is not exposed to customers yet until all parts are hooked up?

We might start giving out the Replay Debugger to some users who have early access to 214 so we don't want to expose anything to them that is not ready.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/breakpointUtil.ts
JimSuplizio: Yes, done. It'll be updated with the next iteration.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
JimSuplizio: Fixed.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
JimSuplizio: See comment about for the same thing. Fixed.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
JimSuplizio: Nick, I changed the class to be a singleton class. Previously, this class was only used in the adapter and the instance was stored on the apedReplayDebug. With the addition of processing breakpoints in the extension it was needed in both processes and had to be setup in each. In the extension code it's loaded on activation and then used in the breakpoint event, creating the singleton was better, cleaner way to do this. Being a singleton also has the added bonus of making it easier to reload once we get any filesystem changed events hooked up.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Yeah, I'll move this but funny story: The only time any of these arrays contains more than one object is when VS Code first loads the project and is loading beakpoints which is specific to add. Otherwise, from the UI they happen one at a time.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
vazexqi: Yup the singleton makes sense. We usually create a Singleton and just export that instance out. See https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/channels/index.ts#L9

We don't have classes calling `.getInstance()` directly.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/breakpointUtil.ts
JimSuplizio: Yes. Those would be the tests above with the incorrect removal comment.

packages/salesforcedx-vscode-replay-debugger/adaptertest/unit/adapter/apexReplayDebug.test.ts
JimSuplizio: Initially I thought these would be removed but when it turned out that the breakpoint info was needed in the adapter and the extension I'd forgotten to remove this comment. The tests aren't going to be removed, they'd just needed to be slightly tweaked.

packages/salesforcedx-vscode-replay-debugger/src/core/logContext.ts
JimSuplizio: Sure. Done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Fixed. Changed things to a stub and removed makeCommandCall altogether.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: James and I talked about this. It's staying. This looks to be the right call.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Removed.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+    // the error was already reported in the function, return
+    return;
+  }
+  console.log('in activate, successfully activated request service');
+
+  const breakpointsSub = vscode.debug.onDidChangeBreakpoints(
+    processBreakpointChangedForCheckpoints
+  );
+  context.subscriptions.push(breakpointsSub);
+  console.log(
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -107,6 +114,49 @@ function getDialogStartingPath(): vscode.Uri | undefined {
...
 
+async function retrieveLineBreakpointInfo(): Promise<boolean> {
+  const sfdxApex = vscode.extensions.getExtension(
+    'salesforce.salesforcedx-vscode-apex'
+  );
+  if (sfdxApex && sfdxApex.exports) {
+    let expired = false;
+    let i = 0;
+    while (!sfdxApex.exports.isLanguageClientReady() && !expired) {
+      await imposeSlightDelay(100);
```
vazexqi: @JimSuplizio - What is the root cause for needing this? This seems brittle. Is this a temporary solution that got checked in by mistake?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -44,39 +45,19 @@ function registerDebugHandlers(): vscode.Disposable {
     async event => {
       if (event && event.session && event.session.type === DEBUGGER_TYPE) {
         if (event.event === GET_LINE_BREAKPOINT_INFO_EVENT) {
-          const sfdxApex = vscode.extensions.getExtension(
-            'salesforce.salesforcedx-vscode-apex'
+          console.log('in registerDebugHandlers, getting line breakpoint info');
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
     debugConfigProvider,
     checkpointsView
   );
+
+  console.log('in activate, getting line breakpoint info');
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+  if (!await retrieveLineBreakpointInfo()) {
+    console.log(
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+      'in activate, did not retrieve line breakpoint info, returning'
+    );
+    // the error was already reported in the function, return
+    return;
+  }
+  console.log('in activate, successfully retrieved line breakpoint info');
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+
+  console.log('in activate, activating request service');
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -94,6 +75,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+  if (!await checkpointService.activateRequestService()) {
+    console.log('in activate, activating request service failed, returning');
```
vazexqi: @JimSuplizio - Can we remove this `console.log`?

CC: @DatGuyJonathan , @lcampos

packages/salesforcedx-vscode-replay-debugger/src/index.ts
JimSuplizio: @vazexqi This was necessary because the language client may not actually be ready. The replay extension and the replay DA request source/line information at different times. When checkpoints were added we needed the source/line information much sooner than the DA, which only needs it when it starts replay-debugging. The problem here is that the extension was trying to invoke getLineBreakpointInfo when the extension wasn't ready and failing. The difference here was less than a couple hundred milliseconds. The extensions, themselves, have an onReady event that signals when they're ready (look at the changes for packages/salesforcedx-vscode-apex/src/index.ts, in this review), for that event I set the flag. In the replay debugger extension we will wait up to 3 seconds in 100 millisecond intervals to check the status. It either expires and reports an error that we're unable to get source line information from the language server OR the extension is ready and we can call to get the source line information.

PR #367, opened 2018-04-07
Update Bootstrap command to new folder structure

guw:
This new layout is what the APEX LS will read/parse for indexing packages with
their namespace


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -155,7 +157,9 @@ export const messages = {
   error_creating_packagexml: 'Error creating package.xml. %s',
   error_extracting_org_source: 'Error extracting downloaded Apex source. %s',
   error_extracting_packages: 'Error extracting packages: %s',
-  error_updating_sfdx_project: 'Error updating sfdx-project.json: %s',
+  error_updating_sfdx_project: 'Error updateing sfdx-project.json: %s',
```
ruthemmanuelle: The version without the extra 'e' was correct: "updating"

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -79,6 +93,38 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+    data: IsvDebugBootstrapConfig
+  ): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(
+        nls.localize(
+          'isv_debug_bootstrap_step2_configure_project_retrieve_namespace'
+        )
+      )
+      .withArg('force:data:soql:query')
+      .withFlag('--query', 'SELECT NamespacePrefix FROM Organization LIMIT 1')
```
DatGuyJonathan: @guw Is the limit clause necessary?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -192,6 +256,36 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+    const orgNamespaceInfoResponseJson = await this.executeCommand(
+      this.buildQueryForOrgNamespacePrefixCommand(response.data),
+      { cwd: projectPath },
+      cancellationTokenSource,
+      cancellationToken
+    );
+    try {
+      const sfdxProjectJsonFile = path.join(projectPath, 'sfdx-project.json');
+      const sfdxProjectConfig = JSON.parse(
+        fs.readFileSync(sfdxProjectJsonFile).toString()
```
DatGuyJonathan: Can this be given the encoding to get the string form straight from `readFileSync`?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: In theory it isn't. However, I consider it good practise from my SQL times. You'd like to see it removed?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: Might be a good idea. I'll check the api and enforce UTF-8.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
DatGuyJonathan: Nope, I was just curious if it was returning more than one.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
guw: @DatGuyJonathan I applied encoding to all file ops (read and write). Good catch. Even though we shouldn't have unexpected chars in those files it's probably best to have it for Windows.

PR #369, opened 2018-04-09
sObject Explorer

ntotten:
### What does this PR do?
Adds an extension that provides a treeview for browsing sObjects. 

N/A

Submitting this for feedback and code review.


vazexqi:
QQ: Why is this done in a new salesforcedx-vscode-sobject-explorer extension instead of being folded into one of the existing extension? Do we not want this functionality by default?

ntotten:
@vazexqi Updated to be part of vscode-core. Let me know how it looks. Still need to add tests.

vazexqi:
Cool. Let's create a mirror internal work item to track this and then let's discuss when this should go in and the long-term plan. We can also talk about when to bring this in during our backlog grooming this week.

PR #372, opened 2018-04-11
Fix getting the uri for a signature with inner class

DatGuyJonathan:
### What does this PR do?
Sample signature is `BotController.findMatchingHandlerIfPossible(BotController.HandlerMapping, String, Map<String,String>, List<String>)`, so we need to remove the params then the method name.

@W-4852805@


packages/salesforcedx-apex-debugger/test/integration/config/variables/source/permissionsets/DebugApex.permissionset-meta.xml
```diff
@@ -583,6 +583,10 @@
         <enabled>true</enabled>
         <name>ModifyAllData</name>
     </userPermissions>
+    <userPermissions>
```
vazexqi: Ah yes this is new for 214.

PR #375, opened 2018-04-16
Changes found in Git after npm run bootstrap

guw:
Reformatted `package.json`


vazexqi:
You should run `scripts/reformat-with-prettier.js`. Perhaps we should also hook that up with husky.

guw:
@vazexqi Then this change should go in. It fixes the formatting from 3ba4d2f69cafa324e436f620a0d5e92e7fe07238.

PR #379, opened 2018-04-17
Generate launch.json

guw:
### What does this PR do?
Simplify setup work with pre-generated launch.json for connecting Apex debugger.

W-4894080

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
@@ -424,6 +424,44 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       }
     }
 
+    // 8: generate launch configuration
+    channelService.appendLine(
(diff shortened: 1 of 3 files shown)


DatGuyJonathan:
Okay, I understand. This should be fine.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -424,6 +424,44 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
...
+              request: 'launch',
+              userIdFilter: [],
+              requestTypeFilter: [],
+              entryPointFilter: '',
+              sfdxProject: '${workspaceRoot}',
+              connectType: 'ISV_DEBUGGER'
+            }
+          ]
+        }),
+        { encoding: 'utf-8' }
```
DatGuyJonathan: Nitpick: This can be just `'utf-8'` because `writeFileSync` has an overload whose second param is a string to specify the encoding. I've seen uses of `fs` in our other extensions follow that approach.

PR #384, opened 2018-04-17
Update to remove checkpoint from VS Code if it fails to create

JimSuplizio:
### What does this PR do?
1. Update to remove the checkpoint correctly if it fails to create.
2. The result message from the checkpoint create can actually result in a message that is not a ApexExecutionOverlayFailureResult but, is instead, is just a string. This can happen in the case where maybe the user loses network connectivity and can't connect to the appserver. The message returned in that case is just a string. If we're unable to parse the error message as an ApexExecutionOverlayFailureResult then just give the error string as the error message.
3. Last but not least, instead of showInformationMessage when we retrieve the breakpoint information, just print it to the console like the other informational messages.

@W-4719064@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
@@ -232,9 +232,9 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
     // is a plain string. It'll cause the parsing to fail.
     if (errorString) {
       try {
-        const result = <ApexExecutionOverlayFailureResult[]>JSON.parse(
+        const result = JSON.parse(
           errorString
-        );
+        ) as ApexExecutionOverlayFailureResult[];
         // If the node is a dupe, then delete it and recreate it. If the failure reason
         // was something else that we don't handle then just set the command result.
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -228,29 +228,37 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
-        // then call the delete again with the recreation flag unset
-        theNode.setActionCommandResult(duplicateId, undefined);
-        await this.executeRemoveApexExecutionOverlayActionCommand(theNode);
-        // clear out the action command result
-        theNode.setActionCommandResult(undefined, undefined);
-        await this.executeCreateApexExecutionOverlayActionCommand(
-          theNode,
-          false /* recreateIfDupe */
+      try {
+        const result = <ApexExecutionOverlayFailureResult[]>JSON.parse(
```
DatGuyJonathan: Our standard for [type assertion](https://www.typescriptlang.org/docs/handbook/basic-types.html) is to use `as`. So this should be `JSON.parse(errorString) as ApexExecutionOverlayFailureResult[]`.

PR #386, opened 2018-04-18
Handle both sfdx.cmd and sfdx.exe on Windows

### LINKED ISSUE #383: Getting a Java Bind issue on launch of VS code
### Summary
_Short summary of what is going on or to provide context_.
Getting a Java Bind issue on launch of VS code

Ha da lot of trouble getting this going when I run any of the sfdx commands from the command palate I get this:
" 
Starting SFDX: Push Source to Default Scratch Org

sfdx force:source:push
sfdx force:source:push ended with error spawn sfdx ENOENT
The SFDX CLI is not installed. Install it from https://developer.salesforce.com/tools/sfdxcli
"

was running into a lot of java issues up front, have reinstalled the apps  quite a few times. I think i worked through that by changing the key in my user settigns to this:
"salesforcedx-vscode-apex.java.home": "/Program Files/Java/jre1.8.0_171"

if this isn't your issue please point me in the right direction

thanks!

### Steps To Reproduce:
1. Launch VS code
1. open debugger

### Expected result
_Describe what should have happened_.
No Java biding Issue

### Actual result
_Describe what actually happened instead_.
Seeing a java bind issue (screen shot attached)

java bind error

### Additional information

**VS Code Version**:
1.22.2
**SFDX CLI Version**:
sfdx-cli/6.11.0-9a1ef8794d (windows-x64) node-v8.9.4
**OS and version**:
Windows 10
Version	10.0.16299 Build 16299

vazexqi:
### What does this PR do?

The new Salesforce CLI installers no longer produce a sfdx.exe. Instead it produces a sfdx.cmd. When using a .cmd on Windows, we need a different way to launch this. See https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows

We also need to make sure that we use the old way for customers who are still on the older installers, which are still on sfdx.exe

@W-4914336@
Resolves #383

### DIFF
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -9,7 +9,8 @@
   "categories": ["Other"],
   "dependencies": {
     "rxjs": "^5.4.1",
-    "tree-kill": "^1.1.0"
+    "tree-kill": "^1.1.0",
+    "which": "1.3.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
@@ -17,6 +18,7 @@
     "@types/mocha": "^2.2.38",
     "@types/node": "^6.0.40",
     "@types/shelljs": "^0.7.4",
+    "@types/which": "1.3.1",
     "chai": "^4.0.2",
     "decache": "^4.1.0",
     "glob": "^7.1.2",
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -8,17 +8,17 @@
   "license": "BSD-3-Clause",
   "categories": ["Other"],
   "dependencies": {
+    "cross-spawn": "6.0.5",
     "rxjs": "^5.4.1",
-    "tree-kill": "^1.1.0",
-    "which": "1.3.0"
+    "tree-kill": "^1.1.0"
   },
   "devDependencies": {
     "@types/chai": "^4.0.0",
+    "@types/cross-spawn": "6.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "^2.2.38",
     "@types/node": "^6.0.40",
     "@types/shelljs": "^0.7.4",
-    "@types/which": "1.3.1",
     "chai": "^4.0.2",
     "decache": "^4.1.0",
     "glob": "^7.1.2",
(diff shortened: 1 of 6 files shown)


vazexqi:
We will do a first run first using the old appveyor build which is using the old installers. If that passes, then I will update the appveyor build to use the new installer and then do another run.

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -9,14 +9,16 @@
   "categories": ["Other"],
   "dependencies": {
     "rxjs": "^5.4.1",
-    "tree-kill": "^1.1.0"
+    "tree-kill": "^1.1.0",
+    "which": "1.3.0"
```
vazexqi: This package helps us determine which sfdx is running, the sfdx.exe or the sfdx.cmd one.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -24,11 +26,26 @@ export interface CancellationToken {
 
 export class CliCommandExecutor {
   private readonly command: Command;
-  private readonly options: ExecOptions;
+  private readonly options: SpawnOptions;
```
vazexqi: I made a mistake earlier and used ExecOptions instead of SpawnOptions. We should use SpawnOptions.
@guw I see that you duplicate this code in your ISV boostrap command. You have to change it yourself.

guw: @vazexqi Can you point to the duplication? Are you referring to the overriden `executeCommand` method? I think it would be good if such refactorings are captured in the same PR for traceability in code history, though.

vazexqi: What was the reason for code duplication? Could you eliminate that duplication?

vazexqi: I see where it is now. I can fix it in this PR.

vazexqi:
First run (with old installers but new code) passed on https://ci.appveyor.com/project/guw/salesforcedx-vscode/build/release/v42.12.0-2375?fullLog=true

Changing the appveyor build to use the new installers now (require PowerShell-fu so🤞 and might take a few iterations).

vazexqi:
@DatGuyJonathan - Do you know if there are any places where we invoke the CLI but do not go through the `CliCommandExecutor`?

DatGuyJonathan:
No, I think we've done pretty well at using `CliCommandExecutor` consistently.

PR #387, opened 2018-04-18
Update CHANGELOG.md

vazexqi:
### What does this PR do?

Adds a last minute fix to CHANGELOG.md

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixed
 #### salesforcedx-vscode-core
 
-* Support invoking the new Salesforce CLI which were installed using the new installers on Windows ([PR #386](https://github.com/forcedotcom/salesforcedx-vscode/pull/386)
+* Support invoking Salesforce CLI commands for CLI installations that used the new installers on Windows  ([PR #386](https://github.com/forcedotcom/salesforcedx-vscode/pull/386))
 
 ## 42.11.0 - April 12, 2018
 ### Added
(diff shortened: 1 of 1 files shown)


vazexqi:
This issue is being fixed in #386. Basically the _new_ Salesforce CLI installers from https://developer.salesforce.com/tools/sfdxcli install a different binary compared to the old installers. More concretely, the old installers used a `sfdx.exe` but the new installers use `sfdx.cmd`. Invoking the .exe vs .cmd require different mechanisms on Windows. See https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows

packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 * Go To Definition and Find All References for types used with instanceOf ([PR #376](https://github.com/forcedotcom/salesforcedx-vscode/pull/376))
 * Go To Definition for implicit constructors across Apex classes ([PR #376](https://github.com/forcedotcom/salesforcedx-vscode/pull/376))
 
+### Fixed
+#### salesforcedx-vscode-core
+
+* Support invoking the new Salesforce CLI which were installed using the new installers on Windows ([PR #386](https://github.com/forcedotcom/salesforcedx-vscode/pull/386)
```
ruthemmanuelle: The new Salesforce CLI what? Would this be accurate? 

* Support invoking Salesforce CLI commands for CLI installations that used the new installers on Windows ([PR #386](https://github.com/forcedotcom/salesforcedx-vscode/pull/386)

ruthemmanuelle: Also, we're missing an ending parenthesis.

PR #392, opened 2018-04-24
Replay Debugger Bug Fixes

JimSuplizio:
### What does this PR do?
There are 3 bug fixes here.
W-4919185 - The checkpoint result was useful for diagnostics when doing the initial work on the feature. Because we're not creating the checkpoint if there's a failure (we remove it) it's far less useful. Move the checkpoint result onto the CheckpointNode instead of making it its own CheckpointInfoNode.

W-4919352 - The difference in behavior is due to not checking canSetLineBreakpoint before passing things off to the appserver. Once that check is in place if the user tries to create a checkpoint on an invalid line it'll give them an error "Checkpoints can be set only on a valid line of Apex source." 
checkpointinvalidlineerror

W-4919978 - This error message is "confusing" if you're not getting in from SFDX after passing in an invalid typeRef to create a checkpoint from the command line. When in VS Code, if the error code from the AppServer is FIELD_INTEGRITY_EXCEPTION that means that the source is out of sync with the server. Translate the error to something more user friendly for VS Code "The local source is out of sync with the server. Push any changes you've made locally to your org, and pull any changes you've made in the org into your local project."
sourceoutofsync

W-4920046 - W-4920046 - This fails on Windows because of this VS Code issue: https://github.com/Microsoft/vscode-languageserver-node/issues/105. I'm pretty much stealing this function from https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lwc/src/index.ts#L107 which fixes the issue. I've already tested this on a Windows VM and on Linux.

@W-4919185, @W-4919352@ @W-4919978@ @W-4920046@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
@@ -34,6 +34,7 @@ import {
 import {
   CHECKPOINT,
   DUPLICATE_VALUE,
+  FIELD_INTEGRITY_EXCEPTION,
   MAX_ALLOWED_CHECKPOINTS
 } from '../constants';
 import { nls } from '../messages';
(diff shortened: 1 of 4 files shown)


JimSuplizio:
Adding bug fixes
@w-4919185 @w-4919352@ @w-4919978@ @w-4920046@

PR #395, opened 2018-04-25
Update changelog for v42.13.0 release

lcampos:
### What does this PR do?
Update changelog for v42.13.0 release

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 42.13.0 - April 26, 2018
+### Fixed
+#### salesforcedx-vscode-apex
+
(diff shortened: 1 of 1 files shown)


vazexqi:
LGTM but let's wait until we have confirmation on W-4863416 (I mentioned you on the bug in GUS as well).

vazexqi:
OK. I've gotten confirmation on W-4863416. It's not something we fixed on the extensions but it has impact. Here's how I would summarize it:

Fixed

salesforcedx-vscode

Copying-and-pasting Salesforce CLI output from the _embedded terminal_ in VS Code would embed terminal escape characters in the pasted text. We have fixed this in the Salesforce CLI and the VS Code team has made a similar fix in VS Code. If you see any issues with copy-and-pasting, please be sure to update both VS Code and the Salesforce CLI.

@ruthemmanuelle - Please apply word-smithing changes as necessary.

PR #402, opened 2018-04-30
Enable existing commands to work with ISV project

guw:
### What does this PR do?
It introduces a way to allow setting environment variables for all commands globally without affecting the VS Code process environment. Additionally, it detects an ISV project and sets SFDX_DEFAULTUSERNAME and SFDX_INSTANCE_UR accordingly.

W-4744274


packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -44,6 +57,34 @@ export class CliCommandExecutor {
...
+  protected static patchEnv(
+    options: SpawnOptions,
+    baseEnvironment: Map<string, string>
+  ): SpawnOptions {
+    // start with current process environment
+    const env = Object.create(null);
+
+    // inherit current process environment
+    Object.keys(process.env).forEach(key => {
+      env[key] = process.env[key];
```
DatGuyJonathan: Can `Object.assign` be used instead of these `forEach` loops?

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -265,6 +275,43 @@ function notifyDebuggerSessionFileChanged(): void {
...
 }
 
+async function notifyIsvAuthChanged() {
+  if (
+    vscode.workspace.workspaceFolders instanceof Array &&
+    vscode.workspace.workspaceFolders.length > 0
+  ) {
+    const forceConfig = await new ForceConfigGet().getConfig(
+      vscode.workspace.workspaceFolders[0].uri.fsPath,
+      'isvDebuggerSid',
```
DatGuyJonathan: Thoughts on using constants for these? They're in three packages now. It makes more sense to put the constant in a debugger package, but core doesn't depend on debugger. The only common package is utils.

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
guw: I started with `Object.create(process.env)` but that made the object look weird in the debugger. `Object.assign` works for this one but not for the following one (because it's a map).

PR #403, opened 2018-04-30
User org name as default project name.

guw:
### What does this PR do?
Pre-populate the project name inbox with the org name.

W-4681065

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -28,6 +28,7 @@
     "glob": "^7.1.2",
     "moment": "2.20.1",
     "rxjs": "^5.4.1",
+    "sanitize-filename": "^1.6.1",
     "shelljs": "^0.8.1"
   },
   "devDependencies": {
@@ -37,6 +38,7 @@
     "@types/mocha": "2.2.38",
     "@types/node": "^7.0.0",
     "@types/path-exists": "^1.0.29",
+    "@types/sanitize-filename": "^1.1.28",
     "@types/shelljs": "^0.7.8",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -35,7 +35,7 @@
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "2.2.38",
-    "@types/node": "^8.9.4",
+    "@types/node": "^7.0.0",
     "@types/path-exists": "^1.0.29",
     "@types/shelljs": "^0.7.8",
     "@types/sinon": "^2.3.2",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -35,7 +35,7 @@
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "2.2.38",
-    "@types/node": "^6.0.40",
+    "@types/node": "^8.9.4",
     "@types/path-exists": "^1.0.29",
     "@types/shelljs": "^0.7.8",
     "@types/sinon": "^2.3.2",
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "2.2.38",
-    "@types/node": "^6.0.40",
+    "@types/node": "^8.9.4",
```
DatGuyJonathan: What prompted this change?

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -47,7 +47,7 @@ class ForceApexExecuteExecutor extends SfdxCommandletExecutor<{}> {
     }).execute(cancellationToken);
 
     execution.processExitSubject.subscribe(async data => {
-      fs.unlink(response.data.fileName);
+      fs.unlink(response.data.fileName, err => null);
```
DatGuyJonathan: What prompted this change?

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -512,30 +511,55 @@ export type IsvDebugBootstrapConfig = ProjectNameAndPath & ForceIdeUri;
...
+      if (loginUrl && sessionId) {
         const protocolPrefix =
-          parameter.secure === '0' ? 'http://' : 'https://';
+          parameter.get('secure') === '0' ? 'http://' : 'https://';
+        this.forceIdUrl = {
+          loginUrl: loginUrl.toLowerCase().startsWith('http')
+            ? loginUrl
+            : protocolPrefix + loginUrl,
+          sessionId: sessionId,
+          orgName: url.hostname
```
DatGuyJonathan: I think this can have dashes, which should be okay on all operating systems. Can this have any other special characters, like a colon (local appserver with port)?

packages/salesforcedx-vscode-core/src/messages/index.ts
```diff
@@ -41,6 +41,9 @@ function loadMessageBundle(config?: Config): Message {
   }
 }
 
-export const nls = new Localization(
-  loadMessageBundle(JSON.parse(process.env.VSCODE_NLS_CONFIG))
-);
+function getNlsConfig(): Config | undefined {
+  const procNlsConfig = process.env.VSCODE_NLS_CONFIG;
+  return procNlsConfig ? JSON.parse(procNlsConfig) : null;
```
DatGuyJonathan: What prompted this change?

packages/salesforcedx-vscode-core/package.json
vazexqi: I'm not sure you can/should be using Node 8 API. VS Code insiders is still using Node 7.9.0. See https://github.com/Microsoft/vscode/issues/11660#issuecomment-246265598

guw: I am consistent with what we define in `.nvmrc`. Is this the wrong one for our builds and tests?

vazexqi: We don't use any of the new Node 8 APIs even if we are running on .nvmrc.
See the discussion at W-4395117.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -28,15 +28,17 @@
...
+    "sanitize-filename": "^1.6.1",
     "shelljs": "^0.8.1"
   },
   "devDependencies": {
     "@types/adm-zip": "^0.4.31",
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "2.2.38",
-    "@types/node": "^6.0.40",
+    "@types/node": "^7.0.0",
```
DatGuyJonathan: I would prefer not to mix-and-match `@types/node` version in our packages. I'm not entirely sure of the repercussions of different versions, but it seems safer to be consistent. Thoughts?

Also, is it safe to just go to 7.9.0 to match the node version that Electron ships with?

guw: There is no 7.9.0 available of that packages. 7.0.x is the highest 7.x one.

I was thinking of consistent as well but then held back because of the compile issues I noticed just in this module. If we do it global the change might introduce too much noise for the actual change. IMHO it would be better to file a Tech Debt work item and look at this when VS Code is done upgrading to Electron 2.0

PR #404, opened 2018-04-30
Apex LS changes to support completion and visibility with supertypes

GaryFNC:
Add in the latest Apex LS jar that provides completion support for
super types - both for inherited methods and inherited fields.
Also, completion now doesn't suggest methods or fields that are not
visible (ex. private in the supertype, etc.)


vazexqi:
👍 Let's get the jar in first and then if there are issues on the individual work items, we can handle that separately.

PR #406, opened 2018-05-04
Ignore warnings and CLI update messages when using --json

DatGuyJonathan:
### What does this PR do?
This change standardizes CLI calls that need --json to use `CommandBuilder.withJson()`. In `withJson()`, it will append `--loglevel fatal` to ignore warnings from CLI so we avoid getting multiple JSON objects.

Additionally, CLI outputs messages when it's updating to the latest version. When this happens with a command that needs --json, the messages are in human-readable format and it cannot be supressed with `--loglevel fatal`. You would see something like this:

```
sfdx-cli: Updating to ...
{ ... the JSON we care about ... }
```

The update messages actually go to stderr, so the fix is to ignore stderr.

@W-4485495@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
@@ -42,7 +42,11 @@ export class CommandBuilder {
   }
 
   public withArg(arg: string): CommandBuilder {
-    this.args.push(arg);
+    if (arg === '--json') {
+      this.withJson();
+    } else {
+      this.args.push(arg);
+    }
     return this;
   }
 
@@ -53,6 +57,7 @@ export class CommandBuilder {
 
   public withJson(): CommandBuilder {
     this.args.push('--json');
+    this.args.push('--loglevel', 'fatal');
     return this;
   }
 
--- a/packages/salesforcedx-utils-vscode/src/cli/commandOutput.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandOutput.ts
@@ -8,14 +8,15 @@
 import { CommandExecution } from './commandExecutor';
 
 export class CommandOutput {
-  private buffer = '';
+  private stdoutBuffer = '';
+  private stderrBuffer = '';
 
   public async getCmdResult(execution: CommandExecution): Promise<string> {
     execution.stdoutSubject.subscribe(realData => {
-      this.buffer += realData.toString();
+      this.stdoutBuffer += realData.toString();
     });
     execution.stderrSubject.subscribe(realData => {
-      this.buffer += realData.toString();
+      this.stderrBuffer += realData.toString();
     });
 
     return new Promise<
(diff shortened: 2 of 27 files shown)


packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
```diff
@@ -42,7 +42,11 @@ export class CommandBuilder {
   }
 
   public withArg(arg: string): CommandBuilder {
-    this.args.push(arg);
+    if (arg === '--json') {
+      this.withJson();
```
DatGuyJonathan: I noticed a bunch of calls like `withArg('--json')`, so I figured I'd add this here in case someone accidentally uses withArg in the future, instead of withJson.

packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
```diff
@@ -53,6 +57,7 @@ export class CommandBuilder {
 
   public withJson(): CommandBuilder {
     this.args.push('--json');
+    this.args.push('--loglevel', 'fatal');
```
DatGuyJonathan: Ignore CLI warnings.

packages/salesforcedx-utils-vscode/src/cli/commandOutput.ts
```diff
@@ -8,24 +8,25 @@
...
+  private stderrBuffer = '';
 
   public async getCmdResult(execution: CommandExecution): Promise<string> {
     execution.stdoutSubject.subscribe(realData => {
-      this.buffer += realData.toString();
+      this.stdoutBuffer += realData.toString();
     });
     execution.stderrSubject.subscribe(realData => {
-      this.buffer += realData.toString();
+      this.stderrBuffer += realData.toString();
```
DatGuyJonathan: updateoutputgoestostdderr

When the CLI updates, the update messages go into stderr and the command output goes to stdout. So, if the exit code is okay, then we can just use stdout and ignore the CLI update messages in stderr.

DatGuyJonathan:
@vazexqi There's a test failure in lwc-next eslint.test.ts that also exists in develop branch. The test is `Should configure eslint.nodePath on sfdx-simple`. It seems like https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lwc-next/src/index.ts#L171 isn't updating the workspace config.

The reason I say it also exists in develop branch is because the build for https://github.com/forcedotcom/salesforcedx-vscode/commit/5b920dea5b15a0abf5669904f722d97a3230196d passed, then I re-ran and it failed, which usually suggests a dependency got updated. It's not clear to me what dependency of lwc-next could've broken. Could you provide some insight?

vazexqi:
@DatGuyJonathan - I've verified manually on my Windows VM that the code is doing what it is doing. The test is having some issues waiting for the right conditions - this could be a change in the underlying set up for VS Code that is no longer waiting. 

I am going to remove that integration test in this branch. There is enough coverage from the unit tests.

PR #407, opened 2018-05-07
Remove  "salesforcedx-utils-vscode" dependency in Apex Debugger Extension

guw:
### What does this PR do?
There's some odd behavior with regards to adding a dependency in `"@salesforce/salesforcedx-utils-vscode"`. Removing the depenency by moving the file watchers into core as a workaround makes the problem go away.

W-4965215

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/src/index.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/src/index.ts
@@ -266,56 +266,13 @@ function registerFileWatchers(): vscode.Disposable {
   trgWatcher.onDidChange(uri => notifyDebuggerSessionFileChanged());
   trgWatcher.onDidCreate(uri => notifyDebuggerSessionFileChanged());
   trgWatcher.onDidDelete(uri => notifyDebuggerSessionFileChanged());
-  const isvAuthWatcher = vscode.workspace.createFileSystemWatcher(
-    '**/.sfdx/sfdx-config.json'
-  );
-  isvAuthWatcher.onDidChange(uri => notifyIsvAuthChanged());
-  isvAuthWatcher.onDidCreate(uri => notifyIsvAuthChanged());
-  isvAuthWatcher.onDidDelete(uri => notifyIsvAuthChanged());
-  return vscode.Disposable.from(clsWatcher, trgWatcher, isvAuthWatcher);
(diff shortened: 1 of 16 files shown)


guw:
@DatGuyJonathan I verified that the changes works with a locally built VSIX
`git clean -xfd; npm install; npm run compile; npm run vscode:package`.

vazexqi:
@guw or @DatGuyJonathan  - What was the error that you see with the previous configuration? Without seeing that I won't be able to suggest any other possible alternatives. The current way of doing it seems a big weird since it brings things that don't need to be into vscode-core.

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -22,16 +22,6 @@ import {
   WORKSPACE_SETTINGS_REQUEST,
   WorkspaceSettings
 } from '@salesforce/salesforcedx-apex-debugger/out/src';
-import {
-  ENV_SFDX_DEFAULTUSERNAME,
-  ENV_SFDX_INSTANCE_URL,
-  SFDX_CONFIG_ISV_DEBUGGER_SID,
-  SFDX_CONFIG_ISV_DEBUGGER_URL
-} from '@salesforce/salesforcedx-utils-vscode/out/src';
```
DatGuyJonathan: Interesting. If utils was can't be a direct dependency of salesforcedx-vscode-apex-debugger, I would think you can do `from '@salesforce/salesforcedx-apex-debugger/node_modules/@salesforce/salesforcedx-utils-vscode/out/src`. Does that not work?

DatGuyJonathan:
I packaged this branch and I'm still seeing the error. Am I missing something?

vazexqi:
@DatGuyJonathan - I haven't had a chance to look into this fix yet. But, you might need to install the other vsix as well (not just the debugger one) because of the restructuring that is going on.
It's best to just install all the vsix that are affected from this change.

vazexqi:
Replaced by #408.

guw:
Reopening since the solution in #408 does not work.

guw:
@DatGuyJonathan I reopened the PR. This moves everything into the core vsix, which makes the ISV auth logic as well as command execution central into one extension.

PR #410, opened 2018-05-09
Variable reassignment on new reference bug fix

Sweetman:
### What does this PR do?
Fixes an issue where reference reassignment does not work properly when reassigning a variable sharing a container to a new reference.
@W-4974520@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
@@ -41,14 +41,6 @@ export class VariableAssignmentState implements DebugLogState {
       if (logContext.getStaticVariablesClassMap().has(className)) {
         map = logContext.getStaticVariablesClassMap().get(className)!;
         container = map.get(varName)! as ApexVariableContainer;
-        if (container.variablesRef !== 0 && !logContext.getRefsMap().has(ref)) {
-          container = new ApexVariableContainer(
-            container.name,
-            '',
-            container.type
-          );
-          map.set(varName, container);
-        }
       } else if (frameInfo.locals.has(varName)) {
         // if name does not contain '.' (i.e. this.attr or a.Name), it should be in locals and we can update the value
         map = frameInfo.locals;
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -34,13 +37,38 @@ export class VariableAssignmentState implements DebugLogState {
...
+        map = frameInfo.locals;
+        container = map.get(varName) as ApexVariableContainer;
+      }
+
+      if (
+        container &&
+        map &&
+        container.variablesRef !== 0 &&
+        !logContext.getRefsMap().has(ref)
+      ) {
```
Sweetman: Without this check, the container for the old reference would be the value for the new ref in the map. This would later be overwritten with the values of the new ref which we don't want in case any other variables are still pointing to the old ref.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -34,13 +37,38 @@ export class VariableAssignmentState implements DebugLogState {
         ref = this.fields[5];
       }
       let container: ApexVariableContainer | undefined;
+      let map: Map<String, VariableContainer> | undefined;
       if (logContext.getStaticVariablesClassMap().has(className)) {
-        const statics = logContext.getStaticVariablesClassMap().get(className)!;
-        container = statics.get(name)! as ApexVariableContainer;
+        map = logContext.getStaticVariablesClassMap().get(className)!;
```
JimSuplizio: We're calling logContext.getStaticVariablesClassMap().get(className) twice, here and in the if statement above. Can't we just set map = logContext.getStaticVariablesClassMap().get(className) and check to see if map is defined?

Sweetman: Yeah I thought this was a little easier to read and verify that the container and map were always tied together at the time but that makes sense too, I'll update it thanks

PR #412, opened 2018-05-10
VSCode engine version update and status bar

JimSuplizio:
### What does this PR do?
1. Update the **engine** version the vscode to 1.22. For the replay debugger extension this was necessary to get access to the progress window which is only in 1.22. All of the other extensions except for salesforcedx-vscode-visualforce were updated. For salesforcedx-vscode-visualforce there will be a work item since that appears to be more involved.
2. Add a progress window that'll show progress the user creates checkpoints (either through the command or through the button on the checkpoints view).
3. Prevent another checkpoint update if one is already in progress. 
4. Add the missing copyright to the batchDeleteExeistingOverlayActionCommand* files.

@W-4979514@ @W-4979533@

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -6,7 +6,7 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.17.0"
+    "vscode": "^1.22.0"
   },
   "categories": ["Debuggers"],
   "dependencies": {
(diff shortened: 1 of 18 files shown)


vazexqi:
@JimSuplizio - Minor but important, it's supposed to be `@W-4979514, W-4979533@` not `@w-4979514@ @w-4979533@`. The auto-tracker looks at the actual squashed commit. So make sure when you squash and merge that the commit has that in there. It doesn't look at the PR comments.

Sweetman:
LGTM

JimSuplizio:
@w-4979514, W-4979533@

PR #413, opened 2018-05-14
Handle Custom SObjects with a namespace

GaryFNC:
Update the ApexLS jar to handle SObjects with a namespace.

Custom SObjects and Fields in a namespace have names like
mynamespace__mysobject__c.

The SObjects are built from faux classes that are compiled
and indexed with those long names in them.
But inside the same namespace, they can be referenced with
a short form of the name - ex. mysobject__c.

Currently, completions only offer the full names.


vazexqi:
@GaryFNC - Is there a reason you bypass the PR template? It always ask for the work item ID. Please follow that template.

PR #416, opened 2018-05-15
Try to pull memory references as a value from reference map

Sweetman:
### What does this PR do?
Tries to pull relevant memory references that were stored during the creation of objects. This is useful for multiply nested objects that would show reference addresses for fields instead of the actual field value pairs.
@W-4873610@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
@@ -37,7 +37,14 @@ export class VariableAssignmentState implements DebugLogState {
         ref = this.fields[5];
       }
       let container: ApexVariableContainer | undefined;
-      let map: Map<string, VariableContainer> | undefined;
+      let map: Map<String, VariableContainer> | undefined;
+
+      if (ref && !logContext.getRefsMap().has(ref)) {
+        logContext
+          .getRefsMap()
+          .set(ref, new ApexVariableContainer(name, value, '', ref));
+      }
+
       if (logContext.getStaticVariablesClassMap().has(className)) {
         map = logContext.getStaticVariablesClassMap().get(className)!;
         container = map.get(varName)! as ApexVariableContainer;
@@ -73,21 +80,30 @@ export class VariableAssignmentState implements DebugLogState {
         !container.type.startsWith('List<') &&
         !container.type.startsWith('Set<')
       ) {
-        if (!logContext.getRefsMap().has(ref)) {
-          logContext.getRefsMap().set(ref, container);
-        } else {
-          // if the ref already exists then that means we are assigning another pointer to a ref so we need to merge the info from the ref into the variable container
-          const refContainer = logContext.getRefsMap().get(ref)!;
-          container.variables = refContainer.variables;
-          container.variablesRef = refContainer.variablesRef;
-          container.name = varName;
-        }
+        // get ref container and map container's variables to the container's
+        const refContainer = logContext.getRefsMap().get(ref)!;
+        container.variables = refContainer.variables;
+        container.variablesRef = refContainer.variablesRef;
+        container.name = varName;
         if (value.indexOf('{') === 0 && value !== '{}') {
           container.value = '';
           container.variablesRef = logContext
             .getVariableHandler()
             .create(container);
-          this.parseAndPopulate(value, container);
+          this.parseAndPopulate(value, container, logContext);
+        } else if (logContext.getRefsMap().has(value)) {
+          const rc = logContext.getRefsMap().get(value)!;
+          const tmpContainer = new ApexVariableContainer(
+            varName,
+            rc.value,
+            rc.type,
+            rc.ref
+          );
+          tmpContainer.variables = rc.variables;
+          tmpContainer.variablesRef = logContext
+            .getVariableHandler()
+            .create(tmpContainer);
+          container.variables.set(varName, tmpContainer);
         }
         // assigning to variable's fields, currently only going to work for a variable in local
       } else if (name.indexOf('.') !== -1 && ref !== '0') {
@@ -105,7 +121,7 @@ export class VariableAssignmentState implements DebugLogState {
             topLevel.variablesRef = logContext
               .getVariableHandler()
               .create(topLevel);
-            this.parseAndPopulate(value, topLevel);
+            this.parseAndPopulate(value, topLevel, logContext);
           } else {
             container.variables.set(
               varName,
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,128 @@ export class VariableAssignmentState implements DebugLogState {
...
+            .set(ref, new ApexVariableContainer('', '', '', ref));
         }
-        if (value.indexOf('{') === 0 && value !== '{}') {
-          container.value = '';
-          container.variablesRef = logContext
-            .getVariableHandler()
-            .create(container);
-          this.parseAndPopulate(value, container);
+        const refContainer = refMap.get(ref)!;
+        if (value !== '{}') {
```
Sweetman: I refactored this a bit to arrange it in a way that made sense. Previously there were too many conditions to keep track of and it wasn't straight forward when you had a variable container and how you got it. I hope this is a little easier to follow

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,128 @@ export class VariableAssignmentState implements DebugLogState {
...
+            if (value.indexOf('{') === 0) {
+              const topLevel = new ApexVariableContainer(varName, '', '');
+              refContainer.variables.set(varName, topLevel);
+              topLevel.variablesRef = logContext
+                .getVariableHandler()
+                .create(topLevel);
+              this.parseJSONAndPopulate(value, topLevel, logContext);
+            } else {
+              if (refMap.has(value)) {
+                const pulledRef = refMap.get(value) as ApexVariableContainer;
```
Sweetman: Here is one location where we check if an assigned value could be a previously seen memory reference

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,128 @@ export class VariableAssignmentState implements DebugLogState {
...
+  ) {
     try {
       const obj = JSON.parse(value);
       Object.keys(obj).forEach(key => {
-        container.variables.set(
-          key,
-          new ApexVariableContainer(key, String(obj[key]), '')
-        );
+        if (logContext.getRefsMap().has(String(obj[key]))) {
+          const refContainer = logContext.getRefsMap().get(String(obj[key]))!;
```
Sweetman: Here is the other location where we check for a previously seen memory reference. This is in case we get a json value where key value pairs lead to an address

Sweetman:
I will get rid of the config changes since those are unnecessary.

Sweetman:
This PR includes the changes to use a previously seen memory address. After playing around with reference reassignments, there are some constraints with doing this due to the logs.

`public class RefAssignment {
    public Account instanceAcc1 = new Account(Name='instanceAcc1');
    public Account instanceAcc2 = new Account(Name='instanceAcc2');
    public Account instanceAcc3 = new Account(Name='instanceAcc3');

    public void method1() {
        instanceAcc3 = instanceAcc1;
        instanceAcc3.Name = 'changed in method2';
        System.debug(instanceAcc3.Name);
        System.debug(instanceAcc1.Name);
        System.debug('hi');
    }
}`

`10:59:31.59 (79079023)|VARIABLE_ASSIGNMENT|[30]|this|{"instanceAcc1":"0x16873bdc","instanceAcc2":"0x15e819ff","instanceAcc3":"0x7421eb84","m":"0x75a4d320"}|0x71ea6b4c
10:59:31.59 (79090256)|STATEMENT_EXECUTE|[30]
10:59:31.59 (79092245)|STATEMENT_EXECUTE|[31]
10:59:31.59 (79105651)|VARIABLE_ASSIGNMENT|[31]|this.instanceAcc3|{"Name":"instanceAcc1"}|0x71ea6b4c
10:59:31.59 (79114129)|STATEMENT_EXECUTE|[32]
10:59:31.59 (79117838)|HEAP_ALLOCATE|[32]|Bytes:18
10:59:31.59 (79173569)|VARIABLE_ASSIGNMENT|[32]|this.Name|"changed in method2"|0x16873bdc`

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
DatGuyJonathan: Does using `as string` work here? Also, it looks like you can cast to string once right after the forEach then re-use the string constant below.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
DatGuyJonathan: @Sweetman , would it help if you add some comments in this area so the next time you revisit, you're recognizing the scenarios instead of recalling?

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,141 @@ export class VariableAssignmentState implements DebugLogState {
...
         }
-        // assigning to variable's fields, currently only going to work for a variable in local
-      } else if (name.indexOf('.') !== -1 && ref !== '0') {
-        container = logContext.getRefsMap().get(ref)!;
-        if (container) {
+
+        // update the correct toplevel container
+        if (
+          container &&
+          !container.type.startsWith('Map<') &&
```
JimSuplizio: This just looks odd to me. We're looking to see whether or not the collection is generic which is going to have the pattern of <SomeString>, can't we just match that? I'd consider moving those 3 checks into an their own call.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
JimSuplizio: James, I'd previously asked you about this. Map.get is going to return undefined, instead of making a call to .has and then a call to .get, effectively double tapping the map, we should be able to make a call to get and check if refContainer is defined.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
Sweetman: Added

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,150 @@ export class VariableAssignmentState implements DebugLogState {
...
+              refContainer.variables.set(varName, topLevel);
+              topLevel.variablesRef = logContext
+                .getVariableHandler()
+                .create(topLevel);
+              this.parseJSONAndPopulate(value, topLevel, logContext);
+            } else {
+              // if it's not nested then we check if the value is a reference
+              if (refMap.has(value)) {
+                const pulledRef = refMap.get(value) as ApexVariableContainer;
+                const tmpContainer = new ApexVariableContainer(
```
DatGuyJonathan: I see you copy containers here and in `parseJSONAndPopulate`. Could you pull that out to a common method, maybe in ApexVariableContainer? If the parts that are copied change in the future, we can be rest assured both are changed.

vazexqi:
Just FYI. I made some changes to the way we use `String` vs `string` as part of my standardize tslint changes.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -26,106 +26,161 @@ export class VariableAssignmentState implements DebugLogState {
...
+          );
         }
-        // assigning to variable's fields, currently only going to work for a variable in local
-      } else if (name.indexOf('.') !== -1 && ref !== '0') {
-        container = logContext.getRefsMap().get(ref)!;
-        if (container) {
+
+        // update toplevel container if it's not this and not a collection
+        // or if the this variable has not been assigned a reference yet
+        if (
```
Sweetman: @DatGuyJonathan I just now added this in because I realized that the current frame's local `this` would be overwritten sometimes if nested variables were changed in a different object. This will just check that if we are updating `this`, the container is only updated if this is the first assignment.

PR #417, opened 2018-05-16
Bump vscode-debug dependencies to 1.28

DatGuyJonathan:
### What does this PR do?
Replay Debugger's VS Code dependency was bumped to 1.21 to detect breakpoint changes without an active debugger session. However, from 1.21 and above, VS Code changed their `DebugProtocol.SetBreakpointsArguments.source.path`. It used to give lower case Windows drive (e.g. c:\foo.cls). But now it gives C:\foo.cls.

That path is converted to a URI that is then validated against language server. The language server returns lower case Windows drive, so the upgrade to 1.21 broke the validation. Changing vscode-debugadapter version to 1.28 makes sure the debugger extension converts that path to the same URI provided by language server.

@W-4995585@


DatGuyJonathan:
Apex Interactive Debugger is already using vscode-debug* 1.28.

PR #419, opened 2018-05-16
Replace + in project name for bootstrap command

guw:
### What does this PR do?
Replace + sign in org name when creating the default project name.

W-4991435


DatGuyJonathan:
What's the underlying issue? Is the URI from LS something like `hello+there.cls`, but the debugger is using `hello%2Bthere`?

guw:
@DatGuyJonathan root cause identified and files as https://github.com/Microsoft/vscode-debugadapter-node/issues/182.

PR #420, opened 2018-05-16
Remember the last open log folder for log picker

JimSuplizio:
### What does this PR do?
There are a couple of things here
1. If the user has previously loaded a log to debug, save that path off. If the user starts a new debug session and is prompted to load a log then the getLogFileName will use that path as the default, if it still exists.
2. If the user has used the SFDX command to download logs, they'll downloaded to the .sfdx/tools/debug/logs sub-directory of the current workspace. If that directory exists and there's no previously loaded log folder, then we'll use that. Otherwise we'll stick with the .sfdx which was the previous default.

@W-4995493@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/index.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/index.ts
@@ -21,6 +21,8 @@ import {
   LINE_BREAKPOINT_INFO_REQUEST
 } from './constants';
 import { nls } from './messages';
+import pathExists = require('path-exists');
+let lastOpenedLogFolder: string | undefined;
 
 function registerCommands(): vscode.Disposable {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -20,6 +20,8 @@ import {
   LINE_BREAKPOINT_INFO_REQUEST
 } from './constants';
 import { nls } from './messages';
+import pathExists = require('path-exists');
```
DatGuyJonathan: Can this be `import * as pathExists from 'path-exists';`?

JimSuplizio: Done.

JimSuplizio:
@w-4995493@

PR #358, opened 2018-03-30
Check apex code and visualforce log levels on launch

DatGuyJonathan:
### What does this PR do?
Parse the log levels in a debug log and inform user if they don't meet the requirements of apex code and visualforce at the finest level.

### What issues does this PR fix or reference?
@W-4817382@

PR #423, opened 2018-05-17
Launch replay debugger for current log file

DatGuyJonathan:
### What does this PR do?
launchfromlogfile

@W-5004376@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-replay-debugger/package.json
@@ -78,7 +78,8 @@
       {
         "id": "apexlog",
         "aliases": ["ApexLog", "DebugLog"],
-        "extensions": [".log"]
+        "extensions": [".log"],
+        "firstLine": "^\\d{2}.*APEX_CODE,FINEST;.*VISUALFORCE,(FINER|FINEST);.*"
       }
     ],
     "grammars": [
@@ -128,7 +129,7 @@
       },
       {
         "command": "sfdx.launch.replay.debugger.logfile",
-        "title": "%launch_snippet_name%"
+        "title": "%launch_from_log_file%"
       }
     ],
     "views": {
--- a/packages/salesforcedx-vscode-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-replay-debugger/package.json
@@ -95,6 +95,10 @@
           "command": "sfdx.create.checkpoints",
           "when":
             "sfdx:project_opened && sfdx:replay_debugger_checkpoints_enabled"
+        },
+        {
+          "command": "sfdx.launch.replay.debugger.logfile",
+          "when": "sfdx:project_opened && editorLangId == 'apexlog'"
         }
       ],
       "view/title": [
@@ -108,7 +112,14 @@
       "editor/context": [
         {
           "command": "sfdx.launch.replay.debugger.logfile",
-          "when": "editorLangId == 'apexlog'"
+          "when": "sfdx:project_opened && editorLangId == 'apexlog'"
+        }
+      ],
+      "explorer/context": [
+        {
+          "command": "sfdx.launch.replay.debugger.logfile",
+          "when":
+            "sfdx:project_opened && resourceIsFile && resourceLangId == 'apexlog'"
         }
       ]
     },
--- a/packages/salesforcedx-vscode-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-replay-debugger/package.json
@@ -74,6 +74,20 @@
         "language": "apex"
       }
     ],
+    "languages": [
+      {
+        "id": "apexlog",
+        "aliases": ["ApexLog", "DebugLog"],
+        "extensions": [".log"]
+      }
+    ],
+    "grammars": [
+      {
+        "language": "apexlog",
+        "scopeName": "text.log",
+        "path": "./syntaxes/apexlog.tmLanguage.json"
+      }
+    ],
     "menus": {
       "commandPalette": [
         {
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-replay-debugger/package.nls.json
```diff
@@ -4,6 +4,7 @@
   "launch_snippet_description_text":
     "A new configuration for launching Apex Replay Debugger",
   "launch_snippet_name": "Launch Apex Replay Debugger",
+  "launch_from_log_file": "Launch Apex Replay Debugger with current file",
```
DatGuyJonathan: @ruthemmanuelle In the .GIF, this is the label for the menu item.

ruthemmanuelle: The other menu items use title-case capitalization, so I'd go with "Launch Apex Replay Debugger with Current File"—or, if you don't think it's too long, "Launch Apex Replay Debugger with Current Log File".

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -74,6 +74,21 @@
         "language": "apex"
       }
     ],
+    "languages": [
+      {
+        "id": "apexlog",
+        "aliases": ["ApexLog", "DebugLog"],
+        "extensions": [".log"],
```
vazexqi: I haven't had the chance to test this out but what happens to other files with .log extension? Does this also get applied there or does the `firstLine` attribute invalidate it?

I remember (and could be wrong) that VS Code calculate the langId of a file fuzzily by combining different criteria (extension, firstline, etc).

packages/salesforcedx-vscode-replay-debugger/syntaxes/apexlog.tmLanguage.json
```diff
@@ -0,0 +1,88 @@
+{
+  "version":
+    "https://github.com/Microsoft/vscode/blob/master/extensions/log/syntaxes/log.tmLanguage.json",
```
vazexqi: Could you also add a OSSREADME.md similar to https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-visualforce/OSSREADME.json at the top-level of packages/salesforcedx-vscode-replay-debugger?

Basically just link to the actual commit SHA where we took this version from. 

It's good for attribution and also to help us identify which SHA we took this file from.

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -105,6 +126,10 @@
         },
         "when":
           "sfdx:project_opened && sfdx:replay_debugger_checkpoints_enabled"
+      },
+      {
+        "command": "sfdx.launch.replay.debugger.logfile",
```
vazexqi: This command is **not** showing up in the palette right (just on right-click menu)?

If it is, then we need to check to ensure that if the user can invoke it, then we need more checks to see if there is an active editor and whether the contents of that active editor is a apexLog type.
It it is not, then we should be OK.

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -89,6 +104,12 @@
             "view == sfdx.force.view.checkpoint && sfdx:project_opened && sfdx:replay_debugger_checkpoints_enabled",
           "group": "navigation"
         }
+      ],
+      "editor/context": [
```
vazexqi: For completeness, should we also make this show up in [explorer/menu](https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributesmenus) when the user is selecting a .log file?
Or do we not have enough information to properly determine the langId of a file before it's opened?

packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -89,6 +104,12 @@
...
+        {
+          "command": "sfdx.launch.replay.debugger.logfile",
+          "when": "editorLangId == 'apexlog'"
```
vazexqi: Should we also check if this is a sfdx project? I am not sure if we can hook into that context from the salesforcedx-core extension. Maybe sfdx_project_opened is universal and applicable.

Could you check?

packages/salesforcedx-vscode-replay-debugger/package.json
DatGuyJonathan: I'll add some checks to make sure the text editor is apexlog type.

packages/salesforcedx-vscode-replay-debugger/package.json
DatGuyJonathan: `firstLine` doesn't invalidate it. I can still launch from context menu, but the launch will error out because of this other regex check we put in a while back: https://github.com/forcedotcom/salesforcedx-vscode/pull/358

packages/salesforcedx-vscode-replay-debugger/package.json
DatGuyJonathan: I can add to `explorer/menu`. `resourceLangId` looks feasible.

packages/salesforcedx-vscode-replay-debugger/package.json
DatGuyJonathan: Seems like `sfdx:project_opened` is universal. The checkpoints view uses that context as well.

packages/salesforcedx-vscode-replay-debugger/package.nls.json
DatGuyJonathan: Okay, I'm also going to prefix it with `SFDX:` because this command shows up in the command palette as well.

PR #424, opened 2018-05-17
Jsuplizio/version update and status bar

JimSuplizio:
### What does this PR do?

1. Add SFDX: Toggle Checkpoint command
2. Log the Create Checkpoints command to the output in the active replay debugger panel. Encapsulate the messages inside of start/end messages. The output will look as follows:
    Starting SFDX: Creating checkpoints
    Creating checkpoints: retrieving org info
    Creating checkpoints: retrieving source/line info
    Creating checkpoints: setting typeRefs for checkpoints
    Creating checkpoints: clearing existing checkpoints
    Creating checkpoints: uploading checkpoints
    Creating checkpoints: processing checkpoints completed successfully
    Ending SFDX: Creating checkpoints
3.  The call to vscode.window.withProgress needed an await to prevent it from immediately returning. Without the await it would return immediatel which causes the code in the finally to get executed and would ultimately allow multiple concurrent calls. This is exactly what I was trying to prevent.

@W-4979514, W-4992208, W-4949704@


packages/salesforcedx-vscode-replay-debugger/package.json
```diff
@@ -96,10 +96,15 @@
           "when":
             "sfdx:project_opened && sfdx:replay_debugger_checkpoints_enabled"
         },
+        {
+          "command": "sfdx.toggle.checkpoint",
+          "when":
+            "sfdx:project_opened && sfdx:replay_debugger_checkpoints_enabled && editorLangId==apex"
```
DatGuyJonathan: In other places, we have spaces before and after `==`, and the language ID is in single quotes.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -104,18 +107,20 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
       } catch (error) {
         const result = JSON.parse(error) as OrgInfoError;
-        vscode.window.showErrorMessage(
-          nls.localize('unable_to_retrieve_org_info') + ': ' + result.message
-        );
+        const errorMessage = `${nls.localize(
+          'unable_to_retrieve_org_info'
+        )} : ${result.message}`;
+        writeToDebuggerOutputWindow(errorMessage);
+        vscode.window.showErrorMessage(errorMessage);
```
DatGuyJonathan: Could we combine `writeToDebuggerOutputWindow(errorMessage);` and `vscode.window.showErrorMessage(errorMessage);` to a single method? I see these two paired up multiple times.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -714,28 +728,35 @@ export async function sfdxCreateCheckpoints() {
...
+    await vscode.window.withProgress(
       {
         location: vscode.ProgressLocation.Notification,
         title: localizedProgressMessage,
         cancellable: false
       },
       async (progress, token) => {
-        console.log('sfdxCreateCheckpoints: retrieving org info');
+        writeToDebuggerOutputWindow(
+          `Starting SFDX: ${localizedProgressMessage}`
```
DatGuyJonathan: I'm a bit confused why some messages are partly localized and partly hardcoded.

packages/salesforcedx-vscode-replay-debugger/package.json
JimSuplizio: It will be done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -774,15 +801,116 @@ export async function sfdxCreateCheckpoints() {
...
+    // There's already a breakpoint at this line
+    if (bp) {
+      // If the breakpoint is a checkpoint then remove it and return
+      if (
+        bp.condition &&
+        bp.condition!.toLowerCase().indexOf(CHECKPOINT) >= 0
+      ) {
+        bpRemove.push(bp);
+        await vscode.debug.removeBreakpoints(bpRemove);
+        return;
```
DatGuyJonathan: Can this be combined with the await?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: They aren't paired up everywhere but you're right, I'll add a boolean flag to writeToDebuggerOutputWindow that'll control whether or not to display an error message.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -774,15 +801,116 @@ export async function sfdxCreateCheckpoints() {
...
+  return;
+}
+
+// This methods was broken out of sfdxToggleCheckpoint for testing purposes.
+function fetchActiveEditorUri(): vscode.Uri | undefined {
+  const editor = vscode.window.activeTextEditor;
+  if (editor) {
+    return editor.document.uri;
+  }
+  return undefined;
```
DatGuyJonathan: This is implicit because of the return types of this method.

packages/salesforcedx-vscode-replay-debugger/src/index.ts
```diff
@@ -172,23 +183,36 @@ export async function retrieveLineBreakpointInfo(): Promise<boolean> {
...
     return false;
   }
 }
 
 function imposeSlightDelay(ms = 0) {
   return new Promise(r => setTimeout(r, ms));
 }
+
+const sfdxCoreExtension = vscode.extensions.getExtension(
+  'salesforce.salesforcedx-vscode-core'
```
DatGuyJonathan: You had questions about the core extension not being loaded before replay debugger extension. Are those concerns still valid? Does this have to be done elsewhere?

JimSuplizio: I had to change the way it was used, hence the way this was written. The pattern that other extensions are using is to declare the following:
const sfdxCoreExports = vscode.extensions.getExtension(
  'salesforce.salesforcedx-vscode-core'
)!.exports;
const channelService = sfdxCoreExports.channelService;
Notice the !.exports? I was originally doing that but it was causing problem that were surfacing when trying to run extension tests.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Sure, I'll remove it.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: Done.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -774,15 +853,119 @@ export async function sfdxCreateCheckpoints() {
...
+}
+exports.fetchActiveEditorUri = fetchActiveEditorUri;
+
+// This methods was broken out of sfdxToggleCheckpoint for testing purposes.
+function fetchActiveSelectionLineNumber(): number | undefined {
+  const editor = vscode.window.activeTextEditor;
+  if (editor && editor.selection) {
+    return editor.selection.start.line;
+  }
+  return undefined;
```
DatGuyJonathan: Could you also remove the implicit return from this method and the one below?

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -774,15 +853,119 @@ export async function sfdxCreateCheckpoints() {
...
+}
+
+// This methods was broken out of sfdxToggleCheckpoint for testing purposes.
+function fetchActiveEditorUri(): vscode.Uri | undefined {
+  const editor = vscode.window.activeTextEditor;
+  if (editor) {
+    return editor.document.uri;
+  }
+}
+exports.fetchActiveEditorUri = fetchActiveEditorUri;
```
Sweetman: Normally we use the `export` keyword with the function if it's something that needs to be shared across classes so line 932 would be something `function fetchActiveEditorUri(): vscode.Uri | undefined {`

This should let you call them directly without having to do `export.function`

packages/salesforcedx-vscode-replay-debugger/src/messages/i18n.ts
```diff
@@ -33,8 +33,20 @@ export const messages = {
...
+    'Step 2 of 6: Retrieving source and line information',
+  checkpoint_creation_status_setting_typeref:
+    'Step 3 of 6: Setting typeRefs for checkpoints',
+  checkpoint_creation_status_clearing_existing_checkpoints:
+    'Step 4 of 6: Clearing existing checkpoints',
+  checkpoint_creation_status_uploading_checkpoints:
+    'Step 5 of 6: Uploading checkpoints',
+  checkpoint_creation_status_processing_complete_success:
+    'Step 6 of 6: Confirming successful checkpoint creation',
+  sfdx_create_checkpoints_end: 'Ending SFDX: Create Checkpoints', // These strings are going to be re-worked to become better, Salesforce appropriate, error messages.
```
ruthemmanuelle: Maybe there should be a newline before the comment? The comment isn't related to this message.

JimSuplizio: Doh. Yeah, there should have been a newline. There is now.

packages/salesforcedx-vscode-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: As per our conversation this is the way it is in order to be able to sinon.stub the method for testing.

Sweetman:
LGTM aside from the return statements that Jonathan mentioned

JimSuplizio:
@w-4979514, W-4992208, W-4949704@

PR #425, opened 2018-05-17
Add note about disabling Live Share

ruthemmanuelle:
### What does this PR do?
Adds a note about disabling the Live Share extension before using the Apex Debugger.

@W-5001188@

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/README.md
+++ b/packages/salesforcedx-vscode-apex-debugger/README.md
@@ -15,6 +15,11 @@ Before you set up the Apex Debugger, make sure that you have these essentials.
 * [Visual Studio Code](https://code.visualstudio.com/download) v1.17 or later
 * The latest versions of the [salesforcedx-vscode-core](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode-core) and [salesforcedx-vscode-apex](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode-apex) extensions (we suggest that you install all extensions in the [salesforcedx-vscode](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode) extension pack)
 
+---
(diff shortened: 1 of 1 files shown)


lostintangent:
Instead of users needing to disable the Live Share extension, you could alternatively instruct them to set the “liveshare.excludedDebugTypes” setting to “[‘apex-replay’]”. That setting prevents Live Share from injecting itself into the specified debug adapters, without needing to disable Live Share entirely. That way, users could still share their projects, co-edit, use shared terminals, etc., but be unblocked to debug their apps locally.

packages/salesforcedx-vscode-apex-debugger/README.md
```diff
@@ -15,6 +15,13 @@ Before you set up the Apex Debugger, make sure that you have these essentials.
 * [Visual Studio Code](https://code.visualstudio.com/download) v1.17 or later
 * The latest versions of the [salesforcedx-vscode-core](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode-core) and [salesforcedx-vscode-apex](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode-apex) extensions (we suggest that you install all extensions in the [salesforcedx-vscode](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode) extension pack)
 
+---
+**NOTE:** Having an active installation of the Visual Studio Live Share extension stops the Apex Debugger from working. Before you use the Apex Debugger, set the `liveshare.excludedDebugTypes` setting to `["apex-replay"]`. For instructions, see [User and Workspace Settings](https://code.visualstudio.com/docs/getstarted/settings) in the Visual Studio Code docs.
```
vazexqi: We need to include the Apex Interactive Debugger as well. You would need to include https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-apex-debugger/package.json#L79

So, it should be `["apex-replay", "apex"]`

As an update, I'm still investigating the fix from Microsoft but haven't had a chance to do that yet. 

We can use this file and if I make progress we can just take this out from the release/v42.17.0 branch.

PR #427, opened 2018-05-19
Standardize tslint across SFDX projects

vazexqi:
### What does this PR do?

Uses the standard tslint rules for Salesforce DX projects.

@W-4956792@

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -1,83 +1,11 @@
-// From https://github.com/palantir/tslint/blob/master/docs/sample.tslint.json
 {
+  "extends": "@salesforce/dev-config/tslint",
   "rules": {
-    "class-name": true,
-    "curly": true,
-    "forin": true,
-    "jsdoc-format": true,
-    "label-position": true,
-    "member-access": true,
-    "new-parens": true,
+    "no-reserved-keywords": false,
     "no-any": false,
-    "no-arg": true,
-    "no-bitwise": true,
-    "no-conditional-assignment": true,
-    "no-consecutive-blank-lines": false,
-    "no-console": [true, "debug", "info", "time", "timeEnd", "trace"],
-    "no-construct": true,
-    "no-debugger": true,
-    "no-duplicate-variable": true,
-    "no-empty": true,
-    "no-eval": true,
-    "no-inferrable-types": [true, "ignore-params"],
-    "no-internal-module": true,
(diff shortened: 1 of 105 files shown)


tslint.json
```diff
@@ -1,83 +1,11 @@
...
-    "whitespace": [
-      true,
-      "check-branch",
-      "check-decl",
-      "check-operator",
-      "check-separator",
-      "check-type"
-    ]
+    "quotemark": false,
+    "member-ordering": false,
```
vazexqi: This turns off to create too much noise. We need to disable this.

tslint.json
```diff
@@ -1,83 +1,11 @@
...
   "rules": {
-    "class-name": true,
-    "curly": true,
-    "forin": true,
-    "jsdoc-format": true,
-    "label-position": true,
-    "member-access": true,
-    "new-parens": true,
+    "no-reserved-keywords": false,
     "no-any": false,
```
vazexqi: This one also created too much noise. We should be taken care of by our strict rules in tslint.json

tslint.json
```diff
@@ -1,83 +1,11 @@
...
-    ],
-    "whitespace": [
-      true,
-      "check-branch",
-      "check-decl",
-      "check-operator",
-      "check-separator",
-      "check-type"
-    ]
+    "quotemark": false,
```
vazexqi: This one interferes with prettier.

tslint.json
```diff
@@ -1,83 +1,11 @@
...
+    "member-ordering": false,
+    "no-string-based-set-timeout": false,
```
vazexqi: I think there is a bug with this rule.

tslint.json
```diff
@@ -1,83 +1,11 @@
...
+    "no-floating-promises": false
```
vazexqi: This is temporary while I try some things out.

vazexqi:
There were some reordering due to having `member-ordering: true` (it's now set to `false` since there was too much churn) in some of the earlier iterations. I left those changes in since I think those small ones do provide value in ordering.

vazexqi:
@DatGuyJonathan - These are mostly tslint changes. You don't have to go through each change. This is just a formality to ensure that you can see what has changed.

I've tested it manually (in addition to the automated tests). I'm going to merge this in.

PR #428, opened 2018-05-21
Upgrade vscode dev-dependency to 1.1.17

vazexqi:
### What does this PR do?

Brings in https://github.com/Microsoft/vscode-extension-vscode/issues/106

@W-4979583@


vazexqi:
I won't be able to test the `vscode:publish` until when we actually release but I can let CircleCI take care of running `vscode:package` to ensure that nothing is wrong.

PR #429, opened 2018-05-21
Single quote strings and refactor common code

Sweetman:
### What does this PR do?
Changes strings being shown in double quotes to single quotes and refactors the copying of a reference container that was lost in a rebase
@W-4995429@

### DIFF
--- a/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
@@ -29,10 +29,7 @@ export class VariableAssignmentState implements DebugLogState {
         name.indexOf('.') > -1 ? name.substring(0, name.lastIndexOf('.')) : '';
       const varName =
         nameSplit.length > 0 ? nameSplit[nameSplit.length - 1] : name;
-      let value = this.fields[4].replace('(^")|("$)', "'");
-      if (value.startsWith('"')) {
-        value = "'" + value.substring(1, value.length - 1) + "'";
-      }
+      const value = this.fields[4].replace(/^"/, "'").replace(/"$/, "'");
       let ref;
       if (this.fields.length === 6) {
         ref = this.fields[5];
--- a/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
+++ b/packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
@@ -29,7 +29,10 @@ export class VariableAssignmentState implements DebugLogState {
         name.indexOf('.') > -1 ? name.substring(0, name.lastIndexOf('.')) : '';
       const varName =
         nameSplit.length > 0 ? nameSplit[nameSplit.length - 1] : name;
-      const value = this.fields[4];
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -29,7 +29,10 @@ export class VariableAssignmentState implements DebugLogState {
         name.indexOf('.') > -1 ? name.substring(0, name.lastIndexOf('.')) : '';
       const varName =
         nameSplit.length > 0 ? nameSplit[nameSplit.length - 1] : name;
-      const value = this.fields[4];
+      let value = this.fields[4].replace('(^")|("$)', "'");
+      if (value.startsWith('"')) {
```
Sweetman: If the value is just primitive then we see if it starts with a double quote and swap it for single quotes.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -164,21 +162,20 @@ export class VariableAssignmentState implements DebugLogState {
...
+            logContext
           );
-          tmpContainer.variables = refContainer.variables;
-          tmpContainer.variablesRef = logContext
-            .getVariableHandler()
-            .create(tmpContainer);
           container.variables.set(key, tmpContainer);
         } else {
+          let varValue = obj[key];
+          if (typeof varValue === 'string') {
```
Sweetman: If the value is in a json object, check if the parsed value is a string and then append single quotes

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -188,4 +185,21 @@ export class VariableAssignmentState implements DebugLogState {
       container.variables.clear();
     }
   }
+  private copyReferenceContainer(
```
Sweetman: This refactoring was lost while rebasing my last PR.

packages/salesforcedx-vscode-replay-debugger/src/states/variableAssignmentState.ts
DatGuyJonathan: Can this if statement be true if the double quote is replaced by the line before this?

Sweetman: Oh I added the if check because the replace regex wasn't working. I am going to remove the replace call.

DatGuyJonathan: You can probably do this `replace(/^"(.*)"$/, "'$1'")`

Sweetman: Going to use `.replace(/^"/, "'").replace(/"$/, "'")`  as we discussed since the `$` syntax is not a standard feature
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/n

PR #430, opened 2018-05-22
Rename replay debugger

DatGuyJonathan:
### What does this PR do?
Changes the extension name. I'm not changing the folder name yet; I think that can be done when the adapter is separated from the extension.

@W-5014405@

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -356,7 +356,7 @@
       "request": "launch",
       "cwd": "${workspaceRoot}",
       "program":
-        "${workspaceRoot}/packages/salesforcedx-vscode-replay-debugger/out/src/adapter/apexReplayDebug.js",
+        "${workspaceRoot}/packages/salesforcedx-vscode-apex-replay-debugger/out/src/adapter/apexReplayDebug.js",
       "args": ["--server=4712"],
       "sourceMaps": true,
       "outFiles": ["${workspaceRoot}/packages/*/out/src/**/*.js"],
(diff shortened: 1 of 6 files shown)


.vscode/launch.json
```diff
@@ -356,7 +356,7 @@
       "request": "launch",
       "cwd": "${workspaceRoot}",
       "program":
-        "${workspaceRoot}/packages/salesforcedx-vscode-replay-debugger/out/src/adapter/apexReplayDebug.js",
+        "${workspaceRoot}/packages/salesforcedx-vscode-apex-replay-debugger/out/src/adapter/apexReplayDebug.js",
```
vazexqi: @DatGuyJonathan - I'm confused. You said that you're not changing the name of the folder yet but I see that you change the name here in launch.json so the folder name needs to change as well or else we won't be able to launch anymore?

DatGuyJonathan: Oops, reverting.

vazexqi:
@DatGuyJonathan - I have a question after thinking about this a bit more.

Could you elaborate more on why shouldn't we just rename the folder as well to salesforcedx-vscode-apex-replay-debugger? Even if we decide to split out the adapter, won't the name just be salesforcedx-apex-replay-debugger (without the vscode in it just like how we treat the interactive debugger adapter folder)?

Asking because right now it feels a bit weird with the folder name being out of sync with the package.json file.

vazexqi:
Got it. Thanks for elaborating. Let's do it after those changes are in then.

PR #431, opened 2018-05-22
Cleanup temp. files when done

guw:
### What does this PR do?
This changes the ISV bootstrap command to clean up any temp. files that were created/downloaded as part of retrieving and converting code.

@W-4963327@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
@@ -434,6 +434,20 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       }
     }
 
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -434,6 +434,20 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       }
     }
 
+    // 7c: cleanup temp files
+    try {
+      shell.rm('-rf', path.join(projectMetadataTempPath));
```
DatGuyJonathan: I don't think this needs a path.join. It should already be a valid path:

```
const projectMetadataTempPath = path.join(
       projectPath,
       this.relativeMetdataTempPath
     );
```

PR #437, opened 2018-05-24
Turn off logging will delete the traceflag

Sweetman:
### What does this PR do?
Delete the traceflag that was created/updated by the turn on command instead of updating the levels to whatever they were set to previously.
@W-4715221@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
@@ -77,7 +77,7 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
         const userId = await getUserId(
           vscode.workspace.workspaceFolders![0].uri.fsPath
         );
-        developerLogTraceFlag.createTraceFlagInfo();
+        developerLogTraceFlag.validateDates();
         resultJson = await this.subExecute(new CreateTraceFlag(userId).build());
         developerLogTraceFlag.setTraceFlagId(resultJson.result.id);
       }
@@ -168,7 +168,7 @@ export class UpdateDebugLevelsExecutor extends SfdxCommandletExecutor<{}> {
     return new SfdxCommandBuilder()
       .withArg('force:data:record:update')
       .withFlag('--sobjecttype', 'DebugLevel')
-      .withFlag('--sobjectid', developerLogTraceFlag.getDebugLevelId())
+      .withFlag('--sobjectid', developerLogTraceFlag.getDebugLevelId()!)
       .withFlag(
         '--values',
         `ApexCode=${APEX_CODE_DEBUG_LEVEL} Visualforce=${VISUALFORCE_DEBUG_LEVEL}`
@@ -184,7 +184,7 @@ export class UpdateTraceFlagsExecutor extends SfdxCommandletExecutor<{}> {
     return new SfdxCommandBuilder()
       .withArg('force:data:record:update')
       .withFlag('--sobjecttype', 'TraceFlag')
-      .withFlag('--sobjectid', developerLogTraceFlag.getTraceFlagId())
+      .withFlag('--sobjectid', developerLogTraceFlag.getTraceFlagId()!)
       .withFlag(
         '--values',
         `StartDate='${developerLogTraceFlag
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -61,9 +61,7 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
           traceflag.Id,
           traceflag.StartDate,
           traceflag.ExpirationDate,
-          traceflag.DebugLevelId,
-          traceflag.DebugLevel.ApexCode,
-          traceflag.DebugLevel.Visualforce
+          traceflag.DebugLevelId
```
Sweetman: No longer need to store previous levels for apex/visualforce since we are deleting the traceflag

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -74,7 +72,7 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
       } else {
         resultJson = await this.subExecute(new CreateDebugLevel().build());
         const debugLevelId = resultJson.result.id;
-        developerLogTraceFlag.setDebugLevelInfo(debugLevelId);
+        developerLogTraceFlag.setDebugLevelId(debugLevelId);
```
Sweetman: Renaming since only the id is relevant now

packages/salesforcedx-vscode-core/src/commands/forceStopApexDebugLogging.ts
```diff
@@ -65,16 +65,12 @@ export async function restoreDebugLevels(): Promise<void> {
...
-    .withArg('force:data:record:update')
-    .withFlag('--sobjecttype', 'DebugLevel')
-    .withFlag('--sobjectid', developerLogTraceFlag.getDebugLevelId())
-    .withFlag(
-      '--values',
-      `ApexCode=${developerLogTraceFlag.getPrevApexCodeDebugLevel()} Visualforce=${developerLogTraceFlag.getPrevVFCodeDebugLevel()}`
-    )
+    .withArg('force:data:record:delete')
+    .withFlag('--sobjecttype', 'TraceFlag')
+    .withFlag('--sobjectid', developerLogTraceFlag.getTraceFlagId()!)
```
Sweetman: Changing command to delete the traceflag instead of updating the levels

Sweetman: Using the `!` since because we know it's set because to reach this it passes the parameter gatherer.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
```diff
@@ -32,38 +31,21 @@ export class DeveloperLogTraceFlag {
...
-    );
-  }
-
   public setTraceFlagDebugLevelInfo(
     id: string,
     startDate: string,
     expirationDate: string,
-    debugLevelId: string,
-    oldApexCodeDebugLevel: string,
-    oldVFDebugLevel: string
```
Sweetman: No longer need to keep track of previous levels

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -186,7 +184,7 @@ export class UpdateTraceFlagsExecutor extends SfdxCommandletExecutor<{}> {
     return new SfdxCommandBuilder()
       .withArg('force:data:record:update')
       .withFlag('--sobjecttype', 'TraceFlag')
-      .withFlag('--sobjectid', developerLogTraceFlag.getTraceFlagId())
+      .withFlag('--sobjectid', developerLogTraceFlag.getTraceFlagId()!)
```
vazexqi: Could you comment on why the `!` is safe here?
I like to keep track of when we explicitly cast something that can be null|undefined.

packages/salesforcedx-vscode-core/src/commands/forceStopApexDebugLogging.ts
vazexqi: Same with the explicit `!` here.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
Sweetman: The traceflagId should be set in the execute method prior to calling this executor when running the query executor. If the field is not present in the JSON  that it's parsing, then it should catch the error and surface a failure to the user.

packages/salesforcedx-vscode-core/src/commands/forceStopApexDebugLogging.ts
Sweetman: Similar to the UpdateTraceFlag, this is called after the query executor and so the id should be set before calling. Otherwise it should surface an error to the user

PR #439, opened 2018-05-25
Add command to debug last used log

JimSuplizio:
### What does this PR do?
Added the command "SFDX: Launch Apex Replay Debugger with Last Log File" which will start a new debugging session with the last log debugged. Note: This command is only available if there isn't an active debug session. The reason for this is that the Restart Debugging command (through the menu option or through the button on the debug panel) will already restart the session with the same parameters. 

As part of this fix, store the last debugged log and the last debugged log directory in the workspaceState Memento. This will persist both variables for the workspace similar to the way breakpoints are persisted which is to say if you were to unload/reload the same project we'd still have the last log and last log directory if they were previously set (otherwise we'd prompt).

Also, fixed an issue with SFDX: Launch Apex Replay Debugger with Current File where it didn't update the last debugged log/directory. (The scenario: Debug Log A through the normal means. Open up Log B through the Open File menu and select to debug with current file. Stop debugging and then debug with last log file and instead of opening up B again it would open up A and start debugging). 

@W-5001363@

### DIFF
--- a/packages/salesforcedx-vscode-apex-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/package.json
@@ -106,6 +106,10 @@
         {
           "command": "sfdx.launch.replay.debugger.logfile",
           "when": "sfdx:project_opened && editorLangId == 'apexlog'"
+        },
+        {
+          "command": "sfdx.launch.replay.debugger.last.logfile",
+          "when": "sfdx:project_opened && !inDebugMode"
         }
       ],
       "view/title": [
@@ -152,6 +156,10 @@
       {
         "command": "sfdx.launch.replay.debugger.logfile",
         "title": "%launch_from_log_file%"
+      },
+      {
+        "command": "sfdx.launch.replay.debugger.last.logfile",
+        "title": "%launch_from_last_log_file%"
       }
     ],
     "views": {
--- a/packages/salesforcedx-vscode-apex-replay-debugger/package.nls.json
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/package.nls.json
@@ -5,6 +5,8 @@
     "A new configuration for launching Apex Replay Debugger",
   "launch_snippet_name": "Launch Apex Replay Debugger",
   "launch_from_log_file": "SFDX: Launch Apex Replay Debugger with Current File",
+  "launch_from_last_log_file":
+    "SFDX: Launch Apex Replay Debugger with Last Log File",
   "logfile_text":
     "Path to an Apex Debug Log file (*.log) in the workspace folder.",
   "stop_on_entry_text":
--- a/packages/salesforcedx-vscode-apex-replay-debugger/src/commands/launchFromLogFile.ts
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/src/commands/launchFromLogFile.ts
@@ -8,7 +8,7 @@
 import * as vscode from 'vscode';
 import { DebugConfigurationProvider } from '../adapter/debugConfigurationProvider';
 
-export function launchFromLogFile(editorUri: vscode.Uri) {
+export function launchFromLogFile(logFile?: string) {
   if (
     !vscode.debug.activeDebugSession &&
     vscode.workspace.workspaceFolders &&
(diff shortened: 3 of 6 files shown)


packages/salesforcedx-vscode-apex-replay-debugger/package.json
```diff
@@ -106,6 +106,10 @@
         {
           "command": "sfdx.launch.replay.debugger.logfile",
           "when": "sfdx:project_opened && editorLangId == 'apexlog'"
+        },
+        {
+          "command": "sfdx.launch.replay.debugger.last.logfile",
+          "when": "sfdx:project_opened && !inDebugMode"
```
JimSuplizio: This command is available when you're in an SFDX project and not already debugging something. If you're already debugging something the Restart Debugging commands effectively do the same thing, restart your debugging session with the same parameters.

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/launchFromLogFile.ts
```diff
@@ -8,15 +8,15 @@
 import * as vscode from 'vscode';
 import { DebugConfigurationProvider } from '../adapter/debugConfigurationProvider';
 
-export function launchFromLogFile(editorUri: vscode.Uri) {
+export function launchFromLogFile(logFile?: string) {
```
JimSuplizio: DebugConfigurationProvider.getConfig only needs the log file name. Change the argument from the URI to the file name and it can be used for both "SFDX: Launch Apex Replay Debugger with Current File" and "SFDX: Launch Apex Replay Debugger with Last Log File" commands.

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -23,7 +23,9 @@ import {
   LINE_BREAKPOINT_INFO_REQUEST
 } from './constants';
 import { nls } from './messages';
-let lastOpenedLogFolder: string | undefined;
+let extContext: vscode.ExtensionContext;
+const LAST_OPENED_LOG_KEY = 'LAST_OPENED_LOG_KEY';
```
DatGuyJonathan: Could you add this to constants.ts?

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -44,24 +46,53 @@ function registerCommands(): vscode.Disposable {
         defaultUri: getDialogStartingPath()
       });
       if (fileUris && fileUris.length === 1) {
-        lastOpenedLogFolder = path.dirname(fileUris[0].fsPath);
+        extContext.workspaceState.update(
```
DatGuyJonathan: If you extract this out to its own method, then (1) you can re-use it in `launchFromLogFileCmd` below and (2) you can add a little test for it. The method can take an ExtensionContext and a string path.

Your test can do something like https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lwc-next/test/eslint.test.ts#L16. You're defining a context object and its update method which can simply add whatever key value it gets to an array or something. Then you assert the array contains LAST_OPENED_LOG_KEY, LAST_OPENED_LOG_FOLDER_KEY and their values. The test will verify the production method uses the context object to save those key value pairs.

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
JimSuplizio: It will be done.

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/launchFromLogFile.ts
JimSuplizio: Done, as per requested

packages/salesforcedx-vscode-apex-replay-debugger/package.nls.json
```diff
@@ -5,6 +5,8 @@
     "A new configuration for launching Apex Replay Debugger",
   "launch_snippet_name": "Launch Apex Replay Debugger",
   "launch_from_log_file": "SFDX: Launch Apex Replay Debugger with Current File",
+  "launch_from_last_log_file":
+    "SFDX: Launch Apex Replay Debugger with Last Log File",
```
vazexqi: @ruthemmanuelle - Could you review, if Jim hasn't pinged you offline already? Thanks!

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
vazexqi: Wow, I haven't used this API so I could be wrong. But these calls are synchronous (instead of asynchronous and returning a promise)? 

Interesting. The updates to the keystore must be pretty fast then.

PR #442, opened 2018-05-25
Fix live share bug

DatGuyJonathan:
### What does this PR do?
If LiveShare is enabled, custom events from a debug adapter to the debugger extension goes through LiveShare so we need to detect the underlying debugger type. More details in `https://github.com/MicrosoftDocs/live-share/issues/396`.

livesharefix

@W-5013801@


DatGuyJonathan:
Their ETA is 5/29. I'll monitor the original github issue for their update and merge after they give the green light.

vazexqi:
@DatGuyJonathan - Corret me if I am wrong – this seems like it should be safe to merge in with/without the changes from LiveShare. 

The only real thing we need to change is the note we have in the CHANGELOG.md from https://github.com/forcedotcom/salesforcedx-vscode/pull/425 once they have merged this in.

vazexqi:
Per https://github.com/MicrosoftDocs/live-share/issues/396#issuecomment-392938107

@ruthemmanuelle - I will submit a change to the README.md as well when I write the release notes for v42.18.0 this week.

PR #447, opened 2018-05-28
Separate replay debugger adapter from extension

DatGuyJonathan:
### What does this PR do?
- Separates the adapter from the extension, similar to Apex Interactive Debugger
- The extension imports `salesforcedx-utils-vscode` through `salesforcedx-apex-replay-debugger`
- Fixes copyright blocks in some checkpoint related files
- Removes some checkpoint messages and constants that aren't used anywhere

@W-4750398@


packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -336,6 +336,7 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
               if (deleteResult) {
                 const result = JSON.parse(deleteResult) as BatchDeleteResponse;
                 if (result.hasErrors) {
+                  // What is this message?
                   const errorMessage = nls.localize(
                     'cannot_delete_existing_overlay_action'
```
DatGuyJonathan: @JimSuplizio , what is this message supposed to be? It doesn't exist in i18n.ts.

JimSuplizio: @DatGuyJonathan , that should be the constant from i18n.ts cannot_delete_existing_checkpoint. It looks like it was overlooked back when messages had overlay

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/apexExecutionOverlayActionCommand.ts
```diff
@@ -46,12 +46,10 @@ export class ApexExecutionOverlayActionCommand extends BaseCommand {
     }
   }
 
-  // For this particular message the query string should be undefined
```
JimSuplizio: For ApexExecutionOverlayActionCommand and BatchDeleteExistingOverlayActionCommand you deleted the comment about the query strings being undefined, why? That's a perfectly valid comment that describes part of the behavior of both commands.

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/apexExecutionOverlayActionCommand.ts
```diff
@@ -22,7 +23,6 @@ export interface ApexExecutionOverlaySuccessResult {
 export class ApexExecutionOverlayActionCommand extends BaseCommand {
   private readonly apiPath = 'services/data/v43.0/tooling/sobjects';
```
vazexqi: Could you extract this into a constant? It would be very weird if this is the only one that uses 43.0 when everything else is updated later.

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/apexExecutionOverlayActionCommand.ts
DatGuyJonathan: Because I was able to read the code which says you're passing in undefined to the parent constructor. That undefined argument is the query string. In fact, I think I'll remove `getQueryString()` all together because you're not doing anything different than baseCommand.ts

PR #448, opened 2018-05-29
only have one create checkpoints command

JimSuplizio:
### What does this PR do?
When I'd added the button to the checkpoints view I'd inadvertently added a second create checkpoints command that would show up in the command palette. Consolidate the commands so there's only one. 

There was one message that needed to get fixed in the checkpointService. There were two checkpoint_creation_status_source_line_info messages and the second one should have been checkpoint_creation_status_setting_typeref.

Updated "SFDX: Create Checkpoints" command to "SFDX: Update Checkpoints in Org"

@W-5001294, W-5000736@


packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -79,11 +79,31 @@ function registerCommands(): vscode.Disposable {
...
+
+  if (checkpointsEnabled) {
+    const sfdxCreateCheckpointsCmd = vscode.commands.registerCommand(
+      'sfdx.create.checkpoints',
+      sfdxCreateCheckpoints
+    );
+    const sfdxToggleCheckpointCmd = vscode.commands.registerCommand(
+      'sfdx.toggle.checkpoint',
+      sfdxToggleCheckpoint
+    );
```
Sweetman: I think you could remove this check to register the command by attaching a when clause to the context being set for checkpoints in package.json. We do this as well to verify sfdx projects are opened and if the replay extension is installed

JimSuplizio: @Sweetman This won't be an issue when checkpoints are fully enabled. Right now I'm just leaving this the way it is.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -771,18 +771,18 @@ export async function sfdxCreateCheckpoints() {
   // The status message isn't changing, call to localize it once and use the localized string in the
   // progress report.
   const localizedProgressMessage = nls.localize(
-    'sfdx_create_checkpoints_start'
+    'sfdx_update_checkpoints_in_org'
   );
   // Wrap everything in a try/finally to ensure creatingCheckpoints gets set to false
   try {
+    writeToDebuggerOutputWindow(`Starting ${localizedProgressMessage}`);
```
DatGuyJonathan: The core extension uses https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/messages/i18n.ts#L20. Can you add yours to i18n.ts also?

JimSuplizio: It will be done.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -858,7 +858,7 @@ export async function sfdxCreateCheckpoints() {
       }
     );
   } finally {
-    writeToDebuggerOutputWindow(nls.localize('sfdx_create_checkpoints_end'));
+    writeToDebuggerOutputWindow(`Ending ${localizedProgressMessage}`);
```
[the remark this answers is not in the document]
JimSuplizio: It will be done.

PR #451, opened 2018-05-31
Introduce salesforcedx-webview-ui package

vazexqi:
### What does this PR do?

This is the package that will contain all the .html that we will use for each webview. We will use this to create webviews that will be surfaced in VS Code. This will be based on the approach documented at https://code.visualstudio.com/docs/extensions/webview

@W-5013822@


packages/salesforcedx-webview-ui/README.md
```diff
@@ -0,0 +1,122 @@
+# Introduction
```
vazexqi: This is the main file that you should read to get the gist of this PR.
The other parts I will illustrate to the team via an internal demo.

vazexqi: @lcampos - Would appreciate your thoughts on this. 

We can still merge this in and we can continue the discussion later but it would be appreciated if you could share some thoughts on this based on your work with sutro - I'm not sure if you were using webviews there (or browserview).

Either way, the links in the README.md should be useful.

packages/salesforcedx-webview-ui/README.md
```diff
@@ -0,0 +1,187 @@
...
+* [react-script-ts](https://github.com/wmonk/create-react-app-typescript) - Provides the template for the project structure
+* [create-react-app](https://github.com/facebook/create-react-app) - The basis for the project structure. Read this to understand how things work underneath.
+* [Blueprint](http://blueprintjs.com/docs/v2/) - For UI components
+* [Jest](https://facebook.github.io/jest/) - For testing
+
+# Useful Reading
+
+* [`<webview>` Tag](https://electronjs.org/docs/api/webview-tag)
+* [Interop’s Labyrinth: Sharing Code Between Web & Electron Apps](https://slack.engineering/interops-labyrinth-sharing-code-between-web-electron-apps-f9474d62eccc)
+* [Growing Pains: Migrating Slack’s Desktop App to BrowserView](https://slack.engineering/growing-pains-migrating-slacks-desktop-app-to-browserview-2759690d9c7b)
```
vazexqi: There are some bugs with webview in Electron. See https://github.com/electron/electron/issues?q=is%3Aopen+is%3Aissue+label%3Acomponent%2Fwebview

However, I searched on the VS Code repo and there were [no](https://github.com/Microsoft/vscode/issues?utf8=%E2%9C%93&q=browserview) discussions about considering using `BrowserView` instead of `WebView`. So that option doesn't seem to exist for VS Code.

packages/salesforcedx-webview-ui/README.md
lcampos: I did look at using webviews but I was able to make it work by using TextDocumentContentProvider to render my html/js. I didn't need to modify anything outside of the html/js files I was rendering.

lcampos: Ok, seems WebViews is the "new" way to go based on this https://code.visualstudio.com/updates/v1_21#_webview-api

packages/salesforcedx-webview-ui/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Developer Flow
+
+## Rapid Iteration of the UI
```
lcampos: This is currently just showing me a blank page, is it expected ? I do see the manifest editor when I run the extensions tho.

packages/salesforcedx-webview-ui/config/webpack.config.dev.js
```diff
@@ -0,0 +1,266 @@
...
+          // In production, they would get copied to the `build` folder.
+          // This loader doesn't use a "test" so it will catch all modules
+          // that fall through the other loaders.
+          {
+            // Exclude `js` files to keep "css" loader working as it injects
+            // its runtime that would otherwise processed through "file" loader.
+            // Also exclude `html` and `json` extensions so they get processed
+            // by webpacks internal loaders.
+            exclude: [/\.(js|jsx|mjs)$/, /\.html$/, /\.json$/],
+            loader: require.resolve('file-loader'),
```
lcampos: Did you try something like 
{ test: /\.(jpe?g|png|gif|svg|woff|eot|ttf)$/i, loader: 'file-loader?name=/public/[name].[ext]' }
? I have it working like that in another project but it might be because I'm on a more recent version of file-loader.

packages/salesforcedx-webview-ui/package.json
```diff
@@ -0,0 +1,118 @@
...
+  "name": "@salesforce/salesforcedx-webview-ui",
+  "description": "UI components for Salesforce Extensions for VS Code",
+  "version": "42.18.0",
+  "publisher": "salesforce",
+  "license": "BSD-3-Clause",
+  "private": true,
+  "engines": {
+    "vscode": "^1.17.0"
+  },
+  "dependencies": {
```
lcampos: Everything webpack, loaders, jest etc should be under devDependencies.

packages/salesforcedx-webview-ui/package.json
```diff
@@ -0,0 +1,118 @@
...
+    "style-loader": "0.19.0",
+    "ts-jest": "22.0.1",
+    "ts-loader": "2.3.7",
+    "tsconfig-paths-webpack-plugin": "2.0.0",
+    "tslint": "5.7.0",
+    "tslint-config-prettier": "1.10.0",
+    "tslint-react": "3.2.0",
+    "uglifyjs-webpack-plugin": "1.1.8",
+    "url-loader": "0.6.2",
+    "webpack": "3.8.1",
```
lcampos: 😍

packages/salesforcedx-webview-ui/package.json
vazexqi: Interesting. I got this from create-react-app. This was the section that they put it in. I can move them into dev-dependencies.

packages/salesforcedx-webview-ui/config/webpack.config.dev.js
vazexqi: Will try. I wasn't sure if you could pack svg|woff|eot|ttf. That would help.

packages/salesforcedx-webview-ui/README.md
vazexqi: Did you navigate to http://localhost:3000/ManifestEditor

packages/salesforcedx-webview-ui/src/entries/ManifestEditor/index.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
```
lcampos: We should change this to import { render } from 'react-dom'; so it will only import the render method and nothing else.

packages/salesforcedx-webview-ui/README.md
lcampos: That was it, thanks

packages/salesforcedx-webview-ui/package.json
lcampos: I guess it depends on how we are bundling the extension, are we keeping them as part of node_modules when creating the extension?

vazexqi: Nope, we only bundle the stuff that is optimized in the /build folder. So nothing from node_modules leaks in. Everything has to be bundled in a way that you can just load the .html/.js/etc. All the JS is bundled as UMD, not That's why I think it's OK and that is why create-react-app did it that way. I will investigate more but I don't think I change this from how they were doing it.

packages/salesforcedx-webview-ui/src/entries/ManifestEditor/index.tsx
vazexqi: Doesn't the tree shaking and dead-code elimination from webpack solve the issue of bringing too much? Or am I overestimating what it can do?

lcampos: Tree shaking will remove that, but I didn't see anything for that on the webpack config files. Is babel compiling to ES6? If it is then webpack will automatically do tree shaking.

vazexqi: https://github.com/forcedotcom/salesforcedx-vscode/blob/e84a5178d55e911c2ba489598d514d88ad039132/packages/salesforcedx-webview-ui/config/webpack.config.prod.js#L284

Looks like we are still compressing/outputting to ecma5. Let me get it into ecma 6. We should use es6 to be consistent with the rest of the output from vs code. 

I see now that tree shaking is only enabled for es6 or es2015. https://webpack.js.org/guides/tree-shaking/

packages/salesforcedx-webview-ui/package.json
vazexqi: I'm looking at https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/package.json and https://github.com/wmonk/create-react-app-typescript/blob/master/packages/react-scripts/package.json 

It seems that react, react-dom, and typescript are in the devDependencies. Let me move those back in there to follow that (not sure how it got moved into dependencies).

I will keep webpack and rest in dependencies for now since that seems to be how the template it.

vazexqi: More info at https://github.com/facebook/create-react-app/issues/1764

packages/salesforcedx-webview-ui/config/webpack.config.dev.js
vazexqi: I might be misunderstanding here but this should be OK because this is the _last_ branch of the _oneOf_ branch. That means we shouldn't need to add more tests. In fact, this test is the `not` of the other branches. 

So, we are already serving the svg, woff, eot, and ttf files. For jpeg, png, gif we actually convert embed them as data URLS to avoid any requests. See https://github.com/forcedotcom/salesforcedx-vscode/blob/e84a5178d55e911c2ba489598d514d88ad039132/packages/salesforcedx-webview-ui/config/webpack.config.dev.js#L121 and https://webpack.js.org/loaders/url-loader/

What I've done is also add woff|eot|ttf so that they get embedded.

I verified that we must be serving them properly because blueprint.js uses some custom font for the icons and we are seeing those icons in the manifest editor.

vazexqi:
@lcampos - The current version of create-react-app uses webpack 3.8.1 and not webpack 4 yet. We can decide to update this later if needed. I can't find an official upgrade doc like I could for webpack 2 -> 3 (https://webpack.js.org/migrate/3/)

lcampos:
@vazexqi there's not migration guide yet but there's some progress being made and tracked here (https://github.com/webpack/webpack.js.org/issues/1706). Right now, unless we see build/bundle perf issues I'd say we can wait to move to v4.

PR #452, opened 2018-06-01
Bump vscode dependency to 1.23

vazexqi:
### What does this PR do?

Webview Panels in VS Code are only available in version 1.23 and higher.
To support custom UI we need to be able to use webview panels.

I separated this from #451 to make it clearer than this is a version bump.

@W-5013822@


vazexqi:
When we do any manual testing we have been using 1.23.x so I'm less worried about any fallout from this. It does mean that unless users update to 1.23, they won't be able to get the new goodness -- which is fine since we need to do this soon anyway.

PR #454, opened 2018-06-01
Send InitializedEvent when LogContext is ready

DatGuyJonathan:
### What does this PR do?
With a Live Share session enabled, the ordering of requests/events doesn't appear to be consistent. Live Share team is still looking into it; their suggestion is to fix our `InitializedEvent`. The issue did expose a flaw in when the replay debugger sends an `InitializedEvent` to VS Code. It should do it after `LogContext` is initialized because it's actually ready for breakpoints and etc.

untitled

@W-5013801@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -191,12 +191,14 @@ export class ApexReplayDebug extends LoggingDebugSession {
   ): void {
     response.success = false;
     this.setupLogger(args);
-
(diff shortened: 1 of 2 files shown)


vazexqi:
This looks good to me. Correct me if I am wrong but from https://github.com/MicrosoftDocs/live-share/issues/480 it seems that they don't know why the ordering changes but there is at least a definite last-event that we can rely on and that is what we are using right?

So, even if the order changes a bit in the future, we should be resilient to it?

DatGuyJonathan:
Yep, that's my thinking as well. This change logically makes sense to me regardless of what Live Share changes in the future.

PR #457, opened 2018-06-04
Add an event listener for the creation of the sfdx-config file

allileong:
### What does this PR do?
This pr adds an event listener to detect the creation of the config file where the default scratch org is specified. That event is used to update the scratch org shortcut in the vscode status bar. 

W-4959520: The scratch org shortcut was not showing up in the vscode status bar in cases where the <workspace-dir>/.sfdx/sfdx-config.json was created after the creation of the scratch org. This is because we only had an event listener listening to updates made to the sfdx-config.json file, and no event listener detecting the creation of the file.


vazexqi:
Quick comment: @DatGuyJonathan could you walk @allileong on how to squash and merge so that the actual commit has @W-xxxxx@ in the commit message for out git2gus to work?

An example would be this commit message: https://github.com/forcedotcom/salesforcedx-vscode/commit/d07a21be2f35006418d0bd7675bc36317af1d5b9

PR #461, opened 2018-06-05
Add an extra quickPick for the api type (REST vs. Tooling API)

allileong:
### What does this PR do?
Adds an extra dropdown for the user to select whether they want to query the "REST API" or "Tooling API". 

W-4365988: Needs a way to be able to specify --usetoolingapi flag for SOQL query

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
@@ -17,59 +17,98 @@ import {
 import * as vscode from 'vscode';
 import { nls } from '../messages';
 import {
+  CompositeParametersGatherer,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
 class ForceDataSoqlQueryExecutor extends SfdxCommandletExecutor<{}> {
-  public build(data: QueryInput): Command {
-    return new SfdxCommandBuilder()
+  public build(data: QueryAndApi): Command {
+    let command = new SfdxCommandBuilder()
       .withDescription(nls.localize('force_data_soql_query_input_text'))
       .withArg('force:data:soql:query')
-      .withFlag('--query', `${data.input}`)
-      .build();
+      .withFlag('--query', `${data.query}`);
+    if (data.api === ApiType.Tooling) {
+      command = command.withArg('--usetoolingapi');
+    }
+    return command.build();
   }
 }
 
-export type QueryInput = {
-  input: string;
-};
-
-export class GetQueryInput implements ParametersGatherer<{ input: string }> {
+export class GetQueryInput implements ParametersGatherer<QueryInput> {
   public async gather(): Promise<
-    CancelResponse | ContinueResponse<{ input: string }>
+    CancelResponse | ContinueResponse<QueryInput>
   > {
     const editor = await vscode.window.activeTextEditor;
 
-    let input;
+    let query;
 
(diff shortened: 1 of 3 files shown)


allileong:
screen shot 2018-06-05 at 1 10 46 pm
Hi @vazexqi , this change adds a new dropdown selection step for the user to select the API that they want to query after they have input their query string. Based on this [documentation](https://trailhead.salesforce.com/en/modules/api_basics/units/api_basics_overview), I named the APIs "Rest API" and "Tooling API". Is this the best way to let the user specify --usetoolingapi? Should "Rest API" be renamed to something more descriptive?

vazexqi:
@allileong - Let's go with this way. The approach is sound.

The other alternative I was considering was to have two separate commands: 

1. Execute SOQL Query... 
1. Execute SOQL Query with Tooling API... 

but that seems like it's going to force too many commands. 

You might need to modify some of the system tests at https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/system-tests/scenarios/forceSoqlQueryUi.test.ts that rely on the sequence of inputs.

I will review again when you have added the tests to give the final code review.

packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
```diff
@@ -17,59 +17,98 @@ import {
...
+}
+
+export interface QueryInput {
+  query: string;
+}
+export interface ApiInput {
+  api: ApiType;
+}
+
+export class SelectApi implements ParametersGatherer<ApiInput> {
```
vazexqi: Actually, this new gatherer seems very specific for this use case -- so there is little reuse. Is it possible to just fold it into the previous one instead of composing them with the `CompositeParameterGatherer`?

allileong: @vazexqi No problem, I will make that change. For learning purposes, why is it better to fold it into the other gatherer?

vazexqi: _This is a minor issue hence I didn't even notice it on first review_

It's mostly whether there is reuse for a gatherer. It seem (and I could be wrong) that the gatherer to select Rest vs Tooling API doesn't have any reuse outside of this case. Do you know of any other command where we might want to use it again (maybe the data commands?). If there is, then let's leave it as-is.

Another reason why we might keep it as-is is for testing. If you find it easier to test it if you separate it, it's fine to keep it this way too.

allileong: Gotcha. The commands where I see the --usetoolingapi flag are the `sfdx force:data:record:*` commands, and the `sfdx force:schema:sobject:describe` command. Are there plans to implement those as vscode commands?

vazexqi: `force:data:record:*` would not likely be surfaced as a top-level command using this UX mechanism – there would be too much typing involved with the current UX and is too error-prone. If we surface this we would have a UI layer on top.

`force:schema:sobject:describe` would also likely not be surfaced at the top-level as a command. 

I'm following the philosophy of the Git integration in VS Code. Git has many CLI commands. But not all need to be surfaced. If the user needs those commands, they can still bounce to the CLI as needed.

allileong: @vazexqi sounds good! These changes are ready for another review

PR #467, opened 2018-06-07
Turn on Synk monitoring on each build for new vulnerabilities

vazexqi:
### What does this PR do?

Turn on Synk monitoring on each build for new vulnerabilities. This will run [`snyk monitor`](https://snyk.io/docs/using-snyk#monitor) and upload the results to https://snyk.io/org/vazexqi for now. I've already enquired about a global one for forcedotcom.

This will continuously update the security checks on each build (and, if no builds happen, it's set to check each day).

I've added @lcampos and @DatGuyJonathan as admins to https://snyk.io/org/vazexqi for now.

Trust.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "synk:test":
       "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
     "synk:monitor":
-      "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false",
+      "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false --org=salesforce",
     "start-webview": "npm run start --prefix=packages/salesforcedx-webview-ui",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,6 +43,10 @@
     "eslint-check":
       "eslint --print-config .eslintrc.json | eslint-config-prettier-check",
     "reformat": "node scripts/reformat-with-prettier.js",
+    "synk:test":
+      "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
```
vazexqi: The only reason that salesforcedx-vscode is ignored is because it's just an extension pack. Synk works by scanning the node_modules (so it needs to run after `npm install`).

vazexqi:
Hmm, the free https://snyk.io/plans is not that good for this method.

> How do you count tests?
We count a test whenever you make a change to your dependencies on your default branch, and depending on your settings, whenever you create a pull request which changes those dependencies. But for the Snyk CLI, we count each call to snyk test or snyk monitor as a test.

We cannot do it on each CI build for now. We could do it daily or weekly but we cannot do it on each build. I'll trim it down and investigate options.

vazexqi:
Just fount out about https://github.com/snyk/snyk/issues/67#issuecomment-295115124 - Will try.

PR #468, opened 2018-06-07
Use vscode-debugadapter-testsupport 

DatGuyJonathan:
### What does this PR do?
Test the replay debugger adapter with a real debug log. I'm basically goldfiling the stack and variables to make asserts easier (so I don't have to assert one variable at a time in the test code).

@W-4688243@


DatGuyJonathan:
@Sweetman , I'd like your feedback on this approach and test coverage for variables in general. The main goal of doing this now is so we can catch regressions that can come from the upcoming heapdump work.

Sweetman:
@DatGuyJonathan integration tests for static variables would be good to add as well. Static variables are shared between frames that are in the same class so it would be worthwhile to add a test checking that they're shared and that updates to one static variable will show up in both frames.

The best scenario for this test would probably be to have:
Class A
-static int myVar=10
-method1()
-method2()
ClassB
-static int myVar=5
-method1()

where A.method1 calls B.method1 which calls A.method2. This way we can check that static variables are only updated for their class. There should be a unit test for this but might be good to have an integration test as well where there are method entry frames into other classes.

DatGuyJonathan:
Thanks for the feedback @Sweetman . I've added a test for that scenario.

Sweetman:
LGTM

PR #469, opened 2018-06-07
Jsuplizio/checkpoints and locks

JimSuplizio:
### What does this PR do?
There are two changes here.
For W-5020642 - Added a 'wrap up' message to be displayed to the user if there's a failure or problem when calling SFDX: Update Checkpoints in Org. While the individual errors during the processing will pop up and be displayed in the console, this is just a message more or less saying something went wrong and you've got to make fixes and call the update command again.

For W-5049315 - This was a little tricky. These changes are guards against the user adding/deleting/changing things while the checkpoints are being updated in the org. There are two ways this can happen, the user executes SFDX: Toggle Checkpoint while the update command is running or the user clicks in the gutter and adds/removes/changes a checkpoint. In the first case a warning message is displayed to the user since, for that, we're the ones responsible for both the breakpoint and checkpoint. In the second case, where the user manipulates one from the gutter, that was a bit tricky because we're tapping into the onDidChangeBreakpoints which means VS Code is going to do whatever and then hit the callback. For that I'd had to introduce an AsyncLock that'll prevent modifications to the checkpoint list due to breakpoint changes while a checkpoint upload was in effect but would then process the change after the checkpoint upload was finished.

@W-5020642, W-5049315@


DatGuyJonathan:
For documentation purposes, `async-lock` isn't a new library to this repo. I introduced it to Apex Interactive Debugger a while ago, and the usages in Replay Debugger are to fix a similar problem.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -644,24 +653,26 @@ export async function processBreakpointChangedForCheckpoints(
...
+            bp.enabled,
+            uri,
+            filename,
+            checkpointOverlayAction
+          );
+        }
+      });
     } else {
       // The breakpoint is no longer a SourceBreakpoint or is no longer a checkpoint. Call to delete it if it exists
       checkpointService.deleteCheckpointNodeIfExists(breakpointId);
```
DatGuyJonathan: Could you explain why this doesn't need a lock?

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -879,6 +918,14 @@ export async function sfdxCreateCheckpoints() {
 //    and the logMessage. The logMessage is scrapped since this ends up being taken over by
 //    checkpoints for user input SOQL or Apex.
 export async function sfdxToggleCheckpoint() {
+  if (creatingCheckpoints) {
```
DatGuyJonathan: I thought toggle is similar to the user creating/removing a conditional breakpoint from the gutter. Here, you disallow toggle while there's an upload in progress. But if the user does it from the gutter, you queue the action using async-lock. Why are the behaviors different?

packages/salesforcedx-vscode-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -40,6 +40,10 @@ export const messages = {
     'Step 5 of 6: Uploading checkpoints',
   checkpoint_creation_status_processing_complete_success:
     'Step 6 of 6: Confirming successful checkpoint creation',
+  checkpoint_upload_in_progress:
+    'A checkpoint update is currently in progress for your org. Wait for the update to complete before you add or remove checkpoints.',
+  checkpoint_upload_error_wrap_up_message:
+    'Your checkpoints have errors. Fix the errors listed in the output, then run SFDX: Update Checkpoints in Org again.',
```
DatGuyJonathan: You can replace `SFDX: Update Checkpoints in Org` with `%s`. When you call `nls.localize` with this string, you can pass in an additional param that contains the localized string for that command.

DatGuyJonathan:
Can you look into the tests? In the apex interactive debugger, I was able to spy on async-lock's `acquire` method to verify it gets called with the correct lock key.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: It does. Fixed in next push.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
JimSuplizio: I actually fully described this in the description above.
"These changes are guards against the user adding/deleting/changing things while the checkpoints are being updated in the org. There are two ways this can happen, the user executes SFDX: Toggle Checkpoint while the update command is running or the user clicks in the gutter and adds/removes/changes a checkpoint. In the first case a warning message is displayed to the user since, for that, we're the ones responsible for both the breakpoint and checkpoint. In the second case, where the user manipulates one from the gutter, that was a bit tricky because we're tapping into the onDidChangeBreakpoints which means VS Code is going to do whatever and then hit the callback. For that I'd had to introduce an AsyncLock that'll prevent modifications to the checkpoint list due to breakpoint changes while a checkpoint upload was in effect but would then process the change after the checkpoint upload was finished."

packages/salesforcedx-vscode-apex-replay-debugger/src/messages/i18n.ts
JimSuplizio: It'll be done.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
DatGuyJonathan: It seems like everything in `processBreakpointChangedForCheckpoints` needs a lock. Can the whole method be locked instead of having multiple `lock.acquire`? Similar to what you did in `sfdxCreateCheckpoints`.

DatGuyJonathan: Thoughts on ^^^?

JimSuplizio: Not everything does. Remember that even non-checkpoint breakpoint changes will still go through processBreakpointChangedForCheckpoints. The guards here are if and only if it will require checkpoint changes. Breakpoints can still be added/removed and if they're not checkpoints then that won't stop in here.

PR #474, opened 2018-06-11
Update to apex-tmLanguage v1.0.1

lcampos:
### What does this PR do?
Adds tokenization support for usage of apex support classes, methods, etc in the following scenarios:
- Casting
- Return type
- Parameters
- Method declarations
- Apex Triggers
- Field declarations
- Property declarations
- Catch clauses
- Local variable declarations
- Object creation
- Array creation
- As type arguments

support-class

support-function

@W-4998669@


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -75,11 +75,13 @@
       {
         "language": "apex",
         "scopeName": "source.apex",
-        "path": "./syntaxes/apex.tmLanguage"
+        "path":
+          "./node_modules/@salesforce/apex-tmlanguage/grammars/apex.tmLanguage"
```
vazexqi: @lcampos - Could you do sanity check on the salesforcedx-vscode-core .vsix from CircleCI once they are ready? I just want to make sure that when we run `vscode:package` that the node_modules for this is packaged properly and not pruned away.

lcampos: Everything looks good (verified/tested with the circleci artifact), I'll merge this now.

PR #485, opened 2018-06-13
Address more precise types from @types/node

vazexqi:
### What does this PR do?

Address the type changes due to https://github.com/DefinitelyTyped/DefinitelyTyped/commit/09ff4cfc0a0677e498f537640dfefa7fb40fbba1#diff-8811b50169fd98160ad78b073b4d0163

@W-5086993@

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -98,7 +98,7 @@ function loadMessageBundle(config?: Config): Message {
 }
 
 export const nls = new Localization(
-  loadMessageBundle(process.env.VSCODE_NLS_CONGIF ? JSON.parse(process.env.VSCODE_NLS_CONFIG!) : undefined)
+  loadMessageBundle(process.env.VSCODE_NLS_CONFIG ? JSON.parse(process.env.VSCODE_NLS_CONFIG!) : undefined)
 );
 
 ```
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -98,7 +98,7 @@ function loadMessageBundle(config?: Config): Message {
 }
 
 export const nls = new Localization(
-  loadMessageBundle(JSON.parse(process.env.VSCODE_NLS_CONFIG))
+  loadMessageBundle(process.env.VSCODE_NLS_CONGIF ? JSON.parse(process.env.VSCODE_NLS_CONFIG!) : undefined)
 );
 
 ```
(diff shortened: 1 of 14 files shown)


vazexqi:
@lcampos @DatGuyJonathan - Could you review this since it's a blocker for tomorrow's release? We should also look to see if we need to forward port this to develop first.

docs/localization.md
```diff
@@ -98,7 +98,7 @@ function loadMessageBundle(config?: Config): Message {
 }
 
 export const nls = new Localization(
-  loadMessageBundle(JSON.parse(process.env.VSCODE_NLS_CONFIG))
+  loadMessageBundle(process.env.VSCODE_NLS_CONGIF ? JSON.parse(process.env.VSCODE_NLS_CONFIG!) : undefined)
```
lcampos: @vazexqi think that ! at the end of JSON.parse is an issue.

vazexqi:
sketch

Verified that even if you switch locales, nothing malfunctions. We don't have language files for any other language so the default behavior is to fall back on en.

docs/localization.md
vazexqi: @lcampos - Sorry, why would that be an issue? The bang is the non-null assertion operator. See https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator

There is an issue with `process.env.VSCODE_NLS_CONGIF` though, which I will fix.

packages/salesforcedx-apex-debugger/src/messages/index.ts
```diff
@@ -42,5 +42,5 @@ function loadMessageBundle(config?: Config): Message {
 }
 
 export const nls = new Localization(
-  loadMessageBundle(JSON.parse(process.env.VSCODE_NLS_CONFIG))
+  loadMessageBundle(process.env.VSCODE_NLS_CONGIF ? JSON.parse(process.env.VSCODE_NLS_CONFIG!) : undefined)
```
DatGuyJonathan: I think `CONGIF` needs to be `CONFIG`, here and other places as well

DatGuyJonathan: Oops, yea, you caught it faster!

DatGuyJonathan:
Verified that I can clean, install, compile, and package this branch, so I think I should be okay for the release.

vazexqi:
The build is failing because of the SLDS issue that was fixed in #479 but not backported to this branch. I'm going to close this and also cherry-pick this into develop.

vazexqi:
- [ ] Initial code review done 
- [ ] Changes based on code review
- [ ] Verify if this preference will also be picked up by the Apex Language Server for the code lens

PR #495, opened 2018-06-18
W-5067151: Update request-light and eslint dependencies for security patches

allileong:
### What does this PR do?
This pr updates the version numbers for our request-light and eslint dependencies. 

The updates include fixes for security vulnerabilities found by Snyk.

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -13,7 +13,7 @@
     "@salesforce/salesforcedx-utils-vscode": "43.3.0",
     "async-lock": "1.0.0",
     "faye": "1.1.2",
-    "request-light": "0.2.1",
+    "request-light": "0.2.3",
     "vscode-debugadapter": "1.28.0",
     "vscode-debugprotocol": "1.28.0"
   },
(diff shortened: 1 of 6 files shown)


DatGuyJonathan:
Snyk doesn't run on top level `node_modules`, but I think the `eslint` version should be updated as well for consistency: https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/package.json#L5

vazexqi:
"Snyk doesn't run on top level node_modules" <-- We can make it run on the top-level ones as well. Perhaps we should do that?

vazexqi:
```
lerna ERR! npm WARN ajv-keywords@2.1.1 requires a peer of ajv@^5.0.0 but none is installed. You must install peer dependencies yourself.
lerna ERR! 
lerna ERR! Error: Command failed: npm list --production --parseable --depth=99999
lerna ERR! npm ERR! peer dep missing: ajv@^5.0.0, required by ajv-keywords@2.1.1
```

This one looks serious. Might be related to #322 where we also hit an issue with ajy. If CircleCI doesn't pass, that means that we won't be able to package and publish.

vazexqi:
By the way, @allileong (and anyone else contributing), we have a preference on how we structure the git commit messages. See https://chris.beams.io/posts/git-commit/ I realized I never made that explicit but it's good to write it down. I will make a PR to the docs folder to include that.

So usually we don't include the bug number in the subject/title so that it's easier to make sense of what a PR is without having to think about bug numbers.

We can discuss more when I send the PR on the preference of git messages (think of it as a request-for-comments).

PR #497, opened 2018-06-19
Wrap environment setup for isv debugger in a try-catch

DatGuyJonathan:
### What does this PR do?
salesforcedx-vscode-core does some environment setup for ISV Debugger. The setup allows SFDX commands to work in an ISV Debugger project. If there's an exception during the setup, it prevents salesforcedx-vscode-core extension to activate, therefore other extensions that depend on it also fail to activate.

So, this just catches the error, outputs it to a console in VS Code Developer Tools, and tells the user there was a problem with the setup. If there's an error, the user can still start and use the isv debugger, but SFDX commands won't work.

@W-5093261@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -152,7 +152,7 @@ export const messages = {
   isv_debug_bootstrap_open_project:
     'Opening project in new Visual Studio Code window',
   isv_debug_config_environment_error:
-    'There was a problem configuring the environment for ISV Customer Debugger. Some features may not work. See VS Code Developer Tools and Salesforce CLI logs for details.',
+    'Salesforce Extensions for VS Code encountered a problem while configuring your environment. Some features might not work. For details, click Help > Toggle Developer Tools or check the Salesforce CLI logs in ~/.sfdx/sfdx.log.',
 
   force_apex_log_get_text: 'SFDX: Get Apex Debug Logs...',
   force_apex_log_get_no_logs_text: 'No Apex debug logs were found',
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -151,6 +151,8 @@ export const messages = {
   isv_debug_bootstrap_generate_launchjson: 'Creating launch configuration',
   isv_debug_bootstrap_open_project:
     'Opening project in new Visual Studio Code window',
+  isv_debug_config_environment_error:
+    'There was a problem configuring the environment for ISV Customer Debugger. Some features may not work. See VS Code Developer Tools and Salesforce CLI logs for details.',
```
DatGuyJonathan: image

vazexqi: It seems that the check that we do is only `if (sfdxApexDebuggerExtension && sfdxApexDebuggerExtension.id)`. So the user might **not** even be using the ISV Debugger and this message could happen.
Is there a way we can re-word this differently?

ruthemmanuelle: Maybe something like this?
`'Salesforce Extensions for VS Code encountered a problem while configuring your environment. Some features might not work. For details, select Salesforce CLI in the Output panel or click Help > Toggle Developer Tools.'`

vazexqi: Yes, that looks good. @DatGuyJonathan could you ensure that the logging has enough details to identify this at the ISV Debugger culprit code path.

A tech debt that we need to file is to provide a better way to dump logs. Just like what VS LIve has with the Dump Logs command.

ruthemmanuelle: I guess so. We try to avoid mentioning the hidden .sfdx folder, since it's not really designed to be messed with, but it sounds like the best solution here. How about this?

`'Salesforce Extensions for VS Code encountered a problem while configuring your environment. Some features might not work. For details, click Help > Toggle Developer Tools or check the Salesforce CLI logs in ~/.sfdx/sfdx.log.'`

vazexqi: Looks good. Thanks @ruthemmanuelle

PR #389, opened 2018-04-19
Refactor some common code out of salesforcedx-apex-debugger

JimSuplizio:
### What does this PR do?
Refactor some common code out of salesforcedx-apex-debugger into salesforcedx-utils-vscode.

Refactor the BaseCommand, RequestService, ForceConfigGet and ForceOrgDisplay 

The RequestSerivce is basically a REST command executor and is used by the Apex debugger and the Apex Replay Debugger. It was odd for the replay debugger to have a dependency on the Apex debugger package. I moved the RequestService and BaseCommands into the utils-vscode extension. Further, the ForceConfigGet and ForceOrgDisplay being CLI commands were also moved. Both were moved but only the ForceOrgDisplay is currently being used by multiple extensions.

The vast majority of files in this PR are changes to imports. 

The project.json file for salesforcedx-utils-vscode did require dependency updates for sinon, mock-spawn and request-light. Because of these changes the salesforcedx-vscode-replay-debugger/tsconfig.json needed to have "dom" added to it's list of libs.

### What issues does this PR fix or reference?
@W-4897179@

PR #500, opened 2018-06-21
Rest call to get heapdumps, initial

JimSuplizio:
### What does this PR do?
This PR contains the necessary changes to scan the log for heap_dump lines and make the necessary calls, on initialize request, to pull the heap dumps to the client.

@W-5068047@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/package.json
+++ b/packages/salesforcedx-apex-replay-debugger/package.json
@@ -13,6 +13,7 @@
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "43.4.0",
+    "request-light": "0.2.1",
     "vscode-debugadapter": "1.28.0",
     "vscode-debugprotocol": "1.28.0",
     "vscode-uri": "1.0.1"
--- a/packages/salesforcedx-apex-replay-debugger/package.json
+++ b/packages/salesforcedx-apex-replay-debugger/package.json
@@ -13,7 +13,6 @@
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "43.4.0",
-    "request-light": "0.2.1",
     "vscode-debugadapter": "1.28.0",
     "vscode-debugprotocol": "1.28.0",
     "vscode-uri": "1.0.1"
@@ -29,6 +28,7 @@
     "mocha-junit-reporter": "^1.13.0",
     "mocha-multi-reporters": "^1.1.4",
     "nyc": "^11.0.2",
+    "request-light": "0.2.1",
     "sinon": "^2.3.6",
     "typescript": "2.6.2",
     "vscode": "1.1.17",
--- a/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -22,7 +22,8 @@ import {
   Variable
 } from 'vscode-debugadapter';
 import { DebugProtocol } from 'vscode-debugprotocol';
-import { breakpointUtil } from '../breakpoints';
+import { breakpointUtil, LineBreakpointEventArgs } from '../breakpoints';
+import { ApexExecutionOverlayResultCommandSuccess } from '../commands';
 import {
   GET_LINE_BREAKPOINT_INFO_EVENT,
   LINE_BREAKPOINT_INFO_REQUEST
@@ -163,6 +164,55 @@ export class ScopeContainer extends VariableContainer {
   }
 }
 
+export class ApexHeapDump {
+  private readonly heapDumpId: string;
+  private readonly className: string;
+  private readonly namespace: string;
+  private readonly line: number;
+  private overlaySuccessResut:
+    | ApexExecutionOverlayResultCommandSuccess
+    | undefined;
+  public constructor(
+    heapDumpId: string,
+    className: string,
+    namespace: string,
+    line: number
+  ) {
+    this.heapDumpId = heapDumpId;
+    this.className = className;
+    this.namespace = namespace;
+    this.line = line;
+  }
+  public getHeapDumpId(): string {
+    return this.heapDumpId;
+  }
+  public getClassName(): string {
+    return this.className;
+  }
+  public getNamespace(): string {
+    return this.namespace;
+  }
+  public getLine(): number {
+    return this.line;
+  }
+  public getOverlaySuccessResult():
+    | ApexExecutionOverlayResultCommandSuccess
+    | undefined {
+    return this.overlaySuccessResut;
+  }
+  public setOverlaySuccessResult(
+    overlaySuccessResult: ApexExecutionOverlayResultCommandSuccess
+  ): void {
+    this.overlaySuccessResut = overlaySuccessResult;
+  }
+  /* tslint:disable */
+  public toString = (): string => {
+    return `HeapDumpId: ${this.heapDumpId}, ClassName: ${this
+      .className}, Namespace: ${this.namespace}, Line: ${this.line}`;
+  }; /* This semi-colon is the reason for the tslint, formatting keeps adding it */
+  /* tslint:enable */
+}
+
 export class ApexReplayDebug extends LoggingDebugSession {
   public static THREAD_ID = 1;
   protected logContext: LogContext;
@@ -170,6 +220,7 @@ export class ApexReplayDebug extends LoggingDebugSession {
   protected traceAll = false;
   private initializedResponse: DebugProtocol.InitializeResponse;
   protected breakpoints: Map<string, number[]> = new Map();
+  protected projectPath: string | undefined;
 
   constructor() {
     super('apex-replay-debug-adapter.log');
@@ -185,10 +236,10 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.sendEvent(new Event(GET_LINE_BREAKPOINT_INFO_EVENT));
   }
 
-  public launchRequest(
+  public async launchRequest(
     response: DebugProtocol.LaunchResponse,
     args: LaunchRequestArguments
-  ): void {
+  ): Promise<void> {
     response.success = false;
     this.setupLogger(args);
     this.log(
@@ -211,6 +262,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.printToDebugConsole(
       nls.localize('session_started_text', this.logContext.getLogFileName())
     );
+    // If the projectPath isn't set then don't bother with heap dump processing
+    if (this.projectPath) {
+      if (this.logContext.scanLogForHeapDumpLines()) {
+        if (
+          !await this.logContext.fetchOverlayResultsForApexHeapDumps(
+            this.projectPath
+          )
+        ) {
+          this.errorToDebugConsole(
+            nls.localize('heap_dump_error_wrap_up_text')
+          );
+        }
+      }
+    }
     response.success = true;
     this.sendResponse(response);
   }
@@ -465,7 +530,11 @@ export class ApexReplayDebug extends LoggingDebugSession {
     switch (command) {
       case LINE_BREAKPOINT_INFO_REQUEST:
         if (args) {
-          breakpointUtil.createMappingsFromLineBreakpointInfo(args);
+          const lineBreakpointEventArgs = args as LineBreakpointEventArgs;
+          breakpointUtil.createMappingsFromLineBreakpointInfo(
+            lineBreakpointEventArgs.lineBreakpointInfo
+          );
+          this.projectPath = lineBreakpointEventArgs.projectPath;
         } else {
           this.initializedResponse.success = false;
           this.initializedResponse.message = nls.localize(
--- a/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -23,7 +23,7 @@ import {
 } from 'vscode-debugadapter';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { breakpointUtil, LineBreakpointEventArgs } from '../breakpoints';
-import { ApexExecutionOverlayResultCommandSuccess } from '../commands';
+
 import {
   GET_LINE_BREAKPOINT_INFO_EVENT,
   LINE_BREAKPOINT_INFO_REQUEST
@@ -164,55 +164,6 @@ export class ScopeContainer extends VariableContainer {
   }
 }
 
-export class ApexHeapDump {
-  private readonly heapDumpId: string;
-  private readonly className: string;
-  private readonly namespace: string;
-  private readonly line: number;
-  private overlaySuccessResut:
-    | ApexExecutionOverlayResultCommandSuccess
-    | undefined;
-  public constructor(
-    heapDumpId: string,
-    className: string,
-    namespace: string,
-    line: number
-  ) {
-    this.heapDumpId = heapDumpId;
-    this.className = className;
-    this.namespace = namespace;
-    this.line = line;
-  }
-  public getHeapDumpId(): string {
-    return this.heapDumpId;
-  }
-  public getClassName(): string {
-    return this.className;
-  }
-  public getNamespace(): string {
-    return this.namespace;
-  }
-  public getLine(): number {
-    return this.line;
-  }
-  public getOverlaySuccessResult():
-    | ApexExecutionOverlayResultCommandSuccess
-    | undefined {
-    return this.overlaySuccessResut;
-  }
-  public setOverlaySuccessResult(
-    overlaySuccessResult: ApexExecutionOverlayResultCommandSuccess
-  ): void {
-    this.overlaySuccessResut = overlaySuccessResult;
-  }
-  /* tslint:disable */
-  public toString = (): string => {
-    return `HeapDumpId: ${this.heapDumpId}, ClassName: ${this
-      .className}, Namespace: ${this.namespace}, Line: ${this.line}`;
-  }; /* This semi-colon is the reason for the tslint, formatting keeps adding it */
-  /* tslint:enable */
-}
-
 export class ApexReplayDebug extends LoggingDebugSession {
   public static THREAD_ID = 1;
   protected logContext: LogContext;
@@ -263,17 +214,13 @@ export class ApexReplayDebug extends LoggingDebugSession {
       nls.localize('session_started_text', this.logContext.getLogFileName())
     );
     // If the projectPath isn't set then don't bother with heap dump processing
-    if (this.projectPath) {
-      if (this.logContext.scanLogForHeapDumpLines()) {
-        if (
-          !await this.logContext.fetchOverlayResultsForApexHeapDumps(
-            this.projectPath
-          )
-        ) {
-          this.errorToDebugConsole(
-            nls.localize('heap_dump_error_wrap_up_text')
-          );
-        }
+    if (this.projectPath && this.logContext.scanLogForHeapDumpLines()) {
+      if (
+        !await this.logContext.fetchOverlayResultsForApexHeapDumps(
+          this.projectPath
+        )
+      ) {
+        this.errorToDebugConsole(nls.localize('heap_dump_error_wrap_up_text'));
       }
     }
     response.success = true;
--- a/packages/salesforcedx-apex-replay-debugger/src/breakpoints/index.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/breakpoints/index.ts
@@ -13,3 +13,8 @@ export interface LineBreakpointInfo {
 export { BreakpointUtil } from './breakpointUtil';
 import { BreakpointUtil } from './breakpointUtil';
 export const breakpointUtil = BreakpointUtil.getInstance();
+
+export interface LineBreakpointEventArgs {
+  lineBreakpointInfo: LineBreakpointInfo[];
+  projectPath: string | undefined;
+}
--- a/packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { BaseCommand } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
+import { SOBJECTS_URL } from '../constants';
+
+export interface ApexExecutionOverlayResultCommandFailure {
+  message: string;
+  errorCode: string;
+}
+
+export interface ApexExecutionOverlayResultCommandSuccess {
+  attributes: Attributes;
+  Id: string;
+  IsDeleted: boolean;
+  CreatedDate: Date;
+  CreatedById: string;
+  LastModifiedDate: Date;
+  LastModifiedById: string;
+  SystemModstamp: Date;
+  UserId: string;
+  RequestedById: string;
+  OverlayResultLength: number;
+  HeapDump: HeapDump;
+  ApexResult: HeapDumpApexResult | null;
+  SOQLResult: HeapDumpSOQLResult | null;
+  Line: number;
+  Iteration: number;
+  ExpirationDate: Date;
+  IsDumpingHeap: boolean;
+  ActionScript: string | null;
+  ActionScriptType: string;
+  ClassName: string;
+  Namespace: string;
+}
+
+export interface HeapDump {
+  className: string;
+  extents: HeadpDumpExtents[];
+  heapDumpDate: Date;
+  namespace: string;
+}
+
+export interface HeadpDumpExtents {
+  collectionType: string | null;
+  count: number;
+  definition: HeadpDumpCollectionTypeDefinition[];
+  extent: HeadpDumpExtent[];
+  totalSize: number;
+  typeName: string;
+}
+
+export interface HeadpDumpCollectionTypeDefinition {
+  name: string;
+  type: string;
+}
+
+export interface HeadpDumpExtent {
+  address: string;
+  size: number;
+  symbols: string[] | null;
+  value: HeadpDumpExtentValue;
+}
+
+// The Extent value has to be any. The value
+export interface HeadpDumpExtentValue {
+  value?: any;
+  entry?: HeadpDumpExtentValueEntry[];
+}
+
+export interface HeadpDumpExtentValueEntry {
+  keyDisplayValue: string;
+  value: HeadpDumpExtentValue;
+}
+
+export interface HeapDumpApexResult {
+  apexError: string | null;
+  apexExecutionResult: HeapDumpApexExecutionResult | null;
+}
+
+export interface HeapDumpApexExecutionResult {
+  column: number;
+  compileProblem: string | null;
+  compiled: boolean;
+  exceptionMessage: string | null;
+  exceptionStackTrace: string | null;
+  line: number;
+  success: boolean;
+}
+
+// If the queryError is returned then the queryMetadata and queryResult will both be null.
+export interface HeapDumpSOQLResult {
+  queryError: string | null;
+  queryMetadata: HeapDumpSOQLResultQueryMetadata | null;
+  queryResult: HeapDumpSOQLResultQueryResult[] | null;
+}
+
+export interface HeapDumpSOQLResultQueryMetadata {
+  columnMetadata: HeapDumpSOQLResultColumnMetadata[];
+  entityName: string;
+  groupBy: boolean;
+  idSelected: boolean;
+  keyPrefix: string;
+}
+
+export interface HeapDumpSOQLResultColumnMetadata {
+  aggregate: boolean;
+  apexType: string;
+  booleanType: boolean;
+  columnName: string;
+  custom: boolean;
+  displayName: string;
+  foreignKeyName: string | null;
+  insertable: boolean;
+  joinColumns: string[];
+  numberType: boolean;
+  textType: boolean;
+  updatable: boolean;
+}
+
+// A note about the HeapDumpSOQLResultQueryResult. The fields from a SOQL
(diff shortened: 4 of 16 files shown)


packages/salesforcedx-apex-replay-debugger/package.json
```diff
@@ -13,6 +13,7 @@
   "categories": ["Debuggers"],
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "43.4.0",
+    "request-light": "0.2.1",
```
JimSuplizio: Added for XHROptions and XHRResponse required for testing.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -211,6 +262,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.printToDebugConsole(
       nls.localize('session_started_text', this.logContext.getLogFileName())
     );
+    // If the projectPath isn't set then don't bother with heap dump processing
+    if (this.projectPath) {
```
JimSuplizio: I was torn between sending the boolean of whether or not checkpoints were enabled versus using an undefined projectPath. The fact is, if the path is somehow undefined then we're not going to be able to get the heap dumps anyways. Rather than passing around the enabled status, on the extension side if checkpoints aren't enabled then it passes over an undefined project path.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -211,6 +262,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+      if (this.logContext.scanLogForHeapDumpLines()) {
+        if (
+          !await this.logContext.fetchOverlayResultsForApexHeapDumps(
```
JimSuplizio: This is the reason why launchRequest had to be made async. The REST calls are all async and we want to wait on them.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -211,6 +262,20 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+            this.projectPath
+          )
+        ) {
+          this.errorToDebugConsole(
+            nls.localize('heap_dump_error_wrap_up_text')
```
JimSuplizio: If there's an error fetching heap dumps that's reported at the time it happens, this is a rollup error.

packages/salesforcedx-apex-replay-debugger/src/breakpoints/index.ts
```diff
@@ -13,3 +13,8 @@ export interface LineBreakpointInfo {
 export { BreakpointUtil } from './breakpointUtil';
 import { BreakpointUtil } from './breakpointUtil';
 export const breakpointUtil = BreakpointUtil.getInstance();
+
+export interface LineBreakpointEventArgs {
```
JimSuplizio: The line breakpoint event response now contains the line breakpoint information and the project path. This is the interface.

packages/salesforcedx-apex-replay-debugger/src/commands/index.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export enum ActionScriptEnum {
```
JimSuplizio: Both the ActionScriptEnum and OrgInfoError had to be moved from the debugger extension (salesforcedx-vscode-apex-replay-debugger) to the adapter (salesforcedx-apex-replay-debugger) to prevent a circular reference. The extension can pull things from the adapter but not the reverse.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
     );
   }
 
+  public getHeapDumps(): ApexHeapDump[] {
+    return this.apexHeapDumps;
+  }
+
+  public scanLogForHeapDumpLines(): boolean {
```
JimSuplizio: This matches the "|HEAP_DUMP|" log tags. If it finds one it splits the line and sets all of the information. Since logging changes more or less have to be additive the ordering here isn't going to change.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
...
+          splitLine[5] /* namespace */,
+          Number(splitLine[6]) /* line */
+        );
+        this.apexHeapDumps.push(heapDump);
+      }
+    }
+    return this.apexHeapDumps.length > 0;
+  }
+
+  public async fetchOverlayResultsForApexHeapDumps(
```
JimSuplizio: The actual method to fetch the heap dumps. I'd considered a bulk call however that would be a potentially hideous amount of combined information. Developer Tools, inspecting the responses from single heap dump calls already has issues displaying larger heap dumps. With 5 large sized heap dumps it takes about 1.5-2.2 seconds (total) on startup to pull them all to the client.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebug.test.ts
```diff
@@ -62,6 +63,15 @@ export class MockApexReplayDebug extends ApexReplayDebug {
   public shouldStopForBreakpoint(): boolean {
     return super.shouldStopForBreakpoint();
   }
+
+  public setProjectPath(projectPath: string | undefined): void {
+    this.projectPath = projectPath;
+  }
+
+  public getProjectPath(): string | undefined {
+    const foo = this.projectPath;
```
JimSuplizio: This has already been changed locally to just return this.projectPath and will be updated with the next push that'll be based on feedback. This was a leftover tweak from when I'd had to debug something.

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -142,9 +145,26 @@ function registerDebugHandlers(): vscode.Disposable {
...
+            }
+            const config = vscode.workspace.getConfiguration();
+            const checkpointsEnabled = config.get(
+              'salesforcedx-vscode-apex-replay-debugger-checkpoints.enabled',
+              false
+            );
+            const returnArgs: LineBreakpointEventArgs = {
+              lineBreakpointInfo: lineBpInfo,
+              // for the moment always send undefined if checkpoints aren't enabled.
+              projectPath: checkpointsEnabled ? fsPath : undefined
```
JimSuplizio: This is what I'd mentioned above. When checkpoints are a real boy and no longer need the check this will just return the project path, if there is one. On the adapter side nothing will need to change.

packages/salesforcedx-apex-replay-debugger/test/unit/commands/apexExecutionOverlayResultCommand.test.ts
```diff
@@ -0,0 +1,786 @@
+/*
```
JimSuplizio: A general note about things in here. There are certain things that are hard coded rather than create a hojillion constants. The interface for the heap dumps is far from trivial and scenarios had to be crafted to ensure things like the SOQL query results and Apex results wouldn't clash (Apex action script tries to commit, this doesn't play well with pending rollbacks from inserts done as part of tests). The code used to generate these and checkpoint locations is at the bottom of the file in comments.

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -23,5 +23,10 @@ export const messages = {
   no_log_file_text:
     'The log file either is missing or does not have any log lines in it.',
   incorrect_log_levels_text:
-    'The log must be generated with log categories Apex code at the FINEST level and Visualforce at the FINER level.'
+    'The log must be generated with log categories Apex code at the FINEST level and Visualforce at the FINER level.',
+  unable_to_retrieve_org_info: 'Unable to retrieve OrgInfo',
+  heap_dump_error:
+    'Problems were encountered while retrieving heap dump information. Message=%s, ErrorCode=%s. Heap Dump Information=%s',
+  heap_dump_error_wrap_up_text:
```
JimSuplizio: It's worth nothing that I'd worked with Ruth on this error message. She's OOO which is why I didn't add her to the review.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -163,13 +164,63 @@ export class ScopeContainer extends VariableContainer {
   }
 }
 
+export class ApexHeapDump {
```
DatGuyJonathan: Could you extract this out to its own file? I don't want `apexReplayDebug.ts` to get bloated.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
DatGuyJonathan: Could you collapse these all into one if statement?

packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+export interface HeadpDumpExtent {
+  address: string;
+  size: number;
+  symbols: string[] | null;
+  value: HeadpDumpExtentValue;
+}
+
+// The Extent value has to be any. The value
```
DatGuyJonathan: I think you're missing some stuff in this comment.

packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
```diff
@@ -0,0 +1,162 @@
...
+// There's nothing remarkable in the request body or the query string. This
+// particular command is simply the URL with a query string that is
+// _<CurrentTimeInMilliseconds>
+export class ApexExecutionOverlayResultCommand extends BaseCommand {
+  private readonly commandName = 'ApexExecutionOverlayResult';
+  private readonly heapdumpKey: string;
+
+  public constructor(heapdumpKey: string) {
+    // The query string is '_<milliseconds>'
+    super('_=' + Date.now());
```
DatGuyJonathan: Is the timestamp necessary? How does it affect the response?

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
...
 
+  public getHeapDumps(): ApexHeapDump[] {
+    return this.apexHeapDumps;
+  }
+
+  public scanLogForHeapDumpLines(): boolean {
+    const heapDumpRegex = RegExp(/\|HEAP_DUMP\|/);
+    for (const line of this.logLines) {
+      if (heapDumpRegex.test(line)) {
+        const splitLine = line.split('|');
```
DatGuyJonathan: Could you edit the regex so it makes sure you have all these indexes before accessing them?

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
...
+  ): Promise<boolean> {
+    let success = true;
+    try {
+      const orgInfo = await new ForceOrgDisplay().getOrgInfo(projectPath);
+      const requestService = new RequestService();
+      requestService.instanceUrl = orgInfo.instanceUrl;
+      requestService.accessToken = orgInfo.accessToken;
+
+      for (const heapDump of this.apexHeapDumps) {
+        this.session.printToDebugConsole(`fetching: ${heapDump.toString()}`);
```
DatGuyJonathan: Could you externalize the strings?

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
...
+            this.session.errorToDebugConsole(errorMessage);
+          } catch (error) {
+            const errorMessage = `${error}. ${errorString}. ${heapDump.toString()}`;
+            this.session.errorToDebugConsole(errorMessage);
+          }
+        }
+      }
+    } catch (error) {
+      success = false;
+      const result = JSON.parse(error) as OrgInfoError;
```
DatGuyJonathan: Can this be any other error besides from `ForceOrgDisplay`? Even if you don't have a repro, I think it's best to be conservative here because this is called from `launchRequest`

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
DatGuyJonathan: Please make such conversations public next time for traceability. It's perfectly fine to have your draft of the text in the PR and iterate with @ruthemmanuelle on it.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebug.test.ts
```diff
@@ -208,6 +232,101 @@ describe('Replay debugger adapter - unit', () => {
...
+      expect(hasLogLinesStub.calledOnce).to.be.true;
+      expect(meetsLogLevelRequirementsStub.calledOnce).to.be.true;
+      expect(scanLogForHeapDumpLinesStub.calledOnce).to.be.true;
+      expect(fetchOverlayResultsForApexHeapDumpsStub.calledOnce).to.be.true;
+      expect(errorToDebugConsoleStub.calledOnce).to.be.true;
+      const errorMessage = errorToDebugConsoleStub.getCall(0).args[0];
+      expect(errorMessage).to.equal(
+        nls.localize('heap_dump_error_wrap_up_text')
+      );
+      errorToDebugConsoleStub.restore();
```
DatGuyJonathan: Do you need the restore here? You already restore in the `afterEach`.

packages/salesforcedx-apex-replay-debugger/test/unit/commands/apexExecutionOverlayResultCommand.test.ts
```diff
@@ -0,0 +1,786 @@
...
+      expect(extent[i].value.entry![0].keyDisplayValue).to.equal('Name');
+      expect(
+        extent[i].value.entry![0].value.value!.startsWith('okToDelete')
+      ).to.equal(true);
+
+      // Verity the AccountNumber keyDisplayValue and the value is xxx
+      expect(extent[i].value.entry![1].keyDisplayValue).to.equal(
+        'AccountNumber'
+      );
+      // JRS expect(extent[i].value.entry![1].value.value).to.equal('xxx');
```
DatGuyJonathan: Should this be removed?

packages/salesforcedx-apex-replay-debugger/test/unit/core/logContext.test.ts
```diff
@@ -183,6 +183,29 @@ describe('LogContext', () => {
...
   });
 
+  it('Should detect and parse HEAP_DUMP log entries', () => {
+    readLogFileStub.restore();
+    readLogFileStub = sinon
+      .stub(LogContextUtil.prototype, 'readLogFile')
+      .returns([
+        '43.0 APEX_CODE,FINEST;...;VISUALFORCE,FINER;..',
+        '<TimeInfo>|HEAP_DUMP|[11]|<HeapDumpId1>|<ClassName1>|<Namespace1>|11',
+        '<TimeInfo>|HEAP_DUMP|[22]|<HeapDumpId2>|<ClassName2>|<Namespace2>|22'
```
DatGuyJonathan: You were looking into if these two line numbers can ever be different. Any findings from that?

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: @DatGuyJonathan I'm open to suggestions here. What would you like to see?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
JimSuplizio: @DatGuyJonathan we did it quickly over slack late Wednesday afternoon because I noticed her OOO message otherwise we'd have done it over email, as per usual.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: I'm leaving the regex as is. I can check the split number to ensure the line has the minimum number of fields.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: Sure. I'll create a heapDump in core since I'm pretty sure I'm going to end up needing a util file.

packages/salesforcedx-apex-replay-debugger/test/unit/core/logContext.test.ts
JimSuplizio: Long story short, I don't see a way where the two line numbers can be different that we'd actually take the heap_dump. The second one is from the checkpoint, itself, and that's the one I feel is safest to use because we know, very specifically, where it comes from.

packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
JimSuplizio: Updated.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: I collapsed the first two into a single, the last one it still it's own

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: Done.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
DatGuyJonathan: Interesting.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
DatGuyJonathan: I have a thought, but I was hoping you'd mull over it and propose something. Did you get a chance to do that?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -23,5 +23,14 @@ export const messages = {
...
+    'Problems were encountered while retrieving heap dump information. See the messages above this one for details. Detailed heap dump variable information is unavailable for the specified line or lines.',
+  fetching_heap_dump:
+    'Retrieving heap dump from server. Heap dump information: %s',
```
DatGuyJonathan: This and the one below are new messages in your latest commit. Do you need Ruth to review?

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,88 @@ export class LogContext {
...
+  public scanLogForHeapDumpLines(): boolean {
+    const heapDumpRegex = RegExp(/\|HEAP_DUMP\|/);
+    for (const line of this.logLines) {
+      if (heapDumpRegex.test(line)) {
+        const splitLine = line.split('|');
+        const heapDump = new ApexHeapDump(
+          splitLine[3] /* heapDumpId */,
+          splitLine[4] /* className */,
+          splitLine[5] /* namespace */,
+          Number(splitLine[6]) /* line */
```
vazexqi: This is more just as a comment to anyone reading (no changes required). See https://stackoverflow.com/questions/369220/why-should-you-not-use-number-as-a-constructor on why we should not use `new Number(xx)`. What you have with Number(xx) is OK.

packages/salesforcedx-apex-replay-debugger/package.json
vazexqi: If this is only for testing does it need to be in dependencies? Or can it go into devDependencies?

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -163,13 +164,63 @@ export class ScopeContainer extends VariableContainer {
...
+    | undefined {
+    return this.overlaySuccessResut;
+  }
+  public setOverlaySuccessResult(
+    overlaySuccessResult: ApexExecutionOverlayResultCommandSuccess
+  ): void {
+    this.overlaySuccessResut = overlaySuccessResult;
+  }
+  /* tslint:disable */
+  public toString = (): string => {
```
vazexqi: What is the reason for defining `toString` this way instead of `public toString():string {...}`? Won't that avoid the issue with the semicolon?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -23,5 +23,14 @@ export const messages = {
   no_log_file_text:
     'The log file either is missing or does not have any log lines in it.',
   incorrect_log_levels_text:
-    'The log must be generated with log categories Apex code at the FINEST level and Visualforce at the FINER level.'
+    'The log must be generated with log categories Apex code at the FINEST level and Visualforce at the FINER level.',
+  unable_to_retrieve_org_info: 'Unable to retrieve OrgInfo',
+  heap_dump_error:
+    'Problems were encountered while retrieving heap dump information. Message=%s, ErrorCode=%s. Heap Dump Information=%s',
```
ruthemmanuelle: Is there a reason there's a period instead of a comma after the error code message, and why there are spaces in `Heap Dump Information` but not in `ErrorCode`?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
ruthemmanuelle: @JimSuplizio , there's no need to add an extra step by asking me to review the messages via email or Slack. We can just review them on the PR in the future. If you think there's a good reason to review them elsewhere, we can talk about it. Maybe we should put it on our parking lot agenda for standup?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
ruthemmanuelle: `heap_dump_error_wrap_up_text` is the message I reviewed with Jim before being out of the office for two days. I'll suggest a couple of minor changes to the other messages.

packages/salesforcedx-apex-replay-debugger/package.json
JimSuplizio: I put it in the same place it was used for the vscode-apex-debugger extension

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: It's reporting the error because it was unable to get the OrgInfo, and it adds the error message to the aforementioned string. I don't know what else to do here aside from report the message. Keep in mind that none of the checkpoint processing failing will prevent the replay debugger from starting it's session, it is reported to the user and the debugging sessions will happen but without the heap dumps (or if it were just one of several that it couldn't get, without that particular one).

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
JimSuplizio: Good catch Ruth, I'll fix them both.

packages/salesforcedx-apex-replay-debugger/src/core/heapDump.ts
```diff
@@ -0,0 +1,56 @@
...
+    | undefined {
+    return this.overlaySuccessResut;
+  }
+  public setOverlaySuccessResult(
+    overlaySuccessResult: ApexExecutionOverlayResultCommandSuccess
+  ): void {
+    this.overlaySuccessResut = overlaySuccessResult;
+  }
+  /* tslint:disable */
+  public toString = (): string => {
```
vazexqi: I thought I left a comment here but I must not have clicked "Add Single Comment".

Why do we need to write the method using the `public toString = (): string => {..}` style? Would a normal override work? That would seem to eliminate the issues that you are seeing with tslint.

See https://stackoverflow.com/questions/35361482/typescript-override-tostring/45320960 and the fiddle at https://jsfiddle.net/sy8wttvw/

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,98 @@ export class LogContext {
...
+    const heapDumpRegex = RegExp(/\|HEAP_DUMP\|/);
+    this.logLines.forEach((line, index) => {
+      if (heapDumpRegex.test(line)) {
+        const splitLine = line.split('|');
+        if (splitLine.length >= 7) {
+          const heapDump = new ApexHeapDump(
+            splitLine[3] /* heapDumpId */,
+            splitLine[4] /* className */,
+            splitLine[5] /* namespace */,
+            Number(splitLine[6]) /* line */
```
vazexqi: This is more of a comment (no action required).

https://stackoverflow.com/questions/369220/why-should-you-not-use-number-as-a-constructor

This is OK since you are doing `Number(xx)` instead of `new Number(xx)`.

packages/salesforcedx-apex-replay-debugger/package.json
DatGuyJonathan: apex-debugger used to have its REST services in that package, so request-light was a dependency. That stuff got moved to utils (https://github.com/forcedotcom/salesforcedx-vscode/pull/389) and all that's left in apex-debugger that needs request-light are tests. Changing the dependency probably just got missed.

You should try changing this to devDependency.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
DatGuyJonathan: Ok, since you're sure all the errors are caught, there's nothing to change here.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -99,6 +113,98 @@ export class LogContext {
...
+    let success = true;
+    try {
+      const orgInfo = await new ForceOrgDisplay().getOrgInfo(projectPath);
+      const requestService = new RequestService();
+      requestService.instanceUrl = orgInfo.instanceUrl;
+      requestService.accessToken = orgInfo.accessToken;
+
+      for (const heapDump of this.apexHeapDumps) {
+        this.session.printToDebugConsole(
+          nls.localize(`fetching_heap_dump', ${heapDump.toString()}`)
```
DatGuyJonathan: There are two backticks and a single quote. How does this message look in the debug console?

packages/salesforcedx-apex-replay-debugger/src/core/heapDump.ts
JimSuplizio: @vazexqi 
public toString(): string { *versus* public toString = (): string => {
The basic difference between the two is that the arrow function doesn't bind their own this, so this is what you'd expect. It's actually discussed in the first answer of the link you posted.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: Ack! It's leftover from copy/paste. It's weird that the syntax checker didn't flag that. The same problem was on 137 which I've fixed as well. Good catch.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
JimSuplizio: @vazexqi FYI I was curious so I'd checked to see if it it would allow either and it doesn't. new Number(xx) would get flagged as incorrect syntax.

packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
JimSuplizio: It doesn't. I've modified this to pass in undefined for the super.

packages/salesforcedx-apex-replay-debugger/src/core/heapDump.ts
vazexqi: Does yours produce a different answer if you use `public toString(): string`?

I have a fiddle here where I am using `public toString(): string` and I don't see any difference. https://jsfiddle.net/sy8wttvw/55/

I'm trying to understand why we need to use that style in our use case.

JimSuplizio: @vazexqi, I changed the function it's no longer an arrow function.

Here's a good read on arrow functions as to why I'd done what I'd done. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

vazexqi: I _think_ I understand arrow functions and when to use them. There are two things here so I don't want to conflate them:

1. Why are we using an arrow function _here_? It stood out to me not because of the tslint but because it's a different format for defining the method. That's what prompted my question to help me understand why we need that. It seems that we don't need it (no loss of functionality, or malfunction) so the normal function declaration is better.
1. The tslint issue is true and could be annoying. It means that there is a mismatch between prettier and tslint. This one should be fixed. I don't know off the top of my head where it is. It might be good to just try it in a simple context and start disabling prettier and see what happens. Then we can trace down the exact tslint rule. That's how I would do it if I were in this situation. I won't be able to tell you the rule without doing that process myself either.

DatGuyJonathan:
Oh, I'm not blocked. I really only need the heapdump interfaces and I can take that as-is, so if there's an issue that needs further discussion, feel free to continue.

PR #505, opened 2018-06-26
Update changelog for release v43.5.0

lcampos:
### What does this PR do?
Updates the changelog for release v43.5.0

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -1,3 +1,21 @@
+## 43.5.0 - June 28, 2018
+
+### Fixed
+
+#### salesforcedx-vscode-core
+
+- Extensions failing to activate after an exception happens during setup ([PR #497](https://github.com/forcedotcom/salesforcedx-vscode/pull/497))
+
+#### salesforcedx-vscode-apex, salesforcedx-vscode-visualforce, salesforcedx-vscode-lwc
+
+- Specify a document selector for Apex, Visualforce, and LWC ([PR #502](https://github.com/forcedotcom/salesforcedx-vscode/pull/502))
+
+### Added
+
+#### salesforcedx-vscode-apex
+
+- Syntax highlighting for merge function and Apex switch statements ([PR #503](https://github.com/forcedotcom/salesforcedx-vscode/pull/503))
+
 ## 43.4.0 - June 21, 2018
 
 ### Added
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -4,11 +4,7 @@
 
 #### salesforcedx-vscode-core
 
-- Extensions failing to activate after an exception happens during setup ([PR #497](https://github.com/forcedotcom/salesforcedx-vscode/pull/497))
-
-#### salesforcedx-vscode-apex, salesforcedx-vscode-visualforce, salesforcedx-vscode-lwc
-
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+
+### Fixed
+
+#### salesforcedx-vscode-core
+
+- Extensions failing to activate after an exception happens during setup ([PR #497](https://github.com/forcedotcom/salesforcedx-vscode/pull/497))
+
+#### salesforcedx-vscode-apex, salesforcedx-vscode-visualforce, salesforcedx-vscode-lwc
+
+- Specify a document selector for Apex, Visualforce, and LWC ([PR #502](https://github.com/forcedotcom/salesforcedx-vscode/pull/502))
```
lcampos: This change involves several extensions, do we want to break it per extension or should we not even have it as part of the changelog ?

ruthemmanuelle: Maybe we can put it at the top level, under `salesforcedx-vscode`, even though it's technically for individual extensions? Also, how does this help users (beyond preventing confusing output in the console)?

packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+## 43.5.0 - June 28, 2018
+
+### Fixed
+
+#### salesforcedx-vscode-core
+
+- Extensions failing to activate after an exception happens during setup ([PR #497](https://github.com/forcedotcom/salesforcedx-vscode/pull/497))
```
ruthemmanuelle: Would this be accurate? `Extensions no longer stop working after an exception is encountered during setup`

packages/salesforcedx-vscode/CHANGELOG.md
lcampos: This is more of following vscode's suggested practices when the extension needs to interact with a file.

I'd be ok if everything is top level under `salesforcedx-vscode`.

packages/salesforcedx-vscode/CHANGELOG.md
lcampos: This used to happen when reloading the editor or opening it, so not sure 'no longer stop working' captures that the extension doesn't even start/activates based from the user's perspective.

vazexqi: How about: Guard against salesforcedx-vscode-core failing to launch if there is an issue with configuring the ISV Debugger.

That seems to be more specific on what the issue is and what we fixed in that PR.

packages/salesforcedx-vscode/CHANGELOG.md
vazexqi: Let's put it at the top-level.

I don't think this directly affects the users for now. It does make our language server activation sequence safer since we only do it for files now and not other schemes.

ruthemmanuelle: If it doesn't affect users, maybe it doesn't belong in the change log?

packages/salesforcedx-vscode/CHANGELOG.md
ruthemmanuelle: Sure, Nick's version sounds good.

packages/salesforcedx-vscode/CHANGELOG.md
vazexqi: Yeah, let's not include this one.

PR #507, opened 2018-06-29
Appveyor

Sweetman:
### What does this PR do?
Stores the junit files as an artifact, displays them in the test dashboard, and determines whether a failure occurred after all the tests have been ran.
@ W-5029369@


.appveyor.yml
```diff
@@ -41,6 +42,9 @@ test_script:
       "Running npm run test:without-system-tests..."
       "--------------------------------------------"
       npm run test:without-system-tests
+      npm run coverage:system-tests
+      npm run aggregateJUnit
+      junit-merge -d junit-aggregate -o junit-aggregate.xml
```
Sweetman: Using junit-merge to merge all results into one file. There's an issue when it does it recursively so I aggregate them all into a single directory with the aggregateJUnit script.

.appveyor.yml
```diff
@@ -49,23 +53,21 @@ test_script:
...
+      { 
+          $host.SetShouldExit(1)
+          Write-Host "Forcing build failure due to test failure(s) :'("
       }
-      # Fail script in case of test failures
-      if($NpmTestsExitCode -ne 0) {
-        #$Error | Format-List * -Force
-        exit $NpmTestsExitCode
-      }
-  - npm run coverage:system-tests
```
Sweetman: Modified this to run all the tests first and then upload the single aggregate junit result file.

Sweetman: This will also fail the build if there's a test failure. I couldn't find an easy way to do this with a junit parser and found this here:
https://help.appveyor.com/discussions/questions/2504-how-to-fail-build-if-uploaded-test-results-have-failures

.appveyor.yml
```diff
@@ -49,23 +53,21 @@ test_script:
...
 
 artifacts:
   - path: '**\*.vsix'
+  - path: junit-aggregate/
+    name: test-results
+    type: zip
```
Sweetman: Points to the path with all the result files and stores them as a zip

config/mochaConfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "reporterEnabled": "mocha-junit-reporter, xunit, spec",
+  "xunitReporterOptions": {
+    "output": "xunit.xml"
+  },
+  "mochaJunitReporterReporterOptions": {
+    "mochaFile": "junit-custom.xml"
+  }
+}
```
Sweetman: Top level mochaConfig that is needed for anything running mocha directly. This is necessary for the non-vscode packages that do not run VSCode's test bin which run mocha programmatically.

package.json
```diff
@@ -51,11 +52,15 @@
...
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/forcedotcom/salesforcedx-vscode.git"
   },
-  "license": "BSD-3-Clause"
+  "license": "BSD-3-Clause",
+  "dependencies": {
+    "junit-xml-parser": "0.0.3"
+  }
```
Sweetman: I will remove this on my next commit. I tried playing around with a parser to remove querying the test dashboard for failures.

packages/salesforcedx-utils-vscode/src/test/testrunner.ts
```diff
@@ -34,6 +34,23 @@ function configure(mochaOpts: any): void {
...
+      mochaJunitReporterReporterOptions: {
+        mochaFile: xmlPath
+          ? paths.join(xmlPath, 'junit-custom.xml')
+          : 'junit-custom.xml'
+      },
+      xunitReporterOptions: {
+        output: xmlPath ? paths.join(xmlPath, 'xunit.xml') : 'xunit.xml'
+      }
+    };
+  }
```
Sweetman: Sets up the reporter options. This is essentially what the config.json files do for the non-programmatic runs of mocha. There were some oddities on Windows where the mocha execution would be inside the downloaded version of vscode which is why you see the if case.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,12 @@
 {
-  "extensions.autoUpdate": false
+  "extensions.autoUpdate": false,
```
Sweetman: Will remove this on the next commit.

packages/system-tests/src/mocha-runner.ts
```diff
@@ -10,11 +10,24 @@ import * as glob from 'glob';
...
+  reporterOptions: {
+    reporterEnabled: 'mocha-junit-reporter, xunit, spec',
+    mochaJunitReporterReporterOptions: {
+      mochaFile: 'junit-custom.xml'
+    },
+    xunitReporterOptions: {
+      output: 'xunit.xml'
+    }
+  }
+});
```
Sweetman: System tests defined its own programmatic way of running the tests so I added the necessary reporter options so it will produce test results as well.

scripts/aggregate-junit-xml.js
```diff
@@ -0,0 +1,26 @@
...
+    process.cwd(),
+    'packages',
+    dir,
+    'junit-custom.xml'
+  );
+  shell.cp(
+    currPackagePath,
+    path.join(process.cwd(), `junit-aggregate/${dir}-junit-custom.xml`)
+  );
+});
```
Sweetman: This checks each package for a junit-custom file and moves it to a top level folder if it exists with its package name.

packages/salesforcedx-utils-vscode/src/test/testrunner.ts
lcampos: Would be good to add that as a comment.

packages/system-tests/package.json
```diff
@@ -25,6 +25,8 @@
     "istanbul": "^0.4.5",
     "mkdirp": "0.5.1",
     "mocha": "3.2.0",
+    "mocha-junit-reporter": "^1.17.0",
```
DatGuyJonathan: These versions are different than the other packages. Is there a feature/bug fix in the newer version that we need for system tests?

Sweetman: No need for the later version. Downgraded it to 1.13

PR #511, opened 2018-07-02
Lcampos/telemetry

lcampos:
### What does this PR do?
This is an early PR for adding telemetry to our extensions

Adds telemetry to our vscode extensions.

Preview of some metrics being sent.

screen shot 2018-07-02 at 12 05 03 pm

screen shot 2018-07-02 at 12 04 16 pm

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/src/telemetry/index.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/src/telemetry/index.ts
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { TelemetryService } from './telemetry';
+
+export const telemetryService = TelemetryService.getInstance();
--- a/packages/salesforcedx-vscode-apex-debugger/src/telemetry/index.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/src/telemetry/index.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, salesforce.com, inc.
+ * Copyright (c) 2018, salesforce.com, inc.
  * All rights reserved.
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
--- a/packages/salesforcedx-vscode-apex-debugger/test/unit/telemetry/telemetry.test.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/test/unit/telemetry/telemetry.test.ts
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { assert, SinonStub, stub } from 'sinon';
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { TelemetryService } from '../../../src/telemetry/telemetry';
+
+describe('Telemetry', () => {
+  let reporter: TelemetryReporter;
+  let sendEvent: SinonStub;
+
+  beforeEach(() => {
+    reporter = new TelemetryReporter('salesforcedx-vscode', 'v1', 'test567890');
+    sendEvent = stub(reporter, 'sendTelemetryEvent');
+  });
+
+  afterEach(() => {
+    sendEvent.restore();
+    reporter.dispose();
+  });
+
+  it('Should send telemetry data', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, true);
+
+    telemetryService.sendExtensionActivationEvent();
+    assert.calledOnce(sendEvent);
+  });
+
+  it('Should not send telemetry data', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, false);
+
+    telemetryService.sendCommandEvent('debugger_command');
+    assert.notCalled(sendEvent);
+  });
+
+  it('Check telemetry sendCommandEvent data format', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, true);
+
+    telemetryService.sendCommandEvent('debugger_command');
+    assert.calledOnce(sendEvent);
+
+    const expectedData = {
+      extensionName: 'salesforcedx-vscode-apex-debugger',
+      commandName: 'debugger_command'
+    };
+    assert.calledWith(sendEvent, 'commandExecution', expectedData);
+  });
+
+  it('Check telemetry sendExtensionActivationEvent data format', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, true);
+
+    telemetryService.sendExtensionActivationEvent();
+    assert.calledOnce(sendEvent);
+
+    const expectedData = {
+      extensionName: 'salesforcedx-vscode-apex-debugger'
+    };
+    assert.calledWith(sendEvent, 'activationEvent', expectedData);
+  });
+
+  it('Check telemetry sendExtensionDeactivationEvent data format', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, true);
+
+    telemetryService.sendExtensionDeactivationEvent();
+    assert.calledOnce(sendEvent);
+
+    const expectedData = {
+      extensionName: 'salesforcedx-vscode-apex-debugger'
+    };
+    assert.calledWith(sendEvent, 'deactivationEvent', expectedData);
+  });
+});
--- a/packages/salesforcedx-vscode-apex-debugger/test/unit/telemetry/telemetry.test.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/test/unit/telemetry/telemetry.test.ts
@@ -38,7 +38,7 @@ describe('Telemetry', () => {
     assert.notCalled(sendEvent);
   });
 
-  it('Check telemetry sendCommandEvent data format', async () => {
+  it('Should send correct data format on sendCommandEvent', async () => {
     const telemetryService = TelemetryService.getInstance();
     telemetryService.initializeService(reporter, true);
 
@@ -52,7 +52,7 @@ describe('Telemetry', () => {
     assert.calledWith(sendEvent, 'commandExecution', expectedData);
   });
 
-  it('Check telemetry sendExtensionActivationEvent data format', async () => {
+  it('Should send correct data format on sendExtensionActivationEvent', async () => {
     const telemetryService = TelemetryService.getInstance();
     telemetryService.initializeService(reporter, true);
 
@@ -65,7 +65,7 @@ describe('Telemetry', () => {
     assert.calledWith(sendEvent, 'activationEvent', expectedData);
   });
 
-  it('Check telemetry sendExtensionDeactivationEvent data format', async () => {
+  it('Should send correct data format on sendExtensionDeactivationEvent', async () => {
     const telemetryService = TelemetryService.getInstance();
     telemetryService.initializeService(reporter, true);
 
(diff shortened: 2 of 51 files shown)


packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -181,6 +182,18 @@ function registerDebugHandlers(): vscode.Disposable {
 export async function activate(context: vscode.ExtensionContext) {
   console.log('Apex Replay Debugger Extension Activated');
 
+  const sfdxCoreExt = vscode.extensions.getExtension(
```
lcampos: This imports telemetry service from core extension which is mainly used to show the telemetry message and check for user's telemetry setting. We then initialize the extension's telemetry service.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+  ): TelemetryReporter | undefined {
+    this.context = context;
+
+    // TelemetryReporter is not initialized if user has disabled telemetry setting.
+    if (this.reporter === undefined && isTelemetryEnabled) {
+      const extensionPackage = require(this.context.asAbsolutePath(
+        './package.json'
+      ));
+
+      this.reporter = new TelemetryReporter(
```
lcampos: Initializing TelemetryReporter this way since that's how other Microsoft extensions do it (e.g. https://github.com/Microsoft/vscode-azure-account/blob/bb54795b948a4cdb111dba46f677d592aa673687/src/telemetry.ts)

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -432,6 +439,8 @@ export async function activate(context: vscode.ExtensionContext) {
 export function deactivate(): Promise<void> {
   console.log('SFDX CLI Extension Deactivated');
 
+  // Send metric data.
+  telemetryService.sendExtensionDeactivationEvent();
```
lcampos: I've seen some issues gathering this specific metric and found this issue on vscode (https://github.com/Microsoft/vscode/issues/47881).

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+        extensionPackage.version,
+        extensionPackage.aiKey
+      );
+      this.context.subscriptions.push(this.reporter);
+    }
+
+    return this.reporter;
+  }
+
+  public sendExtensionActivationEvent(): void {
```
lcampos: I want avoid having to create configuration data when using the Telemetry service across the extension so decided to create this methods. Let me know what you think of this approach.

packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -34,6 +35,16 @@ export class SfdxCoreSettings {
     return this.getConfigValue<boolean>(SHOW_CLI_SUCCESS_INFO_MSG, true);
   }
 
+  // checks for Microsoft's telemetry setting as well as Salesforce's telemetry setting.
+  public getTelemetryEnabled(): boolean {
```
lcampos: I check for vscode's telemetry.enabled setting as well as our own telemetry setting.

packages/salesforcedx-vscode-apex-replay-debugger/test/unit/telemetry/telemetry.test.ts
```diff
@@ -0,0 +1,89 @@
+import * as path from 'path';
```
vazexqi: Could you add the copyright header to this new file?

packages/salesforcedx-vscode-apex-replay-debugger/package.json
```diff
@@ -14,13 +14,16 @@
     "color": "#ECECEC",
     "theme": "light"
   },
+  "aiKey": "b093559c-0c91-43cb-ad24-67bb5a5cbc80",
```
vazexqi: This aiKey would change when Nate gets us an official account right. The key right now is your personal experimental appinsights hub?

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+  public initializeService(
+    context: vscode.ExtensionContext,
+    isTelemetryEnabled: boolean
+  ): TelemetryReporter | undefined {
+    this.context = context;
+    this.isTelemetryEnabled = isTelemetryEnabled;
+
+    // TelemetryReporter is not initialized if user has disabled telemetry setting.
+    if (this.reporter === undefined && this.isTelemetryEnabled) {
+      const extensionPackage = require(this.context.asAbsolutePath(
```
vazexqi: Is this to get the ai key? Or what else does it need from package.json?

vazexqi: NVM. I see that you are using the `require` technique to read this JSON into an object to be used later.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,62 @@
+import vscode = require('vscode');
```
vazexqi: Is there a reason we need to use `require` here instead of `import ... from .. .`?

packages/salesforcedx-vscode-core/src/index.ts
vazexqi: OK. It's probably not a deal-breaker if we can't get the deactivation command. I don't think important things happen in there for our extensions.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.context.globalState.update(TELEMETRY_GLOBAL_VALUE, true);
+  }
+
+  public showTelemetryMessage(): void {
+    // check if we've ever shown Telemetry message to user
+    const showTelemetryMessage = this.getHasTelemetryMessageBeenShown();
+
+    if (showTelemetryMessage) {
+      // this means we need to show the message and set telemetry to true;
+      const readMoreBtn = 'Read More';
```
vazexqi: Should we localize this?

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export class TelemetryService {
```
vazexqi: So this file needs to be duplicated (just like we do for the messages.ts) for each extension that wants to use the telemetry service, right?

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
vazexqi: I think this is _safer_ and more useful. So, let's go with smaller and YAGNI first.

We might need to directly expose https://github.com/Microsoft/vscode-extension-telemetry/blob/master/src/telemetryReporter.ts#L88 in the future. But we can do that later as required. I'm just worried that we if we expose too much now, it would be hard to correct things.

In particular, we need to be careful of the keys, of `public sendTelemetryEvent(eventName: string, properties?: { [key: string]: string }, measures?: { [key: string]: number })`. If the caller of that screws it up, the data might not be easily processed.

By the way, how _powerful_ is AppInsights compared to splunk? Do they have good querying tools?

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public static getInstance() {
+    if (!TelemetryService.instance) {
+      TelemetryService.instance = new TelemetryService();
+    }
+    return TelemetryService.instance;
+  }
+
+  public initializeService(
```
vazexqi: I see some opportunities to reduce code duplication. I understand why this class needs to be in the context of each extension (so that it reads the right package.json using require, which is static).

But for the other methods like `sendCommandEvent`, `sendExtensionActivationEvent`, and `sendExtensionDeactivationEvent`, we might be able to either:

1. Put them into a base class that this class extends from (see https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lwc-next/src/commands/forceLightningLwcCreate.ts#L72)
1. Use constructor dependency injection and inject different loggers into this class in the initializeService. Basically, each logging method that you have defined so far only needs access to `this.reporter` and `this.isTelemetryEnabled` so those can be passed in.

Which do you prefer?

packages/salesforcedx-vscode-apex-replay-debugger/package.json
lcampos: Right, I'm waiting on Nate to get the official app insights key and using my dev account for now.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
lcampos: No, I'll update this to use import from

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
lcampos: Yeah, will add it to the i18n file.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
lcampos: Yes, this is because we want TelemetryReporter to be initialized with each extension's version and name instead of having that information being a custom field.

Also, core telemetry service has more logic into it that will expose how to show the legal notice message and user setting config to the rest of the extensions while the rest of the extension will have a stripped down version of that.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
lcampos: Agreed, I'll try extending from a class since it seems easier to read.

lcampos: I'm going with the constructor dependency injection approach since I failed miserably on setting the core extension as a test dependency for the base class approach.

packages/salesforcedx-vscode-apex-replay-debugger/package.json
vazexqi: We should have the official keys now? @ntotten

packages/salesforcedx-vscode-apex-debugger/src/telemetry/index.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
vazexqi: Nitpick. Should say 2018.
Same with the other new files that you added.

packages/salesforcedx-vscode-apex-replay-debugger/package.json
lcampos: @vazexqi this is something I missed removing, the official keys are on salesforcedx-vscode-code/package.json

packages/salesforcedx-vscode-apex-debugger/test/unit/telemetry/telemetry.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  it('Should not send telemetry data', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, false);
+
+    telemetryService.sendCommandEvent('debugger_command');
+    assert.notCalled(sendEvent);
+  });
+
+  it('Check telemetry sendCommandEvent data format', async () => {
```
vazexqi: Nitpick. Is there a reason why we no longer use the word `Should` in the assert like we do in the previous two assertions?

packages/salesforcedx-vscode-apex-replay-debugger/package.json
ntotten: Yeah, we have had them for a while. `ec3632a4-df47-47a4-98dc-8134cacbaf7e` is for the prod account.

PR #513, opened 2018-07-03
Show non-static heapdump variables in an Apex class for topmost frame

DatGuyJonathan:
### What does this PR do?
heapdump

@W-4549497@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -25,6 +25,10 @@ import { DebugProtocol } from 'vscode-debugprotocol';
 import { breakpointUtil, LineBreakpointEventArgs } from '../breakpoints';
 
 import {
+  HeadpDumpExtentValue,
+  HeadpDumpExtentValueEntry
+} from '../commands/apexExecutionOverlayResultCommand';
+import {
   GET_LINE_BREAKPOINT_INFO_EVENT,
   LINE_BREAKPOINT_INFO_REQUEST
 } from '../constants';
@@ -82,11 +86,23 @@ export class ApexDebugStackFrameInfo {
   public readonly signature: string;
   public statics: Map<string, VariableContainer>;
   public locals: Map<string, VariableContainer>;
+
   public constructor(frameNumber: number, signature: string) {
     this.frameNumber = frameNumber;
     this.signature = signature;
     this.statics = new Map<string, VariableContainer>();
     this.locals = new Map<string, VariableContainer>();
+  }
+
+  public copy(): ApexDebugStackFrameInfo {
+    const me = new ApexDebugStackFrameInfo(this.frameNumber, this.signature);
+    this.statics.forEach((value, key) => {
+      me.statics.set(key, value.copy());
+    });
+    this.locals.forEach((value, key) => {
+      me.locals.set(key, value.copy());
+    });
+    return me;
   }
 }
 
@@ -117,6 +133,15 @@ export abstract class VariableContainer {
     });
     return result;
   }
+
+  public copy(): VariableContainer {
+    const me = Object.assign(Object.create(Object.getPrototypeOf(this)));
+    me.variables = new Map<string, VariableContainer>();
+    this.variables.forEach((value, key) => {
+      me.variables.set(key, value.copy());
+    });
+    return me;
+  }
 }
 
 export class ApexVariableContainer extends VariableContainer {
@@ -138,6 +163,16 @@ export class ApexVariableContainer extends VariableContainer {
     this.type = type;
     this.ref = ref;
     this.variablesRef = variablesRef;
+  }
+
+  public copy(): ApexVariableContainer {
+    const me = super.copy() as ApexVariableContainer;
+    me.name = this.name;
+    me.value = this.value;
+    me.type = this.type;
+    me.ref = this.ref;
+    me.variablesRef = this.variablesRef;
+    return me;
   }
 }
 
@@ -289,10 +324,19 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
-  protected async scopesRequest(
+  public async scopesRequest(
     response: DebugProtocol.ScopesResponse,
     args: DebugProtocol.ScopesArguments
   ): Promise<void> {
+    try {
+      if (this.logContext.hasHeapDumpForTopFrame()) {
+        this.logContext.copyStateForHeapDump();
+        this.replaceVariablesWithHeapDump();
+      }
+    } catch (error) {
+      this.warnToDebugConsole(nls.localize('reconcile_heapdump_error', error));
+    }
+
     response.success = true;
     const frameInfo = this.logContext.getFrameHandler().get(args.frameId);
     if (!frameInfo) {
@@ -323,7 +367,7 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
-  protected async variablesRequest(
+  public async variablesRequest(
     response: DebugProtocol.VariablesResponse,
     args: DebugProtocol.VariablesArguments
   ): Promise<void> {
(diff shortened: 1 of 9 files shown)


packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -289,10 +324,19 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
-  protected async scopesRequest(
+  public async scopesRequest(
     response: DebugProtocol.ScopesResponse,
     args: DebugProtocol.ScopesArguments
   ): Promise<void> {
+    try {
+      if (this.logContext.hasHeapDumpForTopFrame()) {
```
DatGuyJonathan: When VS Code asks for variables, it always asks for the scopes first. By this time, we have a call stack and variables from the debug log. So, we check if there’s a heapdump for the topmost frame and use that heapdump for variables. The variables from the debug log get copied by value before they are replaced by variables from heapdump.

It doesn’t verify that the current stopped line has a checkpoint. The user may have removed the checkpoint from the client, but not upload the changes to the server, so I don’t know for sure. Open for discussion.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -514,6 +558,159 @@ export class ApexReplayDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
+  public replaceVariablesWithHeapDump(): void {
```
DatGuyJonathan: Let me know if you spot refactoring opportunities.

packages/salesforcedx-apex-replay-debugger/src/core/handles.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export class Handles<T> {
```
DatGuyJonathan: This was copied from vscode-debugadapter. I needed to bring this over to add method that can copy by value. Handles is used to store stuff about frames and variables.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -303,6 +351,9 @@ export class LogContext {
   }
 
   public updateFrames(): void {
+    if (this.hasHeapDump()) {
```
DatGuyJonathan: `updateFrames` is where we continue parsing the debug log when the user steps or runs the debugger. So we revert back to the saved state.

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -32,5 +32,7 @@ export const messages = {
   fetching_heap_dump:
     'Retrieving heap dump from server. Heap dump information=%s',
   malformed_log_line:
-    "Encountered a malformed HEAP_DUMP log line, skipping. Log line number=%d, log line='%s'."
+    "Encountered a malformed HEAP_DUMP log line, skipping. Log line number=%d, log line='%s'.",
+  reconcile_heapdump_error:
+    'Problems were encountered while using heapdump: %s.'
```
DatGuyJonathan: A catch-all message in case something goes wrong with parsing the heapdump. The user should probably remove the ApexExecutionOverlayResult object from their org.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: At the point you start debugging, checkpoints (in VS Code or on the server) are 100% moot and the only thing that matters are the HEAP_DUMP entries in the logs. The fact is, the user didn't have to use VS Code to set checkpoints to create the logs, that can still all be done server side. They only things they really need to use the replay-debugger is to have the code locally and a log file.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -514,6 +558,159 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+    const topFrame = this.logContext.getTopFrame();
```
vazexqi: Based on our discussion the other day to make the replacement generic across different frames, it might be good for this to be in its on class. Also, I think it would be useful to pass the frame in as a parameter (so that it doesn't need to be the topmost frame).

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -514,6 +558,159 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+              }
+              for (const variableEntry of innerExtent.value.entry) {
+                refContainer.variables.forEach((value, key) => {
+                  if (
+                    variableEntry.keyDisplayValue ===
+                    (value as ApexVariableContainer).name
+                  ) {
+                    if (
+                      typeof variableEntry.value.value === 'string' &&
+                      (variableEntry.value.value as string).startsWith('0x')
```
vazexqi: Nitpick. I think instead of `.starsWith('0x')` it is better to refactor this into its own _intention revealing method_. See https://learn.co/lessons/intention-revealing-methods. 

In other words, we don't really care that it starts with 0x. What we care about is that it indicates that this could be a memory reference.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -289,10 +324,19 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
     response: DebugProtocol.ScopesResponse,
     args: DebugProtocol.ScopesArguments
   ): Promise<void> {
+    try {
+      if (this.logContext.hasHeapDumpForTopFrame()) {
+        this.logContext.copyStateForHeapDump();
+        this.replaceVariablesWithHeapDump();
+      }
+    } catch (error) {
+      this.warnToDebugConsole(nls.localize('reconcile_heapdump_error', error));
```
JimSuplizio: Can we get some specific information like the log line number added to the error? I realize that they're probably either stepping or running to a breakpoint but this seems like at least outputting the log line number would be good here.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
JimSuplizio: We should probably make the 'string' and '0x' constants and compare against those.

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
JimSuplizio: This is very minor but we should probably use "heap dump", two words, to be consistent with the rest of the messages.

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
vazexqi: How big is the save state? I assume we do a rather deep copy to copy all the values?

packages/salesforcedx-apex-replay-debugger/src/messages/i18n.ts
ruthemmanuelle: Thanks for suggesting what the user should do next, @DatGuyJonathan . Would there be more than one ApexExecutionOverlayResult in the org at a time? If so, how would the user know which one to delete? 

Yes, "heap dump" should be two words.

DatGuyJonathan: @ruthemmanuelle , they can have more than one ApexExecutionOverlayResult in the org, but the error will only occur for one ApexExecutionOverlayResult. We have the ID of the ApexExecutionOverlayResult object so we can possibly suggest that for their query.

ruthemmanuelle: Ok, cool. How about something like this? (I assumed the %s in your original message was a description of the problems, but if not then we can eliminate it from the first sentence of my suggested message.)

'Problems were encountered while using a heap dump: %s. Use Tooling API to delete the ApexExecutionOverlayResult record with the ID %s: From a terminal, run "sfdx force:data:record:delete -t -s ApexExecutionOverlayResult -i %s".'

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
DatGuyJonathan: Yea, I have to do a deep copy so the size of the save state depends on how much variables they've built up from the debug log :grimacing: . It probably isn't as bad as it sounds because variables from the debug log are limited (e.g. concatenated strings, collections are just the toString representation).

PR #515, opened 2018-07-05
Add SFDX: Create Change-set-based Project command

vazexqi:
### What does this PR do?

Adds a command that calls out to the Salesforce CLI to generate a manifest for working with non-scratch orgs. For those curious about this feature, it requires the latest version (unpublished) of the Salesforce CLI to work.

@W-5070910@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -51,6 +51,8 @@
     "SFDX: Execute Anonymous Apex with Currently Selected Text",
   "feature_previews_title": "Salesforce Feature Previews",
   "force_project_create_text": "SFDX: Create Project",
+  "force_project_create_change_set_based_text":
+    "SFDX: Create Change-set-based Project",
   "force_apex_trigger_create_text": "SFDX: Create Apex Trigger",
   "show_cli_success_msg_description":
     "Specifies whether status messages for Salesforce CLI commands run using the VS Code command palette will appear as pop-up information messages (true) or as status bar messages (false).",
@@ -63,11 +65,10 @@
     "SFDX: Create and Set Up Project for ISV Debugging",
   "force_apex_log_get_text": "SFDX: Get Apex Debug Logs...",
 
-  "force_manifest_editor_show_text":
-    "SFDX: Show Package Manifest Editor (Experimental)",
-  "show_experimental_webviews_description":
-    "Specifies whether the commands for the experimental webviews are available to the user.",
-
   "retrieve_test_code_coverage_text":
-    "Specifies whether code coverage results are calculated and retrieved when you run Apex tests."
+    "Specifies whether code coverage results are calculated and retrieved when you run Apex tests.",
+
+  "change_set_based_tools.enabled_description":
+    "Specifies whether the commands for change-set-based development are available to the user.",
+  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor"
 }
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -52,7 +52,7 @@
   "feature_previews_title": "Salesforce Feature Previews",
   "force_project_create_text": "SFDX: Create Project",
   "force_project_create_change_set_based_text":
-    "SFDX: Create Change-set-based Project",
+    "SFDX: Create Change-Set-Based Project",
   "force_apex_trigger_create_text": "SFDX: Create Apex Trigger",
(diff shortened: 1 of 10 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -51,6 +51,8 @@
     "SFDX: Execute Anonymous Apex with Currently Selected Text",
   "feature_previews_title": "Salesforce Feature Previews",
   "force_project_create_text": "SFDX: Create Project",
+  "force_project_create_change_set_based_text":
+    "SFDX: Create Change-set-based Project",
```
vazexqi: @ruthemmanuelle - Is this the proper casing for the command?

packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
```diff
@@ -31,14 +31,32 @@ import {
...
-    return new SfdxCommandBuilder()
+    const builder = new SfdxCommandBuilder()
       .withDescription(nls.localize('force_project_create_text'))
       .withArg('force:project:create')
       .withFlag('--projectname', data.projectName)
-      .withFlag('--outputdir', data.projectUri)
-      .build();
+      .withFlag('--outputdir', data.projectUri);
+
+    if (this.options.isChangeSetBasedProject) {
```
vazexqi: Main change is here. This provides the option to also pass the `--manifest` flag.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -555,7 +555,7 @@ export class EnterForceIdeUri implements ParametersGatherer<ForceIdeUri> {
     }
 
     return null; // all good
-  }
+  };
```
vazexqi: See https://github.com/palantir/tslint/issues/3216.

More info below in tslint.json.

tslint.json
```diff
@@ -5,6 +5,7 @@
     "no-any": false,
     "quotemark": [true, "single", "avoid-escape"],
     "member-ordering": false,
-    "no-string-based-set-timeout": false
+    "no-string-based-set-timeout": false,
```
vazexqi: Basically without this, there is a conflict between prettier and tslint.
Prettier will format and add the semicolon (and if you have it set to format on save in VS Code, this does it automatically).
But, tslint, will complain that the semicolon is unnecessary.

This change fixes it so that tslint wil allow that semicolon and play nice with prettier.

I'm opening a PR to accept this in @salesforce/dev-config but for now we can address it locally.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: No. In title-case capitalization, words that would be capitalized if they weren't hyphenated should be capitalized when they follow hyphens. This should be "SFDX: Create Change-Set-Based Project".

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -63,11 +65,10 @@
...
-  "show_experimental_webviews_description":
-    "Specifies whether the commands for the experimental webviews are available to the user.",
-
   "retrieve_test_code_coverage_text":
-    "Specifies whether code coverage results are calculated and retrieved when you run Apex tests."
+    "Specifies whether code coverage results are calculated and retrieved when you run Apex tests.",
+
+  "change_set_based_tools.enabled_description":
+    "Specifies whether the commands for change-set-based development are available to the user.",
+  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor"
```
ruthemmanuelle: Although this is accurate, I fear that it might be confusing. Change-set-based development is being presented as an alternative to package-based development (i.e., second-generation packaging). Can we say something like, "SFDX: Show Manifest Editor for package.xml"?

vazexqi: _This one probably doesn't need to be solved immediately_.

This one is tricky since it's not just limited to a file called package.xml. package.xml is the default one but the user can call it whatever she wants and can create multiple manifests. So putting package.xml in the command name might limit it from being used for differently named XML manifests.

ruthemmanuelle: Ah, I see. Maybe @ntotten has thoughts on this.

vazexqi:
@allileong - I merged this in earlier since it was a small change and there were some test fixes that were in it. Feel free to review after and we can still make changes since this is gated behind a preference.

PR #517, opened 2018-07-06
Update test doc and make build fail after uploading artifacts

Sweetman:
### What does this PR do?
Add docs for the test and make it so that failing build doesnt block artifact upload
@W-5029369@


Sweetman:
@lcampos the added paragraphs should summarize what we discussed.

I also noticed that a test failure will block the test artifact so I moved failing the build to the very end.

PR #523, opened 2018-07-11
Add SFDX: Authenticate an Org

allileong:
Renamed the existing forceAuthWebLogin to forceAuthDevHub so that
forceAuthWebLogin could be repurposed for the `SFDX: Authorize an Org` 
command.

### What does this PR do?
* Adds a new command that authenticates an org and sets it as the 
default username
* Supports demo mode, similar to the existing command `SFDX: Authorize a Dev Hub`
* Hides the new command behind the `salesforcedx-vscode-core.change_set_based_tools.enabled` perm

For change set development, users should be able to authorize their production and 
sandbox orgs.
@W-5070919@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -138,7 +138,7 @@
         {
           "command": "sfdx.force.auth.web.login",
           "when":
-            "config.salesforcedx-vscode-core.change_set_based_tools.enabled"
+            "sfdx:project_opened && config.salesforcedx-vscode-core.change_set_based_tools.enabled"
         },
         {
           "command": "sfdx.force.auth.logout.all",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -132,8 +132,13 @@
       ],
       "commandPalette": [
         {
-          "command": "sfdx.force.auth.web.login",
+          "command": "sfdx.force.auth.dev.hub",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.auth.web.login",
+          "when":
+            "config.salesforcedx-vscode-core.change_set_based_tools.enabled"
         },
         {
           "command": "sfdx.force.auth.logout.all",
@@ -298,8 +303,12 @@
     },
     "commands": [
       {
-        "command": "sfdx.force.auth.web.login",
+        "command": "sfdx.force.auth.dev.hub",
         "title": "%force_auth_web_login_authorize_dev_hub_text%"
+      },
+      {
+        "command": "sfdx.force.auth.web.login",
+        "title": "%force_auth_web_login_authorize_org_text%"
       },
       {
         "command": "sfdx.force.auth.logout.all",
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -1,74 +1,50 @@
 {
   "running_tasks_title_text": "Running Tasks",
-
   "force_auth_web_login_authorize_dev_hub_text": "SFDX: Authorize a Dev Hub",
+  "force_auth_web_login_authorize_org_text": "SFDX: Authorize an Org",
   "force_auth_logout_all_text": "SFDX: Log Out from All Authorized Orgs",
-  "force_org_create_default_scratch_org_text":
-    "SFDX: Create a Default Scratch Org...",
+  "force_org_create_default_scratch_org_text": "SFDX: Create a Default Scratch Org...",
   "force_org_open_default_scratch_org_text": "SFDX: Open Default Scratch Org",
-  "force_source_pull_default_scratch_org_text":
-    "SFDX: Pull Source from Default Scratch Org",
-  "force_source_pull_force_default_scratch_org_text":
-    "SFDX: Pull Source from Default Scratch Org and Override Conflicts",
-  "force_source_push_default_scratch_org_text":
-    "SFDX: Push Source to Default Scratch Org",
-  "force_source_push_force_default_scratch_org_text":
-    "SFDX: Push Source to Default Scratch Org and Override Conflicts",
-  "force_source_status_text":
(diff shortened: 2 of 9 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -1,74 +1,50 @@
 {
   "running_tasks_title_text": "Running Tasks",
-
   "force_auth_web_login_authorize_dev_hub_text": "SFDX: Authorize a Dev Hub",
+  "force_auth_web_login_authorize_org_text": "SFDX: Authorize an Org",
```
allileong: This line is the only change that I made to this file. The rest of the changes are a result of automatic formatting. Should I undo those changes?

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -87,6 +83,42 @@ export class ForceAuthWebDemoModeLoginExecutor extends SfdxCommandletExecutor<{}
...
+      .withArg('force:auth:web:login')
+      .withFlag('--setalias', data.alias)
+      .withArg('--setdefaultusername')
+      .withArg('--noprompt')
+      .withJson()
+      .build();
+  }
+}
+
+export class AliasGatherer implements ParametersGatherer<Alias> {
```
allileong: This code was copied from forceOrgCreate.ts so that I could customize the default alias. Is there a better way for me to reuse the AliasGatherer from that file while still being able to customize the default alias?

packages/salesforcedx-vscode-core/package.nls.json
vazexqi: Are you using prettier to format? Did you configure any other options for prettier? The only option we have is to use single quotes instead of double.

vazexqi: It appears that your default file length for prettier might be longer than the default 80 that we use.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -2,6 +2,7 @@
   "running_tasks_title_text": "Running Tasks",
 
   "force_auth_web_login_authorize_dev_hub_text": "SFDX: Authorize a Dev Hub",
+  "force_auth_web_login_authorize_org_text": "SFDX: Authorize an Org",
```
allileong: @vazexqi, it turns out I needed to install the prettier extension for vscode. It's wrapping the lines as expected now.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -132,9 +132,14 @@
...
       "commandPalette": [
         {
-          "command": "sfdx.force.auth.web.login",
+          "command": "sfdx.force.auth.dev.hub",
           "when": "sfdx:project_opened"
         },
+        {
+          "command": "sfdx.force.auth.web.login",
+          "when":
+            "config.salesforcedx-vscode-core.change_set_based_tools.enabled"
```
vazexqi: Is this a mistake? Looking at the diff it seems that sfdx.force.auth.web.login existed before but it had a different when clause. If we change it to "config.salesforcedx-vscode-core.change_set_based_tools.enabled" that means that it will only show when that pref is set. Is that intentional?

packages/salesforcedx-vscode-core/package.nls.json
vazexqi: @ruthemmanuelle, @ntotten  - Should this say "Authorize a Prod/Scratch Org" to make it more clear what kind of org it is that we are authorizing against?

packages/salesforcedx-vscode-core/package.json
vazexqi: NVM. I think I see what you are doing. You are actually changing the semantics of this command to do something different. That was what confused me. So previously, sfdx.force.auth.web.login was use to authorize a dev hub but now you are using it to actually authenticate against a prod/scratch org.

Two things:

1. The when clause should be config.salesforcedx-vscode-core.change_set_based_tools.enabled && sfdx:project_opened since you want this to be only invocable if both conditions are true. You still want an sfdx-project.json to be present.
2. In the future, we cannot repurpose command names. When the analytics from @lcampos goes in and he logs the commands, it will be very confusing to see the command names do different things. Since the analytics hasn't gone in yet, this should be OK.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
vazexqi: I think this is fine. The alternative would be to subclass (not a fan of subclassing) or to create a new method that can be delegated to.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Does that mean that people can't authorize sandboxes? Is that because we don't prompt for the --instanceurl parameter? Do we intend to let them authorize sandboxes using this command in the future?

ntotten: They should be able to authorize a sandbox also. I feel like we should just keep it as org. Basically, there are orgs and dev hubs.

packages/salesforcedx-vscode-core/package.json
allileong: Sorry for the confusion about the renaming of the command name. That is a great point about analytics for the future. I have updated the conditions of the `when` clause to check that the user is in an sfdx project.

packages/salesforcedx-vscode-core/package.nls.json
vazexqi: Sorry I wrote wrongly. I meant "Authorize a Prod/Sandbox Org". The reason I was pedantic about this is because I don't want customers to use this command this to to authorize a scratch org. That's why I wanted to say Prod/Sandbox.

But, I'm OK with the current Authorize an Org.

If we don't see the instanceurl it will use the one from sfdx-project.json. But that's a good point about the instanceurl. We should see if we need to add that to the prompt. My inclination is not to. If they need that they can set the sfdx-project.json.

PR #526, opened 2018-07-12
Use package-lock on repo to lock module versions being used.

lcampos:
### What does this PR do?
Use package-lock at top level of the repo to lock down module version's being used. This is to address current security issues with eslint-scope (https://status.npmjs.org/incidents/dn7c1fgrr7ng)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6177 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@salesforce/dev-config": {
+      "version": "1.1.0",
+      "resolved": "http://npm.lwcjs.org/@salesforce/dev-config/-/dev-config-1.1.0/d3277925b6086114b3a3f33c37cb97f6b19e5c4f.tgz",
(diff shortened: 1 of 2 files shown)


lcampos:
@vazexqi unfortunately our use of lerna is preventing us from having per package package-lock.json files (https://github.com/lerna/lerna/issues/1462). I'll continue investigating how to accomplish this since lwc & lwc-next also use eslint.

package-lock.json
```diff
@@ -0,0 +1,6177 @@
...
+      "dev": true,
+      "requires": {
+        "ajv": "5.5.2",
+        "babel-code-frame": "6.26.0",
+        "chalk": "2.4.1",
+        "concat-stream": "1.6.2",
+        "cross-spawn": "5.1.0",
+        "debug": "3.1.0",
+        "doctrine": "2.1.0",
+        "eslint-scope": "3.7.3",
```
lcampos: Locking down to this version which already contains the fix (https://github.com/eslint/eslint-scope/issues/42)

PR #527, opened 2018-07-17
Alli/retrieve command

allileong:
### What does this PR do?
* Adds a new `SFDX: Retrieve Source From Org` command that retrieves source for a given package manifest file or source path in the workspace
* Replaces the CancellableStatusBar with a cancellable progress notification that will pop up at the start of a command and disappear when the command returns

@W-5070934@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -88,6 +88,13 @@
           "when": "view == sfdx.force.tasks.view"
         }
       ],
+      "editor/context": [
+        {
+          "command": "sfdx.force.source.retrieve.current.file",
```
allileong: The command appears in the dropdown menu when a user right-clicks anywhere inside of the an open editor.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -128,6 +135,11 @@
           "command": "sfdx.force.apex.trigger.create",
           "when":
             "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lightningcomponents && sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.source.retrieve",
```
allileong: The command also appears in the dropdown menu when a user right-clicks on a file from the Explorer side-panel

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -71,5 +71,8 @@
 
   "change_set_based_tools.enabled_description":
     "Specifies whether the commands for change-set-based development are available to the user.",
-  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor"
+  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor",
+  "force_source_retrieve_text": "SFDX: Retrieve Source from Org",
+  "force_source_retrieve_this_source_file_text":
+    "SFDX: Retrieve This Source File from Org"
```
allileong: @ruthemmanuelle , this is the text for when a user right clicks in the context of an open file in the editor. The idea is the signal that if they call the command from here, only the current file will be retrieved, but I'm not sure how best to phrase it.
screen shot 2018-07-17 at 4 13 36 pm

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -71,5 +71,8 @@
 
   "change_set_based_tools.enabled_description":
     "Specifies whether the commands for change-set-based development are available to the user.",
-  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor"
+  "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor",
+  "force_source_retrieve_text": "SFDX: Retrieve Source from Org",
```
allileong: This is the text for when a user right-clicks on a file path in the file explorer context. The file that the user clicks on could be a source file or a manifest file, so the message is more general.
screen shot 2018-07-17 at 4 17 30 pm

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -334,18 +334,23 @@ export abstract class SfdxCommandletExecutor<T>
       execution,
       cancellationToken
     );
-    CancellableStatusBar.show(execution, cancellationTokenSource);
+    await ProgressNotification.show(execution, cancellationTokenSource);
+    // CancellableStatusBar.show(execution, cancellationTokenSource); TODO: Remove this line when ProgressNotification is stable
```
allileong: Nick suggested commenting out the CancellableStatusBar for now, until we can be certain that the new progress notification works as expected.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -24,6 +24,8 @@ export const messages = {
   channel_end_with_error: 'ended with error %s',
   channel_end: 'ended',
 
+  progress_notification_text: 'Running %s',
```
allileong: screen shot 2018-07-17 at 4 29 03 pm

packages/salesforcedx-vscode-core/src/notifications/progressNotification.ts
```diff
@@ -0,0 +1,40 @@
...
+import { nls } from '../../src/messages';
+
+import { CommandExecution } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+
+export class ProgressNotification {
+  public static async show(
+    execution: CommandExecution,
+    token: vscode.CancellationTokenSource
+  ) {
+    await vscode.window.withProgress(
```
allileong: vscode.window.withProgress() is async, so I had to update the `execute()` and `attachExecution()` methods to also be async.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Oh, thanks for the info. Is there a reason this isn't in the i18n.ts file? I usually only look at changes to messages files when PRs come through, so I hadn't seen this change when I approved your PR. This change looks ok though.

ruthemmanuelle: Thanks for the info, @allileong . If possible I'd prefer that all customer-facing text be in the i18n.ts files. That way if someone decides to translate our extensions into other languages they'll get all of the text, and it also makes my job a lot easier, because whenever pull requests come through I just look at the messages files. (Not everyone always remembers to add me as a reviewer or to point out what has changed.)

ruthemmanuelle: Ah, interesting. Ok, thanks for trying.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
amphro: Does cancel kill the process?

packages/salesforcedx-vscode-core/test/commands/forceSourceRetrieve.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    );
+    expect(sourceRetrieveCommand.description).to.equal(
+      nls.localize('force_source_retrieve_text')
+    );
+  });
+});
+
+describe('Manifest or Sourcepath Gatherer', () => {
+  it("Should return an object of type 'Source'", async () => {
+    if (vscode.workspace.rootPath) {
```
lcampos: Replace this with workspaceFolders.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
allileong: @amphro Yep, it kills the process.

PR #528, opened 2018-07-18
Show child variables of an object

DatGuyJonathan:
### What does this PR do?
Most of the logic was already in place. I did a bunch of refactoring so I can call the same methods for parent and child variables.

@W-5068072@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
@@ -40,42 +40,42 @@ export interface ApexExecutionOverlayResultCommandSuccess {
 
 export interface HeapDump {
   className: string;
-  extents: HeadpDumpExtents[];
+  extents: HeapDumpExtents[];
   heapDumpDate: Date;
   namespace: string;
 }
 
-export interface HeadpDumpExtents {
+export interface HeapDumpExtents {
   collectionType: string | null;
   count: number;
-  definition: HeadpDumpCollectionTypeDefinition[];
-  extent: HeadpDumpExtent[];
+  definition: HeapDumpCollectionTypeDefinition[];
+  extent: HeapDumpExtent[];
   totalSize: number;
   typeName: string;
 }
 
-export interface HeadpDumpCollectionTypeDefinition {
+export interface HeapDumpCollectionTypeDefinition {
   name: string;
   type: string;
 }
 
-export interface HeadpDumpExtent {
+export interface HeapDumpExtent {
   address: string;
   size: number;
   symbols: string[] | null;
-  value: HeadpDumpExtentValue;
+  value: HeapDumpExtentValue;
 }
 
 // The Extent value has to be an any. The value can be a single value of varying
 // types or an array of values of varying types from a collection.
-export interface HeadpDumpExtentValue {
+export interface HeapDumpExtentValue {
   value?: any;
-  entry?: HeadpDumpExtentValueEntry[];
+  entry?: HeapDumpExtentValueEntry[];
 }
 
-export interface HeadpDumpExtentValueEntry {
+export interface HeapDumpExtentValueEntry {
   keyDisplayValue: string;
-  value: HeadpDumpExtentValue;
+  value: HeapDumpExtentValue;
 }
 
 export interface HeapDumpApexResult {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-apex-replay-debugger/src/commands/apexExecutionOverlayResultCommand.ts
```diff
@@ -40,42 +40,42 @@ export interface ApexExecutionOverlayResultCommandSuccess {
 
 export interface HeapDump {
   className: string;
-  extents: HeadpDumpExtents[];
```
DatGuyJonathan: Fixing typos

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -136,49 +127,123 @@ export class HeapDumpService {
...
-        if (extentValueEntry.keyDisplayValue === varContainer.name) {
-          return this.updateVariableContainerWithExtentValueOrEntry(
+        if (this.isContainerForExtentEntry(varContainer, extentValueEntry)) {
+          return this.updateContainerWithExtentValueOrEntry(
             varContainer,
             extentValueEntry.value.value,
             extentValueEntry.value.entry
           );
         }
         varContainer.variables.forEach((value, key) => {
```
JimSuplizio: This inner loop isn't correct. What this will end up doing is that it'll end up assigning each variable in the container to the extent values that it's looping through in the outer loop. When it's done, every variable will end up having the value of the last extent. It should only be updating the value for the key that matches the keyDisplayValue in the HeapDumpExtentValueEntry

DatGuyJonathan: Did you fix that in your current work iteM? The other day you mentioned fixing a couple of existing bugs. One was `if (heapDumpExtentValue.value) {` is with a boolean value. What's the other one?

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -136,49 +127,123 @@ export class HeapDumpService {
...
+  public updateContainer(
     varContainer: ApexVariableContainer,
-    heapDumpExtentValue: HeadpDumpExtentValue,
-    type?: string
+    heapDumpExtentValue: HeapDumpExtentValue,
+    extentToRevisit: Map<string, HeapDumpExtentValue>,
+    referenceToRevisit: Map<string, ApexVariableContainer>,
+    variableTypes: Map<string, string>
   ): void {
     if (heapDumpExtentValue.value) {
```
JimSuplizio: Whenever we're checking for the existence of an actual variable value we should be using if (typeof heapDumpExtentValue.value !== 'undefined'). The reason for this is that if (heapDumpExtentValue.value) would evaluate to false if the value was a boolean set to false.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -136,49 +127,123 @@ export class HeapDumpService {
...
 
-  public updateVariableContainerWithExtentValueOrEntry(
+  public updateContainerWithExtentValueOrEntry(
     varContainer: ApexVariableContainer,
     extentValue: any,
-    extentEntry: HeadpDumpExtentValueEntry[] | undefined
+    extentEntry: HeapDumpExtentValueEntry[] | undefined,
+    type?: string
   ): void {
     if (extentValue) {
```
JimSuplizio: Same issue where if the value happens to be a boolean set to false. We should be checking whether or not typeof extentValue !== 'undefined'

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
DatGuyJonathan: @JimSuplizio ^^^

JimSuplizio: @DatGuyJonathan At this point I've got the fixes locally. If this is working for you then check it in and I'll update for the loop and value checks in my version of your branch.

DatGuyJonathan:
@JimSuplizio - For future reference, please don't announce you fixed an issue then comment on someone else's PR about that same issue for them to fix also. That was very confusing :unamused:

PR #536, opened 2018-07-30
*WIP* Fix Hanging Progress Bar

allileong:
### What does this PR do?
Fixes the hanging progress bar that prevents 
* `SFDX: Refresh SObject Definitions`
* `SFDX: Turn on Apex Debug Log for Replay Debugger`
* `SFDX: Get Apex Debug Logs...`

from executing. 
W-5227494

### DIFF
--- a/packages/salesforcedx-vscode-core/src/notifications/progressNotification.ts
+++ b/packages/salesforcedx-vscode-core/src/notifications/progressNotification.ts
@@ -14,16 +14,17 @@ export class ProgressNotification {
     execution: CommandExecution,
     token: vscode.CancellationTokenSource
   ) {
-    await vscode.window.withProgress(
+    return vscode.window.withProgress(
       {
         title: nls.localize('progress_notification_text', execution.command),
         location: vscode.ProgressLocation.Notification,
         cancellable: true
       },
-      async (progress, cancellationToken) => {
+      (progress, cancellationToken) => {
         return new Promise(resolve => {
           cancellationToken.onCancellationRequested(() => {
             token.cancel();
+            return resolve();
           });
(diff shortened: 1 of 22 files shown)


packages/salesforcedx-vscode-core/src/notifications/progressNotification.ts
```diff
@@ -10,20 +10,21 @@ import { nls } from '../../src/messages';
 import { CommandExecution } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 
 export class ProgressNotification {
-  public static async show(
+  public static show(
     execution: CommandExecution,
     token: vscode.CancellationTokenSource
   ) {
-    await vscode.window.withProgress(
+    return vscode.window.withProgress(
```
allileong: In my original pr (https://github.com/forcedotcom/salesforcedx-vscode/pull/527), I made the assumption that because `withProgress` returns a promise, I had to await the resolution of the promise before continuing on with the execution of the remaining command code. However, this was a wrong assumption. `WithProgress` sets up listeners that listen for the success/error exit codes of commands before resolving,  and for a few commands, waiting on `withProgress` to resolve actually prevented the command from even being executed at all. As a result, the listeners on `withProgress` never received exit subjects, `withProgress` never resolved, the command code was blocked from being executed, and the looping continued.

allileong:
@vazexqi I did not look into adding that preference to the progress notification. Right now, if you select that preference on one of the success popups, subsequent success messages will only be shown in the status bar, but the progress will still be shown in the progress notification popup. Should I add that preference to the progress notification make the experience more consistent?

PR #537, opened 2018-07-30
Fix SFDX: Deploy This Source File To Org

allileong:
### What does this PR do?
Previously, the command code was not properly wired and selecting
`SFDX: Deploy This Source File To Org` from the right click menu resulted in no action.

@W-5070947@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -97,6 +97,10 @@ function registerCommands(
     'sfdx.force.source.deploy',
     forceSourceDeploy
   );
+  const forceSourceDeployCurrentFileCmd = vscode.commands.registerCommand(
+    'sfdx.force.source.deploy.current.file',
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -97,6 +97,10 @@ function registerCommands(
     'sfdx.force.source.deploy',
     forceSourceDeploy
   );
+  const forceSourceDeployCurrentFileCmd = vscode.commands.registerCommand(
```
allileong: The rest of the implementation was checked in as a part of this PR: https://github.com/forcedotcom/salesforcedx-vscode/commit/36f05c627e92267d10238a225129b939bd620a35

I noticed during manual testing that when right-clicking on a file and selecting `SFDX: Deploy This Source File To Org` produced no action because the command was not initialized properly. 
screen shot 2018-07-30 at 5 11 25 pm

PR #538, opened 2018-07-31
First update for Trigger context vars in heapdumps

JimSuplizio:
### What does this PR do?
This is the matching VS Code side changes for adding the Trigger context variables to the heap dump. Like the live debugger these variables show up under the Global variables. 

It's probably worth adding here that collections are not displayed correctly as of yet. That's the next bug I'm working on.

triggercontext

@W-5186304@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
@@ -10,10 +10,11 @@ import { ApexVariableContainer } from '../adapter/apexReplayDebug';
 import {
   ApexExecutionOverlayResultCommandSuccess,
   HeapDumpCollectionTypeDefinition,
+  HeapDumpExtents,
   HeapDumpExtentValue,
   HeapDumpExtentValueEntry
 } from '../commands/apexExecutionOverlayResultCommand';
-import { ADDRESS_PREFIX } from '../constants';
+import { ADDRESS_PREFIX, EXTENT_TRIGGER_PREFIX } from '../constants';
 import { LogContext } from './logContext';
 
 export class HeapDumpService {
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -170,7 +224,7 @@ export class HeapDumpService {
     referenceToRevisit: Map<string, ApexVariableContainer>,
     variableTypes: Map<string, string>
   ): void {
-    if (heapDumpExtentValue.value) {
+    if (typeof heapDumpExtentValue.value !== 'undefined') {
```
JimSuplizio: The reason for this change was that the original if statement was checking heapDumpExtentValue.value and if that value was a boolean with a value of false it would fail the if statement. I'd run into this processing the Trigger context boolean values. There's another if statement that had to be changed below.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -320,4 +377,19 @@ export class HeapDumpService {
   ) {
     return varContainer.name === extentValueEntry.keyDisplayValue;
   }
+
```
JimSuplizio: The Trigger context variables are Boolean, List and Map types and will have symbols lists of variables that starts with "Trigger.".

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
vazexqi: ✅

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -320,4 +377,19 @@ export class HeapDumpService {
...
+  public isTriggerExtent(outerExtent: HeapDumpExtents) {
+    if (
+      (outerExtent.typeName === 'Boolean' ||
+        outerExtent.typeName.startsWith('List') ||
```
vazexqi: Could I have a custom class called `ListA` and could this mess things up?

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -320,4 +377,19 @@ export class HeapDumpService {
...
+        outerExtent.typeName.startsWith('Map')) &&
```
vazexqi: Ditto.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: You're right. The List and Map typenames should both have the "<" as part of their comparison strings. Good catch @vazexqi!

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: Commented previously. The List and Map typenames should both have the "<" as part of their comparison strings.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -320,4 +377,19 @@ export class HeapDumpService {
   ) {
     return varContainer.name === extentValueEntry.keyDisplayValue;
   }
+
+  public isTriggerExtent(outerExtent: HeapDumpExtents) {
+    if (
+      (outerExtent.typeName === 'Boolean' ||
```
vazexqi: We control the output of the trigger variables so we are certain that we can use `Boolean`, `List<`, etc and not have to do a case-insensitive comparison, right? Or do we also need to do case-insensitive comparisons?

JimSuplizio: That's a good question. The answer is that we do not need a case insensitive comparison. While Apex, itself, may see Boolean and boolean in code as the same thing, the type is still Boolean. That being said, when outputting the types in the extent there is no variance in casing, it uses the actual type. These are the types for the Trigger context variables directly from the heapdump extents: 
"typeName": "List<Account>"
"typeName": "Map<Id,Account>"
"typeName": "Boolean"

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -320,4 +377,19 @@ export class HeapDumpService {
...
+        outerExtent.typeName.startsWith('List<') ||
+        outerExtent.typeName.startsWith('Map<')) &&
```
vazexqi: And the reason we don't have `Set<` is because none of the trigger variables are sets, right (https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_triggers_context_variables.htm)?

Would it make sense to make this more future-proof and just add `outerExtent.typeName.startsWith('Set<')` as well?

JimSuplizio: I had a quick chat with Mike Boilen on this, the List and Map for types would not be changing to Set. We're good here.

JimSuplizio:
As far as the testing goes, the unit style is great for the base extents parsing and since that was already done by Jonathan I'd opted for the example heapdump. We're always going to have the 6 booleans, and at least 1 list and 1 map.

PR #541, opened 2018-08-01
Change 'web' to 'lwc' for LWC

vazexqi:
### What does this PR do?

We are changing the contract for the CLI for LWC in 218 and beyond. Thus, this change is done in lwc-next and will go out in 218. 

@W-5267747@


vazexqi:
@ruthemmanuelle - Correct. That is why this is in the lwc-next first so that it won't impact current users.

ruthemmanuelle:
@vazexqi - Ok, cool. Just checking, since your PR description mentions 218, and v44 is 216.

vazexqi:
Aha! My bad. I've edited it to make it clearer.

PR #542, opened 2018-08-01
Adding Testing Sidebar for ApexClasses

jkalloor0701:
### What does this PR do?
This PR adds a sidebar that allows you to view, run, and interact with apex tests in your project.
PR fixes the lack of visibility in terms of running tests and viewing their errors.


jkalloor0701:
@vazexqi @allileong @lcampos

.appveyor.yml
```diff
@@ -54,6 +54,13 @@ test_script:
       # Upload results for AppVeyor test tab
       "Uploading Test Results:`n"
       (New-Object 'System.Net.WebClient').UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)" , (Resolve-Path .\junit-aggregate.xml) )
+      [xml]$results = Get-Content junit-aggregate.xml
```
vazexqi: Please revert this. Don't change any of our build files.

.vscode/settings.json
```diff
@@ -14,5 +14,14 @@
   "editor.tabSize": 2,
   "editor.formatOnSave": true,
   "prettier.singleQuote": true,
-  "rewrap.wrappingColumn": 80
+  "rewrap.wrappingColumn": 80,
```
vazexqi: Revert this.

package.json
```diff
@@ -47,6 +47,11 @@
       "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
     "synk:monitor":
       "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false --org=vazexqi",
+    "start-webview": "npm run start --prefix=packages/salesforcedx-webview-ui",
```
vazexqi: I think you need to rebase onto the latest version from `develop`.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -49,9 +52,89 @@
...
+          "id": "sfdx.force.test.view",
+          "name": "Apex Tests"
+        }
+      ]
+    },
+    "menus": {
+      "view/title": [
+        {
+          "command": "sfdx.force.test.view.run",
+          "when": "view == sfdx.force.test.view",
```
vazexqi: We need to come up with an analysis of whether we can support multiple test panels since that will inform us on how to continue.

package.json
lcampos: These scripts were previously removed, any reason you need them?

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -49,9 +52,89 @@
...
+          "group": "inline"
+        }
+      ]
+    },
+    "commands": [
+      {
+        "command": "sfdx.force.test.view.run",
+        "title": "Run Tests",
+        "icon": {
+          "light": "resources/light/play-button.svg",
```
vazexqi: Where did we get the resources from? Are they OK from a copyright/IP perspective?

packages/salesforcedx-vscode-apex/scripts/prep-jar.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-JORJE_LOCATION=/Users/nchen/Development/IdeaProjects/apex-jorje
+JORJE_LOCATION=/Users/jkalloor/apex-jorje
```
vazexqi: Please revert this. Or not check this file in.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -12,23 +12,84 @@ import {
   DEBUGGER_LINE_BREAKPOINTS
 } from './constants';
 import * as languageServer from './languageServer';
+import { ApexTestOutlineProvider } from './views/testOutline';
+
+export type ApexTestRequestInfo = {
```
vazexqi: TODO: Need to reconcile this with the changes that are happening on the jorje side.

packages/salesforcedx-vscode-apex/package.json
jkalloor0701: We got them mocha sidebar. @ntotten said that there is no problem with using them

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -12,23 +12,84 @@ import {
...
+  }
+
+  const apexClassesDocs = await vscode.workspace.findFiles('**/*.cls');
+  const testOutlineProvider = new ApexTestOutlineProvider(
+    rootPath,
+    apexClassesDocs,
+    apexClasses
+  );
+
+  const testViewItems = new Array<vscode.Disposable>();
```
vazexqi: Could you refactor the following parts about running test into its own method to make it cleaner? The `activate` method is getting too long.

packages/salesforcedx-vscode-apex/scripts/prep-jar.sh
lcampos: Let's not check in changes to this file

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -41,6 +102,14 @@ async function getLineBreakpointInfo(): Promise<{}> {
   return Promise.resolve(response);
 }
 
+export async function getApexTests(): Promise<{}> {
+  let response = {};
+  if (languageClient) {
+    response = await languageClient.sendRequest('apextest/getTestMethods');
```
vazexqi: TODO: Need to reconcile this with the changes that are happening on the jorje side.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -41,6 +102,14 @@ async function getLineBreakpointInfo(): Promise<{}> {
   return Promise.resolve(response);
 }
 
+export async function getApexTests(): Promise<{}> {
```
vazexqi: Should this be a more specific type rather than `{}`?

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -24,6 +24,7 @@ export const messages = {
   source_missing_text: '%s points to a missing folder',
   java_runtime_missing_text:
     'Java runtime could not be located. Set one using the salesforcedx-vscode-apex.java.home VS Code setting.',
+  force_apex_test_run_codeAction_description_text: 'Run Apex test(s)',
```
vazexqi: @ruthemmanuelle - Text review?

packages/salesforcedx-vscode-apex/test/views/fakeFiles.ts
```diff
@@ -0,0 +1,69 @@
+const FILE_0 = `public class file0 {
```
vazexqi: Ignoring this file for now until we get the design/architecture nailed down.

packages/salesforcedx-vscode-apex/tsconfig.json
```diff
@@ -1,22 +1,16 @@
...
-        "strict": true
-    },
-    "exclude": [
-        "node_modules",
-        ".vscode-test",
-        "out"
-    ]
-}
\ No newline at end of file
+  "compilerOptions": {
```
vazexqi: Was this just a formatting or did we change anything here?

packages/system-tests/package.json
```diff
@@ -48,4 +48,4 @@
   "activationEvents": [
     "*"
   ]
-}
+}
```
vazexqi: Revert this file.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  }
+
+  public getHead(): Test {
+    if (this.head === null) {
+      return this.getAllApexTests(this.path);
+    }
+    return this.head;
+  }
+
+  public getChildren(element: Test): Thenable<Test[]> {
```
vazexqi: Must we use `Thenable`? Can we use a combination of async/await? Shouldn't mix different styles of programming paradigms if possible.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+    if (element) {
+      return Promise.resolve(element.children);
+    } else {
+      if (this.head && this.head.children.length > 0) {
+        return Promise.resolve(this.head.children);
+      } else {
+        const emptyArray = new Array<ApexTest>();
+        emptyArray.push(new ApexTest('No tests in folder', null, 0));
```
vazexqi: @jkalloor0701 - All messages should be externalized.
@ruthemmanuelle - Comments?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  private _onDidChangeTreeData: vscode.EventEmitter<
+    Test | undefined
+  > = new vscode.EventEmitter<Test | undefined>();
+  public readonly onDidChangeTreeData: vscode.Event<Test | undefined> = this
+    ._onDidChangeTreeData.event;
+
+  private eventsEmitter = new events.EventEmitter();
+
+  private apexTestMap: Map<string, Test> = new Map<string, Test>();
+  private head: Test | null;
```
vazexqi: Pick a better name than `head`.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  NamespacPrefix: string;
+};
+
+export type ApexGroupInfo = {
+  outcome: string;
+  summary: TestSummary;
+};
+
+export class ApexTestOutlineProvider implements vscode.TreeDataProvider<Test> {
+  private _onDidChangeTreeData: vscode.EventEmitter<
```
vazexqi: Don't use the `_convention` for private methods/fields.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -12,23 +12,84 @@ import {
...
+  }
+
+  // Test View
+  const rootPath = vscode.workspace.rootPath || '';
+  let apexClasses: ApexTestRequestInfo[] | null = null;
+  if (isLanguageClientReady()) {
+    apexClasses = (await getApexTests()) as ApexTestRequestInfo[];
+  }
+
+  const apexClassesDocs = await vscode.workspace.findFiles('**/*.cls');
```
vazexqi: What is this for? Also, this is not correct. You cannot just find all the .cls files in a project. You need to look into sfdx-project.json and only look into parts that are in the packageDirectories. That is what the Apex LSP does.

So, if you do it this way, there is a possibility that the results will be different.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+    }
+  }
+
+  public getTreeItem(element: Test): vscode.TreeItem {
+    if (element) {
+      return element;
+    } else {
+      this.getAllApexTests(this.path);
+      if (!(this.head && this.head.children.length > 0)) {
+        this.head = new ApexTest('No tests in folder', null, 0);
```
vazexqi: @jkalloor0701 - All messages should be externalized.
@ruthemmanuelle - Comments?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+
+  public async refresh() {
+    this.head = null; // Reset tests
+    this.apexTestMap.clear();
+    ApexTestOutlineProvider.testStrings.clear();
+    this.apexTestInfo = null;
+    if (isLanguageClientReady()) {
+      this.apexTestInfo = (await getApexTests()) as ApexTestRequestInfo[];
+    }
+    this.apexClasses = await vscode.workspace.findFiles('**/*.cls');
```
vazexqi: This is wrong. See previous comment on the explanation why.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -49,9 +52,89 @@
     "test": "node ./node_modules/vscode/bin/test",
     "test:unit": "node ./node_modules/vscode/bin/test"
   },
-  "activationEvents": ["workspaceContains:sfdx-project.json"],
+  "activationEvents": [
+    "workspaceContains:sfdx-project.json",
+    "onCommand:sfdx.force.test.view.showError",
```
lcampos: Why do we need to add this new activation events if the extension would already get activated if the project is in sfdx shape ?

packages/salesforcedx-vscode-apex/src/index.ts
jkalloor0701: I know sometimes when I used debug to run an extension, I would run into problems with using lsp. I am using this as a backup. If that is not necessary I can get rid of this.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+          if (this.head) {
+            this.head.children.push(newApexTestGroup);
+          }
+        }
+      });
+    }
+    return this.head;
+  }
+
+  private addTests(fileContent: string, apexTestGroup: ApexTestGroup): void {
```
vazexqi: Do we really need this? It seems like this is very error-prone since it's not actually parsing but matching regex. You could be confused if you have a commented portion of code like

```
// @IsTest
```

or

```
/**
* @IsTest
**/

File: packages/salesforcedx-vscode-apex/src/views/testOutline.ts [hunk 24]
```diff
@@ -0,0 +1,666 @@
...
+        channelService.appendLine(
+          '-----------------------------------------------------------'
+        );
+        channelService.showChannelOutput();
+      }
+    }
+
+    if (test.file) {
+      vscode.window.showTextDocument(test.file).then(() => {
+        this.eventsEmitter.emit('Show Highlight', position, isRow);
```
vazexqi: @jkalloor0701 - All messages should be externalized.
@ruthemmanuelle - Comments?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+        const apexTest = new ApexTest(name, apexTestGroup.file, ind);
+        apexTest.name = apexTestGroup.label + '.' + name;
+        this.apexTestMap.set(apexTest.name, apexTest);
+        apexTestGroup.children.push(apexTest);
+      }
+      match = regEx.exec(fileContent);
+    }
+  }
+
+  private deleteFolderRecursive(path: string) {
```
vazexqi: What is this for? What are you deleting?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  public readJSONFile(folderName: string) {
+    if (vscode.workspace.rootPath) {
+      const fullFolderName = ospath.resolve(
+        vscode.workspace.rootPath,
+        folderName
+      );
+      const jsonSummary = this.getJSONFileOutput(folderName);
+      this.UpdateTestsFromJSON(jsonSummary);
+      this._onDidChangeTreeData.fire();
+      this.eventsEmitter.emit('Delete Folder', fullFolderName);
```
vazexqi: @jkalloor0701 - All messages should be externalized.
@ruthemmanuelle - Comments?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: I output results to a temporary output directory, and then I go and delete it. This is so that we can still output the human output to the console. @cwallsfdc mentioned that this would be a good option to use

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+        }
+      }
+    }
+    groups.forEach(group => {
+      group.updatePassFailLabel();
+      group.description = this.summarize(jsonSummary.summary, group);
+    });
+  }
+
+  private summarize(summary: TestSummary, group: ApexTestGroup): string {
```
vazexqi: Move this into its class in a different file to make it more easily testable.
In general I'm not a fan of just using string concatenation.
Is there a different approach that we can use?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  }
+}
+
+export abstract class Test extends vscode.TreeItem {
+  public children = new Array<Test>();
+  public description: string;
+  public name: string;
+
+  constructor(
+    public label: string,
```
vazexqi: Don't use this method of declaring fields.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  }
+
+  public abstract contextValue: string;
+}
+
+export class ApexTestGroup extends Test {
+  public passing: number = 0;
+
+  constructor(
+    public label: string,
```
vazexqi: Don't use this style of declaring fields.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+  }
+}
+
+export class ApexTest extends Test {
+  public errorMessage: string = '';
+  public stackTrace: string = '';
+  public passed = false;
+
+  constructor(
+    public label: string,
```
vazexqi: Don't use this style of declaring fields.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+
+export type FullTestResult = {
+  summary: TestSummary;
+  tests: TestResult[];
+};
+
+export type TestSummary = {
+  outcome: string;
+  testsRan: number;
+  passing: number;
```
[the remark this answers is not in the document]
jkalloor0701: That's just what it is called in the actual json output, so I left that as that

packages/salesforcedx-vscode-apex/package.json
vazexqi: Could you send the link to that resource so I can look at the license and copyright?
This might seem pedantic but we need to be very clear on this to avoid legal issues.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -12,23 +12,84 @@ import {
...
+    const handle = languageClient.start();
+    context.subscriptions.push(handle);
+
+    languageClient.onReady().then(() => {
+      languageClientReady = true;
+    });
+  }
+
+  // Test View
+  const rootPath = vscode.workspace.rootPath || '';
```
lcampos: Use `vscode.workspace.workspaceFolders[0]` instead of rootPath since it's deprecated (https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs)

packages/salesforcedx-vscode-apex/tsconfig.json
vazexqi: Got it. And this is formatted using prettier?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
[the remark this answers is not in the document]
vazexqi: OK. Maybe we should move all these files that are just mapping the JSON into a separate file called TestDataAccessObjects.ts so that it's clearer that this is just mapping to the JSON since you have no control over that.

The JSON is "stable" right? And won't change?

packages/salesforcedx-vscode-apex/src/index.ts
vazexqi: Could you elaborate more on the issue with the LSP? 
I think it would be nice to not have to use this manual approach.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
vazexqi: Is there no utility that we can use to delete this short of writing our own recursive delete? Maybe you can search for `shelljs` and see how we use it there.

vazexqi: Also, if you are using the OS temp folders, aren't those deleted automatically for us?

vazexqi:
@jkalloor0701 - Have you tested this on Windows?

packages/salesforcedx-vscode-apex/src/index.ts
lcampos: Look at how commands are registered on activate for salesforcedx-vscode-apex-replay-debugger and salesforcedx-vscode-core extensions.

packages/salesforcedx-vscode-apex/package.json
jkalloor0701: https://github.com/maty21/mocha-sidebar

packages/salesforcedx-vscode-apex/tsconfig.json
jkalloor0701: Yeah, I ran node scripts/reformatwithprettier.js so it should be formatted with that

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -49,9 +52,89 @@
...
+    "onView:sfdx.force.test.view"
+  ],
   "main": "./out/src",
   "contributes": {
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "test",
+          "title": "Test Viewer",
```
allileong: We keep labels in our package.json files in corresponding package.nls.json files so that we can more easily apply localization. You can look at line 140 of this file as an example. Could you please extract the user facing strings in this file to the package.nls.json file?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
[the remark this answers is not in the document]
jkalloor0701: It shouldn't change unless there is a change to the actual sfdx force:apex cli commands themselves

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
```diff
@@ -0,0 +1,666 @@
...
+      this.apexTestOutline.readJSONFile(this.outputToJson);
+    });
+
+    channelService.streamCommandOutput(execution);
+
+    if (this.showChannelOutput) {
+      channelService.showChannelOutput();
+    }
+
+    await ProgressNotification.show(execution, cancellationTokenSource);
```
allileong: In a recent change, I removed the `await` on ProgressNotification.show and it should be removed here as well. After removing the `await`, the `async` should also be removed from line 640.

packages/salesforcedx-vscode-apex/src/index.ts
jkalloor0701: When I first launch the extension, isLanguageClientReady() will  return false (probably has to do with the fact that activate is async and and createLanuageServer is called in activate). When that happens, I use the manual parsing approach.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: Yeah, I didn't want to attempt to refine in case we didn't need it, but I outline why we need it as of now above.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: this.eventsEmitter.emit is not a message, it launches the deleteFolder function.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: shell.rm('-rf', path) is not working. I could use the /tmp OS temp folder for Mac and Linux. For windows, the temp folder is stored in the TEMP environment variable. Not sure if that is a good approach either

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: Well, it's the object that contains all the Test Classes and tests, so '<container>'?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
vazexqi: All right, let's tackle the issue holistically here since I might be missing something.

Intent
===
Purpose (please confirm/refute): we need a way to store temporary results so that we can still output human results (since JSON is not human readable results, I assume). 

Questions:
1. How bad is the JSON results from a human-readable perspective?
1. How come we didn't choose to use the TAP format from the test runner? That one seems to be able to stream results through and is also pretty human-readable.


Implementation
===

If we do need to store the results somewhere temporarily, could we use some of the features of the OS to simplify our implementation? Though, some posts like [this](https://superuser.com/questions/296824/when-is-a-windows-users-temp-directory-cleaned-out) seems to suggest that Windows doesn't clean out the temp folders automatically.

1. Does https://nodejs.org/docs/latest-v7.x/api/os.html#os_os_tmpdir not work? Note, I am using Node 7's doc because that is the version of Node that VS Code ships with as of August 2018.
1. "shell.rm('-rf', path) is not working" <-- what is the error that you are seeing?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
vazexqi: ✅ - OK nothing to do here then.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
vazexqi: Is this the rootNode for all child tree items (tests, test classes, etc)?
Maybe `rootNode` or `treeRoot`?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
ruthemmanuelle: @jkalloor0701 - Like other messages, to provide feedback I'd like to know:
1. When users will encounter this message.
2. Where and how the message will be displayed.
And, yes, all messages should be in a messages file. That way we'll be able to translate them if we localize the extension pack, and it's easier for me to keep an eye on any changes to the messages that people make in the future.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: screen shot 2018-08-02 at 12 49 48 pm
The message is shown in the sidebar itself, and the user encounters it when there are no tests in the project

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
jkalloor0701: screen shot 2018-08-02 at 12 57 11 pm
This is shown as a loading bar when the process is run. I can definitely change it to not have the (s), but I did grab this from other run test commands in the extension (just a heads up).

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: Intent:
1. The JSON results are definitely not very readable, and especially if I want to later allow any invocation of these tests to update the test view (such as the code lens button).
2. I followed the example of the other run test examples in the extension.

Implementation:

1. Yes, that works. I will have to do some tests on Windows to see if that folder gets cleaned automatically.

2. There is no elaboration on the error, even if I turn on the developer tools in the extension host, it just does not remove the folder.

vazexqi: "I want to later allow any invocation of these tests to update the test view (such as the code lens button)." <-- this is a good point. Does the code lens button fire off any events now that the test view can list for?

jkalloor0701: Well, it uses the forceApexTestRunCodeActionExecutor, so I was going to utilize that

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
ruthemmanuelle: Got it. Yeah, I think we talked about this decided that following the style guideline for the other run test commands would cause confusion. I think this is OK how it is. Thanks for the screenshot.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
ruthemmanuelle: Thanks, @jkalloor0701 . Would it make sense to give the user a bit more info here? We could add some language like this, which I adapted from a similar situation we encountered in the past.

No Tests Found
Your project doesn't contain any Apex test methods. To run Apex tests, open a project that contains methods with @isTest annotations or the testMethod keyword.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
ruthemmanuelle: @jkalloor0701 - Is this the same situation as the one above?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
ruthemmanuelle: @jkalloor0701 - Just making sure you saw this comment, since you already responded to my other questions.

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: I could add that. It might be a little long for that small amount of space. Can I have that be the description when you hover over?

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: Yes, it is

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
jkalloor0701: Oh whoops, forgot to respond to this one. This is actually not a message but the event emitter launches a different function to run. @ruthemmanuelle

packages/salesforcedx-vscode-apex/src/views/testOutline.ts
ruthemmanuelle: Sure, having the sentence be hover text works. In that case, please change the label to "No Apex Tests Found".

jkalloor0701:
Whoops, some reverts didn't work. Let me just change those

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -24,6 +24,11 @@ export const messages = {
...
+  force_test_view_loading_message: 'Loading Apex Tests ...',
```
ruthemmanuelle: We usually use sentence-case capitalization for progress messages: "Loading Apex tests..."

vazexqi:
Closed in favor of #552

PR #543, opened 2018-08-06
Activate Extension on 'SFDX: Create a Change-Set-Based Project' Command

allileong:
### What does this PR do?

Extensions are activated lazily in VS Code. As a result, we need to
provide the context in which the extension should be activated. Our
extension is usually activated by the presence of the sfdx-project.json
file, but since the 'SFDX: Create a Change-Set-Based Project' command
can be called from outside of an sfdx project, the command should
be added to the list of events that activate the extension.

@W-5231415@


allileong:
@vazexqi No, the command still respects the "when" condition and will only show up if the pref is set to `true`.

PR #545, opened 2018-08-06
Create uri with 'file' scheme for use with 'vscode.openFolder'

allileong:
### What does this PR do?

Creates a uri with the 'file' scheme to be passed to `vscode.openFolder`. Without specifying the scheme, we see the following error:

screen shot 2018-08-06 at 9 42 14 am

@W-5296710@
https://github.com/Microsoft/vscode/issues/55891#issuecomment-410799312


allileong:
@vazexqi @lcampos, I looked into adding testing for this change, but it is difficult to unit test. A system-test also seems to heavy-handed. Is it ok to check this in without tests?

PR #548, opened 2018-08-09
Alli/rename create command

allileong:
### What does this PR do?
Renames `SFDX: Create Change-Set-Based Project` to `SFDX: Create Project with Manifest`

@W-5230413@


allileong:
@ntotten I left the perm `salesforcedx-vscode-core.change_set_based_tools.enabled` as is, even though it still references 'change-set-based' projects. `change_set_based_tools` seems like it still accurately describes the set of commands that are guarded by the perm. What do you think?

ntotten:
@allileong Yes, that is fine to leave. You're reasoning is correct, the tools are still "change set based dev tools".

PR #549, opened 2018-08-09
Add metrics to commands being executed

lcampos:
### What does this PR do?
Adds telemetry for command execution.

@W-5280764@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
@@ -9,11 +9,13 @@ export class Command {
   public readonly command: string;
   public readonly description?: string;
   public readonly args: string[];
+  public readonly logName?: string;
 
   public constructor(builder: CommandBuilder) {
     this.command = builder.command;
     this.description = builder.description;
     this.args = builder.args;
+    this.logName = builder.logName;
   }
 
   public toString(): string {
@@ -31,6 +33,7 @@ export class CommandBuilder {
   public readonly command: string;
   public description?: string;
   public args: string[] = [];
+  public logName?: string;
 
   public constructor(command: string) {
     this.command = command;
@@ -58,6 +61,11 @@ export class CommandBuilder {
   public withJson(): CommandBuilder {
     this.args.push('--json');
     this.args.push('--loglevel', 'fatal');
+    return this;
+  }
+
+  public withLogName(logName: string): CommandBuilder {
+    this.logName = logName;
     return this;
   }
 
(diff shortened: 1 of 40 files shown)


packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts
```diff
@@ -9,11 +9,13 @@ export class Command {
   public readonly command: string;
   public readonly description?: string;
   public readonly args: string[];
+  public readonly logName?: string;
```
lcampos: Expanding the command builder to support adding log name when we build the command.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -338,6 +339,10 @@ export abstract class SfdxCommandletExecutor<T>
     taskViewService.addCommandExecution(execution, cancellationTokenSource);
   }
 
+  public logMetric(logName?: string) {
```
lcampos: New method that will log command metrics if a logName was declared. Will be "automatically" called if `execute` is not overwritten, else we'll have to call it from the new `execute`.

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -68,6 +69,7 @@ export class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
       cwd: vscode.workspace.rootPath
     }).execute(cancellationToken);
     this.attachExecution(execution, cancellationTokenSource, cancellationToken);
+    this.logMetric(execution.command.logName);
```
lcampos: Directly calling logMetric since we are not leveraging `SfdxCommandletExecutor`.

lcampos:
There are certain commands we use that are built & executed on the utilities packages that I'm currently not addressing since I didn't want to add a circular dependency because of where we have the telemetryService. E.g. FauxCommandGenerator executes a couple of commands and it is part of `salesforcedx-sobject-faux-generator` but it is called from forceGenerateFauxClasses, where we do have some level of tracking.

@vazexqi do you have any suggestions for the scenario described above ?

packages/salesforcedx-vscode-core/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -79,7 +79,8 @@ export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{
       .withFlag('--tests', this.test)
       .withFlag('--resultformat', 'human')
       .withArg('--synchronous')
-      .withFlag('--loglevel', 'error');
+      .withFlag('--loglevel', 'error')
+      .withLogName('force_apex_test_run_codeaction');
```
allileong: Should this be 'force_apex_test_run_code_action'?

lcampos: Yeah, let me change that

PR #550, opened 2018-08-09
Add No response bot to manage inactive issues. Update version of pret…

lcampos:
…tier to enable yaml formatting.

@W-5315918@

### What does this PR do?
Adds a configuration file for No Response bot (https://github.com/probot/no-response) usage on the repo.
Also updates prettier to the latest version and enables yaml file formatting.

@W-5315918@


.github/no-response.yml
```diff
@@ -0,0 +1,10 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
```
lcampos: This config means the bot will look at issues with `more information required` label on it and if it's been inactive for 7 days it will close it with the comment listed here.

vazexqi:
Is the VS Code bot from Microsoft available? I recall that one being smarter in not just closing things with more-information-required but also anything that just didn't have any reasonable activity.

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 sudo: required
```
vazexqi: I thought we removed Travis completely. How come we still have a .travis.yml?

lcampos: Not sure but I can remove it as part of this PR.

lcampos:
They use several bots but based on their config I think they might have their own bots. The closest one I found was this and another one called Stale (https://probot.github.io/apps/stale/). Stale looks at both issues & PRs and closes them based on certain configurable params. I thought it was a bit too much based on how many open issues we have that are waiting on further info from the author.

vazexqi:
OK. Let's go with this one first and see if we need to change. I don't have any better recommendations and we can always switch later.

PR #555, opened 2018-08-13
*DO NOT MERGE* Disable source commands when they do not match the default org type

allileong:
### What does this PR do?
Disable source commands when they do not match the default org type

If the default org is a scratch org, expose: 
* push
* pull
* status

If the default org is not a scratch org, expose:
* deploy
* retrieve

If no workspace default is set, use the global default org. (I am not sure if this is the behavior that we want, but it is consistent with how commands like `org open` behave)

W-5070915


packages/salesforcedx-utils-vscode/src/cli/forceOrgList.ts
```diff
@@ -0,0 +1,63 @@
+import { SfdxCommandBuilder } from './commandBuilder';
```
allileong: I added this command in salesforce-utils-vscode instead of salesforcedx-vscode-core because I am not directly exposing the command to users.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,59 @@
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+import {
+  ForceConfigGet,
+  ForceOrgList
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+
+export async function setupWorkspaceOrgType() {
```
allileong: I wasn't sure where to put these methods, so I put them in a new `context` directory. I'm totally open to suggestions on where they really belong.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -38,7 +39,7 @@
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "2.2.38",
-    "@types/node": "^7.0.0",
+    "@types/node": "^8.0.0",
```
allileong: @salesforce/core wouldn't compile without updating @types/node

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+  const defaultUsernameOrAlias = await getDefaultUsernameOrAlias();
+  const defaultUsernameIsSet = typeof defaultUsernameOrAlias !== 'undefined';
+  let isScratchOrg = false;
+  if (defaultUsernameIsSet) {
+    const username = await getUsername(defaultUsernameOrAlias!);
+    try {
+      isScratchOrg = await isAScratchOrg(username);
+    } catch (e) {
+      if (e.name === 'NamedOrgNotFound') {
+        // TODO: what should we do in this case? Expose all source commands by default?
```
allileong: What should we do if a default workspace username is set, but no AuthInfo file is found for the username in the file system? For now, I am exposing all of the source commands in this case.

packages/salesforcedx-vscode-core/package.json
vazexqi: VS Code just updated to the latest electron which contains Node 8. So I think we can change this.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
vazexqi: Yeah, this is the only reasonable path since we have no other info to decide otherwise.
The other extreme alternative is to not expose anything, which is not acceptable.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+        throw e;
+      }
+    }
+  }
+  setDefaultOrgIsScratchOrg(defaultUsernameIsSet && isScratchOrg);
+  setDefaultOrgIsNonScratchOrg(defaultUsernameIsSet && !isScratchOrg);
+}
+
+async function isAScratchOrg(username: string): Promise<boolean> {
+  const authInfo = await AuthInfo.create(username);
```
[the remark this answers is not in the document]
vazexqi: username can be an alias and also a someUser@somedomain.com?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+    return forceConfig.get('defaultusername');
+  }
+}
+
+export function registerDefaultOrgWatcher(context: vscode.ExtensionContext) {
+  if (
+    vscode.workspace.workspaceFolders instanceof Array &&
+    vscode.workspace.workspaceFolders.length > 0
+  ) {
+    const sfdxConfigWatcher = vscode.workspace.createFileSystemWatcher(
```
vazexqi: What happens if this file is not available yet?
AFAIK .sfdx/sfdx-config.json is only available after you have executed one of the auth commands.
Prior to that it doesn't exist.

Can we register a watcher first (when the files don't exist) and when the files are created, the watcher is smart enough to register?

packages/salesforcedx-vscode-core/package.json
vazexqi: @lcampos - Could we upgrade all the @types/node to 8.00? Also, let's not use the ^ since that has caused us issues in the past when someone changed the type in a non-backwards compatible way and caused compile to fail.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+  setDefaultOrgIsScratchOrg(defaultUsernameIsSet && isScratchOrg);
+  setDefaultOrgIsNonScratchOrg(defaultUsernameIsSet && !isScratchOrg);
+}
+
+async function isAScratchOrg(username: string): Promise<boolean> {
+  const authInfo = await AuthInfo.create(username);
+  const authInfoFields = authInfo.getFields();
+  return Promise.resolve(typeof authInfoFields.devHubUsername !== 'undefined');
```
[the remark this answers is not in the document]
vazexqi: Could this be `null` or `undefined`? Should we check for both?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Returns the non-aliased username
+ * @param usernameOrAlias
+ */
+async function getUsername(usernameOrAlias: string): Promise<string> {
```
vazexqi: How come this one gets a special jsdoc comment?

allileong: I wasn't sure if it was obvious why I needed to get the username from a username. Would it be better if I removed this, or added jsdoc to all the methods for consistency?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,93 @@
...
+  const username = await Aliases.fetch(usernameOrAlias);
+  if (username) {
+    return Promise.resolve(username);
+  }
+  return Promise.resolve(usernameOrAlias);
+}
+
+function setDefaultOrgIsScratchOrg(val: boolean) {
```
vazexqi: Might it be better to track based on source tracking availability vs whether it's a scratch org?
That might make this  more future proof to when we actually add source tracking to other non-scratch orgs.
So, the action item here might just be to name things like _setDefaultUsernameHasChangeTracking_.
I think it's called `defaultusername` right in .sfdx/sfdx-config.json?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "categories": ["Other"],
   "dependencies": {
+    "@salesforce/core": "^0.18.2",
```
vazexqi: @DatGuyJonathan, @lcampos - This is an important change since it fulfills one of my architectural goals on making the VS Code commands faster – use the in-memory version of @salesforce/core instead of delegating to the CLI whenever possible. This can potentially shave 1.5 seconds from invocations.

What Alli is doing can be the stepping stone toward using this approach later.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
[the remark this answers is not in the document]
allileong: `username` here cannot not be an alias, or else we will get an `ENOENT` error. Line 12 of workspaceOrgType.ts ensures that what we have is not an alias, but looking at it now, should that be moved into this method?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
allileong: Yea, the onDidCreate() listener picks up on create events and I manually verified that it works

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
allileong: Yep, that makes sense.

packages/salesforcedx-vscode-core/package.json
lcampos: @vazexqi sure, I'll create a separate PR for that.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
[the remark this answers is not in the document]
allileong: The authInfo objects for non-scratch orgs do not have a field called `devHubUsername`, so checking for `undefined` should be sufficient.

packages/salesforcedx-vscode-core/package.json
allileong: @vazexqi Just curious, where does the 1.5 second statistic come from?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
allileong: Your suggestion to test on windows was a good call. I was previously using `vscode.RelativePattern()` to create the pattern passed to `vscode.workspace.createFileSystemWatcher` and that was not working on Windows. The new implementation works on Mac and Windows.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -338,14 +342,22 @@ function registerCommands(
...
-  isvAuthWatcher.onDidCreate(uri => setupGlobalDefaultUserIsvAuth());
-  isvAuthWatcher.onDidDelete(uri => setupGlobalDefaultUserIsvAuth());
-  return vscode.Disposable.from(isvAuthWatcher);
+function registerIsvAuthWatcher(context: vscode.ExtensionContext) {
+  if (
+    vscode.workspace.workspaceFolders instanceof Array &&
+    vscode.workspace.workspaceFolders.length > 0
+  ) {
+    const configPath = path.join(
+      vscode.workspace.workspaceFolders[0].uri.fsPath,
```
allileong: In my testing, I found that we need to pass the full workspace path to `vscode.workspace.createFileSystemWatcher`,rather than the relative path from the workspace dir to the config file

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+async function getUsername(usernameOrAlias: string): Promise<string> {
+  const username = await Aliases.fetch(usernameOrAlias);
+  if (username) {
+    return Promise.resolve(username);
+  }
+  return Promise.resolve(usernameOrAlias);
+}
+
+function setDefaultUsernameHasChangeTracking(val: boolean) {
```
allileong: Renamed these methods and their corresponding context variables to `setDefaultUsernameHasChangeTracking` and `setDefaultUsernameHasNoChangeTracking`.

allileong:
Hi @lcampos and @vazexqi , this pr is ready for official code review.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -0,0 +1,103 @@
...
+  const defaultUsernameOrAlias = await getDefaultUsernameOrAlias();
+  const defaultUsernameIsSet = typeof defaultUsernameOrAlias !== 'undefined';
+  let isScratchOrg = false;
+  if (defaultUsernameIsSet) {
+    const username = await getUsername(defaultUsernameOrAlias!);
+    try {
+      isScratchOrg = await isAScratchOrg(username);
+    } catch (e) {
+      if (e.name === 'NamedOrgNotFound') {
+        // If the info for a default username cannot be found,
```
lcampos: I'm guessing this will happen when the cli doesn't have any linked scratch orgs/devhub ?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
allileong: Not with the current way that I'm using it. `sfdx:default_username_has_change_tracking` is `true` only if 2 conditions are satisfied: 1) a default username is set, and 2) the org that the username is set to has change tracking. It'll return `false` if no default username is set, or the default username is set and the org it is set to doesn't having change tracking. 

Because of this, I can't use `!sfdx:default_username_has_change_tracking` to expose the change-set-based commands because that expression could return `true` if no default username is set

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
allileong: Yeah, this will happen if no local auth info file can be found for the username specified as the defaultusername.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
lcampos: And this is because we want to show both change set and non change set commands in certain situations, right ?

allileong: Yep, there are times when we want to show both and times when we want to hide both too.

PR #561, opened 2018-08-14
Update format configuration for prettier

lcampos:
### What does this PR do?
Updates the version of prettier to 1.6.0 to address reformat diffs.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "printWidth": 80,
+  "tabWidth": 2,
+  "singleQuote": true
+}
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "husky": "0.14.3",
     "istanbul": "1.1.0-alpha.1",
     "lerna": "2.4.0",
-    "prettier": "1.14.2",
+    "prettier": "1.6.0",
     "remap-istanbul": "^0.9.5",
     "shx": "0.2.2",
     "snyk": "^1.82.1",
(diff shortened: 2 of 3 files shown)


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
lcampos: Adding this config file so we can leverage the same options on the scripts & code editors

lcampos:
I'm thinking I'll have to make this change in release branch as well to avoid any weird merges with the publishing process.

package.json
```diff
@@ -7,7 +7,7 @@
     "husky": "0.14.3",
     "istanbul": "1.1.0-alpha.1",
     "lerna": "2.4.0",
-    "prettier": "1.14.2",
+    "prettier": "1.6.0",
```
vazexqi: @lcampos - How come we had to downgrade prettier?

I thought you wanted the newer version to be able to use it with yaml?

vazexqi: NVM. I saw your explanation on Slack:

`@nchen it looks like the newest version of prettier (1.14.2) does not wrap lines the same way version 1.6.0 used to. I'll revert to the previous version (1.6.0) since we want to keep the current format of our files`

vazexqi:
> I'm thinking I'll have to make this change in release branch as well to avoid any weird merges with the publishing process.

@lcampos - I think you're right. You want to do this in both the release and also the develop branch so that everything is in sync.

Issue #562, opened 2018-08-15
Use `extensionPack` instead of `extensionDependencies` property for defining pack


sandy081:
From release v1.26, defining an Extension Pack now uses a new property called `extensionPack` instead of `extensionDependencies` in package.json. This is because extensionDependencies is mainly used to define functional dependencies and an Extension Pack should not have any functional dependencies with its bundled extensions and they should be manageable independent of the pack. 

So please use `extensionPack` property for defining the pack.

For more details refer to [Release notes](https://code.visualstudio.com/updates/v1_26#_extension-packs-revisited)

vazexqi:
@sandy081 - Is the new `extensionPack` key only available to v1.26 and above? Or can this also work with older versions? Basically do we need to update our package.json to say 

```
"engines": {
  "vscode": "^1.26.0"
},
```

We are asking because our current dependency is on version 1.23 for API functionality and we want to determine if we need to bump everything up to 1.26 to take advantage of this new `extensionPack`.

sandy081:
@vazexqi Good question, yes you have to set your engine > 1.26.0

PR #563, opened 2018-08-15
Use new `extensionPack` key

vazexqi:
### What does this PR do?

See the changes in https://code.visualstudio.com/updates/v1_26#_extension-packs-revisited

#562

### DIFF
--- a/packages/salesforcedx-vscode/package.json
+++ b/packages/salesforcedx-vscode/package.json
@@ -28,7 +28,7 @@
     "vscode:publish": "node ../../scripts/publish-vsix.js"
   },
   "categories": ["Extension Packs"],
-  "extensionDependencies": [
+  "extensionPack": [
     "salesforce.salesforcedx-vscode-apex",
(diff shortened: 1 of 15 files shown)


packages/salesforcedx-vscode/package.json
```diff
@@ -28,7 +28,7 @@
     "vscode:publish": "node ../../scripts/publish-vsix.js"
   },
   "categories": ["Extension Packs"],
-  "extensionDependencies": [
+  "extensionPack": [
```
lcampos: Don't we need to update vscode engine version as well ?

vazexqi: That's a good question. Let me check. I somehow thought it was retroactively applied since this is an extension pack. This is just a key per my understanding.

vazexqi: Confirmed on #562 that we do need to bump up to the latest version.
I think this is OK since our stats indicate that our customers are apt to upgrade to the latest version of VS Code quickly.

PR #566, opened 2018-08-15
Prevent metrics being logged from development environments

lcampos:
### What does this PR do?
Prevents metrics from being logged from development environments

@W-5335810@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
+++ b/packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
@@ -27,9 +27,20 @@ export class TelemetryService {
 
   public initializeService(context: vscode.ExtensionContext): void {
     this.context = context;
-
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -25,11 +25,18 @@ export class TelemetryService {
...
   }
 
-  public initializeService(context: vscode.ExtensionContext): void {
+  public initializeService(
+    context: vscode.ExtensionContext,
+    machineId: string
+  ): void {
     this.context = context;
-
+    const isDevMode = machineId === 'someValue.machineId';
```
lcampos: @vazexqi after looking at our data and testing all the relevant scenarios it looks like your suggestion of using `someValue.machineId` is good enough.

PR #567, opened 2018-08-16
Update node types to version 8.0.0

lcampos:
### What does this PR do?
Updates node types to 8.0.0 now that vscode has updated to a newer electron version (https://code.visualstudio.com/updates/v1_26#_nodejs-update)

Comments on PR #555


vazexqi:
@lcampos - Can we actually use 8.9.3? I just checked and the `process.env` in VS Code echoes out 8.9.3. So to be safer we should probably use that version.

I checked using `npm view @types/node versions | grep 8.9.3` and there is indeed that matching version.

PR #571, opened 2018-08-20
Selectively show apex test view

vazexqi:
### What does this PR do?

The Apex Test View should only show up when you have an SFDX project opened. Without this fix, the Apex Test View shows up for all projects.

@W-5296138@

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -70,7 +70,8 @@
       "test": [
         {
           "id": "sfdx.force.test.view",
-          "name": "%test_view_name%"
+          "name": "%test_view_name%",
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -70,7 +70,8 @@
       "test": [
         {
           "id": "sfdx.force.test.view",
-          "name": "%test_view_name%"
+          "name": "%test_view_name%",
+          "when": "sfdx:project_opened"
```
vazexqi: This was a recently added bug fix from VS Code (see https://github.com/Microsoft/vscode/issues/49145)

The `sfdx:project_opened` comes from the main core extension.

PR #585, opened 2018-08-21
Re-enable system tests

lcampos:
### What does this PR do?
Re-enables system tests and addresses an api issue with spectron & electron 2.0+

@W-5335782@

### DIFF
--- a/packages/system-tests/src/helpers/screenshot.ts
+++ b/packages/system-tests/src/helpers/screenshot.ts
@@ -40,7 +40,11 @@ export class Screenshot {
         : `${this.counter++}.png`
     );
 
-    const image = await this.spectron.app.browserWindow.capturePage();
+    const image = await this.spectron.app.browserWindow
+      .capturePage()
+      .catch(err => {
+        console.log('Error creating test snapshot');
+      });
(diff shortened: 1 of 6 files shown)


packages/system-tests/src/helpers/screenshot.ts
```diff
@@ -40,7 +40,11 @@ export class Screenshot {
         : `${this.counter++}.png`
     );
 
-    const image = await this.spectron.app.browserWindow.capturePage();
+    const image = await this.spectron.app.browserWindow
+      .capturePage()
```
lcampos: `capturePage` is using `toPng` which is not available in electron 2.0+ which we are on after the vscode update. A fix has been merged to spectron (electron/spectron#297) but hasn't been released yet so I'm adding a catch to stop tests from failing.

lcampos:
FYI there is a test failure in `lwc-next` unrelated to system tests, I'll take a look at it too.

packages/system-tests/src/helpers/screenshot.ts
vazexqi: Sounds reasonable.

Issue #587, opened 2018-08-21
Null pointer exceptions from Apex Language Service


keirbowden:
### Summary

Started up VS Code with the pre-release version of the Salesforce CLI to try out the new manifest deployment commands.

### Steps To Reproduce:

1.  Checked out my source from version control
1.  Created a manifest sfdx project
1. Converted the source code to SalesforceDX source format
1. Opened VS Code

### Expected result

Apex Language Service starts okay and features such as Go to definition work.

### Actual result
Many null pointer exceptions and features not working. Here's one example:

13:00:25.613 [Apex Prelude Service STARTING] INFO  a.j.l.i.s.ApexPreludeManagedService - Scanning built-in and system Apex types took 2875 ms.
13:00:26.164 [ApexIndexer STARTING] ERROR a.jorje.lsp.impl.index.ApexIndexer - exception thrown in traversing codeunit for references
java.lang.NullPointerException: null
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:212)
	at apex.jorje.semantic.ast.expression.MethodCallExpression.getMethod(MethodCallExpression.java:504)
	at apex.jorje.lsp.impl.index.ApexIndexer$1.visitEnd(ApexIndexer.java:473)
	at apex.jorje.lsp.impl.index.ApexIndexer$1.visitEnd(ApexIndexer.java:268)
	at apex.jorje.semantic.ast.expression.MethodCallExpression.traverse(MethodCallExpression.java:157)
	at apex.jorje.semantic.ast.statement.VariableDeclaration.traverse(VariableDeclaration.java:82)
	at apex.jorje.semantic.ast.statement.VariableDeclarationStatements.traverse(VariableDeclarationStatements.java:88)
	at apex.jorje.semantic.ast.statement.BlockStatement.traverse(BlockStatement.java:141)
	at apex.jorje.semantic.ast.statement.MethodBlockStatement.traverse(MethodBlockStatement.java:76)
	at apex.jorje.semantic.ast.member.Method.traverse(Method.java:104)
	at apex.jorje.semantic.ast.compilation.UserClass.traverse(UserClass.java:128)
	at apex.jorje.lsp.impl.index.ApexIndexer.insertCodeUnit(ApexIndexer.java:268)
	at apex.jorje.lsp.impl.index.ApexIndexer.buildCacheInternal(ApexIndexer.java:196)
[...]

Issue #581, opened 2018-08-21
ApexIndexer throwing NullPointerException


PhilippOrtnerUp2Go:
### Summary

When starting VSCode in a valid folder with a sfdx-config.json I am missing most of the LS features since I can observe an error in the output of the Apex Language Server in the output window.

Other features such as Debugging or the CLI plugin work perfectly fine.

### Steps To Reproduce:

1.  Open sfdx-config.json containing folder (connection and CLI works in this folder).
2.  Observe output window of Apex Language Server. 

### Expected result

Full LS features / no errors.

### Actual result

Multiple of these NullPointerExceptions:

[ApexIndexer STARTING] ERROR a.jorje.lsp.impl.index.ApexIndexer - exception thrown in traversing codeunit for references
java.lang.NullPointerException: null

### Additional information

[apex-ls.log](https://github.com/forcedotcom/salesforcedx-vscode/files/2306031/apex-ls.log)

**VS Code Version**:

1.26.1

**SFDX CLI Version**:

sfdx-cli/6.29.0 (linux-x64) node-v10.6.0

**OS and version**:

Ubuntu 16.04 LTS 
4.15.0-32-generic

PhilippOrtnerUp2Go:
Thanks for the quick reply @vazexqi .

Seems to be similar to issue [#388](https://github.com/forcedotcom/salesforcedx-vscode/issues/388) which was solved.

If you need any other logging info or want me to test something I'll happily help you. 

Will look into opening a support case, need to talk to my supervisors, also not really sure about looking at the code at the moment.

vazexqi:
Duplicate of #587.

PR #598, opened 2018-08-24
Fix NPE in Apex LSP

vazexqi:
### What does this PR do?

Fix a `NullPointerException` in the LSP due to an internal refactoring.

@W-5373295@
#587 
#581


vazexqi:
Closes #587 .

PR #599, opened 2018-08-27
Replay debugger metrics

DatGuyJonathan:
### What does this PR do?
Send metrics for replay debugger when:
- Debugger session is launched
- Size of debug log
- Uploading checkpoints and error message (if any)
- Error message when using heapdump variables (if any)

@W-4872066@


packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -56,4 +56,32 @@ export class TelemetryService {
...
+        errorMessage: errorMsg
+      });
+    }
+  }
+
+  public sendErrorEvent(errorMsg: string): void {
+    if (this.reporter !== undefined && this.isTelemetryEnabled) {
+      this.reporter.sendTelemetryEvent('error', {
+        extensionName: EXTENSION_NAME,
+        errorMessage: errorMsg
```
DatGuyJonathan: @lcampos - Do you have a preference on how you want to log errors in AppInsights? I'm sending an error message in an error event. We can also have a callstack, but it's in JS source.

lcampos: This format is ok since it'll allow us to filter by and build graphs using extensionName, errorMessage, logSize, etc.

packages/salesforcedx-vscode-apex-replay-debugger/src/telemetry/telemetry.ts
```diff
@@ -56,4 +56,32 @@ export class TelemetryService {
       });
     }
   }
+
```
lcampos: It would be good to add a field for error number/code if we are getting one.

DatGuyJonathan: Unfortunately, we don't get an error code. We would have to create our own codes.

errorobj

lcampos: Ok, then let's leave that as future work.

PR #602, opened 2018-08-28
Fix vscode-lightning test failure

lcampos:
### What does this PR do?
Updates tests in vscode-lightning extension that were failing. Root cause of the failure is an increase in the time it takes to activate salesforcedx-vscode-core extension.

### DIFF
--- a/packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
+++ b/packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
@@ -26,14 +26,18 @@ describe('SLDS Deprecated Class Name', () => {
     }
   });
 
-  it('Should create SFDX fix deprecated class command', async () => {
-    if (coreExtension && !coreExtension.isActive) {
-      await coreExtension.activate();
-    }
+  it('coreExtension activation', async () => {
+    await coreExtension.activate();
+    expect(coreExtension.isActive);
+  });
 
-    if (lightningExtension && !lightningExtension.isActive) {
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-lightning/test/deprecatedClassName.test.ts
```diff
@@ -11,20 +11,36 @@ import * as vscode from 'vscode';
...
+
       lightningExtension = vscode.extensions.getExtension(
         'salesforce.salesforcedx-vscode-lightning'
       ) as vscode.Extension<any>;
     }
   });
 
+  it('coreExtension activation', async function() {
+    // tslint:disable-next-line:no-invalid-this
+    this.timeout(0);
```
lcampos: Broke down the original test into multiple tests. I also introduced this line which disables mocha's timeout restriction. This is needed since there's been an increase in the time needed to activate the core extension, which delays the lightning extension's activate.

lcampos: Based on appveyor, core extension activate is taking ~6800ms

allileong: Would it be better to set a timeout that is longer than the core extension activation without disabling the timeout restriction altogether?

lcampos: It could be, that way we can catch if we make a change that significantly delays activation of core. Let me try it.

PR #604, opened 2018-08-28
Add SFDX: Delete Source from Org command

allileong:
### What does this PR do?
Adds a new `SFDX: Delete from Project and Org` command

W-5070968

### DIFF
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -74,9 +74,11 @@
     "Specifies whether the commands for change-set-based development are available to the user.",
   "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor",
   "force_source_retrieve_text": "SFDX: Retrieve Source from Org",
-  "force_source_retrieve_this_source_file_text":
-    "SFDX: Retrieve This Source File from Org",
+  "force_source_retrieve_this_source_text":
+    "SFDX: Retrieve This Source from Org",
   "force_source_deploy_text": "SFDX: Deploy Source to Org",
-  "force_source_deploy_this_source_file_text":
-    "SFDX: Deploy This Source File to Org"
+  "force_source_deploy_this_source_text": "SFDX: Deploy This Source to Org",
+  "force_source_delete_text": "SFDX: Delete from Project and Org",
+  "force_source_delete_this_source_text":
+    "SFDX: Delete this from Project and Org"
 }
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
@@ -66,8 +66,8 @@ export class ManifestChecker implements PreconditionChecker {
 export class ConfirmationAndSourcePathGatherer
   implements ParametersGatherer<{ filePath: string }> {
   private explorerPath: string;
-  private readonly PROCEED = 'Delete source';
-  private readonly CANCEL = 'Cancel';
+  private readonly PROCEED = nls.localize('confirm_delete_source_button_text');
+  private readonly CANCEL = nls.localize('cancel_delete_source_button_text');
 
   public constructor(explorerPath: any) {
     this.explorerPath = explorerPath.fsPath;
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as path from 'path';
+
+import {
+  Command,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import {
+  CancelResponse,
+  ContinueResponse,
+  ParametersGatherer,
+  PreconditionChecker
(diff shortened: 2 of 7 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -74,9 +74,11 @@
     "Specifies whether the commands for change-set-based development are available to the user.",
   "force_manifest_editor_show_text": "SFDX: Show Package Manifest Editor",
   "force_source_retrieve_text": "SFDX: Retrieve Source from Org",
-  "force_source_retrieve_this_source_file_text":
-    "SFDX: Retrieve This Source File from Org",
```
allileong: @ruthemmanuelle Previously, if a user right clicked inside of an open file in the text editor context, the command text for deploy and retrieve were "SFDX: Retrieve This Source File from Org" and "SFDX: Deploy This Source File to Org". I realized that this isn't completely accurate in the case where a user right clicks on an open manifest file, since it is not just the manifest file but also all of the metadata entities specified in the manifest that get deployed or retrieved. To account for that, I removed the word `File` from both strings. Does that sound ok? 
screen shot 2018-08-28 at 1 38 10 pm

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -74,9 +74,11 @@
...
+  "force_source_retrieve_this_source_text":
+    "SFDX: Retrieve This Source from Org",
   "force_source_deploy_text": "SFDX: Deploy Source to Org",
-  "force_source_deploy_this_source_file_text":
-    "SFDX: Deploy This Source File to Org"
+  "force_source_deploy_this_source_text": "SFDX: Deploy This Source to Org",
+  "force_source_delete_text": "SFDX: Delete from Project and Org",
```
allileong: @ruthemmanuelle This was the command name defined by the work item. Do you think it should refer specifically to "Source" like the deploy and retrieve command names?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -75,6 +75,11 @@ export const messages = {
 
   force_source_deploy_text: 'SFDX: Deploy Source to Org',
   force_source_retrieve_text: 'SFDX: Retrieve Source from Org',
+  force_source_delete_text: 'SFDX: Delete from Project and Org',
+  force_source_delete_manifest_unsupported_message:
```
allileong: @ruthemmanuelle Unlike the source retrieve and deploy commands, users cannot delete source by right-clicking on a manifest file because the `--manifest` option is not supported by the CLI. This is the message that will be displayed to the user if they do attempt to select the source delete command by right-clicking on a manifest file.
screen shot 2018-08-28 at 1 47 16 pm

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -75,6 +75,11 @@ export const messages = {
...
+    'SFDX: Delete from Project and Org is not supported for manifest files. Select a source file or directory to delete',
+  force_source_delete_confirmation_message:
+    'This operation will delete files on your computer and in your org. Are you sure you want to proceed?',
```
allileong: This text is used to ask the user for confirmation that they want to delete source
screen shot 2018-08-28 at 1 48 50 pm

packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
```diff
@@ -0,0 +1,100 @@
...
+      return true;
+    }
+    return false;
+  }
+}
+
+export class ConfirmationAndSourcePathGatherer
+  implements ParametersGatherer<{ filePath: string }> {
+  private explorerPath: string;
+  private readonly PROCEED = 'Delete source';
```
lcampos: We usually localize the items (buttons) used on showInformationMessage, is there any reason not to do it here ?

packages/salesforcedx-vscode-core/test/commands/forceSourceDelete.test.ts
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: nitpick: update to 2018

allileong: 👍  thanks for catching that

packages/salesforcedx-vscode-core/src/commands/forceSourceDelete.ts
allileong: That makes sense, I'll change it to localize the button text.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Looks good, but the second sentence should end in a period.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -75,6 +75,13 @@ export const messages = {
...
+  confirm_delete_source_button_text: 'Delete source',
```
ruthemmanuelle: "Source" should be capitalized: 'Delete Source'.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
allileong: That sounds great, I'll make that change now.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: I just saw this question. I think this is OK how it is.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Yes, this is good. Sorry I missed your question before.

PR #608, opened 2018-08-29
Ajha/telemetrymsg

AnanyaJha:
### What does this PR do?
Updates the telemetry message to allow user to click on opt out link. 

W-5313636


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -185,6 +185,7 @@ export const messages = {
   REST_API_description: 'Execute the query with REST API',
   tooling_API_description: 'Execute the query with Tooling API',
   telemetry_legal_dialog_message:
-    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to opt out.',
+    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to ' +
+    '[opt out.](https://github.com/forcedotcom/salesforcedx-vscode/wiki/Telemetry)',
```
allileong: Nitpick: could you move the period to outside of the link? 

That way this will match the screenshot from https://salesforce.quip.com/pboBAoxGwycm.

AnanyaJha: Will do!

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -192,6 +192,7 @@ export const messages = {
   REST_API_description: 'Execute the query with REST API',
   tooling_API_description: 'Execute the query with Tooling API',
   telemetry_legal_dialog_message:
-    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to opt out.',
+    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to ' +
+    '[opt out](https://github.com/forcedotcom/salesforcedx-vscode/wiki/Telemetry).',
```
lcampos: The opt out link is the same as `TELEMETRY_OPT_OUT_LINK` used in the `then` method of the `showInformationMessage` in telemetry.ts (~line 96). Could we add a token here (e.g. `(0)`) and then replace it when using this string on telemetry.ts ?

If we need to update that link in the future then we would only have to do it in one place.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -192,6 +192,7 @@ export const messages = {
   REST_API_description: 'Execute the query with REST API',
   tooling_API_description: 'Execute the query with Tooling API',
   telemetry_legal_dialog_message:
-    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to opt out.',
+    'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to ' +
```
lcampos: Why do we have this as two strings ? `'You agree ... Learn how to ' + '[opt out]...'`

AnanyaJha: Good point

PR #611, opened 2018-08-31
Only use our npm for @salesforce packages

vazexqi:
### What does this PR do?

This makes it fetch from the public npm for the public packages, possibly reducing any mirroring issues.

Build issues

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-registry=https://npm.lwcjs.org/
+@salesforce:registry=https://npm.lwcjs.org/


packages/salesforcedx-vscode-lwc-next/package.json
```diff
@@ -25,9 +25,10 @@
   "dependencies": {
     "@salesforce/salesforcedx-utils-vscode": "43.14.0",
     "ajv": "^6.1.1",
-    "eslint": "4.16.0",
-    "eslint-plugin-lwc": "0.4.0",
-    "lwc-language-server": "1.6.6",
+    "eslint-plugin-lwc":
```
vazexqi: @lcampos - npm also has a feature to specify a tarball URL. See https://docs.npmjs.com/files/package.json#urls-as-dependencies

So I used that instead of bundling it directly. I got this url using, e.g., `npm view --registry=https://npm.lwcjs.org/ lwc-language-server@1.6.1`

packages/salesforcedx-vscode-lwc-next/package.json
```diff
@@ -26,8 +26,10 @@
     "@salesforce/salesforcedx-utils-vscode": "43.14.0",
     "ajv": "^6.1.1",
     "eslint": "4.16.0",
-    "eslint-plugin-lwc": "0.4.0",
-    "lwc-language-server": "1.6.6",
+    "eslint-plugin-lwc":
+      "http://npm.lwcjs.org/eslint-plugin-lwc/-/eslint-plugin-lwc-0.4.0/643f5e18247ca8c0a60bf972f4d3795c97e8e6fe.tgz",
```
vazexqi: @rsalvador - Due to possibly, https://github.com/jdxcode/npm-register/issues/129, the npm.lwcjs.org registry is not properly sending back the packages for @babel and @types. Thus, instead of using the package that way, we embed the tar.url.

This was done by getting the dist.tarball using, e.g., `npm view --registry=https://npm.lwcjs.org/ lwc-language-server@1.6.1`. The tar.gz URL option is supported through https://docs.npmjs.com/files/package.json#urls-as-dependencies

packages/salesforcedx-vscode-lwc-next/package.json
```diff
@@ -26,8 +26,10 @@
...
+    "lwc-language-server":
+      "http://npm.lwcjs.org/lwc-language-server/-/lwc-language-server-1.6.1/48af811b08f90d445c0a5185071129d29df73d39.tgz",
```
lcampos: We were using version 1.6.6 and this tarball is 1.6.1, any reason for not using 1.6.6 tarball ?

vazexqi: Good catch. No this is a typo when I was copy-and-pasting. We should use 1.6.6. Let me fix this.

packages/salesforcedx-vscode-lwc-next/package.json
vazexqi: When you are back from the weekend, it might be worthwhile to see if we want to use https://docs.npmjs.com/files/package.json#bundleddependencies to the lwc-language-server so that we can do it without specifying all its dependencies chain. I think this is a consequence of the refactor that was going on. lwc-language-server@1.5.1 didn't need all these other dependencies brought in.

dgautsch:
@vazexqi we found a fix for jdxcode/npm-register#129 in this PR https://github.com/jdxcode/npm-register/pull/130

packages/salesforcedx-vscode-lwc-next/package.json
rsalvador: Ok, will try that when publishing the next lwc-language-server version

Issue #588, opened 2018-08-21
Include force:source:deploy errors in problem view


keirbowden:
**Is your feature request related to a problem? Please describe.**
Using the preview force:source:deploy functionality, any errors encountered sit in the output view rather than the problem view, which means that no files get decorated with error markers.

**Describe the solution you'd like**
This makes sense when there are high level deployment errors (dependent class invalid, for example), but where there are aura linter or apex compilation errors, which include line/column information, it would be really helpful to have these in the problem view - not only to mark up the file but also to be able to click on the problem and jump straight to it. 

**Describe alternatives you've considered**
None that I can think of.

**Additional context**

ChuckJonas:
I might try to pick this one up.  Been looking for a good issue to start contributing with.  I think we just need to create a `createDiagnosticCollection` and parse the errors in `packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts`.

ntotten:
@ChuckJonas That would be great! I think the feature makes sense. The one issue is that I think the errors in problems view should ONLY be issues related to linter or compilation errors. There are other errors that could happen that shouldn't be there (i.e. auth error, connection issue, etc.)  @vazexqi any objections?

vazexqi:
Yes, this would be a great addition. I haven't thought of all the semantics yet but there might be some issue with the JSON results (some of them might not contain the proper file name, line number, or column number). I've not done an exhaustive check of all the errors but those are some of the caveats. When you encounter them, we can fix them as well.

CC: @lcampos

ChuckJonas:
@vazexqi could you point me to some code where I might find the result of the `ForceSourceDeployExecutor` command?  At first glance, I though results would be [returned here](https://github.com/forcedotcom/salesforcedx-vscode/blob/d806cb9f0bbafb772c5d7d9c1edd3ff5001b8a2f/packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts#L49), but that apparently is just kicking the command off.

ChuckJonas:
@ntotten @vazexqi 

Is there a way to silence the update message on `stderr` or move the sfdx `force:source:deploy` compile errors to `stdout`?

I've added `withJson()` to the forceSourceDeploy cmd, and that gives me JSON for compile errors to `stderr` but the update message is going to make it a PITA to parse (although it's kinda unexpected as without the `withJson` they get put in `stdout`).

[...]

Issue #397, opened 2018-04-25
'Invoke Apex Tests' VS error message states that failed to run when tests have actually failed instead


frackham:
### Summary
When running apex tests through the VS code command palette, any failing tests will throw a misleading error message.

### Steps To Reproduce:
1. Run a test class (Palette, Invoke Apex Tests) with at least 1 failing test.
1. Review error message

### Expected result
'Successfully ran with failing tests'  (warning symbol?)

### Actual result
'SFDX: Invoke Apex Tests... failed to run'

Output shows 'sfdx force:apex:test:run --resultformat human --loglevel error ended with exit code 100'

### Additional information
![image](https://user-images.githubusercontent.com/2105204/39237291-f34a88a8-4872-11e8-8fb3-0d242f53952c.png)

**VS Code Version**:
1.22.2
SF Extensions: 42.12.0

**SFDX CLI Version**:
v42.0

**OS and version**:
Windows 10

vazexqi:
@Sweetman - I believe that you also raised this up. Do we already have the mechanisms in place to fix this? Or does this require a larger effort?

Sweetman:
The mechanisms in place can be altered to fix this. Right now we only check for a 0 exit code in the notificationService and the CLI seems to have established 100 for test failures for apex/lightning. We would just be altering the notificationService.reportExecutionStatus to account for the test exit code. Another option is to write a separate reportExecutionStatus method for test commands if we want to separate it out for some reason.

phardakerffdc:
Any update on this? We're also seeing this in our Blueocean (Jenkins) pipelines.

Are there any known workarounds?

PR #613, opened 2018-09-01
Include force:source:deploy errors in problem view

ChuckJonas:
### What does this PR do?
Shows compile errors in problem view (apex only!).  

#588


packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
+      .withJson();
     if (data.type === FileType.Manifest) {
       commandBuilder.withFlag('--manifest', data.filePath);
     } else {
       commandBuilder.withFlag('--sourcepath', data.filePath);
     }
     return commandBuilder.build();
   }
+
+  public execute(response: ContinueResponse<SelectedPath>): void {
```
ChuckJonas: this was a bit weird to me, overriding the execute statements means you pretty much loose all the "sandwich" code... would be nice if it instead returned the `execution` instead of `void` so you could extend without changing the default behavior...

```
let execution = super.execute(response);
execution.processExitSubject.subscribe(async exitCode => {
//....
});
```

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
+        results[err.filePath] = [err];
+      }
+      return results;
+    }, {});
+  }
+
+  private createDiagnosticFromResult(err: Result) {
+    const ln = Number(err.lineNumber) - 1;
+    const col = Number(err.columnNumber) - 1;
+    const range = new vscode.Range(new vscode.Position(ln, col), new vscode.Position(ln, col));
```
ChuckJonas: salesforce only returns a position, not a range

ChuckJonas:
@ntotten  @vazexqi 
I think there is still more to do on this, but I would love a quick review.  I wasn't sure about some of the coding conventions... Tried to follow best what I saw in a few of the other places I looked.  

Also, I couldn't get the test to run in my env

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
-
-export class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
-  SelectedPath
-> {
+import { notificationService, ProgressNotification } from '../notifications';
+import { taskViewService } from '../statuses';
+import { SfdxCommandlet, SfdxCommandletExecutor, SfdxWorkspaceChecker } from './commands';
+import { FileType, ManifestOrSourcePathGatherer, SelectedPath } from './forceSourceRetrieve';
+
+interface Result {
```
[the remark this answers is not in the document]
ChuckJonas: Also, these types are only for apex and I'm not sure they are even complete in that regard...

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
+      (execution.stderrSubject as any) as Observable<Error | undefined>
+    );
+    this.logMetric(execution.command.logName);
+    channelService.showChannelOutput();
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+
+  private getDeployResultData(stdErr: string) {
```
ChuckJonas: will likely need to expand for different meta-data types

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
     }
     return commandBuilder.build();
   }
+
+  public execute(response: ContinueResponse<SelectedPath>): void {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: vscode.workspace.rootPath
```
lcampos: `workspace.rootPath` is deprecated (https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs), you can use `workspace.workspaceFolders[0]` for now since we need to make further changes for the extensions to support multiple workspaces.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
lcampos: CLI output will always be in the format `{ status: 0|1, result: /* command execution result */ }`. I think just doing a JSON.parse (with some error handling) might be enough.

Something I'm still looking into is if all possible errors are posted in the shape of the `Result` & `ForceSourceDeployErrorResult` interfaces you've added. So far they work for apex, custom objects and aura (with some info missing that might be a bug/gap). I'll get back to you once I'm able to confirm this.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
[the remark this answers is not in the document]
lcampos: I think these interfaces and the parsing methods should be moved to the `salesforcedx-utils-vscode` module under `cli` since we can potentially use them for parsing results from other command executions.

As mentioned below, I've tested errors on different metadata error results but will get back to you on what changes might be needed for the interfaces.

lcampos:
These changes also need tests, you can look at `forceSourceDeploy.test.ts` to see how we are currently writing tests.

Thanks for contributing to the project and feel free to ping me if you have questions/comments, looking forward to getting this to a point in which can be merged.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
+// move this elsewhere?
+vscode.workspace.onDidChangeTextDocument((e => {
+  if (ForceSourceDeployExecutor._errorCollection.has(e.document.uri)) {
+    ForceSourceDeployExecutor._errorCollection.delete(e.document.uri);
+  }
+}));
+
+export class ForceSourceDeployExecutor extends SfdxCommandletExecutor<SelectedPath> {
+
+  public static _errorCollection = vscode.languages.createDiagnosticCollection('deploy-errors');
```
vazexqi: Let's not use the `_` in front of variables names.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
vazexqi: Luis is right here. There are some deploy errors that do not conform to the shape. We have to do a best case approximation – what happens if there is no file URI (or no line/column) to associate with those.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -5,37 +5,129 @@
...
+    );
+    this.logMetric(execution.command.logName);
+    channelService.showChannelOutput();
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+
+  private getDeployResultData(stdErr: string) {
+    const stdErrLines = stdErr.split('\n');
```
vazexqi: Is this safe on Windows?

vazexqi:
Also, just as an FYI, the "continuous-integration/appveyor/pr " isn't going to pass for external PRs since the autobuild has some encrypted values that are not accessible to external contributors.

As this PR gets further along, we will work out a plan on what to do for CI – we might just bring the branch back instead of a fork.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
ChuckJonas: had the same thought... I'm not sure.  Regardless I should use `require('os').EOL` instead

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
ChuckJonas: whats your preferred naming convention for statics?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
ChuckJonas: @lcampos @vazexqi  What are your thoughts about the line below (for some reason github won't let me start a new comment)?

There's gotta be a better way of handling the update + other junk before the response json.  IMO, it would be easier if the error results were in stdout... 

How have you guys handled this in other commands?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
ChuckJonas: Awesome, ya looks like it does work with objects deployments as well.

As mentioned below, I'm struggling to find a simple & consistent way to only grab the response JSON from stderr.  Here's an example stderr from a object deploy command.  You can see it has extra junk before the response. 

```
▸    sfdx-cli: update available from 6.27.0 to 6.31.0-32aa4b4408
[xmldom error]	end tag name: type maybe not complete 
@#[line:9,col:15]
{"message":"Deploy failed.","status":1,"stack":"DeployFailed: Deploy failed.\n    at ALMError (/Users/jonas/.local/share/sfdx/plugins/node_modules/salesforce-alm/dist/lib/almError.js:44:19)\n    at MetadataRegistry.initializeMetadataTypeInfos.then.catch.e (/Users/jonas/.local/share/sfdx/plugins/node_modules/salesforce-alm/dist/lib/source/sourceApiCommand.js:58:31)\n    at tryCatcher (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/util.js:16:23)\n    at Promise._settlePromiseFromHandler (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/promise.js:510:31)\n    at Promise._settlePromise (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/promise.js:567:18)\n    at Promise._settlePromise0 (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/promise.js:612:10)\n    at Promise._settlePromises (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/promise.js:687:18)\n    at Async._drainQueue (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/async.js:138:16)\n    at Async._drainQueues (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/async.js:148:10)\n    at Immediate.Async.drainQueues [as _onImmediate] (/Users/jonas/.local/share/sfdx/plugins/node_modules/bluebird/js/release/async.js:17:14)\n    at runCallback (timers.js:773:18)\n    at tryOnImmediate (timers.js:734:5)\n    at processImmediate [as _immediateCallback] (timers.js:711:5)","name":"DeployFailed","result":[{"columnNumber":"28","lineNumber":"13","error":"Error parsing file: 'asdasd' is not a valid value for the enum 'ActionOverrideType' (13:28)","fullName":"abc__c","type":"CustomObject","filePath":"force-app/main/default/objects/abc__c/abc__c.object-meta.xml"}],"warnings":[]}
```

vazexqi: We usually use `--loglevel fatal` to silence any other information (see https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-utils-vscode/src/cli/commandBuilder.ts#L63)

@amphro - Any thoughts on `[xmldom error]	end tag name: type maybe not complete `? I've not seen that one before.

ChuckJonas: I'm using with `withJson()` which sets the `--loglevel fatal`, but I'm still getting the update and xml parsing junk... is this a bug with the CLI?

amphro: I haven't seen `[xmldom error]	end tag name: type maybe not complete 
@#[line:9,col:15]` but the update available is well known and doesn't respect the `--loglevel fatal` either. We are picking up the story to always have the json response in stdout, which will simplify a lot of this. It should be fixed in the next month with an envvar to enable the new functionality to allow tools to slowly move over. 

Unfortunately, until that happens, there isn't a great workaround short of stipping everything before the first `{:` and hoping that is good enough.

ChuckJonas:
@lcampos 

- moved the error parsing to the utils package
- added unit testing and fixed the unit tests for error parsing
- fixed broken test for SourceDeployCommand
-- I wasn't sure how best to actually test executing the command so I don't have any coverage around the creation of Diagnostics...

ntotten:
Related to #397

lcampos:
I'm closing this PR since these changes got merged as part of PR #717.

PR #615, opened 2018-09-04
initial PR for collections and heap dumps

JimSuplizio:
### What does this PR do?
This PR changes how we deal with non-primitive variables when parsing heap dumps. I've simplified the processing which now creates a bunch of leaf references and pieces variables together from those. The formal tests have not been written for this feature yet, this PR is to review the design and to allow the branch to be pulled for the dreamforce replay-debugger demo. 

@W-5068084@


JimSuplizio:
The NPE is fixed with the second commit. My apologies. The detection for whether or not something is a reference is a bit rudimentary and, unfortunately, has to be. Basically, when processing the variables if something was flagged as a reference while creating the leaf references but isn't in the reference map when we're processing the variables then treat it as a string, set the variable's value to that and clear out the ref field from the variable.

DatGuyJonathan:
Thanks. I'm running into an issue with expanding a collection. Could you try these repro steps?

1. Clone https://github.com/forcedotcom/sfdx-dreamhouse
2. Checkout the branch nick/trailheadx-debuggers
3. Auth to your Dev Hub
4. Create a default scratch org
5. Run `scripts/init.sh`
6. Turn on logging
7. Create a checkpoint on https://github.com/forcedotcom/sfdx-dreamhouse/blob/nick/trailheadx-debuggers/force-app/main/default/classes/HandlerFindProperties.cls#L55
8. Open scratch org
9. Open Dreamhouse app
10. At the bottom left corner, open Dreambot
11. Enter `find house`
12. Enter `Boston`
13. Select `Single Family`
14. Enter `100000`
15. Enter `1000000`
16. Get latest debug log
17. Launch debugger session with that log file
18. Run to the checkpoint
19. Expand a local variable called `properties`

It's not showing the list of custom objects and it removes the static and global scopes. Here's a .gif  starting from step 9.

expandlist

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -146,10 +146,7 @@ export class HeapDumpService {
               // the value to a string which seems to work just fine for everything
               // except boolean values. If the boolean values aren't toString'd then
               // they can cause issues in the variable's window.
```
DatGuyJonathan: Can you elaborate on these issues?

JimSuplizio: I removed the comment but I can comment on the issues. This was the 'weirdness' I'd seen in variables where sometimes things didn't show up correctly. The variable.value is a string and for apparently typescript allows assigning true/false to that string which resulted in the variable sometimes showing up, sometimes not.

DatGuyJonathan:
I think these are the remaining tasks:
- [ ] Fix issue where some strings aren't in quotes
- [ ] Tests
- [ ] Try large collections. IIRC, the Apex Interactive Debugger can paginate.

JimSuplizio:
Fixed the issues with the quotes and added tests. The pagination of large collections is going to get a separate bug. That's going to require some digging into the live debugger to see the when/how. I don't see the changes being particularly hard to do once the exact conditions and manner of pagination are known.

vazexqi:
We have created a separate bug for pagination (it's not a blocker but a nice to have)

1. Pagination is often used as a form of lazy loading to make things load faster. However, since we load *all* the heapdumps upfront anyway, there is no performance gains from pagination.
1. Pagination is also used as a form of reducing needless information. However, since you can only see the heapdump at one instance (the top most frame) and only for that line, this issue seems less useful since you are browsing just at that line.

PR #617, opened 2018-09-05
Cleaning up Apex Execute

muenzpraeger:
### What does this PR do?

This PR fixes the wrong error output when running the Execute Anonymous Apex command.

W-5347904

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -6,7 +6,6 @@
  */
 
 import {
-  CliCommandExecutor,
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -38,29 +33,6 @@ class ForceApexExecuteExecutor extends SfdxCommandletExecutor<{}> {
       .withLogName('force_apex_execute')
       .build();
   }
-
-  public execute(response: ContinueResponse<TempFile>): void {
```
vazexqi: @muenzpraeger - Was this just dead code then?

vazexqi: Gotcha. I think that's actually useful to persist for debugging and, perhaps, an "Execute last command" that can persist throughout VS Code sessions.

But, let me not get ahead of myself :-) There is actually an official way for extensions to store information that could persist across sessions. See https://code.visualstudio.com/docs/extensionAPI/vscode-api#ExtensionContext if you are interested.

PR #630, opened 2018-09-07
Add Folding regions for Apex Classes

### LINKED ISSUE #508: Add Folding regions for Apex Classes
**Is your feature request related to a problem? Please describe.**
It would be nice if Apex supported Folding regions

**Additional context**
Example: https://code.visualstudio.com/updates/v1_17#_folding-regions

ntotten:
### What does this PR do?
Adds support for code folding regions in Apex Code: https://code.visualstudio.com/updates/v1_17#_folding-regions

#508


packages/salesforcedx-vscode-apex/syntaxes/apex.configuration.json
```diff
@@ -1,62 +1,19 @@
...
+    "lineComment": "//",
+    "blockComment": ["/*", "*/"]
+  },
+  // symbols used as brackets
+  "brackets": [["{", "}"], ["[", "]"], ["(", ")"]],
+  // symbols that are auto closed when typing
+  "autoClosingPairs": [["{", "}"], ["[", "]"], ["(", ")"], ["'", "'"]],
+  // symbols that that can be used to surround a selection
+  "surroundingPairs": [["{", "}"], ["[", "]"], ["(", ")"], ["'", "'"]],
+  "folding": {
```
lcampos: This is cool. Do we need to communicate that this is not Apex language syntax and that it is exclusive to IDE interpretation ?

ntotten: Its probably worth noting in the docs, but it's technically just a comment in apex so it wont cause any issues. The format is:

```
//#region MyRegion
code...
//#endregion
```

lcampos: Right, I just want the user to understand this is syntax supported by the IDE and not the language. So if they attempt to use this on dev console, other apex code editors, etc it won't work.

vazexqi: Since this sets a new standard (even if it's just comments like ApexDoc is), we are checking with the Apex team to get their feedback first.

[Question] Is this a feature of Monaco (desktop and/or web) or VS Code? Asking to find out how portable this is across different implementations of the code editor

[Minor] This is probably too small an issue but regions could have weird impact later when we try to implement refactoring like "Extract Method" where we have to carry the regions as comments. It might lead to weird _unbalanced_ regions if the user places //#region //#endregion wrongly.

ntotten: Assuming just vscode since it uses the language config that’s unique to vscode.

I don’t think the refactoring is a big deal since this isn’t a real code construct, it’s just a UI helper.

praksb: In C# and VB this is supported in the language/compiler so there is some validation support in the compiler for unbalanced regions as well as some validation during refactoring which alters the code regions. I am assuming that we have no plans to do this in Apex compiler and hence going this route. 

I am fine with this solution for Apex in VS code but just a caveat as Nick mentioned that this will mess up the regions in VS code when the user perform certain refactorings like "Extract Method" as we move comments. This will happen even when the user placed the //#region and //#endregion correctly to begin with since the code being refactored may include just one of the comments which may get moved to a different place in the source. 
I do see that other languages like F# and JS/TS have a similar solution in VS Code and I am assuming they also likely suffer from the same drawbacks and their customers are fine with that.

ntotten: I’ll add a wiki article explaining the use and drawbacks of this job st so we ha e it written down.

Issue #573, opened 2018-08-20
lightning:stuff tags not highlighted in .cmp viles


mshanemc:
### Summary

opening a lightning component since last Thursday, none of the the non-html tags are highlighted anymore 

### Steps To Reproduce:

1.  Find a repo that has a lightning component with markup, or create one
2. Open in vscode. 

### Expected result

syntax highlighting on LC.cmp
### Actual result

no highlitghting
### Additional information

_Feel free to attach a screenshot_.
<img width="1676" alt="screen shot 2018-08-16 at 2 25 41 pm" src="https://user-images.githubusercontent.com/4261788/44366046-12f41780-a491-11e8-8baa-acb66a643adf.png">

**VS Code Version**: 1.26.1

**SFDX CLI Version**: sfdx-cli/6.29.0-0a2482692c (darwin-x64) node-v8.9.4

**OS and version**: macos 10.13.6

vazexqi:
@mshanemc - Did you mean to include a _different_ screenshot?

mshanemc:
![screen shot 2018-08-20 at 3 52 18 pm](https://user-images.githubusercontent.com/4261788/44371377-3bd1d800-a4a4-11e8-8ed5-5db2a65d1a8d.png)

yep, sorry about that.

daeschwed:
@mshanemc As a workaround, add the following snippet to your user settings in VS Code. That fixes the highlighting issue.

`"files.associations": { "*.cmp": "visualforce" }`

vazexqi:
#534 might actually be related to this. If they did change the grammar, then it seems that marking `<namespace:tagname>` could cause a spike if the entire line is now errorneous.... 🤔

erocheleau:
@ntotten The fix suggested to force the file association of *.cmp to "visualforce" has worked for me and gotten rid of the red <aura: tags.

https://github.com/forcedotcom/salesforcedx-vscode/issues/573#issuecomment-416990920

ntotten:
@lcampos This might be a good short term fix ^. The visualforce language is using an older version of the HTML grammar & language server from before it was broken for `foo:baz` elements.

erocheleau:
I don't know if VSCode has gotten better at letting you "extend" grammar and languages.
Last time I tried it was not possible so you had to re-do everything.

ntotten:
[...]

PR #637, opened 2018-09-13
Fix lightning component syntax highlighting issues

lcampos:
### What does this PR do?
Forks the previous version of the html syntax file (https://github.com/textmate/html.tmbundle) and includes it as part of the lightning extension.

#573

### DIFF
--- a/packages/salesforcedx-vscode-lightning/package.json
+++ b/packages/salesforcedx-vscode-lightning/package.json
@@ -69,6 +69,13 @@
           ".tokens"
         ]
       }
+    ],
+    "grammars": [
+      {
+        "language": "html",
+        "scopeName": "text.html.basic",
+        "path": "./syntaxes/html.tmLanguage.json"
+      }
     ]
   }
 }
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -69,6 +69,13 @@
           ".tokens"
         ]
       }
+    ],
+    "grammars": [
+      {
+        "language": "html",
+        "scopeName": "text.html.basic",
+        "path": "./syntaxes/html.tmLanguage.json"
```
ntotten: Is this scoped to just the above extensions or will it replace all html?

lcampos: Yes, it only applies to those file extensions.

packages/salesforcedx-vscode-lightning/syntaxes/html.tmLanguage.json
```diff
@@ -0,0 +1,746 @@
+{
```
vazexqi: Is this use completely as-is or did we have to modify some things? Asking to see if there is anything I need to focus on.

lcampos: I modified nothing about it, only took the previous version of it which doesn't contain the rules that check for invalid tags.

PR #640, opened 2018-09-17
Create temp directory for anonymous apex

### LINKED ISSUE #639: Anonymous Apex execution throws error when no Apex classes are in the project
### Summary

The execution of Anonymous Apex fails, if a project doesn't contain any Apex classes.

### Steps To Reproduce:

1. Open a project that doesn't contain any Apex classes (and with that no apex.db in the tools folder)
2. Run Anonymous Apex

### Expected result

Apex should be executed in any case. If the folder .sfdx/tools doesn't exist it should be automatically created.

muenzpraeger:
Fixes #639

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -78,6 +78,10 @@ type TempFile = {
 
 export function writeFileAsync(fileName: string, inputText: string) {
   return new Promise((resolve, reject) => {
+    const toolsDir = path.dirname(fileName);
+    if (!fs.existsSync(toolsDir)) {
+      fs.mkdirSync(toolsDir);
+    }
     fs.writeFile(fileName, inputText, err => {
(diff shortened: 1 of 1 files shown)


muenzpraeger:
As I wrote in the referenced issue it's IMHO because the project where I used the command didn't contain any Apex classes.

Adding a `mode` would totally make sense. What's the mode that's set by the Apex Language Server? I couldn't find it in the project as it's created outside of node.

vazexqi:
@muenzpraeger - I created a new empty SFDX project (`sfdx force:project:create -n simple`) and opened it in VS Code. The Apex language server did launch (even without any existing apex classes) and it created the folder `./sfdx/tools` folder with apex.db.

Could you check the output of Help > Toggle Developer Tools? I think your Apex language server didn't launch and I'd like to see if there is a different underlying bug.

vazexqi:
@muenzpraeger - As for permissions, use 075. That is what the Apex language server users when it creates it from Java side.

PR #646, opened 2018-09-20
Track Apex LSP startup time, errors and apex extension startup time

lcampos:
### What does this PR do?
Adds apex extension activation time, Apex LSP errors and startup time to insights.

@W-4883803@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/index.ts
+++ b/packages/salesforcedx-vscode-apex/src/index.ts
@@ -28,6 +28,7 @@ const sfdxCoreExtension = vscode.extensions.getExtension(
 let languageClient: LanguageClient | undefined;
 
 export async function activate(context: vscode.ExtensionContext) {
+  const extensionHRStart = process.hrtime();
   const rootPath = vscode.workspace.workspaceFolders![0].name;
   const testOutlineProvider = new ApexTestOutlineProvider(rootPath, null);
   // Telemetry
@@ -40,8 +41,7 @@ export async function activate(context: vscode.ExtensionContext) {
     );
   }
 
-  telemetryService.sendExtensionActivationEvent();
-
+  const langClientHRStart = process.hrtime();
   languageClient = await languageServer.createLanguageServer(context);
   LanguageClientUtils.setClientInstance(languageClient);
   const handle = languageClient.start();
@@ -52,9 +52,11 @@ export async function activate(context: vscode.ExtensionContext) {
     .then(async () => {
       LanguageClientUtils.languageClientReady = true;
       await testOutlineProvider.refresh();
+      telemetryService.sendApexLSPActivationEvent(langClientHRStart);
     })
     .catch(err => {
       // Handled by clients
+      telemetryService.sendApexLSPError(err);
     });
 
   context.subscriptions.push(await registerTestView(testOutlineProvider));
@@ -65,6 +67,8 @@ export async function activate(context: vscode.ExtensionContext) {
     isLanguageClientReady,
     getApexTests
   };
+
+  telemetryService.sendExtensionActivationEvent(extensionHRStart);
   return exportedApi;
 }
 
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -65,6 +67,8 @@ export async function activate(context: vscode.ExtensionContext) {
     isLanguageClientReady,
     getApexTests
   };
+
+  telemetryService.sendExtensionActivationEvent(extensionHRStart);
```
lcampos: This times are close (not exact) as the ones shown by running `Developer: Startup Performance` in vscode.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -52,9 +52,11 @@ export async function activate(context: vscode.ExtensionContext) {
     .then(async () => {
       LanguageClientUtils.languageClientReady = true;
       await testOutlineProvider.refresh();
+      telemetryService.sendApexLSPActivationEvent(langClientHRStart);
```
lcampos: Time reported is different, but I think more accurate, than the once reported on Output tab in vscode by the lsp.

packages/salesforcedx-vscode-apex/test/unit/telemetry/telemetry.test.ts
```diff
@@ -77,4 +78,32 @@ describe('Telemetry', () => {
...
+  it('Should send correct data format on sendApexLSPActivationEvent', async () => {
+    const telemetryService = TelemetryService.getInstance();
+    telemetryService.initializeService(reporter, true);
+
+    telemetryService.sendApexLSPActivationEvent([0, 50]);
+    assert.calledOnce(sendEvent);
+
+    const expectedData = {
+      extensionName: 'salesforcedx-vscode-apex',
+      startupTime: match.string
```
lcampos: Had to use matchers (https://sinonjs.org/releases/v2.3.6/matchers/) or the test would never pass.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -52,9 +52,11 @@ export async function activate(context: vscode.ExtensionContext) {
     .then(async () => {
```
allileong: @lcampos I've been meaning to ask this general question. It seems strange that we mix `promise.then()...` with async await. When we are making changes to a method/file where we see both being used, is it ok to refactor the promise chain to async await?

lcampos: It really depends on what you're trying to achieve. Using `await` will cause the execution to pause until the promise is resolved. In this case we chose to wait for the language server to be created (line 45) but we do not want to wait for it to be ready and be done parsing the project, doing that would mean the activation of the extension is delayed.

I'd say that if you need to have the value from the promise to continue executing then rest of the code (synchronous) then you should use await, if not and it's expensive/time consuming then having it being asynchronous is the way to go.

vazexqi: I like that idea. As another data point, we should also see if we are _over_ awaiting and making things serial. If we are, then we should use https://developers.google.com/web/fundamentals/primers/async-functions#careful_avoid_going_too_sequential

packages/salesforcedx-vscode-apex/src/index.ts
vazexqi: I think this time will take into account the _whole_ initialization cycle. The one at the output only considers the startup time from jorje. Also, remember, jorje does some things in an async way (by design so that we don't block), so the time reported might still be different.

For the large code base sample I sent, what is the difference?

lcampos: Apex LSP returns 2 times (Scanning built-in and system Apex types& Scanning user-defined types). I'm comparing against scanning user-defined types. For small/medium size projects the difference in metrics is 1-2 seconds. For the large project I'm getting a significant discrepancy e.g. metric ~34s vs jorje metric ~80s.

vazexqi: The larger measure is also useful to know since that is the total time for Apex jorje to complete. It's not the full time for the user since Apex jorje asyncs some of the work (and doesn't block the UI). But, that number is good to record if possible.

We could use the number you are able to obtain more easily for now (but we need to interpret it with those caveats).

PR #659, opened 2018-10-04
Fix for circular references

JimSuplizio:
### What does this PR do?
There was an issue where circular references weren't being hooked up correctly. This was causing the value of the name/value pair displayed in VS Code to display the wrong thing. Initially I'd opted stop hooking things up when there was a circular reference but this ended leaving a dangling reference. This also caused the value, when it was computed, to be incorrect as well. Right now, with the way it's hooked up it'll mimic the live debugger which is to say that you can just open up the variables until rapture. There is a slight difference in the replay debugger vs live debugger here. The live debugger does something special when it encounters a circular reference and puts it in braces (see picture) instead of keeping it in the same list/variable format as non-recusive references.
livedebuggercircref

The replay-debugger keeps the live debugging format but still identifies the circular dependency. Mainly because the code to special case and add braces would be painful to deal with (the live debugger has a bit more leniency when dealing with variables since it doesn't have to compute the values from heap dump extent entries.
replaydebuggercircref

@W-5491876@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
@@ -134,7 +134,8 @@ export class HeapDumpService {
                 const updatedVariable = this.createVariableFromReference(
                   symbolName,
                   refVar,
-                  new Map<string, null>()
+                  new Map<string, null>(),
+                  new Array<ApexVariableContainer>()
                 );
                 if (updatedVariable) {
                   frameInfo.locals.set(localVar.name, updatedVariable);
@@ -173,7 +174,8 @@ export class HeapDumpService {
                   const updatedVariable = this.createVariableFromReference(
                     symbolName,
                     refVar,
-                    new Map<string, null>()
+                    new Map<string, null>(),
+                    new Array<ApexVariableContainer>()
                   );
                   if (updatedVariable) {
                     statics.set(staticVar.name, updatedVariable);
@@ -210,7 +212,8 @@ export class HeapDumpService {
                       const updatedVariable = this.createVariableFromReference(
                         symName,
                         refVar,
-                        new Map<string, null>()
+                        new Map<string, null>(),
+                        new Array<ApexVariableContainer>()
                       );
                       if (updatedVariable) {
                         frameInfo.globals.set(symName, updatedVariable);
@@ -298,7 +301,8 @@ export class HeapDumpService {
   // cloned into a new variable that has a variable name.
   private copyReferenceContainer(
     refContainer: ApexVariableContainer,
-    varName: string
+    varName: string,
+    createVarRef = true
   ) {
     const tmpContainer = new ApexVariableContainer(
       varName,
@@ -311,9 +315,15 @@ export class HeapDumpService {
     // set so it can be expanded in the variables window. Note this check
     // is kind of superfluous but it's better to be safe than sorry
     if (!this.isPrimitiveType(tmpContainer.type)) {
-      tmpContainer.variablesRef = this.logContext
-        .getVariableHandler()
-        .create(tmpContainer);
+      if (createVarRef) {
+        tmpContainer.variablesRef = this.logContext
+          .getVariableHandler()
+          .create(tmpContainer);
+      } else {
+        // If the variable is being cloned for a name change then
+        // it needs to use the same variablesRef as the parent
+        tmpContainer.variablesRef = refContainer.variablesRef;
+      }
     }
     return tmpContainer;
   }
@@ -327,7 +337,8 @@ export class HeapDumpService {
   //   1: '2'
   //   2: '3'
   private createStringFromVarContainer(
-    varContainer: ApexVariableContainer
+    varContainer: ApexVariableContainer,
+    visitedMap: Map<string, null>
   ): string {
     // If the varContainer isn't a reference or is a string references
     if (
@@ -336,36 +347,55 @@ export class HeapDumpService {
     ) {
       return varContainer.value;
     }
-    // For a list or set the name string is going to end up being (<val1>,...<valX)
-    // For a objects, the name string is going to end up being <TypeName>: {<Name1>=<Value1>,...<NameX>=<ValueX>}
-    const isListOrSet =
-      varContainer.type.toLowerCase().startsWith('list<') ||
-      varContainer.type.toLowerCase().startsWith('set<');
-    // The live debugger, for collections, doesn't include their type in variable name/value
-    const containerType = this.isCollectionType(varContainer.type)
-      ? ''
-      : `${varContainer.type}:`;
-    let returnString = isListOrSet ? '(' : `${containerType}{`;
-    let first = true;
-    // Loop through each of the container's variables to get the name/value combinations, calling to create
-    // the string if another collection is found.
-    for (const entry of Array.from(varContainer.variables.entries())) {
-      const valueAsApexVar = entry[1] as ApexVariableContainer;
-      if (!first) {
-        returnString += ', ';
-      }
-      if (valueAsApexVar.ref) {
-        // if this is also a ref then create the string from that
-        returnString += this.createStringFromVarContainer(valueAsApexVar);
+
+    // If the varContainer is a ref and it's already been visited then return the string
+    if (varContainer.ref) {
+      if (visitedMap.has(varContainer.ref)) {
+        return 'already output';
       } else {
-        // otherwise get the name/value from the variable
-        returnString += isListOrSet
-          ? valueAsApexVar.value
-          : `${valueAsApexVar.name}=${valueAsApexVar.value}`;
+        visitedMap.set(varContainer.ref, null);
       }
-      first = false;
     }
-    returnString += isListOrSet ? ')' : '}';
+    let returnString = '';
+    try {
+      // For a list or set the name string is going to end up being (<val1>,...<valX)
+      // For a objects, the name string is going to end up being <TypeName>: {<Name1>=<Value1>,...<NameX>=<ValueX>}
+      const isListOrSet =
+        varContainer.type.toLowerCase().startsWith('list<') ||
+        varContainer.type.toLowerCase().startsWith('set<');
+      // The live debugger, for collections, doesn't include their type in variable name/value
+      const containerType = this.isCollectionType(varContainer.type)
+        ? ''
+        : `${varContainer.type}:`;
+      returnString = isListOrSet ? '(' : `${containerType}{`;
+      let first = true;
+      // Loop through each of the container's variables to get the name/value combinations, calling to create
+      // the string if another collection is found.
+      for (const entry of Array.from(varContainer.variables.entries())) {
+        const valueAsApexVar = entry[1] as ApexVariableContainer;
+        if (!first) {
+          returnString += ', ';
+        }
+        if (valueAsApexVar.ref) {
+          // if this is also a ref then create the string from that
+          returnString += this.createStringFromVarContainer(
+            valueAsApexVar,
+            visitedMap
+          );
+        } else {
+          // otherwise get the name/value from the variable
+          returnString += isListOrSet
+            ? valueAsApexVar.value
+            : `${valueAsApexVar.name}=${valueAsApexVar.value}`;
+        }
+        first = false;
+      }
+      returnString += isListOrSet ? ')' : '}';
+    } finally {
+      if (varContainer.ref) {
+        visitedMap.delete(varContainer.ref);
+      }
+    }
     return returnString;
   }
 
@@ -577,7 +607,10 @@ export class HeapDumpService {
     }
     // If the reference doesn't contain any child references then it's value can set here.
     if (!hasInnerRefs) {
-      refContainer.value = this.createStringFromVarContainer(refContainer);
+      refContainer.value = this.createStringFromVarContainer(
+        refContainer,
+        new Map<string, null>()
+      );
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -298,7 +301,8 @@ export class HeapDumpService {
   // cloned into a new variable that has a variable name.
   private copyReferenceContainer(
     refContainer: ApexVariableContainer,
-    varName: string
+    varName: string,
+    createVarRef = true
```
JimSuplizio: This was necessary because the variablesRef (which is what tells VS Code that the variable can be expanded) didn't need to be recreated for a variable that's part of the circular reference. The reason cloning was necessary was due to the name changing. For instance local var MyVar of type Foo has field that's a List<Foo> and we add MyVar to its own list. The name of the local variable is MyVar but once added to the list the name is '0' or whatever number in the list it is.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
         ) as ApexVariableContainer;
         expect(updatedNonRefVariable.value).to.be.eq('5');
       });
+
+      it('Should correctly deal with circular references and variable values', () => {
```
vazexqi: Please clean up needless comments.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -597,24 +630,44 @@ export class HeapDumpService {
   public createVariableFromReference(
     varName: string,
     refVariable: ApexVariableContainer,
-    visitedMap: Map<string, null>
+    visitedMap: Map<string, ApexVariableContainer | null>,
+    updateAfterVarCreation: ApexVariableContainer[]
```
JimSuplizio: UpdateAfterVarCreation is used in the finally clause. I ended up getting into a chicken and egg situation here and some variable values cannot be updated until the variable is finished being created. This is only applicable to circular references. createVariableFromReference is a recursive function and because of the visitedMap we when we remove the last entry we know we're done with the initial variable being created. It's at this that time it's safe to update the values of any variables that we couldn't update during creation because we know the variable is complete. Trying to create the value earlier, when the variable was incomplete, would end up creating a value that was also incomplete.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -134,7 +134,8 @@ export class HeapDumpService {
                 const updatedVariable = this.createVariableFromReference(
                   symbolName,
                   refVar,
-                  new Map<string, null>()
+                  new Map<string, null>(),
```
vazexqi: The `replaceFrameVariablesWithHeapDump` is _quite_ long. 
It is worth considering breaking this up into smaller methods, which each method having a more descriptive name than having all the comments.

I don't know if you have time to do this now but, in general, this code type of long method code is not maintainable.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
JimSuplizio: Nothing in that test is a 'needless' comment. Please be specific about which ones you think are needless.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+        topFrame.id = id;
+        frameInfo.locals.set(localRefVariable.name, localRefVariable);
+        // Update it with the heap dump
+        heapDumpService.replaceVariablesWithHeapDump();
+
+        // Verify the variable was updated and that there is a cicular reference that
+        // was set up correctly. The local variable cf1, of type CircularReference contains
+        // a field that's a List<CircularReference>. After creating cf1 we add cf1 to its own
+        // list.
+        const expectedVariableValue =
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+            'someInt'
+          ) as ApexVariableContainer).value
+        ).to.be.eq('5');
+        const listChildVar = updatedLocRefVariable.variables.get(
+          'cfList'
+        ) as ApexVariableContainer;
+
+        // Verify the value is set correctly and there's exactly 1 item
+        // in the list
+        expect(listChildVar.value).to.be.eq(expectedListVarValue);
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+          '(CircularReference:{already output, someInt=5})';
+
+        const updatedLocRefVariable = frameInfo.locals.get(
+          localRefVariable.name
+        ) as ApexVariableContainer;
+        // Verify the variable's value has been set
+        expect(updatedLocRefVariable.value).to.be.eq(expectedVariableValue);
+
+        // There should be 2 children, 1 list named cfList and 1 integer named someInt
+        expect(updatedLocRefVariable.variables.size).to.be.eq(2);
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+        ) as ApexVariableContainer;
+
+        // Verify the value is set correctly and there's exactly 1 item
+        // in the list
+        expect(listChildVar.value).to.be.eq(expectedListVarValue);
+        expect(listChildVar.variables.size).to.be.eq(1);
+
+        // There should be 1 child item on the list and it should have the same
+        // value as cf1, expectedVariableValue.
+        const listElementVar = listChildVar.variables.get(
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+          '0'
+        ) as ApexVariableContainer;
+        expect(listElementVar.value).to.be.eq(expectedVariableValue);
+        // Lastly, verify that it's list has the same value as cfList, expectedListVarValue
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+          '',
+          'CircularReference',
+          '0x717304ef'
+        );
+        const frameInfo = new ApexDebugStackFrameInfo(0, 'Foo');
+        const id = frameHandler.create(frameInfo);
+        topFrame.id = id;
+        frameInfo.locals.set(localRefVariable.name, localRefVariable);
+        // Update it with the heap dump
+        heapDumpService.replaceVariablesWithHeapDump();
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
...
+
+      it('Should correctly deal with circular references and variable values', () => {
+        // Create the heapdump
+        const heapdump = createHeapDumpWithCircularRefs();
+        getHeapDumpForThisLocationStub = sinon
+          .stub(LogContext.prototype, 'getHeapDumpForThisLocation')
+          .returns(heapdump);
+
+        // Create a ref variable to update
+        const localRefVariable = new ApexVariableContainer(
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
```diff
@@ -636,6 +637,71 @@ describe('Replay debugger adapter variable handling - unit', () => {
         ) as ApexVariableContainer;
         expect(updatedNonRefVariable.value).to.be.eq('5');
       });
+
+      it('Should correctly deal with circular references and variable values', () => {
+        // Create the heapdump
+        const heapdump = createHeapDumpWithCircularRefs();
```
vazexqi: Remove this comment.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: I can create a work item to do this but it's not happening right now. There are other bugs that need to get immediately looked at from the Blitz.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
JimSuplizio: I'll remove.

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
JimSuplizio: I'll remove.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -327,7 +337,8 @@ export class HeapDumpService {
   //   1: '2'
   //   2: '3'
   private createStringFromVarContainer(
-    varContainer: ApexVariableContainer
+    varContainer: ApexVariableContainer,
+    visitedMap: Map<string, null>
```
vazexqi: Is there a reason why we use it as a `Map<string, null>`? I think we don't ever use the key so whatever it is fine. But, in that case, why not use a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) since you are just checking for membership?

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
```diff
@@ -336,36 +347,55 @@ export class HeapDumpService {
...
-        returnString += ', ';
-      }
-      if (valueAsApexVar.ref) {
-        // if this is also a ref then create the string from that
-        returnString += this.createStringFromVarContainer(valueAsApexVar);
+
+    // If the varContainer is a ref and it's already been visited then return the string
+    if (varContainer.ref) {
+      if (visitedMap.has(varContainer.ref)) {
+        return 'already output';
```
vazexqi: I am guessing that this is the text matches the interactive debugger?

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
JimSuplizio: Removed.

JimSuplizio:
@vazexqi sure, I may tend to over comment and I admit to that. With simple code doing simple things, sure, it's 'self commenting' but I've had to deal with a lot of code that wasn't simple code doing simple things that had no comments. In the case of dealing heap dumps and extents I'd much rather over comment than under comment, especially when we're dealing with things like
heapdump.extents.extent.value.entry.value.value or heapdump.extents.extent.value.value.value. Also, as you're previously pointed out, there's the curse of knowledge, and while I may know the code but there should be enough comments for someone else to pick up and understand.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: Yes. The pictures in the description illustrate live VS replay which is why I'd used "already output"

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: Habit and slight ignorance. Both Map and Set are close to O(1) in lookup times and I keep seeing Set and thinking **Array**

packages/salesforcedx-apex-replay-debugger/test/unit/adapter/apexReplayDebugVariablesHandling.test.ts
JimSuplizio: Gone.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
vazexqi: Cool, please change it then. Using a `Set`is much clearer rather than a `Map<string, null>` since there is nothing keyed off the value.

DatGuyJonathan:
I'm trying to figure out where in this change it would've ran into the error `RangeError: Maximum call stack size exceeded`. Do you have your SFDX project attached to the bug? The bug's description says it's attached somewhere, but I can't find it.

packages/salesforcedx-apex-replay-debugger/src/core/heapDumpService.ts
JimSuplizio: It's changed with the latest commit.

JimSuplizio:
@DatGuyJonathan the project wasn't added initially because I'm the only that ended up fixing the bug. It's been added now. The Max Stacksize exceeded was due to in an infinite loop in in the heap dump processing, not something coming from the server.

JimSuplizio:
> Okay, my question was about W-5491858 which got duped out to W-5491876. So, if W-5491858 is not fixed yet, you should probably re-open it.

It was duped because the issues were related and I fixed both them with this PR.

PR #663, opened 2018-10-05
Fix log variable processing for object properties and parameters

JimSuplizio:
### What does this PR do?
There was an issue with log variable processing that happened when the object had properties with the same names as method parameters. If a class had myInteger is a property and a method on that class had myInteger as a parameter, the processing would get confused. Take the following log line segment:
VARIABLE_ASSIGNMENT|[6]|this.myInteger|1|0x759bccda
Instead of processing the variable assignment on "this.myInteger" what ended up happening is that the variable myInteger, which is the parameter, would end up getting processed. Because the log line has a reference, which points back to 'this', we ended up with a myInteger that expanded with all of this' properties. 
incorrectvariableassignments

The fix was relatively simple, before pulling the local variable, look to see if processing 'this'.
withfix

The goldfile test actually contains the same inner class code that @lcampos was running when he hit this. As it turns the scenario is in the goldfile since we're no longer needlessly creating the extra variableRefs, those needed to be updated. I'd ended up creating a project with these classes and debugged it locally to verify correctness of these changes. There were quite a few superfluous variableRef's being created that shouldn't have been.  

@W-5497724@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/states/variableAssignmentState.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/states/variableAssignmentState.ts
@@ -47,7 +47,6 @@ export class VariableAssignmentState implements DebugLogState {
         // If the className is 'this' that means the variable being split was
         // this.<something>. We need to check the className for 'this' otherwise
         // a propery on 'this' would get incorrectly processed as a local variable.
-        // It'll correclty fall into the local variable
       } else if (className !== 'this' && frameInfo.locals.has(varName)) {
         map = frameInfo.locals;
         container = map.get(varName) as ApexVariableContainer;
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-apex-replay-debugger/src/states/variableAssignmentState.ts
```diff
@@ -44,7 +44,11 @@ export class VariableAssignmentState implements DebugLogState {
       if (logContext.getStaticVariablesClassMap().has(className)) {
         map = logContext.getStaticVariablesClassMap().get(className)!;
         container = map.get(varName)! as ApexVariableContainer;
-      } else if (frameInfo.locals.has(varName)) {
+        // If the className is 'this' that means the variable being split was
+        // this.<something>. We need to check the className for 'this' otherwise
+        // a propery on 'this' would get incorrectly processed as a local variable.
+        // It'll correclty fall into the local variable
+      } else if (className !== 'this' && frameInfo.locals.has(varName)) {
```
JimSuplizio: To be clear this was tested to levels of indirection as well. this.myInteger with myInteger as a parameter as well as this.myInnerClass.myInteger with myInteger is a parameter.

JimSuplizio:
The second push just trimmed an unnecessary line of comment.

PR #667, opened 2018-10-09
Re-enable system tests

lcampos:
### What does this PR do?
Reenables system tests on appveyor

@W-5511648@

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -41,7 +41,7 @@ test_script:
       "Running npm run test:without-system-tests..."
       "--------------------------------------------"
       npm run test:without-system-tests
-      #npm run coverage:system-tests
+      npm run coverage:system-tests
       npm run aggregateJUnit
       junit-merge -d junit-aggregate -o junit-aggregate.xml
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -41,7 +41,7 @@ test_script:
       "Running npm run test:without-system-tests..."
       "--------------------------------------------"
       npm run test:without-system-tests
-      #npm run coverage:system-tests
+      npm run coverage:system-tests
```
lcampos: I've confirmed that the system tests are running correctly both on appveyor and locally. I believe the issues seen a few days ago were related exclusively to appveyor since 1) I've confirmed vscode didn't make any changed to the DOM & 2) all system tests leverage the same piece of code that looks for `.monaco-workbench` in the DOM and only the first 2 categories of system tests failed consistently

vazexqi: Thanks! Sorry for the confusion. I only saw the `.monaco-workbench` and concluded wrongly that it was a change in the DOM.

Issue #645, opened 2018-09-19
Apex unit test shows blue/red when passing


specimen151:
### Summary

I'm running my unit tests, and they all pass. 100% pass. But two tests show as blue, and the overall icon show up as red. I tried running the tests in the developer console, and they are all green with green checkmark.

I've looked for differences between the tests that get the green circle and the ones that don't, but I can't find any (of course different test).

Even copied one of the green tests (remember ALL tests pass, but at least this one gets a green circle). And it still showed up as blue.

Perhaps there is some cached meta-data somewhere? My project originated as a convert from mdapi, and the test that get the green circle are the ones created before convert to DX. The ones which pass but get blue are written after conversion to DX.

**VS Code Version**:

1.27.2

**SFDX CLI Version**:

6.31.0, node 8.9.4

**OS and version**:

Windows 10

vazexqi:
This might be related to W-5360967, where the Apex Test View doesn't distinguish between unrun (blue) and failing test (red) in some conditions.

specimen151:
I’ll try the refresh and get back with the results. It strange that the
tests that were written before converting to DX (in developer console, then
converted with mdapi) works (green) but even the very same teat code
(copy-paste with another method name) gets the blues.

ons. 19. sep. 2018 kl. 22:39 skrev Nick Chen <notifications@github.com>:

vazexqi:
@lcampos - Per our discussion, our suspect is that it's picking up the wrong json file at the /tmp directory?

lcampos:
@vazexqi yes, we have a way to replicate the issue and logged a bug to track and fix it (W-5463028)
cc @ntotten

venkatpolisetti:
**Is there any way to sort the list of test classes displayed in that sidebar?** 

Currently you cannot search in the tests area sidebar for a specific test class to run. If there is a way to sort them or by default sort them alphabetically so that we know where to scroll to.

Also, **is there a way to show the code coverage on the actual class with red and green highlights just like developer console**? Force.com IDE plugin used to have that feature.

Right now to see the covered lines, I am running the test in developer console to see the covered lines and had to switch back and forth between vscode and dev console.

lcampos:
Hi @venkatpolisetti, thanks for the feedback. Please follow issue #605 for search/sort of tests in the sidebar and #677 for showing code coverage.

I'm going to close this since the fix for the main issue is out.

lcampos:
Fix is out as part of v44.4.0 release.

PR #670, opened 2018-10-10
Ajha/sidebar folder

AnanyaJha:
### What does this PR do?
This pr changes where apex test files are saved. Now, it uses the .sfdx folder structure and saves in a folder named 'apexTests.'

@W-5463028@

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,11 +26,6 @@
     "test": "lerna exec --concurrency 1 --stream --bail=false -- npm run test",
     "test:unit": "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
     "test:integration": "lerna run test:integration --concurrency 1 --stream",
-<<<<<<< HEAD
-    "test:without-system-tests": "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test",
-    "coverage:system-tests": "lerna run coverage --scope system-tests --concurrency 1 --stream",
-    "vscode:package": "lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1 && node scripts/reformat-with-prettier",
-=======
     "test:without-system-tests":
       "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test",
     "test:system-tests":
@@ -39,7 +34,6 @@
       "lerna run coverage --scope system-tests --concurrency 1 --stream",
     "vscode:package":
       "lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1 && node scripts/reformat-with-prettier",
->>>>>>> develop
     "vscode:sha256": "lerna run vscode:sha256 --concurrency 1",
     "vscode:publish": "lerna run vscode:publish --concurrency 1",
     "watch": "lerna run --parallel watch",
--- a/package.json
+++ b/package.json
@@ -16,15 +16,18 @@
     "vsce": "^1.36.3"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
-    "bootstrap": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "postinstall":
+      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "bootstrap":
+      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
     "clean": "lerna run clean",
     "compile": "lerna run --stream compile",
     "lint": "lerna run lint",
     "publish": "node scripts/publish.js",
     "prepush": "npm run lint",
     "test": "lerna exec --concurrency 1 --stream --bail=false -- npm run test",
-    "test:unit": "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
+    "test:unit":
+      "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
     "test:integration": "lerna run test:integration --concurrency 1 --stream",
     "test:without-system-tests":
       "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,18 +16,15 @@
     "vsce": "^1.36.3"
   },
   "scripts": {
-    "postinstall":
```
allileong: @AnanyaJha, could you please run `node scripts/reformat-with-prettier.js` in your salesforcedx-vscode project? This will undo the automatic formatting changes in this file.

package.json
```diff
@@ -40,18 +37,19 @@
...
+    "synk:monitor": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false --org=vazexqi",
     "aggregateJUnit": "node scripts/aggregate-junit-xml.js"
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/forcedotcom/salesforcedx-vscode.git"
   },
-  "license": "BSD-3-Clause"
+  "license": "BSD-3-Clause",
+  "dependencies": {
```
lcampos: We shouldn't need this at the top level package.json

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -174,6 +174,7 @@
     "expand-home-dir": "0.0.3",
     "find-java-home": "0.2.0",
     "path-exists": "3.0.0",
+    "shelljs": "^0.8.1",
```
lcampos: This is making the circle ci build fail. Take a look at the stack trace to see what needs to be changed.

AnanyaJha: Thanks, looking into it now

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -77,7 +79,20 @@ export class ApexTestRunner {
   }
 
   public getTempFolder(): string {
-    return os.tmpdir();
+    if (vscode.workspace && vscode.workspace.workspaceFolders) {
+      const workspaceRootPath = path.join(
```
allileong: Could you rename this variable (maybe to something like 'sfdxDir')?

AnanyaJha: Definitely!

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -174,6 +174,7 @@
     "expand-home-dir": "0.0.3",
     "find-java-home": "0.2.0",
     "path-exists": "3.0.0",
+    "shelljs": "^0.7.8",
```
allileong: @lcampos What is the official process for bringing in new modules? Is there paperwork involved?

allileong: Oops, just realized that it's being used elsewhere.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -77,7 +79,20 @@ export class ApexTestRunner {
...
+        '.sfdx'
+      );
+      const apexTestPath = path.join(sfdxDir, 'apexTests');
+
+      if (!pathExists.sync(apexTestPath)) {
+        mkdir('-p', apexTestPath);
+      }
+      return apexTestPath;
+    } else {
+      throw new Error();
```
allileong: Could you add an error message to this error?

packages/salesforcedx-vscode-apex/package.json
lcampos: @allileong no worries, but for future reference, if we were to add a completely new module we would have to see if it's already part of the list of pre-approved modules (approved by legal, etc). If it's not part of that list then we would have to go through the process of getting it approved. There's an internal link for this that I can add to our internal docs.

lcampos: @AnanyaJha why did you choose a different version than the one used on `salesforcedx-vscode-core` ?

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
AnanyaJha: Yeah, I'll add in a message for the error.

lcampos:
Fix for #645

packages/salesforcedx-vscode-apex/package.json
AnanyaJha: @lcampos I actually had the same version (`"^0.8.1"`) listed earlier, but that was what had caused the circleci build to fail so I switched it back to the current one. In the `node_modules/package.json` for `salesforcedx-vscode-apex` the version listed seemed to be `"^0.7.8"`.

 I also saw that in the package.json for `salesforcedx-vscode-core` it looks like the version in `devDependencies` is `"^0.7.8"` while the version in `dependencies` is `"^0.8.1"`.

lcampos: The problem is that you are listing shelljs twice, once in `dependencies` and again in `devDependencies`. I think the problem in circleci went away cause you now have the same version in both places but you should only have it as part of `dependencies`.

After you sort that out, let's update it to have the same versions as `salesforcedx-vscode-core` for `shelljs` and `@types/shelljs`.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -5,14 +5,17 @@
...
+import { mkdir } from 'shelljs';
 import * as vscode from 'vscode';
+import { nls } from '../messages';
 import { ReadableApexTestRunExecutor } from './readableApexTestRunExecutor';
 import {
   ApexTestNode,
   ApexTestOutlineProvider,
   TestNode
 } from './testOutlineProvider';
+import pathExists = require('path-exists');
```
lcampos: Change this to use `from` instead of `require` and (if possible) just load the function/method you are going to use e.g. sync.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -77,7 +80,20 @@ export class ApexTestRunner {
   }
 
   public getTempFolder(): string {
-    return os.tmpdir();
+    if (vscode.workspace && vscode.workspace.workspaceFolders) {
+      const sfdxDir = path.join(
+        vscode.workspace.workspaceFolders[0].uri.fsPath,
+        '.sfdx'
+      );
+      const apexTestPath = path.join(sfdxDir, 'apexTests');
```
lcampos: I think we should change the path to be `.sfdx/tools/testresults/apex`. The reason for this is 1) I want to follow the current pattern we have for debug logs where we store them under `.sfdx/tools/` and 2) we might need to store test results that are not specific to apex e.g `aura`.

AnanyaJha: Okay sounds good, I'll make that change!

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
AnanyaJha: Okay will do!

PR #681, opened 2018-10-16
Apex refactor & semantic error capabilities

lcampos:
### What does this PR do?
Changes needed to gate Apex Refactor & semantic error capabilities. 

apex_rename

@W-5530830@


packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -135,6 +135,13 @@ function protocol2CodeConverter(value: string) {
 export async function createLanguageServer(
   context: vscode.ExtensionContext
 ): Promise<LanguageClient> {
+  const isInsiders: boolean = /insiders/i.test(vscode.env.appName);
```
lcampos: Decided to add this piece here since we currently have only one feature to gate based on vscode version & feature flag but let me know if we rather move this to it's own class/service.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -141,6 +141,11 @@
           "type": ["string", "null"],
           "default": null,
           "description": "%java_home_description%"
+        },
```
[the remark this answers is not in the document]
vazexqi: Let's go with this for now since we only have rename. Once we have others, we can decide what level of granularity. The only downside is that the name, i.e., `salesforcedx-vscode-apex.enable-refactor` will be kept in their settings.

packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -6,6 +6,6 @@
   "refresh_test_title": "Refresh Tests",
   "run_single_test_title": "Run Single Test",
   "show_error_title": "Display Error",
-  "java_home_description":
-    "Specifies the folder path to the Java 8 runtime used to launch the Apex Language Server (for example, /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home)."
-}
+  "java_home_description": "Specifies the folder path to the Java 8 runtime used to launch the Apex Language Server (for example, /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home).",
+  "apex_refactor_description": "Enable Apex refactor capabilities, currently in beta. This feature flag is exclusive for VSCode; feature is auto-enabled in VSCode Insiders."
```
vazexqi: CC: @ruthemmanuelle ^

packages/salesforcedx-vscode-apex/src/languageServer.ts
vazexqi: I think we don't need the  `: boolean` since it's a primitive. I think tslint might catch this and say it's inferable.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -152,6 +163,12 @@ export async function createLanguageServer(
     }
   };
 
+  if (enableApexRename) {
```
lcampos: Had to add it this way or else the `Rename Symbol` option will always show in the menu regardless of the status of the feature flag.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -35,6 +35,10 @@ async function createServer(
       `${requirementsData.java_home}/bin/java`
     );
     let args: string[];
+    const enableSemanticErrors: boolean = vscode.workspace
```
lcampos: Allow users to enable semantic errors for Apex LSP.

PR #683, opened 2018-10-17
Update test sidebar 

AnanyaJha:
### What does this PR do?

This PR addresses 4 main areas:

1) It ensures that the sidebar reads the results from the correct file for each run.
2) It updates the individual and class icons on the test sidebar with the appropriate colors, depending on the result of the test(s). Also refreshes the sidebar after each run. 
3) Since the exact purpose of the hover text has not been determined yet, it hides the hover text for now.
4) It updates the icons to display orange for a skipped test (if something has changed in the org and not locally, if a test does an http callout, etc.)

@W-5360967@


packages/salesforcedx-vscode-apex/src/views/testDataAccessObjects.ts
```diff
@@ -35,20 +35,21 @@ export class TestSummarizer {
     group: ApexTestGroupNode
   ): string {
     let summString = '';
-    const failing = group.children.length - group.passing;
+    /*
```
lcampos: Remove commented out code.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -196,7 +191,7 @@ export class ApexTestOutlineProvider
     }
     groups.forEach(group => {
       group.updatePassFailLabel();
-      group.description = TestSummarizer.summarize(jsonSummary.summary, group);
+      // group.description = TestSummarizer.summarize(jsonSummary.summary, group);
```
lcampos: Remove commented out code.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -152,19 +155,11 @@ export class ApexTestOutlineProvider
...
-    for (const file of files) {
-      if (
-        file !== 'test-result-codecoverage.json' &&
-        ospath.extname(file) === '.json' &&
-        file.startsWith('test-result')
-      ) {
-        fileName = file;
-      }
-    }
+    const testIdFile = path.join(fullFolderName, 'test-run-id.txt');
```
lcampos: Let's change the name of these variables to testRunIdFile (from testIdFile) & testRunId (from testId). It'll map directly to the name of the file we are working with and would be easier to read the code.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -262,17 +265,22 @@ export class ApexTestGroupNode extends TestNode {
 
   public updatePassFailLabel() {
     this.passing = 0;
+    this.failing = 0;
+    this.skipping = 0;
```
lcampos: What is the purpose of adding `this.skipping` if we never check for the outcome being 'Skip' and increment the count ? How can I test for this scenario ?

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -40,9 +43,9 @@ export class ApexTestOutlineProvider
   private path: string;
   private apexTestInfo: ApexTestMethod[] | null;
 
-  constructor(path: string, apexTestInfo: ApexTestMethod[] | null) {
+  constructor(xpath: string, apexTestInfo: ApexTestMethod[] | null) {
```
lcampos: `xpath` aka `this.path` seems to be exclusively used for `this.getAllApexTests` which really doesn't need that parameter. Please check if `this.path` is used somewhere other than `getAllApexTests` and remove it if it's not needed.

packages/salesforcedx-vscode-apex/resources/dark/testSkip.svg
```diff
@@ -0,0 +1 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 114 114"><defs><style>.cls-1{fill:#ff8b00;}</style></defs><title>Artboard 1 copy</title><path class="cls-1" d="M57,7.5A49.5,49.5,0,1,0,106.5,57,49.5,49.5,0,0,0,57,7.5Zm0,87.42A37.93,37.93,0,1,1,94.93,57,37.93,37.93,0,0,1,57,94.93Z"/><circle class="cls-1" cx="57" cy="57" r="25.88"/></svg>
```
lcampos: Are these testSkip.svg files just copies of similar svg's in the project and you just manually updated their style ?

packages/salesforcedx-vscode-apex/src/views/testDataAccessObjects.ts
```diff
@@ -35,20 +35,21 @@ export class TestSummarizer {
     group: ApexTestGroupNode
```
lcampos: `group` parameter is now unused, let's remove it.

packages/salesforcedx-vscode-apex/src/views/testDataAccessObjects.ts
AnanyaJha: Removing!

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
AnanyaJha: Yup!

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
AnanyaJha: Okay makes sense

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -152,19 +155,11 @@ export class ApexTestOutlineProvider
...
-        file !== 'test-result-codecoverage.json' &&
-        ospath.extname(file) === '.json' &&
-        file.startsWith('test-result')
-      ) {
-        fileName = file;
-      }
-    }
+    const testIdFile = path.join(fullFolderName, 'test-run-id.txt');
+    const testId = fs.readFileSync(testIdFile);
+    let fileName = 'test-result-' + testId + '.json';
```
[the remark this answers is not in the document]
AnanyaJha: Sounds good

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
AnanyaJha: Okay I'll check for that

packages/salesforcedx-vscode-apex/resources/dark/testSkip.svg
AnanyaJha: Yeah, they're copies of the other svg's we already had in the project and I manually updated the colors. Is there a different way I should do it?

packages/salesforcedx-vscode-apex/src/views/testDataAccessObjects.ts
AnanyaJha: Okay

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
AnanyaJha: Ah that was an oversight, I meant to increment `test.skipping` as well. I was wondering if we want to mimic the CLI's behavior here as well: if a test is run individually, it will give the user a message if the test was changed in the org and not locally. If a test is run as part of the whole test class, it does not give the user any message that a specific test name was altered. Instead, it shows that all tests were run. This leads to the status of the local test (that was not renamed) to display as `Outcome: Not Run`.

packages/salesforcedx-vscode-apex/resources/dark/testSkip.svg
lcampos: I thank that's ok

PR #684, opened 2018-10-17
Update changelog for release v44.3.0

lcampos:
### What does this PR do?
Changelog updates for release v44.3.0

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### salesforcedx-vscode-apex
 
-- Store Apex Test sidebar execution results in .sfdx folder ([PR #670](https://github.com/forcedotcom/salesforcedx-vscode/pull/670))
+- We fixed some minor under-the-hood bugs.
 
 ## 44.2.0 - October 13, 2018
 
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 44.3.0 - October 18, 2018
+
+### Fixed
+
+#### salesforcedx-vscode-apex
+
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 44.3.0 - October 18, 2018
+
+### Fixed
+
+#### salesforcedx-vscode-apex
+
+- Store Apex Test sidebar execution results in .sfdx folder ([PR #670](https://github.com/forcedotcom/salesforcedx-vscode/pull/670))
```
vazexqi: Should we rephrase this in from the _why_ perspective instead of the _what/how_ we did it?

I think the end result of this is that results from multiple test runs don't interfere with one another anymore?

lcampos: Well, we broke it up into 2 different changes. The one that takes care of picking the correct result is still under review so we are exclusively releasing having the files saved on .sfdx folder.

vazexqi: Ah, got it. What you have is good then.

ruthemmanuelle: Do we need to mention this, since the more impactful change isn't ready yet? We try to avoid talking about the hidden .sfdx folder. Maybe we can just say that we fixed an under-the-hood bug? Or are people likely to need to access these execution results?

ruthemmanuelle: I'm on the east coast, so I'm not going to be online more tonight, but I'll be on bright and early tomorrow if you want me to look at anything. If you think that people need to know the location of these execution results, this looks ok. I'd put `.sfdx` in tick marks, though, since we format file names and paths as code.

lcampos: Thanks for the feedback @ruthemmanuelle, we'll go with under the hood fixes.

PR #693, opened 2018-10-24
[WIP] Refactor replay debugger initialize

lcampos:
### What does this PR do?
It removes the event that gets information from Apex LSP on replay debugger initializeRequest phase. This is replaced by an internal command that is triggered on resolveDebugConfiguration method (executes before debugger initializeRequest). The new command creates a file that contains all the information needed from Apex LSP and returns the file path. The result (file path) is then sent to launchRequest which parses the data for breakpoint validations and file navigation in later phases of the debugger.

I decided on this approach because the information from Apex LSP is used for breakpoint validation and file navigation once the debugger has started; replacing this would mean a big refactor on our debugger implementation. I also tested using a file watcher that would create a file with this data on any local update to apex code but that showed performance issues when testing with big projects (1000+ apex files).

This is an early review, still need to do clean up and update the tests but I wanted to get your opinions on this approach early on.

@W-5402619@

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
+++ b/packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
@@ -59,6 +59,7 @@ export interface LaunchRequestArguments
   logFile: string;
   stopOnEntry?: boolean | true;
   trace?: boolean | string;
+  __privateData: any;
 }
 
 export class ApexVariable extends Variable {
@@ -222,7 +223,6 @@ export class ApexReplayDebug extends LoggingDebugSession {
     args: DebugProtocol.InitializeRequestArguments
   ): void {
     this.initializedResponse = response;
-    // this.sendEvent(new InitializedEvent());
     this.initializedResponse.body = {
       supportsConfigurationDoneRequest: true,
       supportsCompletionsRequest: false,
@@ -241,14 +241,24 @@ export class ApexReplayDebug extends LoggingDebugSession {
     };
     this.initializedResponse.success = true;
     this.sendResponse(this.initializedResponse);
-    // this.sendEvent(new Event(GET_LINE_BREAKPOINT_INFO_EVENT));
   }
 
   public async launchRequest(
     response: DebugProtocol.LaunchResponse,
     args: LaunchRequestArguments
   ): Promise<void> {
-    this.sendEvent(new Event(GET_LINE_BREAKPOINT_INFO_EVENT));
+    // TODO: need to remove __privateData from args because it's being used later for other purposes.
+    // TODO: also might want to move this somewhere along lines 275 for validation purposes
+    if (args) {
+      const lineBreakpointEventArgs = args.__privateData as LineBreakpointEventArgs;
+      breakpointUtil.createMappingsFromLineBreakpointInfo(
+        lineBreakpointEventArgs.lineBreakpointInfo
+      );
+      this.projectPath = lineBreakpointEventArgs.projectPath;
+    } else {
+      response.message = nls.localize('session_language_server_error_text');
+    }
+
     response.success = false;
     console.log('----------------');
     console.log('launchRequest');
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -149,39 +144,8 @@ function registerDebugHandlers(): vscode.Disposable {
...
-              LINE_BREAKPOINT_INFO_REQUEST,
-              returnArgs
-            );
-            console.log(
-              'in registerDebugHandlers, retrieved line breakpoint info from language server'
-            );
-          }
-        } else if (event.event === SEND_METRIC_LAUNCH_EVENT && event.body) {
+
+        if (event.event === SEND_METRIC_LAUNCH_EVENT && event.body) {
```
lcampos: We can keep these events since their triggered after initializing the debugger.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -71,6 +78,34 @@ export async function activate(context: vscode.ExtensionContext) {
   return exportedApi;
 }
 
+async function updateProjectLineBreakpointFile() {
```
lcampos: The command that calls Apex LSP and creates a json file with the results. Decided to add it to `salesforcedx-vscode-apex` since this will also be needed for `salesforcedx-vscode-apex-debugger`.

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -59,6 +58,8 @@ export interface LaunchRequestArguments
   logFile: string;
   stopOnEntry?: boolean | true;
   trace?: boolean | string;
+  projectBreakpointFilePath?: string;
```
lcampos: Extending this interface to add the file path to the new file containing Apex LSP info & the project path. These used to be included as part of `LineBreakpointEventArgs` but I didn't use that since I would have to port it into `salesforcedx-vscode-apex` to avoid circular dependencies.

packages/salesforcedx-vscode-apex-replay-debugger/src/adapter/debugConfigurationProvider.ts
```diff
@@ -40,13 +40,21 @@ export class DebugConfigurationProvider
     config.name = config.name || nls.localize('config_name_text');
     config.type = config.type || DEBUGGER_TYPE;
     config.request = config.request || DEBUGGER_LAUNCH_TYPE;
+    config.projectBreakpointFilePath = '${command:sfdx.updateBreakpoint}';
```
vazexqi: @lcampos - I think this is they 🔑 idea here. Could you elaborate more on why `${command:sfdx.updateBreakpoint}` is guaranteed to invoke the command? I think it's because of https://code.visualstudio.com/docs/extensions/example-debuggers#_anatomy-of-the-packagejson-of-a-debug-extension

> The variables contribution binds 'variables' to 'commands'. These variables can be used in the launch configuration using the ${command:Xxxx} syntax and the variables are substituted by the value returned from the bound command when a debug session is started.

1. Are there any limits to how many configuration variables can be "bound" by such dynamic commands?
1. Do the "bound" commands execute in order (or in parallel)?
1. What happens if one of the "bound" command fails? Does the rest of the debugger sequence continue trying to invoke other "bound" commands? Or it bails?

vazexqi: Also, if we put this in a config, would it also show up when the user is typing the configuration? I think VS Code will try to provide some auto-completion for you when you are typing out the debugger configuration.

In the grand scheme of things this is not a stopper. But it would be good if we could prevent the user from overriding the `projectBreakpointFilePath` key in the Debug Launch Configuration so that nothing bad will happen.

lcampos: I didn't find much documentation on it so the following is based off of code and comments on threads for the feature.

1. Looking at the code that collects and executes the variables used in the launch config there are no limits to how many commands can be included. The following does apply; the command is only executed once if it's declared multiple times on the same launch configuration, the debug session is canceled if the command errors or is itself canceled (e.g. if the command provides a UI to do so), if the variable can't be mapped to a registered command the debug session execution is canceled. 

Here are some test for multiple commands per configuration https://github.com/Microsoft/vscode/blob/90171dab977e4cea933a01824ed0313d86eccdc8/src/vs/workbench/services/configurationResolver/test/electron-browser/configurationResolverService.test.ts

2. Commands get executed sequentially https://github.com/Microsoft/vscode/blob/2d452e0a7e981a675a59617e296b51de5c40bbb7/src/vs/workbench/services/configurationResolver/electron-browser/configurationResolverService.ts#L109

3. If one of the commands fails then the debug session is also canceled.

Here are the threads on when the feature was added (https://github.com/Microsoft/vscode/issues/6569, https://github.com/Microsoft/vscode/issues/7026). It's been refactored since then but the specs still apply (look at links in answer 2).

lcampos: @vazexqi I haven't test if this new field will be available for auto-completion by the user, I'll get back to you on that (good feedback).

lcampos: I've confirmed that this new property is not available to the end user since it is not defined on the extension's `package.json`. In case they add it to their config it will be labeled as invalid and that field will be ignored when running the replay debugger.

lcampos: @JimSuplizio that seem like a good change. I remember trying to do the async call here but don't exactly remember what issue I ran into. Let me try again and do some testing.

vazexqi: Thanks for the suggestion @JimSuplizio. Keep us posted @lcampos on whether the suggested approach works — or, if it doesn't, why it can't work for our case.

lcampos: I've confirmed that @JimSuplizio suggestion works. The reason I was initially having issues with this was that I was trying to directly convert resolveDebugConfiguration into an async method and typescript was throwing errors. Moving the configuration to an async method that is then resolved on resolveDebugConfiguration's return does the trick :)

I'll be closing this PR and opening a new one with all the changes discussed here.

cc @vazexqi

vazexqi: Excellent!

PR #698, opened 2018-10-25
LWC folder rename

muenzpraeger:
### What does this PR do?

Changes the folder to lwc

### What issue does this PR fix?

@W-5501730@


lcampos:
Do we need to add a validation for the `lwc` keyword on the commands that create apex & visualforce ? There is an existing validation for `lightningcomponents` & `aura`.

muenzpraeger:
For GA, yes. Now it works fine for the limited audience. IMHO

PR #700, opened 2018-10-26
Updated package names and readmes

ntotten:
### What does this PR do?
Updates the extension names and titles in the readmes to be in line with what other extensions on the Marketplace use.

The second goal of this is to help people find the right extension. Currently, we are seeing higher numbers of downloads on some of the individual extension rather than the extension pack. My hope is this clarification in naming will ensure that when users search "Salesforce" they will find the extension pack as the most likely option.

### DIFF
--- a/packages/salesforcedx-vscode-apex-replay-debugger/README.md
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/README.md
@@ -1,33 +1,38 @@
 # Apex Replay Debugger for Visual Studio Code
+
 Apex Replay Debugger simulates a live debugging session using a debug log that is a recording of all interactions in a transaction. You no longer need to parse through thousands of log lines manually. Instead, Apex Replay Debugger presents the logged information similarly to an interactive debugger, so you can debug your Apex code. The debugging process is a repetition of editing your Apex code, pushing or deploying the code to your org, reproducing the buggy scenario, downloading the resulting debug log, and launching Apex Replay Debugger with that debug log.
 
 ## Prerequisites
+
 Before you set up Apex Replay Debugger, make sure that you have these essentials.
 
-* **Salesforce CLI**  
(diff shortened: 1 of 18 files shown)


packages/salesforcedx-vscode-apex-replay-debugger/README.md
```diff
@@ -1,33 +1,40 @@
 # Apex Replay Debugger for Visual Studio Code
+
 Apex Replay Debugger simulates a live debugging session using a debug log that is a recording of all interactions in a transaction. You no longer need to parse through thousands of log lines manually. Instead, Apex Replay Debugger presents the logged information similarly to an interactive debugger, so you can debug your Apex code. The debugging process is a repetition of editing your Apex code, pushing or deploying the code to your org, reproducing the buggy scenario, downloading the resulting debug log, and launching Apex Replay Debugger with that debug log.
 
+**For best results, do not install this extension directly. Install the complete [Salesforce Extension Pack] (https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode) instead.**
```
lcampos: Why do we really need this ? Are extension dependencies not getting properly downloaded ?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "salesforcedx-vscode-core",
-  "displayName": "Salesforce CLI Integration for Visual Studio Code",
+  "displayName": "CLI Integration",
```
lcampos: We should rename it Salesforce CLI Integrations since that's the official name for the cli.

packages/salesforcedx-vscode-apex-replay-debugger/README.md
ntotten: Yeah, its actually something we had already on most of the extension, but it was missing on this one. Based on download numbers some extensions seem to be getting installed more than the extension pack which seems weird. Maybe its the way Microsoft counts downloads, but I want to make sure as I have encountered times when customers were confused because they only added one extension instead of the pack.

PR #711, opened 2018-10-31
Set UserID and SessionID in AppInsights Telemetry 

AnanyaJha:
### What does this PR do?

This pr updates the `vscode-extension-telemetry` to collect data for unique users through `UserID` and `SessionID` and populate AppInsights reports with this information. With this pr, the extension is included as a file in our project until it is possible to merge these changes into the official Microsoft extension.

@W-5313565@

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,33 +16,42 @@
     "vsce": "^1.36.3"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
-    "bootstrap": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "postinstall":
+      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "bootstrap":
+      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
     "clean": "lerna run clean",
     "compile": "lerna run --stream compile",
     "lint": "lerna run lint",
     "publish": "node scripts/publish.js",
     "prepush": "npm run lint",
     "test": "lerna exec --concurrency 1 --stream --bail=false -- npm run test",
-    "test:unit": "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
+    "test:unit":
+      "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
     "test:integration": "lerna run test:integration --concurrency 1 --stream",
-    "test:without-system-tests": "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test",
-    "test:system-tests": "lerna run test --scope system-tests --concurrency 1 --stream",
-    "coverage:system-tests": "lerna run coverage --scope system-tests --concurrency 1 --stream",
-    "vscode:package": "lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1 && node scripts/reformat-with-prettier",
+    "test:without-system-tests":
+      "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test",
+    "test:system-tests":
+      "lerna run test --scope system-tests --concurrency 1 --stream",
+    "coverage:system-tests":
+      "lerna run coverage --scope system-tests --concurrency 1 --stream",
+    "vscode:package":
+      "lerna exec --scope @salesforce/salesforcedx-* -- npm prune --production && lerna run vscode:package --concurrency 1 && node scripts/reformat-with-prettier",
     "vscode:sha256": "lerna run vscode:sha256 --concurrency 1",
     "vscode:publish": "lerna run vscode:publish --concurrency 1",
     "watch": "lerna run --parallel watch",
-    "eslint-check": "eslint --print-config .eslintrc.json | eslint-config-prettier-check",
+    "eslint-check":
+      "eslint --print-config .eslintrc.json | eslint-config-prettier-check",
     "reformat": "node scripts/reformat-with-prettier.js",
-    "snyk:test": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-core/OSSREADME.json
```diff
@@ -0,0 +1,30 @@
...
+      "copies or substantial portions of the Software.",
+      "",
+      "THE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR",
+      "IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,",
+      "FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE",
+      "AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER",
+      "LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,",
+      "OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE."
+    ],
+    "version": "1.26.0",
```
lcampos: This should be `0.1.0`.

package.json
```diff
@@ -16,42 +16,33 @@
...
+    "snyk:test": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
+    "snyk:monitor": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false --org=vazexqi",
     "aggregateJUnit": "node scripts/aggregate-junit-xml.js"
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/forcedotcom/salesforcedx-vscode.git"
   },
-  "license": "BSD-3-Clause"
+  "license": "BSD-3-Clause",
```
lcampos: We don't need to modify this file as part of this changes.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -135,7 +136,7 @@ export class TelemetryService {
 
   public dispose(): void {
     if (this.reporter !== undefined) {
-      this.reporter.dispose();
+      this.reporter.dispose().catch();
```
lcampos: Why do we need this `.catch()` during dispose ?

package.json
AnanyaJha: It keeps automatically updating the formatting for me :confused: Is there a way to stop that from happening for all future changes too? I remember Alli had asked me to run `node scripts/reformat-with-prettier.js` is that something I should be running each time?

JimSuplizio: If you need to, you can create a local git ignore file for this. The number of times I've had to discard changes to this file is nuts.

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,174 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+'use strict';
+
+// process.env['APPLICATION_INSIGHTS_NO_DIAGNOSTIC_CHANNEL'] = true;
```
lcampos: This seems to be there because of how `applicationinsights` uses the diagnostic channel (https://github.com/Microsoft/vscode-extension-telemetry/issues/16). Did you look into this ?

package.json
lcampos: We should do something about this file getting reformatted every time we run `npm install` but this was added because line 47 was added.

packages/salesforcedx-vscode-core/OSSREADME.json
AnanyaJha: Got it!

package.json
AnanyaJha: Ah okay, I see the change you're referring to. Deleting it now, thanks.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
AnanyaJha: Without the `.catch()` I was getting an `UnhandledPromiseRejection` error because this `dispose()` function calls the one declared in `telemetryReporter.ts`. In telemetryReporter's dispose, there is a Promise that expects to be handled, so the `no-floating-promises` linting rule applies here. Alli was helping me figure all this out and we decided to ignore a similar error I was getting in `telemetryReporter.ts` with `// tslint:disable-next-line:no-floating-promises` but kept the `.catch()` here. Is there a better way to handle this scenario, rather than with an empty `catch()`?

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
AnanyaJha: I haven't looked into this, but I will thanks for the link

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
lcampos: I see now. You can change it to `.catch(err => {console.log(err)})` or something similar with which we get some visibility if there's an error. I don't think we can do further handling since this is mainly triggered when the editor is closing or the extension is getting deactivated.

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
AnanyaJha: From what I'm understanding, the original need for that line was as a workaround to ensure that there isn't any monkey patching, but as of `1.0.5` of the `applicationinsights` module this workaround is no longer needed. We're currently using `1.0.6` so I believe we shouldn't need to include this line

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
AnanyaJha: Okay, sounds good I'll do that!

PR #717, opened 2018-11-02
Include deploy errors in problem view (source:deploy)

lcampos:
### What does this PR do?
Takes external PR #613 for showing deploy errors in the problem view and builds on top of it. Code was refactored to be able to reuse it for other commands (e.g. source:push) in future PRs, adds coverage for other deploy error message results and expands test coverage.

#588, @W-5411035@

Scenario 1: Deploy fails but filePath = N/A, we link the errors to the manifest or source path used (e.g. folder).
filepath_not_provided

Scenario 2: Deploy fails and all errors have valid filePath data.
filepath_provided

### DIFF
--- a/packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
+++ b/packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
@@ -12,18 +12,23 @@ import * as vscode from 'vscode';
 export function handleDiagnosticErrors(
   errors: ForceSourceDeployErrorResult,
   workspacePath: string,
-  sourcePath: string
-) {
-  const errorCollection = vscode.languages.createDiagnosticCollection('tmp');
+  sourcePath: string,
+  errorCollection: vscode.DiagnosticCollection
+): vscode.DiagnosticCollection {
+  errorCollection.clear();
   const diagnosticMap: Map<string, vscode.Diagnostic[]> = new Map();
   if (errors.hasOwnProperty('result')) {
     errors.result.forEach(error => {
-      // source:deploys sometimes returns N/A as filePath
+      // source:deploy sometimes returns N/A as filePath
       const fileUri =
         error.filePath === 'N/A'
           ? sourcePath
           : path.join(workspacePath, error.filePath);
-      const range = getRange(error.lineNumber, error.columnNumber);
+      const range = getRange(
+        error.lineNumber || '1',
+        error.columnNumber || '1'
+      );
+
       const diagnostic = {
         message: error.error,
         severity: vscode.DiagnosticSeverity.Error,
--- a/packages/salesforcedx-utils-vscode/src/cli/deployErrorParser.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/deployErrorParser.ts
@@ -1,9 +1,15 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
 export interface DeployError {
-  columnNumber: string;
+  columnNumber?: string;
   error: string;
   filePath: string;
-  fullName: string;
-  lineNumber: string;
+  fullName?: string;
+  lineNumber?: string;
   type: string;
 }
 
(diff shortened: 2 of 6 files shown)


packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export function handleDiagnosticErrors(
+  errors: ForceSourceDeployErrorResult,
+  workspacePath: string,
+  sourcePath: string,
+  errorCollection: vscode.DiagnosticCollection
+): vscode.DiagnosticCollection {
+  errorCollection.clear();
+  const diagnosticMap: Map<string, vscode.Diagnostic[]> = new Map();
+  if (errors.hasOwnProperty('result')) {
```
lcampos: Not all deploy errors have `result` array.

packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+    errors.result.forEach(error => {
+      // source:deploy sometimes returns N/A as filePath
+      const fileUri =
+        error.filePath === 'N/A'
```
lcampos: Not all deploy errors have a valid filePath, which maps to your project, so we do the best we can and map to manifest or sourcePath provided for running the deploy command.

packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+          ? sourcePath
+          : path.join(workspacePath, error.filePath);
+      const range = getRange(
```
lcampos: Same as with filePath, we don't always get location data to build a valid range which is a required field to build Diagnostics.

allileong:
Luis and I spoke about two areas of concern regarding this pr:

1. Now that we are calling the `force:source:deploy` command with the `--json` flag, we do not get the human readable table formatted output for our successfully deployed source. 
screen shot 2018-11-02 at 11 20 18 am

Instead, all of our output is in json format which is more difficult to read:
screen shot 2018-11-02 at 11 22 02 am

The formatting is currently handled by the CLI, so Luis is going to look into whether or not there is some manipulation we can do on our end to produce a more readable output on success. 

2. In some cases, our problem results include 'N/A' for the filepath associated with an error result. In that case, clicking on the link to the problem from the problems view displays an error such as "Unable to open 'force-app': File is a directory."

There is little we can do to make the experience better when the problem result is missing the filepath information. Luis is considering filing a bug on the Metadata API team to include the filepath with the error.

ntotten:
@allileong Is there a way to just not show the JSON output? We have effectively displayed all the useful information in the problems view anyway. In the event of a success, we could just write a simple message "Deploy succeeded" or something.

allileong:
That's a good idea and would definitely be a cleaner output. I'm so used to the output being in the table format that the difference felt jarring, but I think it's also a fair point that people may only care that the deploy succeeded. 

What's the plan to extend the same functionality to `source:push` for scratch orgs? Since we do source tracking there, the command output includes a little extra information about the file status (whether it was Added, Changed, or Deleted). Also, since push acts on all files that have changed, do you think it's more useful in that case to notify people of exactly which files they just pushed since they have less control over that?

@ntotten @lcampos

lcampos:
@allileong specifically for `source:push`, I've so far only looked at displaying errors in a similar way. For the additional info given by that command's output, Diagnostics provides more ways of surfacing data and even tagging them as Info message instead of Errors. Looking at what else we can do specific for that command is the main reason why I chose to address it as a separate PR and will make sure to include looking into all the things you brought up.

For the success output on `source:deploy`, I'll spend some time today to see what we can do and update the thread with what I find.

lcampos:
Update: I've looked at how we can format success messages, we can in fact to something about it but it involves changes in several areas of our code base. We can't leverage the existing services (notificationService, channelService) for this because the execution data is `readonly`. We would have to re-work those services to be able to use them in this scenario or expand them to support it. At the same time, we would have to add code to format the json output into the table output (or other more readable output).

I'll be opening some work items to detail the options around this and we can then prioritize that work without stopping this change from getting merged.

cc @ntotten @allileong

rapsacnz:
The change to make the output more readable would be really nice... currently it's pretty difficult to parse out the actual problem from the push details.

Issue #287, opened 2018-02-02
Improved error handling for invalid sfdx-project.json files


Youpala:
### Summary
The only command showing up when i press F1 is SFDX: Create Project, even though my project has all the necessary files. 

What i tried:
Opening the Dreamforce and Simple projects.
Reinstalling the Salesforce extensions.
Reinstalling VSCode
Reinstalling SFDX CLI
Setting up the scratch org through sfdx.exe.

Manually pulling and pushing through sfdx.exe worked as expected.

### Steps To Reproduce:
Create a project and open it in VSCode.

### Expected result
The full set of SFDX commands available.

### Actual result
Can only access "SFDX: Create Project" command.

### Additional information

![capture](https://user-images.githubusercontent.com/1337479/35734119-445f773c-0820-11e8-8293-1aa340265131.PNG)

**VS Code Version**: Stable 1.19.3

**Java JDK Version**: 8u161

**SFDX CLI Version**: sfdx-cli/6.1.19-4341549 (windows-x64) node-v8.9.4

**OS and version**: Windows 10 (16299.192)

**Internal Work Item**: **[#W-5171263](https://gus.my.salesforce.com/a07B0000005PDd0IAG)**

vazexqi:
@Youpala - Very weird. I see that you have sfdx-project.json available at the top level.

This might be related to https://github.com/forcedotcom/salesforcedx-vscode/wiki/Troubleshooting#extensions-malfunction-after-update 

**BUT** before you do that, could you try the following:

1. What happens if you try to execute the SFDX: Create Project command? I suspect that it actually fails but could you confirm?
1. In the VS Code menu bar, select Help > Toggle Developer Tools, click Console, and search for relevant messages. What errors do you see?

vazexqi:
@Youpala - Any updates on your side? I've not been able to reproduce this issue on my side.

Youpala:
I willl try to look into it this week end but I couldn’t reproduce this issue on my personal computer either. Only my office labtop had this issue. It might be caused by some faulty parameters on that particular computer.

De : Nick Chen<mailto:notifications@github.com>
Envoyé le :jeudi 8 février 2018 17:14
À : forcedotcom/salesforcedx-vscode<mailto:salesforcedx-vscode@noreply.github.com>
Cc : Stanley Horwood<mailto:stanley.horwood@hotmail.fr>; Mention<mailto:mention@noreply.github.com>
[...]

PR #719, opened 2018-11-02
Added JSON schema validation for sfdx-project.json and scratch def files

ntotten:
### What does this PR do?
Adds JSON schema validation for `sfdx-project.json` files and `*-scratch-def.json` files.

@W-4541490@, #287

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -579,6 +579,18 @@
           "description": "%telemetry_enabled_description%"
         }
       }
-    }
+    },
+    "jsonValidation": [
+      {
+        "fileMatch": "sfdx-project.json",
+        "url":
+          "https://forcedotcom.github.io/schemas/sfdx-project.json/sfdx-project.schema.json"
+      },
+      {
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -579,6 +579,18 @@
           "description": "%telemetry_enabled_description%"
         }
       }
-    }
+    },
+    "jsonValidation": [
+      {
+        "fileMatch": "sfdx-project.json",
+        "url":
+          "https://forcedotcom.github.io/schemas/sfdx-project.json/sfdx-project.schema.json"
```
lcampos: I think we should have these schemas as a module that gets added as a dependency or how would this work in an offline scenario ?

ntotten: We could include them, but even MS uses web versions for most of the built-in schemas. I'm pretty sure they cache them for some period. https://github.com/Microsoft/vscode/blob/master/extensions/javascript/package.json#L101-L129

lcampos: Ok, let's go with this and see if we need to include them locally.

Issue #710, opened 2018-10-30
Turn On Apex Debug Log for Replay Debugger not setting correct debug level


matthewjcavalier:
When I run the command 'SFDX: Turn On Apex Debug Log for Replay Debugger' everything looks to be fine. Then I run a single test that fails, run the command 'SFDX: Get Apex Debug Logs...' and select the last log like normal. The issue is though that when the file is downloaded the APEX_CODE log level is DEBUG and the VISUALFORCE log level is INFO, both of which is incorrect for debugging.

### Additional information
**Internal Work Item**: [W-5540032](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005X5ceIAC/view)

lcampos:
@matthewjcavalier I'm closing this issue since the fix is out as part of version 44.8.0 of the extensions. Thanks for reporting this and let us know if there's anything else we can help with.

PR #743, opened 2018-11-13
Pass in null for TraceFlag StartDate

JimSuplizio:
### What does this PR do?
When creating/updating the TraceFlag's StartTime use null instead of the local system time on whatever machine happens to be running VS Code. The problem here is that that local machine and appserver are more than likely not going to be in sync. This means that creating TraceFlag StartTime from the local system time, especially one whose time is ahead of the appserver, would end up causing a delay for when logging started. Instead, passing null for the start time which will tell logging to use the now time on the server side which means the very next command will log appropriately.

@W-5540032@


lcampos:
Fix for #710

PR #744, opened 2018-11-13
Update spectron app handler to pick VSCode window

lcampos:
### What does this PR do?
Updates our Spectron application handler and re-enables system tests in appveyor. The changes address a Spectron Windows OS issue (https://github.com/electron/spectron/issues/60) where multiple empty terminals get opened when running Spectron, causing our tests to fail in appveyor.

@W-5511648@


packages/system-tests/src/spectron/application.ts
```diff
@@ -120,15 +120,28 @@ export class SpectronApplication {
...
-
-      if (/bootstrap\/index\.html/.test(await this.webclient.getUrl())) {
-        break;
+    if (count > 1) {
+      for (let i = 0; i < count; i++) {
+        await this.webclient.windowByIndex(i);
+        const title = await this.webclient.getTitle();
+
+        if (
+          process.platform === 'win32' &&
```
lcampos: NodeJs always returns `win32` regardless of which windows version is being used (https://nodejs.org/docs/latest-v8.x/api/process.html#process_process_platform).

packages/system-tests/src/spectron/application.ts
```diff
@@ -120,15 +120,28 @@ export class SpectronApplication {
...
 
-    // Pick the first workbench window here
+    // Pick the window that loads VSCode,
+    // added code to address Spectron limitation in Windows OS (https://github.com/electron/spectron/issues/60)
     const count = await this.webclient.getWindowCount();
-
-    for (let i = 0; i < count; i++) {
-      await this.webclient.windowByIndex(i);
-
-      if (/bootstrap\/index\.html/.test(await this.webclient.getUrl())) {
```
allileong: This could be a basic question, but why is the window that we want in focus during the Spectron test going to be named "...bootstrap/index.html"? Is that a Spectron convention?

Issue #761, opened 2018-11-19
Issue turning ON Replay Debugger


lflorezcyberark:
I am having a similar issue with version 44.7.0. I am connected to a Sandbox ORG and I am able to deploy from a project locally to the ORG, however when I am triggering the command SFDX: Turn on Apex debug log for Replay Debugger, I get the following error message:

```
{"message":"Provided external ID field does not exist or is not accessible: null","status":1,"stack":"NOT_FOUND: Provided external ID field does not exist or is not accessible: null\n at HttpApi.getError (C:\Users\luis_\AppData\Local\sfdx\client\node_modules\jsforce\lib\http-api.js:250:13)\n at C:\Users\luis_\AppData\Local\sfdx\client\node_modules\jsforce\lib\http-api.js:95:22\n at tryCallOne (C:\Users\luis_\AppData\Local\sfdx\client\node_modules\promise\lib\core.js:37:12)\n at C:\Users\luis_\AppData\Local\sfdx\client\node_modules\promise\lib\core.js:123:15\n at flush (C:\Users\luis_\AppData\Local\sfdx\client\node_modules\asap\raw.js:50:29)\n at _combinedTickCallback (internal/process/next_tick.js:131:7)\n at process._tickCallback (internal/process/next_tick.js:180:9)","name":"NOT_FOUND","warnings":["All --json output, including on error, is moving to stdout. Use the SFDX_JSON_TO_STDOUT environment variable to move over to the new functionality. This will become the default in version 45."]}

SFDX: Turn On Apex Debug Log for Replay Debugger ended with exit code 1
```

I am also able to download a debug log from VS Code but launching from the .log downloaded doesn't do anything.

This is my launch.json:

`{ // Use IntelliSense to learn about possible attributes. // Hover to view descriptions of existing attributes. // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387 "version": "0.2.0", "configurations": [ { "name": "Launch Apex Replay Debugger", "type": "apex-replay", "request": "launch", "logFile": "07L3C000005c8RgUAI_20181119115110.log", "stopOnEntry": true, "trace": true } ] }`

07L3C000005c8RgUAI_20181119115110.log is the downloaded log file.

allileong:
[...]

PR #765, opened 2018-11-20
Show better error message for SFDX: Turn On Apex Debug Log for Replay Debugger when updating a trace flag that is missing a debug level

allileong:
### What does this PR do?
This PR changes the `SFDX: Turn On Apex Debug Log for Replay Debugger` command to throw an error when the query for trace flags returns a trace flag with a null debugLevelId. 

https://github.com/forcedotcom/salesforcedx-vscode/issues/761

`SFDX: Turn On Apex Debug Log for Replay Debugger` first queries TraceFlag for any existing DEVELOPER_LOG type logs. If it finds any existing trace flags, it will take the first one from query results, and it will update that trace flag's associated DebugLevel. 

If the trace flag has no associated DebugLevel, then the command is currently returning the error referenced by the issue above. Since trace flags without Debug Levels are now considered invalid, the fix in this pr will notify the user to delete the invalid trace flags with a helpful error message.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -192,5 +192,8 @@ export const messages = {
   tooling_API_description: 'Execute the query with Tooling API',
   telemetry_legal_dialog_message:
     'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to [opt out](%s).',
-  telemetry_legal_dialog_button_text: 'Read more'
+  telemetry_legal_dialog_button_text: 'Read more',
+
+  invalid_debug_level_id_error:
+    "You have an invalid trace flag in your org. Please delete the trace flag with Id: %s and any others that are missing an associated Debug Level before running this command again. You can find trace flags that are missing debug levels by running the soql query: 'SELECT Id FROM TraceFlag WHERE debuglevelid = null' or visiting the Debug Logs page in your org's Setup UI."
 };
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -195,5 +195,5 @@ export const messages = {
   telemetry_legal_dialog_button_text: 'Read more',
 
   invalid_debug_level_id_error:
-    "You have an invalid trace flag in your org. Please delete the trace flag with Id: %s and any others that are missing an associated Debug Level before running this command again. You can find trace flags that are missing debug levels by running the soql query: 'SELECT Id FROM TraceFlag WHERE debuglevelid = null' or visiting the Debug Logs page in your org's Setup UI."
+    'At least one trace flag in your org doesn\'t have an associated debug level. Before you run this command again, run "sfdx force:data:soql:query -t -q "SELECT Id FROM TraceFlag WHERE DebugLevelId = null"". Then, to delete each invalid trace flag, run "sfdx force:data:record:delete -t -s TraceFlag -i 7tfxxxxxxxxxxxxxxx", replacing 7tfxxxxxxxxxxxxxxx with the ID of each trace flag without a debug level.'
 };
--- a/packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
+++ b/packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
@@ -11,7 +11,7 @@ export class DeveloperLogTraceFlag {
   private static instance: DeveloperLogTraceFlag;
   private active: boolean;
   private traceflagId: string | undefined;
-  private debugLevelId: string | undefined;
+  private debugLevelId: string | undefined | null;
   private startDate: Date;
   private expirationDate: Date;
 
@@ -35,7 +35,7 @@ export class DeveloperLogTraceFlag {
     id: string,
     startDate: string,
     expirationDate: string,
-    debugLevelId: string
+    debugLevelId: string | null
   ) {
     this.traceflagId = id;
     this.startDate = new Date(startDate);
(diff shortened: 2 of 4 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -192,5 +192,8 @@ export const messages = {
   tooling_API_description: 'Execute the query with Tooling API',
   telemetry_legal_dialog_message:
     'You agree that Salesforce Extensions for VS Code may collect usage information, user environment, and crash reports for product improvements. Learn how to [opt out](%s).',
-  telemetry_legal_dialog_button_text: 'Read more'
+  telemetry_legal_dialog_button_text: 'Read more',
+
+  invalid_debug_level_id_error:
+    "You have an invalid trace flag in your org. Please delete the trace flag with Id: %s and any others that are missing an associated Debug Level before running this command again. You can find trace flags that are missing debug levels by running the soql query: 'SELECT Id FROM TraceFlag WHERE debuglevelid = null' or visiting the Debug Logs page in your org's Setup UI."
```
allileong: Hi @ruthemmanuelle , I'd love to get your thoughts on this error message. Basically, trace flags that have no debug level are invalid and should be deleted. In the Setup UI, trace flags that have no debug level would look something like this: 
screen shot 2018-11-20 at 4 10 47 pm

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
```diff
@@ -44,7 +44,7 @@ export class DeveloperLogTraceFlag {
     this.active = true;
   }
 
-  public setDebugLevelId(debugLevelId: string) {
+  public setDebugLevelId(debugLevelId: string | undefined | null) {
```
lcampos: Same as with the change in line 38.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
```diff
@@ -57,6 +57,10 @@ export class DeveloperLogTraceFlag {
     showTraceFlagExpiration(this.getExpirationDate());
   }
 
+  public isValidDebugLevelId() {
+    return this.debugLevelId != null;
```
lcampos: We probably need to also check it's not `undefined` and not an empty string.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
```diff
@@ -35,7 +35,7 @@ export class DeveloperLogTraceFlag {
     id: string,
     startDate: string,
     expirationDate: string,
-    debugLevelId: string
+    debugLevelId: string | null
```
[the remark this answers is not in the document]
allileong: Yea, I changed it so that I could set debugLevelId to `null` in the test.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
allileong: Good point, I'll add a check for the empty string. In JS, `null == undefined`, but `null !== undefined`, so I used the loose equality to check for undefined here. That is not very obvious though, so I can add an explicit check for `this.debugLevelId !== undefined` too.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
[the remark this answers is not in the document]
allileong: I thought about putting the validation inside of the `setTraceFlagDebugLevelInfo()` method so that I wouldn't have to change the type def, but the logic didn't seem like it belonged in the setter.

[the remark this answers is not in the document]
lcampos: I was wrong about this, after taking a closer look at how we use this (`forceStartApexDebugLogging.ts`) we do need this change. Sorry, didn't see that before.

As to the validation, I agree it doesn't fit as part of `setTraceFlagDebugLevelInfo()` so this is good.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
lcampos: No need for changes here, look at comments in line 38.

packages/salesforcedx-vscode-core/src/traceflag/developerLogTraceFlag.ts
lcampos: Awesome, yeah let's make it obvious

packages/salesforcedx-vscode-core/src/messages/i18n.ts
lcampos: I personally would like to have a shorter message (first 2 sentences) and then provide a link to a doc page in case the user doesn't know how to view & delete the traceflags (e.g. https://help.salesforce.com/articleView?id=code_add_users_debug_log.htm)

ruthemmanuelle: I'm reluctant to send the user hunting through the various Setup pages looking for the trace flag (the Debug Logs page, the Apex Classes page, the Apex Triggers page). And, there might not be much point in providing the ID if there might be other trace flags that also need to be deleted. How about something like this?

At least one trace flag in your org doesn't have an associated debug level. Before you run this command again, run "sfdx force:data:soql:query -t -q "SELECT Id FROM TraceFlag WHERE DebugLevelId = null"". Then, to delete each invalid trace flag, run "sfdx force:data:record:delete -t -s TraceFlag -i 7tfxxxxxxxxxxxxxxx", replacing 7tfxxxxxxxxxxxxxxx with the ID of each trace flag without a debug level.

allileong: Thanks! I really like that new message. 

I've updated the code, and this is what it looks like when the user encounters the error:
screen shot 2018-11-21 at 9 53 55 am

PR #770, opened 2018-11-21
Create an autobuild that runs against the insiders' build

AnanyaJha:
### What does this PR do?

This pr updates our AppVeyor configuration and `package.json`s to run tests in VSCode and VSCode Insiders concurrently.

@W-5411499@


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -51,7 +51,11 @@
       "shx rm -rf node_modules && cd out && node ../../../scripts/clean-all-but-jar.js && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "node ../../scripts/run-test-with-top-level-extensions",
-    "test:unit": "node ../../scripts/run-test-with-top-level-extensions"
+    "test:insiders":
+      "cross-env CODE_VERSION=insiders node ../../scripts/run-test-with-top-level-extensions",
+    "test:unit": "node ../../scripts/run-test-with-top-level-extensions",
+    "test:unit-insiders":
```
lcampos: This is missing `cross-env` in the script

packages/system-tests/package.json
```diff
@@ -39,13 +39,17 @@
...
+      "shx rm -rf .vscode-test && shx rm -rf node_modules && shx rm -rf out",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "pretest": "npm run compile && node ../../scripts/download-vscode-for-system-tests",
-    "test": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node out/src/main.js",
-    "coverage": "npm run pretest && node ../../scripts/instrument-salesforcedx-vscode-extensions && cross-env COLLECT_COVERAGE=1 npm run test && node ../../scripts/remap-coverage"
+    "pretest":
+      "npm run compile && node ../../scripts/download-vscode-for-system-tests",
+    "test":
+      "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node out/src/main.js",
+    "test:insiders":
```
lcampos: This is missing `cross-env` as part of the script.

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -53,7 +53,11 @@
       "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "node ../../scripts/run-test-with-top-level-extensions",
-    "test:unit": "node ../../scripts/run-test-with-top-level-extensions"
+    "test:insiders":
```
lcampos: Both `test:insiders` and `test:unit-insiders` run in the same way, why do we need these 2 scripts ? I know you are following the same structure we currently have (`test` & `test:unit`) but I want to make sure we only add the scripts we need.

.appveyor.yml
```diff
@@ -40,7 +43,8 @@ test_script:
       "--------------------------------------------"
       "Running npm run test:without-system-tests..."
       "--------------------------------------------"
-      npm run test:without-system-tests
+  - cmd: npm run %vscode%
```
lcampos: If I understand correctly, this changes allow us to run what was configured in `matrix` in parallel ?

.appveyor.yml
```diff
@@ -40,7 +43,8 @@ test_script:
       "--------------------------------------------"
       "Running npm run test:without-system-tests..."
```
lcampos: Can you update this message to reflect which matrix option is running ?

lcampos:
Some modules are running `cross-env` with version `5.0.4` and others with `^5.0.5`. Do we need to have all modules run in the same version or can we address that as a follow up change ?

.appveyor.yml
AnanyaJha: Yup, will do.

.appveyor.yml
AnanyaJha: Yes, it runs the two different configurations specified in the matrix through `%vscode%

packages/salesforcedx-vscode-apex-debugger/package.json
AnanyaJha: So looking at all of the package.jsons it looks like the `test` and `test:unit` scripts are the exact same for all of them. It seems as though the unit tests are the exact same as the others, so I'm not sure that we need the `test:unit` or `test:unit-insiders` scripts.

lcampos: Ok, for now let's just focus on removing `test:unit-insiders` and log a work item to get rid of `test:unit` in the future.

AnanyaJha:
> Some modules are running `cross-env` with version `5.0.4` and others with `^5.0.5`. Do we need to have all modules run in the same version or can we address that as a follow up change ?

For now, it doesn't look like the two different versions make a difference. I can just update them all to be using the same version right now, otherwise it can be addressed as a follow up change as well.

packages/salesforcedx-vscode-apex-debugger/package.json
AnanyaJha: Sounds good, I'll create a work item for it as well.

.appveyor.yml
AnanyaJha: Actually, I haven't been able to configure it to display different messages based on which matrix option is running. For now, I can update the message to display both the options?

lcampos: Did you try changing `"Running npm run ... "` to `echo Running npm run %vscode%` ? If that doesn't work then we can look at it later.

AnanyaJha: That worked!

Issue #768, opened 2018-11-21
Set the Duration of a Scratch Org


jriordan22849:
**Is your feature request related to a problem? Please describe.**
I don't want to be spinning up a scratch org every 7 days. The default scratch org duration in days is 7. I want to be able to configure this so I do not have to reproduce test data within the org every week.

**Describe the solution you'd like**
A configuration in which I can specify the duration in days for the Scratch Org. When creating a Scratch org In the Salesforce Cli, I can specify the duration in days. I would like to have something similar in vscode. A configuration to set the -d flag(durationdays) and is appended to the sfdx force:org:create.

**Describe alternatives you've considered**

**Additional context**

lcampos:
This was delivered in release `44.11.0` of the extensions, thanks for posting this @jriordan22849

PR #773, opened 2018-11-27
Set the Duration of a Scratch Org

renatoliveira:
### What does this PR do?

This adds another inputBox to the process of creating a scratch org, right after specifying the org's alias. It allows the user to specify the expiration days for the scratch org (other than just using the default 7 every time).

It addresses #768.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -43,7 +43,7 @@ export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
       .withArg('force:org:create')
       .withFlag('-f', `${selectionPath}`)
       .withFlag('--setalias', data.alias)
-      .withFlag('-d', data.expirationDays)
+      .withFlag('--durationdays', data.expirationDays)
       .withArg('--setdefaultusername')
       .withLogName('force_org_create_default_scratch_org')
       .build();
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -87,7 +87,6 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
     ) {
       scratchOrgExpirationInDays = defaultExpirationdate;
     }
-    // return params;
     return {
       type: 'CONTINUE',
       data: {
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -85,7 +85,6 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
       return { type: 'CANCEL' };
     }
     if (
-      scratchOrgExpirationInDays !== undefined &&
       !Number.isNaN(Number.parseInt(scratchOrgExpirationInDays)) &&
       Number.parseInt(scratchOrgExpirationInDays) >= 1 &&
       Number.parseInt(scratchOrgExpirationInDays) <= 30
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -71,7 +71,7 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
       prompt: nls.localize(
         'parameter_gatherer_enter_scratch_org_expiration_days'
       ),
-      placeHolder: defaultExpirationdate.toString()
+      placeHolder: defaultExpirationdate
     } as vscode.InputBoxOptions;
     const alias = await vscode.window.showInputBox(aliasInputOptions);
     // Hitting enter with no alias will use the value of `defaultAlias`
--- a/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
@@ -27,7 +27,7 @@ import {
 } from './commands';
 
 export const DEFAULT_ALIAS = 'vscodeScratchOrg';
-export const DEFAULT_EXPIRATION_DATE = 7;
+export const DEFAULT_EXPIRATION_DATE = '7';
 export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
   AliasAndFileSelection
 > {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -51,6 +51,8 @@ export const messages = {
   parameter_gatherer_enter_username_name: 'Enter target username',
   parameter_gatherer_enter_alias_name:
     'Enter an org alias or use default alias',
+  parameter_gatherer_enter_scratch_org_expiration_days:
+    'Enter the expiration days for the scratch org',
```
ruthemmanuelle: Maybe 'Enter the number of days until scratch org expiration (1–30 days)'?

renatoliveira: "Enter the number of days until scratch org expiration (1-30 days)" sounds a little better in my opinion. What do you think? Also, does it make sense to check if the user inputted a valid number (1 to 30) instead of letting the command fail?

JimSuplizio: @renatoliveira @ruthemmanuelle 
I like message but how about we put the default in there? Something like "**Enter the number of days until scratch org expiration (1-30 days, default 7 days)**". @ruthemmanuelle, just to check, is 30 days the max life?

For the user input, it should default to 7 so the user doesn't have to type anything unless they care to change the default. As for verification, yes, it should definitely verify the input is a number and it falls within the min/max range.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -64,18 +67,39 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
...
+    let scratchOrgExpirationInDays = await vscode.window.showInputBox(
+      expirationDays
+    );
+    if (
+      scratchOrgExpirationInDays === undefined ||
+      scratchOrgExpirationInDays === ''
+    ) {
+      scratchOrgExpirationInDays = defaultExpirationdate;
+    }
+    // return params;
```
JimSuplizio: We should definitely be verifying the user input is a valid number in between the min/max number of days.

JimSuplizio: We should definitely be verifying that the user input was a number and falls between the min/max days.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -42,6 +43,7 @@ export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
       .withArg('force:org:create')
       .withFlag('-f', `${selectionPath}`)
       .withFlag('--setalias', data.alias)
+      .withFlag('-d', data.expirationDays)
```
vazexqi: Please use the full flag instead of the shortcut of `-d`.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
vazexqi: Remove dead code.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -36,16 +41,19 @@ describe('Force Org Create', () => {
     it('Should return cancel if alias is undefined', async () => {
       const gatherer = new AliasGatherer();
       const response = await gatherer.gather();
-      expect(inputBoxSpy.calledOnce).to.be.true;
+      // expect(inputBoxSpy.calledOnce).to.be.true;
```
vazexqi: Remove dead code.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -54,9 +62,12 @@ describe('Force Org Create', () => {
     it('Should return Continue with inputted alias if user input is not undefined or empty', async () => {
       const gatherer = new AliasGatherer();
       const response = await gatherer.gather();
-      expect(inputBoxSpy.calledThrice).to.be.true;
+      // expect(inputBoxSpy.calledThrice).to.be.true;
```
vazexqi: Remove dead code.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -20,13 +20,18 @@ describe('Force Org Create', () => {
...
     before(() => {
       inputBoxSpy = sinon.stub(vscode.window, 'showInputBox');
       inputBoxSpy.onCall(0).returns(undefined);
+
       inputBoxSpy.onCall(1).returns('');
-      inputBoxSpy.onCall(2).returns(TEST_ALIAS);
+      inputBoxSpy.onCall(2).returns('');
+
+      inputBoxSpy.onCall(3).returns(TEST_ALIAS);
+      inputBoxSpy.onCall(4).returns('8');
```
vazexqi: What is this magic number of 8? See https://refactoring.guru/replace-magic-number-with-symbolic-constant on how to make this code easier to read.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -36,16 +41,19 @@ describe('Force Org Create', () => {
...
+      // expect(inputBoxSpy.calledOnce).to.be.true;
+      expect(inputBoxSpy.callCount).to.equal(1);
       expect(response.type).to.equal('CANCEL');
     });
 
     it('Should return Continue with default alias if user input is empty string', async () => {
       const gatherer = new AliasGatherer();
       const response = await gatherer.gather();
-      expect(inputBoxSpy.calledTwice).to.be.true;
+      // expect(inputBoxSpy.calledTwice).to.be.true;
```
vazexqi: Remove dead code.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Yes, 30 days is the max life for scratch orgs. If we're going to allow the user to leave the field blank to accept the default, how about this? 

'Enter the number of days until scratch org expiration (1–30), or press Enter to accept the 7-day default'

Note that there's an en dash, not a hyphen, in "1–30".

And, I agree that we should validate the user's input.

renatoliveira: Or we can also set the default seven day message as the placeholder :)

"Enter the number of days until scratch org expiration (1-30 days)"

placeholder: "Press Enter to accept the default (7 days)"

renatoliveira:
@vazexqi Thanks for the review! I will take those considerations and make the appropriate changes to the code. 👍

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Good idea re: making use of the placeholder so we can make this message shorter. 

For the org alias prompt, we use the default value as the placeholder, and we show "Enter an org alias or use default alias (Press 'Enter' to confirm or 'Escape' to cancel)". The first part is `parameter_gatherer_enter_alias_name`. It doesn't look like we have a message for that latter part, so VS Code must be adding it automatically. (https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts#L63 is where we process `parameter_gatherer_enter_alias_name`.)

Let's do this:

`parameter_gatherer_enter_scratch_org_expiration_days`: 
'Enter the number of days (1–30) until scratch org expiration or use the default value' (followed by the default message about Enter and Escape)

placeholder: 
'7'

renatoliveira:
I feel like it is worth to mention that I did solve the validation thing and how. To check this and to check the day range, I converted the user's input to an integer. So if the user somehow inputs a float, it will get converted to a valid integer. I hope this doesn't look like a hack. 😅

packages/salesforcedx-vscode-core/src/messages/i18n.ts
JimSuplizio: @ruthemmanuelle @renatoliveira 
The message 'Enter the number of days (1–30) until scratch org expiration or use the default value'  is a bit confusing unless you happen to know what the default is.

'Enter the number of days (1–30) until scratch org expiration or use the default value (7)' may be better here.

JimSuplizio:
@vazexqi I've gone ahead and resolved the conversations where @renatoliveira has made the requested changes. I've got one minor comment still on the user message and a minor ask on the number conversion/verification.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
JimSuplizio: @renatoliveira @ruthemmanuelle Things look pretty much good to go except the changed to add (7) to the end of the string. Since it's approved by @ruthemmanuelle I'll probably just make that change locally instead of asking for another iteration.

lcampos:
Merged in #799

PR #777, opened 2018-11-28
Jsuplizio/manifestations

JimSuplizio:
### What does this PR do?
The "SFDX: Retrieve/Deploy Source from Org" calls the underlying CLI "sfdx force:source:retrieve" with --manifest or --sourcepath depending on where the command was invoked from. Right clicking on the project's manifest directory and running the command will attempt to run the underlying command with the --manifest option on the directory which will outright fail.

With these changes I check if the type is manifest and, if so, check if the input "file" is actually a manifest file or the manifest directory. If the input is the manifest directory and there's only one .xml file I'll use that but if there are multiple files I'll pop up a showOpenDialog and allow the user to pick one or cancel. If the user picks one the command will continue as normal. If the user hits cancel instead of choosing a file then the command gets canceled and doesn't run. I've attached a screenshot of the open dialog and  it's worth noting that both the "Select Manifest File" and the "Manifest" filter dropdown will both get localized.
manifestfileselect

@W-5230669@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
@@ -4,6 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
 
@@ -56,11 +57,68 @@ export class ManifestOrSourcePathGatherer
       // corresponding to the explorer path when we begin supporting multiple workspace folders
       const workspaceRootPath = vscode.workspace.workspaceFolders[0].uri.fsPath;
       const manifestPath = path.join(workspaceRootPath, 'manifest');
-      const isManifestFile = this.explorerPath.includes(manifestPath);
-      const type = isManifestFile ? FileType.Manifest : FileType.Source;
-      return { type: 'CONTINUE', data: { filePath: this.explorerPath, type } };
+      const isManifest = this.explorerPath.includes(manifestPath);
+      const type = isManifest ? FileType.Manifest : FileType.Source;
+
+      // If the type is manifest check whether or not the user selected an actual
+      // file or just the manifest directory. If the user selected the manifest
+      // directory then check that directory for manifest files. If there's only
+      // one manifest file then use it, if there none or more than one then give
+      // them a file select dialog and make them select one.
+      if (isManifest) {
+        let manifestFile = this.explorerPath;
+        const stat = fs.lstatSync(manifestFile);
+        if (stat.isDirectory()) {
+          const manifestFileList = this.getManifestFiles(manifestFile);
+          if (manifestFileList.length === 1) {
+            manifestFile = manifestFileList[0];
+          } else {
+            const localizedFilterButtonText = nls.localize(
+              'select_manifest_filter_button_text'
+            );
+            const fileUris = await vscode.window.showOpenDialog({
+              canSelectFiles: true,
+              canSelectFolders: false,
+              canSelectMany: false,
+              openLabel: nls.localize('select_manifest_file_button_text'),
+              filters: { localizedFilterButtonText: ['xml'] },
+              defaultUri: vscode.Uri.file(manifestFile)
+            });
+            if (fileUris && fileUris.length === 1) {
+              manifestFile = fileUris[0].fsPath;
+            } else {
+              return { type: 'CANCEL' };
+            }
+          }
+        }
+        return {
+          type: 'CONTINUE',
+          data: { filePath: manifestFile, type }
+        };
+      } else {
+        return {
+          type: 'CONTINUE',
+          data: { filePath: this.explorerPath, type }
+        };
+      }
     }
     return { type: 'CANCEL' };
+  }
+
+  private getManifestFiles(manifestDir: string): string[] {
+    const manifestFileExtension = '.xml';
+    const manifestFiles: string[] = [];
+    const files = fs.readdirSync(manifestDir);
+    files.forEach(rawFile => {
+      const fileName = path.join(manifestDir, rawFile);
+      const stat = fs.lstatSync(fileName);
+      if (!stat.isDirectory()) {
+        if (path.extname(fileName).toLowerCase() === manifestFileExtension) {
+          manifestFiles.push(fileName);
+        }
+      }
+    });
+    return manifestFiles;
   }
 }
 
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
@@ -81,7 +81,7 @@ export class ManifestOrSourcePathGatherer
               canSelectFolders: false,
               canSelectMany: false,
               openLabel: nls.localize('select_manifest_file_button_text'),
-              filters: { localizedFilterButtonText: ['xml'] },
+              filters: { [localizedFilterButtonText]: ['xml'] },
               defaultUri: vscode.Uri.file(manifestFile)
             });
             if (fileUris && fileUris.length === 1) {
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -82,6 +82,8 @@ export const messages = {
     'Deleting source files deletes the files from your computer and removes the corresponding metadata from your default org. Are you sure you want to delete this source from your project and your org?',
   confirm_delete_source_button_text: 'Delete Source',
   cancel_delete_source_button_text: 'Cancel',
+  select_manifest_file_button_text: 'Select Manifest File',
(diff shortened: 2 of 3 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -82,6 +82,8 @@ export const messages = {
     'Deleting source files deletes the files from your computer and removes the corresponding metadata from your default org. Are you sure you want to delete this source from your project and your org?',
   confirm_delete_source_button_text: 'Delete Source',
   cancel_delete_source_button_text: 'Cancel',
+  select_manifest_file_button_text: 'Select Manifest File',
+  select_manifest_filter_button_text: 'Manifest',
```
JimSuplizio: @ruthemmanuelle both of these strings would need your approval (or correction)

packages/salesforcedx-vscode-core/test/commands/forceSourceRetrieve.test.ts
```diff
@@ -70,26 +72,264 @@ describe('Manifest or Sourcepath Gatherer', () => {
       expect.fail('The workspace root path was undefined');
     }
   });
+});
+
```
JimSuplizio: It's worth noting that I am creating/deleting the manifest directory and manifest files. When running the core tests there's a system-tests/assets/sfdx-simple project directory already being created under salesforcedx-vscode/packages (which is  why the vscode.workspace.workspaceFolders is even set). I'm creating an manifest directory before the batch of Manifest Gather tests and deleting it after. For the manifest files, themselves, I'm just creating/cleaning the files before/after each scenario. I did this because it was more efficient than mocking all of the items from fs that I'd need to mock in order to test the additional code paths in the ManifestOrSourcePathGatherer.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: @JimSuplizio - These look reasonable. In your screenshot, it looks like Manifest is an option in a dropdown menu. Are there other options in that menu?

JimSuplizio: @ruthemmanuelle No, there are no other options, the only valid type of manifest file is an XML file. In VS Code filters are always a dropdownlist which, in this case, translates to a dropdownlist that only has 1 item.

ruthemmanuelle: Ok. I'll approve the PR.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -56,12 +57,69 @@ export class ManifestOrSourcePathGatherer
...
+      // file or just the manifest directory. If the user selected the manifest
+      // directory then check that directory for manifest files. If there's only
+      // one manifest file then use it, if there none or more than one then give
+      // them a file select dialog and make them select one.
+      if (isManifest) {
+        let manifestFile = this.explorerPath;
+        const stat = fs.lstatSync(manifestFile);
+        if (stat.isDirectory()) {
+          const manifestFileList = this.getManifestFiles(manifestFile);
+          if (manifestFileList.length === 1) {
```
allileong: There's an edge case where a user may have one manifest file as well as one or more directories containing other manifest files in their top level manifest directory. In that case, `manifestFileList.length === 1` is true, but there could be other manifests in their manifest directory. This is a weird case, but I could see this happening if users come to really like the file picker feature and use it all the time, and then happen to have a top level manifest and also a number of other manifests organized into subfolders. They may expect to see the file picker, but instead, the top level manifest is deployed.

JimSuplizio: @allileong @lcampos That could certainly be true but if they select the directory and do that, the acceptance criteria (pasted below, copied from the work item) states that I don't prompt in that case. That's definitely up for discussion but we should probably include @ntotten .

Acceptance criteria: 1) In the case that the manifest folder only contains 1 manifest, don't prompt, just use that manifest 2) In the event that the manifest folder contains multiple manifests, prompt to select manifest.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -56,12 +57,69 @@ export class ManifestOrSourcePathGatherer
       // corresponding to the explorer path when we begin supporting multiple workspace folders
       const workspaceRootPath = vscode.workspace.workspaceFolders[0].uri.fsPath;
       const manifestPath = path.join(workspaceRootPath, 'manifest');
-      const isManifestFile = this.explorerPath.includes(manifestPath);
-      const type = isManifestFile ? FileType.Manifest : FileType.Source;
-      return { type: 'CONTINUE', data: { filePath: this.explorerPath, type } };
+      const isManifest = this.explorerPath.includes(manifestPath);
+      const type = isManifest ? FileType.Manifest : FileType.Source;
+
+      // If the type is manifest check whether or not the user selected an actual
```
lcampos: No need for this comment, code is clear enough to follow.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -56,12 +57,69 @@ export class ManifestOrSourcePathGatherer
...
+        const stat = fs.lstatSync(manifestFile);
+        if (stat.isDirectory()) {
+          const manifestFileList = this.getManifestFiles(manifestFile);
+          if (manifestFileList.length === 1) {
+            manifestFile = manifestFileList[0];
+          } else {
+            const localizedFilterButtonText = nls.localize(
+              'select_manifest_filter_button_text'
+            );
+            const fileUris = await vscode.window.showOpenDialog({
```
lcampos: We should be using the `FileSelector` class here or extending it. Look at how we do it for `SFDX: Create a Default Scratch Org ...`; `forceOrgCreate.ts`.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -56,12 +57,69 @@ export class ManifestOrSourcePathGatherer
...
+      // If the type is manifest check whether or not the user selected an actual
+      // file or just the manifest directory. If the user selected the manifest
+      // directory then check that directory for manifest files. If there's only
+      // one manifest file then use it, if there none or more than one then give
+      // them a file select dialog and make them select one.
+      if (isManifest) {
+        let manifestFile = this.explorerPath;
+        const stat = fs.lstatSync(manifestFile);
+        if (stat.isDirectory()) {
+          const manifestFileList = this.getManifestFiles(manifestFile);
```
lcampos: I think that we should show an error message whenever they trigger this from the manifest folder and there's no `.xml` manifest files in it.

lcampos:
@JimSuplizio the acceptance criteria talks about this being for retrieve and deploy but these changes are exclusive to retrieve.

JimSuplizio:
@lcampos your previous comment "the acceptance criteria talks about this being for retrieve and deploy but these changes are exclusive to retrieve"  is incorrect. ManifestOrSourcePathGatherer is used by both retrieve and deploy it just happens to be defined in forceSourceRetrieve.ts and the tests for this class are in forceSourceRetrieve.test.ts.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
JimSuplizio: Sure, why not. The next person to touch this code can just reason out what's going on rather than reading the intent.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
ntotten: This is a really good observation @allileong. I didn't think of this case.

I think you are right that we should show the picker in this case. The acceptance criteria should be amended as I missed this edge case.

JimSuplizio: @ntotten to be clear then, if what the user selects isn't a file then always pop up the picker dialog?

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
JimSuplizio: Adding @ntotten and @allileong considering the discussion above revolves around always popping up a picker unless the user selected a manifest file.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
lcampos: If you think the code itself doesn't communicate intent then change it so it does and we don't have to add this type of comments.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
allileong: Maybe we can open up the picker as long as there exist any files or directories inside of the selected directory, but if the selected directory is empty, then notify the user that the directory does not contain any manifests.

lcampos:
@JimSuplizio my bad, I must've done something wrong while testing it 👍

Issue #759, opened 2018-11-16
Time Stamp on SFDX commands in the VSCode Output tab. 


dannybusiness:
**Is your feature request related to a problem? Please describe.**
I often will be editing multiple files and changing windows between a browser, VSCode and other apps. Sometimes I will have made several different code edits and not know when was the last time I pushed and if the server has the latest version of my code. 

**Describe the solution you'd like**
It would be helpful in the output to have a timestamp for each operation. 

**Describe alternatives you've considered**
I haven't thought of any other alternatives. This isn't critical but a nice to have.

lcampos:
This was delivered in release `44.11.0` of the extensions, thanks for posting the idea @dannybusiness.

dannybusiness:
Thank you! I'm so happy to have this feature generally available.

PR #780, opened 2018-11-29
Adding timestamps to command executions in output…

lcampos:
… view

### What does this PR do?
Adds timestamps to every sfdx command execution that's being printed in the Output panel.

As the title suggests, this is a work in progress cause I want to run the automation and see if I missed any test failures and get feedback.

#759 

timestamp_preview_highlight

### DIFF
--- a/packages/salesforcedx-vscode-core/src/channels/channelService.ts
+++ b/packages/salesforcedx-vscode-core/src/channels/channelService.ts
@@ -44,10 +44,14 @@ export class ChannelService {
     this.channel.appendLine(execution.command.toString());
     this.channel.appendLine('');
 
-    this.channel.appendLine(execution.command.toCommand());
+    this.channel.appendLine(
+      this.getExecutionTime() + ' ' + execution.command.toCommand()
+    );
 
     execution.processExitSubject.subscribe(data => {
-      this.channel.append(execution.command.toCommand());
+      this.channel.append(
+        this.getExecutionTime() + ' ' + execution.command.toCommand()
+      );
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -79,6 +85,19 @@ export class ChannelService {
     });
   }
 
+  private getExecutionTime() {
```
lcampos: Using the same time format as we're using in apex lsp and debug logs.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -44,10 +44,14 @@ export class ChannelService {
...
     this.channel.appendLine('');
 
-    this.channel.appendLine(execution.command.toCommand());
+    this.channel.appendLine(
+      this.getExecutionTime() + ' ' + execution.command.toCommand()
+    );
 
     execution.processExitSubject.subscribe(data => {
-      this.channel.append(execution.command.toCommand());
+      this.channel.append(
```
lcampos: I looked at having this as part of the classes in `commandExecutor.ts`, having an internal variable (per class) to track start and end of execution and then leverage them whenever we wanted to print them. It turned out to be more complex and un-intuitive in case we wanted to use those timestamps somewhere else other than for the output view.
I then turned to the simplest approach which is adding it here on the right execution methods. Let me know if I should look at moving this somewhere else where it makes more sense.

cc @vazexqi

PR #783, opened 2018-11-30
[WIP] Easy Deploy on Save

allileong:
I'd like to get feedback on my implementation approach for Easy Deploy on Save in case my current approach requires major changes.

### What does this PR do?
* Adds a workspace configuration called `salesforcedx-vscode-core.push-or-deploy-on-save.enabled` that is set to `false` by default
* Works in both scratch orgs and non-scratch orgs

Still left to do:
* Update @salesforce/core package to the latest version (in this PR it has been bumped from 0.18.2 to 0.23.1, but the latest version available in npm is 0.24.8)
* Update typescript in all repo packages
* Add testing
* Test on OS's other than Mac
* Test out the alternative way to batch created/deleted files for a single deploy/push; Right now, whenever I detect a create/delete, I extend the wait time by 1 second. When the wait time of 1 second has passed without detecting any additional creates/deletes, then I initiate a single deploy/push

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005wp9BIAQ/view

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -23,7 +23,7 @@
   },
   "categories": ["Other"],
   "dependencies": {
-    "@salesforce/core": "0.18.2",
+    "@salesforce/core": "0.23.1",
     "@salesforce/salesforcedx-sobjects-faux-generator": "44.8.0",
     "@salesforce/salesforcedx-utils-vscode": "44.8.0",
     "adm-zip": "0.4.7",
@@ -50,7 +50,7 @@
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
     "sinon": "^2.3.6",
-    "typescript": "2.6.2",
+    "typescript": "2.9.1",
     "vscode": "1.1.17"
   },
   "scripts": {
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -158,7 +158,7 @@
             "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
         },
         {
-          "command": "sfdx.force.source.deploy",
+          "command": "sfdx.force.source.deploy.manifest.or.source.path",
           "when":
             "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
         },
@@ -346,7 +346,7 @@
           "when": "false"
         },
         {
-          "command": "sfdx.force.source.deploy",
+          "command": "sfdx.force.source.deploy.manifest.or.source.path",
           "when": "false"
         },
         {
@@ -545,7 +545,7 @@
         "title": "%force_source_retrieve_this_source_text%"
       },
       {
-        "command": "sfdx.force.source.deploy",
+        "command": "sfdx.force.source.deploy.manifest.or.source.path",
         "title": "%force_source_deploy_text%"
       },
       {
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -577,6 +577,11 @@
           "type": ["boolean"],
           "default": true,
           "description": "%telemetry_enabled_description%"
+        },
+        "salesforcedx-vscode-core.push-or-deploy-on-save.enabled": {
+          "type": ["boolean"],
+          "default": false,
+          "description": "%push_or_deploy_on_save_enabled_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
@@ -86,7 +86,7 @@ export class ForceApexTestRunCommandFactory {
   private data: ApexTestQuickPickItem;
   private getCodeCoverage: boolean;
   private builder: SfdxCommandBuilder = new SfdxCommandBuilder();
-  private testRunExecutorCommand: Command;
+  private testRunExecutorCommand: Command | undefined;
 
   constructor(data: ApexTestQuickPickItem, getCodeCoverage: boolean) {
     this.data = data;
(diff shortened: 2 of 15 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
```diff
@@ -86,7 +86,7 @@ export class ForceApexTestRunCommandFactory {
   private data: ApexTestQuickPickItem;
   private getCodeCoverage: boolean;
   private builder: SfdxCommandBuilder = new SfdxCommandBuilder();
-  private testRunExecutorCommand: Command;
+  private testRunExecutorCommand: Command | undefined;
```
allileong: This was required as a result of updating Typescript.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -50,7 +50,7 @@
     "mock-spawn": "0.2.6",
     "nyc": "^11.0.2",
     "sinon": "^2.3.6",
-    "typescript": "2.6.2",
+    "typescript": "2.9.1",
```
vazexqi: @allileong - If we are upgrading TypeScript, we should do it consistently across all modules. Usually when I do that, I do it in a separate commit to let the test run fully.

Is there a reason why we only update TypeScript for this one module?

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -0,0 +1,135 @@
...
+
+async function createSourceFileWatcher(): Promise<vscode.FileSystemWatcher | null> {
+  if (
+    vscode.workspace.workspaceFolders &&
+    vscode.workspace.workspaceFolders.length > 0
+  ) {
+    const sfdxProjectPath = vscode.workspace.workspaceFolders[0].uri.fsPath;
+    const sfdxProject = await SfdxProject.resolve(sfdxProjectPath);
+    const sfdxProjectJson = (await sfdxProject.resolveProjectConfig()) as SfdxProjectJson;
+    const packageDirectoryPaths = sfdxProjectJson.packageDirectories.map(
```
allileong: I am using the @salesforce/core api to fetch the packageDirectories info from the sfdx-project.json in the workspace. I then create a fileWatcher to detect any creations/changes/deletions made to files inside of the packageDirectories. This approach is not able to differentiate between which files are and aren't source.

packages/salesforcedx-vscode-core/package.json
allileong: @vazexqi I plan to complete that as I finish this work. I just wanted to open a WIP review to make sure that I'm on the right track with my approach.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -97,9 +100,35 @@ export class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
...
+      type: 'CONTINUE',
+      data: { filePath: sourcePaths, type: FileType.Source }
+    };
+  }
+}
+
 const workspaceChecker = new SfdxWorkspaceChecker();
 
-export async function forceSourceDeploy(explorerPath: any) {
+export async function forceSourceDeployMultipleSourcePaths(uris: vscode.Uri[]) {
```
allileong: This new deploy command is only called when a file within an sfdx packageDirectory is created/changed/deleted so it does not need to use the ManifestOrSourcePathGatherer to determine whether the passed in paths belong to source or manifests.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -10,34 +10,58 @@ import * as vscode from 'vscode';
 
 import { ForceConfigGet } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 
-export async function setupWorkspaceOrgType() {
+export enum OrgType {
+  SourceTracked,
+  NonSourceTracked
+}
+
+export async function getWorkspaceOrgType(): Promise<OrgType> {
```
allileong: I pulled out this getWorkspaceOrgType method so that I could use it inside of pushOrDeployOnSave.ts

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -0,0 +1,135 @@
...
+        );
+      }
+    }
+  } catch (e) {
+    if (e.name === 'NamedOrgNotFound') {
+      notificationService.showErrorMessage(
+        nls.localize('error_fetching_auth_info_text')
+      );
+    } else if (e.name === 'NoDefaultusernameSet') {
+      // Do nothing.
```
allileong: @ntotten, if no defaultusername is set in the workspace, should I notify the customer that their changes are not being deployed/pushed to any org?

packages/salesforcedx-vscode-core/package.json
allileong: @vazexqi that's a good idea. When I finalize my upgrade, I will do it in a separate commit

packages/salesforcedx-vscode-core/test/telemetry/MockContext.ts
```diff
@@ -29,6 +29,7 @@ class MockMemento implements Memento {
 export class MockContext implements ExtensionContext {
   constructor(mm: boolean) {
     this.globalState = new MockMemento(mm);
+    this.workspaceState = new MockMemento(mm);
```
allileong: Another change resulting from the TS upgrade

packages/salesforcedx-vscode-core/src/statuses/taskView.ts
```diff
@@ -88,8 +88,8 @@ export class TaskViewService implements TreeDataProvider<Task> {
 }
 
 export class Task extends TreeItem {
-  public readonly label: string;
-  public readonly collapsibleState: TreeItemCollapsibleState;
+  public readonly label: string | undefined;
```
allileong: Another change resulting from the TS upgrade

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
ntotten: I think it would be a good idea. Maybe something in the console? "Deploy on save is enabled, but no org is authorized. No files deployed to org." //cc @ruthemmanuelle

PR #788, opened 2018-12-04
Keep the order of test methods per test class in Apex Test sidebar

lcampos:
### What does this PR do?
Fixes Apex Test sidebar sorting so that the test methods are kept in the same order as they are in the test class.

@W-5652397@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
+++ b/packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
@@ -118,32 +118,34 @@ export class ApexTestOutlineProvider
     }
     this.rootNode.children = new Array<TestNode>();
     if (this.apexTestInfo) {
-      this.apexTestInfo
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -118,32 +118,34 @@ export class ApexTestOutlineProvider
...
+        if (
+          this.rootNode &&
+          !(this.rootNode.children.indexOf(apexGroup) >= 0)
+        ) {
+          this.rootNode.children.push(apexGroup);
+        }
+        this.testStrings.add(apexGroup.name);
+      });
+      // Sorting independently so we don't loose the order of the test methods per test class.
+      this.apexTestInfo.sort((a, b) =>
```
lcampos: Moving the sort to happen after we create the groups & test method maps so that we don't loose the order of the test methods per class. 

Addressing it in a different way would mean having a more complex sort/for each function that considers the classes name and each method's code location.

PR #794, opened 2018-12-05
Use newer lwc-language-server

vazexqi:
### What does this PR do?

Use the new lwc-language-server.

@W-5675238@


lcampos:
@vazexqi looks like that version is not published yet. Here's the error from circleci `No matching version found for lwc-language-server@1.8.2`

PR #795, opened 2018-12-05
deploy/retrieve using manifest

JimSuplizio:
### What does this PR do?
SFDX: Deploy Source to Org and SFDX: Retrieve Source From Org commands need to work on sources and directories but the processing was such that, when it came to dealing with manifest files, they were allowing incorrect inputs. What I did for this change was created 2 new commands **SFDX: Deploy Source to Org using Manifest** and **SFDX: Retrieve Source from Org using Manifest** and a new language identifier. The language identifier utilizes the filenamePatterns to specifically identify XML files under the project's manifest directory as "forcesourcemanifest" language files. This allowed WHEN clauses to be added to the new commands to only enable them in the explorer commands when the user right clicks on an XML in the manifest directory. Further, this allowed me to disable the original commands when an XML file was selected under the manifest directory. This also had the side benefit of requiring far less code changes. It's worth noting that the original commands are still enabled in some places where they shouldn't be but there's a separate bug for that.

This first snapshot is what it looks like when the user selects an XML file under the manifest directory. Notice that only the "using Manifest" commands are enabled.
manifestdirwithxmlfile

The second snapshot is what it looks like when the user select any other file under the manifest directory (or in the project) and right clicks. Notice that the original commands are enabled (they're enabled everywhere as they previously were). 
manifestdirnonxmlfile

@W-5230669@


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -73,8 +73,12 @@
   "force_source_retrieve_text": "SFDX: Retrieve Source from Org",
   "force_source_retrieve_this_source_text":
     "SFDX: Retrieve This Source from Org",
+  "force_source_retrieve_using_manifest_text":
+    "SFDX: Retrieve Source from Org using Manifest",
   "force_source_deploy_text": "SFDX: Deploy Source to Org",
   "force_source_deploy_this_source_text": "SFDX: Deploy This Source to Org",
+  "force_source_deploy_using_manifest_text":
+    "SFDX: Deploy Source to Org using Manifest",
```
JimSuplizio: @ruthemmanuelle These are two new commands that are specific to source deploy/retrieve using the manifest file. I'd just taken the existing commands and added "using Manifest".

ruthemmanuelle: Since "using" is more than four characters long, it should be capitalized even though it's a preposition. But maybe "SFDX: Retrieve Source in Manifest from Org" and "SFDX: Deploy Source in Manifest to Org" would be better? That way they'd be a bit more similar to the existing "... This Source ..." deploy/retrieve commands.

ntotten:
What happens when I right click on the non-`.xml` files and try to run the regular retrieve/deploy. Do I get a nice error message now or is that another work item?

packages/salesforcedx-vscode-core/package.nls.json
JimSuplizio: Adding @allileong to this. "Deploy Source in Manifest" isn't really grammatically correct. You're deploying source identified by the manifest there's no actual source in the manifest. Initially I'd thought about Deploy Source With Manifest but that wasn't correct either. You're using the manifest to identify what source you're retrieving or deploying which is why using made sense.

allileong: I like "Using" better since it does seem the most accurate, but I think both are good options, and the user would likely understand how to use the command in either case.

JimSuplizio:
> What happens when I right click on the non-`.xml` files and try to run the regular retrieve/deploy. Do I get a nice error message now or is that another work item?

@ntotten I commented on that in the original description 
"It's worth noting that the original commands are still enabled in some places where they shouldn't be but there's a separate bug for that."
It'll pass whatever the user selected off the **source** version command and return the error that the CLI gives but as mentioned, that source commands need to be limited and currently aren't.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -598,6 +624,11 @@
       {
         "id": "ignore",
         "filenames": [".forceignore"]
+      },
+      {
+        "id": "forcesourcemanifest",
+        "aliases": ["ForceSourceManifest", "forcesourcemanifest"],
+        "filenamePatterns": ["**/manifest/**/*.xml"]
```
lcampos: This is a poorly documented gem, nice find @JimSuplizio

packages/salesforcedx-vscode-core/package.json
```diff
@@ -155,17 +155,27 @@
         {
           "command": "sfdx.force.source.retrieve",
           "when":
-            "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
+            "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest'"
```
lcampos: We should also add `&& resourceFilename != manifest` so that these commands (retrieve, deploy, delete) are not showed when right-clicking the manifest folder.

JimSuplizio: I agree but right now, all of those other commands should be taken care of with the other work item. I've done the minimum to prevent them from being on the menu when the *Manifest* commands are active.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Well, the source is "in" the manifest in the same sense in which chapters in a book could be "in" a table of contents. But I'm ok with "SFDX: Deploy Source to Org Using Manifest", since you guys think it's better. Just make sure that you capitalize "Using".

ruthemmanuelle: @JimSuplizio - I know I said in my last comment that I was ok with the "...Using Manifest" option, but I really think that "SFDX: Retrieve Source in Manifest from Org" and "SFDX: Deploy Source in Manifest to Org" are better. Since @ntotten liked my last comment, I assume he does too? 

We want the user to be able to grasp the purpose of the command as quickly as possible while they're scanning the command palette, and putting "Manifest" earlier in the message would help with that. The "from Org"/"to Org" part doesn't distinguish these commands from the others.

And, if users don't know what a manifest is then using "in Manifest" would help them understand that the types to deploy/retrieve are listed in the manifest. 

"SFDX: Retrieve Source in Manifest from Org" and "SFDX: Deploy Source in Manifest to Org" also sound better (that is, they read more smoothly) than "SFDX: Retrieve Source from Org Using Manifest" and "SFDX: Deploy Source to Org Using Manifest".

JimSuplizio: @ruthemmanuelle I'll change them to "SFDX: Retrieve Source in Manifest from Org" and "SFDX: Deploy Source in Manifest to Org"

PR #799, opened 2018-12-06
Jsuplizio/renatoliveira org duration

JimSuplizio:
### What does this PR do?
This change is for a merge of a PR made by @renatoliveira. The changes are to allow setting the duration of a scratch org. The original PR is here: https://github.com/forcedotcom/salesforcedx-vscode/pull/773

I did have to make a couple of minor updates which I'll comment on inline.

@W-5648418@


packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -69,13 +72,44 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
...
       return { type: 'CANCEL' };
     }
-    return alias === ''
-      ? { type: 'CONTINUE', data: { alias: defaultAlias } }
-      : { type: 'CONTINUE', data: { alias } };
+    const expirationDays = {
+      prompt: nls.localize(
+        'parameter_gatherer_enter_scratch_org_expiration_days'
+      ),
+      value: defaultExpirationdate
```
JimSuplizio: For the expiration days, which has an actual default, use **value** in the InputBoxOptions instead of **placeHolder**. The difference here is that hitting enter on the placeholder returns an undefined and since we're defaulting to 7 there's no need to force the user to do additional typing, they can just hit enter to accept the default value. The InputBox does highlight the default value so if the user opts to change it they don't have to delete it.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -18,45 +18,99 @@ import { nls } from '../../src/messages';
...
     const TEST_ALIAS = 'testAlias';
     const TEST_WORKSPACE = 'sfdxsimple'; // FYI: This test uses the workspace created by the system tests to run
+    const TEST_ORG_EXPIRATION_DAYS = '7';
+    const TEST_ORG_EXPIRATION_DAYS_PLUS_ONE_DAY = '8';
+    const TEST_ORG_EXPIRATION_DAYS_INPUT_FLOAT = '8.2';
+    const TEST_ORG_EXPIRATION_DAYS_INPUT_INVALID_RANGE = '31';
     let inputBoxSpy: sinon.SinonStub;
 
-    before(() => {
+    beforeEach(() => {
```
JimSuplizio: The before/after were changed to beforeEach and afterEach. There were several reasons for this, the least of which was the running total on the inputBoxSpy number of calls should not be across scenarios. Similarly, the returns for the inputBoxSpy calls should be in the individual scenarios which makes them both easier to read and diagnose if there's a failure.

packages/salesforcedx-vscode-core/test/commands/forceOrgCreate.test.ts
```diff
@@ -18,45 +18,99 @@ import { nls } from '../../src/messages';
...
       inputBoxSpy.restore();
     });
 
     it('Should return cancel if alias is undefined', async () => {
+      inputBoxSpy.onCall(0).returns(undefined);
       const gatherer = new AliasGatherer();
       const response = await gatherer.gather();
-      expect(inputBoxSpy.calledOnce).to.be.true;
-      expect(response.type).to.equal('CANCEL');
+      expect(inputBoxSpy.callCount).to.equal(1);
```
vazexqi: Is there a reason to change it from `expect(inputBoxSpy.calledOnce).to.be.true` to `expect(inputBoxSpy.callCount).to.equal(1)`. I just want to understand if this is a preference or is there is a technical reason.

JimSuplizio: @vazexqi You'd have to look at the intermediate changes. There was previously a running total of count that went through all of the tests that was part of the original PR. I didn't actually change the .to.be.true to a count, myself. When I refactored before/after to beforeEach and afterEach everything was already checking the count for the number of calls (which, in most cases was a total of 2) and so I just went uniform checking the call count the same way through everything instead of calledOnce or calledTwice etc.

PR #800, opened 2018-12-06
Update Test Sidebar Results for Code Lens, Command Palette, and Sidebar

AnanyaJha:
### What does this PR do?

This pr updates the test sidebar to display results for any tests run through the sidebar, code lens, or command palette. 

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005QaelIAC/view


packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -111,6 +113,22 @@ export class ApexTestOutlineProvider
...
+  ) {
+    await this.refresh();
+    const testRunIdFile = path.join(apexTestPath, 'test-run-id.txt');
+    const testRunId = fs.readFileSync(testRunIdFile);
+    const testResultFilePath = path.join(
+      apexTestPath,
+      'test-result-' + testRunId + '.json'
+    );
+    if (testResultFile === testResultFilePath) {
+      this.readJSONFile(testResultFile);
```
AnanyaJha: Currently, I'm creating a watcher on the creation of the new json test result file and verifying that the file path returned from that matches the test id in the `test-result-id.txt` file. I haven't handled the scenario where the two don't match yet; `readJSONFile()` doesn't get called and the tests don't get updated right now.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -4,10 +4,8 @@
 //  * Licensed under the BSD 3-Clause license.
 //  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 //  */
+import { TestRunner } from '@salesforce/salesforcedx-utils-vscode/out/src/cli/';
 import * as events from 'events';
-import * as path from 'path';
-import * as pathExists from 'path-exists';
```
lcampos: Moving `TestRunner` to the `salesforcedx-utils-vscode` module allows us to remove `path-exists` from this module's `package.json`.

packages/salesforcedx-vscode-core/src/commands/forceApexTestRun.ts
```diff
@@ -107,9 +114,10 @@ export class ForceApexTestRunCommandFactory {
         );
         break;
       case TestType.Class:
-        this.builder = this.builder
-          .withFlag('--classnames', `${this.data.label}`)
-          .withArg('--synchronous');
```
lcampos: If I remember correctly we're removing this because it doesn't allow us to get the results in a json file. If that's true, did we ask the cli team about that behavior ?

AnanyaJha: Yeah, there is a bug with that functionality; including the synchronous argument should still result in a json file creation. I confirmed it was a bug with the cli team and filed it in gus.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
AnanyaJha: Actually, it looks like `path-exists` is needed for the `requirements.ts` file within the `salesforcedx-vscode-apex` module.

packages/salesforcedx-vscode-apex/test/views/readableApexTestRunExecutor.test.ts
```diff
@@ -0,0 +1,83 @@
+/*
```
AnanyaJha: This file adds tests to check the commands being generated by `readableApexTestRunExecutor` to run the apex tests through the sidebar.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -111,6 +110,22 @@ export class ApexTestOutlineProvider
     this.onDidChangeTestData.fire();
   }
 
+  public async onResultFileCreate(
+    apexTestPath: string,
+    testResultFile: string
+  ) {
+    await this.refresh();
```
lcampos: Shouldn't we only refresh the sidebar if the new file matches the data in `test-run-id.txt` ?

AnanyaJha: Yeah, I tested it and makes more sense to refresh only if the new file matches the correct data. The only difference is that now, the refresh happens immediately before the new results are displayed instead of immediately after the user starts a new test run- so it may not be _as_ apparent from the user's perspective that a refresh is occurring.

PR #802, opened 2018-12-07
Adding LWC snippets

muenzpraeger:
### What does this PR do?

Adds LWC snippets

W-5680524


packages/salesforcedx-vscode-lwc/snippets/lwc.json
```diff
@@ -0,0 +1,224 @@
+{
```
vazexqi: @muenzpraeger - What did we settle on for prettier? I'm asking because some of the snippets below look like they are using 4 spaces for the indentation (prettier's default is 2). I know that prettier would just do what it needs to do and format it but it would be nice to have the snippets be as close as possible.

packages/salesforcedx-vscode-lwc/snippets/lwc.json
```diff
@@ -0,0 +1,224 @@
...
+      "    }",
+      "}"
+    ]
+  },
+  "lwc-wire-apex-function": {
+    "prefix": "lwc-wire-apex-function",
+    "description": "Creates a @wire'd function for Apex with data modification",
+    "body": [
+      "@wire(${1:apex-method-name}, {$0})",
+      "apex${2:apex-function-name} ({error, data}) {",
```
allileong: Hi @muenzpraeger 
I believe this is a typo -- there is an extra "apex" at the beginning of the line. 

Also, are `apex-function-name` and `apex-method-name` referring to the same name? If so, I think it would be better to choose one to use consistently.

packages/salesforcedx-vscode-lwc/snippets/lwc.json
```diff
@@ -0,0 +1,224 @@
...
+      "    message: '${2:message}',",
+      "    variant: 'warning'",
+      "}));"
+    ]
+  },
+  "lwc-imperative-apex": {
+    "prefix": "lwc-imperative-apex",
+    "description": "Creates an imperative Apex call",
+    "body": [
+      "${1:method name}()",
```
allileong: @muenzpraeger Apologies, I'm not familiar with what lwc-imperative-apex is, so I want to double check. Should 'method' here be 'method', or should this be 'function', or have a hyphen between method and name instead?

vazexqi: It should be a method name that is exposed through Apex. There are more details in the lwc-recipes repo (soon to be public).

packages/salesforcedx-vscode-lwc/snippets/lwc.json
muenzpraeger: Methods exist in Apex, functions in JavaScript. Hence the distinction.

And yes, I've chosen to prefix things with `apex` Not necessary, but helps to distinct.

PR #811, opened 2018-12-11
Update right click menu label based on test status

brpowell:
Looking for some feedback on how I went about this. Also noticed there aren't any tests for the functionality of going to the error/definition, I'll work on that as well.

### What does this PR do?
When right clicking on a test that hasn't run, passed, or is skipped, the menu option now says "Go to Definition" instead of "Display Error", the latter which is shown only when the test has failed. 

W-5368845

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -95,8 +95,13 @@
           "when": "view == sfdx.force.test.view"
         },
         {
+          "command": "sfdx.force.test.view.runClassTests",
+          "when": "view == sfdx.force.test.view && viewItem == apexTestGroup",
+          "group": "inline"
+        },
+        {
           "command": "sfdx.force.test.view.runSingleTest",
-          "when": "view == sfdx.force.test.view",
+          "when": "view == sfdx.force.test.view && viewItem == apexTest",
           "group": "inline"
         }
       ]
@@ -124,6 +129,14 @@
         "icon": {
           "light": "resources/light/document/notRun.svg",
           "dark": "resources/dark/document/notRun.svg"
+        }
+      },
+      {
+        "command": "sfdx.force.test.view.runClassTests",
+        "title": "%run_tests_title%",
+        "icon": {
+          "light": "resources/light/play-side.svg",
+          "dark": "resources/dark/play-side.svg"
         }
       },
       {
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -92,16 +92,23 @@
       "view/item/context": [
         {
           "command": "sfdx.force.test.view.showError",
-          "when": "view == sfdx.force.test.view"
+          "when":
+            "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)_Fail/"
+        },
+        {
+          "command": "sfdx.force.test.view.showDefinition",
+          "when":
+            "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)(_Pass|_Skip|\\b)/"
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -92,16 +92,23 @@
...
         {
           "command": "sfdx.force.test.view.showError",
-          "when": "view == sfdx.force.test.view"
+          "when":
+            "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)_Fail/"
+        },
+        {
+          "command": "sfdx.force.test.view.showDefinition",
+          "when":
+            "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)(_Pass|_Skip|\\b)/"
```
brpowell: VSCode doesn't exactly have a way of dynamically updating the labels for the same command in a context menu. I created another command that runs the same method, but has the "Go to Definition" label. Then conditionally render one command or the other based on the where clause.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -246,6 +246,9 @@ export abstract class TestNode extends vscode.TreeItem {
         dark: DARK_ORANGE_BUTTON
       };
     }
+
+    const contextValue = this.contextValue.split('_');
+    this.contextValue = `${contextValue}_${outcome}`;
```
brpowell: Changed the name of this method to highlight it's doing more than updating an icon now. I'm appending the outcome of a test to the contextValue of the test node so i can access this in the commands' where clauses as defined in the package.json.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -34,11 +35,22 @@ export class ApexTestRunner {
   }
 
   public async showErrorMessage(test: TestNode) {
```
brpowell: When a test class is marked as failed, the "Display Error" option will still show up upon right clicking it. It seemed confusing to have Display Error bring you to the top of the class file, so I changed it to go to the first failed test case.

packages/salesforcedx-vscode-apex/package.json
vazexqi: 1. Yeah, there isn't a way to dynamically update the labels. 
1. Does this mean that you won't be able to see the "Go to Definition" menu if you are on a failing node?
1. What about calling it `goToDefinition`. I think you're really going to the definition more than showing it in-line.

packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -6,6 +6,7 @@
   "refresh_test_title": "Refresh Tests",
   "run_single_test_title": "Run Single Test",
   "show_error_title": "Display Error",
+  "show_definition_title": "Go to Definition",
```
vazexqi: "show_definition_title" -> "go_to_definition_title"?

brpowell: 2. Right, you'll only ever see "Go to Definition" or "Display Error"
3. Yeah I should be consistent, I'll update that.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -33,12 +34,23 @@ export class ApexTestRunner {
     this.eventsEmitter.on('sfdx:update_selection', this.updateSelection);
   }
 
-  public async showErrorMessage(test: TestNode) {
+  public showErrorMessage(test: TestNode) {
```
lcampos: Very cool to get rid of the `async` here.

PR #818, opened 2018-12-14
Authorize orgs against different URLs (production, sandbox, or custom)

brpowell:
### What does this PR do?

When users authorize an org, they now choose to log-in against production, sandbox, or enter a custom URL.

W-5459248


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -54,6 +54,8 @@ export const messages = {
   parameter_gatherer_enter_username_name: 'Enter target username',
   parameter_gatherer_enter_alias_name:
     'Enter an org alias or use default alias',
+  parameter_gatherer_enter_custom_url:
+    'Enter a custom login URL or use default URL',
```
ruthemmanuelle: I assume the default URL will show in the text input box?

brpowell: Right, it shows as the placeholder in the input box similar to the org alias.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -102,8 +108,31 @@ export class ForceAuthWebLoginDemoModeExecutor extends ForceAuthDemoModeExecutor
...
 }
 
-export class AliasGatherer implements ParametersGatherer<Alias> {
-  public async gather(): Promise<CancelResponse | ContinueResponse<Alias>> {
+export class AuthParamsGatherer implements ParametersGatherer<AuthParams> {
+  public async gather(): Promise<
+    CancelResponse | ContinueResponse<AuthParams>
+  > {
+    const orgTypeItems: vscode.QuickPickItem[] = [
+      { label: 'Production', detail: 'login.salesforce.com' },
```
lcampos: These should also be labels in the `i18n` file since this is the text that we show the user while running the command. cc @ruthemmanuelle 

screen shot 2018-12-14 at 2 08 06 pm

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -102,8 +108,31 @@ export class ForceAuthWebLoginDemoModeExecutor extends ForceAuthDemoModeExecutor
...
+      { label: 'Sandbox', detail: 'test.salesforce.com' },
+      { label: 'Custom', detail: 'Use a custom login URL' }
+    ];
+    const selection = await vscode.window.showQuickPick(orgTypeItems);
+    const orgType = (selection as vscode.QuickPickItem).label;
+    let loginUrl: string | undefined;
+    if (orgType === 'Custom') {
```
lcampos: I think we should validate that what the user enters is a `url` or else the command just hangs and the user doesn't receive any error. We might also want to ask the cli team about it cause I never got an error back from the cli when using an invalid url.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
vazexqi: Does this tie into the sfdx-project.json in any way? By that I mean do you read that file first to see what the user has put into the `sfdxLoginUrl` field and make that the default option in the prompt?

It seems that it should at least read what you have provided in the sfdx-project.json (including any custom urls) and not need to override it each time.

What I fear is that the user has already coded some url in sfdx-project.json and now in this flow, you are forcing them to have to type it into the custom field again.

Thoughts?

vazexqi: I think the sfdx-core module might provide facilities for parsing the sfdx-project.json to help with this use case.

@allileong - please keep me honest here.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
ruthemmanuelle: New Salesforce CLI plug-ins built on oclif can validate URLs. https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_plugins.meta/sfdx_cli_plugins/cli_plugins_customize_parameters_types.htm 

Not sure that helps us at the moment, though.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
allileong: Yes, this code has not yet been submitted, but it shows an example of how to read properties from the `sfdx-project.json` using the sfdx-core module.

brpowell: Cool, I think that's a good idea to surface the value in `sfdx-project.json` as the placeholder when you select "Custom"

brpowell: screen shot 2018-12-17 at 2 00 14 pm
screen shot 2018-12-17 at 2 07 29 pm

How do we feel about something like this? @vazexqi

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
brpowell: @lcampos I could be wrong, but this doesn't seem like something we'd have to implement if the CLI handled it as expected by throwing an error. If we still want to implement the validation though I can do so.

lcampos: @brpowell yeah, I'll log a bug on the cli and let's keep this as is.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
vazexqi: @brpowell - This looks good to me. I think using the second screenshot might be better since it's more consistent with the positions of Production and Sandbox, i.e., the URL is on the second line.

I will leave the wording to @ruthemmanuelle.

Thanks for making this change!

ruthemmanuelle: Looks good. Just change "url" to "URL".

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -102,8 +111,90 @@ export class ForceAuthWebLoginDemoModeExecutor extends ForceAuthDemoModeExecutor
...
+export class AuthParamsGatherer implements ParametersGatherer<AuthParams> {
+  public readonly orgTypes = {
+    project: new OrgTypeItem('auth_project_label', 'auth_project_detail'),
+    production: new OrgTypeItem('auth_prod_label', 'auth_prod_detail'),
+    sandbox: new OrgTypeItem('auth_sandbox_label', 'auth_sandbox_detail'),
+    custom: new OrgTypeItem('auth_custom_label', 'auth_custom_detail')
+  };
+
+  public static readonly validateUrl = (url: string) => {
+    const expr = /https?:\/\/(.*)/;
```
lcampos: I see what you did there, this will work for internal envs too 🕺

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -102,8 +111,90 @@ export class ForceAuthWebLoginDemoModeExecutor extends ForceAuthDemoModeExecutor
...
+      return null;
+    }
+    return nls.localize('auth_invalid_url');
+  };
+
+  private getProjectUrl(): string | undefined {
+    let projectUrl: string | undefined;
+    if (workspace.rootPath) {
+      const sfdxProjectJsonFile = path.join(
+        workspace.rootPath,
```
lcampos: `rootPath` is deprecated, lets change this for `workspaceFolders[0]`

packages/salesforcedx-utils-vscode/src/cli/sfdxProjectJsonParser.ts
```diff
@@ -0,0 +1,52 @@
...
+import { SfdxProject } from '@salesforce/core';
+import { JsonArray, JsonMap } from '@salesforce/ts-types';
+
+import * as path from 'path';
+
+export class SfdxProjectJsonParser {
+  public async getValue(workspacePath: string, key: string) {
+    const sfdxProject = await SfdxProject.resolve(workspacePath);
+    const sfdxProjectJson = await sfdxProject.resolveProjectConfig();
+    return sfdxProjectJson[key];
```
lcampos: This is awesome. I tried to break it by deleting `sfdcLoginUrl` on my `sfdx-project.json` file and it turns out that the cli has a default for this type of scenarios.

brpowell: Yeah I was hoping that would be the case and it looks like it is! I'm taking advantage of it when determining to render the "Project Default" option with the org type selection.

Issue #577, opened 2018-08-20
Deploy on Save


ChuckJonas:
**Is your feature request related to a problem? Please describe.**
I'm running the new [change_set_based_tools](https://github.com/forcedotcom/salesforcedx-vscode/wiki/Develop-Against-Any-Org-in-Visual-Studio-Code) and a compile on save is desperately needed.  

**Describe the solution you'd like**
Setting you can add to cause it to run `deploy this source file to org` command on save

**Describe alternatives you've considered**
Could probably use a "Run on save extension" but it would be much nicer to have it supported out of the box

**Internal Work Item**: [W-5552627](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005wp9BIAQ/view)

renatoliveira:
If this ever gets implemented I think it would be a great addition to have the option to **deploy only files that were modified since the last successful push**.

For example: I get three files, and worked on two of them. Instead of deploying all three, or deploying each individual file, the extension can check which files were modified since the last successful deployment (this is important, because if a deployment fails due to a compile error, it would still need to be marked as "changes that we can try to deploy next time").

And of course, this would be great if it were optional as well. Three modes to work with:

1. Deploy everything everytime;
2. Deploy just the single file the user specifies (current);
3. Deploy just the modified files in the project folder.

I don't think there are extensions for this kind of workflow (yet).

toadgeek:
That would be a great improvement!

ChuckJonas:
@renatoliveira that would be great; I imagine it would require some type of data structure to track everything.   I opened #578 as a different solution to a similar problem.

ntotten:
@omniphx You're also welcome to send a PR to add it to the official package if you want. We are aiming to do it pretty soon though if not.

smrobert09:
@renatoliveira @omniphx @ntotten We have an extension that adds the file system tracking. It creates a hash table of the folder structure, detects changes, and allows a 'deploy all changes' type command. 

SFDX creates the hash file automatically on 'force:source:retrieve --manifest' (the sourcePathInfos.json file under .sfdx/orgs/<orgUsername>/sourcePathInfos.json). The problem is that it deletes as soon as the command finishes, making us rebuild it or try to capture/copy it. I know sandboxes aren't source tracking orgs, meaning that we override changes on every push, but if that file were left in place, this would be an easy win. Is this something we can request from the CLI team @ntotten ?

[...]

PR #822, opened 2018-12-14
Easy Deploy on Save

allileong:
### What does this PR do?
When the workspace setting for deploy-on-save is enabled, changes made to files inside of a package directory will trigger an automatic push or deploy, depending on the type of org that is connected to the workspace. 

**Note:** This functionality is not working in Windows due to a problem with the GlobPattern and I have followed the following issue with VS Code: https://github.com/Microsoft/vscode/issues/65240

W-5552627, #577


packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -0,0 +1,192 @@
...
+export enum FileEventType {
+  Create,
+  Change,
+  Delete
+}
+
+const WAIT_TIME_IN_MS = 50;
+
+export async function registerPushOrDeployOnSave() {
+  if (sfdxCoreSettings.getPushOrDeployOnSaveEnabled()) {
```
allileong: We only register the file watcher if the push-or-deploy-on-save setting is already set to true on core activation. This means that if a user wants to turn on the feature, they will have to restart VS Code. 

Disablement of the feature will be respected immediately will not require a restart of the extension.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +0,0 @@
-{
```
lcampos: Why do we need to delete this ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -7,60 +7,34 @@
...
 
 vscode.workspace.onDidChangeTextDocument(e => {
   if (ForceSourceDeployExecutor.errorCollection.has(e.document.uri)) {
     ForceSourceDeployExecutor.errorCollection.delete(e.document.uri);
   }
 });
 
-export class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
-  SelectedPath
+export abstract class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
```
lcampos: Awesome, this makes every type of deploy command render any deploy errors in the problem view.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -0,0 +1,39 @@
...
+  SfdxWorkspaceChecker
+} from './commands';
+import { ForceSourceDeployExecutor } from './forceSourceDeploy';
+
+export class ForceSourceDeployManifestExecutor extends ForceSourceDeployExecutor {
+  public build(manifestPath: string): Command {
+    const commandBuilder = new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_source_deploy_text'))
+      .withArg('force:source:deploy')
+      .withLogName('force_source_deploy')
```
lcampos: Lets change this to `fource_source_deploy_manifest`

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -0,0 +1,69 @@
...
+} from './commands';
+import { ForceSourceDeployExecutor } from './forceSourceDeploy';
+import { SourcePathChecker } from './forceSourceRetrieveSourcePath';
+
+export class ForceSourceDeploySourcePathExecutor extends ForceSourceDeployExecutor {
+  public build(sourcePath: string): Command {
+    const commandBuilder = new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_source_deploy_text'))
+      .withArg('force:source:deploy')
+      .withLogName('force_source_deploy')
```
lcampos: Lets change this to `force_source_deploy_with_sourcepath` to keep consistency with the log ids for the retrieve commands.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -0,0 +1,91 @@
...
+        for (const packagePath of fullPackagePaths) {
+          if (sourcePath.startsWith(packagePath)) {
+            sourcePathIsInPackageDirectory = true;
+            break;
+          }
+        }
+        if (sourcePathIsInPackageDirectory) {
+          return inputs;
+        }
+      } catch (error) {}
```
lcampos: We should log this error and/or send it to our telemetry service.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
allileong: Thanks for catching this. I did not mean to delete this and will put it back.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -0,0 +1,91 @@
...
+          sfdxProjectPath
+        );
+        const fullPackagePaths = packageDirectoryPaths.map(
+          packageDirectoryPath =>
+            path.join(sfdxProjectPath, packageDirectoryPath)
+        );
+
+        let sourcePathIsInPackageDirectory = false;
+        for (const packagePath of fullPackagePaths) {
+          if (sourcePath.startsWith(packagePath)) {
```
lcampos: Wouldn't `packagePath` be something like `force-app` and `sourcePath` be something like `Users/myuser/repo_name/force-app` ?

lcampos: Disregard this, I've checked and I was wrong 👍

PR #829, opened 2018-12-17
Log activation time per extension

lcampos:
### What does this PR do?
Adds logging to track the time it takes every extension to activate. This is the first set of changes on performance enhancements.

@W-5702036@


lcampos:
@brpowell @JimSuplizio yeah the more we add to this the more I feel bad. Initially we did try and have it as an interface that every module could extend but that code was too complex to maintain and test so ended up with this approach. I'll keep this in mind and will look at alternatives next time we modify the metrics code, which will be soon.

Issue #816, opened 2018-12-13
Execute SOQL Query With Selected Text... doesn't work for queries on multiple lines of file


boxfoot:
### Summary

"Execute SOQL query with selected text" does not work for multiline queries

### Steps To Reproduce:

1.  Type this in editor:
        `SELECT Id FROM Contact LIMIT 10`
1.  Select it and run "Execute SOQL Query with selected text..."
1. Works as expected
1. Now repeat with this text:

        SELECT Id 
        FROM Contact
        LIMIT 10

### Expected result

Query executes successfully

### Actual result

ERROR at Row:1:Column:XX
unexpected token: '<EOF>'.

### Additional information

_Feel free to attach a screenshot_.

**VS Code Version**: 1.26.0

**SFDX CLI Version**: sfdx-cli/6.43.1 (windows-x64) node-v11.0.0

**OS and version**: Windows 10

boxfoot:
Note, however, that you can do this without VS Code --

Save this as `query.soql`:

        SELECT Id
        FROM Contact
        LIMIT 10

Then run this command: `sfdx force:data:soql:query --query "$(<query.soql)"`

So the SFDX CLI can handle the multiline query, but the VSCode extension is getting confused by the input...

Thanks for looking into this.

lcampos:
@boxfoot this was released today as part of version `44.15.0` of the extensions. Thanks again for the contribution.

boxfoot:
@lcampos thanks for the update, and for making it easy to contribute.

PR #832, opened 2018-12-18
Allow multiline queries in 'Execute SOQL query'

boxfoot:
### What does this PR do?

Allows selecting a multiline query when using the "Execute SOQL query with selected text" command.

https://github.com/forcedotcom/salesforcedx-vscode/issues/816

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
@@ -63,7 +63,7 @@ export class GetQueryAndApiInputs
         query = document.getText(editor.selection);
       }
     }
-    query = query!.replace('[', '').replace(']', '').replace('\n', ' ');
+    query = query!.replace('[', '').replace(']', '').replace(/(\r\n|\n)/g, " ");
 
     if (!query) {
       return { type: 'CANCEL' };
--- a/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
@@ -63,7 +63,7 @@ export class GetQueryAndApiInputs
         query = document.getText(editor.selection);
       }
     }
-    query = query!.replace('[', '').replace(']', '').replace(/(\r\n|\n)/g, " ");
+    query = query!.replace('[', '').replace(']', '').replace(/(\r\n|\n)/g, ' ');
 
     if (!query) {
       return { type: 'CANCEL' };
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
```diff
@@ -63,7 +63,7 @@ export class GetQueryAndApiInputs
         query = document.getText(editor.selection);
       }
     }
-    query = query!.replace('[', '').replace(']', '');
+    query = query!.replace('[', '').replace(']', '').replace('\n', ' ');
```
vazexqi: What happens on Windows that use \r\n as the line break?

boxfoot: Good question.  I'm developing on windows and it seems to work, but may be because I have my git and editorconfig set to enforce `\n` even on windows.  Easy enough to fix.

```suggestion
    query = query!.replace('[', '').replace(']', '').replace(/(\r\n|\n)/g, " ");
```

packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
```diff
@@ -63,7 +63,7 @@ export class GetQueryAndApiInputs
         query = document.getText(editor.selection);
       }
     }
-    query = query!.replace('[', '').replace(']', '');
+    query = query!.replace('[', '').replace(']', '').replace(/(\r\n|\n)/g, " ");
```
vazexqi: Thanks for making the change.
I think this will fail the linter that we have – did you run `npm run lint`?

We prefer to use single quotes and not double quotes. As you can see from the previous calls to `replace`.

packages/salesforcedx-vscode-core/src/commands/forceDataSoqlQuery.ts
boxfoot: @vazexqi you can see the original submission here had single quotes... not sure why I changed that when I switched to regex!

lcampos:
Thanks @boxfoot for opening the issue and the PR to address it. I'll work on getting this merged, which involves creating a branch with your changes so we can run additional automation. I'll make sure to keep you up to date.

lcampos:
@vazexqi didn't see your post before I did mine. I've created PR #833 to get this merged, just waiting on all the tests to run. 
cc @boxfoot

lcampos:
Closing since #833 is merged, thanks again for the contribution @boxfoot

boxfoot:
Cool, thanks much!  FFR, should I branch in your repo rather than forking?  (Do I even have access to?  I didn't try)

Also - no pressure but I'm curious based on your release schedule when this is likely to be available in my normal VS Code?    Thanks again!

lcampos:
@boxfoot the way you created your PR is correct (fork then send a PR), we need to sort somethings out with the way we run our automation so that we don't have to create an additional PR to be able to run everything.

As for releasing the change, our next release is scheduled for Jan 10th, 2019. We normally release on a weekly basis but that's currently paused because of the holiday season.

PR #865, opened 2019-01-11
[WIP] Reclassify tests

allileong:
### What does this PR do?
* Adds a new category of test (vscode-integration) for tests that require a vscode instance
* Moves strict unit tests into a subdirectory of the `test` directory named `unit` 
* Moves vscode-integration tests out of any directories named `unit`

W-5743308: Correctly classify tests in our current suite


package.json
```diff
@@ -30,6 +30,8 @@
       "lerna exec --concurrency 1 --stream --bail=false -- npm run test:insiders",
     "test:unit":
       "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit",
+    "test:vscode-integration":
```
allileong: I added a new test category: vscode-integration for tests that were previously considered 'unit' tests, but are now considered 'integration' tests since they depend on a running vscode instance.

packages/salesforcedx-utils-vscode/test/unit/cli/forceOrgDisplay.test.ts
```diff
@@ -7,7 +7,7 @@
 
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import { ForceOrgDisplay, OrgInfo, SfdxCommandBuilder } from '../../src/cli';
+import { ForceOrgDisplay, OrgInfo, SfdxCommandBuilder } from '../../../src/cli';
```
allileong: I moved all strict unit tests into a `test/unit` subdirectory within each package.

packages/salesforcedx-vscode-apex-debugger/test/telemetry/telemetry.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 import { assert, match, SinonStub, stub } from 'sinon';
 import TelemetryReporter from 'vscode-extension-telemetry';
-import { TelemetryService } from '../../../src/telemetry/telemetry';
+import { TelemetryService } from '../../src/telemetry/telemetry';
```
allileong: We do not have any strict unit tests in any of our packages with names starting with "salesforcedx-vscode-*" (in other words, our extension packages), so I have moved all tests under the `unit` test subdirectory into the top level `test` directory and removed the `test/unit` directory.

allileong:
Hi @AnanyaJha and @lcampos , please let me know what you think of these changes. If they look ok to you, then I will go ahead and update our documentation to reflect the new test categories.

package.json
AnanyaJha: The test category `vscode-integration` looks good to me. Since these tests depend on a running vscode instance, I believe we actually have two scripts in most of the "salesforcedx-vscode-*" `package.json`s that would fall into this category: `test:insiders:` and the `test:unit` that was renamed. Would it be helpful to rename the `test:insiders` script to also reflect this new category?

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -53,7 +53,8 @@
     "test": "node ../../scripts/run-test-with-top-level-extensions",
```
AnanyaJha: Is there a difference between the `test` specified here and the new `test:vscode-integration` category? Or are they both running the same script?

package.json
allileong: @AnanyaJha, that's a good point. What do you think of renaming the `test:insiders` category to `test:insiders-integration`?

AnanyaJha: Yeah, I think that sounds good!

packages/salesforcedx-vscode-lightning/test/telemetry/telemetry.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 import { assert, match, SinonStub, stub } from 'sinon';
 import TelemetryReporter from 'vscode-extension-telemetry';
-import { TelemetryService } from '../../../src/telemetry/telemetry';
+import { TelemetryService } from '../../src/telemetry/telemetry';
```
lcampos: I thought the telemetry tests didn't require vscode to run since we mocked some of the needed vscode apis.

packages/salesforcedx-vscode-apex/package.json
allileong: I think that it would be nice to keep a script that runs all of the tests (regardless of category) within a package. I could change the `test` script to call all the scripts for the test sub-categories, but I ran into problems with code coverage when I tried that. Let me look into it some more.

Issue #742, opened 2018-11-12
command 'sfdx.force.auth.dev.hub' not found (Other commands fail too)


eltoroit:
### Summary

Error "command 'sfdx.force.auth.dev.hub' not found" in new Windows machine when trying to execute a SFDX command from inside VS Code. Other commands also fail.

I have a brand new Windows 8 machine where I have already installed Java, CLI, VS Code, etc. But when I try to execute the command in VS Code to authorize the Org or other commands then I got an error saying the comand is not found. I am sure I am missing some setting but it's not clear what I need to do. 

In a mac, I had a similar issue and I set the user settings with this "salesforcedx-vscode-apex.java.home": "/Library/Java/JavaVirtualMachines/jdk1.8.0_191.jdk/Contents/Home" but not sure what to do in Windows.

### Steps To Reproduce:

1.  In the command palette, choose a command (> SFDX: Authrorize Dev Hub)  
1.  This message appears immediately.

### Expected result

No errors :-)

### Actual result

An error as shown on the screenshot below

### Additional information

![windowsproblem](https://user-images.githubusercontent.com/7660838/48379084-ecfd9e00-e6a0-11e8-9c75-bd0357965706.png)

**VS Code Version**: 1.28.2

**SFDX CLI Version**: 6.38

**OS and version**: Windows 8

lcampos:
Hi @eltoroit, the configuration that you show in the screenshot looks correct. The error you are seeing usually means there was an error while loading the extensions. Do you get any other error message in the `Debug Console` or `Output` view ?

eltoroit:
Thanks @lcampos for confirming. Knowing that the configuration was correct, I upgraded my old machine to Windows 10, installed all the OS patches, reinstalled Java, CLI and VS Code... now it works fine. It could of have been a configuration error. But I was not sure if the settings in  VS Code, and I kept trying all combinations ;-)

lcampos:
Awesome, glad you got it working. I'll close this issue now.

prateekmathur1991:
@lcampos @JodieM I am also facing the same issue.

My configuration is -
OS - MacOS Mojave 10.14.2
VS Code version - 1.30.1
SFDX CLI version - sfdx-cli/6.44.0-dead729fb5 (darwin-x64) node-v8.9.4

JAVA_HOME and salesforcedx-vscode-apex.java.home settings -

[...]

PR #868, opened 2019-01-13
Handle cli errors when loading workspace context

lcampos:
### What does this PR do?
Fixes current salesforce-vscode-core issue during activation when the cli throws errors like `GenericKeychainServiceError: The service and acount specified in key.json do not match the version of the toolbelt ...`

#742 , [W-5767373](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005jRQwIAM/view)

### DIFF
--- a/packages/salesforcedx-vscode-core/test/context/workspaceOrgType.test.ts
+++ b/packages/salesforcedx-vscode-core/test/context/workspaceOrgType.test.ts
@@ -266,7 +266,6 @@ describe('setupWorkspaceOrgType', () => {
     const getConfigStub = getGetConfigStub(
       new Map([['defaultusername', username]])
     );
-    expect(await getDefaultUsernameOrAlias()).to.equal(username);
 
     const error = new Error();
     error.name = 'GenericKeychainServiceError';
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/test/context/workspaceOrgType.test.ts
```diff
@@ -228,6 +257,28 @@ describe('setupWorkspaceOrgType', () => {
     authInfoCreateStub.restore();
     executeCommandStub.restore();
   });
+
+  it('should set both sfdx:default_username_has_change_tracking and sfdx:default_username_has_no_change_tracking contexts to false for cli config error', async () => {
```
allileong: Won't this error be handled by the default switch case on lines 53 & 54 of workspaceOrgType https://github.com/forcedotcom/salesforcedx-vscode/pull/868/files#diff-6145aff21335781c022072ea6842d8d2R53, in which case `sfdx:default_username_has_change_tracking` and  `sfdx:default_username_has_no_change_tracking` will be set to true, not false?

lcampos: That's correct, not sure why this test is passing. Let me take a look at it.

PR #793, opened 2018-12-05
Bump typescript and @salesforce/core versions

allileong:
* Update typescript to 3.1.6 to match @salesforce/core
* Update @salesforce/core to 1.0.1

PR #872, opened 2019-01-15
Refactor to enable “strictPropertyInitialization" typescript rule

dehru:
### What does this PR do?

After the upgrade to TypeScript 3, the strictPropertyInitialization flag had to be disabled.  This PR enables that flag, and attempts to fix the issues in the source code.

* set tsconfig compiler flag ```strictPropertyInitialization:true```
* initialize the variable depending on how it was used.  
    * If it was string or boolean and seemed to be harmless, I initialized it to empty string or false.  
    * If it had logic that checked for undefined I modify it’s type with ``` string | undefined ```.  
    * For more complex object or service I had to use the definite assignment assertion (!).

Each of these approaches was described here: 
https://blog.mariusschulz.com/2018/05/20/typescript-2-7-strict-property-initialization

Resolves: W-5676805
See: W-5552627, https://github.com/forcedotcom/salesforcedx-vscode/pull/793

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -123,10 +123,10 @@ export interface SetExceptionBreakpointsArguments {
 export class ApexDebugStackFrameInfo {
   public readonly requestId: string;
   public readonly frameNumber: number;
-  public globals: Value[];
-  public statics: Value[];
-  public locals: LocalValue[];
-  public references: Reference[];
+  public globals: Value[] = [];
+  public statics: Value[] = [];
+  public locals: LocalValue[] = [];
+  public references: Reference[] = [];
   constructor(requestId: string, frameNumber: number) {
     this.requestId = requestId;
     this.frameNumber = frameNumber;
@@ -528,10 +528,10 @@ export class MapTupleContainer implements VariableContainer {
 
 export class ApexDebug extends LoggingDebugSession {
   protected myRequestService = new RequestService();
-  protected mySessionService: SessionService;
-  protected myBreakpointService: BreakpointService;
+  protected mySessionService!: SessionService;
+  protected myBreakpointService!: BreakpointService;
   protected myStreamingService = StreamingService.getInstance();
-  protected sfdxProject: string;
+  protected sfdxProject: string = '';
   protected requestThreads: Map<number, string>;
   protected threadId: number;
 
@@ -540,7 +540,7 @@ export class ApexDebug extends LoggingDebugSession {
   protected variableContainerReferenceByApexId = new Map<number, number>();
 
   private static LINEBREAK = `${os.EOL}`;
-  private initializedResponse: DebugProtocol.InitializeResponse;
+  private initializedResponse: DebugProtocol.InitializeResponse | undefined;
 
   private trace: string[] | undefined;
   private traceAll = false;
--- a/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
@@ -99,7 +99,7 @@ export class SessionService {
       new SfdxCommandBuilder()
         .withArg('force:data:record:update')
         .withFlag('--sobjecttype', 'ApexDebuggerSession')
-        .withFlag('--sobjectid', this.sessionId || '')
+        .withFlag('--sobjectid', this.sessionId)
         .withFlag('--values', "Status='Detach'")
         .withArg('--usetoolingapi')
         .withJson()
--- a/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
@@ -13,11 +13,11 @@ import {
 import { RequestService } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
 
 export class SessionService {
-  private userFilter: string;
-  private requestFilter: string;
-  private entryFilter: string;
-  private project: string;
-  private sessionId: string;
+  private userFilter: string = '';
+  private requestFilter: string = '';
+  private entryFilter: string = '';
+  private project: string = '';
+  private sessionId: string = '';
   private connected = false;
   private readonly requestService: RequestService;
 
@@ -99,7 +99,7 @@ export class SessionService {
       new SfdxCommandBuilder()
         .withArg('force:data:record:update')
         .withFlag('--sobjecttype', 'ApexDebuggerSession')
-        .withFlag('--sobjectid', this.sessionId)
+        .withFlag('--sobjectid', this.sessionId || '')
         .withFlag('--values', "Status='Detach'")
         .withArg('--usetoolingapi')
         .withJson()
--- a/packages/salesforcedx-apex-debugger/src/core/streamingService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/streamingService.ts
@@ -18,8 +18,8 @@ export class StreamingService {
   public static DEFAULT_TIMEOUT = 14400;
   private static instance: StreamingService;
   private readonly apiVersion = '41.0';
-  private systemEventClient: StreamingClient;
-  private userEventClient: StreamingClient;
+  private systemEventClient!: StreamingClient;
+  private userEventClient!: StreamingClient;
 
   public static getInstance() {
     if (!StreamingService.instance) {
@@ -100,7 +100,7 @@ export class StreamingService {
     return Promise.resolve(this.isReady());
   }
 
-  private removeTrailingSlashURL(instanceUrl: string) {
+  private removeTrailingSlashURL(instanceUrl = '') {
     return instanceUrl ? instanceUrl.replace(/\/+$/, '') : '';
   }
 
--- a/packages/salesforcedx-apex-debugger/src/core/streamingService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/streamingService.ts
@@ -100,7 +100,7 @@ export class StreamingService {
     return Promise.resolve(this.isReady());
   }
 
-  private removeTrailingSlashURL(instanceUrl = '') {
+  private removeTrailingSlashURL(instanceUrl: string | undefined) {
     return instanceUrl ? instanceUrl.replace(/\/+$/, '') : '';
   }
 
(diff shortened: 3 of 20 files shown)


packages/salesforcedx-apex-debugger/src/core/sessionService.ts
```diff
@@ -13,11 +13,11 @@ import {
 import { RequestService } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
 
 export class SessionService {
-  private userFilter: string;
-  private requestFilter: string;
-  private entryFilter: string;
-  private project: string;
-  private sessionId: string;
+  private userFilter: string = '';
```
vazexqi: Just be careful of this since we might actually check for null or undefined somewhere in the code. And with the current initializations those checks might no longer work.

packages/salesforcedx-apex-debugger/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "outDir": "out",
     "preserveConstEnums": true,
     "strict": true,
-    "strictPropertyInitialization": false
+    "strictPropertyInitialization": true
```
vazexqi: I think we might be able to just remove this setting. The default inherited setting is to be true.

vazexqi: CC: @allileong

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
vazexqi: CC: @DatGuyJonathan

packages/salesforcedx-vscode-core/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "outDir": "out",
     "preserveConstEnums": true,
     "strict": true,
-    "strictPropertyInitialization": false
+    "strictPropertyInitialization": true
```
vazexqi: Ditto for this one.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,12 @@
 {
-  "extensions.autoUpdate": false
+  "extensions.autoUpdate": false,
```
vazexqi: Don't check this in. This should be reverted.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -528,10 +528,10 @@ export class MapTupleContainer implements VariableContainer {
 
 export class ApexDebug extends LoggingDebugSession {
   protected myRequestService = new RequestService();
-  protected mySessionService: SessionService;
-  protected myBreakpointService: BreakpointService;
+  protected mySessionService!: SessionService;
+  protected myBreakpointService!: BreakpointService;
   protected myStreamingService = StreamingService.getInstance();
-  protected sfdxProject: string;
+  protected sfdxProject: string = '';
```
vazexqi: Double-check that we do check that string == null or undefined to see if it is not initialized in the code. Just want to make sure that we don't assume an empty string means that it is set and ready to go.

packages/salesforcedx-apex-debugger/tsconfig.json
dehru: ok.  I'll make that change.

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
dehru: I tried to look around and figure out the uses of these variables.  Some of them are handled in the builder functions like:
```
public withUserFilter(filter?: string): SessionService {
    this.userFilter = filter || '';
    return this;
}
```
but a couple of them did check for undefined elsewhere in the code, and on those I changed their type to:
```
 private _proxyUrl: string | undefined;
```

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
dehru: Ah ok.  It's weird because the file is already checked in, but vscode modified it...it looks like people have been battling this file for a while.  

I can revert it to it's previous state and add it to .gitignore?
Or I could completely remove it and add it to .gitignore and make it go away forever?

Here's it's git history
screen shot 2019-01-15 at 3 30 14 pm

vazexqi: You can add it to .gitignore. 
Don't remove it.

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
vazexqi: Yup. I'm cool with the `protected mySessionService!: SessionService;`. It's just the ones where we initialize the string to `''`  that I need another pair of 👀

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
dehru: The usages of this do not seem to check if this is ever undefined.  ...which is kinda scary.  

ultimately, it ends up here:
```
const execution = new CliCommandExecutor(
      new SfdxCommandBuilder()
        .withArg('force:data:record:create')
        .withFlag('--sobjecttype', 'ApexDebuggerBreakpoint')
        .withFlag(
          '--values',
          `SessionId='${sessionId}' FileName='${typeref}' IsEnabled='true' Type='Exception'`
        )
        .withArg('--usetoolingapi')
        .withJson()
        .build(),
      { cwd: projectPath, env: this.requestService.getEnvVars() }
    ).execute();
```

dehru: another option is for me to use ```string | undefined```.  

But to make Typescript happy when I pass it into functions that take ```string```,  I'll need to pass it like this ```this.sfdxProject || ''```, which brings back to the same issue.

From what I can tell there is no downstream undefined checks.

packages/salesforcedx-apex-debugger/tsconfig.json
dehru: done

packages/salesforcedx-vscode-core/src/statuses/taskView.ts
```diff
@@ -88,9 +88,6 @@ export class TaskViewService implements TreeDataProvider<Task> {
 }
 
 export class Task extends TreeItem {
-  public readonly label: string;
```
vazexqi: Why did we remove this? Was it unused?

packages/salesforcedx-vscode-core/src/statuses/taskView.ts
```diff
@@ -88,9 +88,6 @@ export class TaskViewService implements TreeDataProvider<Task> {
...
-  public readonly collapsibleState: TreeItemCollapsibleState;
```
vazexqi: Why did we remove this? Was it unused?

packages/salesforcedx-apex-debugger/test/unit/core/sessionService.test.ts
```diff
@@ -142,7 +142,7 @@ describe('Debugger session service', () => {
           '{"message":"There was an error", "action":"Try again"}'
         );
         expect(service.isConnected()).to.equal(false);
-        expect(service.getSessionId()).to.an('undefined');
+        expect(service.getSessionId()).to.equal('');
```
vazexqi: @DatGuyJonathan - Could you do a quick sanity check here? Reason for asking is that the unit tests explicitly check for undefined and now we change it to be `''`.

packages/salesforcedx-apex-debugger/src/core/streamingService.ts
```diff
@@ -100,7 +100,7 @@ export class StreamingService {
     return Promise.resolve(this.isReady());
   }
 
-  private removeTrailingSlashURL(instanceUrl: string) {
+  private removeTrailingSlashURL(instanceUrl = '') {
```
vazexqi: I'm not sure why we need this change since the `return instanceUrl ? ...` should check for null | undefined so we don't need a default parameter?

packages/salesforcedx-vscode-core/src/statuses/taskView.ts
dehru: Yeah, these are completely unused.  The TreeItem takes them in the super() constructor.  So I got nervous about instantiating them here, since that might override TreeItem.label and TreeItem.collapsibleState in some way.

packages/salesforcedx-apex-debugger/src/core/streamingService.ts
dehru: This is one of those cases where i couldn't use an empty string as a default, because there is an undefined check on this property.
```
if (instanceUrl) {
      envVars[ENV_SFDX_INSTANCE_URL] = instanceUrl;
    }
```

I'll update the method to take ```string | undefined``` and let it's internals deal with the undefined.
```
private removeTrailingSlashURL(instanceUrl: string | undefined) {
    return instanceUrl ? instanceUrl.replace(/\/+$/, '') : '';
  }
```

dehru: done

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
DatGuyJonathan: This LGTM. It will either be the same valid value as `sfdxProject` in the user's launch.json or the default set in our package.json. The user can already set an empty string as their `sfdxProject` and it just falls back to the default.

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
DatGuyJonathan: The code paths for these session filter strings LGTM.

packages/salesforcedx-apex-debugger/test/unit/core/sessionService.test.ts
DatGuyJonathan: These look appropriate.

packages/salesforcedx-utils-vscode/src/requestService/requestService.ts
```diff
@@ -104,7 +104,7 @@ export class RequestService {
     restHttpMethodEnum: RestHttpMethodEnum = RestHttpMethodEnum.Post
   ): Promise<string> {
     if (this.proxyUrl) {
-      configure(this._proxyUrl, this._proxyStrictSSL);
+      configure(this._proxyUrl || '', this._proxyStrictSSL);
```
DatGuyJonathan: Seems like this should be `this._proxyUrl!` because of the if-check.

dehru:
I'm going to close this pull request, and instead of assigning some strings a default value, or updating their type to be ``` string | undefined ```, I can use ```field!``` definitely assigned or possibly ```?``` to get Typescript 3 compiler to kinda "ignore" the fact that we aren't initializing a value.

PR #873, opened 2019-01-15
Refactor to enable no-floating-promises typescript rule

rvwatch:
### What does this PR do?
- removes "no-floating-promises" key in tslint.json. TSlint should check this by default
- adds await to functions throughout the project that are flagged by tslint
- ignores 2 linter issues that should not have been flagged 

W-5676815 - Refactor non-debugger vscode extensions to conform to "no-floating-promises" tslint rule

### DIFF
--- a/packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
+++ b/packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
@@ -148,9 +148,8 @@ describe('Replay debugger adapter - integration', function() {
         classBPath,
         classBValidLines
       );
-
-      await dc.configurationDoneRequest({});
-
+      // tslint:disable-next-line:no-floating-promises
+      dc.configurationDoneRequest({});
       // Verify stopped on the first line of debug log
       const stackTraceResponse = await dc.assertStoppedLocation('entry', {
         path: logFilePath,
@@ -234,7 +233,8 @@ describe('Replay debugger adapter - integration', function() {
         classStaticVarsAValidLines
       );
 
-      await dc.configurationDoneRequest({});
+      // tslint:disable-next-line:no-floating-promises
+      dc.configurationDoneRequest({});
 
       // Verify stopped on the first line of debug log
       const stackTraceResponse = await dc.assertStoppedLocation('entry', {
@@ -251,6 +251,7 @@ describe('Replay debugger adapter - integration', function() {
         classStaticVarsAPath,
         classStaticVarsAValidLines[0]
       );
+
     } finally {
       const disconnectResponse = await dc.disconnectRequest({});
       expect(disconnectResponse.success).to.equal(true);
--- a/packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
+++ b/packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
@@ -56,7 +56,7 @@ describe('Replay debugger adapter - integration', function() {
 
   after(async () => {
     if (dc) {
-      dc.stop();
+      await dc.stop();
     }
   });
 
@@ -149,7 +149,7 @@ describe('Replay debugger adapter - integration', function() {
         classBValidLines
       );
 
-      dc.configurationDoneRequest({});
+      await dc.configurationDoneRequest({});
 
       // Verify stopped on the first line of debug log
       const stackTraceResponse = await dc.assertStoppedLocation('entry', {
@@ -234,7 +234,7 @@ describe('Replay debugger adapter - integration', function() {
         classStaticVarsAValidLines
       );
 
-      dc.configurationDoneRequest({});
+      await dc.configurationDoneRequest({});
 
       // Verify stopped on the first line of debug log
       const stackTraceResponse = await dc.assertStoppedLocation('entry', {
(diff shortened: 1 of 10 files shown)


packages/salesforcedx-vscode-visualforce/src/extension.ts
```diff
@@ -106,7 +106,7 @@ export async function activate(context: ExtensionContext) {
 
   let disposable = client.start();
   toDispose.push(disposable);
-  client.onReady().then(() => {
+  await client.onReady().then(() => {
```
vazexqi: I'm not sure if this is the right way to do this. This would make it sequential, I think.

tslint.json
```diff
@@ -7,6 +7,6 @@
     "member-ordering": false,
     "no-string-based-set-timeout": false,
     "semicolon": [true, "always", "ignore-bound-class-methods"],
-    "no-floating-promises": false
+    "no-floating-promises": true
```
vazexqi: Shouldn't need to set this to true. The whole line (with the false) can be removed and it will default to true.

CC: @allileong

rvwatch: Just verified this. I'll remove that line in the next commit.

packages/salesforcedx-vscode-visualforce/src/extension.ts
```diff
@@ -106,79 +106,86 @@ export async function activate(context: ExtensionContext) {
...
-            return presentations.map(p => {
-              const presentation = new ColorPresentation(p.label);
-              presentation.textEdit =
-                p.textEdit &&
-                client.protocol2CodeConverter.asTextEdit(p.textEdit);
-              presentation.additionalTextEdits =
-                p.additionalTextEdits &&
-                client.protocol2CodeConverter.asTextEdits(
-                  p.additionalTextEdits
+  client
```
rvwatch: The auto formatter is driving me crazy with this one. I'll get this settled in another commit.

packages/salesforcedx-vscode-visualforce/src/extension.ts
```diff
@@ -106,79 +106,86 @@ export async function activate(context: ExtensionContext) {
...
-      tagRequestor,
-      { visualforce: true },
-      'visualforce.autoClosingTags'
-    );
-    toDispose.push(disposable);
-  });
-
+      toDispose.push(disposable);
+    })
+    .catch(err => {
```
rvwatch: FYI, here is the actual code change to this file.

packages/salesforcedx-vscode-visualforce/src/extension.ts
rvwatch: I've updated this code to use a catch that will send an error to telemetry instead of an await.
```
.catch(err => {
      // Handled by clients
      telemetryService.sendExtensionActivationEvent(err);
    });
```

packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
```diff
@@ -148,9 +148,8 @@ describe('Replay debugger adapter - integration', function() {
         classBPath,
         classBValidLines
       );
-
+      // tslint:disable-next-line:no-floating-promises
```
vazexqi: Could you provide some context on why you choose to ignore the next line?

packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
```diff
@@ -234,6 +233,7 @@ describe('Replay debugger adapter - integration', function() {
         classStaticVarsAValidLines
       );
 
+      // tslint:disable-next-line:no-floating-promises
```
vazexqi: Could you provide some context on why you choose to ignore the next line?

rvwatch: This is the function in question. TS is flagging this as a floating promise. 

```public configurationDoneRequest(
    response: DebugProtocol.ConfigurationDoneResponse,
    args: DebugProtocol.ConfigurationDoneArguments
  ): void {
    if (this.logContext.getLaunchArgs().stopOnEntry) {
      // Stop in the debug log
      this.logContext.updateFrames();
      this.sendEvent(new StoppedEvent('entry', ApexReplayDebug.THREAD_ID));
    } else {
      // Set breakpoints first, then try to continue to the next breakpoint
      this.continueRequest({} as DebugProtocol.ContinueResponse, {
        threadId: ApexReplayDebug.THREAD_ID
      });
    }
  }
```

rvwatch: I've been reading up a bit more on the debug adapter protocol. I think that is what lay at the core of this being seen as a promise by TS. Specifically in the debugadapter-testsupport module: 

> The module provides a toolkit with **_Promise-based_** building blocks for individual protocol requests (e.g. `stepInRequest`) and for common request sequences (e.g. `hitBreakpoint`). These building blocks can be easily configured for a specific adapter and combined to form complex scenarios.

vazexqi: If it is a promise then it should be handled as such. However, for this one, I _think_ I do agree that this might be a misclassification and it's better to just tell tslint to ignore the line.

packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
vazexqi: I still don't know why this line can be ignored safely, other than the fact that this is just in the test so a floating promise is less catastrophic. Could you provide more details.

PR #878, opened 2019-01-16
Enables Typescript Compiler flag for strictPropertyInitialization

dehru:
### What does this PR do?

Since the move to TS 3, properties are required to be initialized or specific market as type | undefined.  So in order to get around some existing code that did not initialize properties, this compiler option was turned off.  This PR fixes the existing code and re-enables the strictPropertyInitialization for the compiler.

* Enables Typescript Compiler flag for strictPropertyInitialization
* The absolute minimal implementation changes with those properties that need initialization
* Uses these constructs:
    * the ```field!``` - "definitely assigned" marker on the field...er...leave us alone we've got this
    * the ```field?``` - possibly undefined, shortcut for ``` type | undefined ```
    * or in just a couple cases, setting boolean to false, number to a default constant, array to [], or function to empty () => {}

Resolves: W-5676805
See: W-5552627, https://github.com/forcedotcom/salesforcedx-vscode/pull/793

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
@@ -99,7 +99,7 @@ export class SessionService {
       new SfdxCommandBuilder()
         .withArg('force:data:record:update')
         .withFlag('--sobjecttype', 'ApexDebuggerSession')
-        .withFlag('--sobjectid', this.sessionId || '')
+        .withFlag('--sobjectid', this.sessionId)
         .withFlag('--values', "Status='Detach'")
         .withArg('--usetoolingapi')
         .withJson()
--- a/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
+++ b/packages/salesforcedx-apex-debugger/src/core/sessionService.ts
@@ -13,11 +13,11 @@ import {
 import { RequestService } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
 
 export class SessionService {
-  private userFilter: string;
-  private requestFilter: string;
-  private entryFilter: string;
-  private project: string;
-  private sessionId: string;
+  private userFilter: string = '';
+  private requestFilter: string = '';
+  private entryFilter: string = '';
+  private project: string = '';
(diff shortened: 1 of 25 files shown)


packages/salesforcedx-apex-debugger/src/core/sessionService.ts
```diff
@@ -13,11 +13,11 @@ import {
 import { RequestService } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
 
 export class SessionService {
-  private userFilter: string;
-  private requestFilter: string;
-  private entryFilter: string;
-  private project: string;
-  private sessionId: string;
+  private userFilter?: string;
```
vazexqi: @dehru - Were you able to check what @ntotten said and whether our tslint rules could have suggested to use `userFilter?` instead of `userFildter: string | undefined`?

packages/salesforcedx-apex-debugger/src/core/streamingService.ts
```diff
@@ -18,8 +18,8 @@ export class StreamingService {
   public static DEFAULT_TIMEOUT = 14400;
   private static instance: StreamingService;
   private readonly apiVersion = '41.0';
```
vazexqi: @lcampos @DatGuyJonathan - Should we bump this up to the latest API (not in this PR but in a future one?)

packages/salesforcedx-apex-debugger/src/core/sessionService.ts
dehru: I do not see that option for tslint in the core rules.

vazexqi:
@allileong and @lcampos - Could you take a quick look at this when you get the chance? @DatGuyJonathan and I have already reviewed. Would like to get this in soon so you guys can take advantage of more TypeScript 3 goodness.

lcampos:
@vazexqi @dehru I looked at this and ran some quick tests and it all looks good. The appveyor run was also successful so not sure why it didn't get updated here. I'll merge this today.

dehru:
Thanks for giving me an easy one!

PR #879, opened 2019-01-17
Fix debug log filename being set to '07Lxxxxxxxxxxxxxxx_Invalid date.…

lcampos:
…log'

### What does this PR do?
Fixes the Debug Log filename containing `Invalid date` as part of it. This was introduced was part of the fix to enrich the Debug Log list data.

@W-5714727@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
@@ -116,6 +116,7 @@ export type ApexDebugLogObject = {
 
 interface ApexDebugLogItem extends vscode.QuickPickItem {
   id: string;
+  startTime: string;
 }
 
 export class LogFileSelector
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -131,6 +132,9 @@ export class LogFileSelector
         return {
           id: logInfo.Id,
           label: icon + logInfo.LogUser.Name + ' - ' + logInfo.Operation,
+          startTime: moment(new Date(logInfo.StartTime)).format(
```
allileong: Does adding this cause the startTime to now appear twice in the quickpick dropdown?

AnanyaJha: I had the same question as well

lcampos: No, this is because the quickpick dropdown only prints the info in the default fields of the QuickPickItem interface and ignores the ones we add to it when we extend it as part of `ApexDebugLogItem`.

screen shot 2019-01-17 at 9 52 15 am

allileong: Gotcha, thanks. LGTM 👍

PR #882, opened 2019-01-17
Add Vscode-Integration Test Category

allileong:
### What does this PR do?
With this change we now have the following test categories:

**Non-system-tests (tests in all packages except for the `system-tests` package):**
* test:unit - strict unit tests that must be placed in the `<package-name>/test/unit` directory
* test:integration - integration tests (not requiring a vscode instance) that must be placed in the `<package-name>/test/integration` directory
* test:vscode-integration - integration tests requiring a vscode instance that **must be placed in the `<package-name>/test/vscode-integration` directory**
* test:vscode-insiders-integration - runs the vscode-integration tests, but using vscode-insiders rather than vscode

**System-tests (tests in the `system-tests` package):**
* test:system-tests - all system tests in the `system-tests` package
* test:vscode-insiders-system-tests - all system tests in the `system-tests` package


allileong:
Hi @lcampos and @AnanyaJha, I've created a new pr for the test scripts work because I have made several more changes since the last PR. There's a lot going on here, so please let me know if you'd like to have a meeting to discuss the changes. Similar to my last pr, I'd like to get your approval on the new structure before I update our documentation.

scripts/run-vscode-integration-tests.js
```diff
@@ -0,0 +1,16 @@
+const shell = require('shelljs');
+shell.set('-e');
+shell.set('+v');
+
+const path = require('path');
+const cwd = process.cwd();
+
+// Executes the tests in the out/test/vscode-integration directory
+shell.exec(
+  `cross-env CODE_TESTS_PATH='${path.join(
```
allileong: @lcampos - I was wrong when I mentioned to you in Bellevue that the `vscode/bin/test` script necessarily runs all tests inside of the `out/test` directory. If I set the CODE_TESTS_PATH environment variable, I can override that default, and scope the tests being run to the `out/test/vscode-integration` directory instead. This is great because we can now have unit tests, integration tests, and vscode-integration test all inside of the `test` directory and still be able to run them separately.

lcampos: Very cool 👍

lcampos:
This looks really good. What else are you missing if you've added the shelljs scripts, the package.json scripts and moved the tests to the correct folder ?

allileong:
@lcampos cool, thanks for the feedback. I was waiting to update the doc (https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/docs/tests.md) in case any changes were needed.

packages/salesforcedx-slds-linter/package.json
```diff
@@ -35,8 +36,11 @@
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "test": "node ./node_modules/vscode/bin/test",
-    "test:unit": "node ./node_modules/vscode/bin/test"
+    "test": "npm run test:vscode-integration",
```
AnanyaJha: Question - is the `test` script meant to run all the tests in each package? ie. in this one, should it be running the `vscode-integration` and `vscode-insiders-integration` scripts?

AnanyaJha:
I had one question, but otherwise the changes look great!

packages/salesforcedx-slds-linter/package.json
allileong: @AnanyaJha That's a good point. I've been considering VS Code (and not Insiders) to be the default, but don't have a great reason for doing that. The only caveat I have for combining the two scripts is that before running `npm run test` locally, users will have to quit out of both VS Code and Insiders since the tests fail when another instance of the same editor is open in another window. This isn't really a problem, though, since the scripts to only run the test against VS Code or Insiders exist. 

I'll add `npm run test:vscode-insiders-integration` to the `test` scripts.

allileong: We've decided to combine the `vscode-integration` and `vscode-insiders-integration` scripts under the `test` script, but not until this work (W-4889431) is done to break up our autobuild into multiple stages. Combining the scripts now would effectively add the vscode-insiders-integration tests to our current build, which would make the build time significantly longer.

allileong:
@lcampos & @AnanyaJha, I added changes to `.vscode/launch.json` and `docs/tests.md`. This PR is ready for official review.

PR #890, opened 2019-01-21
Org Picker UI

AnanyaJha:
## What does this PR do?
This PR creates an org picker allowing users to `Authorize an Org`, `Create a Scratch Org`, or switch to a previously authenticated org . It shows the alias if it exists, otherwise defaults to the username. It displays scratch orgs associated with the current `defaultdevhub` and hides all others. If no default dev hub is specified, it gives the user the option to set a default dev hub. 

 @W-5459244@


package.json
```diff
@@ -16,41 +16,29 @@
     "vsce": "^1.36.3"
   },
   "scripts": {
-    "postinstall":
-      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
-    "bootstrap":
-      "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
+    "postinstall": "lerna bootstrap -- --no-package-lock && node scripts/reformat-with-prettier",
```
AnanyaJha: These were accidental changes, I'll edit them back.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+    return authInfoObjects;
+  }
+
+  public async filterAuthInfo(authInfoObjects: FileInfo[]) {
+    authInfoObjects = authInfoObjects.filter(fileData =>
+      isNullOrUndefined(fileData.scratchAdminUsername)
+    );
+    authInfoObjects = authInfoObjects.filter(fileData =>
+      isNullOrUndefined(fileData.isDevHub)
```
AnanyaJha: This filters out dev hubs from appearing in the dropdown list for orgs, because we only display the `defaultusername` field in the status bar. I can see why including the dev hubs would be useful as well if a developer wants to create a scratch org in a different dev hub.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -469,6 +482,10 @@ export async function activate(context: vscode.ExtensionContext) {
   await setupWorkspaceOrgType();
   registerDefaultUsernameWatcher(context);
 
+  // Set display for defaultusername org
+  showOrg();
+  // displayDefaultUsername();
```
lcampos: Remove commented out code.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -222,5 +222,8 @@ export const messages = {
...
   error_source_path_not_in_package_directory_text:
-    'Error deploying or retrieving source: The file or directory that you tried to deploy or retrieve isn\'t in a package directory that\'s specified in your sfdx-project.json file. Add this location to your "packageDirectories" value, or deploy or retrieve a different file or directory. For details about sfdx-project.json, see: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm'
+    'Error deploying or retrieving source: The file or directory that you tried to deploy or retrieve isn\'t in a package directory that\'s specified in your sfdx-project.json file. Add this location to your "packageDirectories" value, or deploy or retrieve a different file or directory. For details about sfdx-project.json, see: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm',
+  error_retrieving_previous_authorizations:
+    'Error retrieving previous authorizations. Run "SFDX: Authorize an Org" or "SFDX: Create a Default Scratch Org"',
+  missing_default_org: 'No Default Org Set'
 };
```
AnanyaJha: @ruthemmanuelle Hi Ruth, this text will show up if the user has no previous authorizations set and tries to access the authorization list. Let me know if you'd like any changes, thanks!

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
...
+  const editor = window.activeTextEditor;
+  if (!editor) {
+    statusBarItem.hide();
+    return;
+  }
+  displayDefaultUsername();
+}
+
+export async function displayDefaultUsername() {
+  const defaultUsernameorAlias = await getDefaultUsernameOrAlias();
```
AnanyaJha: This method will return the global `defaultusername` when one cannot be found for the current project. Because of this, there is a situation in which the user has run `SFDX: Log Out from All Authorized Orgs` so there are no auth files available but the `sfdx-config.json` file does not remove/update the global 'defaultusername.' In that case, the user will see a default in the status bar, but when they click on it they will receive an error. If we'd rather that not happen, I can also change it to only refer to the local config file for the `defaultusername` info.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -0,0 +1,40 @@
...
+  public constructor(arg: string) {
+    super();
+    this.arg = arg;
+  }
+
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription('SFDX: Set a Default Org')
+      .withArg('force:config:set')
+      .withArg('defaultusername=' + this.arg)
```
lcampos: Let's change this from string concatenation to template literals (`defaultusername=${this.arg}`)

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -0,0 +1,40 @@
...
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+export class ForceConfigSetExecutor extends SfdxCommandletExecutor<{}> {
+  private arg: string;
+
+  public constructor(arg: string) {
```
lcampos: Can we change `arg` to a more meaningful name e.g. username, defaultUsername, etc ?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
+import { Aliases, AuthInfo } from '@salesforce/core';
+import {
+  CancelResponse,
+  ContinueResponse,
+  ParametersGatherer
```
lcampos: ParametersGatherer is not used, lets remove it.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
+import { Aliases, AuthInfo } from '@salesforce/core';
```
lcampos: When we create a new module like this we normally expose the methods using an index.ts file. Look at how we do it for other features like diagnostics, notifications, etc.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
...
+      return Promise.reject(
+        new Error(nls.localize('error_retrieving_previous_authorizations'))
+      );
+    }
+    const authInfoObjects = Promise.all(
+      authFilesArray.map(fileName => {
+        const filePath = path.join(os.homedir(), '.sfdx', fileName);
+        return readFileAsync(filePath, 'utf8')
+          .then(fileData => JSON.parse(fileData))
+          .catch(err => null);
```
lcampos: We should log this error instead of not doing anything with it.

lcampos:
We need to provide a brief description on what the user needs to do (1. on the screenshot attached below).
I think we should add something that distinguishes the `SFDX` commands from the org options (2.). For example, the git command add a `+` sign to distinguish between creating a new branch and selecting an existing branch. 

current_quickpick_highlight

lcampos:
In a scenario where a user has a clean sfdx cli (has no org auths) and opens vscode we show `No Default Org Set` in the footer (1. on the screenshot attached below). If they then click on this, we show an error message in the bottom-right corner (2. screenshot).
cli_without_credentials_highlight
 
Instead of the current behavior, I suggest we show the command palette with exclusively the `SFDX` commands for authorizing an org or creating a scratch org.

cc @ntotten

lcampos:
The current feature is designed to exclusively switch defaultusername (sfdx force:config:set defaultusername=<alias or username>). This will be a limitation for users that have multiple dev hubs or for projects where the default dev hub has not been set. How do we want to address those scenarios @ntotten  ?

Should we add `SFDX: Authorize a Dev Hub` to the command options we show in the command palette if there's no default dev hub ?

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
AnanyaJha: Yeah, sounds good.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
AnanyaJha: Yup!

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
AnanyaJha: Ah okay, got it

AnanyaJha:
> The current feature is designed to exclusively switch defaultusername (sfdx force:config:set defaultusername=). This will be a limitation for users that have multiple dev hubs or for projects where the default dev hub has not been set. How do we want to address those scenarios @ntotten ?
> 
> Should we add `SFDX: Authorize a Dev Hub` to the command options we show in the command palette if there's no default dev hub ?

I think it would make sense to include that command as an option and to display the dev hubs as part of the dropdown list as well. The only thing that could be confusing is that the status bar on the bottom only displays the `defaultusername` not the `defaultdevhubusername`. So even if the user selects a new default dev hub, the bar on the bottom would currently not be updated to reflect that.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,125 @@
...
+      })
+    );
+    return authInfoObjects;
+  }
+
+  public async filterAuthInfo(authInfoObjects: FileInfo[]) {
+    authInfoObjects = authInfoObjects.filter(fileData =>
+      isNullOrUndefined(fileData.scratchAdminUsername)
```
allileong: What does it mean if an authInfo has a scratchAdminUsername field?

AnanyaJha: That field indicates that the file represents a new user, other than the default admin, within the org.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -0,0 +1,40 @@
...
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+export class ForceConfigSetExecutor extends SfdxCommandletExecutor<{}> {
```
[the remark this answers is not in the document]
AnanyaJha: Yeah I can definitely do that, I think we're still waiting back on hearing whether we should have the ability to also set the `defaultdevhubusername` field through this ui as well. In that case, I was planning on having this executor handle both scenarios, would the rename still be a good idea?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Hi @AnanyaJha . How will these two new messages be displayed?

AnanyaJha: The `org_select_text` will be the placeholder in the text box for the drop down, which allows them to select an org or scratch org. The `missing_default_org` will display in the bottom status bar when there is no default org set for the project.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -0,0 +1,46 @@
...
+  private usernameOrAlias: string;
+
+  public constructor(usernameOrAlias: string) {
+    super();
+    this.usernameOrAlias = usernameOrAlias;
+  }
+
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription('SFDX: Set a Default Org')
```
lcampos: This text needs to be a label (e.g. `nls.localize('label_name')`).

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+import { Aliases, AuthInfo } from '@salesforce/core';
+import {
+  CancelResponse,
+  ContinueResponse
+} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import * as os from 'os';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import fs = require('fs');
```
lcampos: Lets change this to `import { readFile } from 'fs';`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+const readFileAsync = promisify(fs.readFile);
+export class OrgList {
+  public async getAuthInfoObjects() {
+    const authFilesArray = await AuthInfo.listAllAuthFiles().catch(err => null);
+
+    if (authFilesArray === null || authFilesArray.length === 0) {
+      return null;
+    }
+    const authInfoObjects = Promise.all(
```
lcampos: We can change this from being asynchronous to synchronous if we use `fs.readFileSync`. What's the reason for having this as async ?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,139 @@
...
+    vscode.commands.executeCommand('sfdx.force.org.create');
+    return { type: 'CONTINUE', data: {} };
+  } else {
+    const usernameOrAlias = selection.split(' ', 1);
+    vscode.commands.executeCommand('sfdx.force.config.set', usernameOrAlias);
+    return { type: 'CONTINUE', data: {} };
+  }
+}
+
+export async function showOrg() {
```
lcampos: Should probably change this from `showOrg` to `showOrgListQuickPick` or something like that. Right now it can get confused by the method in `scratch-org-decorator.ts` which is also used in `salesforcedx-vscode-core/src/index.ts`

packages/salesforcedx-vscode-core/src/orgPicker/index.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export {
+  OrgList,
+  FileInfo,
```
lcampos: We're not using `FileInfo` outside of `orgList.ts`, lets just export the things that are used outside of `orgList.ts`.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -593,6 +593,10 @@
       {
         "command": "sfdx.force.source.delete.current.file",
         "title": "%force_source_delete_this_source_text%"
+      },
+      {
+        "command": "sfdx.force.set.default.org",
+        "title": "SFDX: Set Default Org"
```
lcampos: This title text needs to be a label.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  const selection = await vscode.window.showQuickPick(quickPickList, {
+    placeHolder: nls.localize('org_select_text')
+  });
+
+  if (!selection) {
+    return { type: 'CANCEL' };
+  }
+  if (
```
lcampos: We could change this to a switch statement but other than that it looks good.

AnanyaJha: Cool, thanks. I can change it to a switch statement and merge it in.

Issue #883, opened 2019-01-18
Overzealous deploy on save


MariuszAndrzejewski:
### Summary

Save on deploy triggers separately for any changed or created files in package directory for projects with manifest. Deploy on save tries to deploy even filest that are not salesforce metadata.

What is worst retrieving source or changing branches in git makes extension attempt to deploy all changed sources, running deploy separately for each. If there are large differences between branches separate deploys may not compile and even freeze visual studio code.

This makes this feature pretty much unusable when working with source control.

### Steps To Reproduce:

1. Set "salesforcedx-vscode-core.push-or-deploy-on-save.enabled" to true
2. Create new project with manifest or open existing and authorize an org.
3. Setup git repository for project
4. Make any code changes and commit
5. Create new branch and checkout.
6. Make code changes in different files and commit
7. Checkout between two branches.
8. Observe deploy source running for all files.

OR

1. Set "salesforcedx-vscode-core.push-or-deploy-on-save.enabled" to true
2. Create new project with manifest and authorize an org.
3. Right click on package.xml and choose SFDX: Retrieve source in manifest for org.
4. Observe deploy source running for changed files.

OR

1. Set "salesforcedx-vscode-core.push-or-deploy-on-save.enabled" to true
2. Create new project with manifest or open existing and authorize an org.
3. Create new empty text file in project package directory
4. Observe deploy source attempting to deploy file even if it is not salesforce source file.

OR

1. Set "salesforcedx-vscode-core.push-or-deploy-on-save.enabled" to true
2. Setup projects with manifest
3. Copy any sources into project package directory.
4. Observe deploy source running for all files.

### Actual result

Deploy source to org is executed on any file modification.

### Expected result

Deploy source to org should only execute when actively saving file in visual studio code.

### Additional information

**VS Code Version**:
1.30.2
**SFDX CLI Version**:
6.5.0
**OS and version**:
Ubuntu 18.10

lcampos:
Thanks for the feedback @MariuszAndrzejewski, we're currently working on some enhancements related to that feature and will include this as part of that work.

ChuckJonas:
Wow... I wish I had seen this issue BEFORE doing a similar steps.... I basically just nuked and entire UAT ORG.  WTF...

ntotten:
Sorry about this everyone. We ha e a fix coming out next week per the above PR. ^

steals:
[...]

PR #895, opened 2019-01-23
Fix deploy-on-save feature to deploy on save event only

allileong:
### What does this PR do?
Previously, deploys were being triggered by:
* File creation
* File deletion
* File updates triggered by source retrieves and branch changes

With this change, the deploy on save feature will only deploy source that has been explicitly saved, and will group all changes made within 3 seconds in a single deploy. 

W-5777512 (https://github.com/forcedotcom/salesforcedx-vscode/issues/883)
W-5756483


packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -9,99 +9,55 @@ import { channelService } from '../channels';
...
 import * as vscode from 'vscode';
 
-export enum FileEventType {
-  Create,
-  Change,
-  Delete
-}
-
-const WAIT_TIME_IN_MS = 50;
+const WAIT_TIME_IN_MS = 4500;
```
allileong: I played around with different wait times and settled on 4500ms, but I am very open to suggestions for other wait times. In my own development, I don't tend to make a lot of rapid changes within 4.5 seconds of each other, so my way of development will still trigger many deploys, but when I extended the wait time much longer than 4.5 seconds, the time between my last change and the deployment felt like a noticeable lag.

ntotten: Sounds good. We could consider making it configurable if it’s not too difficult, but 4500 seems like a good default.

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -176,8 +105,23 @@ function displayError(message: string) {
   channelService.showChannelOutput();
 }
 
-function ignorePath(uri: vscode.Uri) {
-  return isDotFile(uri) || isDirectory(uri);
+function ignorePath(uri: vscode.Uri, packageDirectories: string[]) {
+  return (
+    isDotFile(uri) ||
+    isDirectory(uri) ||
```
lcampos: Did you try using uri.scheme === file in `isDirectory` method ? It would be faster if we didn't have to use `fs`.

Issue #803, opened 2018-12-07
Automatically perform 'SObject Definitions Refresh' after 'Retrieve Source from Org'


SalesforceBobLightning:
I would like to suggest that the SalesforceDX VS Code extension automatically performs the 'SObject Definitions Refresh' command after the user performs the 'Retrieve Source from Org' command, because you need the objects refreshed to be able to develop against them.

What are the reasons for not doing this?

renatoliveira:
I can think of some reasons it does not work this way today:

1. One does not necessarily need to refresh the definitions every time they retrieve the source from the organization.
2. It can take a long time on big projects. This is especially troublesome if people are still developing against a single sandbox (this is, AFAIK, pretty common for big pre-DX projects).

This does not mean I don't think this is a good idea. It really can help. But I think this can be **optional**. Maybe a `refreshDefinitionsOnRetrieve` flag on the project's configuration file.

SalesforceBobLightning:
A flag sounds like a good solution.

What would the ETA on that be? :)

lcampos:
Hi @SalesforceBobLightning, thanks for the feedback. I agree with you that this would make the overall experience better. Just to provide more context, the reason we didn't tackle it before were performance and priority. 

For performance, it's in line with what @renatoliveira already pointed out but we have some ideas on how to make it work in a more performant way. For priority, we wanted to focus on providing more value in the Apex LSP so at the time we focused on delivering Find All References, Go to Definition, etc. Now that we're working in Apex Rename Refactor this has come up internally as one of the things to tackle and it's currently in our queue. I'll make sure to update this thread once we have something close to be released.

ntotten:
SObject refresh now happens on first load of a project. We don't pull on every refresh at this time, but its something being considered.

jisaacs1:
I would propose you only refresh the definitions of SObjects included in the pull (CustomObject or their CustomFields). That would reduce the performance impact on pulls and greatly reduce if not eliminate the need for full SObject refreshes after the first load of a project.

PR #899, opened 2019-01-24
[WIP] Automatically refresh SObject definitions with increased performance

brpowell:
### What does this PR do?
Currently, updating SObject definitions for code completion is a manual process and involves a long running describe call that fetches every SObject. These changes now monitor the local filesystem for changes to objects and regenerate faux classes on the fly to instantly update code completion. In certain cases where callouts to the org are necessary, the describe call now only makes callouts for the specific object(s) that need updating to reduce wait time. 

W-5681053
#803 

I am still reorganizing code, looking for bugs, and need to write tests. Any early feedback would be helpful for all three of these things!


packages/salesforcedx-vscode-core/src/generator/objectFileWatcher.ts
```diff
@@ -0,0 +1,155 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { channelService } from '../channels';
+import { nls } from '../messages';
+import { notificationService } from '../notifications';
+import { SfdxProjectJsonParser } from '../util';
+
+const WAIT_TIME_IN_MS = 50;
+
```
brpowell: Thinking about generalizing and putting source file watcher stuff into a util for future use

packages/salesforcedx-vscode-core/src/generator/objectFileWatcher.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (remoteRefreshObjects.size > 0) {
+      vscode.commands.executeCommand(
+        'sfdx.force.internal.refreshsobjects',
+        Array.from(remoteRefreshObjects)
+      );
+    }
+  }
+}
+
+export async function getPackageDirectoriesRelativePattern(): Promise<
```
brpowell: This too into a util. I copied it from Alli's deploy on save functionality ;)

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -14,41 +14,41 @@ import { xhr, XHROptions, XHRResponse } from 'request-light';
...
-  activateable: boolean;
-  childRelationships: ChildRelationship[];
-  compactLayoutable: boolean;
-  createable: boolean;
-  custom: boolean;
-  customSetting: boolean;
-  deletable: boolean;
-  deprecatedAndHidden: boolean;
-  feedEnabled: boolean;
+  actionOverrides?: any[];
```
vazexqi: Could you elaborate on the reasoning for making these optional?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -61,7 +50,16 @@ export class FauxClassGenerator {
     // TBD what are these mapped to and how to create them
     // ['calculated', 'xxx'],
     // ['masterrecord', 'xxx'],
-    ['complexvalue', 'Object']
+    ['complexvalue', 'Object'],
```
vazexqi: How did you arrive at the extensive list of all the different types?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGeneratorUtil.ts
```diff
@@ -0,0 +1,90 @@
...
+  private static stripId(name: string): string {
+    if (name.endsWith('Id')) {
+      return name.slice(0, name.length - 2);
+    } else {
+      return name;
+    }
+  }
+
+  private static capitalize(input: string): string {
+    return input.charAt(0).toUpperCase() + input.slice(1);
```
vazexqi: Does this work with unicode or the different encoding that we support? Just in case the user uses a different character set, what does this do?

Talk to @JimSuplizio. I think he knows more about unicode and the strange behaviors of upper casing certain things. We want to be careful on our assumptions here.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
brpowell: I wanted to take advantage of a lot of the existing generator code that was built around the Field interface. When creating Field objects during local generation, we don't have all of the extra information that comes from a describe call. I don't see the purpose of having a lot of these fields as required if so many of them aren't actually being used by the generation process.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGeneratorUtil.ts
brpowell: Hm I'm not totally sure, but this was code that was already present in the generator class I just moved it to another file.

vazexqi:
@brpowell - Remember to add new copyright headers to new files.

I have some general comments about this feature (based on my experiences with the faux generator):

1. Don't assume that the custom object file that you have is always valid XML. The user could be typing halfway. The user (or something external like a git merge) could change the file, resulting in a state that is not parseable. So you need to be resilient to that.
1. Be careful with the file watchers. Having too many can be expensive. 
1. Be careful with file watchers on directories that can be deleted (and handle those properly). Same with files (sometimes Windows might not allow you to delete the file if it is opened in VS Code).
1. Be careful of bursts of file changes since you need to process all of them.
1. It is _extremely_ helpful to log errors. The customer is going to inadvertently have some XML that we never anticipated. It would be good to log it in a way that the customer can easily send to us for debugging purposes.

I just want to be careful about this feature since you are dealing now with many moving parts now that can be affected. So the processing needs to be resilient.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
brpowell: Referencing: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_field_types.htm#meta_type_fieldtype

Still need to handle:
- MetadataRelationship
- IndirectLookup
- Hierarchy
- File
- CustomDataType
- Html

Do you know what the status of these are?

PR #986, opened 2019-01-30
Automatically refresh SObject definitions on extension activation if there are no faux classes

brpowell:
### What does this PR do?

When the core extension first activates, it will check to see if the current project has initially generated any faux classes. If not, the definitions will refresh automatically in the background with progress shown in the status bar.

W-5754599


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -493,6 +494,10 @@ export async function activate(context: vscode.ExtensionContext) {
     decorators.showDemoMode();
   }
 
+  // Refresh SObject definitions if there aren't any faux classes
+  const projectPath = vscode.workspace!.workspaceFolders![0].uri.fsPath;
+  await initSObjectDefinitions(projectPath);
```
brpowell: FYI this does not mean wait until the definitions have been completely refreshed, it's just waiting until executeCommand was called (or not called) in the function.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -42,10 +78,32 @@ class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}> {
...
+    notificationService.reportCommandExecutionStatus(
+      execution,
+      cancellationToken
+    );
+
+    let progressLocation = vscode.ProgressLocation.Notification;
+    const refreshSource = response.data;
+    if (refreshSource !== SObjectRefreshSource.MANUAL) {
+      progressLocation = vscode.ProgressLocation.Window;
+      telemetryService.sendAutomaticSObjectRefreshEvent(refreshSource);
```
brpowell: Is this okay as a separate telemetry event?

packages/salesforcedx-vscode-core/src/index.ts
lcampos: Do we need to do wait for that ? `initSObjectDefinitions` currently runs ForceConfigGet (via `workspaceOrgType`) which is running `sfdx force:config:get` which can be very slow depending on how many credentials you have in your local cli.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
lcampos: Specific to telemetry, I think we should use `this.logMetric` and modify the command text we send when this is triggered outside of the command palette (e.g. startup, etc). Because this will currently cause 2 different metrics for one execution (here and on line 121).

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -61,18 +119,40 @@ class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}> {
...
+    (await getDefaultUsernameOrAlias()) !== undefined;
+  if (projectPath && hasDefaultUsernameSet) {
+    const sobjectFolder = path.join(
+      projectPath,
+      SFDX_DIR,
+      TOOLS_DIR,
+      SOBJECTS_DIR
+    );
+    if (!fs.existsSync(sobjectFolder)) {
+      vscode.commands.executeCommand(
```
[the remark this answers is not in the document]
brpowell: In the execute method of the command there's stuff like handling the channel output, prevention of running a refresh while one is already active, the notification service, task view service and showing progress that I figured could be leveraged rather than duplicated. Is there a reason to not do it this way?

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -7,6 +7,7 @@
 
 import * as util from 'util';
 import vscode = require('vscode');
+import { SObjectRefreshSource } from '../commands/forceGenerateFauxClasses';
```
lcampos: We don't need this enum here, let's remove it.

packages/salesforcedx-vscode-core/test/vscode-integration/telemetry/index.test.ts
```diff
@@ -7,6 +7,7 @@
 import { expect } from 'chai';
 import { assert, match, SinonStub, stub } from 'sinon';
 import { window } from 'vscode';
+import { SObjectRefreshSource } from '../../../src/commands/forceGenerateFauxClasses';
```
lcampos: Same as above, let's remove it this enum.

amitkumarj1979:
Is there a way i can disable the refresh ? I have a big sandbox with almost 60% apex code usage, Aura components and also 100+ sobjects with some of them having 500+ fields. When the classes are generated the Java process takes up almost 5GB memory and the Apex Language server does not work and crashes. If I delete the faux classes the memory usage goes down  2 GB.

lcampos:
Hi @amitkumarj1979 , we don't have a feature flag for this at the moment but will look into fixing the issue and provide a better experience. In the meantime, I think you can just leave an empty `sobjects` directory under your project's `.sfdx/tools/` and that should prevent the background sobject refresh from getting triggered.

cc @ntotten

amitkumarj1979:
Thanks @lcampos - Thats what i did for now :-)

PR #987, opened 2019-01-30
Fix Deploy on Save to respect changes made to the sfdx-project.json file

allileong:
### What does this PR do?
This PR creates a new SfdxProjectConfig singleton class that auto-updates whenever the `sfdx-project.json` file is changed. 
This PR fixes W-5796697: Deploy on Save does not respect new packageDirectories added to the sfdx-project.json


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -220,5 +220,7 @@ export const messages = {
...
   error_source_path_not_in_package_directory_text:
-    'Error deploying or retrieving source: The file or directory that you tried to deploy or retrieve isn\'t in a package directory that\'s specified in your sfdx-project.json file. Add this location to your "packageDirectories" value, or deploy or retrieve a different file or directory. For details about sfdx-project.json, see: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm'
+    'Error deploying or retrieving source: The file or directory that you tried to deploy or retrieve isn\'t in a package directory that\'s specified in your sfdx-project.json file. Add this location to your "packageDirectories" value, or deploy or retrieve a different file or directory. For details about sfdx-project.json, see: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm',
+  error_parsing_sfdx_project_file:
+    'Sfdx project file seems invalid. Unable to parse %s. %s'
```
allileong: Hi, @ruthemmanuelle. I copied this text from an error message that I saw coming from the LWC extension that reads: 
screen shot 2019-01-30 at 11 17 44 am

The error message that I'm adding reads:
screen shot 2019-01-30 at 11 17 08 am

allileong: Please let me know what you think.

packages/salesforcedx-vscode-core/src/predicates/salesforcePredicates.ts
```diff
@@ -34,21 +34,3 @@ export class IsSfdxProjectOpened implements Predicate<typeof workspace> {
     }
   }
 }
-export class IsInSfdxPackageDirectory implements Predicate<string> {
```
allileong: Moved this logic to the SfdxPackageDirectories class.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectConfig.ts
```diff
@@ -0,0 +1,82 @@
...
+import { notificationService } from '../notifications';
+import { isSfdxProjectOpened } from '../predicates';
+import { SfdxProjectPath } from '../sfdxProject';
+import { telemetryService } from '../telemetry';
+
+/**
+ * Class representing the local sfdx-project.json file.
+ * Does not contain global values.
+ */
+export default class SfdxProjectConfig {
```
allileong: I looked into using vscode's extension [data storage](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage) capabilities to store the sfdxProjectConfig in memory, but the workspaceState, globalState, and storagePath exist on the ExtensionContext and that is only available to the `activate` function in the extensions.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectConfig.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  private static async initializeSfdxProjectConfig() {
+    if (
+      !SfdxProjectConfig.instance &&
+      isSfdxProjectOpened.apply(vscode.workspace).result
+    ) {
+      const sfdxProjectPath = SfdxProjectPath.getPath();
+      try {
+        const sfdxProject = await SfdxProject.resolve(sfdxProjectPath);
+        SfdxProjectConfig.instance = await sfdxProject.retrieveSfdxProjectJson();
```
allileong: Previously, we were using `sfdxProject.resolveProjectConfig()`, which fetches an aggregation of the global + local config values. Using `sfdxProject.retrieveSfdxProjectJson()` instead will only pull in the values present in the local `sfdx-project.json` config file.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectPath.ts
```diff
@@ -4,5 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import * as vscode from 'vscode';
 
-export { SfdxProjectJsonParser } from './sfdxProjectJsonParser';
+export default class SfdxProjectPath {
+  public static getPath() {
+    return vscode.workspace!.workspaceFolders![0].uri.fsPath;
+  }
```
brpowell: This is much nicer than repeatedly copying and pasting that long statement!

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: It's a bit long, and "seems invalid" feels kind of wishy-washy. How about this?

`Couldn't parse sfdx-project.json file (%s). Parse error: %s`

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectConfig.ts
brpowell: With this new way do you think default values for missing entries (sfdcLoginUrl for example) will still be returned? Was that case actually a default value, or something from a global config?

allileong: @brpowell If a value isn't found in the sfdx-project.json, then getValue() will just return `undefined`. I tested the auth command, and when the sfdx-project.json doesn't contain a `sfdcLoginUrl`, then no default is provided in the dropdown.

allileong:
@brpowell That's a good question. There's still a work item out there to move the sfdxProjectJsonParser (now SfdxProjectConfig) out of the core package and into util, so I'm thinking of leaving that work to be completed for that work item.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectConfig.ts
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { SfdxProject, SfdxProjectJson } from '@salesforce/core';
+import { AnyJson, JsonMap } from '@salesforce/ts-types';
```
lcampos: JsonMap is not used, lets remove it.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectConfig.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public static async getInstance(): Promise<SfdxProjectJson> {
+    if (!SfdxProjectConfig.instance) {
+      await SfdxProjectConfig.initializeSfdxProjectConfig();
+    }
+    return SfdxProjectConfig.instance;
+  }
+
+  public static async getValue(key: string): Promise<AnyJson | undefined> {
```
lcampos: Overall the changes here look good. Since `sfdxProjectConfig.ts` will become the way we interact with `sfdx-project.json` what do we think about accessing its values by doing `SfdxProjectConfig.getValue('packageDirectories')` versus having getters specific to the values the file can have.

I'm asking right now cause we're already using two of those values; packageDirectories and sfdcLoginUrl.

allileong: @lcampos Are you suggesting that for any value that we're retrieving from the `sfdx-project.json`, we create a specific getter method in this file? For example, since we are already using packageDirectories and sfdcLoginUrl, then we have a method called `SfdxProjectConfig.getPackageDirectories()` and `SfdxProjectConfig.getSfdcLoginUrl()`, and whenever we need to access a new value, we just add another method for that value?

lcampos: I'm not suggesting anything yet, I have some questions regarding how we want/need to consume `sfdx-project.json` data. The reason why I bring this up is because some of these values need to be validated before we can use them in other parts of the project. One example here is how we have some checks in `sfdxPackageDirectories.getPackageDirectoryPaths()`. If I want to consume package directory 
data I would use sfdxPackageDirectories instead of directly calling for that array using `SfdxProjectConfig.getValue('packageDirectories')` because it gives me all the error handling.

Opposite to this, if I need the `sfdcLoginUrl` data, its validations would have to be done in `forceAuthWebLogin` since we do not have something like `sfdxPackageDirectories` for it.

Is this because we need the flexibility of calling `SfdxProjectConfig.getValue('<value name>')` across the project or should we make additional changes so that we use interfaces that will run those validations every time we need to use one of these values ?

lcampos: Update: Alli and I synced "offline", we decided to keep the current changes and monitor how we   access `sfdx-project.json` values before making any further changes. If we need to refactor to make it simpler to consume those values then we'll work on them as needed.

PR #989, opened 2019-01-30
Added execution time to command execution telemetry

ntotten:
### What does this PR do?
Tracks execution time for command execution.

@W-5825408@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
@@ -63,13 +63,18 @@ export class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
   public async execute(
     response: ContinueResponse<ApexDebugLogIdStartTime>
   ): Promise<void> {
+    const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
     const execution = new CliCommandExecutor(this.build(response.data), {
       cwd: vscode.workspace.rootPath
     }).execute(cancellationToken);
     this.attachExecution(execution, cancellationTokenSource, cancellationToken);
-    this.logMetric(execution.command.logName);
+
+    execution.processExitSubject.subscribe(async data => {
+      this.logMetric(execution.command.logName, startTime);
+    });
+
     const result = await new CommandOutput().getCmdResult(execution);
     const resultJson = JSON.parse(result);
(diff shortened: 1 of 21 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -63,13 +63,18 @@ export class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
...
+    const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
     const execution = new CliCommandExecutor(this.build(response.data), {
       cwd: vscode.workspace.rootPath
     }).execute(cancellationToken);
     this.attachExecution(execution, cancellationTokenSource, cancellationToken);
-    this.logMetric(execution.command.logName);
+
+    execution.processExitSubject.subscribe(async data => {
```
allileong: @lcampos I don't think it will break the functionality, but the 'async' keyword isn't needed here and is used in several other places where it isn't necessary. 

Eslint has a rule 'require-await' that flags the usage of 'async' without 'await', but tslint does not have the same rule.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -120,8 +121,7 @@ export class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}>
     if (refreshSource !== SObjectRefreshSource.Manual) {
       logName += '_' + refreshSource;
     }
-    this.logMetric(logName);
-
+    this.logMetric(logName, startTime);
```
allileong: @lcampos Is there a reason why this instance of `this.logMetric` isn't inside of a listener for ` execution.processExitSubject()`?

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
lcampos: Good catch, let me update all the classes that have a similar pattern. Might end up even removing data, if possible, since I'm not using it.

packages/salesforcedx-vscode-core/src/commands/forceGenerateFauxClasses.ts
lcampos: Forgot to refactor this piece, thanks for pointing it out.

Issue #962, opened 2018-10-15
Re-run Apex test command and showing coverage option don't work in Apex test view.


shunkosa:
### Summary
The commands after you run Apex tests, **SFDX: Re-Run Last Invoked Apex Test Class** and **SFDX: Re-Run Last Invoked Apex Test Method**, don't work if you run Apex test in Apex Test side bar view. **salesforcedx-vscode-core.retrieve-test-code-coverage** option doesn't work either. But they work well when you run test from link text.

### Steps To Reproduce:

1.  Set salesforcedx-vscode-core.retrieve-test-code-coverageShow as true in settings.json
2. Show Apex test sidebar from beaker icon.
3. Run a Apex test.

### Expected result

--codecoverage option should be added and re-run commands should be available when you run test in Apex test sidebar. Now they work well in "Run All Tests" or "Run Test" texts in the editor.

### Actual result

* --codecoverage option is not added.
* Re-run commands are not available.

### Additional information

_Feel free to attach a screenshot_.

**VS Code Version**: 1.28.1

**SFDX CLI Version**: 44.2.2

**OS and version**: Mac OS X (10.13.6 High Sierra)

**Internal Work Item**: [W-5296419](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005QaeqIAC/view)

JamieBriersSageCom:
As this work has now been merged to develop, is the potential release still Winter 19, or could we expect this fix sooner?

Thanks

ntotten:
@JamieBriersSageCom We release every Thursday. This didn't make the cut for this release (we fork on Monday so we can test, etc.), but it will be in next Thursday.

JamieBriersSageCom:
Brilliant news, thank you

brpowell:
Hi @JamieBriersSageCom. The merge from yesterday specifically fixes the code coverage option not being honored when running tests from the test sidebar. The separate issue of the SFDX: Re-Run Last Invoked Apex Test Method/Class not working for test sidebar tests has been filed as separate work.

JamieBriersSageCom:
Okay, thanks for the clarification.  Personally, its the ability to use the sidebar whilst getting coverage data which has been my biggest worry, so I'm happy that's resolved and I look forward to further fixes. 

Thanks very much for your hard work.

lcampos:
Released as part of version `45.6.0`

PR #1018, opened 2019-02-05
Running tests in the test side bar honors code coverage setting

brpowell:
### What does this PR do?
Running tests from the test sidebar now outputs code coverage if a user has the code coverage setting turned on.

W-5296419
#962

### DIFF
--- a/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
@@ -8,6 +8,7 @@
 import * as vscode from 'vscode';
 import {
   PUSH_OR_DEPLOY_ON_SAVE_ENABLED,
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -54,6 +55,10 @@ export class SfdxCoreSettings {
     return this.getConfigValue<boolean>(PUSH_OR_DEPLOY_ON_SAVE_ENABLED, false);
   }
 
+  public getRetrieveTestCodeCoverage(): boolean {
```
lcampos: This is good, lets also update `forceApexTestRun.ts` and `forceApexTestRunCodeAction.ts` to use it instead of having `sfdxCoreSettings.getConfiguration().get('retrieve-test-code-coverage')` there.

PR #1022, opened 2019-02-07
Run integration tests with vscode 1.30.0

lcampos:
### What does this PR do?
Makes the integration tests run against vscode 1.30 while we sort out a way to fix terminal configuration requiring an additional step. 

https://github.com/Microsoft/vscode/issues/68032

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-debugger/package.json
@@ -54,7 +54,7 @@
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
     "test:vscode-integration":
-      "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
+      "cross-env CODE_VERSION='1.30.0' node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
     "test:vscode-insiders-integration":
       "cross-env CODE_VERSION=insiders npm run test:vscode-integration"
   },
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -54,7 +54,7 @@
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
     "test:vscode-integration":
-      "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
+      "cross-env CODE_VERSION='1.30.0' node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
```
allileong: Could this be added directly to the `run-vscode-integration-tests-with-top-level-extensions` script? I also noticed that isn't being added in the `salesforcedx-vscode-visualforce` package. Does that package need to be treated differently?

lcampos:
@AnanyaJha we're currently not running the insiders tests in appveyor so I didn't make the change for those.

packages/salesforcedx-vscode-apex-debugger/package.json
lcampos: We could move it there, I just didn't cause I focused only on the test suites that were failing. I think the `salesforcedx-vscode-visualforce` ones are still passing on version 1.31.0 because they do not rely on running a command via the cli, so the terminal changes do not affect them or the new info message is not messing with them.

lcampos: @allileong I ended up following your suggestion to fix appveyor timeouts, thanks :)

PR #1025, opened 2019-02-08
Move to node 10.2.0 and npm 6.7.0

lcampos:
### What does this PR do?
Updates extension config to use nodejs v10.2.0. Main motivation for this is vscode v1.31.0 now using that same version.

@W-5837736@


.appveyor.yml
```diff
@@ -18,7 +18,7 @@ environment:
     #- vscode: test:insiders-without-system-tests
 
 install:
-  - ps: Install-Product node 8 x64
+  - ps: Install-Product node 10.2.0 x64
   - appveyor-retry npm install
```
lcampos: I'm thinking about updating this as well cause appveyor is installing version 5.6.0 of npm but would first like to get some feedback about the version changes.

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/async-lock": "0.0.20",
     "@types/chai": "^4.0.0",
     "@types/mocha": "2.2.38",
-    "@types/node": "8.9.3",
+    "@types/node": "10.1.4",
```
lcampos: @vazexqi want to point out this last change before I merge this change. I didn't see any code changes needed after updating to this types/node version.

lcampos: FYI: we synced offline and decided to make this change later as a separate PR since we don't want to encourage usage of node 10 api's just yet.

PR #1026, opened 2019-02-08
[PoC] Use new HTML Data instead of Custom Visualforce LSP

ntotten:


lcampos:
I'm closing this since it was a proof of concept.

PR #1041, opened 2019-02-12
Initial changes for code cov

AnanyaJha:
### What does this PR do?

This PR does the initial changes to calculate code coverage across all types of tests.

- Updates `nyc` configs with basic flags for `remap-istanbul` across all packages with `vscode-integration` tests
- Updates `remap-coverage` script to work across packages
- Adds `coverage` script to all `package.json`s with `vscode-integration` tests
- Adds missing `coverconfig.json` across packages for `mocha` 
- Removes `run-vscode-integration-tests` script
- Removes `skipped` test from `salesforcedx-apex-debugger`


packages/system-tests/assets/sfdx-simple/.forceignore
```diff
@@ -1 +1,4 @@
-force-app/main/default/lightningcomponents/.gitkeep
\ No newline at end of file
+force-app/main/default/lightningcomponents/.gitkeep
+force-app/main/default/lightningcomponents/jsconfig.json
+
+force-app/main/default/lightningcomponents/.eslintrc.json
```
AnanyaJha: This was accidentally added to the commit

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,15 @@
 {
-  "extensions.autoUpdate": false
-}
+    "extensions.autoUpdate": false,
+    "terminal.integrated.env.osx": {
```
AnanyaJha: This was accidentally added to the commit

AnanyaJha:
@allileong I deleted the `run-vscode-integration-tests` script and replaced any packages that were dependent on it with `run-vscode-integration-tests-with-top-level-extensions` because the two scripts seemed to be for the same purpose. Let me know if that looks okay bc I'm not sure if we'd kept them both there for another reason!

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -42,7 +42,7 @@
     "watch": "tsc -watch -p .",
     "clean":
       "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
-    "test": "npm run test:unit && npm run test:integration",
+    "test": "npm run test:unit",
```
lcampos: Why are we now exclusively running unit tests with the `test` script ?

scripts/remap-coverage.js
```diff
@@ -15,10 +15,7 @@ const remapIstanbulExecutable = path.join(
 );
 
 const coverageParentFolder = path.join(
-  __dirname,
```
lcampos: I'm assuming this is because `remap-coverage` is being used across the project and not exclusively on system tests ?

packages/salesforcedx-slds-linter/package.json
```diff
@@ -38,12 +38,13 @@
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
     "test:vscode-integration":
-      "node ../../scripts/run-vscode-integration-tests",
+      "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
```
lcampos: One thing I just noticed is that changing this to run with `run-vscode-integration-test-with-top-level-extensions` will install all the other extensions in the vscode test instance. Do we need that for this ?

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -35,7 +35,8 @@
     "cross-env": "^5.0.5",
     "mocha": "3.2.0",
     "sinon": "^2.3.6",
-    "vscode": "1.1.17"
+    "vscode": "1.1.17",
+    "nyc": "^11.0.2"
```
lcampos: Can we set the `nyc` version without using the carrot ? Also, is this the same `nyc` version used by the sfdx cli ?

packages/salesforcedx-vscode-apex-replay-debugger/coverconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
allileong: When these changes stablilize, could you please add an explanation of these configurations to our testing documenation?

packages/salesforcedx-apex-debugger/package.json
AnanyaJha: There were two integration tests in this package that were being `skipped,` which was stopping codecov from reporting. I was going to comment the tests out, but I realized those are the only two `integration` tests in the package so I adjusted the `test` script instead.

packages/salesforcedx-vscode-apex-replay-debugger/coverconfig.json
AnanyaJha: Yup, definitely will update the documentation!

packages/salesforcedx-apex-debugger/package.json
lcampos: Got it, can you log a child story so we can fix them later ?

scripts/remap-coverage.js
AnanyaJha: Yeah, the `coverage` script (added to packages previously missing it) uses `remap-coverage`. Once we get the other components working, we'll have to use `remap-coverage` across the project.

packages/salesforcedx-slds-linter/package.json
AnanyaJha: Ah okay, so it looks like this test does need a vscode instance, but it doesn't need all the other extensions for the test. In that case, should I reverse this change?

packages/salesforcedx-vscode-apex-debugger/package.json
AnanyaJha: We can use the same version as sfdx cli- I'm going to try configuring ours using `ts-node` as well so this dependency will be removed. In case it doesn't work, I'll update this version number.

PR #1062, opened 2019-02-14
Remove use of sfdx-vscode context for cli usage

lcampos:
### What does this PR do?
- Reformats package.json files using prettier
- Remove use of sfdx-vscode context for cli usage inside of vscode
- Run all tests using latest version of vscode

@W-5858110@


allileong:
Will there be an alternative way for the CLI logging to know that commands are coming from VS Code?

lcampos:
@allileong not immediately, we think we can replace those metrics with the ones we've been adding directly to vscode.

PR #1067, opened 2019-02-15
Update prettier and husky

lcampos:
### What does this PR do?
Updates prettier and husky versions at the top level to fix current formatting issues.

@W-5862907@


docs/package.json
```diff
@@ -4,18 +4,8 @@
...
+    "reformat": "node ../scripts/reformat-docs-with-prettier.js"
   },
   "author": "",
-  "license": "ISC",
-  "devDependencies": {
-    "husky": "^1.3.1",
-    "prettier": "^1.16.4",
-    "pretty-quick": "^1.10.0"
-  },
-  "husky": {
```
lcampos: husky can not be isolated to one package/directory when the project is a mono repo. I've moved this to the top level of the repo.

docs/package.json
```diff
@@ -4,18 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "prettier": "prettier --write --config \"./.prettierrc\" \"**/*.{yml,json,html,css,js,md}\""
```
lcampos: Since everything's moving to the top level, I'm also replacing this with a script in order to use the same prettier version across the project.

package.json
```diff
@@ -62,5 +48,10 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "node": "10.2.0"
+  },
+  "husky": {
```
lcampos: husky config that applies across the project, this replaces having `prepush` under scripts since that's being deprecated.

contributing/tests.md
```diff
@@ -126,24 +126,24 @@ The important args are:
 
 - The first, optional, parameter is a location to a folder that will serve as
   the workspace to run the tests. If you omit this, it just uses a clean
-<<<<<<< HEAD
```
allileong: Looks like this file wasn't properly cleaned up in a previous merge.

lcampos: This is just part of the tests.md doc but it's incredible confusing when reviewing it :)
screen shot 2019-02-19 at 2 33 52 pm

allileong: It looks like there was a three-way merge unrelated to this PR in https://github.com/forcedotcom/salesforcedx-vscode/commit/7ff38a9b7c7f4ed8f887f4d5f14c026c74249389#diff-7f9d343d396a3b97d4b2960c77dfb89aR83 that wasn't resolved properly. The `contributing/tests.md` is a bit broken even in the develop branch https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/contributing/tests.md, but that can be fixed in a separate PR.

lcampos: I see what you mean. I'll just fix it here, thanks for catching that 👍

PR #1080, opened 2019-02-20
Revise setting description for sobject refresh on startup

brpowell:
### What does this PR do?

Revised the setting description based on Ruth's feedback.

W-5861128


lcampos:
@ruthemmanuelle @brpowell I'm taking care of prettier configuration changes in PR #1067, line sizes will change based on config and the new version of prettier we're moving to so don't spend time on it here.

brpowell:
@ruthemmanuelle thanks for doing that! Since "sObject" seems to be the definitive casing we should probably change it to so in other places where it's cased "SObject".

PR #1085, opened 2019-02-20
Show deploy status in both the problem view and output view in a human readable format

brpowell:
### What does this PR do?

Shows the deploy status for force:source:deploy in both the problem view and the output view in a human readable table.

W-5594811


packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -31,39 +39,76 @@ export abstract class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
...
+  }
+
+  private outputResult(parser: ForceDeployResultParser) {
+    const table = new Table();
+    const errors = parser.getErrors();
+    const successes = parser.getSuccesses();
+    const deployedSource = successes ? successes.result.deployedSource : undefined;
+
+    if (deployedSource) {
+      const outputTable = table.createTable(deployedSource as unknown as Row[], [
```
brpowell: `deployedSource as unknown as Row[]` is strange looking, but it's how to force cast. Because the Row type is dynamic with no specific keys, this will work without having to validate the keys of deployedSource.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -31,39 +39,76 @@ export abstract class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
       : '';
     const execFilePathOrPaths = response.data;
     const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: workspacePath
+      cwd: workspacePath,
+      env: { SFDX_JSON_TO_STDOUT: 'true' }
```
brpowell: Forcing all output to stdout for the execution

packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -0,0 +1,68 @@
+/*
```
brpowell: Renamed this class to reflect that it parses errors and successes now.

brpowell:
screen shot 2019-02-22 at 12 56 03 pm

Sample output

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
lcampos: Very cool

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
allileong: Awesome, it's really cool that this can be extended and reused for source:push results too.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -31,39 +39,76 @@ export abstract class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
...
+        { key: 'fullName', label: nls.localize('table_header_full_name') },
+        { key: 'type', label: nls.localize('table_header_type') },
+        { key: 'filePath', label: nls.localize('table_header_project_path') }
+      ]);
+      channelService.appendLine('=== Deployed Source');
+      channelService.appendLine(outputTable);
+    }
+
+    if (errors) {
+      const outputTable = table.createTable(errors.result as unknown as Row[], [
```
lcampos: When running a deploy from a `package.xml` and it has deploy errors, all the error parsing seems correct up until this point since the `errors` object does not have a `result` property and causes the rendering of the table to fail.

screen shot 2019-02-26 at 9 26 20 am

packages/salesforcedx-utils-vscode/src/output/table.ts
```diff
@@ -0,0 +1,118 @@
...
+export class Table {
+  public createTable(rows: Row[], cols: Column[]): string {
+    const maxColWidths = this.calculateMaxColumnWidths(rows, cols);
+
+    let columnHeader = '';
+    let headerSeparator = '';
+    cols.forEach((col, index, arr) => {
+      const width = maxColWidths.get(col.key);
+      if (width) {
+        const isLastCol = index < arr.length - 1;
```
allileong: This variable naming is a little bit misleading - `isLastCol` is `true` when the index does not correspond to the last column. Maybe isLastCol should equal `index === arr.length-1`, and the logic in line 113 could be based on `!isLastCol`?

packages/salesforcedx-utils-vscode/test/unit/cli/deployResultParser.test.ts
```diff
@@ -0,0 +1,206 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
allileong: Nit: Change year to 2019

packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -0,0 +1,68 @@
...
+  result: {
+    deployedSource: DeploySuccess[];
+  };
+}
+
+export class ForceDeployResultParser {
+  private result: any;
+
+  constructor(stdOut: string) {
+    const stdErrLines = stdOut.split(require('os').EOL);
```
lcampos: Lets change this `require` to be an `import` at the top of the file instead of inline.

packages/salesforcedx-utils-vscode/src/output/table.ts
brpowell: Ahhh yes that was a hiccup 😛 thanks!

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -31,39 +39,92 @@ export abstract class ForceSourceDeployExecutor extends SfdxCommandletExecutor<
...
+              key: 'filePath',
+              label: nls.localize('table_header_project_path')
+            },
+            { key: 'error', label: nls.localize('table_header_errors') }
+          ]
+        );
+        channelService.appendLine('=== Deploy Errors');
+        channelService.appendLine(outputTable);
+      } else if (name && message) {
+        channelService.appendLine(`${name}: ${message}\n`);
```
brpowell: If there aren't any results in the error, at least show the name and message of the error to show what happened.

packages/salesforcedx-utils-vscode/test/unit/cli/deployResultParser.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      error: 'Invalid dependency ...',
+      lineNumber: '10',
+      columnNumber: '23',
+      type: 'ApexClass',
+      fullName: 'Testing'
+    });
+
+    const parser = new ForceDeployResultParser(
+      `sfdx force:source:deploy --json --loglevel fatal --manifest /Users/username/manifest/package.xml ${
+        require('os').EOL
```
lcampos: If possible, lets change this to a top level import instead of an inline require.

brpowell: Bleh, i didn't notice it was in a test too. Thanks for catching

PR #1089, opened 2019-02-21
Fix 'manifests' vs. 'manifest' and one other typo

tet3:
When creating a new project with manifest, the directory created is called 'manifest', not 'manifests', and the right-click commands on package.xml will not work if the directory is called 'manifests'. Also removed a stray "in" before the link to Metadata API docs.

### What does this PR do?

*Docs changes only, to one article*. 
Replaces two instances of 'manifests' with 'manifest' in the Org Development Model article.
Removes awkward 'in' before Metadata API docs link.


lcampos:
Thanks for contributing @tet3 !

PR #1091, opened 2019-02-21
[WIP] refactor deprecated workspace.rootPath api

dehru:
### What does this PR do?

This PR deals with the deprecation of workspace.rootPath, and updates the code to use vscode.workspace.workspaceFolders[].  

In this current PR, wherever we use ```vscode.workspace.rootPath```, I’ve changed it to essentially ( through a utility api ) to use ```vscode.workspace.workspaceFolder[0].uri.path```.  And in places where the previous code used possibly undefined (!) ```vscode.workspace!.workspaceFolder![0].uri.fsPath```, I updated it to use the new utility methods and return an empty string instead of possibly throwing an error.

In order for these vscode extensions to work with the Theia ide, it cannot use the deprecated workspace.rootPath api.  

However, my biggest question for the more experienced developers on this project, what exactly IS workspace.rootPath ( is it uri.path or uri.fsPath )?   Should I have always used uri.fsPath?

Thanks for your review!


packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -230,7 +231,7 @@ export abstract class SelectDirPath
   public async gather(): Promise<
     CancelResponse | ContinueResponse<{ outputdir: string }>
   > {
-    const rootPath = vscode.workspace.rootPath;
+    const rootPath = getRootWorkspacePath();
```
dehru: per my question in my PR description, this uses ( through the util ) workspace.workspaceFolders[0].uri.path and not uri.fsPath.

lcampos: We shouldn't use `uri.path` since that breaks file paths in windows, it doesn't return proper file path separators based on platform. Let's change it so everything uses `uri.fsPath`. 

Here's more info on that if needed https://code.visualstudio.com/api/references/vscode-api#Uri

packages/salesforcedx-vscode-core/src/util/rootWorkspace.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function getRootWorkspace(): WorkspaceFolder {
+  return hasRootWorkspace() ? workspace.workspaceFolders![0] : {} as WorkspaceFolder;
+}
+
+export function getRootWorkspaceFsPath(): string {
+  return getRootWorkspace().uri ? getRootWorkspace().uri.fsPath : '';
+}
+
+export function getRootWorkspacePath(): string {
```
lcampos: I think we should remove this getter in order to prevent usage of `uri.path`.

dehru: OK, I ended up removing the other getter ```getRootWorkspaceFsPath()```, and kept ```getRootWorkspacePath()```, but had it dig out the uri.fsPath property and return that.   

I guess to me it seemed to express the user's intent without unnecessary specificity of the underlying model.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,12 @@
 {
-  "extensions.autoUpdate": false
+  "extensions.autoUpdate": false,
```
lcampos: This change should not be part of the PR

packages/system-tests/assets/sfdx-simple/sfdx-project.json
```diff
@@ -6,4 +6,4 @@
   ],
```
lcampos: This change should not be part of the PR

packages/salesforcedx-vscode-core/test/vscode-integration/util/rootWorkspace.test.ts
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
lcampos: Nit pick, copyright needs to change to 2019

packages/salesforcedx-vscode-core/src/util/rootWorkspace.ts
```diff
@@ -0,0 +1,15 @@
+import { workspace, WorkspaceFolder } from 'vscode';
```
lcampos: Need to add the copyright header.

dehru:
This PR was from a fork, and @lcampos asked me to redo it as a branch from this repo.  So I'm closing this PR, and linking back to it from the description of the new PR for reference.

PR #1099, opened 2019-02-22
Dehru/fix deprecated rootworkspace api

dehru:
NOTE: My original PR was from a fork, and so I was asked to create a branch from this remote.  In the process, I can't keep the comments from my original PR.  

Here's the link to the original PR with the original comments:
[https://github.com/forcedotcom/salesforcedx-vscode/pull/1091](https://github.com/forcedotcom/salesforcedx-vscode/pull/1091)

### What does this PR do?
This PR deals with the deprecation of workspace.rootPath, and updates the code to use vscode.workspace.workspaceFolders[].

In this current PR, wherever we use `vscode.workspace.rootPath`, I’ve changed it to essentially ( through a utility api ) to use `vscode.workspace.workspaceFolder[0].uri.path`. And in places where the previous code used possibly undefined (!) `vscode.workspace!.workspaceFolder![0].uri.fsPath`, I updated it to use the new utility methods and return an empty string instead of possibly throwing an error.

In order for these vscode extensions to work with the Theia ide, it cannot use the deprecated workspace.rootPath api.

However, my biggest question for the more experienced developers on this project, what exactly IS workspace.rootPath ( is it uri.path or uri.fsPath )? Should I have always used uri.fsPath?

Thanks for your review!


packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectPath.ts
```diff
@@ -4,10 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import * as vscode from 'vscode';
+import { getRootWorkspacePath } from '../util';
 
 export default class SfdxProjectPath {
   public static getPath() {
-    return vscode.workspace!.workspaceFolders![0].uri.fsPath;
+    return getRootWorkspacePath();
```
lcampos: Instead of only doing this update I think we should replace `SfdxProjectPath`'s usage in favor of the `rootWorkspace` utility in order to have only 1 way of getting `vscode.workspace!.workspaceFolders![0].uri.fsPath` in the core extension.

@allileong what do you think ?

allileong:
Aside from the comment about consolidating the RootWorkspace and SfdxProjectPath utilities, I think the changes look great. Thanks for cleaning up all the different ways that we were validating and getting the workspace root.

allileong:
And regarding `uri.path` vs `uri.fsPath`, the current implementation of the deprecated `vscode.workspace.rootPath` gets the rootPath from `uri.fsPath`. I didn't look at the implementation back when rootPath hadn't been deprecated yet, but based on the current implementation and the doc on the differences between path and fsPath, I suspect that it has always used `fsPath`.

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxProjectPath.ts
dehru: How do you guys want to handle the change to replace SfdxProjectPath with this new utility?  In this PR or in a separate PR?

lcampos: Lets have that as part of this PR.

allileong: I would be fine with doing the refactoring in a separate PR. What do you think @lcampos?

allileong: Oops, sorry, must not have refreshed my page.

dehru: I've removed SfdxProjectPath, references, and tests.  Here's the specific commit of that work.

https://github.com/forcedotcom/salesforcedx-vscode/pull/1099/commits/c21af85f37cb9a0d6eb7198097a506b2e277639b

Let me know if this is sufficient to merge this PR in.

allileong: Awesome, thanks for making that change. This pr looks good to me 👍

lcampos:
@dehru PR looks good, go ahead and merge it once all the tests run successfully.

PR #1107, opened 2019-02-26
Fix extensions failing to activate when sfdx cli is missing

lcampos:
### What does this PR do?
Fixes core extension not getting activated when the salesforce cli is not installed or can't be found, which currently causes all the other extensions to fail.
With the new behavior, we check if the cli is installed before executing cli commands during extension activation. If the cli is not installed, a warning message is surfaced to the user pointing them to where they can download the cli.

cli_not_installed_message

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/src/index.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/src/index.ts
@@ -268,23 +268,25 @@ export async function activate(context: vscode.ExtensionContext) {
     )
   );
 
-  console.log('Setting up ISV Debugger environment variables');
-  // register watcher for ISV authentication and setup default user for CLI
-  // this is done in core because it shares access to GlobalCliEnvironment with the commands
-  // (VS Code does not seem to allow sharing npm modules between extensions)
-  try {
-    registerIsvAuthWatcher(context);
-    console.log('Configured file watcher for .sfdx/sfdx-config.json');
-    await setupGlobalDefaultUserIsvAuth();
-  } catch (e) {
-    console.error(e);
-    vscode.window.showWarningMessage(
-      nls.localize('isv_debug_config_environment_error')
-    );
-  }
-
-  // Telemetry
   if (sfdxCoreExtension && sfdxCoreExtension.exports) {
+    if (sfdxCoreExtension.exports.isCLIInstalled()) {
+      console.log('Setting up ISV Debugger environment variables');
+      // register watcher for ISV authentication and setup default user for CLI
+      // this is done in core because it shares access to GlobalCliEnvironment with the commands
+      // (VS Code does not seem to allow sharing npm modules between extensions)
+      try {
+        registerIsvAuthWatcher(context);
+        console.log('Configured file watcher for .sfdx/sfdx-config.json');
+        await setupGlobalDefaultUserIsvAuth();
+      } catch (e) {
+        console.error(e);
+        vscode.window.showWarningMessage(
+          nls.localize('isv_debug_config_environment_error')
+        );
+      }
+    }
+
+    // Telemetry
     sfdxCoreExtension.exports.telemetryService.showTelemetryMessage();
 
     telemetryService.initializeService(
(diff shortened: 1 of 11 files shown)


packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -268,23 +268,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
-  } catch (e) {
-    console.error(e);
-    vscode.window.showWarningMessage(
-      nls.localize('isv_debug_config_environment_error')
-    );
-  }
-
-  // Telemetry
   if (sfdxCoreExtension && sfdxCoreExtension.exports) {
+    if (sfdxCoreExtension.exports.isCLIInstalled()) {
```
lcampos: After fixing `salesforcedx-vscode-core` I noticed that `salesforcedx-vscode-apex-debugger` would also fail to activate when the cli is not present.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -223,5 +223,7 @@ export const messages = {
   missing_default_org: 'No Default Org Set',
   force_config_set_org_text: 'SFDX: Set a Default Org',
   error_parsing_sfdx_project_file:
-    "Couldn't parse sfdx-project.json file (%s). Parse error: %s"
+    "Couldn't parse sfdx-project.json file (%s). Parse error: %s",
+  sfdx_cli_not_found:
```
lcampos: @ruthemmanuelle this message is similar to `channel_end_with_sfdx_not_found` which is at the top of this same file but I created a new one in order to make the text contain a clickable link.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { which } from 'shelljs';
```
lcampos: Using `shelljs` allows us to have a cross platform way of figuring out if the cli is available in the system.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: Cool. Both of these messages should say just "Salesforce CLI" instead of "The Salesforce CLI". (Similarly, we wouldn't say "The VS Code".)

packages/salesforcedx-vscode-apex-debugger/src/index.ts
brpowell: Are these the only two packages that require the installation of the CLI?

lcampos: These are the only two packages that execute a cli command in their `activation` methods.

brpowell: Gotcha. Also I think you're missing the `showCLINotInstalledMessage()` call if it's not installed

allileong: Are we calling CLI commands from extensions that don't require it in the activation specifically? Do we need to warn users in those cases too?

lcampos: @brpowell since `salesforcedx-vscode-apex-debugger` depends on `salesforcedx-vscode-core` I'm only using that in the core extension to only show the warning message once.

@allileong these two are the only extensions that calling the cli during activation. I think we can eventually move some of this out of activation but not immediately since some of those calls are being used to set the context used to show/hide commands and require a bigger refactor.
What type of warning are you thinking ?

allileong: Sorry, my mistake. I looked in the code, and was thinking of a case that doesn't exist.

PR #1117, opened 2019-03-04
Show source:push errors in problem view and output view

brpowell:
### What does this PR do?

force:source:push errors are now displayed in the problem view in addition to the output view.

W-5594854


packages/salesforcedx-utils-vscode/test/unit/cli/deployResultParser.test.ts
```diff
@@ -223,4 +223,37 @@ describe('force:source:deploy parser', () => {
...
+          type: 'ApexClass',
+          fullName: 'MyClass',
+          filePath: 'src/classes/MyClass.cls'
+        }]
+      }
+    };
+
+    const parser = new ForceDeployResultParser(JSON.stringify(response));
+    const successes = parser.getSuccesses();
+    if (successes) {
```
allileong: Could you add an else case to handle no successes?

brpowell: Ah yes good catch, thanks!

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -24,19 +27,41 @@ import { telemetryService } from '../telemetry';
 import { getRootWorkspacePath } from '../util';
 import { SfdxCommandletExecutor } from './commands';
 
+export interface DeployParams {
+  sourcePush: boolean;
+  sourcePaths: string;
+}
+
+export class DeployParamsGatherer implements ParametersGatherer<DeployParams> {
```
allileong: Hey Bryan, as an alternative to creating a new `DeployParamsGatherer` and `SourcePathDeployChecker` in order to plumb the `sourcePush` boolean through the entire command, how would you feel about instead having forceSourceDeploy.ts's `outputResult()` call out to another instance method `getTableType()` to get the table type? You could implement something like `protected getTableType() { return 'deploy' }` in ForceSourceDeployExecutor and override the function in ForceSourcePushExecutor to be something like `protected getTableType() { return 'push' };

brpowell: @allileong I like that idea a lot better!

allileong:
Hey @brpowell, I hadn't thought of this until now, but `source:push` is different from `source:deploy` in that `source:push` reports conflicts. Could you update your changes to handle that case as well? 

Currently, with your changes, we aren't giving the end user any information about their conflicts in that error scenario:
screen shot 2019-03-06 at 1 31 46 pm

When I call the `source:push` command directly from the terminal, the output that I see is:
screen shot 2019-03-06 at 1 29 06 pm

allileong:
@brpowell sorry about that. I meant to hit `Comment`, not `Close and comment` 🙊

packages/salesforcedx-vscode-core/src/commands/forceSourcePush.ts
```diff
@@ -9,15 +9,16 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as vscode from 'vscode';
```
lcampos: This import is not needed.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -13,7 +13,10 @@ import {
   Row,
   Table
 } from '@salesforce/salesforcedx-utils-vscode/out/src/output';
-import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import {
+  ContinueResponse,
+  ParametersGatherer
```
lcampos: `ParametersGatherer` is not used, you can remove it.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
```diff
@@ -13,7 +13,10 @@ import {
   Row,
```
lcampos: All the changes done lately around deploy and push have left the name of this file a bit obsolete, it doesn't reflect the code's purpose anymore. What do you think about renaming it ?

lcampos: cc @allileong

packages/salesforcedx-vscode-core/src/commands/forceSourcePush.ts
```diff
@@ -9,15 +9,16 @@ import {
...
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
-  SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
+import { DeployType, ForceSourceDeployExecutor } from './forceSourceDeploy';
 
-export class ForceSourcePushExecutor extends SfdxCommandletExecutor<{}> {
+export class ForceSourcePushExecutor extends ForceSourceDeployExecutor {
```
lcampos: Very cool change in order to reuse what we have for deploy 👍

lcampos:
@brpowell I noticed that we currently display the table (output view) in between the command that gets printed at the beginning of the execution and the command line that gets printed when the execution ends.
screen shot 2019-03-07 at 6 52 34 pm

With these changes, we'll display the table after printing both commands. Can we control this on the override of `execute` on `forceSourceDeploy.ts` ?

screen shot 2019-03-07 at 6 52 47 pm

brpowell:
@lcampos It has changed because before the command was using the execution output, whereas now the extension is the one generating the output.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -230,5 +230,12 @@ export const messages = {
...
   table_header_full_name: 'FULL NAME',
-  table_header_state: 'STATE'
+  table_header_state: 'STATE',
+  table_no_results_found: 'No results found',
+  table_title_deployed_source: 'Deployed Source',
+  table_title_deploy_errors: 'Deploy Errors',
+  table_title_pushed_source: 'Pushed Source',
+  table_title_push_errors: 'Push Errors',
+  push_conflicts_error:
+    "We couldn't complete the push operation due to conflicts. Verify that you want to keep the local versions, then run SFDX: Push Source to Default Scratch Org and Override Conflicts."
```
brpowell: @ruthemmanuelle Added an error message for source:push conflicts that closely resembles the one in the CLI. Main different is instead of telling the user to run `sfdx force:source:push -f` it specifies the corresponding VSCode command

packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -6,65 +6,71 @@
  */
 import { EOL } from 'os';
 
-export interface DeployError {
+export const CONFLICT_ERROR_NAME = 'sourceConflictDetected';
+
+export interface DeployResult {
```
brpowell: A conflict has the same fields as a successful deploy. I just decided to make everything a "DeployResult"

packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -6,65 +6,71 @@
...
 
-export interface DeploySuccess {
-  state: string;
-  fullName: string;
-  type: string;
-  filePath: string;
-}
-
-export interface ForceSourceDeployErrorResult {
+export interface ForceSourceDeployErrorResponse {
```
brpowell: DeployResult and ForceSourceDeployErrorResult were confusing to me. Changed it to ForceSourceDeployErrorResponse to more accurately reflect what it represents. Same for the success response

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploy.ts
brpowell: Do you think it should be something more like "baseDeployCommand"?

allileong: That sounds good to me 👍

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ruthemmanuelle: How about this, @brpowell ? 

We couldn’t push your source due to conflicts. Make sure that you want to overwrite the metadata in your org with your local files, then run "SFDX: Push Source to Default Scratch Org and Override Conflicts".

PR #1119, opened 2019-03-04
Add CPUs and total system memory 

allileong:
### What does this PR do?
Add CPUS and total system memory to our extension telemetry

W-5870048

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -73,7 +73,12 @@ import { setDefaultOrg, showDefaultOrg } from './orgPicker';
 import { registerPushOrDeployOnSave, sfdxCoreSettings } from './settings';
 import { taskViewService } from './statuses';
 import { telemetryService } from './telemetry';
-import { getRootWorkspacePath, hasRootWorkspace, isCLIInstalled, showCLINotInstalledMessage } from './util';
+import {
+  getRootWorkspacePath,
+  hasRootWorkspace,
+  isCLIInstalled,
+  showCLINotInstalledMessage
+} from './util';
 
 function registerCommands(
   extensionContext: vscode.ExtensionContext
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -73,7 +73,12 @@ import { setDefaultOrg, showDefaultOrg } from './orgPicker';
 import { registerPushOrDeployOnSave, sfdxCoreSettings } from './settings';
 import { taskViewService } from './statuses';
 import { telemetryService } from './telemetry';
-import { getRootWorkspacePath, hasRootWorkspace, isCLIInstalled, showCLINotInstalledMessage } from './util';
+import {
```
allileong: This is just an auto-formatting change

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
```diff
@@ -112,6 +112,16 @@ export default class TelemetryReporter extends vscode.Disposable {
       /^(\d+)(\.\d+)?(\.\d+)?(.*)/,
       '$1$2$3'
     );
+    const cpus = os.cpus();
```
allileong: screen shot 2019-03-04 at 1 26 57 pm
This is what the newly added fields look like in AppInsights.

The information about CPUs and total memory is fetched once when telemetry is first initialized on extension activation. It is also possible to get information on the amount of *available* memory from the `os` package, but it seems like that information would be most useful if we fetched it each time of each telemetry event so that we can get a snapshot of available memory as the commands are being executed. @lcampos  Is this information that I should add to each event?

lcampos: I think this is ok. The telemetryReporter.ts is supposed to be a temporary fork of https://github.com/Microsoft/vscode-extension-telemetry since we wanted to add some capabilities to it and were waiting on those changes to get merged. Once we merge this we should open an issue on the repo and open a PR to see if we can get them merged.

lcampos: For the available memory info, I agree that it makes more sense to be captured at a per command basis. Lets handle it as a separate work item.

PR #1121, opened 2019-03-05
Enable code coverage for vscode-integration tests

AnanyaJha:
### What does this PR do?
This PR enables code coverage to run across our vscode-integration tests

- Fixes file reference issues
- Adjusts for Mocha exit event
- Creates a top level config file
- Instruments all `salesforcedx-vscode-*` extensions for system tests coverage
- Bumps `nyc` version to be consistent with `sfdx-core`
- Removed unused dependencies
@W-5812998@


AnanyaJha:
There is another issue where unit tests + coverage are not being reported on appveyor or codecov because the output file (`junit-custom.xml`) is being overwritten by integration tests. Depending on the package, we may run `npm run test:unit & npm run test:vscode-integration`. Each of those scripts individually writes to a `junit-custom.xml` file with the name of the test, time, and some other info. These `junit-custom.xml` files are the ones that populate appveyor with the test data so we cant view the test list on our builds. Each test also generates code coverage reports, which get sent to the `coverage` directory in each package. Because we run both `integration` and `unit` tests in some directories, whichever command runs second is overwriting the results of the first command run- so we don’t actually see our `unit` tests listed in the total number of tests in the appveyor builds or total code coverage in codecov. So far, I’ve confirmed this behavior in all the packages that run `npm run test:unit && npm run test:vscode-integration` so `salesforcedx-apex-debugger`, `salesforcedx-apex-replay-debugger`, and `salesforcedx-sobjects-faux-generator.` 
@sfsholden The changes to breakup the autobuilds may take care of this issue by itself. Otherwise, I think we should start writing to two separate files based on the type of tests.

PR #1128, opened 2019-03-06
Configuration changes to fix issues with appveyor builds

lcampos:
### What does this PR do?
- Updates `cross-env` version across packages
- Updates tests that didn't properly handle stub restore when timing out
- Updates mocha config
- Sets specific vscode version to run against when building all the extensions
- Updates appveyor config in order to use the same node version across script executions & set clone depth in order to speed up build setup.

@W-5917635@

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,6 +4,9 @@ version: '{branch}-{build}'
 # Do not build on tags (GitHub and BitBucket)
 skip_tags: true
 
+# Limit the number of last commits when cloning the repo/branch
+clone_depth: 1
+
 # Do not build feature branch with open Pull Requests
 skip_branch_with_pr: true
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,6 +30,7 @@ install:
 build_script:
   - npm run compile
   - npm run lint
+  - npm config set scripts-prepend-node-path true
 
 test_script:
   - node --version
--- a/packages/salesforcedx-vscode-apex-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/package.json
@@ -38,7 +38,7 @@
     "@types/path-exists": "^1.0.29",
     "@types/sinon": "^2.3.2",
     "chai": "^4.0.2",
-    "cross-env": "^5.0.5",
+    "cross-env": "5.2.0",
     "mocha": "3.2.0",
     "sinon": "^2.3.6",
     "vscode": "1.1.17"
--- a/packages/salesforcedx-vscode-apex-replay-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/package.json
@@ -58,7 +58,7 @@
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
-    "test:vscode-integration": "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
+    "test:vscode-integration": "node ../../scripts/run-vscode-integration-tests",
     "test:vscode-insiders-integration": "cross-env CODE_VERSION=insiders npm run test:vscode-integration"
   },
   "activationEvents": [
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
@@ -175,16 +175,17 @@ describe('Set Default Org', () => {
     orgListStub = sinon.stub(OrgList.prototype, 'updateOrgList');
     quickPickStub = sinon.stub(vscode.window, 'showQuickPick');
   });
-  afterEach(() => {
-    orgListStub.restore();
-    quickPickStub.restore();
-  });
 
   it('should return Cancel if selection is undefined', async () => {
-    orgListStub.returns(orgList);
-    quickPickStub.returns(undefined);
-    const response = await setDefaultOrg();
-    expect(response.type).to.equal('CANCEL');
+    try {
+      orgListStub.returns(orgList);
+      quickPickStub.returns(undefined);
+      const response = await setDefaultOrg();
+      expect(response.type).to.equal('CANCEL');
+    } finally {
+      orgListStub.restore();
+      quickPickStub.restore();
+    }
   });
 
   it('should return Continue and call force:auth:web:login command if SFDX: Authorize an Org is selected', async () => {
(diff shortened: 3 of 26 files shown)


.appveyor.yml
```diff
@@ -4,6 +4,9 @@ version: '{branch}-{build}'
 # Do not build on tags (GitHub and BitBucket)
 skip_tags: true
 
+# Limit the number of last commits when cloning the repo/branch
+clone_depth: 1
```
lcampos: This makes clone go from 2+mins to 10sec or less in appveyor. More info here https://www.appveyor.com/docs/how-to/repository-shallow-clone/

.appveyor.yml
```diff
@@ -30,6 +33,7 @@ install:
 build_script:
   - npm run compile
   - npm run lint
+  - npm config set scripts-prepend-node-path true
```
lcampos: Enforces all npm scripts run with the node version configured here and not the default one.

packages/salesforcedx-vscode-apex-replay-debugger/package.json
```diff
@@ -58,7 +58,7 @@
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
-    "test:vscode-integration": "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
+    "test:vscode-integration": "node ../../scripts/run-vscode-integration-tests",
```
lcampos: The tests on this package do not need the extensions being spinned up with vscode.

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
```diff
@@ -175,16 +175,17 @@ describe('Set Default Org', () => {
...
-    orgListStub.restore();
-    quickPickStub.restore();
-  });
 
   it('should return Cancel if selection is undefined', async () => {
-    orgListStub.returns(orgList);
-    quickPickStub.returns(undefined);
-    const response = await setDefaultOrg();
-    expect(response.type).to.equal('CANCEL');
+    try {
```
lcampos: These tests are a bit fragile and once one fails it causes all of them to fail because the stubs didn't get restored.

packages/salesforcedx-vscode-lightning/test/vscode-integration/deprecatedClassName.test.ts
```diff
@@ -15,7 +15,7 @@ describe('SLDS Deprecated Class Name', () => {
   let lightningExtension: vscode.Extension<any>;
 
   before(async () => {
-    if (vscode.workspace.rootPath) {
+    if (vscode.workspace.workspaceFolders![0]) {
```
lcampos: Updating test to remove the use of a deprecated api.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -59,9 +59,8 @@
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "pretest": "node ../../scripts/download-vscode-for-system-tests",
     "test": "npm run test:vscode-integration",
-    "test:vscode-integration": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
+    "test:vscode-integration": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-vscode-integration-tests",
```
lcampos: These tests also do not need all the extensions to be spinned up with vscode.

scripts/run-vscode-integration-tests-with-top-level-extensions.js
```diff
@@ -10,7 +10,7 @@ const cwd = process.cwd();
 // Executes the test, using the top-level packages as the CODE_EXTENSIONS_PATH
 
 shell.exec(
-  `cross-env CODE_EXTENSIONS_PATH='${path.join(
+  `cross-env CODE_VERSION='1.30.0' CODE_EXTENSIONS_PATH='${path.join(
```
lcampos: Spinning up vscode 1.31.1 with all the code under `packages` is causing it to crash in Windows environments, which is what we have appveyor running on. I'm setting it to use `1.30.0` while I find a fix for this. FYI, linux and macos builds of vscode do not crash.

.appveyor.yml
lcampos: cc @sfsholden first step towards faster appveyor runs :)

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
AnanyaJha: Ah so the afterEach doesn't run if a test fails?

.appveyor.yml
sfsholden: Only thing I noticed on this page was a warning about setting the number too low. I don't know if this would cause an issue for us or not.

"Be aware that if you do a lot of commits producing queued builds and depth number is too small git checkout operation following git clone can fail because requested commit is not present in a cloned repository."

scripts/run-vscode-integration-tests-with-top-level-extensions.js
JimSuplizio: Does this mean we're going to have to update this as the code version changes?

.appveyor.yml
lcampos: Yeah, I haven't seen us commit fast enough to run into this but if we start running into issues we can always modify this.

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
lcampos: That's correct, the only config we could use would be to set `--bail` on the mocha options but that would make the test suite stop on the first error. Didn't use `--bail` since there is value in executing the whole test suite when doing significant changes.

AnanyaJha: Cool, makes sense.

scripts/run-vscode-integration-tests-with-top-level-extensions.js
lcampos: No, this is a stop gap while we figure out how to run these tests in the latest vscode version. Also, not all test suites run using this script, most are still running on the latest stable vscode version (currently 1.31.1).

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
JimSuplizio: So, I understand the reasoning for this change but I don't quite understand why it was necessary? I thought Mocha was supposed to run the afterEach regardless of whether or not the test passed? I can see the executeCommand stub needing a catch where its being used since it's not used everywhere.

lcampos: That's not how mocha behaves when it's configured the way we've done it for our test suites. We can change the configuration later if we think we need to.

PR #1133, opened 2019-03-07
Changelog for release v45.5.0

lcampos:
### What does this PR do?
Updates the changelog for release 45.5.0.

I'm updating it directly into the `develop` branch this time since I wasn't able to merge it into the release branch this week.

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -4,25 +4,26 @@
 
 #### salesforcedx-vscode-apex
 
-- Hover capability to show Apex Javadoc information when hovering on symbols ([PR #1106](https://github.com/forcedotcom/salesforcedx-vscode/pull/1106))
+- Show Apex Javadoc information when hovering on symbols ([PR #1106](https://github.com/forcedotcom/salesforcedx-vscode/pull/1106))
 
 ### Fixed
 
-#### salesforcedx-vscode-docs
+#### salesforcedx-vscode/docs
 
-- Fix manifest typos on Org Development Model documentation ([PR #1089](https://github.com/forcedotcom/salesforcedx-vscode/pull/1089))—Contribution by [@tet3](https://github.com/tet3)
+- Fix typos in Org Development Model documentation ([PR #1089](https://github.com/forcedotcom/salesforcedx-vscode/pull/1089))—Contribution by [@tet3](https://github.com/tet3)
 
 #### salesforcedx-vscode-apex
 
-- Fix Apex Language Server performance issue that caused high CPU load: handle document change requests queuing up ([PR #1086](https://github.com/forcedotcom/salesforcedx-vscode/pull/1086), [Issue #1047](https://github.com/forcedotcom/salesforcedx-vscode/issues/1047))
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
...
+
+### Added
+
+#### salesforcedx-vscode-apex
+
+- Hover capability to show Apex Javadoc information when hovering on symbols ([PR #1106](https://github.com/forcedotcom/salesforcedx-vscode/pull/1106))
+
+### Fixed
+
+#### salesforcedx-vscode-docs
```
lcampos: @ruthemmanuelle now that we moved the docs to github pages I was thinking about having a dedicated section for it on the change log, what do you think ?

ruthemmanuelle: Sounds great. Maybe it should be `salesforcedx-vscode/docs`, though? I'm suggesting that because the docs aren't their own extension, so they don't have a Marketplace page with a URL that includes `salesforcedx-vscode-docs`, but their files live in `salesforcedx-vscode/docs` in the repo.

lcampos: Yeah, changing it to `salesforcedx-vscode/docs` makes sense 👍

ruthemmanuelle: Cool, I'll make that change along with some other edits.

packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+## 45.5.0 - March 7, 2019
+
+### Added
+
+#### salesforcedx-vscode-apex
+
+- Show Apex Javadoc information when hovering on symbols ([PR #1106](https://github.com/forcedotcom/salesforcedx-vscode/pull/1106))
```
praksb: i would change "Apex Javadoc" to "Apex doc comment"

lcampos: But doesn't the comment have to follow the Javadoc format ?

praksb: I don't think we officially say Javadoc anywhere in our public documentation. Also, unlike Java doc comments or C# doc comments we don't explicitly support any tags. I think Javadoc is just a format that some customers are using to document Apex source. As for implementation we right now just show any block comments associated with a symbol definition we don't validate or process any of the tags that are used.

lcampos: Ok, let me change it to something like Apex document block comment to reflect what we currently support.

ruthemmanuelle: Does it use the [ApexDoc format](https://github.com/SalesforceFoundation/ApexDoc#documenting-class-files)? Should we say "ApexDoc"?

lcampos: Well, what Prakash mentioned about no support for tags, etc would still apply. Also, that ApexDoc format is not official, right ? By official I mean part of the Apex language.

ruthemmanuelle: Yeah, ApexDoc is a community-driven thing. The "Show Apex block comment information" version you added works, @lcampos .

PR #1135, opened 2019-03-08
Enable SFDX:Re-run commands to work with Test Sidebar

JimSuplizio:
### What does this PR do?
The SFDX: Re-run test commands were not being enabled when tests were run through the test side bar. The reason for this due to the last run class/method were only being cached when tests were run though the source links. As part of this change I moved the source link code actions over from packages/salesforcedx-vscode-core to packages/salesforcedx-vscode-apex which is where the test console lives. 

@W-5828443@


packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extensions.autoUpdate": false
-}
+    "extensions.autoUpdate": false,
+    "eslint.nodePath": "/home/jsuplizio/DevToolsGit/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules",
```
brpowell: Sneaky file I'm guessing from running tests, remember to revert.

packages/salesforcedx-vscode-apex/src/commands/index.ts
```diff
@@ -0,0 +1,15 @@
+/*
```
brpowell: Glad to see this stuff moved to the apex package 👍

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -33,8 +34,8 @@ function isNotEmpty(value: string): boolean {
 // cache last test class and test method values to
 // enable re-running w/o command context via built-in LRU
 class ForceApexTestRunCacheService {
```
brpowell: I might suggest moving the cache service into its own file since multiple commands are utilizing it.

brpowell:
@JimSuplizio when i run SFDX: Invoke Apex Tests, the last class isn't being cached, but everything in the sidebar seems to be working as expected.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
JimSuplizio: Interesting as I've no idea what made this update or why there's no test failures. I'll make sure this is reverted when I update the PR.

JimSuplizio:
>> when i run SFDX: Invoke Apex Tests, the last class isn't being cached, but everything in the sidebar seems to be working as expected

@brpowell This particular bug was only dealing with the test sidebar. If there's yet one more way to run tests then it probably needs a separate bug.

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
JimSuplizio: Sure. It'll be in the updated PR.

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
JimSuplizio: This file is reverted with the latest PR. I wasn't aware that running tests locally modified this file.

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
JimSuplizio: Moved into its own file testRunCache/forceApexTestRunCacheService.ts

packages/salesforcedx-vscode-apex/src/testRunCache/forceApexTestRunCacheService.ts
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: Lets change this to 2019 instead of 2017.

packages/salesforcedx-vscode-apex/src/testRunCache/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: Lets change this to 2019 instead of 2017.

packages/salesforcedx-vscode-apex/src/testRunCache/forceApexTestRunCacheService.ts
JimSuplizio: It will be done.

packages/salesforcedx-vscode-apex/src/testRunCache/index.ts
JimSuplizio: It will be done.

PR #1139, opened 2019-03-11
Lag time for SFDX: Set Default Org command

AnanyaJha:
### What does this PR do?
This PR speeds up the response time for the org picker button and refactors how the command is registered.

@W-5820434@


packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,70 @@
...
+        rootFolder: rootPath + '.sfdx/',
+        filename: 'sfdx-config.json'
+      });
+      const localDefault = myLocalConfig.get('defaultusername');
+      return JSON.stringify(localDefault).replace(/\"/g, '');
+    }
+
+    const aggregator = await ConfigAggregator.create();
+    const globalDefault = aggregator.getPropertyValue('defaultusername');
+    return JSON.stringify(globalDefault).replace('""', '');
```
AnanyaJha: This should also be `.replace(/\"/g, '')`, I was getting double quotes when converting to a string so this is to remove those.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,70 @@
...
+  ConfigAggregator,
+  ConfigFile
+} from '@salesforce/core';
+import * as vscode from 'vscode';
 export class OrgAuthInfo {
   public static async getDefaultUsernameOrAlias(
     vscodePath: string
   ): Promise<string | undefined> {
-    const forceConfig = await new ForceConfigGet().getConfig(
+    if (
```
lcampos: Use `hasRootWorkspace()`

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,70 @@
...
+    return forceConfig.get('defaultusername');*/
   }
 
-  public static async getDefaultDevHubUsernameOrAlias(
-    vscodePath: string
-  ): Promise<string | undefined> {
+  public static async getDefaultDevHubUsernameOrAlias(): Promise<
+    string | undefined
+  > {
+    if (
```
lcampos: Use `hasRootWorkspace()`

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,70 @@
...
   public static async getDefaultUsernameOrAlias(
     vscodePath: string
   ): Promise<string | undefined> {
-    const forceConfig = await new ForceConfigGet().getConfig(
+    if (
+      vscode.workspace &&
+      vscode.workspace.workspaceFolders &&
+      vscode.workspace.workspaceFolders[0]
+    ) {
+      const rootPath = vscode.workspace.workspaceFolders[0].uri.fsPath + '/';
```
lcampos: You can replace this with `getRootWorkspacePath()`. 

I'm not sure building the path like this and adding `.sfdx/` on line 26 will work on windows. Can we replace it by building the uri using `path.join()` ?

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,70 @@
...
       'defaultusername'
     );
-    return forceConfig.get('defaultusername');
+    return forceConfig.get('defaultusername');*/
   }
 
-  public static async getDefaultDevHubUsernameOrAlias(
-    vscodePath: string
-  ): Promise<string | undefined> {
+  public static async getDefaultDevHubUsernameOrAlias(): Promise<
```
lcampos: The comments on `getDefaultUsernameOrAlias` also apply here.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
AnanyaJha: Yeah, that's a good point. I'll use `getRootWorkspacePath()` instead and `path.join()` where needed

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -383,12 +386,14 @@ export async function activate(context: vscode.ExtensionContext) {
     sfdxProjectOpened
   );
 
+  // register org picker commands and set up filewatcher for defaultusername
+  const orgList = new OrgList();
+  context.subscriptions.push(registerOrgPickerCommands(context, orgList));
+  await orgList.registerDefaultUsernameWatcher(context);
```
lcampos: I think we should only register the file watcher if the cli is installed. What do you think ?

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,51 @@
...
   public static async getDefaultUsernameOrAlias(
     vscodePath: string
   ): Promise<string | undefined> {
-    const forceConfig = await new ForceConfigGet().getConfig(
-      vscodePath,
-      'defaultusername'
-    );
-    return forceConfig.get('defaultusername');
+    if (hasRootWorkspace()) {
+      const rootPath = path.join(getRootWorkspacePath(), '/');
```
lcampos: Sorry I wasn't clear. Once you use `path.join` you shouldn't need to add a `/` to the path, that utility should build the correct paths for you. So in line 20 you would only need to call `getRootWorkspacePath()` and then you just use `path.join` for creating the `rootFolder` value.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,51 @@
...
+import {
+  Aliases,
+  AuthInfo,
+  ConfigAggregator,
+  ConfigFile
+} from '@salesforce/core';
+import * as path from 'path';
+import { getRootWorkspacePath, hasRootWorkspace } from './index';
 export class OrgAuthInfo {
   public static async getDefaultUsernameOrAlias(
```
lcampos: Both here and on `getDefaultDevHubUsernameOrAlias` we need to account for the possibility of the local configuration not being in place or incomplete. For example, when you create a new project we do not create a `.sfdx-config.json` file. It's until the user created a scratch org or auths to an org that this gets created.
If the file is present, then there's also a possibility of `defaultusername` and `defaultdevhubusername` not being present.

lcampos:
I left a couple of comments, other than that I think the structure of the changes make sense. Let me know if you want me to look at anything else.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
AnanyaJha: No worries, I'd actually kept it like that originally with only `getRootWorkspacePath()` but I was seeing some weird behavior. When it hit the `path.join` for building the `rootFolder` path, it was getting built without the `/`. So when it was looking for the `rootFolder` path, the result was `/Users/a.jha/Documents/repos/dreamhouse-sfdx.sfdx/sfdx-config.json`. Should I be building the path in a different way?

lcampos: That's weird, I changed it to 
```
const rootPath = getRootWorkspacePath();
const rootFolder = path.join(rootPath, '.sfdx');
```
and `rootFolder` is set to `"/Users/lcamposguajardo/Desktop/testNewOrgPicker/.sfdx"`. Do you have any other changes in your local that could be causing the result to be different ?

AnanyaJha: Ah I think it must not have compiled properly because it still thought the , was a +. All good now, thanks.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
AnanyaJha: That's a good point, I'll add in checks for the file not being present and usernames not being present as well.

packages/salesforcedx-vscode-core/src/index.ts
AnanyaJha: I think that makes sense- I believe we should also only register the org picker commands if the cli is present as well. None of the methods triggered by the filewatcher directly use the cli anymore, but later on the org picker does use `force:config:set` which will fail without the cli

AnanyaJha:
There is a scenario in which the local `defaultusername` is not found, so we switch to using the global `defaultusername`. The user is notified of this behavior with work item W-5856821, but that only displays a message for the global `defaultusername` configuration. In the situation where the local `defaultdevhubusername` is not set either, we also default to using the global value. If we decide to stick with that behavior as well, I think it would be useful to update W-5856821 to also notify the user of the `defaultdevhubusername` setting. The only scratch orgs we display in the quickpick are those associated with the `defaultdevhubusername`, so it could be confusing to only display those associated with the global value.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,28 +4,55 @@
...
-  public static async getDefaultUsernameOrAlias(
-    vscodePath: string
-  ): Promise<string | undefined> {
-    const forceConfig = await new ForceConfigGet().getConfig(
-      vscodePath,
-      'defaultusername'
-    );
-    return forceConfig.get('defaultusername');
+  public static async getDefaultUsernameOrAlias(): Promise<string | undefined> {
+    try {
```
AnanyaJha: If we can't find a local `sfdx-config.json` or there is no `defaultusername`/`defaultdevhubusername` set, we default to calling the global values. If the global default is not found either, I thought it would make sense to log an error explaining that either the `defaultdevhubusername` or `defaultusername` could not be found. If this behavior makes sense, @ruthemmanuelle what would you recommend the message to display?

ruthemmanuelle: @AnanyaJha and I talked about this via Slack. My suggested language: `No default {org|Dev Hub} is set. Run {"SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org"|"SFDX: Authorize a Dev Hub"} to set one.`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -22,7 +23,38 @@ export interface FileInfo {
...
+    this.statusBarItem = vscode.window.createStatusBarItem(
+      vscode.StatusBarAlignment.Left,
+      49
+    );
+    this.statusBarItem.command = 'sfdx.force.set.default.org';
+    this.statusBarItem.show();
+  }
+
+  public async showDefaultOrg(editor?: vscode.TextEditor) {
+    if (!editor) {
```
lcampos: Why do we need this check ? Currently, this will prevent the org picker from showing when you create a new project and customers would have to reload vscode in order to get this. 

This experience (not showing org picker) is not exclusive to creating a new project.

AnanyaJha: Hmm, okay I went back and read the documentation and I'd understood it to mean that it's just looking for the text editor associated with the last touched source files. I assumed that this would be available in all contexts, unless something odd had happened; I didn't realize that a text editor isn't immediately set/there may not be source files available when you create a new project or update the extensions etc. We should be able to take this check out.

Issue #1103, opened 2019-02-25
Switching to Test panel freeze Vs Code when a large number of test classes are present


chulai:
### Summary

Switching to Test panel freeze Vs Code when a large number of test classes are present. It seems the Salesforce Extension waits for all test to be loaded synchronously to return control to Vs Code.

### Steps To Reproduce:

1.  Open large SFDX Project in Vs Code. 1700+ test classes. 20300+ test methods.
1.  Switch to Test Apex Tests Panel.

### Expected result

Test classes and test methods should come up as they are found. Vs Code should remain responsive at all time.

### Actual result

Vs Code freezes for almost 3 minutes until all test are displayed at once. While waiting a "Loading Apex Tests message" appears. Clicking anywhere shows a modal window with the message 'The window is no longer responding. You can reopen or close the window or keep waiting.".

### Additional information

![image](https://user-images.githubusercontent.com/837749/53345116-cffc7c00-38d9-11e9-9a50-e63776c63139.png)

**VS Code Version**: 1.3.1

**SFDX CLI Version**: sfdx-cli/6.54.4-c01dcc69db (windows-x64) node-v8.9.4

**OS and version**: Windows 10, 64-bit (Windows_NT x64 10.0.17763).

praksb:
When we open the test panel we currently try to scan all the classes in a project to get the complete list of test classes and test methods. This doesn't scale well in case of huge number of test classes and test methods and may sometimes result in Out of Memory error as reported in: https://github.com/forcedotcom/salesforcedx-vscode/issues/1101

We should investigate if we can have a better approach that could scale well in scenarios like this.

chulai:
@praksb  I noted you closed issue #1101 which I also reported. As an experiment, I removed all Apex Test classes from the project and that avoided the issue described here. However, the Out of Memory error described here on #1101 still happened. So I don't think the 2 tickets are related.

praksb:
@chulai From the call stack you provided in issue #1101 you are experiencing out of memory exception at startup. We have a service called StandardTestService that compiles all the apex sources at startup and we have another service called ApexIndexer which does that as well at the same time. The StandardTestService also runs when you open up the test panel. Even with the test classes removed we will still be trying to process a huge number of classes twice (once for the StandardTestService and once for the ApexIndexer) which could result in Out of Memory exception. The fix we are working on would address the out of memory exception you see at startup as well as the test panel freeze that you reported in this issue.

[...]

PR #1144, opened 2019-03-12
Improve test refresh performance by querying db instead of recompiling sources

brpowell:
### What does this PR do?

Instead of recompiling the sources to fetch apex tests, query the information from the db instead. Sidebar will wait until the lsp is done doing an initial compile.

#1103 
W-5884817


packages/salesforcedx-vscode-apex/src/languageClientUtils.ts
```diff
@@ -17,6 +17,7 @@ import {
 
 export class LanguageClientUtils {
   private static instance: LanguageClient | undefined;
+  public static indexing = true;
```
brpowell: This is to prevent the refresh command from running before the lsp is finished indexing the project. Otherwise you'll get a message such as "no tests found" which is more confusing than "Loading tests". Another idea I had was maybe hiding the test bar icon until this process is complete and when it's ready show the icon. Thoughts? @ntotten

Issue #973, opened 2018-09-07
Code Coverage Reporting & UI


ntotten:
It would be helpful to developers if they could see test coverage reports and have UI helpers to determine which parts of their code is tested.

A new command **SFDX: Show Test Coverage** could be added that did the following:

1) Ran a test coverage report and showed the results in the terminal
2) Added coverage indicators in the editor gutter as shown below for the C# example.

![image](https://user-images.githubusercontent.com/282782/45238321-7793dc80-b2af-11e8-9a60-19d90c61f308.png)

mnunezdm:
Hi,

I didn't know if should open a new Issue or posting in this one.

After last update Code Coverage shows correctly on Apex Classes, but an annoying notification appears every time a class with no code coverage is opened

This happens with Test Classes, maybe it could be great to ignore this files that does not require any line to test

![testCodeCoverage](https://user-images.githubusercontent.com/10410852/54773321-a2210380-4c09-11e9-8efa-63107c8754c7.gif)

Also, the org selector is now "Lightning" Fast

ntotten:
This feature shipped two weeks ago.

PR #1145, opened 2019-03-13
Visual display of apex code coverage

lcampos:
### What does this PR do?
It takes the info from the test run results that contain code coverage and highlights which lines those test results cover and which they don't.

code-coverage-colorizer

#973


packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (!fs.existsSync(testResultFilePath)) {
+    throw new Error(
+      `No code coverage information was found for test run ${testRunId}`
+    );
+  }
+  const testResultOutput = fs.readFileSync(testResultFilePath, 'utf8');
+  const codeCoverage = JSON.parse(testResultOutput) as CoverageTestResult;
+  if (codeCoverage.coverage === undefined) {
+    throw new Error(
+      `No code coverage information was found for test run ${testRunId}.`
```
lcampos: Error showed when the test results do not include code coverage information.

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
```diff
@@ -0,0 +1,176 @@
...
+  id: string;
+  name: string;
+  totalLines: number;
+  lines: { [key: string]: number };
+};
+
+function getTestRunId() {
+  const testRunIdFile = path.join(apexDirPath, 'test-run-id.txt');
+  if (!fs.existsSync(testRunIdFile)) {
+    throw new Error('No test run information was found for this project.');
```
lcampos: Error when the project does not have any test run information, this would happen if they've never run tests via vscode/cli.

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (editor && isApexMetadata(editor.document.uri.fsPath)) {
+        const codeCovArray = getCoverageData() as CoverageItem[];
+        const codeCovItem = codeCovArray.find(
+          covItem =>
+            covItem.name === getApexMemberName(editor.document.uri.fsPath)
+        );
+
+        if (!codeCovItem) {
+          throw new Error(
+            'No code coverage information was found for the current file.'
```
lcampos: We'll show this when the test results do have code coverage information but not for the apex file currently in focus.

packages/salesforcedx-vscode-apex/src/codecoverage/decorations.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { window } from 'vscode';
```
lcampos: These are the color definitions used (green & red).

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -104,6 +105,14 @@ export async function activate(context: vscode.ExtensionContext) {
 function registerCommands(
   extensionContext: vscode.ExtensionContext
 ): vscode.Disposable {
+  // Colorize code coverage
+  const statusBarToggle = new StatusBarToggle();
+  const colorizer = new CodeCoverage(statusBarToggle);
+  const colorizerCmd = vscode.commands.registerCommand(
+    'sfdx.force.apex.toggle.colorizer',
```
lcampos: Tied everything under one command so that it's easy to trigger the feature using keyboard shortcuts.

packages/system-tests/assets/sfdx-simple/.sfdx/tools/testresults/apex/test-result-707xx00000000BG.json
```diff
@@ -0,0 +1,105 @@
+{
```
lcampos: Had to add these files under `testresults` in order to run the tests. Let me know if you have a better idea for generating these for testing purposes.

packages/salesforcedx-vscode-apex/src/codecoverage/statusBarToggle.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { Disposable, StatusBarItem, window } from 'vscode';
+
+export class StatusBarToggle implements Disposable {
+  private static readonly toggleCodeCovCommand =
+    'sfdx.force.apex.toggle.colorizer';
+  private static readonly showIcon = '$(tasklist)';
+  private static readonly hideIcon = '$(three-bars)';
+  private static readonly toolTip = 'Apex Code Coverage highlighter';
```
lcampos: Text showed when you hover the status bar icon.

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
ruthemmanuelle: What would we recommend that people do next? Set `"salesforcedx-vscode-core.retrieve-test-code-coverage": true`? Or did something else probably cause this error?

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
ruthemmanuelle: Can we add a next step? Something like, 'No test run information was found for this project. Set "salesforcedx-vscode-core.retrieve-test-code-coverage": true in your user or workspace settings, then run Apex tests from the Apex Tests sidebar or from within a test class file.'

packages/salesforcedx-vscode-apex/src/codecoverage/statusBarToggle.ts
ruthemmanuelle: Other nearby hover text seems to use imperative verbs and title-case capitalization. How about this? 'Highlight Apex Code Coverage'

ruthemmanuelle:
Glad to see this, @lcampos . I added some comments on the messages. Also, shouldn't the messages be externalized into messages files?

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -130,6 +139,7 @@ function registerCommands(
     forceApexTestMethodRunCodeAction
   );
   return vscode.Disposable.from(
+    colorizerCmd,
```
praksb: should this be forceApexToggleColorizerCmd to match the rest?

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -104,6 +105,14 @@ export async function activate(context: vscode.ExtensionContext) {
 function registerCommands(
   extensionContext: vscode.ExtensionContext
 ): vscode.Disposable {
+  // Colorize code coverage
+  const statusBarToggle = new StatusBarToggle();
+  const colorizer = new CodeCoverage(statusBarToggle);
+  const colorizerCmd = vscode.commands.registerCommand(
```
praksb: As per the pattern i am seeing for other commands, i think this could be renamed to forceApexToggleColorizerCmd.

praksb:
@lcampos what happens if we try this feature on a stale run result? Let's say that the apex class has since been edited and a bunch of methods have been removed. Does attempting to map nonexistent lines fail gracefully?

lcampos:
@praksb it does fail gracefully. If the code just shifted but the lines that will be colored are still in the range of the document, then we will just color whatever code, comment, empty space available. If the  document's length has shortened, then it will just display a warning about it if the lines that need to be colored are not part of the range of the document.
I'll add a message to let the customer know they need to refresh their code coverage run but it will only be displayed in the last case described above since we do not do further validations to the code in the file or code coverage run. I wanted to keep the feature as simple as possible in the first version and then see if we needed to add more validations via apex lsp, etc.

lcampos:
@ruthemmanuelle thanks for the review, yes we will have these messages on the extension's `i18n.ts` file. Since I didn't know how many different labels we were gonna need I just kept it on the code.

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
lcampos: Yes, running tests without that setting on.

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -40,5 +40,14 @@ export const messages = {
 
   client_name: 'Apex Language Server',
   cannot_determine_workspace:
-    'Unable to determine workspace folders for workspace'
+    'Unable to determine workspace folders for workspace',
```
lcampos: @ruthemmanuelle here are the new error messages, let me know if any changes are needed.

PR #1151, opened 2019-03-14
Update node version check and remove unneccesary steps in publish script

lcampos:
### What does this PR do?
Updates our publish script to use nodejs version 10.2.0 and removes steps that are now obsolete as a result of upgrading to a newer prettier version.

@W-5942633@

### DIFF
--- a/scripts/publish.js
+++ b/scripts/publish.js
@@ -17,13 +17,13 @@ shell.set('+v');
  *
  */
 
-// Checks that you are running this with Node v8.9.0 and above
+// Checks that you are running this with Node v10.2.0 and above
 const [version, major, minor, patch] = process.version.match(
   /^v(\d+)\.?(\d+)\.?(\*|\d+)$/
 );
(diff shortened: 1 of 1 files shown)


scripts/publish.js
```diff
@@ -122,11 +122,5 @@ shell.exec(`git add SHA256.md`);
 // Git commit
 shell.exec(`git commit -m "Updated SHA256"`);
 
-// Add formatting changes
```
lcampos: These steps now cause the script to stop/fail because there's nothing to add and commit as a result of the changes to the new prettier version. It used to be that we would loose some formatting when updating the package.json version number (https://github.com/forcedotcom/salesforcedx-vscode/commit/4445c6099bdcea45e5f405578bdb3ec72b63182a), that doesn't happen anymore (https://github.com/forcedotcom/salesforcedx-vscode/commit/3cd305db8d9de49d5966c73bfe958b47a3b32f29).

PR #1160, opened 2019-03-18
Aura Language Server

nrkruk:
initial PR for delivery of the aura language server


lcampos:
Why are we adding another vscode extension (`salesforcedx-vscode-aura`) instead of using the existing one (`salesforcedx-vscode-lightning`) ?

packages/salesforcedx-vscode-aura/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
lcampos: We already have a top level `.prettierrc` that applies to the rest of the monorepo. Is there a specific reason a different config is needed ?

packages/salesforcedx-vscode-aura/LICENSE.txt
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2018, Salesforce.com, Inc.
```
lcampos: Shouldn't this be 2019 ?

packages/salesforcedx-vscode-aura/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "url": "https://github.com/forcedotcom/salesforcedx-vscode"
+  },
+  "icon": "images/VSCodeLWC.png",
+  "galleryBanner": {
+    "color": "#ECECEC",
+    "theme": "light"
+  },
+  "version": "45.6.0",
+  "publisher": "salesforce",
+  "preview": true,
```
lcampos: Why are we releasing it as preview ?

packages/salesforcedx-vscode-aura/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "Programming Languages"
+  ],
+  "dependencies": {
+    "@salesforce/salesforcedx-utils-vscode": "45.6.0",
+    "chalk": "^2.4.2",
+    "change-case": "^3.1.0",
+    "lightning-lsp-common": "1.0.4",
+    "lwc-language-server": "1.8.3",
+    "opn": "^5.4.0",
+    "vscode": "1.1.17",
```
lcampos: If this dependency is being used exclusively for testing then we can move it to `devDependencies`.

packages/salesforcedx-vscode-aura/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
+    "postinstall": "node node_modules/.bin/vscode-install",
+    "pretest": "node ../../scripts/download-vscode-for-system-tests",
+    "test": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-test-with-top-level-extensions",
+    "test:insiders": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && cross-env CODE_VERSION=insiders node ../../scripts/run-test-with-top-level-extensions",
+    "test:unit": "node ../../scripts/install-vsix-dependencies dbaeumer.vscode-eslint && node ../../scripts/run-test-with-top-level-extensions"
+  },
+  "activationEvents": [
+    "workspaceContains:sfdx-project.json",
+    "workspaceContains:**/core/workspace-user.xml"
```
lcampos: Where is `/core/workspace-user.xml` generated from ? Is it so it works for internal devs as well ?

packages/salesforcedx-vscode-aura/package.json
```diff
@@ -0,0 +1,104 @@
...
+  "categories": [
+    "Programming Languages"
+  ],
+  "dependencies": {
+    "@salesforce/salesforcedx-utils-vscode": "45.6.0",
+    "chalk": "^2.4.2",
+    "change-case": "^3.1.0",
+    "lightning-lsp-common": "1.0.4",
+    "lwc-language-server": "1.8.3",
+    "opn": "^5.4.0",
```
lcampos: `chalk`, `change-case` and `opn` are new to the project but they are all on MIT licenses. @ntotten do we need to do something regarding 3PP, etc ?

packages/salesforcedx-vscode-aura/src/views/lwc-node.ts
```diff
@@ -0,0 +1,50 @@
...
+    get contextValue() {
+        if ((this.type === NodeType.Component || this.type === NodeType.WebComponent) && !this.uri) {
+            return 'external';
+        }
+        return undefined;
+    }
+
+    get iconPath() {
+        if (this.type === NodeType.Namespace) {
+            return path.join(__dirname, '../../../resources/namespace.svg');
```
lcampos: Doesn't this break on windows ? You can use `path.join` to take care of building the uri in correct format depending on the platform being used, e.g.:

```
path.join(__dirname, '..', '..', '..', 'resources', 'namespace.svg');
```

packages/salesforcedx-vscode-aura/test/extension.test.ts
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: Update copyright to 2019

packages/salesforcedx-vscode-aura/package.json
ntotten: Let's not do preview. This is miles ahead of the current version even.

packages/salesforcedx-vscode-aura/package.json
ntotten: Yes, we need to.

ntotten: Actually, we should not be using `opn` at all. VS Code is asking developers to remove the use of that library and use their API: https://code.visualstudio.com/updates/v1_31#_open-resources-in-a-browser

nrkruk:
@lcampos we wanted to add the aura extension initially in isolation from the lightning stylesheets stuff.  We planned to combine the two, I was just thinking to do that after the initial merge, but I didn't think about the fact that this would technically be an additional plugin. I will merge the two now and update the PR

packages/salesforcedx-vscode-aura/package.json
nrkruk: Yeah, I'll remove this when I merge with the lightning package

packages/salesforcedx-vscode-aura/package.json
nrkruk: Yes, this is strictly an internal dev thing

packages/salesforcedx-vscode-aura/src/views/lwc-node.ts
nrkruk: Good catch. We haven't tested much on windows - is there a windows machine you guys have access to that I can use?

packages/salesforcedx-vscode-aura/src/commands/quickpick/quickpick.ts
```diff
@@ -0,0 +1,55 @@
...
+import { LanguageClient } from 'vscode-languageclient';
+import { LwcNode } from '../../views/lwc-node';
+import { TagItem } from './tag-item';
+
+export function createQuickOpenCommand(client: LanguageClient) {
+    return (arg: any) => {
+        if (arg instanceof LwcNode) {
+            const node = arg as LwcNode;
+            const tag: string = node.label;
+            opn(`https://developer.salesforce.com/docs/component-library/bundle/${tag}`);
```
lcampos: We normally have urls on a `constants.ts` per module. If it's part of a message or it needs to be dynamically modified then we also use `i18n.ts` under each package's `messages` folder. I think it's up to you which of these options we go with for this.

packages/salesforcedx-vscode-aura/src/commands/quickpick/quickpick.ts
```diff
@@ -0,0 +1,55 @@
...
+                                items.push(new TagItem(key, '', value.documentation, Uri.parse(uri)));
+                            }
+                        }
+                        const sorted = items.sort((a, b) => (a.label < b.label ? -1 : 1));
+                        window.showQuickPick(sorted, {}).then((data: TagItem) => {
+                            workspace.openTextDocument(data.uri).then(doc => window.showTextDocument(doc));
+                        });
+                    },
+                    err => {
+                        console.error('Could not request lwc/listComponents - is the LSP the correct version?');
```
lcampos: I think we should also log this error and the one on line 51 using the telemetry service.

packages/salesforcedx-vscode-aura/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+            { language: 'javascript', scheme: 'untitled' }
+        ],
+        synchronize: {
+            fileEvents: [
+                workspace.createFileSystemWatcher('**/*.resource'),
+                workspace.createFileSystemWatcher('**/labels/CustomLabels.labels-meta.xml'),
+                workspace.createFileSystemWatcher('**/aura/*/*.{cmp,app,intf,evt,js}'),
+                workspace.createFileSystemWatcher('**/components/*/*/*.{cmp,app,intf,evt,lib,js}'),
+                // need to watch for directory deletions as no events are created for contents or deleted directories
+                workspace.createFileSystemWatcher('**/', true, true, false),
```
lcampos: Do we have to listen to all the directories or can it be scoped to only aura related ones ?

packages/salesforcedx-vscode-aura/src/views/lwc-node.ts
lcampos: Some of us have personal windows machines, the rest of the team just spins up their own Windows VMs. cc @ntotten

packages/salesforcedx-vscode-aura/test/extension.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        await matchExtensionAsHtml('.tokens');
+    });
+});
+
+describe('Test commands', () => {
+    let coreExtension: vscode.Extension<any>;
+    let auraExtension: vscode.Extension<any>;
+
+    before(async () => {
+        if (vscode.workspace.rootPath) {
```
lcampos: `rootPath` is deprecated, you can replace it with `vscode.workspace.workspaceFolders![0]`.

packages/salesforcedx-vscode-lwc/src/dxsupport/waitForDX.ts
```diff
@@ -0,0 +1,40 @@
+import * as vscode from 'vscode';
+
+let wait: Promise<vscode.Extension<any>>;
+
+export async function waitForDX(activate: boolean = false) {
```
lcampos: This is unreliable since activation times vary based on the machine's resources. Why do we want to activate and then wait for `salesforcedx-vscode-core` to be present instead of having a direct dependency with `salesforcedx-vscode-core`? Doing so will trigger activation of this extension after core is active.

packages/salesforcedx-vscode-lwc/src/telemetry/telemetry.ts
```diff
@@ -44,15 +74,17 @@ export class TelemetryService {
     }
   }
 
-  public sendExtensionDeactivationEvent(): void {
+  public async sendExtensionDeactivationEvent(): Promise<void> {
+    await this.setupVSCodeTelemetry();
```
lcampos: This is not needed if the telemetry setup is triggered during the activation phase of the extension.

packages/salesforcedx-vscode-aura/src/commands/quickpick/quickpick.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function createQuickOpenCommand(client: LanguageClient) {
+    return (arg: any) => {
+        if (arg instanceof LwcNode) {
+            const node = arg as LwcNode;
+            const tag: string = node.label;
+            opn(`https://developer.salesforce.com/docs/component-library/bundle/${tag}`);
+            return;
+        }
+        console.log('Waiting for LSP to be ready');
```
lcampos: If these type of messaging is needed across the extension I'd suggest you add all the text on a `i18n.ts` file. It makes it easier for our doc writers to contribute and for translating the text (in the future). You can see examples of this on each package's `messages` folder.

packages/salesforcedx-vscode-aura/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    // Create the language client and start the client.
+    this.client = client = new LanguageClient('auraLanguageServer', 'Lightning Language Server', serverOptions, clientOptions);
+    // UI customizations
+    context.subscriptions.push(commands.registerCommand('salesforce-lightning-quickopen', createQuickOpenCommand(client)));
+    const componentProvider = new ComponentTreeProvider(client, context);
+    window.registerTreeDataProvider('salesforce-lightning-components', componentProvider);
+
+    client.onReady().then(() => {
```
lcampos: It would be good to track how long it takes for the server to start, error, etc. We can do it similar to how it's done for the Apex LSP https://github.com/forcedotcom/salesforcedx-vscode/blob/51a10f151ef4328b78642b1cdc587fffd6c6d45f/packages/salesforcedx-vscode-apex/src/index.ts#L87

packages/salesforcedx-vscode-lwc/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
lcampos: Same as with the other `.prettierrc` in `salesforcedx-vscode-aura`.

SHA256.md
```diff
@@ -16,20 +16,19 @@ shasum -a 256 <location_of_the_downloaded_file>
```
lcampos: This file only has format changes, let's remove it from this PR.

packages/salesforcedx-vscode-aura/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "sinon": "^2.3.6"
+  },
+  "scripts": {
+    "vscode:prepublish": "npm prune --production",
+    "vscode:package": "vsce package",
+    "compile": "tsc -p ./",
+    "lint": "tslint --project .",
+    "watch": "tsc -watch -p .",
+    "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
+    "postinstall": "node node_modules/.bin/vscode-install",
```
lcampos: This target fails when running it on windows (after running `npm install`) but it doesn't on a mac.

```
lerna ERR! npm run postinstall [salesforcedx-vscode-aura] Output from stderr:
npm WARN lifecycle The node binary used for scripts is C:\Users\lcamp\Documents\github\salesforcedx-vscode\node_modules\node\bin\node.exe but npm is using C:\Program Files\nodejs\node.exe itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
C:\Users\lcamp\Documents\github\salesforcedx-vscode\packages\salesforcedx-vscode-aura\node_modules\.bin\vscode-install:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
    at new Script (vm.js:74:7)
    at createScript (vm.js:246:10)
    at Object.runInThisContext (vm.js:298:10)
    at Module._compile (internal/modules/cjs/loader.js:670:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:10)
    at startup (internal/bootstrap/node.js:238:19)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! salesforcedx-vscode-aura@45.6.0 postinstall: `node node_modules/.bin/vscode-install`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the salesforcedx-vscode-aura@45.6.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

lcampos:
@nrkruk running `npm run compile` fails in both Windows and Mac with the error below. Am I missing some steps before spinning up the extensions ?

```
salesforcedx-vscode-lwc: src/index.ts(8,10): error TS2305: Module '"/Users/lcamposguajardo/github/DevTools/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/node_modules/lightning-lsp-common/lib/shared"' has no exported member 'shared'.
```

packages/salesforcedx-vscode-lwc/src/dxsupport/waitForDX.ts
nrkruk: There isn't much that we use from the salesforcedx-vscode-core plugin (the only thing we need is the TelemetryService and the createLwcComponent command, both of which the language server can function without). Avoiding the direct dependency allows us to do things immediately (like indexing / content assist, for example). Otherwise we have to wait until core is loaded before the language server can do anything.

nrkruk: How big of a sticking point is this? It seems like an optimization, but if you have a use case where you think this would cause bugs, let me know

SHA256.md
```diff
@@ -32,4 +32,4 @@ d7b0af158a166d1c89fc91845fb7baa1e0d46da71e116e9b6aaf44ef5cb1e83b  salesforcedx-v
 5. In Visual Studio Code, from the Extensions view, select ... > Install from
 VSIX.
 
-6. Install the verified VSIX file.
+6. Install the verified VSIX file.
```
nrkruk: I have no idea what is going on but I can't get this file out of the PR. I've tried 3 times now. I give up

packages/salesforcedx-vscode-lwc/src/dxsupport/waitForDX.ts
midzelis: This has two purposes: 1) improve startup performance, 2) make the dependency optional. 

If we have a direct dependency, dx will have to active before we start to activate, slowing down the entire activation process. 

In a core scenario (working on lightning/lwc on a salesforce core workspace) we do not need the vscode-core dependency at all. It would be wrong, actually, since the lightning commands that we use from the toolbelt (like create lightning component) wouldn't work in a core env. 

We can increase the length of time we wait for DX to start. Hopefully it starts within 2 minutes, but we can make it 5 or 10 or whatever you think is appropriate.

ntotten: Regarding the startup time, the release that just went out today pretty much eliminated any startup time - Luis basically removed everything from the activation and loaded everything async. 

Regarding the optional dependency on core, I'm wondering if we can still require it to be installed but basically, it won't do anything - the core extension shouldn't do anything unless it detects an `sfdx-project.json` so it wouldn't add any functionality for core devs. Correct me if I'm wrong luis.

lcampos: Looks like the real concern for adding this was performance. Has this been tested against the latest changes in `develop` branch ? This week's changes significantly improved the performance of `salesforcedx-vscode-core` activation/start up.

lcampos: I also do agree that having a direct dependency in core just because of the telemetry functionality is a bit too much. Might be a good time to start exploring moving that into its own utility module.

package.json
```diff
@@ -41,7 +41,8 @@
     "reformat": "node scripts/reformat-with-prettier.js",
     "snyk:test": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk test --severity-threshold=medium --show-vulnerable-paths=false",
     "snyk:monitor": "lerna exec --bail=false --ignore salesforcedx-vscode -- snyk monitor --severity-threshold=medium --show-vulnerable-paths=false --org=vazexqi",
-    "aggregateJUnit": "node scripts/aggregate-junit-xml.js"
+    "aggregateJUnit": "node scripts/aggregate-junit-xml.js",
+    "link-lsp": "lerna exec yarn link aura-language-server lwc-language-server lightning-lsp-common --scope salesforcedx-vscode-lightning && lerna exec yarn link lwc-language-server lightning-lsp-common --scope salesforcedx-vscode-lwc"
```
lcampos: This project currently uses `npm`, can the use of `yarn` here be replaced ?

packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -24,13 +24,18 @@
     "Programming Languages"
   ],
   "dependencies": {
-    "@salesforce/salesforcedx-slds-linter": "45.7.0",
```
lcampos: Does this mean that the existing slds functionality is being replaced by either `aura-language-server` or `lightning-lsp-common` ?

Since the use of `salesforcedx-slds-linter` is going away, can we also delete that module ?

packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -82,6 +88,40 @@
         "scopeName": "text.html.basic",
         "path": "./syntaxes/html.tmLanguage.json"
       }
+    ],
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "salesforce-lightning-explorer",
+          "title": "Lightning Explorer",
```
lcampos: The text displayed to the end user needs to be replaced by a label that is defined on a `package.nls.json` file, e.g. https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.nls.json
This applies to line 96, 105 and 121

packages/salesforcedx-vscode-lightning/src/commands/quickpick/quickpick.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+import { paramCase } from 'change-case';
+import opn = require('opn');
+import { Uri, window, workspace } from 'vscode';
+import { LanguageClient } from 'vscode-languageclient';
+import { COMPONENT_LIBRARY_BUNDLE_LINK } from '../../constants';
+import { LwcNode } from '../../views/lwc-node';
+import { TagItem } from './tag-item';
+
+export function createQuickOpenCommand(client: LanguageClient) {
```
lcampos: It would be good to add telemetry to track the language server performance e.g. https://github.com/forcedotcom/salesforcedx-vscode/blob/79f718244dd0712836898ce5a51b659e9127f370/packages/salesforcedx-vscode-apex/src/index.ts#L77

packages/salesforcedx-vscode-lightning/package.json
nrkruk: lightning-lsp-common is just a package with common code for both aura and lwc lsps. We are removing the slds-linter as per @ntotten 's request. There is a new version in the works that I believe will be put in soon. But for now the existing one doesn't work and hasn't been maintained. The new slds linter probably belongs in its own package I would think.

I can remove the slds-linter package as well, just wasn't sure if you'd be ok with that or if I should leave it until its fully replaced

nrkruk: To be clear though, with my existing changes, the slds-linter will no longer be present in VSCode

packages/salesforcedx-vscode-aura/src/views/lwc-node.ts
nrkruk: I've fixed these and ones I found in the LSPs, just need to test on windows to make sure ternjs in the language server is working correctly

packages/salesforcedx-vscode-aura/src/commands/quickpick/quickpick.ts
nrkruk: I created a constants file

nrkruk:
Closed this PR and reopend as #1183 so the automation will run correctly

packages/salesforcedx-vscode-lightning/package.json
ntotten: > I can remove the slds-linter package as well, just wasn't sure if you'd be ok with that or if I should leave it until its fully replaced

Let's remove it. The replacement is a complete rewrite anyway.

PR #1164, opened 2019-03-19
Fix deployResultParser for oclif compatibility 

brpowell:
### What does this PR do?
JSON response slightly differs with oclif - updating the parser to handle pre and post oclif.

W-5944509

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
@@ -37,15 +37,8 @@ export interface ForceSourceDeploySuccessResponse {
 export class ForceDeployResultParser {
   private response: any;
 
-  constructor(stdOut: string) {
-    const stdErrLines = stdOut.split(EOL);
-    for (const line of stdErrLines) {
-      if (line.trim().startsWith('{')) {
-        this.response = JSON.parse(line);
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -37,15 +37,8 @@ export interface ForceSourceDeploySuccessResponse {
 export class ForceDeployResultParser {
   private response: any;
 
-  constructor(stdOut: string) {
-    const stdErrLines = stdOut.split(EOL);
```
lcampos: Looks like the `EOL` import can be removed too.

packages/salesforcedx-utils-vscode/src/cli/deployResultParser.ts
```diff
@@ -37,15 +35,18 @@ export interface ForceSourceDeploySuccessResponse {
...
-    for (const line of stdErrLines) {
-      if (line.trim().startsWith('{')) {
-        this.response = JSON.parse(line);
-        return;
-      }
+  constructor(stdout: string) {
+    try {
+      const sanitized = stdout.substring(
```
brpowell: stdout should almost never be anything but the json response, but just in case.

PR #1167, opened 2019-03-20
Fix autobuild issue with not being able to connect to codecov

AnanyaJha:
### What does this PR do?
This PR fixes an issue with lerna failing the build due to a system-test failure. 
Also, when uploading results to codecov fails (ie. network issues), I found that the build does not actually fail- it just displays the failure as a `FetchError`. Since the build does not fail, this pr just adds an `appveyor-retry` command (retry up to 3x) for uploading results to codecov. 

@W-5959903@


sfsholden:
@AnanyaJha Hey Ananya, I don't think this matters that much but I think it'd be good to keep the argument '--scope system-tests' for the 'coverage:system-tests' script in package.json. The scope tells lerna to only execute the script for the system-tests package. Without it, lerna exec is trying to execute 'coverage:system-tests' for all packages.

Issue #998, opened 2019-01-31
Create Default Folder for Template if not Exists


ntotten:
When you run a create command with a template, the default folder is the default option for where to create the file, however, if that folder doesn't exist the command doesn't work. Instead, if you select the default folder (or really any folder), the folder should be created if it doesn't exist.

Issue #852, opened 2019-01-08
Template Commands should be surfaced on only default folders.


ntotten:
Currently, the "SFDX: Create X" commands are shown on every folder in the solution. This is because we allow users to have a flexible structure to their project. However, this is resulting in a very overloaded right-click menu. The proposal is to change it so that the "SFDX: Create X" commands only show up on the default folder for each type. For example, "SFDX: Create Lighting Component" only shows up on the aura folder. "SFDX: Create Apex Class" only shows on the "classes" folder.

PR #1187, opened 2019-03-24
New menu for selecting output directory for "create" commands

brpowell:
### What does this PR do?

Revises the menu for selecting an output directory for the create commands that generates templates. Users are now prompted with a list of default output options. If they want to choose a custom directory, they can select the last option of the menu to show a list of all directories in the project. For components that require their parent folder to be named something specific, it is automatically created whether the chosen output directory has that name or not. 

W-5815448
#852 #998


packages/salesforcedx-vscode-core/src/commands/forceLightningAppCreate.ts
```diff
@@ -92,7 +88,7 @@ const fileNameGatherer = new SelectFileName();
 const lightningFilePathExistsChecker = new LightningFilePathExistsChecker();
 
 export async function forceLightningAppCreate(explorerDir?: any) {
-  const outputDirGatherer = new SelectStrictDirPath(explorerDir, 'aura');
+  const outputDirGatherer = new SelectOutputDir('aura', true);
```
brpowell: Because aura bundles require the "aura" parent folder, we let the constructor know this requirement by specifying "true".

packages/salesforcedx-vscode-core/src/sfdxProject/sfdxPackageDirectories.ts
```diff
@@ -27,7 +27,11 @@ export default class SfdxPackageDirectories {
             if (dirPath.startsWith(path.sep)) {
               dirPath = dirPath.substring(1);
             }
-            packageDirectoryPaths.push(dirPath);
+            if (packageDirectory.default) {
+              packageDirectoryPaths = [dirPath].concat(packageDirectoryPaths);
```
brpowell: Slight modification to have the default package directory in a project be the first element of the list.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcCreate.ts
```diff
@@ -128,10 +128,10 @@ const fileNameGatherer = new SelectFileName();
 const lightningFilePathExistsChecker = new LightningFilePathExistsChecker();
 
 export async function forceLightningLwcCreate(explorerDir?: any) {
-  const outputDirGatherer = new SelectStrictDirPath(explorerDir, 'lwc');
+  const outputDirGatherer = new SelectOutputDir('lwc', true);
   const parameterGatherer = new CompositeParametersGatherer(
-    outputDirGatherer,
-    fileNameGatherer
+    fileNameGatherer,
+    outputDirGatherer
```
brpowell: Switched the order of these gatherers to keep the experience consistent with the other create commands.

brpowell:
@lcampos @ntotten Should we add telemetry data to measure when default folders are chosen vs. custom? I did something similar for the SObject refresh by appending the source to the command execution event - would that be sufficient?

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -216,85 +217,72 @@ export class SelectFileName
...
   implements ParametersGatherer<{ outputdir: string }> {
-  private explorerDir: string | undefined;
-  private globKeyWord: string | undefined;
-
-  public constructor(explorerDir?: vscode.Uri, globKeyWord?: string) {
-    this.explorerDir = explorerDir ? explorerDir.fsPath : explorerDir;
-    this.globKeyWord = globKeyWord;
+  private typeDir: string;
+  private typeDirRequired: boolean | undefined;
+  public readonly defaultOutput = '/main/default';
```
lcampos: I think this is going to break on Windows.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -216,85 +217,72 @@ export class SelectFileName
...
-export class SelectStrictDirPath extends SelectDirPath {
-  public globDirs(srcPath: string, priorityKeyword?: string): string[] {
-    const globPattern = priorityKeyword
-      ? path.join(srcPath, '**/', priorityKeyword + '/')
-      : path.join(srcPath, '**/');
-    const relativeDirs = new glob.GlobSync(globPattern).found.map(value => {
-      let relativePath = path.relative(srcPath, path.join(value, '/'));
-      relativePath = path.join(relativePath, '');
+  public getCustomOptions(rootPath: string): string[] {
+    return new glob.GlobSync(path.join(rootPath, '**/')).found.map(value => {
```
lcampos: I think this is going to break on Windows, can you verify ?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -235,5 +235,6 @@ export const messages = {
   error_no_default_username:
     'No default org is set. Run "SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org" to set one.',
   error_no_default_devhubusername:
-    'No default Dev Hub is set. Run "SFDX: Authorize a Dev Hub" to set one.'
+    'No default Dev Hub is set. Run "SFDX: Authorize a Dev Hub" to set one.',
+  custom_output_directory: 'Select a Custom Directory'
```
lcampos: cc @ruthemmanuelle

packages/salesforcedx-vscode-core/src/commands/commands.ts
brpowell: Yikes good catch

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -216,85 +217,72 @@ export class SelectFileName
...
-
   public async gather(): Promise<
     CancelResponse | ContinueResponse<{ outputdir: string }>
   > {
-    const rootPath = getRootWorkspacePath();
-    let outputdir;
-    if (rootPath) {
+    let packageDirs: string[] = [];
+    try {
+      packageDirs = await SfdxPackageDirectories.getPackageDirectoryPaths();
```
lcampos: Cool handling for multiple packages.

packages/salesforcedx-vscode-core/src/commands/forceApexTriggerCreate.ts
```diff
@@ -90,10 +86,7 @@ const fileNameGatherer = new SelectFileName();
 const filePathExistsChecker = new FilePathExistsChecker(APEX_TRIGGER_EXTENSION);
 
 export async function forceApexTriggerCreate(explorerDir?: any) {
```
lcampos: Do we need to keep the `explorerDir` parameter now that's not being used ? Same question applies to the other create commands being modified.

brpowell: I don't think so? I wasn't really sure what it was being used for in the first place...

lcampos:
@brpowell I do think adding telemetry to see how the feature is used would help. I know we already have some to track the execution time. You can either add more details to that one or add a new log line, it's up to you.

ntotten:
> @lcampos @ntotten Should we add telemetry data to measure when default folders are chosen vs. custom? I did something similar for the SObject refresh by appending the source to the command execution event - would that be sufficient?

I don't think we need to collect that specifically. I'm not sure what it will really tell us? I would be interested in perf numbers though on how long creates take so we can figure out if any of them are slow, but i think we can do that as separate work.

packages/salesforcedx-vscode-core/src/commands/commands.ts
ntotten: @lcampos do we need a path utility in our core library maybe? This seems to be a common problem. Maybe if we prevented the use of `require('path`)` and used our own utility, it would cut down on these windows bugs?

ntotten:
@brpowell Does the new aura extension that was just merged on friday also need this change applied to it? //cc @nrkruk @lcampos

packages/salesforcedx-vscode-core/src/commands/commands.ts
lcampos: No, using `path` is the right way to address this since its a Nodejs module and it does a good job across platforms.

lcampos:
> @brpowell Does the new aura extension that was just merged on friday also need this change applied to it? //cc @nrkruk @lcampos

These changes already cover the create commands for `aura` and `lwc`.

packages/salesforcedx-vscode-core/src/commands/forceApexClassCreate.ts
```diff
@@ -56,7 +57,12 @@ class ForceApexClassCreateExecutor extends SfdxCommandletExecutor<
     }).execute(cancellationToken);
 
     execution.processExitSubject.subscribe(async data => {
-      this.logMetric(execution.command.logName, startTime);
+      const dirType = response.data.outputdir.endsWith(
```
brpowell: These template commands badly need to be refactored to avoid having so much duplicated code. I started doing so but stuck it in another branch to separate it from this work. I'll make a story for it.

packages/salesforcedx-vscode-core/src/commands/forceLightningInterfaceCreate.ts
```diff
@@ -26,13 +26,14 @@ import {
...
   SelectFileName,
-  SelectStrictDirPath,
+  SelectOutputDir,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
 const LIGHTNING_INT_EXTENSION = '.intf';
+const LIGHTNING_METADATA_DIR = 'aura';
```
JimSuplizio: Bryan, I'm seeing a whole bunch of LIGHTNING_METADATA_DIR constant definitions across multiple files (among other LIGHTNING_*_EXTENSION constants). We should create a file for these metadata constants and store them in there especially if we're going to be handling other metadata types in a similar fashion.

brpowell: Yeah I very much agree. I'm going to create a refactor story for these template commands and as part of it I want to create a configuration file the commands can read. It'll contain things like file extension, metadata dir name, requires subdirectory. Since the refactoring has quite a few changes I wanted to keep it separate from this work.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcCreate.ts
lcampos: I would expect this change to break `Should create Lightning LWC` on the `scaffolding.test.ts` I fixed a few days ago.

Issue #918, opened 2018-08-21
SFDX: Retrieve Source from Org with command palette


takahitomiyamoto:
I would like you to add this feature because of more quick coding:
we can run the command “SFDX: Retrieve Source from Org” with command palette.

ntotten:
This has been fixed in the latest release. Please reopen if you see problems.

Issue #662, opened 2018-10-05
Cannot read property 'fspath' of undefined --> when using a Keyboard shortcut to deploy Salesforce file


pranaymistry:
### I added a custom keyboard shortcut to deploy current file using "sfdx.force.source.deploy.current.file" command 

One of our business requirements is to figure out a keyboard shortcut that can do Save and compile of Apex classes from VSCode while doing ORG based development. This is very important for us to roll out VSCode and drive adoption of its use for Salesforce development. This feature is currently not available in VSCode and hence I tried adding a keyboard shortcut for the sfdx command but I am running into an issue. 

### Steps To Reproduce:

1. Go to **Code** --> **Preferences** --> **Keyboard Shortcuts**

<img width="485" alt="screen shot 2018-10-05 at 9 17 56 am" src="https://user-images.githubusercontent.com/40212486/46540656-96ba6580-c87f-11e8-9a3b-dbb846e4dc87.png">

2. Open **keybindings.json** file 

<img width="439" alt="screen shot 2018-10-05 at 9 19 37 am" src="https://user-images.githubusercontent.com/40212486/46540764-d84b1080-c87f-11e8-9f91-d86d57042e29.png">

3. Add the below command in your keybindings.json file

`[
    {
        "key": "shift+cmd+r",
        "command": "sfdx.force.source.deploy.current.file"
    }
]`

4. This is how your keybindings.json file should look like 

<img width="561" alt="screen shot 2018-10-05 at 9 20 54 am" src="https://user-images.githubusercontent.com/40212486/46540821-0a5c7280-c880-11e8-9921-5c8e68049ccb.png">

5. Now open an Apex class or Visualforce page or Trigger file and press Shift+Cmd+r

### Expected result

The file should have been deployed

### Actual result

<img width="511" alt="screen shot 2018-10-05 at 9 23 21 am" src="https://user-images.githubusercontent.com/40212486/46540954-5a3b3980-c880-11e8-97b7-d04dd0bda32c.png">

**VS Code Version**: 

<img width="284" alt="screen shot 2018-10-05 at 9 24 51 am" src="https://user-images.githubusercontent.com/40212486/46541025-8a82d800-c880-11e8-9f46-82866c689ee7.png">

**SFDX CLI Version**:

[...]

PR #1190, opened 2019-03-25
Enable commands for commandPalette

JimSuplizio:
### What does this PR do?
This PR is linked to two different github issues:
https://github.com/forcedotcom/salesforcedx-vscode/issues/662
https://github.com/forcedotcom/salesforcedx-vscode/issues/918

This PR affects the following commands:
- SFDX: Deploy Source in Manifest to Org 
- SFDX: Retrieve Source in Manifest from Org
- SFDX: Delete from Project and Org
- SFDX: Deploy Source to Org
- SFDX: Retrieve Source From Org
The commands were previously only active with the explorer/context (right clicking in file explorer) and editor/context (right clicking in the open file in the editor). Enabling these for the commandPalette in VS Code, for the currently open file, will also enable these commands to be shortcut. The caveat here is that these commands have a **when** clause which restricts when the commands can be run. If the user is defining shortcuts for any commands that currently have a when clause then we need to provide guidance that they should also lift the **when** clause from the package.json that the command is defined in.

@W-5803870@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -326,27 +326,27 @@
         },
         {
           "command": "sfdx.force.source.retrieve.current.source.file",
-          "when": "false"
+          "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest'"
```
allileong: For these commands that are available when `resourceLangId != 'forcesourcemanifest'`, you can also add `editorIsOpen` to the when clause to restrict the command from being available when no file is open in the editor. 

For the commands that require `resourceLangId == 'forcesourcemanifest'`, `editorIsOpen` isn't necessary.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -34,6 +34,12 @@ export class ForceSourceDeployManifestExecutor extends BaseDeployExecutor {
 }
 
 export async function forceSourceDeployManifest(manifestUri: vscode.Uri) {
+  if (!manifestUri) {
+    const editor = vscode.window.activeTextEditor;
+    if (editor) {
```
lcampos: This will not always be `undefined` when triggering the command using keyboard shortcuts and the user doesn't have an open document. You can leverage the existing language id and check for it e.g.`if (editor && editor.document.languageId === 'forcesourcemanifest')`. We should also throw an informative error message when it's not a manifest file or the users will keep getting the `Cannot read property 'fsPath' of undefined` warning message.

packages/salesforcedx-vscode-core/package.json
JimSuplizio: Good call @allileong I'll get this updated with the next PR.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -81,10 +81,20 @@ export const messages = {
     'SFDX: Push Source to Default Scratch Org and Override Conflicts',
 
   force_source_deploy_text: 'SFDX: Deploy Source to Org',
+  force_source_deploy_select_file_or_directory:
```
JimSuplizio: @ruthemmanuelle I've got some user facing messages for you to review please.

ntotten:
@JimSuplizio @ruthemmanuelle This PR has docs implications. We should try to update the docs in the same PR if possible.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -54,6 +57,21 @@ export class MultipleSourcePathsGatherer implements ParametersGatherer<string> {
 }
 
 export async function forceSourceDeploySourcePath(sourceUri: vscode.Uri) {
+  if (!sourceUri) {
+    const editor = vscode.window.activeTextEditor;
+    if (editor && editor.document.languageId !== 'forcesourcemanifest') {
```
lcampos: Only the commands that deal with manifest files should have a check for the language id being forcesourcemanifest e.g. `forceSourceDeployManifest.ts`

lcampos: Update: synced with Jim and we'll keep this condition as it gives a better experience/error message than not having it.

PR #1197, opened 2019-03-27
Update lightning LSPs to 2.0.9

nrkruk:
- Bug fixes for windows
- Add preview preferences for lightning explorer


packages/salesforcedx-vscode-lightning/src/commands/quickpick/quickpick.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { paramCase } from 'change-case';
-import opn = require('opn');
+import opn = require('open');
```
lcampos: Can this be changed to `import ... from ...` instead of require ?

PR #1223, opened 2019-04-02
Retrieve default org's metadata types

AnanyaJha:
I want to get feedback on my changes so far before I add tests. 

### What does this PR do?
This PR adds the functionality to get the metadataTypes for the default org and save the results in the `.sfdx/orgs` folder. It does not include the activation of this process so this code does not get triggered currently. 

@W-5980289@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/index.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/index.ts
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export {
+  forceDescribeMetadata
+} from './orgMetadata';
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/index.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/index.ts
@@ -5,5 +5,5 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 export {
-  forceDescribeMetadata
+  onUsernameChange
 } from './orgMetadata';
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
@@ -9,7 +9,9 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as fs from 'fs';
 import * as path from 'path';
+import { isNullOrUndefined } from 'util';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
@@ -17,7 +19,7 @@ import {
   SfdxWorkspaceChecker
 } from '../commands';
 import { nls } from '../messages';
-import { getRootWorkspacePath, hasRootWorkspace } from '../util';
+import { getRootWorkspacePath } from '../util';
 
 export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<string> {
   public build(data: string): Command {
@@ -45,4 +47,28 @@ export async function forceDescribeMetadata() {
     describeExecutor
   );
   await commandlet.run();
+  buildTypeList();
+}
+
+export type MetadataObject = {
+  directoryName: string;
+  inFolder: boolean;
+  metaFile: boolean;
+  suffix: string;
+  xmlName: string;
+};
+
+export function buildTypeList() {
+  const filePath = path.join(getRootWorkspacePath(), '.sfdx', 'tools', 'metadata', 'metadataTypes.json');
+  if (fs.existsSync(filePath)) {
+    const fileData = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+    const metadataObjects = fileData.metadataObjects;
+    const metadataTypes = [];
+    for (const index in metadataObjects) {
+      if (!isNullOrUndefined(metadataObjects[index].xmlName)) {
+        metadataTypes.push(metadataObjects[index].xmlName);
+      }
+    }
+    console.log(metadataTypes);
+  }
 }
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import {
+  Command,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as path from 'path';
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from '../commands';
+import { nls } from '../messages';
+import { getRootWorkspacePath, hasRootWorkspace } from '../util';
+
+export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<string> {
+  public build(data: string): Command {
+    const filePath = path.join(getRootWorkspacePath(), '.sfdx', 'tools', 'metadata', 'metadataTypes.json');
+    return new SfdxCommandBuilder()
+      .withDescription(
+        nls.localize('')
+      )
+      .withArg('force:mdapi:describemetadata')
+      .withJson()
+      .withFlag('-f', filePath)
+      .withLogName('force_describe_metadata')
+      .build();
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+const describeExecutor = new ForceDescribeMetadataExecutor();
+
+export async function forceDescribeMetadata() {
+  const commandlet = new SfdxCommandlet(
+    workspaceChecker,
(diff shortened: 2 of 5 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,139 @@
...
+  private outputPath: string;
+
+  public constructor(outputPath: string) {
+    super();
+    this.outputPath = outputPath;
+  }
+
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(
```
AnanyaJha: @ruthemmanuelle Just wanted to check if this description was alright.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,139 @@
...
+    const fileData = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+    const metadataObjects = fileData.metadataObjects;
+    const metadataTypes = [];
+    for (const index in metadataObjects) {
+      if (!isNullOrUndefined(metadataObjects[index].xmlName)) {
+        metadataTypes.push(metadataObjects[index].xmlName);
+      }
+    }
+  } else {
+    throw new Error('There was an error retrieving metadata type information.');
```
AnanyaJha: @ruthemmanuelle This error would get thrown if there is an issue finding the file with the metadata type information, do you think this error message should be more explicit?

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
ruthemmanuelle: Yeah, I think it's good. I like that it mirrors the CLI command's name (and the API call's). Shouldn't this be in a messages file, though?

Can you also update the list in this doc file? The commands are in alphabetical order by CLI command. https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/docs/articles/user-guide/cli-commands.md

AnanyaJha:
Assuming that retrieving the metadata types runs in the background whenever the default org is switched - there are a number of places where errors can be thrown ie. no workspace, an issue with finding the default username, error reading the file. Do we want to actively throw those errors or just log them with a `console.error(**)`?

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
ruthemmanuelle: Whenever possible, we try to make error messages include suggested next steps. What should the user do to recover?

ruthemmanuelle: Also, like the other message, I believe this should be in a messages file.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,139 @@
...
+      execution.command.toString(),
+      (execution.stderrSubject as any) as Observable<Error | undefined>
+    );
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+}
+
+async function getDirectory(): Promise<string | undefined> {
```
allileong: I find the name `getDirectory` to be a bit confusing since it's returning the path to a file. Maybe it could be renamed to something like `getMetadataTypesPath`?

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+      workspaceChecker,
+      parameterGatherer,
+      describeExecutor
+    );
+    await commandlet.run();
+  }
+}
+
+async function getMetadataTypesPath(): Promise<string | undefined> {
+  if (hasRootWorkspace()) {
```
allileong: Since you are already checking for `hasRootWorkspace()` here, could you call `OrgAuthInfo.getDefaultUsernameOrAlias()` directly fro line 94 and get rid of the helper function in line 109?

AnanyaJha: Yeah, I think that would definitely make sense!

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
AnanyaJha: Okay sounds good! I just kept the messages in this file until we confirmed them, but I'll move them over to the messages file. This command actually won't be surfaced through the command palette, so would we still want to display it in that doc?

lcampos: This shouldn't be part of that list since the user will never use this through the command palette or any of the other ways we expose cli commands.

packages/salesforcedx-vscode-core/src/orgBrowser/index.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export {
```
lcampos: I think your local setup is not picking up the repo's prettier config.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+
+export async function forceDescribeMetadata(outputPath?: string) {
```
lcampos: Will calling `forceDescribeMetadata` directly be the way in which we'll refresh existing data ?

AnanyaJha: Yes, I think that should be the way we'll refresh existing data but I need to make some changes to the current method to allow for that

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  } else {
+    throw new Error('There was an error retrieving metadata type information.');
+  }
+}
+
+export async function onUsernameChange() {
+  const metadataTypesPath = await getMetadataTypesPath();
+  if (!isNullOrUndefined(metadataTypesPath) && fs.existsSync(metadataTypesPath)) {
+    buildTypesList(metadataTypesPath);
```
lcampos: I don't think we need to load the data (.json file created by the cli command above) in memory every time the org credentials are changed/added to the `.sfdx-config.json`. I'd imagine this would only be the entry point to getting data from the org if it's not available locally.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+export type MetadataObject = {
+  directoryName: string;
+  inFolder: boolean;
+  metaFile: boolean;
+  suffix: string;
+  xmlName: string;
+};
+
+export function buildTypesList(metadataTypesPath: string) {
+  if (!isNullOrUndefined(metadataTypesPath)) {
```
lcampos: It would be good to log the number of metadata types the user is querying for.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+export async function getDefaultUsernameOrAlias(): Promise<string | undefined> {
+  if (hasRootWorkspace()) {
+    return await OrgAuthInfo.getDefaultUsernameOrAlias();
+  }
+}
+
+export type MetadataObject = {
```
lcampos: This type is not being used.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+    const workspaceRootPath = getRootWorkspacePath();
+    const defaultUsernameOrAlias = await getDefaultUsernameOrAlias();
+    const defaultUsernameIsSet = typeof defaultUsernameOrAlias !== 'undefined';
+
+    if (defaultUsernameIsSet) {
+      const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias!);
+      const metadataTypesPath = path.join(workspaceRootPath, '.sfdx', 'orgs', username, 'metadata', 'metadataTypes.json');
+      return metadataTypesPath;
+    } else {
+      throw new Error(nls.localize('error_no_default_username'));
```
lcampos: Log these errors so we can have visibility on what users might run into.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
ruthemmanuelle: Ah, ok. I didn't understand that. No need to add it to the doc if it won't be in the command palette. I think that adding it would confuse people.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
ruthemmanuelle: @AnanyaJha - I saw your question on this PR, "Assuming that retrieving the metadata types runs in the background whenever the default org is switched - there are a number of places where errors can be thrown ie. no workspace, an issue with finding the default username, error reading the file. Do we want to actively throw those errors or just log them with a console.error(**)?" When these errors occur, is there anything the user should do to avoid problems? Or will we have another opportunity to communicate with the user before they actually try to work with the retrieved types about whatever issues they need to address?

Issue #1090, opened 2019-02-21
Move Configs to Project Templates


ntotten:
Today, several of the extensions for VS Code write or update various configuration files from the extension. Examples are the jsconfig.json, .eslintrc from LWC, changing values in .vscode/settings.json from the core extension and LWC.

I find this pattern of having the extensions write config files to be jarring and unfamiliar. I can't think of any other tools that do this. We have also seen problems (https://github.com/forcedotcom/salesforcedx-vscode/issues/1049) from customers when we update these files - specifically for local settings.

Instead of writing configuration files and settings from extensions we should include those files and settings in the project or other templates and the extensions should check for their presence and provide proper warnings when they are not included.

We already have done most of this work with the updated project create template. It now will include the following out of the box:

```
├── README.md
├── .gitignore
├── .forceignore
├── .prettierrc
├── .prettierignore
├── .vscode
│   ├── extensions.json
│   ├── settings.json
│   └── launch.json
├── config
│   └── project-scratch-def.json
├── force-app
│   └── main
│       └── default
│           ├── applications
│           ├── aura
│           ├── classes
│           ├── contentassets
│           ├── flexipages
│           ├── layouts
│           ├── lwc
│           │  └── .eslintrc.json
│           ├── objects
│           ├── permissionsets
│           ├── staticresources
│           ├── tabs
│           └── triggers
└── sfdx-project.json
```

So you will already get most of what you need to have configured in an extension in the default template. If we added a few things such as the `jsconfig.json` file then we could eliminate the need for extensions to write files/settings individually.

This does bring up the problem of what happens if the settings/config files change. We have in the past discussed the idea of a CLI command called `force:project:update` that would update your project files to the latest version. This is useful for changes to the schema of the `sfdx-project.json` or updating/adding config files with new settings.

Additionally, there could be functionality in the command `fource:project:update --check` that does a check of your projects version settings. With this command, we could do a single check that your project was in the right state from the core VS Code extension - one place to manage all the configs and settings. We would then prompt people to upgrade their project if needed. This would remove the need for every extension to implement this themselves.

Additionally, we need to put local user configurations in a new place outside of `.vscode/settings.json`. That file is meant to be used for project settings, not user settings. If we do have local, extension specific settings we should adopt something like `.sfdx/settings.json` that is not checked into source control.

dehru:
I'm just commenting on the .vscode/settings.json file, and my recent experience working within this repo.

[...]

PR #1234, opened 2019-04-05
Set default project create template to standard

ntotten:
### What does this PR do?
Updates the project create to use the standard template

#1090

### DIFF
--- a/docs/articles/getting-started/migrate-from-forcecom-ide.md
+++ b/docs/articles/getting-started/migrate-from-forcecom-ide.md
@@ -79,7 +79,7 @@ The second option for migrating your project is an in-place conversion. This opt
 1. As a first step, you need to add several files (some required, some optional) to support the new project format. The most important is the `sfdx-project.json`. The easiest way to create these files is to create a temporary project to copy them from. Run the following command to do so.
 
    ```bash
-   $ sfdx force:project:create -n ../tempproj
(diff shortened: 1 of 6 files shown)


ruthemmanuelle:
Thanks for updating this, @ntotten . The Salesforce release notes will say that the `standard` template becomes the default (in Salesforce CLI and Salesforce Extensions for VS Code) at R2. I think we should wait until then to merge this PR. We have docs and Trailhead content that refer to the old project structure, and they'll be wrong if we merge this now. Cc: @lcampos

ntotten:
I don't want to wait in VS Code. This isn't going to break documentation or tutorials because the commands don't change, the only difference is that the template will have more stuff in it. The reason we are waiting in CLI is that we have a policy of introducing breaking changes only at R2, we don't have such a policy for VS Code - nor should we.

Let's update the release notes to reflect that the change is only the default in CLI at R2.

ruthemmanuelle:
Here are a couple of Trailhead units that will be wrong if we merge this:
- https://trailhead.salesforce.com/content/learn/projects/find-and-fix-bugs-with-apex-replay-debugger/apex-replay-debugger-set-up-debugger
- https://trailhead.salesforce.com/content/learn/projects/quickstart-vscode-salesforce/use-vscode-for-salesforce

There might be others. Those are just the ones I know about off the top of my head.

lcampos:
Changes merged as PR #1283

PR #1235, opened 2019-04-05
Expose create commands only on default folders

### LINKED ISSUE #852: Template Commands should be surfaced on only default folders.

ntotten:
### What does this PR do?
Limits create commands to only be exposed on default folders

Closes #852, @W-5746531@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -116,15 +116,15 @@
       "explorer/context": [
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename == classes && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.visualforce.component.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != components && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.visualforce.page.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != pages && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.lightning.app.create",
@@ -144,7 +144,7 @@
         },
         {
           "command": "sfdx.force.apex.trigger.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != triggers && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.source.retrieve.source.path",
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -116,15 +116,15 @@
       "explorer/context": [
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename == classes && sfdx:project_opened"
```
lcampos: This is good but looks like it breaks a test so need to get that updated as well.

brpowell: Looks good to me too. The test failure doesn't look related to these changes, I was seeing that fail locally for me on another branch as well...

packages/salesforcedx-vscode-core/package.json
```diff
@@ -116,15 +116,15 @@
...
         },
         {
           "command": "sfdx.force.visualforce.component.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != components && sfdx:project_opened"
```
praksb: shouldn't this be resourceFilename == components

packages/salesforcedx-vscode-core/package.json
```diff
@@ -116,15 +116,15 @@
...
         },
         {
           "command": "sfdx.force.visualforce.page.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != pages && sfdx:project_opened"
```
praksb: resourceFilename == pages

packages/salesforcedx-vscode-core/package.json
```diff
@@ -144,7 +144,7 @@
         },
         {
           "command": "sfdx.force.apex.trigger.create",
-          "when": "explorerResourceIsFolder && resourceFilename != aura && resourceFilename != lwc && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourceFilename != triggers && sfdx:project_opened"
```
praksb: resourceFilename == triggers

packages/salesforcedx-vscode-core/package.json
ntotten: @brpowell Ah that, I was stumped on that one. I couldn't figure out why this change broke some random localization test.

packages/salesforcedx-vscode-core/package.json
ntotten: Good catch. Fixed.

packages/salesforcedx-vscode-core/package.json
ntotten: Good catch. Fixed

PR #1236, opened 2019-04-05
Move sObject Refresh command to apex module and improve telemetry

brpowell:
### What does this PR do?

The command code has been moved to the apex module, and it now kicks off the process once the language server has finished indexing the classes. This is to minimize the language server initialization clashing with pulling down all of the classes. The setting to enable or disable refresh on start is still present, but it is now under the apex configuration. Send additional telemetry data about the refresh process.

W-6017035

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -24,6 +24,21 @@ export interface CancellationToken {
   isCancellationRequested: boolean;
 }
 
+export enum SObjectRefreshSource {
+  Manual = 'manual',
+  Startup = 'startup'
+}
+
+export interface SObjectRefreshResult {
+  data: {
+    source?: SObjectRefreshSource;
+    cancelled: boolean;
+    standardObjects?: number;
+    customObjects?: number;
+  };
+  error?: { message: string; stack?: string };
+}
+
 const SFDX_DIR = '.sfdx';
 const TOOLS_DIR = 'tools';
 const SOBJECTS_DIR = 'sobjects';
@@ -70,16 +85,20 @@ export class FauxClassGenerator {
 
   private emitter: EventEmitter;
   private cancellationToken: CancellationToken | undefined;
+  private result: SObjectRefreshResult;
 
   constructor(emitter: EventEmitter, cancellationToken?: CancellationToken) {
     this.emitter = emitter;
     this.cancellationToken = cancellationToken;
+    this.result = { data: { cancelled: false } };
   }
 
   public async generate(
     projectPath: string,
-    type: SObjectCategory
-  ): Promise<string> {
+    type: SObjectCategory,
+    source: SObjectRefreshSource
+  ): Promise<SObjectRefreshResult> {
+    this.result = { data: { source, cancelled: false } };
     const sobjectsFolderPath = path.join(
       projectPath,
       SFDX_DIR,
@@ -113,8 +132,10 @@ export class FauxClassGenerator {
     try {
       sobjects = await describe.describeGlobal(projectPath, type);
     } catch (e) {
+      const err = JSON.parse(e);
       return this.errorExit(
-        nls.localize('failure_fetching_sobjects_list_text', e)
+        nls.localize('failure_fetching_sobjects_list_text', err.message),
+        err.stack
       );
     }
     let j = 0;
@@ -130,9 +151,9 @@ export class FauxClassGenerator {
           await describe.describeSObjectBatch(projectPath, sobjects, j)
         );
         j = fetchedSObjects.length;
-      } catch (e) {
+      } catch (errorMessage) {
         return this.errorExit(
-          nls.localize('failure_in_sobject_describe_text', e)
+          nls.localize('failure_in_sobject_describe_text', errorMessage)
         );
       }
     }
@@ -146,18 +167,21 @@ export class FauxClassGenerator {
       }
     }
 
+    this.result.data.standardObjects = standardSObjects.length;
+    this.result.data.customObjects = customSObjects.length;
+
     this.logFetchedObjects(standardSObjects, customSObjects);
 
     try {
       this.generateFauxClasses(standardSObjects, standardSObjectsFolderPath);
-    } catch (e) {
-      return this.errorExit(e);
+    } catch (errorMessage) {
+      return this.errorExit(errorMessage);
     }
 
     try {
       this.generateFauxClasses(customSObjects, customSObjectsFolderPath);
-    } catch (e) {
-      return this.errorExit(e);
+    } catch (errorMessage) {
+      return this.errorExit(errorMessage);
     }
 
     return this.successExit();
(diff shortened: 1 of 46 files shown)


packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -49,5 +50,7 @@ export const messages = {
     'It looks like this file has been updated. To update your code coverage numbers, run the tests in this file.',
   colorizer_no_code_coverage_current_file:
     'No code coverage information was found for this file. Set "salesforcedx-vscode-core.retrieve-test-code-coverage": true in your user or workspace settings. Then, run Apex tests that include methods in this file. You can run tests from the Apex Tests sidebar or using the Run Tests or Run All Tests code lens within the file.',
-  colorizer_statusbar_hover_text: 'Highlight Apex Code Coverage'
+  colorizer_statusbar_hover_text: 'Highlight Apex Code Coverage',
+  force_sobjects_no_refresh_if_already_active_error_text:
+    'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.'
```
ruthemmanuelle: 👍

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -425,29 +418,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
-    );
-  }
-
   const api: any = {
-    ProgressNotification,
+    channelService,
     CompositeParametersGatherer,
     EmptyParametersGatherer,
+    getDefaultUsernameOrAlias,
+    getRootWorkspacePath,
```
lcampos: I don't think we should expose `getRootWorkspacePath` to other extensions. It is a nice wrapper to the vscode workspace api but it does not add anything to justify other extensions using it.

praksb:
@brpowell Is it possible to make the change to fetch the files before we start indexing in LSP? Otherwise this change would be a potential source of performance issue since I think an sobject fetch would cause us to reinitialize again. So we would have to run index once at startup and again on an sobject fetch.

brpowell:
> @brpowell Is it possible to make the change to fetch the files before we start indexing in LSP? Otherwise this change would be a potential source of performance issue since I think an sobject fetch would cause us to reinitialize again. So we would have to run index once at startup and again on an sobject fetch.

You mean wait for sObject refresh to be done and then initialize the LSP? Or kick it off in parallel? If it's the former I'd say that might be a burden for the user to have to wait for refresh to finish to start using lsp features.

lcampos:
@praksb @brpowell indexing and recompilations are good things to look out for with these changes. Keep in mind that we don't know how long the fetching of sobjects will take since that'll depend on the org's shape. You can look at existing telemetry to get an idea, but it could be that indexing local code finishes before the sobject fetch call.

praksb:
@brpowell I was thinking of the former (wait for sobject refresh to be done not run in parallel). But, I do see your point that the refresh might take a while. We can check this in as is and wait for customer feedback. The possible perf impact should also be minimal since we do a refresh only once now and that too only when the folder didn't exist.

> > @brpowell Is it possible to make the change to fetch the files before we start indexing in LSP? Otherwise this change would be a potential source of performance issue since I think an sobject fetch would cause us to reinitialize again. So we would have to run index once at startup and again on an sobject fetch.
> 
> You mean wait for sObject refresh to be done and then initialize the LSP? Or kick it off in parallel? If it's the former I'd say that might be a burden for the user to have to wait for refresh to finish to start using lsp features.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -225,6 +226,13 @@
           ],
           "default": false,
           "description": "%apex_semantic_errors_description%"
+        },
+        "salesforcedx-vscode-apex.enable-sobject-refresh-on-startup": {
+          "type": [
+            "boolean"
+          ],
+          "default": true,
+          "description": "%enable_sobject_refresh_on_startup_description%"
```
lcampos: The `sfdx.force.internal.refreshsobjects` commands from the core extension's `package.json` need to be moved to this `package.json`

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -24,6 +24,21 @@ export interface CancellationToken {
   isCancellationRequested: boolean;
```
brpowell: Changes in this file are for improving telemetry data around this command

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -225,6 +230,13 @@
           ],
           "default": false,
           "description": "%apex_semantic_errors_description%"
+        },
+        "salesforcedx-vscode-apex.enable-sobject-refresh-on-startup": {
+          "type": [
+            "boolean"
+          ],
+          "default": false,
```
brpowell: As we discussed this morning, the default behavior will now be off. The PR is mostly focused on moving the code to the apex extension and improving telemetry data now.

packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -112,18 +111,30 @@ export class ForceGenerateFauxClassesExecutor extends SfdxCommandletExecutor<{}>
...
-      console.log('Generate error ' + e);
+      const result = await gen.generate(
+        vscode.workspace.workspaceFolders![0].uri.fsPath,
+        SObjectCategory.ALL,
+        refreshSource
+      );
+      console.log('Generate success ' + result.data);
+      this.logMetric(commandName, startTime, result.data);
+    } catch (result) {
+      console.log('Generate error ' + result.error);
```
brpowell: Now log errors and successes with the number of objects fetched - if the command got far enough to collect that information.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -59,14 +67,10 @@ export async function activate(context: vscode.ExtensionContext) {
   }
 
   // Telemetry
-  if (sfdxCoreExtension && sfdxCoreExtension.exports) {
-    sfdxCoreExtension.exports.telemetryService.showTelemetryMessage();
```
brpowell: @lcampos removed the telemetry message based on our discussion

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -205,6 +206,10 @@
       {
         "command": "sfdx.force.apex.test.last.method.run",
         "title": "%force_apex_test_last_method_run_text%"
+      },
+      {
+        "command": "sfdx.force.internal.refreshsobjects",
```
lcampos: You are missing porting the command under `commandPalette`.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -80,6 +84,17 @@ export async function activate(context: vscode.ExtensionContext) {
       if (languageClient) {
         languageClient.onNotification('indexer/done', async () => {
           LanguageClientUtils.indexing = false;
+
+          // Refresh SObject definitions if there aren't any faux classes
+          const sobjectRefreshStartup: boolean = vscode.workspace
+            .getConfiguration(SFDX_APEX_CONFIGURATION_NAME)
+            .get<boolean>(ENABLE_SOBJECT_REFRESH_ON_STARTUP, true);
```
lcampos: The default should be `false`

brpowell: I'll change that for consistency. FYI I recently discovered this value doesn't actually matter here, the default value is read from the package.json.

packages/salesforcedx-vscode-apex/src/telemetry/telemetry.ts
```diff
@@ -79,6 +87,29 @@ export class TelemetryService {
     }
   }
 
+  public sendErrorEvent(
+    error: { message: string; stack?: string },
+    additionalData?: any,
```
lcampos: I think commandName can be sent as part of additionalData.

Also, is it work deleting the `sendError` error method and update the code to use the new `sendErrorEvent` ?

brpowell: You mean `sendError` in the core module? It has quite a few references, but not a ridiculous amount. I will create separate work for it. I'll modify additionalData though 👍

PR #1237, opened 2019-04-05
Retrieve default org's metadata types

AnanyaJha:
### What does this PR do?
This PR adds the functionality to get the metadataTypes for the default org and save the results in the .sfdx/orgs folder. It does not include the activation of this process so this code will not get triggered.

@W-5980289@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
@@ -9,7 +9,9 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as fs from 'fs';
 import * as path from 'path';
+import { isNullOrUndefined } from 'util';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
@@ -17,7 +19,7 @@ import {
   SfdxWorkspaceChecker
 } from '../commands';
 import { nls } from '../messages';
-import { getRootWorkspacePath, hasRootWorkspace } from '../util';
+import { getRootWorkspacePath } from '../util';
 
 export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<string> {
   public build(data: string): Command {
@@ -45,4 +47,28 @@ export async function forceDescribeMetadata() {
     describeExecutor
   );
   await commandlet.run();
+  buildTypeList();
+}
+
+export type MetadataObject = {
+  directoryName: string;
+  inFolder: boolean;
+  metaFile: boolean;
+  suffix: string;
+  xmlName: string;
+};
+
+export function buildTypeList() {
+  const filePath = path.join(getRootWorkspacePath(), '.sfdx', 'tools', 'metadata', 'metadataTypes.json');
+  if (fs.existsSync(filePath)) {
+    const fileData = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+    const metadataObjects = fileData.metadataObjects;
+    const metadataTypes = [];
+    for (const index in metadataObjects) {
+      if (!isNullOrUndefined(metadataObjects[index].xmlName)) {
+        metadataTypes.push(metadataObjects[index].xmlName);
+      }
+    }
+    console.log(metadataTypes);
+  }
 }
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import {
+  Command,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import * as path from 'path';
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from '../commands';
+import { nls } from '../messages';
+import { getRootWorkspacePath, hasRootWorkspace } from '../util';
+
+export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<string> {
+  public build(data: string): Command {
+    const filePath = path.join(getRootWorkspacePath(), '.sfdx', 'tools', 'metadata', 'metadataTypes.json');
+    return new SfdxCommandBuilder()
+      .withDescription(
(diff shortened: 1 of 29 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,157 @@
...
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+
+export async function forceDescribeMetadata(outputPath?: string) {
```
AnanyaJha: This function would get called for the initial load of the view and whenever the view is refreshed

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export type MetadataObject = {
+  directoryName: string;
+  inFolder: boolean;
+  metaFile: boolean;
+  suffix: string;
+  xmlName: string;
+};
+
+export function buildTypesList(metadataTypesPath: string): string[] {
```
AnanyaJha: This would be used to populate the top level child nodes for the metadata types

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,157 @@
...
+        metadataTypes.push(metadataObject.xmlName);
+      }
+    }
+    telemetryService.sendMetadataTypes(undefined, {
+      metadataTypes: metadataTypes.length
+    });
+    return metadataTypes;
+  } else {
+    const err =
+      'There was an error retrieving metadata type information. Refresh the view to retry.';
```
AnanyaJha: @ruthemmanuelle If we run into an error when trying to retrieve the metadata types from the generated output file, I think we should throw an error. If the user runs into this scenario, they should be able to refresh the view to generate the output file and load the view again.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -0,0 +1,157 @@
...
+export type MetadataObject = {
+  directoryName: string;
+  inFolder: boolean;
+  metaFile: boolean;
+  suffix: string;
+  xmlName: string;
+};
+
+export function buildTypesList(metadataTypesPath: string): string[] {
+  if (fs.existsSync(metadataTypesPath)) {
```
lcampos: Since we're parsing a json, I think we should wrap the method on a try catch that surfaces that throws the error and have the code calling this handle surfacing the message.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -164,6 +164,19 @@ export class TelemetryService {
     }
   }
 
+  public sendMetadataTypes(
```
lcampos: Should this be a more generic method that we use to log things that are not errors or command executions ? It could be something like `sendEventData` or something like that.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/orgMetadata.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  beforeEach(() => {
+    getDefaultUsernameStub = stub(OrgAuthInfo, 'getDefaultUsernameOrAlias');
+    getUsernameStub = stub(OrgAuthInfo, 'getUsername');
+  });
+  afterEach(() => {
+    getDefaultUsernameStub.restore();
+    getUsernameStub.restore();
+  });
+
+  it('returns the path for a given username', async () => {
```
lcampos: Nitpick, change the title to `Should return ...` to follow the current test conventions.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
AnanyaJha: Yeah, I think that's a good idea!

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
AnanyaJha: Okay, that works

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
ruthemmanuelle: Sorry, I just saw this.  Can we tell them how to refresh the view? Or is it obvious? Not sure what this UI looks like.

AnanyaJha: I've taken out this message with my latest change, but at some point we will have to display a message similar to this. I think the view will include a refresh button in the top right corner, so we would want to point the user to that.

ruthemmanuelle: Ok, it sounds like refreshing will be intuitive, so the message you had (or something similar to it) should be good. Let me know when you're ready to add the new message.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/orgMetadata.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    getDefaultUsernameStub = stub(OrgAuthInfo, 'getDefaultUsernameOrAlias');
+    getUsernameStub = stub(OrgAuthInfo, 'getUsername');
+  });
+  afterEach(() => {
+    getDefaultUsernameStub.restore();
+    getUsernameStub.restore();
+  });
+
+  it('should return the path for a given username', async () => {
+    getDefaultUsernameStub.returns('defaultUsername');
```
lcampos: This should always return an username in email format e.g. `test-audskbqx3kttjb@example.com`

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -164,6 +164,16 @@ export class TelemetryService {
     }
   }
 
+  public sendEventData(
```
lcampos: Need to add a test for this in `test/vscode-integration/telemetry/index.test.ts` since no other method is testing sending measures.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/orgMetadata.test.ts
AnanyaJha: I kept the `getDefaultUsernameStub` as `defaultAlias` because it can return an alias or a username, but I updated the `getUsername` stub to be in an email format, since that ensures we only have the username.

PR #1245, opened 2019-04-08
Update LSP jar for NPE that happens during computing references

praksb:
### What does this PR do?
Updates LSP jar with the latest LSP changes.

Fixes an NPE that was being caused by adding references to type/member which are null during computing of references.
@W-6022740@


praksb:
@lcampos Can you please help merge? The test failure is not letting me to merge.

Issue #1210, opened 2019-03-29
Aura Components extension is overwriting .vscode/settings.json


aheber:
### Summary

It seems that since yesterday's update the Aura Components plugin is overwriting the [[projectroot]]/.vscode/settings.json file with the following:
```
{
    "html.suggest.angular1": false,
    "html.suggest.ionic": false,
}
```
Maybe limited to my project only, I don't know.

### Steps To Reproduce:

1. Have a basic settings.json file already configured:
```
{
    "npm.enableScriptExplorer": true
}
```
2. With the Aura Components plugin installed, reload the vscode window
settings.json is rewritten to:
```
{
    "html.suggest.angular1": false,
    "html.suggest.ionic": false
}
```

### Expected result

It is OK to add things to the settings.json, I should not however lose existing setting values.

### Actual result

All of my existing project settings are lost and replaced. This file is tracked in source control so I was able to recover it but it is still being overwritten every time I restart.

### Additional information

_Feel free to attach a screenshot_.

**VS Code Version**:
1.32.3 (user setup)
**SFDX CLI Version**:
@oclif/plugin-commands 1.2.2 (core)
@oclif/plugin-help 2.1.6 (core)
@oclif/plugin-not-found 1.2.2 (core)
@oclif/plugin-plugins 1.7.8 (core)
@oclif/plugin-update 1.3.9 (core)
@oclif/plugin-warn-if-update-available 1.7.0 (core)
@oclif/plugin-which 1.0.3 (core)
@salesforce/sfdx-trust 3.0.2 (core)
analytics 1.1.2 (core)
generator 1.1.0 (core)
salesforcedx 45.8.1 (core)
├─ force-language-services 45.9.0 (core)
└─ salesforce-alm 45.10.0 (core)

sfdx-cli 7.1.3 (core)

**OS and version**:
Windows 10

ntotten:
Fix has been checked in and will go out next week.

AndrewRayCode:
It looks like you've overwritten my local vscode settings in every single project I've worked on. These files aren't version controlled obviously because they're local to me and the project and shouldn't be shared.

How do I get these files back? I had a lot of per-project settings that are now lost.

I've reported this extension so hopefully it gets taken down ASAP

ntotten:
@AndrewRayCode I apologize for the problem this has caused. This was a bug and a mistake on our part. We have updated the extension and the fix will go out in a few days.

[...]

PR #1231, opened 2019-04-04
Update aura and lwc lsps to 2.1.0

nrkruk:
### What does this PR do?
- Updates Aura and LWC Language Servers to 2.0.13
- Updates plugin activation to be more consistent for non-standard workspaces
- Remove core dependency from sfdx-vscode-lightning plugin to allow async startup
- Adds configuration settings for activation

### What issues does this PR fix or reference?
- fixes #1210

PR #1253, opened 2019-04-10
update Lightning LSPs to 2.1.0 / add new activation logic

nrkruk:
### What does this PR do?
- Update the lightning lsps to 2.1.0
- Fix settings.json file being overwritten
- Async startup for aura language server
- Add vscode settings for configuring LSP activation
- Cleanup activation logic for internal and external developers
#1210

### DIFF
--- a/packages/salesforcedx-vscode-lightning/package.json
+++ b/packages/salesforcedx-vscode-lightning/package.json
@@ -142,7 +142,7 @@
             "autodetect",
             "off"
           ],
-          "enumDescription": [
+          "enumDescriptions": [
             "%activation_mode_always_on%",
             "%activation_mode_autodetect%",
             "%activation_mode_off%"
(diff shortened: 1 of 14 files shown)


nrkruk:
This is the same PR as (https://github.com/forcedotcom/salesforcedx-vscode/pull/1231), but with cleanup / comments addressed

packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -28,7 +28,7 @@
     "aura-language-server": "2.1.0",
     "change-case": "^3.1.0",
     "lightning-lsp-common": "2.1.0",
-    "open": "^6.0.0",
+    "open": "6.0.0",
```
lcampos: @AnanyaJha this fixes the compile issue we're seeing this morning with using version 6.2.0.

PR #1255, opened 2019-04-10
Update changelog release 45.10.0

lcampos:
### What does this PR do?
Updates changelog for release v45.10.0

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### salesforcedx-vscode-apex
 
-- Fix NullPointerException when Apex language server was processing references ([PR #1245](https://github.com/forcedotcom/salesforcedx-vscode/pull/1245))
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
+## 45.10.0 - April 10, 2019
+
+### Fixed
+
+#### salesforcedx-vscode-apex
+
+- Fix NullPointerException when Apex language server was processing references ([PR #1245](https://github.com/forcedotcom/salesforcedx-vscode/pull/1245))
+
+#### salesforcedx-vscode-apex-debugger, salesforcedx-vscode-apex-replay-debugger
```
lcampos: The fix applies to both extensions but let me know if we should change the format

ruthemmanuelle: I think it's fine. If we also had items that affected one but not the other I'd want to have separate sections, but it looks like we don't.

PR #1259, opened 2019-04-12
display message based upon defaultname and where it was found or not

JimSuplizio:
### What does this PR do?
Display a message when the user does not have a defaultusername set for the current project OR if the defaultusername is being pulled from the global config instead of the local project config.

If there is no defaultusername in the local or global configs we were already displaying a message to the debug console I just added that message as an informational popup. That message will be displayed if the user create a new project or loads a project that has not authenticated yet.

If the defaultusername is pulled from the global config we will now display a warning message (console and popup) with instructions to authorize an org.

As part of this I refactored the config code out of the OrgAuthInfo class into its own class, ConfigUtil. The fact is, the CLI's ConfigAggregator class should do exactly what ConfigUtil does but it's currently broken and doesn't fetch anything other than the global config.

The informational message
InformationNoDefaultOrg

The warning message
GlobalConfigWarning

@W-5856821@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/messages/i18n.ts
+++ b/packages/salesforcedx-vscode-core/src/messages/i18n.ts
@@ -242,6 +242,8 @@ export const messages = {
   error_no_default_devhubusername:
     'No default Dev Hub is set. Run "SFDX: Authorize a Dev Hub" to set one.',
   custom_output_directory: 'Choose a Custom Directory',
+  warning_using_global_username:
+    'No default username found in the local project config; using the global default username. Run "SFDX: Authorize an Org" to set the username for the local project config.',
   apex_class_message_name: 'Apex Class',
   apex_trigger_message_name: 'Apex Trigger',
   visualforce_component_message_name: 'Visualforce Component',
--- a/packages/salesforcedx-vscode-core/src/util/authInfo.ts
+++ b/packages/salesforcedx-vscode-core/src/util/authInfo.ts
@@ -4,60 +4,53 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import {
-  Aliases,
-  AuthInfo,
-  ConfigAggregator,
-  ConfigFile
-} from '@salesforce/core';
-import * as path from 'path';
+import { Aliases, AuthInfo } from '@salesforce/core';
+import { isUndefined } from 'util';
+import { channelService } from '../channels';
 import { nls } from '../messages';
-import { getRootWorkspacePath } from './index';
+import { notificationService } from '../notifications';
+import { ConfigSource, ConfigUtil } from './index';
+
+const defaultUserNameKey = 'defaultusername';
+const defaultDevHubUserNameKey = 'defaultdevhubusername';
 export class OrgAuthInfo {
   public static async getDefaultUsernameOrAlias(): Promise<string | undefined> {
-    try {
-      const rootPath = getRootWorkspacePath();
-      const myLocalConfig = await ConfigFile.create({
-        isGlobal: false,
-        rootFolder: path.join(rootPath, '.sfdx'),
-        filename: 'sfdx-config.json'
-      });
-      const localDefault = myLocalConfig.get('defaultusername');
-      return JSON.stringify(localDefault).replace(/\"/g, '');
-    } catch {
-      await this.getGlobalDefaults('defaultusername');
-    }
-  }
-
-  private static async getGlobalDefaults(usernameType: string) {
-    try {
-      const aggregator = await ConfigAggregator.create();
-      const globalDefault = aggregator.getPropertyValue(usernameType);
-      return JSON.stringify(globalDefault).replace(/\"/g, '');
-    } catch {
-      if (usernameType === 'defaultusername') {
-        console.error(nls.localize('error_no_default_username'));
-      } else {
-        console.error(nls.localize('error_ no_default_devhubusername'));
+    const defaultUserName = await ConfigUtil.getConfigValue(defaultUserNameKey);
+    if (isUndefined(defaultUserName)) {
+      displayMessage(
+        nls.localize('error_no_default_username'),
+        true,
+        VSCodeWindowTypeEnum.Informational
+      );
+      return undefined;
+    } else {
+      const configSource = await ConfigUtil.getConfigSource(defaultUserNameKey);
+      if (configSource === ConfigSource.Global) {
+        displayMessage(
+          nls.localize('warning_using_global_username'),
+          true,
+          VSCodeWindowTypeEnum.Warning
+        );
       }
     }
+    return JSON.stringify(defaultUserName).replace(/\"/g, '');
   }
 
   public static async getDefaultDevHubUsernameOrAlias(): Promise<
     string | undefined
   > {
-    try {
-      const rootPath = getRootWorkspacePath();
-      const myLocalConfig = await ConfigFile.create({
-        isGlobal: false,
-        rootFolder: path.join(rootPath, '.sfdx'),
-        filename: 'sfdx-config.json'
-      });
-      const localDefault = myLocalConfig.get('defaultdevhubusername');
-      return JSON.stringify(localDefault).replace(/\"/g, '');
-    } catch {
-      await this.getGlobalDefaults('defaultdevhubusername');
+    const defaultDevHubUserName = await ConfigUtil.getConfigValue(
+      defaultDevHubUserNameKey
+    );
+    if (isUndefined(defaultDevHubUserName)) {
+      displayMessage(
+        nls.localize('error_no_default_devhubusername'),
+        true,
+        VSCodeWindowTypeEnum.Error
+      );
+      return undefined;
     }
+    return JSON.stringify(defaultDevHubUserName).replace(/\"/g, '');
   }
 
   public static async getUsername(usernameOrAlias: string): Promise<string> {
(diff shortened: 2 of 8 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -243,5 +243,7 @@ export const messages = {
     'No default org is set. Run "SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org" to set one.',
   error_no_default_devhubusername:
     'No default Dev Hub is set. Run "SFDX: Authorize a Dev Hub" to set one.',
-  custom_output_directory: 'Choose a Custom Directory'
+  custom_output_directory: 'Choose a Custom Directory',
+  warning_using_global_username:
```
JimSuplizio: @ruthemmanuelle this was the warning message you'd looked at this morning when I was writing the tests.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,60 +4,53 @@
...
+    const defaultUserName = await ConfigUtil.getConfigValue(defaultUserNameKey);
+    if (isUndefined(defaultUserName)) {
+      displayMessage(
+        nls.localize('error_no_default_username'),
+        true,
+        VSCodeWindowTypeEnum.Informational
+      );
+      return undefined;
+    } else {
+      const configSource = await ConfigUtil.getConfigSource(defaultUserNameKey);
```
JimSuplizio: This is reason we needed to know where the config item was coming from. If we're pulling the defaultusername from the global config then we need to show the user a warning which instructs them to authorize an org.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -80,3 +73,33 @@ export class OrgAuthInfo {
     }
   }
 }
+
+enum VSCodeWindowTypeEnum {
+  Error = 1,
+  Informational = 2,
+  Warning = 3
+}
+function displayMessage(
```
JimSuplizio: Most of the functions duplicated throughout the core extension for displaying messages end up displaying only a single type of message (warning/error/informational). Since OrgAuthInfo needs to display each type of message, I'd opted for a single function that writes to the output window and optionally displays a messages. This was preemptively done like this because getDefaultUsernameOrAlias ends up getting called twice each time a project is loaded, once by  the OrgPicker and once by the WorkspaceOrgType. Some refactoring needs to get done to reconcile those but there probably needs to be a boolean plumbed through the getDefaultUsernameOrAlias call so the message only gets displayed once. Right now, it's odd though, only one notification gets popped up in spite of there being two calls and two messages in the output window.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!isNullOrUndefined(value)) {
+      return ConfigSource.Global;
+    }
+    return ConfigSource.None;
+  }
+
+  // Given a configuration key check the local then the global configs for the value returning
+  // the value if it exists or undefined otherwise. If the optional source argument if set to Local
+  // or Global then it will only check the appropriate config for the value.
+  public static async getConfigValue(
```
JimSuplizio: This method doesn't have any specific tests. This code is a consolidation of the code in OrgAuthInfo and the methods in there were stubbed for testing instead of trying to mess with creating local/global configs.

packages/salesforcedx-vscode-core/test/vscode-integration/util/configUtil.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const configSource = await ConfigUtil.getConfigSource('key');
+    expect(configSource).to.be.eq(ConfigSource.Local);
+  });
+  it('should return ConfigSource.Global if the key/value is in the global config', async () => {
+    getConfigValueStub.onCall(0).returns(undefined);
+    getConfigValueStub.onCall(1).returns('someValue');
+    const configSource = await ConfigUtil.getConfigSource('key');
+    expect(configSource).to.be.eq(ConfigSource.Global);
+  });
+  it('should return ConfigSource.None if the key/value is in the global config', async () => {
```
JimSuplizio: This should be if the key/value is not in any config. I'll submit another PR to correct this.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -4,60 +4,53 @@
...
-        filename: 'sfdx-config.json'
-      });
-      const localDefault = myLocalConfig.get('defaultdevhubusername');
-      return JSON.stringify(localDefault).replace(/\"/g, '');
-    } catch {
-      await this.getGlobalDefaults('defaultdevhubusername');
+    const defaultDevHubUserName = await ConfigUtil.getConfigValue(
+      defaultDevHubUserNameKey
+    );
+    if (isUndefined(defaultDevHubUserName)) {
```
AnanyaJha: I like that we're displaying the message for `defaultUsername` and `defaultDevHubUsername`, but I was wondering why the message is informational for the `username` and an error for the `devhubusername`?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+import { isNullOrUndefined, isUndefined } from 'util';
+import { getRootWorkspacePath } from './index';
+
+export enum ConfigSource {
+  Local,
+  Global,
+  None
+}
+
+// This class should be reworked or removed once the ConfigAggregator correctly checks
```
AnanyaJha: Maybe we can include this info in the chatter for the bug instead?

packages/salesforcedx-vscode-core/src/util/authInfo.ts
JimSuplizio: Prior to me touching anything in this class it was always writing to console.error for both username and devhubusername. For the defaultusername, error wasn't correct because that's something that a user would see when they create a project or load a project that's been created but not authenticated or used to create an org so informational was correct there. For the case where we're pulling defaultusername from the global config, that's definitely a warning. The debhubusername isn't part of the bug or something I vetted, that was always being output as an error so I left that unchanged.

AnanyaJha: Ah okay makes sense, so the missing devhubusername error was originally being reported using console.error as well, so the user never actually saw it as a pop up. Since surfacing this error to the user isn't part of this bug, are we would still like to display this error using the vscode pop up too?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
JimSuplizio: I want to leave the comments in here. The reason being is that we don't control that class and we don't actually know when it'll be fixed. Further, it states intent. I'd had to ask 2 different people (both you and @lcampos) why we weren't using the ConfigAggregator to begin with because there were no comments in the code indicating why processing had to create the local config and then use the ConfigAggregator for the global.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
JimSuplizio: It seems that if there's an actual error that they should have been notifications to begin with instead of just writing to the console since that's the pattern we've adopted. The way I wrote the function, if people flip the other direction and don't want notifications, the notification can be turned off with a boolean

packages/salesforcedx-vscode-core/src/util/authInfo.ts
brpowell: Is this generic enough to live outside of authInfo in another util?

packages/salesforcedx-vscode-core/src/util/authInfo.ts
AnanyaJha: I agree, I think it makes sense to display the notification to the user. But since we are displaying a notification to the user in case of a missing default username and missing default dev hub username, should we try to make them the same type of notification? As in, should both of them be informational? Or both errors? I'm just trying to understand whether we want the same sort of behavior in both scenarios.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+// local as well as global configs. It's also worth noting that ConfigAggregator, according
+// to its docs checks local, global and environment and, for our purposes, environment may
+// not be viable.
+
+// Given a configuration key, check the local then the global configs for the value and return
```
AnanyaJha: I think these methods are clear enough that the commented explanations may not be necessary

packages/salesforcedx-vscode-core/src/util/authInfo.ts
JimSuplizio: Possibly but due to time constraints it's just going to live here for the moment.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
JimSuplizio: Sure, I've removed them with this PR.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
JimSuplizio: The informational message (no default user in local or global) will get shown when the user first creates a project or if the user loads a project that hasn't been authenticated yet. The fact that it was originally an error message was just simply incorrect. The warning message is because we found a default user but it was found in the global which means you could possibly be changing something in an org that you didn't want to change it. This message should have always been a warning and not an error. Finally, the devhub username doesn't actually enter into this equation for my changes. it was previously an error in the code and I didn't change that. I didn't fully vet the scenario in which this message is displayed but kept it as it was. If the message type when the devhub username isn't found is not correct then that is a change separate from this.

PR #1261, opened 2019-04-13
Address apex extension not activating when java is misconfigured

lcampos:
### What does this PR do?
The current change will address `salesforcedx-vscode-apex` extension not activating when Java is missing or is in an unsupported version. As part of this, we'll now be able to keep track of the activation status of the Apex language server and leverage it when we need to query for information from it (apex test sidebar, both apex debuggers).

Issue #809,  @W-5988133@


packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
```diff
@@ -97,9 +97,15 @@ export class DebugConfigurationProvider
     while (
       this.sfdxApex &&
       this.sfdxApex.exports &&
-      !this.sfdxApex.exports.isLanguageClientReady() &&
+      !this.sfdxApex.exports.languageClientStatus.isReady() &&
       !expired
     ) {
+      if (this.sfdxApex.exports.languageClientStatus.failedToInitialize()) {
```
lcampos: This new check allows us to surface a better error message when the user triggers the debugger. We used to display a language server is not ready message. We will now display the correct error that caused the Apex language server to be unavailable.

debugger_error

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -72,56 +72,64 @@ export async function activate(context: vscode.ExtensionContext) {
...
-  LanguageClientUtils.setClientInstance(languageClient);
-  const handle = languageClient.start();
-  context.subscriptions.push(handle);
+  // Initialize Apex language server
+  try {
+    const langClientHRStart = process.hrtime();
+    languageClient = await languageServer.createLanguageServer(context);
+    languageClientUtils.setClientInstance(languageClient);
+    const handle = languageClient.start();
+    languageClientUtils.setStatus(ClientStatus.Indexing, '');
```
lcampos: LanguageClientUtils is now a singleton that has an Unavailable status as its default. Once we set the language client instance and start it we set the status to indexing.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -72,56 +72,64 @@ export async function activate(context: vscode.ExtensionContext) {
...
+            await testOutlineProvider.refresh();
+          });
+        }
+        languageClientUtils.setStatus(ClientStatus.Ready, '');
+        telemetryService.sendApexLSPActivationEvent(langClientHRStart);
+      })
+      .catch(err => {
+        // Handled by clients
+        telemetryService.sendApexLSPError(err);
+        languageClientUtils.setStatus(ClientStatus.Error, err);
```
lcampos: I haven't test this enough but this will only happen when an internal Apex language server happens. Should we add a generic message instead of directly pushing the error ?

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -224,7 +233,7 @@ async function registerTestView(
   // Refresh Test View command
   testViewItems.push(
     vscode.commands.registerCommand('sfdx.force.test.view.refresh', () => {
-      if (!LanguageClientUtils.indexing) {
+      if (!languageClientUtils.getStatus().isIndexing()) {
```
lcampos: @brpowell I was wondering if we should change this from when it's not indexing to when it's ready. What do you think ?

packages/salesforcedx-vscode-apex/src/languageClientUtils/languageClientUtils.ts
```diff
@@ -0,0 +1,115 @@
+/*
```
lcampos: Refactored some of the original code into a singleton and moved it into a new `languageClientUtils` module.

packages/salesforcedx-vscode-apex/src/languageClientUtils/languageClientUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+  DEBUGGER_EXCEPTION_BREAKPOINTS,
+  DEBUGGER_LINE_BREAKPOINTS
+} from '../constants';
+import {
+  ApexLSPConverter,
+  ApexTestMethod,
+  LSPApexTestMethod
+} from '../views/lspConverter';
+
+export class LanguageClientUtils {
```
lcampos: I wanted `LanguageClientUtils` to only be modified by the `salesforcedx-vscode-apex` extension and then have the debugger extensions exclusively consume the status of the Apex language server (`LanguageClientStatus`). This is why I only expose `languageClientUtils.getStatus()` as part of the extensions api.

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -121,6 +125,16 @@ export class ApexTestRunner {
   }
 
   public async runApexTests(tests: string[], testRunType: TestRunType) {
+    const languageClientStatus = languageClientUtils.getStatus() as LanguageClientStatus;
```
lcampos: In the scenario where Java is not present or not the right version, the test sidebar will still allow the user to refresh or click the run test buttons. This will now surface an info message telling them why the tests are not present and how to fix the configuration.
test_sidebar_message

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
```diff
@@ -121,6 +125,16 @@ export class ApexTestRunner {
...
+    if (!languageClientStatus.isReady()) {
+      if (languageClientStatus.failedToInitialize()) {
+        vscode.window.showInformationMessage(
```
lcampos: Not sure if we should change it to an error message. When the extension is activating and the apex language server fails to start we show it as an error message. What do you think ?

packages/salesforcedx-vscode-apex/src/index.ts
sfsholden: What if we set a generic message and followed it up with the error?

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
ruthemmanuelle: I think you can remove "more".

packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
ruthemmanuelle: If this can't be a link, I'd remove the square brackets, the parentheses, and the last period:
`see Set Your Java Version: https://...`

packages/salesforcedx-vscode-apex/src/views/testRunner.ts
praksb: Error message would be good i think to indicate that the language server failed to start.

packages/salesforcedx-vscode-apex/src/index.ts
praksb: We could combine ready and indexing as well. I think if the language server is ready it should mean it started successfully and completed indexing.

packages/salesforcedx-vscode-apex/src/index.ts
praksb: I think we should add a generic message here if it is going to be visible to customer since the actual error might not make sense for the customer. We have telemetry of the actual error that should help us debug the issue.

packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
lcampos: This is the same error shown in the info message. These changes now keep it around in order to show it when a user tries to execute the debuggers or anything else depending on the Apex language server. If I modify it then it means the info messages won't open a browser window when the user clicks on the link.

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -52,5 +52,7 @@ export const messages = {
     'No code coverage information was found for this file. Set "salesforcedx-vscode-core.retrieve-test-code-coverage": true in your user or workspace settings. Then, run Apex tests that include methods in this file. You can run tests from the Apex Tests sidebar or using the Run Tests or Run All Tests code lens within the file.',
   colorizer_statusbar_hover_text: 'Highlight Apex Code Coverage',
   force_sobjects_no_refresh_if_already_active_error_text:
-    'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.'
+    'A refresh of your sObject definitions is already underway. If you need to restart the process, cancel the running task.',
+  apex_language_server_failed_activate:
```
lcampos: @ruthemmanuelle here's a new generic error message that will be displayed to the user when the error is not specific to Java version misconfiguration. Let me know what you think.

packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
ruthemmanuelle: Ok, that makes sense.

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
ruthemmanuelle: Can we give any advice on what to do next?

lcampos: Not really, this message is shown when something goes bad internally in Apex language server but I haven't been able to set a project in a way in which I end up showing this message.

ruthemmanuelle: Do we want to advise them to file an issue on GitHub? Seems like we'd be curious to know how they got in this state. Maybe say something like, `The Apex Language Server failed to activate. Please file an issue on the salesforcedx-vscode GitHub repository.`?

Issue #1289, opened 2019-04-25
SFDX: Update Checkpoints in Org 


alpha-bytes:
### Summary

Today, when running the `SFDX: Update Checkpoints in Org` command from the VS Code palette, I receive the following error in the Output window: 

**sfdxApex.exports.languageClientStatus is not a function**

Opening the dev tools window reveals the following logged error: 

```
Error: sfdxApex.exports.languageClientStatus is not a function
	at Object.<anonymous> (/Users/<myName>/.vs…ut/src/index.js:150)
	at Generator.next (<anonymous>)
	at __awaiter (/Users/<myName>/.vs…out/src/index.js:13)
	at new Promise (<anonymous>)
	at __awaiter (/Users/<myName>/.vs…/out/src/index.js:9)
	at Object.retrieveLineBreakpointInfo (/Users/<myName>/.vs…ut/src/index.js:145)
	at /Users/<myName>/.vs…pointService.js:561
	at Generator.next (<anonymous>)
	at fulfilled (/Users/<myName>/.vs…ckpointService.js:4)
```

Additionally, even when trying to use the Apex Replay Debugger without breakpoints, I get the following error popup message: **Unable to retrieve breakpoint info from language server, language server is not ready**

**VS Code Version**: 1.33

**SFDX CLI Version**: sfdx-cli/7.5.0-e6ca93eee8 darwin-x64 node-v10.15.3

**OS and version**: Mac OS Mojave 10.14

lcampos:
Thanks for reporting this @alpha-bytes. We've released a fix for it as part of version `45.12.1`, I'll close this for now but let us know if you run into any other issue.

PR #1290, opened 2019-04-25
Fix debuggers to get latest  status on apex language server

lcampos:
### What does this PR do?
Fixes apex debugger and replay debugger not getting the latest status of the Apex Language Server.

#1289

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/index.ts
+++ b/packages/salesforcedx-vscode-apex/src/index.ts
@@ -121,7 +121,7 @@ export async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(commands);
 
   context.subscriptions.push(await registerTestView(testOutlineProvider));
-  const languageClientStatus = languageClientUtils.getStatus();
+  const languageClientStatus = languageClientUtils;
 
   const exportedApi = {
     getLineBreakpointInfo,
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -121,7 +121,7 @@ export async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(commands);
 
   context.subscriptions.push(await registerTestView(testOutlineProvider));
-  const languageClientStatus = languageClientUtils.getStatus();
+  const languageClientStatus = languageClientUtils;
```
lcampos: Publishing `languageClientUtils.getStatus()` would not allow the consumers of this api to get the latest server status. Instead, they would get the status of the server at the point in which the apex extension got activated. Most of the time the status would be `Indexing`.

brpowell: I'd either change the name of the export from languageClientStatus to languageClientUtils, or export just getStatus with `const languageClientStatus = languageClientUtil.getStatus`.

AnanyaJha: So `getStatus()` is instead called by whatever is consuming `languageClientStatus`? ie. `debugConfigurationProvider.ts` to get the status at that point of time?

vazexqi: [nitpick] I agree with Bryan on the name.

lcampos: Cool, yes I'll send a follow up PR to clean up the naming.

lcampos: @AnanyaJha yes

PR #1298, opened 2019-04-30
[Org Browser] Skeleton UI

AnanyaJha:
### What does this PR do?
This PR lays the groundwork for the Org Browser UI.

@W-5767976@


packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+  private _onDidChangeTreeData: vscode.EventEmitter<
+    Node | undefined
+  > = new vscode.EventEmitter<Node | undefined>();
+  public readonly onDidChangeTreeData: vscode.Event<Node | undefined> = this
+    ._onDidChangeTreeData.event;
+
+  constructor() {}
+
+  public async refresh(): Promise<void> {
+    await this.getDefaultUsernameOrAlias();
```
AnanyaJha: refresh for the metadata type nodes should get called here

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  public getTreeItem(element: Node): vscode.TreeItem {
+    return element;
+  }
+
+  public getChildren(element?: Node): Promise<Node[]> {
+    if (isNullOrUndefined(element)) {
+      if (!isNullOrUndefined(this.defaultOrg)) {
+        const org = new Node(this.defaultOrg, NodeType.Org);
+        return Promise.resolve([org]);
```
AnanyaJha: getting the metadata type nodes should be called here

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+      } else {
+        return Promise.resolve([]);
+      }
+    } else {
+      return Promise.resolve(element.children);
```
AnanyaJha: getting metadata components for the selected type node should be called here

AnanyaJha:
So far, this UI will not load and display any data other than the current default org. If the org is switched, the new org does not get displayed right now because the filewatcher on `sfdx-config.json` has not been hooked up yet. The appropriate `when` criteria for displaying the view has been removed  right now so we don't display to users until the entire view is complete.

packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+import * as vscode from 'vscode';
+
+export enum NodeType {
+  Org,
+  MetadataType,
+  MetadataCmp
+}
+
+export class Node extends vscode.TreeItem {
```
brpowell: Another nit - "Node" seems generic when this is specifically for the org browser. Maybe MetadataNode or OrgBrowserNode or something.

packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -0,0 +1,35 @@
...
+  public children: Node[] = [];
+  constructor(label: string, public readonly type: NodeType) {
+    super(label);
+    this.type = type;
+    switch (this.type) {
+      case NodeType.Org:
+        this.collapsibleState = 1;
+        this.tooltip = 'Default Org';
```
brpowell: These tooltips are probably fine for now. When we start populating the treeView though we should remember to change these to either be dynamic based on the data, or localize these strings.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -75,12 +75,28 @@
...
         {
           "id": "sfdx.force.tasks.view",
           "name": "%running_tasks_title_text%"
         }
+      ],
+      "metadata": [
+        {
+          "id": "metadata",
+          "name": "Metadata",
+          "when": "sfdx:display_tree_view"
```
brpowell: I think you mentioned this was temporary. Later it'll probably be something like `sfdx:project_opened && sfdx:default_username_has_no_change_tracking`

packages/salesforcedx-vscode-core/package.json
```diff
@@ -75,12 +75,28 @@
   ],
   "main": "./out/src",
   "contributes": {
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "metadata",
+          "title": "Org Browser",
```
brpowell: Screen Shot 2019-05-01 at 4 16 35 PM
For reference this is where the title shows @ruthemmanuelle @ntotten

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -386,8 +401,9 @@ export async function activate(context: vscode.ExtensionContext) {
   // register org picker commands and set up filewatcher for defaultusername
   const orgList = new OrgList();
   await orgList.displayDefaultUsername();
-  context.subscriptions.push(registerOrgPickerCommands(context, orgList));
-
+  context.subscriptions.push(registerOrgPickerCommands(orgList));
+  await setupOrgBrowser(context);
```
lcampos: I think we should change the way we initialize the org browser and org picker because right now both of them are calling for the same info during activation (defaultUsernameOrAlias). 

We should probably also comment out this since we're executing code during activation that is not available, I want to make sure activation is as streamlined as possible.

packages/salesforcedx-vscode-core/resources/dep.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M12 5c3.453 0 5.891 2.797 5.567 6.78 1.745-.046 4.433.751 4.433 3.72 0 1.93-1.57 3.5-3.5 3.5h-13c-1.93 0-3.5-1.57-3.5-3.5 0-2.797 2.479-3.833 4.433-3.72-.167-4.218 2.208-6.78 5.567-6.78zm0-2c-4.006 0-7.267 3.141-7.479 7.092-2.57.463-4.521 2.706-4.521 5.408 0 3.037 2.463 5.5 5.5 5.5h13c3.037 0 5.5-2.463 5.5-5.5 0-2.702-1.951-4.945-4.521-5.408-.212-3.951-3.473-7.092-7.479-7.092z"/></svg>
```
lcampos: Where did we get this svg from ? I want to know if we need to add any license reference to the core package.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+import { isNullOrUndefined } from 'util';
+import * as vscode from 'vscode';
+import { hasRootWorkspace, OrgAuthInfo } from '../util';
+import { BrowserNode, NodeType } from './index';
+
+export class MetadataOutlineProvider
+  implements vscode.TreeDataProvider<BrowserNode> {
+  private defaultOrg: string | undefined;
+
+  private _onDidChangeTreeData: vscode.EventEmitter<
```
lcampos: We don't use the underscore prefix for private methods, setting it as private is enough.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+    return element;
+  }
+
+  public getChildren(element?: BrowserNode): Promise<BrowserNode[]> {
+    if (isNullOrUndefined(element)) {
+      if (!isNullOrUndefined(this.defaultOrg)) {
+        const org = new BrowserNode(this.defaultOrg, NodeType.Org);
+        return Promise.resolve([org]);
+      } else {
+        return Promise.resolve([]);
```
lcampos: Shouldn't this be a reject ? Are we going to show a message to the user that no org/defaultusername is available for use ?

packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -0,0 +1,35 @@
...
+        break;
+      case NodeType.MetadataCmp:
+        this.collapsibleState = 0;
+        this.tooltip = 'Metadata Component';
+        break;
+      case NodeType.MetadataType:
+        this.collapsibleState = 1;
```
praksb: Instead of setting it to 1, using vscode.TreeItemCollapsibleState.Collapsed would be more readable.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: There were a few places across our repo where I'd seen the underscore prefix be used specifically for this tree view method, but I can remove it. I'll change the name to internalOnDidChangeTreeData or something of the sort so it doesn't conflict with the the public onDidChangeTreeData method.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: I was playing around with actually displaying a message when there is no org set, in which case we would actually be returning a node with the label set to a message and the collapsible state set to none. In that case, I think it would remain Promise.resolve

packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
AnanyaJha: Yeah, that sounds good.

packages/salesforcedx-vscode-core/src/index.ts
AnanyaJha: That makes sense, I was thinking about making a call to retrieve defaultUsernameOrAlias here and just passing the info to org picker + browser. I can make that change as part of this PR if needed.

packages/salesforcedx-vscode-core/resources/dep.svg
AnanyaJha: I'm updating this to use an icon from the Lightning Design System. Will we need to add any license reference in that case?

lcampos: Why do we want an icon from lds ? In this type of scenario we try to use the icons shipped as part of vscode (octicons)

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
lcampos: Cool, thanks for clarifying

packages/salesforcedx-vscode-core/src/index.ts
lcampos: I'm ok if you want to do the refactoring of initializing org picker & browser on a follow up PR. I just want to make sure that this PR does not add code that's not going to be accessible as part of the extension's `activate` phase. 

You could comment `await setupOrgBrowser(context)` or modify it so that nothing inside of `setupOrgBrowser` is executed.

packages/salesforcedx-vscode-core/resources/dep.svg
AnanyaJha: The octicons actually only have cloud upload & cloud download options: 
Screen Shot 2019-05-07 at 11 25 46 AM
Screen Shot 2019-05-07 at 11 27 27 AM

I thought that might be confusing because users are able to do deploy & retrieve with the org browser/working set so I went with just the cloud icon from lds:
Screen Shot 2019-05-07 at 11 28 26 AM
I'm open to using whichever one looks best though!

ntotten: The Salesforce icon is best I think for this case.

AnanyaJha: Cool, I'll leave it as is then.

PR #1316, opened 2019-05-06
[Org Browser] Lazy loading for metadata components

AnanyaJha:
### What does this PR do?
This PR does the groundwork for retrieving metadata component information for a given metadata type. For types that support folders, we are only fetching the components in the top level public folders at this point.

@W-5980323@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
@@ -16,7 +16,6 @@ import * as path from 'path';
 import { Observable } from 'rxjs/Observable';
 import { isNullOrUndefined } from 'util';
 import * as vscode from 'vscode';
-import { channelService } from '../channels';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
@@ -24,7 +23,7 @@ import {
   SfdxWorkspaceChecker
 } from '../commands';
 import { nls } from '../messages';
-import { notificationService, ProgressNotification } from '../notifications';
+import { notificationService } from '../notifications';
 import { taskViewService } from '../statuses';
 import { telemetryService } from '../telemetry';
 import { getRootWorkspacePath, hasRootWorkspace, OrgAuthInfo } from '../util';
@@ -66,15 +65,12 @@ export class ForceListMetadataExecutor extends SfdxCommandletExecutor<string> {
 
     execution.processExitSubject.subscribe(async data => {
       this.logMetric(execution.command.logName, startTime);
-      console.log(this.outputPath);
       buildComponentsList(this.outputPath, this.metadataType);
     });
     notificationService.reportExecutionError(
       execution.command.toString(),
       (execution.stderrSubject as any) as Observable<Error | undefined>
     );
-    channelService.streamCommandOutput(execution);
-    ProgressNotification.show(execution, cancellationTokenSource);
     taskViewService.addCommandExecution(execution, cancellationTokenSource);
   }
 }
@@ -120,10 +116,9 @@ export async function getComponentsPath(
         metadataType + '.json'
       );
       return componentsPath;
-    } catch {
-      const err = nls.localize('error_no_default_username');
-      telemetryService.sendError(err);
-      throw new Error(err);
+    } catch (e) {
+      telemetryService.sendError(e);
+      throw new Error(e);
     }
   } else {
     const err = nls.localize('cannot_determine_workspace');
@@ -149,7 +144,6 @@ export function buildComponentsList(
       { metadataType },
       { metadataComponents: metaComponents.length }
     );
-    console.log(metaComponents);
     return metaComponents;
   } catch (e) {
     throw e;
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
@@ -45,13 +45,22 @@ export class ForceListMetadataExecutor extends SfdxCommandletExecutor<string> {
   }
 
   public build(data: {}): Command {
-    return new SfdxCommandBuilder()
+    let builder = new SfdxCommandBuilder()
       .withArg('force:mdapi:listmetadata')
       .withFlag('-m', this.metadataType)
       .withFlag('-u', this.defaultUsernameOrAlias)
       .withFlag('-f', this.outputPath)
-      .withJson()
-      .build();
+      .withJson();
+
+    if (
+      this.metadataType === 'Dashboard' ||
+      this.metadataType === 'Report' ||
+      this.metadataType === 'EmailTemplate' ||
+      this.metadataType === 'Document'
+    ) {
+      builder = builder.withFlag('--folder', 'unfiled$public');
+    }
+    return builder.build();
   }
 
   public execute(response: ContinueResponse<string>): void {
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,160 @@
...
+    this.outputPath = outputPath;
+    this.defaultUsernameOrAlias = defaultUsernameOrAlias;
+  }
+
+  public build(data: {}): Command {
+    let builder = new SfdxCommandBuilder()
+      .withArg('force:mdapi:listmetadata')
+      .withFlag('-m', this.metadataType)
+      .withFlag('-u', this.defaultUsernameOrAlias)
+      .withFlag('-f', this.outputPath)
```
brpowell: Are we using output files instead of directly passing the data to the tree provider because we're planning on having an org's component tree persist between vs code sessions?

cc: @lcampos

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    );
+    const forceDescribeMetadataCmd = forceListMetadataExec.build({});
+    expect(forceDescribeMetadataCmd.toCommand()).to.equal(
+      `sfdx force:mdapi:listmetadata -m ${metadataType} -u ${defaultUsername} -f ${outputPath} --json --loglevel fatal`
+    );
+  });
+
+  it('Should build list metadata command with folder arg', async () => {
+    const outputPath = 'outputPath';
+    const metadataType = 'Report';
```
brpowell: We should test all of the "folder types" to make sure none of them have a command built without a folder param. Perhaps in the command file there could be an exported set of those types and iterate through each one expecting the right command. Not to mention when building the command (line 55, metadataCmp.ts), the set can be used to check for membership of a given type instead of having a conditional check for each type built.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
lcampos: @brpowell yes, we eventually want to allow users to move across orgs and having all the info in memory is not a good option. We'll store the data under `.sfdx` and will access it and update it based on the user's actions via the UI.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,160 @@
...
+  ) {
+    super();
+    this.metadataType = metadataType;
+    this.outputPath = outputPath;
+    this.defaultUsernameOrAlias = defaultUsernameOrAlias;
+  }
+
+  public build(data: {}): Command {
+    let builder = new SfdxCommandBuilder()
+      .withArg('force:mdapi:listmetadata')
```
praksb: how about having this as a const?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,160 @@
...
+      .withFlag('-m', this.metadataType)
+      .withFlag('-u', this.defaultUsernameOrAlias)
+      .withFlag('-f', this.outputPath)
+      .withJson();
+
+    if (
+      this.metadataType === 'Dashboard' ||
```
praksb: instead of what you have here, how about initializing a Set with 'Dashboard', 'Report' etc.. and doing a set.has(this.metadataType) check?

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
AnanyaJha: Yeah, I think this is a good idea!

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
AnanyaJha: Yes, I think this could work too. I'll incorporate this along with Bryan's suggestion

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
AnanyaJha: I didn't save it as a constant to follow the format of the other commands in our core package. However, I realized that the work item for pr specifies that we do want to display the progress notification for the command to the user. In that case, should we also specify a description field for the command builder?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+    telemetryService.sendEventData(
+      'Metadata Components quantity',
+      { metadataType },
+      { metadataComponents: metaComponents.length }
+    );
+    return metaComponents;
+  } catch (e) {
+    throw e;
```
lcampos: Lets log and throw the error similar to how you're doing in `getComponentsPath`

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,160 @@
...
+      .withFlag('-f', this.outputPath)
+      .withJson();
+
+    if (
+      this.metadataType === 'Dashboard' ||
+      this.metadataType === 'Report' ||
+      this.metadataType === 'EmailTemplate' ||
+      this.metadataType === 'Document'
+    ) {
+      builder = builder.withFlag('--folder', 'unfiled$public');
```
lcampos: There's an issue with this, only EmailTemplate and Report types have an out-of-the-box public folder called `unfiled$public`.
Overall how are we thinking about presenting this to the user ? Is the user going to see a top-level option in the tree-view called 'Report', then see only the reports under `unfiled$public` when they expand it ?

AnanyaJha: Overall, we were thinking of having a top-level option ie. `Report` that would load the reports from `unfiled$public` and the custom folders that contain the rest of the reports. Loading the custom folders under these types will be done by the work item @brpowell is on. 
We were under the impression that the `unfiled$public` folder exists for all four types, Bryan and I will sync on how to handle this now

AnanyaJha: For now, we'll just list the `unfiled$public` data for EmailTemplate and Report. Dashboards and Documents will just display the other folders under them when expanded.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -0,0 +1,168 @@
...
+export async function getComponentsPath(
+  metadataType: string,
+  defaultUsernameOrAlias: string
+): Promise<string> {
+  if (hasRootWorkspace()) {
+    try {
+      const workspaceRootPath = getRootWorkspacePath();
+      const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
+
+      if (isNullOrUndefined(username)) {
```
lcampos: Added this since now `OrgAuthInfo.getUsername(...)` returns string or undefined.

PR #1318, opened 2019-05-07
Update activation logic for default username information

AnanyaJha:
### What does this PR do?
This PR updates the activation logic for the org picker, org browser, and workspace org type to only collect default username information once.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -352,10 +352,12 @@ function registerOrgPickerCommands(orgList: OrgList): vscode.Disposable {
 }
 
 async function setupOrgBrowser(
-  extensionContext: vscode.ExtensionContext
+  extensionContext: vscode.ExtensionContext,
+  defaultUsernameOrAlias?: string
 ): Promise<void> {
-  const metadataTreeProvider = new MetadataOutlineProvider();
-  await metadataTreeProvider.getDefaultUsernameOrAlias();
+  const metadataTreeProvider = new MetadataOutlineProvider(
+    defaultUsernameOrAlias
+  );
   const metadataProvider = vscode.window.registerTreeDataProvider(
     'metadata',
     metadataTreeProvider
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -35,14 +35,8 @@ export class OrgList implements vscode.Disposable {
...
 
-  public async displayDefaultUsername() {
-    let defaultUsernameorAlias: string | undefined;
-    if (hasRootWorkspace()) {
-      defaultUsernameorAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(
-        false
-      );
-    }
-    if (defaultUsernameorAlias) {
+  public async displayDefaultUsername(defaultUsernameorAlias?: string) {
```
lcampos: This doesn't need to be `async` anymore.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -177,8 +171,14 @@ export class OrgList implements vscode.Disposable {
...
-    await setupWorkspaceOrgType();
-    await this.displayDefaultUsername();
+    let defaultUsernameorAlias: string | undefined;
+    if (hasRootWorkspace()) {
+      defaultUsernameorAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(
+        false
+      );
+    }
+    await setupWorkspaceOrgType(defaultUsernameorAlias);
+    await this.displayDefaultUsername(defaultUsernameorAlias);
```
lcampos: You can remove this `await`.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -398,16 +400,21 @@ export async function activate(context: vscode.ExtensionContext) {
...
 
+  let defaultUsernameorAlias: string | undefined;
+  if (hasRootWorkspace()) {
+    defaultUsernameorAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(false);
+  }
+
   // register org picker commands and set up filewatcher for defaultusername
   const orgList = new OrgList();
-  await orgList.displayDefaultUsername();
+  await orgList.displayDefaultUsername(defaultUsernameorAlias);
```
lcampos: You can remove this `await` after removing the `async` from `displayDefaultUsername`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
AnanyaJha: Ah good catch

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
AnanyaJha: Cool

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -352,10 +352,12 @@ function registerOrgPickerCommands(orgList: OrgList): vscode.Disposable {
 }
 
 async function setupOrgBrowser(
```
brpowell: Can this setup function be defined in the orgBrowser module and imported instead? Might be good to reduce crowding this file when we can.

AnanyaJha: I was thinking about that initially too but ended up deciding against it for now. If we want to move these helper setup methods over to their respective files, I can definitely do that in another pr.

Issue #1196, opened 2019-03-27
Context that shows/hides commands is not refreshed when switching orgs


lcampos:
### Summary

The context set for deploy/retrieve commands is not being refreshed when a user moves between different org types or creates a new scratch org on a new project. This is causing users not being able to access commands like `SFDX: Push Source to Default Scratch Org`, `SFDX: Deploy Source in Manifest to Org`, etc.

### Steps To Reproduce:

1.  Create a new project `SFDX : Create Project with Manifest`
2. Create a scratch org `SFDX: Create a Default Scratch Org`
3. Create valid apex code `SFDX: Create Apex Class`
4. Try and deploy the new code to the newly create scratch org by opening the command palette and looking for `SFDX: Push Source to Default Scratch Org`

### Expected result
I see the push command on the command palette and I'm able to push code to the scratch org.

### Actual result
Context is not updated and commands remain hidden to the user. The user needs to reload the vscode window in order for the commands to show.

**Salesforce VS Code Extensions Version**: 45.6.0 and above
**VS Code Version**: 1.32.3
**SFDX CLI Version**: 6.56.0-e3fd846a1f
**OS and version**: Mac OS High Sierra

deeelan:
I'm experience the same thing. Do anyone have the voluntee of fixing this? I have this issue almost three months ago. And any thread of this issue dont solve anything or its closed and reference to another thread with no solution. I've been working on Mac OS High Sierra and Windows 10. 

My solution is: when the options are displaying in the context menu, not update VSCode. It seems that everytime that VSCode launch an update it break it.

deeelan:
Personally I think this not should happen, cuz the coding in Salesforce it supossed to b "full-supported" on VSCode

garviand:
I don't get any of the Push/Deploy options on right-click, I am also not getting the testing options for Apex tests.

Patlatus:
I have the same problem. When I create a Salesforce project, deploy\push\retrieve menu items disappear randomly and I am not able to restore them and get them back. Not sure if this is connected to switching between orgs. I just created a scratch org and I didn't switch org, however, initially those options were present and suddenly they disappear and I am not able to restore them.

Patlatus:
Also my OS is not Mac, I am using Windows.

HernanMorgado:
[...]

PR #1327, opened 2019-05-09
Fix issue with context not being properly updated when creating scrat…

lcampos:
…ch orgs

### What does this PR do?
Refactor needed to address the issue where command context is not properly update when creating scratch orgs. 
Root cause: because of the latest perf updates we've shipped, we are running a bit faster than how the cli sets scratch auth context. This causes us not to receive enough information to set the correct context used to control visibility of some commands (e.g. Push/Pull & Deploy/Retrieve).

Issue #1196, @W-5988233@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
@@ -118,14 +118,20 @@ export async function getUserId(projectPath: string): Promise<string> {
   const defaultUsernameOrAlias = await getDefaultUsernameOrAlias();
   if (isNullOrUndefined(defaultUsernameOrAlias)) {
     const err = nls.localize('error_no_default_username');
-    telemetryService.sendError(err);
+    telemetryService.sendErrorEvent(
+      'Undefined username or alias when starting Apex Replay Debugger',
+      err
+    );
     throw new Error(err);
   }
 
   const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
   if (isNullOrUndefined(username)) {
     const err = nls.localize('error_no_default_username');
-    telemetryService.sendError(err);
+    telemetryService.sendErrorEvent(
+      'Undefined username when starting Apex Replay Debugger',
+      err
+    );
     throw new Error(err);
   }
 
--- a/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
@@ -27,6 +27,7 @@ import {
   SfdxWorkspaceChecker
 } from './commands';
 
+import { isNullOrUndefined } from 'util';
 import { getDefaultUsernameOrAlias } from '../context';
 import { telemetryService } from '../telemetry';
 import { developerLogTraceFlag } from './';
@@ -115,7 +116,19 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
 
 export async function getUserId(projectPath: string): Promise<string> {
   const defaultUsernameOrAlias = await getDefaultUsernameOrAlias();
-  const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias!);
+  if (isNullOrUndefined(defaultUsernameOrAlias)) {
+    const err = nls.localize('error_no_default_username');
+    telemetryService.sendError(err);
+    throw new Error(err);
+  }
+
+  const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
+  if (isNullOrUndefined(username)) {
+    const err = nls.localize('error_no_default_username');
+    telemetryService.sendError(err);
+    throw new Error(err);
+  }
+
   const execution = new CliCommandExecutor(
     new ForceQueryUser(username).build(),
     {
--- a/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
+++ b/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
@@ -14,20 +14,37 @@ export enum OrgType {
   NonSourceTracked
 }
 
+function imposeSlightDelay(ms = 0) {
+  return new Promise(r => setTimeout(r, ms));
+}
+
 export async function getWorkspaceOrgType(
   defaultUsernameOrAlias?: string
 ): Promise<OrgType> {
-  if (!isNullOrUndefined(defaultUsernameOrAlias)) {
-    const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias!);
-    const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username).catch(err =>
-      telemetryService.sendError(err)
-    );
-    return isScratchOrg ? OrgType.SourceTracked : OrgType.NonSourceTracked;
+  if (isNullOrUndefined(defaultUsernameOrAlias)) {
+    const e = new Error();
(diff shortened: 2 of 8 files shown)


packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -13,21 +14,36 @@ export enum OrgType {
...
+    throw e;
+  }
 
-  if (defaultUsernameIsSet) {
-    const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias!);
-    const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username).catch(err =>
-      telemetryService.sendError(err)
+  let username;
+  let counter = 0;
+  while (isNullOrUndefined(username) && counter < 30) {
```
lcampos: Retrying here is what allows us to get the correct info from the cli when creating scratch orgs.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -13,21 +14,36 @@ export enum OrgType {
...
+    username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
+    counter += 1;
+    await imposeSlightDelay(100);
+  }
+
+  if (isNullOrUndefined(username)) {
```
lcampos: Added this to have some visibility in case further tweaking is needed.

packages/salesforcedx-vscode-core/src/orgBrowser/orgMetadata.ts
```diff
@@ -88,34 +88,40 @@ export async function forceDescribeMetadata(outputPath?: string) {
...
-        'metadataTypes.json'
-      );
-      return metadataTypesPath;
-    } else {
-      const err = nls.localize('error_no_default_username');
-      telemetryService.sendError(err);
-      throw new Error(err);
-    }
-  } else {
+  if (!hasRootWorkspace()) {
```
lcampos: Same functionality, I just refactored it in order to have less if/else's and error out as soon as possible.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -78,12 +78,11 @@ export class OrgAuthInfo {
     }
   }
 
-  public static async getUsername(usernameOrAlias: string): Promise<string> {
+  public static async getUsername(
+    usernameOrAlias: string
+  ): Promise<string | undefined> {
```
lcampos: This became an issue after I added the retry in `workspaceOrgType.ts`. Turns out if `username` was `undefined` we ended up returning the original parameter provided, making it difficult to identify if we had the correct info to keep going. This change simplifies the method and is the reason why I had to update all the other code outside of `workspaceOrgType.ts`.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -115,7 +116,19 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
...
-  const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias!);
+  if (isNullOrUndefined(defaultUsernameOrAlias)) {
+    const err = nls.localize('error_no_default_username');
+    telemetryService.sendError(err);
+    throw new Error(err);
+  }
+
+  const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
+  if (isNullOrUndefined(username)) {
+    const err = nls.localize('error_no_default_username');
```
praksb: The error we throw here is the same as in line 120. We won't be able to distinguish from the telemetry data which condition failed.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
praksb: can we do better here? i am not familiar with the code here but i wonder if there is a way to receive a notification when cli sets the scratch org context.

lcampos: Unfortunately no, the cli is not a server that sends notifications we can subscribe to. It's more of a sync api request an we get the data that's currently available.

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
lcampos: That's true, I can make them unique for telemetry purposes but keep it as is for the end user since the message still makes sense in both situations.

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceOrgType.test.ts
```diff
@@ -22,7 +22,7 @@ describe('getUsername', () => {
   it('should return the username when given a username', async () => {
```
brpowell: Should this test description be updated?

lcampos: Yes, i'll do that 👍

packages/salesforcedx-vscode-core/src/commands/forceStartApexDebugLogging.ts
```diff
@@ -115,7 +116,25 @@ export class ForceStartApexDebugLoggingExecutor extends SfdxCommandletExecutor<{
...
+  if (isNullOrUndefined(defaultUsernameOrAlias)) {
+    const err = nls.localize('error_no_default_username');
+    telemetryService.sendErrorEvent(
+      'Undefined username or alias when starting Apex Replay Debugger',
+      err
+    );
+    throw new Error(err);
+  }
+
+  const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
```
brpowell: Is this being called to guarantee we get a username back and not an alias?

AnanyaJha: Yeah, the getUsername method returns the associated username if it's provided an alias. Otherwise, it'll return the same username back

AnanyaJha:
The changes look good to me! With the activation logic PR we just merged, would it be a good idea for us to update the onSfdxConfigEvent filewatcher as well? That filewatcher now attempts to retrieve the defaultusername info and passes that workspaceOrgType and orgPicker

lcampos:
> The changes look good to me! With the activation logic PR we just merged, would it be a good idea for us to update the `onSfdxConfigEvent` filewatcher with the logic currently added to `getWorkspaceOrgType()` as well? That filewatcher now attempts to retrieve the defaultusername info and passes that workspaceOrgType and orgPicker.

Good catch, I'll take a look at that and get it updated as part of it.

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -41,7 +42,13 @@ export async function registerPushOrDeployOnSave() {
 
 export async function pushOrDeploy(filesToDeploy: vscode.Uri[]): Promise<void> {
   try {
-    const orgType = await getWorkspaceOrgType();
+    let defaultUsernameorAlias: string | undefined;
```
lcampos: Updated this scenario since I forgot to in the previous changes.

lcampos:
@AnanyaJha I checked and the `onSfdxConfigEvent` filewatcher didn't need to get updated but I did find a missing update on `pushOrDeployOnSave.ts`. Let me know if I'm missing something about the filewatcher and it still needs to get updated.

PR #1329, opened 2019-05-10
Bp/org picker msgs

brpowell:
### What does this PR do?

Removes warning message when opening the org picker that a default dev hub is not set. Every time the list was opened the message was written twice in the output console and a notification pops up. Also only show option for creating a scratch org if we have a dev hub set, or if we don't have a dev hub set show the option to authorize one.

W-6104719

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
+++ b/packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
@@ -165,12 +165,6 @@ export class OrgList implements vscode.Disposable {
     }
   }
 
-  public async getDefaultUsernameOrAlias() {
-    if (hasRootWorkspace()) {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -160,9 +165,15 @@ export class OrgList implements vscode.Disposable {
     }
   }
 
+  public async getDefaultUsernameOrAlias() {
```
AnanyaJha: Where is this method getting used? It looks like it has the same functionality as the `getDefaultDevHubUsernameOrAlias(`) in this class

brpowell: Good catch, that was junk i forgot to take out

PR #1340, opened 2019-05-14
[Org Browser] Fix for no components present

AnanyaJha:
### What does this PR do?
This PR addresses the issues found during QA- no components found for a given metadata type and the correct metadata name for Report.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
@@ -28,7 +28,7 @@ import { taskViewService } from '../statuses';
 import { telemetryService } from '../telemetry';
 import { getRootWorkspacePath, hasRootWorkspace, OrgAuthInfo } from '../util';
 
-export const folderTypes = new Set(['EmailTemplate', 'Reports']);
+export const folderTypes = new Set(['EmailTemplate', 'Report']);
 export class ForceListMetadataExecutor extends SfdxCommandletExecutor<string> {
   private metadataType: string;
   private outputPath: string;
@@ -112,8 +112,10 @@ export async function getComponentsPath(
   if (hasRootWorkspace()) {
     try {
       const workspaceRootPath = getRootWorkspacePath();
-      const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
-
+      let username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
+      if (!username) {
+        username = defaultUsernameOrAlias;
+      }
       if (isNullOrUndefined(username)) {
         const err = nls.localize('error_no_default_username');
         telemetryService.sendErrorEvent(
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -112,8 +112,8 @@ export async function getComponentsPath(
   if (hasRootWorkspace()) {
     try {
       const workspaceRootPath = getRootWorkspacePath();
-      const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
-
+      let username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
```
AnanyaJha: This had to be updated due to the changes recently made to the `getUsername` method. If the `defaultUsernameOrAlias` is not an alias, then the `getUsername` method now returns undefined. In that scenario, we want to continue using the default value initially provided.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -112,8 +112,8 @@ export async function getComponentsPath(
...
+      username = username ? username : defaultUsernameOrAlias;
       if (isNullOrUndefined(username)) {
```
lcampos: With these changes the `if (isNullOrUndefined(username)) {` statement is not needed anymore.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -148,11 +148,14 @@ export function buildComponentsList(
   metadataType: string
 ): string[] {
   try {
-    const fileData = JSON.parse(fs.readFileSync(componentsPath, 'utf8'));
     const metaComponents = [];
-    for (const component of fileData) {
-      if (!isNullOrUndefined(component.fullName)) {
-        metaComponents.push(component.fullName);
+    const fileData = fs.readFileSync(componentsPath, 'utf8');
+    if (fileData !== 'undefined') {
```
lcampos: This is for covering when the file gets created bu the content is undefined instead of a JSON structure, right ?

AnanyaJha: Yup, exactly

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
brpowell: A fun trick in cases like this is `let username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias) || defaultUsernameOrgAlias;`. Not necessary though :)

AnanyaJha: Ooooh that’s literally exactly what I was trying to figure out how to do!! You rock I’m updating this

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -148,11 +140,14 @@ export function buildComponentsList(
...
+      const fileObject = JSON.parse(fileData);
+      for (const component of fileObject) {
```
brpowell: Due to the nature of the api, if the response only returns one item fileObject will be an Object instead of a List. The side effect is that you'll be iterating over the keys of the object and not the object itself. This should be handled here

PR #1345, opened 2019-05-15
Add docs for Declare missing methods quickfix

praksb:
### What does this PR do?
Update apex docs to include the recently added "Declare missing methods" quick fix.


lcampos:
Do we need this to be published as part of this week's release ? If so, we'll need to port the change to the release branch as well.

PR #1349, opened 2019-05-15
add Apex doc block comment to auto-closing pairs

no-stack-dub-sack:
### What does this PR do?
Adds doc block comments, e.g. `/** */` as an auto closing pair to Apex's language config.

No issue which is currently open, however, I'm wondering if you guys would be willing to accept this change. Overall, it's pretty innocuous, and not super opinionated, and would come in handy for anyone trying to use a Java-Doc like solution on their Apex code. I'm currently building a VS Code extension which is a TypeScript re-write of the original ApexDoc, which will make using such a utility much easier. Having an this as an auto closing pair will greatly help the extension along. Please let me know your thoughts, and thanks for considering either way.

P.S. the change in format was due to the format on save feature prescribed by the repo's settings


lcampos:
Hi @no-stack-dub-sack , we need you to sign the CLA in order for us to consider merging your change.

no-stack-dub-sack:
@lcampos CLA signed, thanks.

PR #1350, opened 2019-05-16
Add internal dev context. Show lightning create commands and hide pro…

lcampos:
…ject create commands when internal dev context is true.

### What does this PR do?
Create a new context for usage of the extensions for internal development

@W-6151369@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -69,6 +69,7 @@
   },
   "activationEvents": [
     "workspaceContains:sfdx-project.json",
+    "workspaceContains:core-workspace.code-workspace",
```
lcampos: @nrkruk this is the new workspace file that will looking for. Let me know what's the final name we'll go with. I suggest we pick one that hints to be specific to us so we don't activate the extensions unnecessarily.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -364,9 +352,49 @@
           "command": "sfdx.force.source.delete.current.file",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest' && editorIsOpen"
         },
+        {
+          "command": "sfdx.force.set.default.org",
```
lcampos: This command was always displayed as part of the command palette because we forgot to add it here.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -364,9 +352,49 @@
...
+          "when": "sfdx:project_opened"
+        },
         {
           "command": "sfdx.force.lightning.lwc.create",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.force.lightning.lwc.create",
+          "when": "sfdx:internal_dev"
```
lcampos: Consolidated all the lightning create commands here. Unfortunately, vscode's when clause doesn't support the use of OR statements so I had to duplicate the command entries.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -395,6 +395,15 @@ export async function activate(context: vscode.ExtensionContext) {
     sfdxProjectOpened
   );
 
+  // Set internal dev context
+  const internalDev = sfdxCoreSettings.getInternalDev();
```
lcampos: The piece of code that will populate a global context according to whether it finds the new setting `salesforcedx-vscode-core.internal-development` and its value .

packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -117,6 +117,12 @@
           "when": "view == salesforce-lightning-components && viewItem == external",
           "group": "inline"
         }
+      ],
+      "commandPalette": [
```
lcampos: Updated this since `SFDX: Open Lightning Component` was part of the command palette in scenarios where the extension do not activate and ends up throwing errors.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -222,6 +228,38 @@ export class SelectOutputDir
   }
 }
 
+export class SelectOutputInternalDevDir
+  implements ParametersGatherer<{ outputdir: string }> {
+  public constructor() {}
+
+  public async gather(): Promise<
```
lcampos: This is the code that will scan the project, starting from its root, and display a list of all the folders inside it when the user runs a create command and has entered the metadata's name. 

@ntotten or @nrkruk, I need to know how many folders & levels of depth we need to show when a user is running the create commands so showing a list of available folders is as performant as possible.

Here's a screenshot of the step I'm referring to:
Screen Shot 2019-05-16 at 7 28 18 AM

packages/salesforcedx-vscode-core/package.json
nrkruk: Looking at this now, we don't actually need any changes here I don't think.

packages/salesforcedx-vscode-core/src/commands/commands.ts
nrkruk: Core modules look like:
core/ui-global-components/components/{namespace}/{name}/{name}.cmp
core/ui-global-components/modules/one/{namespace}/{name}/{name}.html

So we can filter on directories named components for aura commands and directories named modules for lwc commands.

Also, I think we probably just want the right click -> create component on a folder.

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+import { telemetryService } from '../../telemetry';
+import { AURA_METADATA_TYPE, LWC_METADATA_TYPE } from './metadataTypeConstants';
+
+export class InternalDevWorkspaceChecker implements PreconditionChecker {
+  public check(): boolean {
+    return sfdxCoreSettings.getInternalDev();
+  }
+}
+
+export class FileInternalPathGatherer
```
lcampos: The old approach of presenting a list of available folders, even when filtering to exclusively the valid ones, was not performant. Got rid of that code and replace it with this one which takes the file path from the context of the user right-clicking a folder.

FYI: I might add one more validation here, depending on some tests I'm running.

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+    CancelResponse | ContinueResponse<{ outputdir: string }>
+  > {
+    const outputdir = this.filePath;
+    return outputdir
+      ? { type: 'CONTINUE', data: { outputdir } }
+      : { type: 'CANCEL' };
+  }
+}
+
+export class InternalSourcePathChecker
```
lcampos: Validates that the file path is the correct one based on the type of metadata that the user is trying to create.

lcampos:
@nkruk & @ntotten with the latest updates, this dev model will only allow using the lightning create commands by right-clicking on folders. Showing the commands as part of the command palette was not performant and I will not work on enabling them in this release. At this moment I think the only thing missing is add & updating tests.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -443,18 +470,34 @@
         "command": "sfdx.force.lightning.app.create",
         "title": "%force_lightning_app_create_text%"
       },
+      {
+        "command": "sfdx.internal.lightning.app.create",
+        "title": "%force_lightning_app_create_text%"
```
nrkruk: its unclear to me why these have to be duplicated w/ a separate command id? I thought they could be the same command and internally we just check a setting as to what parameters to pass...

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
nrkruk: Will you still be able to run these form the command palette or is it required now that you right click a folder?

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
nrkruk: I don't think we should enforce this

packages/salesforcedx-vscode-lightning/package.json
nrkruk: I don't think we want to key off of sfdx:project_opened here. Otherwise it won't show up in scenarios that it should. Lets just add an activation even for "onCommand:salesforce-lightning-quickopen" instead

packages/salesforcedx-vscode-core/package.json
lcampos: I want to have a clear separation between how we run a internal command vs a normal one. That meant creating an additional entry point where I can define preconditions, param gatherers and postcondition methods that are exclusive to the internal mode e.g. `forceInternalLightningInterfaceCreate` in `forceLightningInterfaceCreate.ts`.

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
lcampos: No, we won't do command palette in this release. I didn't find a performant way for the commands to work using the command palette in a big project even after ignoring invalid folder paths.

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
lcampos: Why not? This is based on the source path conversation we had earlier, where `aura` metadata should only be created in source paths that contain `/components/{namespace}` and `/modules/{namespace}` for the `lwc` scenarios.

packages/salesforcedx-vscode-lightning/package.json
lcampos: This is causing that command to show in scenarios where nothing related to lightning is present and throws an error if the user runs it.

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
nrkruk: This is true for a majority of projects in core, but lwc can be configured to read modules from any directory. I just don't see any value in limiting where internal devs/ off-core devs can create these components, especially now that you have it working with a right click action on a folder.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -28,8 +28,8 @@
     "@salesforce/salesforcedx-utils-vscode": "45.15.1",
     "ajv": "^6.1.1",
     "eslint": "5.0.0",
-    "lightning-lsp-common": "2.2.2",
-    "lwc-language-server": "2.2.2",
+    "lightning-lsp-common": "2.2.4",
+    "lwc-language-server": "2.2.4",
```
nrkruk: Make sure you update the aura-language-server to 2.2.4 also

nrkruk: In the salesforcedx-vscode-lightning package

packages/salesforcedx-vscode-core/src/commands/templates/internalCommandUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.metadataType = metadataType;
+  }
+  public async check(
+    inputs: ContinueResponse<{ outputdir: string }> | CancelResponse
+  ): Promise<ContinueResponse<{ outputdir: string }> | CancelResponse> {
+    if (inputs.type === 'CONTINUE') {
+      const sourcePath = inputs.data.outputdir;
+      if (
+        this.metadataType === AURA_METADATA_TYPE &&
+        sourcePath.indexOf('/components/') !== -1
```
lcampos: Note: This indexOf will not work on Windows.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -395,6 +435,15 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+  vscode.commands.executeCommand(
+    'setContext',
+    'sfdx:internal_dev',
+    internalDev
+  );
+
   let defaultUsernameorAlias: string | undefined;
```
lcampos: Need to add more logic to activation to avoid enabling features that won't work in this dev mode e.g. org picker.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -55,5 +55,6 @@
   "force_source_deploy_in_manifest_text": "SFDX: Deploy Source in Manifest to Org",
   "force_source_delete_text": "SFDX: Delete from Project and Org",
   "force_source_delete_this_source_text": "SFDX: Delete This from Project and Org",
-  "force_config_set_org_text": "SFDX: Set a Default Org"
+  "force_config_set_org_text": "SFDX: Set a Default Org",
+  "internal_dev_description": "Specifies whether or not to set up internal development mode which provides limited SFDX commands."
```
lcampos: @ruthemmanuelle this is the description users will get when they set the `internal_development` preference.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -368,6 +416,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
+  // Set internal dev context
+  const internalDev = sfdxCoreSettings.getInternalDev();
+
+  vscode.commands.executeCommand(
+    'setContext',
+    'sfdx:internal_dev',
+    internalDev
+  );
+
+  if (internalDev) {
```
lcampos: Doing this allows for setting a limited number of functionality in the core extension when the user has set internal development flag as true. E.g. no org picker or command palette commands.

packages/system-tests/src/spectron/application.ts
```diff
@@ -109,6 +109,10 @@ export class SpectronApplication {
     );
   }
 
+  public waitUI(): Promise<any> {
```
lcampos: Shorter wait method for usage when waiting on a UI element in vscode e.g. wait for the command palette to open.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -364,9 +371,49 @@
           "command": "sfdx.force.source.delete.current.file",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest' && editorIsOpen"
         },
+        {
+          "command": "sfdx.force.set.default.org",
+          "when": "sfdx:project_opened"
```
brpowell: Is my understanding correct? - "sfdx:project_opened" is for regular dx projects that are opened, and sfdx:internal_dev is to surface commands even for internal development?

lcampos: Yes, you are correct. sfdx:internal_dev should only enable the new lightning create commands called `sfdx.internal.lightning.` and use the new entry methods which have the new preconditions and gatherers.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/internalCommandUtils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      const response = await folderPathGatherer.gather();
+      expect(response.type).to.equal('CANCEL');
+    });
+
+    it('Should return Cancel if path does not exist', async () => {
+      existsSyncStub.returns(false);
+      const testDir = Uri.parse('file:///path/to/outside/dir');
+      lstatSyncStub.returns({
+        isDirectory() {
+          return false;
```
brpowell: Very minor nit, but I might make this return true so we guarantee we're not returning cancel because it's not a directory.

lcampos: The reason I set it to false is because this scenario tests for a path that doesn't exist. If the path does not exist (line 93) then isDirectory needs to return false.
I think the other tests cover the rest of the scenarios where the path exists and it's a directory or is a file. Let me know if there's other scenarios you're thinking about.

packages/salesforcedx-vscode-core/package.nls.json
ruthemmanuelle: Hi @lcampos . Does the pref do anything in addition to limiting the commands that are available?

lcampos: @ruthemmanuelle it's also meant for developing only lightning and/or lwc

PR #1368, opened 2019-05-25
Use previous language server versions

lcampos:
### What does this PR do?
Revert to previous language server versions


lcampos:
@vazexqi no, 2.2.3 has the changes to include new flags to settings.json and 2.2.4 is only package.json description updates.

PR #1388, opened 2019-06-10
Documentation for extract constant and extract local variable

sfsholden:
### What does this PR do?
Addresses documentation updates for new Apex LSP feature work - Extract Constant and Extract Local Variable.

@W-6224490@


docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
 
 ![GIF showing a renaming conflict](../../images/apex-rename-conflict.gif)
+
+## Refactor: Extract Constant
+
+You can extract valid literal expressions into constants. Literal expressions include: String, Integer, Long, Double, Decimal, and Boolean.
+
+This constant will be instantiated below the class declaration. If the expression is contained within an innerclass, the constant will be instantiated below the innerclass declaration. The new constant will always be defined with modifiers private, static, and final.
+
+At this time, there is no conflict detection for the new constant.
```
praksb: No need to mention this. "conflict detection" is also a term which may not be familiar to some.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -184,7 +184,10 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       )
       .withArg('force:mdapi:retrieve')
       .withFlag('--retrievetargetdir', this.relativeMetdataTempPath)
-      .withFlag('--packagenames', packageNames.join(','))
+      .withFlag(
```
lcampos: These changes seem unrelated to the documentation ones. We should split this into different PRs ?

praksb: Yes. they should be removed from this PR.

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
```
lcampos: I think we should include how the user would access the quick action e.g. To perform Extract Constant, select ...

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
 If the new name conflicts with an existing identifier name, we try to fully qualify the references to the existing name in contexts where the conflicts exist.
 
 ![GIF showing a renaming conflict](../../images/apex-rename-conflict.gif)
+
+## Refactor: Extract Constant
+
+You can extract valid literal expressions into constants. Literal expressions include: String, Integer, Long, Double, Decimal, and Boolean.
```
praksb: It is not clear to me what "valid literal expressions" mean. We can remove "valid" i think. Also, no need to specify "literal expressions", just "literals" should be fine. The wording her could be "Extract constant refactoring is allowed only on literals". If you do that, then there is no need to specify the types of the literal expressions.

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
+This constant will be instantiated below the class declaration. If the expression is contained within an innerclass, the constant will be instantiated below the innerclass declaration. The new constant will always be defined with modifiers private, static, and final.
```
praksb: "instantiated" is not the right word to use here. I would change it to something like "The new constant will be declared as a field in the containing class from where the refactoring was invoked". Alternately, you could just have an example here similar to what Jetbrains has: https://www.jetbrains.com/help/idea/extract-constant.html

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
+![GIF showing extract constant](../../images/extract-constant.gif)
+
+## Refactor: Extract Local Variable
+
+You can extract valid expression types into local variables. Please note that not all expression types are currently supported.
+
+This new variable will be instantiated above the currently selected line. It will retain the formatting of the expression that is selected.
+
+At this time, there is no conflict detection for the new variable and nested expressions are not supported.
```
praksb: You could remove text on "conflict detection" from here as well. I think you are referring to NestedExpression used in the compiler code here. It is confusing to call it out here since nested expression could also mean expression instead another expression which actually is supported. I would just remove this text as well and we could fix the NestedExpression case as a bug fix later.

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
+At this time, there is no conflict detection for the new constant.
+
+![GIF showing extract constant](../../images/extract-constant.gif)
+
+## Refactor: Extract Local Variable
+
+You can extract valid expression types into local variables. Please note that not all expression types are currently supported.
+
+This new variable will be instantiated above the currently selected line. It will retain the formatting of the expression that is selected.
```
praksb: "instantiated" is not the right word. "declared" or "defined" would be better.

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
+This constant will be instantiated below the class declaration. If the expression is contained within an innerclass, the constant will be instantiated below the innerclass declaration. The new constant will always be defined with modifiers private, static, and final.
+
+At this time, there is no conflict detection for the new constant.
+
+![GIF showing extract constant](../../images/extract-constant.gif)
+
+## Refactor: Extract Local Variable
+
+You can extract valid expression types into local variables. Please note that not all expression types are currently supported.
```
praksb: "expression types" is an implementation detail just "expression" should be fine. Also, it is not clear what "valid" means we could just remove it as well. We don't have to specify that we don't support some expression types since they are just bug fixes which will be handled later.

In general, this is not a specification so there is no need to explicitly go into a lot of details about these features. Our documentation should just be brief enough to describe the feature and how to use it. For ideas you could refer to:
https://www.jetbrains.com/help/idea/extract-variable.html

docs/articles/apex/refactoring.md
```diff
@@ -15,3 +15,25 @@ Validation of the new symbol name is performed before applying the rename. If th
...
+
+This new variable will be instantiated above the currently selected line. It will retain the formatting of the expression that is selected.
+
+At this time, there is no conflict detection for the new variable and nested expressions are not supported.
+
+![GIF showing extract local variable basics](../../images/extract-local-variable-basics.gif)
```
praksb: You could remove this gif. The other one you have for local variables is most interesting.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
sfsholden: Sorry, I thought this was a fresh branch off of develop. This was my mistake.

docs/articles/apex/refactoring.md
sfsholden: I like this. I will follow the example from Jetbrains. It will also mean that we are increasing the number of users that get value from the instructions (ease of access for people that cannot watch the video).

PR #1398, opened 2019-06-13
Addressed workspaceOrgType context issues

AnanyaJha:
### What does this PR do?
This PR adds telemetry to capture information on workspaceOrgType context switching failing due to a username being present instead of an alias. It also increases the retry period to address auth info file latency issues to 6 seconds.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
+++ b/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
@@ -42,7 +42,7 @@ export async function getWorkspaceOrgType(
 
   if (isNullOrUndefined(username)) {
     telemetryService.sendError(
-      `workspaceOrgType.getWorkspaceOrgType ran into an undefined username after ${counter} retries and scratch org username is ${isUsername}`
+      `workspaceOrgType.getWorkspaceOrgType ran into an undefined username after ${counter} retries and default username provided is ${isUsername}`
     );
   } else {
     telemetryService.sendEventData(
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -27,17 +27,22 @@ export async function getWorkspaceOrgType(
     throw e;
   }
 
+  const email = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(
```
lcampos: This regex fails if you have consecutive `-` characters or a combination of `.-`. e.g. test2342--sd@test.co

AnanyaJha: POSIX expressions are not supported by javascript out of the box- you can download an npm package called `posix-character-classes` to use them. For now, I changed the regex to accept multiple special characters for the local part of the email.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -42,7 +42,7 @@ export async function getWorkspaceOrgType(
 
   if (isNullOrUndefined(username)) {
     telemetryService.sendError(
-      `workspaceOrgType.getWorkspaceOrgType ran into an undefined username after ${counter} retries and scratch org username is ${isUsername}`
+      `workspaceOrgType.getWorkspaceOrgType ran into an undefined username after ${counter} retries and default username provided is ${isUsername}`
```
AnanyaJha: Changed the message since the username may not be for a scratch org.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
lcampos: Thanks for checking, I think that just changing the regex for now should be enough

PR #1407, opened 2019-06-18
Introduce webpack in build process

lcampos:
### What does this PR do?
Introduces the use of webpack (exclusively in the `salesforcedx-utils-vscode` module) as part of the extensions build process. We'll now be able to tree-shake, minify and bundle our code when we create the extension vsix files.

The PR also updates the extensions' `.vscodeignore` files to remove all the unnecessary code from the vsix. As a result of this work, the vsix file size is reduced in at least 2.3 MB per extension that uses the `salesforcedx-utils-vscode` module.

| Ext. Name            | Prev Size     | New Size  |
| ---------------------|:-------------:| ---------:|
| Core                 | 14.9 MB       | 12.2 MB   |
| Apex                 | 25.1 MB       | 22.4 MB   |
| Apex Replay Debugger | 3.8 MB        | 724 KB    |
| Apex Debugger        | 6.3 MB        | 3.3 MB    |
| Lightning            | 26 MB         | 23.3 MB   |
| LWC                  | 23.5 MB       | 20.9 MB   |
| Visualforce          | 9.3 MB        | 9 MB      |
 
@W-5846230@


.circleci/config.yml
```diff
@@ -15,7 +15,11 @@ jobs:
       - restore_cache:
           key: dependency-cache-{{ checksum "package.json" }}
       - run: npm install
-      - run: npm run compile
+      # Setup an NODE_ENV to production mode before compiling.
```
lcampos: Update circleci builds so that the vsix files are compiled in production mode. This means we run webpack in production mode.

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -43,7 +43,7 @@
     "vscode": "^1.26.0"
   },
   "scripts": {
-    "compile": "tsc -p ./",
+    "compile": "webpack",
```
lcampos: Compilation of this module is now done with Webpack and not directly with the Typescript compiler.

packages/salesforcedx-utils-vscode/test/unit/i18n/localizationTest.ts
```diff
@@ -6,14 +6,9 @@
...
 import { expect } from 'chai';
-import {
-  BASE_FILE_EXTENSION,
-  BASE_FILE_NAME,
-  Config,
-  DEFAULT_LOCALE,
-  Localization,
-  Message
-} from '../../../src/i18n/localization';
+import { Config, Localization, Message } from '../../../src/i18n';
```
lcampos: Updated the test to remove the dynamic `requires` since that breaks webpack's dependency navigation.

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -0,0 +1,84 @@
+const path = require('path');
+const glob = require('glob');
+const DIST = path.resolve(__dirname);
+
+const getEntryObject = () => {
```
lcampos: Function that builds the entry object used by webpack in order to build the bundles. It adds the test file paths if you run the configuration in `development` mode.

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+  // extensions run in a node context
+  target: 'node',
+  mode: getMode(),
+  entry: getEntryObject(),
+  // vsix packaging depends on commonjs2
+  output: {
+    path: DIST,
+    filename: '[name].js',
+    libraryTarget: 'commonjs2',
+    devtoolModuleFilenameTemplate: `webpack:///salesforcedx-utils-vscode/[resource-path]`
```
lcampos: Format used in the source maps generated by every bundle (output from `getEntryObject`). We then use this format in the `launch.json` file to be able to map to the bundled code and still be able to set breakpoints in VSCode.

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+  // pre-process certain file types using loaders
+  module: {
+    rules: [
+      {
+        test: /\.tsx?$/,
+        exclude: /node_modules|\.d\.ts$/,
+        use: [
+          {
+            loader: 'ts-loader'
```
lcampos: We'll now use `ts-loader` instead of directly calling the Typescript compiler.

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+  output: {
+    path: DIST,
+    filename: '[name].js',
+    libraryTarget: 'commonjs2',
+    devtoolModuleFilenameTemplate: `webpack:///salesforcedx-utils-vscode/[resource-path]`
+  },
+  // include source maps
+  devtool: 'source-map',
+  // excluding dependencies from getting bundled
+  externals: {
```
lcampos: Everything inside of externals will not be included in the output bundles. Most of the time we'll ignore vscode modules e.g. vscode-nls

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+module.exports = {
+  // extensions run in a node context
+  target: 'node',
+  mode: getMode(),
+  entry: getEntryObject(),
+  // vsix packaging depends on commonjs2
+  output: {
+    path: DIST,
+    filename: '[name].js',
+    libraryTarget: 'commonjs2',
```
lcampos: `commonjs2` is the format in which vscode expects the generated JS to be in. If we forget to add this, vscode won't be able to run our generated JS.

packages/salesforcedx-vscode-apex-debugger/.vscodeignore
```diff
@@ -6,3 +6,22 @@ src/**
 **/*.map
 .gitignore
 tsconfig.json
+
+# salesforcedx-apex-debugger files to ignore
```
lcampos: We're only using what's in `*/out/src/**` so I'm removing everything else from the vsix.

packages/salesforcedx-vscode-apex-debugger/.vscodeignore
```diff
@@ -6,3 +6,22 @@ src/**
...
+# salesforcedx-apex-debugger files to ignore 
+node_modules/@salesforce/salesforcedx-apex-debugger/docs/**
+node_modules/@salesforce/salesforcedx-apex-debugger/src/**
+node_modules/@salesforce/salesforcedx-apex-debugger/test/**
+node_modules/@salesforce/salesforcedx-apex-debugger/out/test/**
+node_modules/@salesforce/salesforcedx-apex-debugger/package-lock.json
+node_modules/@salesforce/salesforcedx-apex-debugger/tsconfig.json
+
+# salesforcedx-utils-vscode files to ignore 
+node_modules/@salesforce/salesforcedx-apex-debugger/node_modules/@salesforce/salesforcedx-utils-vscode/node_modules/**
```
lcampos: Now that we create bundles for `salesforcefx-utils-vscode` we can remove its `node_modules`.

scripts/publish.js
```diff
@@ -90,7 +90,7 @@ shell.exec('git clean -xfd -e node_modules');
 shell.exec('npm install');
 
 // Compile
-shell.exec('npm run compile');
+shell.exec('NODE_ENV=production npm run compile');
```
lcampos: This allows us to run webpack in production mode when publishing the extensions.

packages/salesforcedx-vscode-core/src/commands/forceAuthWebLogin.ts
```diff
@@ -11,15 +11,15 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
-import { CliCommandExecutor } from '@salesforce/salesforcedx-utils-vscode/out/src/cli/commandExecutor';
```
lcampos: Updating code references so that the leverage the `index.ts` file per module.

packages/salesforcedx-vscode-apex-debugger/.vscodeignore
brpowell: Is it possible to ignore everything in the package and then whitelist `*/out/src/**` to cut down on the number of lines? Might be nice as we add more modules to the webpack config.

```
# ignore everything but out/src
node_modules/@salesforce/salesforcedx-apex-debugger/**
!node_modules/@salesforce/salesforcedx-apex-debugger/out/src/**
```

lcampos: Yes, this is only temporary and once we have webpack on every package we'll be able to ignore everything under the node_modules, test, etc. since everything will be bundled under the package's `out/src/**`. I did these updates so that we don't have to wait for every package to run via webpack in order to ship only the code needed to run.

PR #1413, opened 2019-06-19
Fix flaky tests in appveyor builds

lcampos:
### What does this PR do?
Updates the custom object metadata used in tests which were causing cli deploy issues. Updates the spectron util which reads command execution results from the Output panel. Updates the soql system tests to run independently and adds a check to the SFDX commands being loaded in the command palette, making the test results more reliable.

@W-6259681@


packages/system-tests/src/areas/common.ts
```diff
@@ -71,7 +71,7 @@ export class CommonActions {
   }
 
   public async getConsoleOutput(): Promise<any> {
-    const htmlTag = `div[class="view-lines"]`;
+    const htmlTag = `div[id="workbench.panel.output"] div[class="view-lines"]`;
```
lcampos: VSCode has 2 different areas which use `view-lines` as a class; the monaco editor & the terminal output panel. This change makes sure we always look for the content in the terminal's output panel.

packages/system-tests/scenarios/forceSoqlQueryUi.test.ts
```diff
@@ -27,17 +27,22 @@ describe(TITLE, () => {
 
   let username: string;
 
-  before(async () => {
+  beforeEach(async () => {
```
lcampos: Updating from `before` to `beforeEach` and `after` to `afterEach` causes the tests to run in independent vscode sessions instead of sharing the same vscode session. This is important because the tests themselves were only asserting the presence of `exit code 0` in the output panel. If the tests share the same vscode session, as long as one of the tests is successful then the rest would pass even if their command execution fails.

PR #1422, opened 2019-06-21
Update launch configurations to fix issues with debugging tests

lcampos:
### What does this PR do?
Updates the project's launch configurations, fixing the issues with running tests from vscode and attaching the debugger to the test execution.

Screen Shot 2019-06-20 at 5 34 51 PM

@W-6268305@

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -330,35 +330,6 @@
       "internalConsoleOptions": "openOnSessionStart"
     },
     {
-      "name": "Launch SObject Generator Tests",
-      "type": "node",
-      "request": "launch",
-      "cwd": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator",
-      "program": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/node_modules/mocha/bin/_mocha",
-      "args": [
-        "-u",
-        "tdd",
-        "--timeout",
-        "100000",
-        "--colors",
-        "--recursive",
-        "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/out/test/"
-      ],
-      "sourceMaps": true,
-      "smartStep": true,
-      "outFiles": [
-        "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/out/test/**/*.js"
-      ],
-      "sourceMapPathOverrides": {
-        "../../../test/unit/": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/out/test/unit/*",
-        "../../../test/integration/": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/out/test/integration/*",
-        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
-        "webpack:///*": "*"
-      },
-      "preLaunchTask": "Compile",
-      "internalConsoleOptions": "openOnSessionStart"
-    },
-    {
       "name": "Launch SObject Generator Unit Tests",
       "type": "node",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -189,24 +239,16 @@
...
       "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
+      "sourceMapPathOverrides": {
+        "../../../test/unit/": "${workspaceFolder}/packages/salesforcedx-visualforce-markup-language-server/out/test/unit/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///*": "*"
+      },
       "preLaunchTask": "Compile",
       "internalConsoleOptions": "openOnSessionStart"
     },
-    {
```
lcampos: Removing since this package was deleted a few months ago.

.vscode/launch.json
```diff
@@ -264,8 +318,16 @@
...
+        "../../../test/vscode-integration/": "${workspaceFolder}/packages/salesforcedx-vscode-apex-debugger/out/test/vscode-integration/*",
+        "../../../../test/vscode-integration/": "${workspaceFolder}/packages/salesforcedx-vscode-apex-debugger/out/test/vscode-integration/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///*": "*"
+      },
+      "preLaunchTask": "Compile",
+      "internalConsoleOptions": "openOnSessionStart"
     },
     {
       "name": "Launch SObject Generator Tests",
```
lcampos: The sobjects-faux-generator package has 3 different configs; 1 that runs both integration & unit tests, 1 for integration tests and another for unit tests. Do we want to consolidate the configs into only one that runs both integration & unit tests or have a dedicated launch config for every test category ?

.vscode/launch.json
```diff
@@ -97,8 +102,15 @@
...
+      "smartStep": true,
       "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
-      "preLaunchTask": "Compile"
+      "sourceMapPathOverrides": {
+        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-core/out/test/vscode-integration/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///*": "*"
+      },
+      "preLaunchTask": "Compile",
+      "internalConsoleOptions": "openOnSessionStart"
```
lcampos: Adding `intergnalConsoleOptions` to the missing configs allows VSCode to open the terminal and focus in the correct panel (Debug).

AnanyaJha: Is this related to the appveyor test failures we were seeing earlier where we couldn't locate the console output?

.vscode/launch.json
AnanyaJha: I think it would be helpful to keep the test configs separate for each category in case we don't need to run both integration and unit tests at the same time.

.vscode/launch.json
lcampos: No, the test failures were caused by the tests executing before the extensions were active. I've fixed that by changing the way we run those tests. This setting will focus in the Debug panel of the terminal once you run the launch config.

.vscode/launch.json
lcampos: Cool, I'll remove the option that runs both test categories at once so we have the same configs across modules.

praksb: I agree. We should have one for unit and one for integration and that's it.

PR #1428, opened 2019-06-24
[Org Browser] Hook up metadata to tree view

AnanyaJha:
### What does this PR do?
This PR retrieves the metadata and displays it in the org browser. Adds refresh functionality at the top level. Org browser will automatically refresh with data when the default org has switched. Tests have not been added yet. The org browser is hidden by default; to review, change 'display_tree_view' to 'true'.

@W-5767978@


packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
+      this.defaultOrg = username;
+      return diff;
+    } else {
+      throw new Error('Workspace could not be found.');
     }
   }
 }
+
+export function parseErrors(error: string): string {
+  const e = JSON.parse(error);
```
AnanyaJha: Do we want to change the error message based on the type of error?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -138,31 +81,27 @@ export type MetadataObject = {
...
-}
-
-export async function onUsernameChange() {
-  const metadataTypesPath = await getTypesPath();
-  if (
-    !isNullOrUndefined(metadataTypesPath) &&
-    !fs.existsSync(metadataTypesPath)
-  ) {
-    await forceDescribeMetadata(metadataTypesPath);
+  const jsonObject = JSON.parse(metadataFile);
```
AnanyaJha: Wasn't sure if I should wrap the JSON parsing here and for the components in a try/catch

packages/salesforcedx-vscode-core/package.json
```diff
@@ -625,6 +632,14 @@
       {
         "command": "sfdx.internal.lightning.lwc.create",
         "title": "%force_lightning_lwc_create_text%"
+      },
+      {
+        "command": "sfdx.force.metadata.view.refresh",
+        "title": "Refresh tests",
```
lcampos: This needs to be a label and the text should be something along the lines of `Refresh Org Browser`. cc @hbc-bindu

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -490,8 +496,8 @@ export async function activate(context: vscode.ExtensionContext) {
   const orgList = new OrgList();
   orgList.displayDefaultUsername(defaultUsernameorAlias);
   context.subscriptions.push(registerOrgPickerCommands(orgList));
-  // await setupOrgBrowser(context, defaultUsernameorAlias);
 
+  await setupOrgBrowser(context, defaultUsernameorAlias);
```
lcampos: I think we should keep this commented out since we're always passing `false` for the tree view context in the following line.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
+    }
+    const nodeList = [];
+    for (const component of componentsList) {
+      const cmpNode = new BrowserNode(component, NodeType.MetadataCmp);
+      nodeList.push(cmpNode);
+    }
+    return nodeList;
+  }
+
+  public async getDefaultUsernameOrAlias(): Promise<boolean> {
```
lcampos: The name and responsibility of this method is a bit confusing. We use it to determine if the user has switched to a different org and also to set the username for the rest of the org browser code to leverage. Can we break this into different methods and name them based on what they are responsible for ?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
-  public async refresh(defaultOrg?: string): Promise<void> {
-    if (!isNullOrUndefined(defaultOrg)) {
-      this.defaultOrg = defaultOrg;
+  public async onViewChange() {
+    if (await this.getDefaultUsernameOrAlias()) {
+      this.internalOnDidChangeTreeData.fire();
     }
+  }
+
+  public async refresh(defaultOrg?: string): Promise<void> {
```
lcampos: Can we visually indicate to the user that a refresh of the org browser data is happening ? Right now, if a user clicks the refresh button there's nothing that lets the user know if the feature is actually refreshing or when it's done refreshing the data.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -138,31 +81,27 @@ export type MetadataObject = {
...
-export function buildTypesList(metadataTypesPath: string): string[] {
-  try {
-    const fileData = JSON.parse(fs.readFileSync(metadataTypesPath, 'utf8'));
-    const metadataObjects = fileData.metadataObjects as MetadataObject[];
-    const metadataTypes = [];
-    for (const metadataObject of metadataObjects) {
-      if (!isNullOrUndefined(metadataObject.xmlName)) {
-        metadataTypes.push(metadataObject.xmlName);
-      }
+export function buildTypesList(
```
lcampos: This method should return a string array in alphabetical order. Currently the types are shown in an order that's difficult to navigate.
Screen Shot 2019-06-24 at 7 12 05 PM

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
lcampos: I think we should expand the try/catch to cover the rest of the code since it's also navigating a json structure that might throw a null/empty value.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -138,31 +81,27 @@ export type MetadataObject = {
...
+  metadataFile?: any,
+  metadataTypesPath?: string
+): string[] {
+  if (isNullOrUndefined(metadataFile)) {
+    try {
+      metadataFile = fs.readFileSync(metadataTypesPath!, 'utf8');
+    } catch (e) {
+      throw e;
```
lcampos: Need to log the exception as well.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -8,124 +8,67 @@
 import {
   CliCommandExecutor,
   Command,
+  CommandOutput,
```
lcampos: The code in this file both runs a describe command and hosts utility methods (e.g. getTypesFolder) that are unrelated to the describe command. Can we create another file specific for the utility methods used across the org browser code ?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
lcampos: Yes, what errors are not already covered here ?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
     }
   }
 }
+
+export function parseErrors(error: string): string {
+  const e = JSON.parse(error);
+  let message: string;
+  switch (e.name) {
+    case 'RefreshTokenAuthError':
+      message = 'Error refreshing authentication token.';
```
lcampos: Let's update this messages to use labels. cc @hbc-bindu

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
+    await this.getDefaultUsernameOrAlias();
     this.internalOnDidChangeTreeData.fire();
   }
 
   public getTreeItem(element: BrowserNode): vscode.TreeItem {
     return element;
   }
 
-  public getChildren(element?: BrowserNode): Promise<BrowserNode[]> {
+  public async getChildren(element?: BrowserNode): Promise<BrowserNode[]> {
```
lcampos: Let's rework this method to make it easier to read.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: This was all of them! Was just checking if we wanted these.

packages/salesforcedx-vscode-core/src/index.ts
AnanyaJha: Ah yes, was meant to be commented still. Thanks

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: Sounds good, let me know if these messages look okay @hbc-bindu and I'll add them to the i8n file

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
AnanyaJha: Yeah I can do that- I'll do the same for the metadataCmp file because it also contains utility code and the list metadata command.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: Hmm yeah that makes sense. I'll look into how to indicate the progress.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -51,115 +36,88 @@ export class ForceListMetadataExecutor extends SfdxCommandletExecutor<string> {
...
-    const fileData = fs.readFileSync(componentsPath, 'utf8');
-    if (fileData !== 'undefined') {
-      const fileObject = JSON.parse(fileData);
-      const cmpList = fileObject instanceof Array ? fileObject : [fileObject];
-      for (const component of cmpList) {
-        if (!isNullOrUndefined(component.fullName)) {
-          metaComponents.push(component.fullName);
-        }
+  const components = [];
+  if (isNullOrUndefined(componentsFile)) {
```
brpowell: `util.isNullOrUndefined` has been deprecated for awhile, we should avoid using it going forward.
https://nodejs.org/api/util.html#util_util_isnullorundefined_object

lcampos: Good catch

packages/salesforcedx-vscode-core/package.json
hbc-bindu: Label looks good to me.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
hbc-bindu: Messages are good; no changes.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
+      nodeList.push(typeNode);
+    }
+    return nodeList;
+  }
+
+  public async getComponents(
+    metadataType: BrowserNode
+  ): Promise<BrowserNode[]> {
+    const username = this.defaultOrg!;
+    const componentsPath = await getComponentsPath(
```
brpowell: I'm wondering if metadataCmp.ts can be responsible for doing more of this work. The provider is really only concerned with having a list of the component names to construct the nodes and shouldn't care whether it comes from the cache or remote.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
brpowell: Whether there's an org switch or refresh, you could clear the tree and present a non-collapsable "loading" node.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
brpowell: Icons?
Screen Shot 2019-06-25 at 2 51 41 PM

No Icons?
Screen Shot 2019-06-25 at 2 52 30 PM

@ntotten

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -24,33 +37,144 @@ export class MetadataOutlineProvider
...
+      }
+    } else if (element.type === NodeType.Org) {
+      const metadataTypes = await this.getTypes();
+      element.children = metadataTypes;
+      return Promise.resolve(element.children);
+    } else if (element.type === NodeType.MetadataType) {
+      const metadataCmps = await this.getComponents(element);
+      if (isNullOrUndefined(metadataCmps) || metadataCmps.length === 0) {
+        const emptyDefault = new BrowserNode(
+          'No components available',
```
brpowell: Do we want to show a node if there aren't any components? In the normal file tree if there aren't any files in a folder nothing is shown.

AnanyaJha:
Still working on visually indicating a refresh is in progress and the `display tree view` context is set to true for now.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export function isEmpty(object: any): object is null | undefined {
```
lcampos: 👍

packages/salesforcedx-vscode-core/src/commands/forceListMetadata.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.metadataType = metadataType;
+    this.defaultUsernameOrAlias = defaultUsernameOrAlias;
+  }
+
+  public build(data: {}): Command {
+    let builder = new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_list_metadata_text'))
+      .withArg('force:mdapi:listmetadata')
+      .withFlag('-m', this.metadataType)
+      .withFlag('-u', this.defaultUsernameOrAlias)
```
lcampos: Let's add a `.withLogName` entry to this command.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: I'm leaning towards displaying a node with a `No components available` label because I found it difficult to easily see that the type node was expanded and no components were below it once the data had been requested.

cc @ntotten Thoughts?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -20,37 +21,107 @@ export class MetadataOutlineProvider
...
+        break;
+      case NodeType.MetadataType:
+        element.children = await this.getComponents(element);
+        break;
+    }
+    return Promise.resolve(element.children);
   }
 
-  public async getDefaultUsernameOrAlias() {
+  public async getTypes(): Promise<BrowserNode[]> {
```
brpowell: getTypes(), getComponents(), and getDefaultUsernameOrAlias() don't need to be public.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -20,37 +21,107 @@ export class MetadataOutlineProvider
...
-      this.defaultOrg = await OrgAuthInfo.getDefaultUsernameOrAlias(false);
+      const username = await OrgAuthInfo.getDefaultUsernameOrAlias(false);
+      return username;
+    } else {
+      throw new Error(nls.localize('cannot_determine_workspace'));
     }
   }
 }
+
+export function parseErrors(error: string): string {
```
brpowell: Does this need to be exported?

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
brpowell: This name is a misnomer. We're not testing if an object is empty, but whether it is null or undefined. An empty object would be `{}`, and this function would return false for such a value. null and undefined are falsy values so it's fine to just test this using bang.
```
> !undefined
true
> !null
true
> !{}
false
```

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
ntotten: I don't have a strong opinion on this. The file tree in VS Code allows you to click expand on the node and just doesn't show anything when the folder is empty. In this case though since we are loading data from the org I almost wonder if not showing anything will make it feel like an err. I'm okay with the current design.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
ntotten: I personally like the icons. Once we expand to "source format" i think it would be good to use the same icons as the file explorer for xml, js, html, cls, etc.

ntotten: One thought, in the file explorer there is an option to show icons or not. Maybe we can use that same setting?

image

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: I'll leave the message in for now and we can adjust if we get feedback on it. Thanks!

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: A few of the methods need to be public so I can access them for testing.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
AnanyaJha: I think it's useful to keep this function since we have been using 'isNullOrUndefined' across the project. I'll adjust the name to be more representative of what the function is doing

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
AnanyaJha: The icons look good to me too, I think we can add them in a future story. I'll log a work item for the icons with the option of exploring that setting ^^

packages/salesforcedx-vscode-core/src/commands/forceListMetadata.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.metadataType = metadataType;
+    this.defaultUsernameOrAlias = defaultUsernameOrAlias;
+  }
+
+  public build(data: {}): Command {
+    let builder = new SfdxCommandBuilder()
+      .withArg('force:mdapi:listmetadata')
+      .withFlag('-m', this.metadataType)
+      .withFlag('-u', this.defaultUsernameOrAlias)
+      .withLogName(nls.localize('force_list_metadata_text'))
```
lcampos: This label will break the current metric name format we use, let's replace it for using something like `force_list_metadata`.

AnanyaJha: Sounds good, will do!

PR #1434, opened 2019-06-26
Update changelog for release/v46.3.0

praksb:
### What does this PR do?
Update changelog for release/v46.3.0

Updates the changelog for the next release

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### salesforcedx-vscode-apex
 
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 46.3.0 - June 27, 2019
+
+### Fixed
+
+- Update launch configurations for running extension tests ([PR #1422](https://github.com/forcedotcom/salesforcedx-vscode/pull/1422))
```
praksb: This change is not part of any package so i don't have a package name to include for this change

lcampos: We don't need to include that change since it's exclusive to how we develop and has no customer impact.

praksb: @lcampos removed that change

PR #1448, opened 2019-07-03
Jhork/org open in container mode

jonnyhork:
### What does this PR do?
This PR will add logic use the native VS Code API to open the browser on the user's host machine when they try to open their default org form the command palette.
@W-6218734@


packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,64 @@
...
+
+  public execute(response: ContinueResponse<string>): void {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: getRootWorkspacePath()
+    }).execute(cancellationToken);
+
+    if (process.env.SFDX_CONTAINER_MODE) {
```
vazexqi: I wonder if we should use this (which we set by ourselves) or if we should use a more general one from VS Code itself to detect if it is running in container mode.

What's the status of the general one from VS Code?

jonnyhork:
I am not sure what the usual pattern for error handling is in the extensions. I used `telemetryService.sendErrorEvent(e.message, e.stack);` based on other examples.
What is the expected behavior we want when
1. There is an error in the CLI, ie no URL is present or JSON is not returned
2. There is an error with VS code opening the browser

Currently, I think error handling currently taken care by the CLI, as long as we display the output.

Also, I don't think there are any test for `forceOrgOpen.ts` is that the case? Do we want a test to at lest make sure the `vscode.env.openExternal` is called when in container mode?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
jonnyhork: oh, forgot there is a general one from VS code, I'll have to look into that.
Is there an advantage to using our own here when we start running the Web IDE, will can we set the same env var?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,64 @@
...
   public build(data: {}): Command {
-    return new SfdxCommandBuilder()
+    const builder = new SfdxCommandBuilder()
       .withDescription(nls.localize('force_org_open_default_scratch_org_text'))
       .withArg('force:org:open')
-      .withLogName('force_org_open_default_scratch_org')
-      .build();
+      .withLogName('force_org_open_default_scratch_org');
+    if (process.env.SFDX_CONTAINER_MODE) {
+      builder.withArg('-r').withJson();
```
jonnyhork: `-r` URL only and `withJson` to parse the stdout form the CLI

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,64 @@
...
+    }).execute(cancellationToken);
+
+    if (process.env.SFDX_CONTAINER_MODE) {
+      execution.stdoutSubject.subscribe(cliResponseJSON => {
+        try {
+          const cliOrgData: CLIOrgData = JSON.parse(cliResponseJSON.toString());
+          const authenticatedOrgUrl: string = cliOrgData.result.url;
+          if (authenticatedOrgUrl) {
+            vscode.env.openExternal(vscode.Uri.parse(authenticatedOrgUrl));
+          }
```
jonnyhork: I wasn't sure what we should do if there is no `authenticatedOrgUrl`

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
dehru: We decided against using the vscode one so that it will work in the web-ide container.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Let's use the long URLs. Not the short one.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Correct me if I am wrong, we decided on using a ENV VARIABLE so that the CLI can detect it and use the device flow for that.

But, I don't think it's mutually exclusive right? When we are in VS Code, we could use its own detection mechanism?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
jonnyhork: I don't think the URL is shortened, at least it looks quite verbose. The `-r` flag just prevents the CLI from opening the browser and only returns the org info in the output. I'm not aware of any differences in the URL it returns.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
jonnyhork: What is advantage to using VS Codes detection system over our own ENV var? The assumption is its more reliable or that if a user didn't start with our supported image then the container adaptations would be broken again?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Sorry, mistyped. I meant use `--url` instead of `-r`.

jonnyhork: ok that makes more sense

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,66 @@
...
+        } catch (e) {
+          telemetryService.sendErrorEvent(e.message, e.stack);
+        }
+      });
+    }
+    this.attachExecution(execution, cancellationTokenSource, cancellationToken);
   }
```
jonnyhork: I needed to call attachExecution so that these methods are called to show the CLI output. However, because of the `--json` flag added to the open command, the output is in json format and not quite as UI friendly as when the output is wrapped in a message.
`channelService.streamCommandOutput(execution);`

   `if (this.showChannelOutput) {
    }`

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,66 @@
...
+          const authenticatedOrgUrl: string = cliOrgData.result.url;
+          if (authenticatedOrgUrl) {
+            vscode.env.openExternal(vscode.Uri.parse(authenticatedOrgUrl));
+          }
+        } catch (e) {
+          telemetryService.sendErrorEvent(e.message, e.stack);
+        }
+      });
+    }
+    this.attachExecution(execution, cancellationTokenSource, cancellationToken);
```
jonnyhork: I needed to call attachExecution so that these methods are called to show the CLI output. However, because of the `--json` flag added to the open command, the output is in json format and not quite as UI friendly as when the output is wrapped in a message.
`channelService.streamCommandOutput(execution);`

`if (this.showChannelOutput) { channelService.showChannelOutput(); }`

vazexqi:
If we bump the VS Code version dependency to 1.36 we can also use https://code.visualstudio.com/updates/v1_36#_apis-for-remote

jonnyhork:
### **Summary of Pending Issues**bitmoji
**1. What do we want to use to determine if we are running in a container?**

- Custom env variable `SFDX_CONTAINER_MODE`, this is probably the simplest option and is easily applicable when the Web IDE is build from dockerfile.
- Official VS code API for remote development https://code.visualstudio.com/updates/v1_36#_apis-for-remote 
It's not clear to me what value I would be looking for based on the release notes above so I would have to debug this a bit. Either `vscode.env.remoteName` or `!ExtensionKind.UI` perhaps.

**2. How do we want to handle messaging to the user in the Output Channel when in a container?**
- Currently the raw `json` response is displayed from CLI 
- I am working on parsing the STDOUT from the CLI and formatting it before being passed to the Output Channel, this would be a specific adaptation for `force:org:open`

**3. What is the plan for testing?**
- As far as I can tell there are no existing tests for `forceOrgOpen.ts`, what does the PDT do in this case?
- Do we want to at least make sure if we are in container mode then `vscode.env.openExternal()` is called? or do we want more extensive testing?
- I'm not sure how much effort is needed mock the appropriate things to test a command like this in VS Code

**4. Error handling, do we need more?**
- Currently most of the error handling is taken care of in the CLI, we would just display the error it produces in the Output Channel.
- I also have `telemetryService.sendErrorEvent(e.message, e.stack);` in a catch block if there is errors with `json` parsing or VS code opening the browser.

CC: @vazexqi @dehru @lcampos

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: getRootWorkspacePath()
+    }).execute(cancellationToken);
+
+    if (process.env.SFDX_CONTAINER_MODE) {
+      channelService.showChannelOutput();
+      channelService.streamCommandStartStop(execution);
```
jonnyhork: This line will handle displaying the start and stop time for command execution and display the exit code.
`11:06:15.720 sfdx force:org:open --urlonly --json --loglevel fatal`
"User message"
`11:06:18.396 sfdx force:org:open --urlonly --json --loglevel fatal ended with exit code 0`

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
+
+      execution.stdoutSubject.subscribe(cliResponseJSON =>
+        this.openBrowserWithCliResponse(cliResponseJSON)
```
jonnyhork: I extracted the logic here to keep the code in the execution and subscription cleaner.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
+  }
+
+  private buildUserMessageWith(orgData: CLIOrgData): string {
+    return nls.localize(
+      'force_org_open_container_mode_message_text',
+      orgData.result.orgId,
+      orgData.result.username,
+      orgData.result.url
+    );
+  }
```
jonnyhork: This method will produce a formatted message based on the JSON response from the CLI to match the message displayed  to the user when they run the command without the `--json` flag.

```
Access org 00D1U000000qfglUAA as user jhork@dev.com with the following URL: https://na85.salesforce.com/secur/frontdoor.jsp?sid=00D1U000000qfgl!AQ4AQAD_1.NwP.soL74jcx5MlM0nXLeBiCxJmn626qsyLXXJK6byVDKseNB8aLSx.VKSHQaiwf5XjUd7p7e1f6hmEjaL5Q.Z
```

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
+import { telemetryService } from '../telemetry';
+import { getRootWorkspacePath } from '../util';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
+interface CLIOrgData {
```
vazexqi: Use a type alias instead of an interface since we aren't really extending this. See https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
 class ForceOrgOpenExecutor extends SfdxCommandletExecutor<{}> {
   public build(data: {}): Command {
-    return new SfdxCommandBuilder()
+    const builder = new SfdxCommandBuilder()
       .withDescription(nls.localize('force_org_open_default_scratch_org_text'))
       .withArg('force:org:open')
-      .withLogName('force_org_open_default_scratch_org')
-      .build();
+      .withLogName('force_org_open_default_scratch_org');
+    if (process.env.SFDX_CONTAINER_MODE) {
```
vazexqi: [Must] Extract this check into its own method in some util class. We use this check at least twice and I can see Dehru using this later as well. By putting it in a single place, you follow the principles of DRY and we can just change it in one place if we need to change how we use it (between `process.env.SFDX_CONTAINER_MODE and/or the new VS Code API).

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: What happens if a user is not using our official docker image and doesn't set the ENV var?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,24 +6,90 @@
...
+
+      if (authenticatedOrgUrl) {
+        vscode.env.openExternal(vscode.Uri.parse(authenticatedOrgUrl));
+      }
+    } catch (e) {
+      telemetryService.sendErrorEvent(e.message, e.stack);
+    }
+  }
+
+  public execute(response: ContinueResponse<string>): void {
```
vazexqi: I looked at the parent execute and it seems that you are breaking the contract? With the way you are doing it, don't you lose all metrics with start time?

vazexqi:
1. What do we want to use to determine if we are running in a container?

I've made a concrete suggestion to pull it into a util class first and use ENV for now.

2. How do we want to handle messaging to the user in the Output Channel when in a container?

Your propose idea is fine.

3. What is the plan for testing?

@lcampos - Could you help here?

4. Error handling, do we need more?

What you have is fine.

[Big change requested]: The current approach you have breaks the existing contract. You are not using the notificationService and not sending back telemetry. Is there a reason you are not doing those things?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
jonnyhork: Then the command would proceed as if its not in a container and try to open a browser, which may not be available in the container.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+      const cliOrgData: CLIOrgData = JSON.parse(cliResponseJSON.toString());
+      const authenticatedOrgUrl: string = cliOrgData.result.url;
+
+      channelService.appendLine(this.buildUserMessageWith(cliOrgData));
+
+      if (authenticatedOrgUrl) {
+        vscode.env.openExternal(vscode.Uri.parse(authenticatedOrgUrl));
+      }
+    } catch (e) {
+      telemetryService.sendErrorEvent(e.message, e.stack);
```
vazexqi: I think it would be worthwhile also presenting so error to the user if this fails to open. That way the user knows that something has failed and can report to use the error message.

Maybe something like

```
channelService.appendLine(`Failed to open org: ${e.message}`);
```

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -30,3 +30,7 @@ export function showCLINotInstalledMessage() {
   );
   window.showWarningMessage(showMessage);
 }
+
+export function isSFDXContainerMode(): boolean {
+  return process.env.SFDX_CONTAINER_MODE === 'true';
```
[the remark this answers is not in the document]
lcampos: Are we doing this instead of evaluating `context.executionContext === vscode.ExtensionExecutionContext.Remote` when activating the extension, like it's mentioned in the [Supporting Remote Development docs](https://code.visualstudio.com/api/advanced-topics/remote-extensions#using-native-node.js-modules), because it's still a proposed/unstable api ?

Is this being set when the user spins up the container ?

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: getRootWorkspacePath()
+    }).execute(cancellationToken);
+
+    execution.processExitSubject.subscribe(() => {
+      this.logMetric(execution.command.logName, startTime);
+    });
+    execution.stdoutSubject.subscribe(cliResponseJSON =>
```
lcampos: The CLI returns a stream with the response, we normally use `execution.stdoutSubject.subscribe` to collect the data into a string variable and later process it in `execute.processExitSubject.subscribe`. You can take a look at how we do that in `forceOrgCreate.ts` and other commands that implement their own `execute` method.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+      this.openBrowserWithCliResponse(cliResponseJSON)
```
lcampos: Once you add error handling to `openBrowserWithCliResponse` (throw an error in catch) you should update the `execute` to catch errors and surface them to the user.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
lcampos: We should avoid adding `e.stack` to the metrics since it might contain user data like file paths, etc.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+import { telemetryService } from '../telemetry';
+import { getRootWorkspacePath, isSFDXContainerMode } from '../util';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './commands';
 
+type CLIOrgData = {
```
lcampos: Once you rebase you could try using a new interface that's very close to this https://github.com/forcedotcom/salesforcedx-vscode/blob/901e9244e04634447bcc2d886f6cfd0e4b83afa9/packages/salesforcedx-utils-vscode/src/cli/orgCreateResultParser.ts#L8

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+      telemetryService.sendErrorEvent(e.message, e.stack);
+    }
+  }
+
+  public execute(response: ContinueResponse<string>): void {
+    const startTime = process.hrtime();
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: getRootWorkspacePath()
```
lcampos: I think you need to add `env: { SFDX_JSON_TO_STDOUT: 'true' }` here so that you also get the error results from the cli in `execution.stdoutSubject.subscribe`

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,17 +6,87 @@
...
+
+    execution.processExitSubject.subscribe(() => {
+      this.logMetric(execution.command.logName, startTime);
+    });
+    execution.stdoutSubject.subscribe(cliResponseJSON =>
+      this.openBrowserWithCliResponse(cliResponseJSON)
+    );
+
+    channelService.showChannelOutput();
+    channelService.streamCommandStartStop(execution);
```
lcampos: With the normal open org command we show a notification message in the bottom right corner to show the user that the command is executing (screenshot below). This current implementation does not show that, is that intentional ?

Screen Shot 2019-07-09 at 11 10 11 PM

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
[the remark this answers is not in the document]
vazexqi: Yes. That new setting requires VS Code API 1.36 (latest). I’ve asked Jonny to abstract the check into the util here so that we can update and use that new setting later when it’s ready. For now we can just use the ENV var.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Good catch. @jonnyhork this is another place where you are breaking the contract by doing something different other than what we have already at https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts#L81

Please double-check that you are not doing anything that deviates from that. And If you are call that out explicitly, such things could be missed during code reviews since we are not always going to compare each line of your code with the base method (it doesn’t show up in the diff).

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Another good catch. This is again another place where the contract is broken from https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts#L49

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: I don’t think this is intentional. Again, we should stick as closely as possible to the contract and not deviate unless we have a good reason (and that needs to be called out).

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
vazexqi: Sorry, I was looking at `baseDeployCommand` which uses that. But the normal command in https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/commands.ts#L326 doesn’t use that.

@lcampos - Do all commands need to use that? In that case, it needs to be changed in the base command?

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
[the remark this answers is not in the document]
jonnyhork: @vazexqi I check for the value of 'true' because I thought it would be weird if it was just any value, for example `SFDX_CONTAINER_MODE = 'false'` or has a value of `"this is not in a container"`
I guess someone would have to override the value for some reason but I figured why not make it `true`.
Just my thought process, I'm open to simply 
``` return process.env.SFDX_CONTAINER_MODE```

**Comment by jonnyhork** [comment]
@vazexqi @lcampos @dehru 
I have refactored `ForceOrgOpen.ts` based on our conversations. I have added a `OrgOpenContainerResultParser` class to handle the stdOut response from the CLI differently based on success or failure.

On success the org url will be parsed and opened by VS code. The command start and stop will be displayed in the output channel. However, because I am handling the CLI response _after_ the process has exited in `processExitSubject` I can not stream the raw json response to the output channel using  `streamCommandOutput`. 

Because we do not want to show the entire unformatted json error response from the CLI could not simply steam the output before I have parsed a success/fail. So I added back in the custom formatted message from `buildUserMessageWith(cliOrgData)`.

On failure, just the error message will be displayed to the user in the output channel.
Org Open Container Success-Fail

File: packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts (hunk 13 above)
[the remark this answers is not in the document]
Comment (vazexqi) [reply packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts]: This is fine for now.

File: packages/salesforcedx-vscode-core/src/messages/i18n.ts [hunk 19]
```diff
@@ -72,6 +72,10 @@ export const messages = {
     'There was an unexpected error when processing the org create response.',
 
   force_org_open_default_scratch_org_text: 'SFDX: Open Default Org',
+  force_org_open_default_scratch_org_container_error:
```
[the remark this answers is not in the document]
jonnyhork: It is net new, but the wording mirrors that of the `force_org_create_result_parsing_error` so hopefully it will be an easy approval. I will @mention the doc writer.

vazexqi:
Let's get some unit tests in. Take a look at how other commands are writing their unit tests. I think for `forceOrgOpen` let's verify that the command is built to what we expect it. And for the CLIResponseParser that it handles different cases.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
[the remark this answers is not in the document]
jonnyhork: @hbc-bindu Would you mind taking a look at the text for `force_org_open_default_scratch_org_container_error`? Thanks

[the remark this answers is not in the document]
vazexqi: @hbc-bindu - To provide context, https://github.com/forcedotcom/salesforcedx-vscode/pull/1448#issuecomment-510595001 has the screenshot on where it occurs.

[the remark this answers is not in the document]
jonnyhork: Yes, thanks Nick. Just to be clear, the text _"Fail to decipher auth data. reason: Bad input string"_ in that screenshot is taken from the message property of the error returned by the CLI.

However the text in the PR here will be displayed in the same way and place as the whats in the screenshot.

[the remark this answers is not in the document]
hbc-bindu: @jonnyhork Thanks for providing the context.
You could rephrase the error message as:
An unexpected error occurred while processing the org open response.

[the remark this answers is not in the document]
jonnyhork: Sure can, thanks you.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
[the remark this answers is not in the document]
vazexqi: I looked at the way that we have been using ENV variables in the CLI (https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_cli_env_variables.htm).

What you have done (checking for `true`) is consistent with what the CLI is doing, so no changes now (or in the future) are necessary.

Just for context, this is not usually how other CLIs do it. Some other ones just need the ENV to be defined.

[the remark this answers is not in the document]
jonnyhork: ok, thanks for checking on that.

packages/salesforcedx-vscode-core/src/commands/forceOrgOpen.ts
```diff
@@ -6,36 +6,123 @@
...
+  public build(data: {}): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_org_open_default_scratch_org_text'))
+      .withArg('force:org:open')
+      .withLogName('force_org_open_default_scratch_org')
+      .build();
+  }
+}
+
+export function getExecutor(): SfdxCommandletExecutor<{}> {
```
jonnyhork: I extracted the ternary into a function to make it easier for testing. I also think its a cleaner implementation and it mimics the pattern in other commands such as authWebLogin.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -30,3 +30,7 @@ export function showCLINotInstalledMessage() {
   );
   window.showWarningMessage(showMessage);
 }
+
+export function isSFDXContainerMode(): boolean {
+  return process.env.SFDX_CONTAINER_MODE ? true : false;
```
jonnyhork: There were previous discussions that checking for the string 'true' was too specific, so I refactored to only check for the presence of the env variable. This is also how other util methods check for env variables.

packages/salesforcedx-vscode-core/test/unit/commands/forceOrgOpen.test.ts
```diff
@@ -0,0 +1,14 @@
+import { expect } from 'chai';
```
jonnyhork: This file was mistakenly left in, I will remove this from the PR. The test was moved to the integration/commands directory.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -64,6 +67,7 @@
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run test:vscode-integration",
+    "test:unit": "./node_modules/.bin/nyc ./node_modules/.bin/_mocha --recursive out/test/unit --reporter mocha-multi-reporters --timeout 20000 --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaUnitTestsConfig.json",
```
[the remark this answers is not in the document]
lcampos: Remove everything that's not needed, if we need it in the future we can take a look at this branch/PR.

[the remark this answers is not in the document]
jonnyhork: Sounds good, thanks.

jonnyhork:
> Let's get some unit tests in. Take a look at how other commands are writing their unit tests. I think for `forceOrgOpen` let's verify that the command is built to what we expect it. And for the CLIResponseParser that it handles different cases. -nchen

@lcampos I have added some tests for forceOrgOpen and the result parser. Looks like there is code coverage issues. Is there a certain target we need to hit? Are there other tests you would like added at this time?

.vscode/launch.json
```diff
@@ -203,6 +203,29 @@
       "preLaunchTask": "Compile",
       "internalConsoleOptions": "openOnSessionStart"
     },
+    {
```
lcampos: This launch configuration is not needed.

jonnyhork: Thanks, I forgot about this change. I'll remove it.

PR #1456, opened 2019-07-08
Refactor setting command context after scratch org creation

lcampos:
### What does this PR do?
The workspace org type context does not switch between non-source tracking context to source tracking when the user is creating a new scratch org in a slow setups. This PR reworks the initial approach to setting the context in order to do it when the CLI is done executing instead of relying on configuration listeners.

@W-6289619@


packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/baseTemplateCommand.test.ts
```diff
@@ -61,8 +61,8 @@ describe('FilePathExistsChecker for Aura bundle', () => {
     });
 
     after(() => {
-      sinon.restore(vscode.workspace);
-      sinon.restore(notificationService);
+      findFilesSpy.restore();
```
lcampos: Fixes some flapping tests caused by new proposed api's in version 1.36 of vscode.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -31,24 +27,11 @@ export async function getWorkspaceOrgType(
     defaultUsernameOrAlias
   );
   const isUsername = email ? true : false;
-
```
lcampos: We can now remove the retry code since the org create command will now set the correct context after it finishes executing.

packages/salesforcedx-utils-vscode/src/cli/orgCreateResultParser.ts
```diff
@@ -0,0 +1,53 @@
...
+      err.name = 'OrgCreateParserFail';
+      throw err;
+    }
+  }
+
+  public createIsSuccessful(): boolean {
+    return this.response && this.response.status === 0 ? true : false;
+  }
+
+  public getResult(): OrgCreateSuccessResult | OrgCreateErrorResult {
```
lcampos: Originally I wanted to use the success result data to set the org type in the execute method of the command. I didn't end up doing that because we currently only run the command for scratch orgs but we'll soon have to since it's also going to support creating sandboxes. 
Do we want to keep this around or should I change it to exclusively returning `OrgCreateErrorResult` ? I'm ok with both options.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -68,6 +68,8 @@ export const messages = {
 
   force_org_create_default_scratch_org_text:
     'SFDX: Create a Default Scratch Org...',
+  force_org_create_result_parsing_error:
```
lcampos: @hbc-bindu This is the only new message that we'll potentially surface to users.

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -53,8 +62,60 @@ export class ForceOrgCreateExecutor extends SfdxCommandletExecutor<
...
+      this.logMetric(execution.command.logName, startTime);
+      try {
+        const createParser = new OrgCreateResultParser(stdOut);
+
+        if (createParser.createIsSuccessful()) {
+          // NOTE: there is a beta in which this command also allows users to create sandboxes
+          // once it's GA this will have to be updated
+          setWorkspaceOrgTypeWithOrgType(OrgType.SourceTracked);
+        } else {
+          const errorResponse = createParser.getResult() as OrgCreateErrorResult;
```
lcampos: Handling the scratch org creation will look something like this:

scratch_org_create_error

.gitignore
```diff
@@ -91,6 +91,9 @@ jspm_packages/
 
 # Project test data
 packages/system-tests/project_*
+packages/system-tests/assets/sfdx-simple/.vscode/
```
lcampos: Adding these so we can ignore those files when running tests locally

packages/salesforcedx-utils-vscode/src/cli/orgCreateResultParser.ts
brpowell: I think it's okay the way you have it now. You left a note to address this when we support creating sandboxes.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
hbc-bindu: @lcampos, you could rephrase the message as:
An unexpected error occurred while processing the org create response.
PR seems to have been merged into develop; please update, if possible.

PR #1459, opened 2019-07-10
Folder support for org browser

brpowell:
### What does this PR do?

Fetch folders and put component nodes in their respective folders for a metadata type that uses folders.

W-5980634


packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -10,26 +10,35 @@ export enum NodeType {
...
+    this.fullName = fullName || label;
     switch (this.type) {
       case NodeType.Org:
         this.collapsibleState = vscode.TreeItemCollapsibleState.Expanded;
-        this.tooltip = 'Default Org';
         break;
       case NodeType.MetadataCmp:
         this.collapsibleState = vscode.TreeItemCollapsibleState.None;
-        this.tooltip = 'Metadata Component';
+        this.iconPath = vscode.ThemeIcon.File;
```
brpowell: File and folder icons based on the user's icon theme:

Screen Shot 2019-07-09 at 4 42 40 PM

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -29,13 +31,21 @@ export class ComponentUtils {
       (await OrgAuthInfo.getUsername(defaultUsernameOrAlias)) ||
       defaultUsernameOrAlias;
 
+    let fileName = metadataType;
+    if (TypeUtils.FOLDER_TYPES.has(metadataType)) {
```
AnanyaJha: Is it possible to just check if a folder is specified instead of also verifying the metadata is a folder type? Is there any scenario in which this method could be called with a folder specified without actually being a folder type? It seems like this could only be triggered with a folder once it passes the check in 'getComponents' of the outline provider.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -86,18 +86,31 @@ export class MetadataOutlineProvider
...
-    );
-    const nodeList = componentsList.map(
-      cmp => new BrowserNode(cmp, NodeType.MetadataCmp)
-    );
-    if (nodeList.length === 0) {
-      nodeList.push(
-        new BrowserNode(nls.localize('empty_components'), NodeType.EmptyNode)
+    let nodeList: BrowserNode[] = [];
+    if (TypeUtils.FOLDER_TYPES.has(metadataType.label!)) {
+      let typeName = metadataType.label!;
```
lcampos: I think this logic for getting the correct typeName for a metadataType that supports folders should in a util instead of having it here. Maybe having it in `TypesUtils`.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -86,18 +86,31 @@ export class MetadataOutlineProvider
...
+      if (typeName === 'EmailTemplate') {
+        typeName = 'Email';
+      }
+
+      const folders = await this.loadAndMapComponents(
+        `${typeName}Folder`,
+        NodeType.Folder
+      );
+      for (const folder of folders) {
```
lcampos: What's the limit of folders per Metadata Type ? Doing this loop represents 1 query per iteration/folder because we're calling `this.loadAndMapComponents`.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -21,6 +21,13 @@ type MetadataObject = {
   xmlName: string;
 };
 export class TypeUtils {
+  public static readonly FOLDER_TYPES = new Set([
```
lcampos: Are these the only types that support folders ?

brpowell: Sources:
https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_folder.htm
https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_listmetadataquery.htm

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -57,50 +57,53 @@ export class MetadataOutlineProvider
...
-      cmp => new BrowserNode(cmp, NodeType.MetadataCmp)
-    );
-    if (nodeList.length === 0) {
-      nodeList.push(
-        new BrowserNode(nls.localize('empty_components'), NodeType.EmptyNode)
-      );
+  public async getComponents(node: BrowserNode): Promise<string[]> {
+    const cmpUtils = new ComponentUtils();
+    try {
+      if (TypeUtils.FOLDER_TYPES.has(node.fullName)) {
```
lcampos: I don't have a strong opinion on this but this could also be a switch where line 103 is the default behavior.

Issue #1451, opened 2019-07-05
salesforcedx-vscode-core.push-or-deploy-on-save.enabled Stop working


Kxa422:
I've notice recently that the settings salesforcedx-vscode-core.push-or-deploy-on-save.enabled have stop working.
It was working in the past but now I need to right click and deploy to org to have the code pushed to the org.

arunmitteam:
@ntotten I'm having the same issue. Should I setup my workspace from scratch again? I updated my extension and CLI and still the push or save is not kicking off the deploy.
As per this issue, it has been taken care but it has unfortunately broken the functionality for some users. https://github.com/forcedotcom/salesforcedx-vscode/issues/1155#issuecomment-508249942

lcampos:
Hi @arunmitteam & @Kxa422, could you provide the following info so we can try and replicate this on our end.

- VS Code version
- OS version
- Salesforce extensions version
- Salesforce CLI version

arunmitteam:
OS Version - macOS 10.14.5 
VS Code version - Version 1.36.0 (1.36.0)
Salesforce extensions - 46.4.0
Salesforce CLI - sfdx-cli/7.13.0-27dbcb37d3 darwin-x64 node-v10.15.3

Kxa422:
OS Version - Windows 10 Pro 1709 (Os Build 16299 1087)
VS code version - 1.36.0
Salesforce CLI - sfdx-cli/7.13.0-27dbcb37d3 win32-x64 node-v10.15.3
salesforcedx 46.3.0 (core)
├─ force-language-services 46.8.0 (core)
└─ salesforce-alm 46.7.0 (core)

lcampos:
@arunmitteam I have a similar config now, can you tell me whether you're setting the `"salesforcedx-vscode-core.push-or-deploy-on-save.enabled": true` flag in your project's `.vscode/settings.json` file or on the User settings.json file (Open Command Palette > Preferences: Open Settings) that applies to all VS Code ?

lcampos:
@Kxa422 what about your Salesforce extensions for vs code ? You can see their version by clicking on the extensions icon in VS Code's left side bar.
![extensions](https://user-images.githubusercontent.com/1041842/60736162-de534800-9f0a-11e9-9290-226a59304407.png)

arunmitteam:
@lcampos 
yes, I have it enabled. See below.
[...]

PR #1464, opened 2019-07-11
Remove dirty doc check for deploy on save and add telemetry

brpowell:
### What does this PR do?

We're suspecting the way the listeners are set up to determine what documents are dirty before kicking of a deploy is problematic for some users. This removes the check for a dirty document before kicking off a deploy and will do so whether or not the file actually changed when a user saves it. To help further investigate, telemetry data is being added. 

W-6330470
#1451

### DIFF
--- a/packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
@@ -47,7 +47,6 @@ export class DeployQueue {
   }
 
   public async enqueue(document: vscode.Uri) {
-    telemetryService.sendEventData('deployOnSaveEnqueue');
     this.queue.add(document);
     await this.wait();
     await this.doDeploy();
@@ -141,7 +140,9 @@ function displayError(message: string) {
   notificationService.showErrorMessage(message);
   channelService.appendLine(message);
   channelService.showChannelOutput();
-  telemetryService.sendError(`DeployOnSaveError: ${message}`);
+  telemetryService.sendError(
+    `DeployOnSaveError: Documents were queued but a deployment was not triggered`
+  );
 }
 
 async function ignorePath(uri: vscode.Uri) {
--- a/packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
@@ -15,6 +15,7 @@ import { SfdxPackageDirectories } from '../sfdxProject';
 import * as path from 'path';
 import { setTimeout } from 'timers';
 import * as vscode from 'vscode';
+import { telemetryService } from '../telemetry';
 import { hasRootWorkspace, OrgAuthInfo } from '../util';
 
 export class DeployQueue {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -45,6 +47,7 @@ export class DeployQueue {
   }
 
   public async enqueue(document: vscode.Uri) {
+    telemetryService.sendEventData('deployOnSaveEnqueue');
```
brpowell: This event will send every enqueue - going back and forth on whether we need this or not.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -190,8 +190,8 @@ export class TelemetryService {
     }
   }
 
-  private getEndHRTime(hrstart: [number, number]): string {
+  public getEndHRTime(hrstart: [number, number]): string {
     const hrend = process.hrtime(hrstart);
-    return util.format('%d%d', hrend[0], hrend[1] / 1000000);
+    return util.format('%d.%d', hrend[0], Math.round(hrend[1] / 1000000));
```
brpowell: I think for this we were concatenating the milliseconds to the seconds. 5s and 430.123ms would look something like 5430.123 instead of 5.430

lcampos: Let's leave this change in format out of this PR since it will also change the way we measure all current perf info and dashboards.

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -86,6 +89,19 @@ export class DeployQueue {
...
+
+        telemetryService.sendEventData(
+          'deployOnSave',
+          {
+            deployType: orgType === OrgType.SourceTracked ? 'Push' : 'Deploy'
+          },
+          {
+            documentsToDeploy: toDeploy.length,
+            waitTimeForLastDeploy: this.deployWaitStart
+              ? parseFloat(telemetryService.getEndHRTime(this.deployWaitStart))
```
lcampos: You might want to move the `telemetry.getEndHRTime` changes to a new method in this file for now. One thing to keep in mind is that the rest of the commands returns any time measure in milliseconds. Sending the time info in milliseconds (5400 for 5 seconds and 400milliseconds) makes is easier to build dashboards and queries using the time info since there's no pre-formatting needed.

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
```diff
@@ -135,6 +141,7 @@ function displayError(message: string) {
   notificationService.showErrorMessage(message);
   channelService.appendLine(message);
   channelService.showChannelOutput();
+  telemetryService.sendError(`DeployOnSaveError: ${message}`);
```
lcampos: I think we should remove the message variable from here. It returns data like Metadata Type info e.g. custom object names

packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts
lcampos: Having the metric in line 93 already contains the number of items to be deployed so I think we could remove this one.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
brpowell: Doh, my bad I didn't notice the dashboard were in ms.

PR #1470, opened 2019-07-11
 [WIP] [Org Browser] Retrieve source from org

AnanyaJha:
### What does this PR do?
This PR retrieves individual metadata components from the org using an inline icon in the org browser.

@W-5768040@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -114,7 +114,8 @@
         },
         {
           "command": "sfdx.force.source.retrieve",
-          "when": "view == metadata && viewItem == component"
+          "when": "view == metadata && viewItem == component",
+          "group": "inline"
         }
       ],
       "editor/context": [
@@ -647,7 +648,11 @@
       },
       {
         "command": "sfdx.force.source.retrieve",
-        "title": "%force_source_retrieve_text%"
+        "title": "%force_source_retrieve_display_text%",
+        "icon": {
+          "light": "resources/light/retrieve.svg",
+          "dark": "resources/dark/retrieve.svg"
+        }
       }
     ],
     "configuration": {
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -111,6 +111,10 @@
         {
           "command": "sfdx.force.task.stop",
           "when": "view == sfdx.force.tasks.view"
+        },
+        {
+          "command": "sfdx.force.source.retrieve",
+          "when": "view == metadata && viewItem == component"
         }
       ],
       "editor/context": [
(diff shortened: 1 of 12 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -111,6 +111,11 @@
         {
           "command": "sfdx.force.task.stop",
           "when": "view == sfdx.force.tasks.view"
+        },
```
AnanyaJha: A heads up- the api to display the icon inline only displays the icon when the user is hovering over it.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -104,4 +105,11 @@ export class ComponentUtils {
     }
     return componentsList;
   }
+
+  public async retrieveComponent(componentNode: BrowserNode) {
+    const type = componentNode.metadataType;
+    const component = componentNode.label; // this will use componentNode.fullName once folder changes are merged
```
AnanyaJha: We'll use the 'fullName' property of the nodes once the folder changes are merged. To retrieve components from folders, we have to specify the folder that the component belongs to. We won't need to do anything special to handle that at this point, because the `fullName` property will contain the information needed ie. 'folderOne/component'.

AnanyaJha:
@ntotten Since the drop down menu on individual components will only have a single menu option 'Retrieve Source from Org, I've put in inline icons instead. 

Screen Shot 2019-07-11 at 3 08 10 PM

This would just be for the beta version, until we have more actions we can populate the right-click drop down menus with. Thoughts?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -89,6 +89,7 @@ export const messages = {
   force_source_deploy_select_manifest:
     'You can run SFDX: Deploy Source in Manifest to Org only on a manifest file.',
   force_source_retrieve_text: 'SFDX: Retrieve Source from Org',
+  force_source_retrieve_display_text: 'Retrieve Source from Org',
```
AnanyaJha: @hbc-bindu This is the text that would display when hovering over the 'Retrieve' icon on individual components.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -0,0 +1,49 @@
...
+import { nls } from '../messages';
+import {
+  EmptyParametersGatherer,
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from './commands';
+
+export class ForceSourceRetrieveExecutor extends SfdxCommandletExecutor<
+  string
```
AnanyaJha: Right now, the user sees the progress notification on the bottom right hand indicating that the retrieve is running. Do we want to display the notification to the user?

PR #1473, opened 2019-07-13
Added flag to enable org browser

AnanyaJha:
### What does this PR do?
This PR adds a flag to display the org browser for internal testing purposes.

@W-6330612@


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -497,8 +497,9 @@ export async function activate(context: vscode.ExtensionContext) {
   orgList.displayDefaultUsername(defaultUsernameorAlias);
   context.subscriptions.push(registerOrgPickerCommands(orgList));
 
-  // await setupOrgBrowser(context, defaultUsernameorAlias);
-  vscode.commands.executeCommand('setContext', 'sfdx:display_tree_view', false);
+  if (sfdxCoreSettings.showOrgBrowser()) {
```
lcampos: I think we should remove checking for the setting here. Having the flag being used in the `when` clause for the tree view will immediately show the option once you modify the setting file but the tree view itself will throw an error if you try to use it since the retrieve commands are not registered since `setupOrgBrowser` didn't run in activation.

AnanyaJha: Sounds good, we can remove the check.

packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -12,6 +12,7 @@ import {
   RETRIEVE_TEST_CODE_COVERAGE,
   SFDX_CORE_CONFIGURATION_NAME,
   SHOW_CLI_SUCCESS_INFO_MSG,
+  SHOW_ORG_BROWSER,
```
lcampos: You can also remove this since `showOrgBrowser` is not used anymore.

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -20,3 +20,4 @@ export const PUSH_OR_DEPLOY_ON_SAVE_ENABLED = 'push-or-deploy-on-save.enabled';
 export const RETRIEVE_TEST_CODE_COVERAGE = 'retrieve-test-code-coverage';
 export const SFDX_CLI_DOWNLOAD_LINK =
   'https://developer.salesforce.com/tools/sfdxcli';
+export const SHOW_ORG_BROWSER = 'show-org-browser';
```
lcampos: You can remove this too

packages/salesforcedx-vscode-core/package.json
```diff
@@ -680,6 +680,12 @@
           ],
           "default": false,
           "description": "%internal_dev_description%"
+        },
+        "salesforcedx-vscode-core.show-org-browser": {
+          "type": [
+            "boolean"
+          ],
+          "default": false
```
brpowell: Will this use a checkbox in the settings? :)

AnanyaJha: This one isn't configured for the checkbox- I thought it would be better to keep it as basic as possible since we don't want customers easily finding it and playing around with it. I'll look into how we can do checkboxes for our other customer-facing flags though!

PR #1479, opened 2019-07-15
[Org Browser] Retrieve Source from Org

AnanyaJha:
### What does this PR do?
This PR retrieves individual metadata components from the org using an inline icon in the org browser.

@W-5768040@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -57,5 +57,6 @@
   "force_source_delete_this_source_text": "SFDX: Delete This from Project and Org",
   "force_config_set_org_text": "SFDX: Set a Default Org",
   "internal_dev_description": "Specifies whether or not to set up internal development mode which provides limited SFDX commands.",
-  "refresh_org_browser_text": "Refresh Org Browser"
+  "refresh_org_browser_text": "Refresh Org Browser",
+  "force_source_retrieve_display_text": "Retrieve Source from Org"
 }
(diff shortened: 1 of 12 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -57,5 +57,6 @@
   "force_source_delete_this_source_text": "SFDX: Delete This from Project and Org",
   "force_config_set_org_text": "SFDX: Set a Default Org",
   "internal_dev_description": "Specifies whether or not to set up internal development mode which provides limited SFDX commands.",
-  "refresh_org_browser_text": "Refresh Org Browser"
+  "refresh_org_browser_text": "Refresh Org Browser",
+  "force_source_retrieve_display_text": "Retrieve Source from Org"
```
AnanyaJha: @hbc-bindu The hover text would look like this: 
Screen Shot 2019-07-15 at 3 14 18 PM

Let me know if that looks okay!

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieve.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(metadataArg: string) {
+    super();
+    this.metadataArg = metadataArg;
+  }
+
+  public build(): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_source_retrieve_text'))
+      .withArg('force:source:retrieve')
+      .withFlag('-m', this.metadataArg)
```
lcampos: Need to add `.withLogName` to this command.

PR #1480, opened 2019-07-15
aura language server - doc updates

nrkruk:
### What does this PR do?
Update documentation for the aura language server


docs/Gemfile.lock
```diff
@@ -252,4 +252,4 @@ DEPENDENCIES
   minima (~> 2.0)
 
 BUNDLED WITH
-   1.17.2
+   2.0.1
```
ntotten: This breaks netlify builds. Bundler should not be updated to version 2 yet.

PR #1481, opened 2019-07-16
Refresh metadata components

brpowell:
### What does this PR do?

Allow refreshing metadata components by metadata type or folder.

W-5768021


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -408,9 +408,13 @@ async function setupOrgBrowser(
...
     }
   });
-  vscode.commands.registerCommand('sfdx.force.metadata.view.refresh', () => {
-    return metadataProvider.refresh();
-  });
+
+  vscode.commands.registerCommand(
+    'sfdx.force.metadata.view.refresh',
+    async node => {
+      await metadataProvider.refresh(node);
```
brpowell: Refresh command based on the node given

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -57,5 +54,5 @@
   "force_source_delete_this_source_text": "SFDX: Delete This from Project and Org",
   "force_config_set_org_text": "SFDX: Set a Default Org",
   "internal_dev_description": "Specifies whether or not to set up internal development mode which provides limited SFDX commands.",
-  "refresh_org_browser_text": "Refresh Org Browser"
```
brpowell: Need opinions here. I wanted to have a separate label for the individual refresh buttons from the global refresh, but that would involve adding another command just to have the different label. Is it okay if I generalize it more to "Refresh"?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -111,6 +111,11 @@
         {
           "command": "sfdx.force.task.stop",
           "when": "view == sfdx.force.tasks.view"
+        },
+        {
+          "command": "sfdx.force.metadata.view.refresh",
+          "when": "view == metadata && viewItem == refreshable",
```
brpowell: Doesn't seem like the when clause supports OR so i had to mark the contextValue for nodes as "refreshable" that support the functionality

AnanyaJha: Currently, the user can view the 'refresh' icon on metadata type nodes even if they haven't expanded the node to view the components yet. If you click on refresh, the refresh is executed but the component nodes still aren't displayed because the user hasn't expanded the type node. I think it can be a little unintuitive to have the ability to refresh even if the initial loading of the components hasn't occurred yet- is it possible to only display the refresh icon when the node is expanded?

AnanyaJha: Also- it looks like the when clause doesn't support OR but it does support regex. If we would like to assign the appropriate context values for folder and type nodes (ie. 'folder' and 'type') we can set the clause to: `"when": "view == metadata && viewItem =~ /(folder|type)/"` and it should work!

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -33,10 +34,15 @@ export class MetadataOutlineProvider
...
   }
 
-  public async refresh(): Promise<void> {
-    const usernameOrAlias = await this.getDefaultUsernameOrAlias();
-    this.defaultOrg = usernameOrAlias;
-    this.internalOnDidChangeTreeData.fire();
+  public async refresh(node?: BrowserNode): Promise<void> {
+    if (node && !node.toRefresh) {
+      node.toRefresh = true;
+    } else if (!this.toRefresh) {
```
AnanyaJha: Is this check necessary? It looks like the refresh function should only get called if the user is attempting to refresh- and if it's not a refresh at the type level, it has to be a refresh at the org level. Can we make this an `else`?

packages/salesforcedx-vscode-core/package.json
brpowell: There's sort of a way to do it but it's awkward. The tree data change event has to be fired a second time which makes this stuttering effect. And even then you have to click out of the tree view and click back in to it to get the icon to actually load.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
brpowell: The condition is to stop people from spamming refresh and making a ton of api calls if one is already in progress.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -57,19 +63,19 @@ export class MetadataOutlineProvider
...
+        this.toRefresh = false;
+        break;
+      case NodeType.Folder:
+      case NodeType.MetadataType:
+        const type = TypeUtils.FOLDER_TYPES.has(element.fullName)
+          ? NodeType.Folder
+          : NodeType.MetadataCmp;
+        element.setChildren(await this.getComponents(element), type);
+        element.toRefresh = false;
+        break;
```
lcampos: Do we need a `default` here ?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -87,20 +93,28 @@ export class MetadataOutlineProvider
...
-        );
-      } else if (node.type === NodeType.Folder) {
-        return await cmpUtils.loadComponents(
-          this.defaultOrg!,
-          node.parent!.fullName,
-          node.fullName
-        );
+      let typeName;
+      let folder;
+      switch (node.type) {
```
lcampos: Nice

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
brpowell: No other node type can be expanded and have it's children fetched, so not necessarily.

PR #1490, opened 2019-07-18
Filter out unsupported org browser types

brpowell:
### What does this PR do?

Some types are considered unsupported with the org browser, so this adds a list to filter them out of the view.

W-6354663


ntotten:
Make sure we document this once we release. //cc @hbc-bindu

PR #1492, opened 2019-07-19
Don’t show "no code coverage" warning when viewing a testclass

Yerrak:
### What does this PR do?
Minor tweak regarding the display of the _No code coverage information was found for this file [...]_ warning when viewing test classes.
Now, the warning is only displayed in case `@istest` is not included inside the opened file.

No reference, just a minor nuisance I happen to run into a lot, always drawing my attention to the warning when switching editor tabs.

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
+++ b/packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
@@ -157,10 +157,17 @@ export class CodeCoverage {
             covItem.name === getApexMemberName(editor.document.uri.fsPath)
         );
 
-        if (!codeCovItem) {
+        const isTestclass = editor.document
+          .getText()
+          .toLowerCase()
+          .includes('@istest');
+
+        if (!codeCovItem && !isTestclass) {
           throw new Error(
             nls.localize('colorizer_no_code_coverage_current_file')
           );
+        } else if (!codeCovItem) {
+          return;
         }
 
         for (const key in codeCovItem.lines) {


packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
```diff
@@ -157,10 +157,17 @@ export class CodeCoverage {
             covItem.name === getApexMemberName(editor.document.uri.fsPath)
         );
 
-        if (!codeCovItem) {
+        const isTestclass = editor.document
+          .getText()
+          .toLowerCase()
+          .includes('@istest');
```
brpowell: The check for if it's a test class should happen before `getCoverageData()` on line 154. One possibility is to replace the `isApexMetadata()` check on 153 with another that still uses it:

```
function canColorize(document: TextDocument) {
  const isTest = document
    .getText()
    .toLowerCase()
    .includes('@istest');
  return isApexMetadata(document.uri.fsPath) && !isTest;
}
```

you could define it right under the isApexMetadata definition.

malte-ST: First of all, sorry for the late response!

I really like your solution and I'd be happy to change the PR accordingly, if you think the following is a non-issue!

I think, there is a case where this could potentially hide coverage data that actually exists. Since checking for the occurrence of `@istest` in a file is technically not an indication whether the file is indeed a testclass (think comments or very old API versions where normal classes were allowed to contain their respective test methods), the proposed adjustment might prevent showing coverage data in these cases.

The current solution would still allow displaying the coverage data (since the `codeCovItem` is defined, even though `@isTest` is included in the current file).

Yerrak: First of all, sorry for the late response!

I really like your solution and I'd be happy to change the PR accordingly, if you think the following is a non-issue!

I think, there is a case where this could potentially hide coverage data that actually exists. Since checking for the occurrence of @istest in a file is technically not an indication whether the file is indeed a testclass (think comments or very old API versions where normal classes were allowed to contain their respective test methods), the proposed adjustment might prevent showing coverage data in these cases.

The current solution would still allow displaying the coverage data (since the codeCovItem is defined, even though @isTest is included in the current file).

brpowell: Hi @Yerrak. I see what you mean, that is a good point. I think I'm still confused on how this remediates the issue though - can you give some repro steps?

Yerrak: Hi @brpowell, happy to clarify!
I see the following cases/types of `.cls` files being affected:
1- no occurrence of `@isTest`: everything works as before because `isTestclass` would always be false.
2- `@isTest` occurs and we do not have coverage data for the file: This is the case I was mainly trying to catch since we never have coverage data for a test class. In this instance through `l. 165` we do not throw an error, but just return. Basically like before, but now the error is suppressed.
3a- `@isTest` occurs (eg. in a comment) and we somehow do have coverage data for the file: Everything works as before, we just colorize.
3b- `@isTest` occurs (eg. in a comment), no coverage data exists and it is actually *not* a test class: In this case, the error is suppressed even though the `.cls` file is not covered.

My only concern would be case `3b` since we (I guess wrongly) suppress the error of having no coverage data available for the current file. From my experience, this is something that should rarely occur.
On a side note: I _think_ `3b` is also a non-issue since we always get (empty/0%) coverage data for classes that are just not covered by the tests. This would mean, that a `codeCovItem` exists which translates to case `3a`. I'm not 100% sure about this behaviour though.

brpowell: Thanks for clarifying Yerrak. One thing to keep in mind is that on line 154 where we get the coverage data, if there isn't a file for the class that method will throw a "no code coverage error". Also I think instead of checking just this annotation text in the document, we should leverage the Apex Language Server to determine if it is a test class. Check out the [languageClientUtils method](https://github.com/forcedotcom/salesforcedx-vscode/blob/bc4e38a5feedebd4248659ade57d6b51d513b44d/packages/salesforcedx-vscode-apex/src/languageClientUtils/languageClientUtils.ts#L93) the test sidebar uses to fetch the test methods. The results contain doc URIs that could be leveraged.

lcampos:
Closing this since there hasn't been any activity on it and it should be addressed by using the Apex language server.

PR #1493, opened 2019-07-19
Only allow unmanaged components

brpowell:
### What does this PR do?

Only allow unmanaged components. i.e. components in unmanaged packages or no package at all.

W-6358467


packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -59,7 +59,10 @@ export class ComponentUtils {
       if (!isNullOrUndefined(cmpArray)) {
         cmpArray = cmpArray instanceof Array ? cmpArray : [cmpArray];
         for (const cmp of cmpArray) {
-          if (!isNullOrUndefined(cmp.fullName)) {
+          if (
+            !isNullOrUndefined(cmp.fullName) &&
+            cmp.manageableState === 'unmanaged'
```
brpowell: Eventually we'll move this to a set and include the enum value for unlocked packaging

PR #1498, opened 2019-07-22
Update Org Browser setting description and surface setting through UI

brpowell:
### What does this PR do?

Update the org browser flag to give a description and surface changing the value through the settings UI.

W-6369516

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -95,7 +95,7 @@
         {
           "id": "metadata",
           "name": "Metadata",
-          "when": "config.salesforcedx-vscode-core.show-org-browser && sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
+          "when": "config.salesforcedx-vscode-core.orgBrowser.enabled && sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
         }
       ]
     },
@@ -707,11 +707,10 @@
           "default": false,
           "description": "%internal_dev_description%"
         },
-        "salesforcedx-vscode-core.show-org-browser": {
-          "type": [
-            "boolean"
-          ],
-          "default": false
+        "salesforcedx-vscode-core.orgBrowser.enabled": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "%org_browser_enabled_description%"
         }
       }
     },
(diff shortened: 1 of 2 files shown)


brpowell:
Screen Shot 2019-07-22 at 4 34 21 PM

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -56,5 +56,6 @@
   "internal_dev_description": "Specifies whether or not to set up internal development mode which provides limited SFDX commands.",
   "refresh_types_text": "Refresh Types",
   "refresh_components_text": "Refresh Components",
-  "force_source_retrieve_display_text": "Retrieve Source from Org"
-}
+  "force_source_retrieve_display_text": "Retrieve Source from Org",
+  "org_browser_enabled_description": "(Beta) Specifies whether to show or hide the Org Browser view when a non-scratch org is set as the default org. See the [documentation](https://forcedotcom.github.io/salesforcedx-vscode/user-guide/org-browser) to learn about the Org Browser."
```
brpowell: @hbc-bindu let me know what you think of this wording. I also may need to update the documentation link depending on feedback from the documentation PR.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -707,11 +707,10 @@
           "default": false,
           "description": "%internal_dev_description%"
         },
-        "salesforcedx-vscode-core.show-org-browser": {
-          "type": [
-            "boolean"
-          ],
-          "default": false
+        "salesforcedx-vscode-core.orgBrowser.enabled": {
```
lcampos: Calling it `orgBrowser.enabled` gives the user that impression that the setting is already enabled when seeing it in the UI. At least we've had users being confused about it in the past.

packages/salesforcedx-vscode-core/package.nls.json
AnanyaJha: Maybe we could say 'Specifies whether to display the...'? not sure

packages/salesforcedx-vscode-core/package.json
brpowell: VS Code uses this convention throughout their settings. I think it was confusing for us before because we weren't surfacing the setting through the UI to see the state.

Screen Shot 2019-07-23 at 9 25 09 AM

packages/salesforcedx-vscode-core/package.nls.json
hbc-bindu: Hi @brpowell, you could rephrase it as:
Indicates whether the Org Browser view is displayed when the default org is a non-scratch org. See Org Browser [documentation] (https://..........)

packages/salesforcedx-vscode-core/package.json
lcampos: I see what you mean, then let's keep it as it is right now.

PR #1499, opened 2019-07-23
[Org Browser] Check if file exists locally before running a retrieve

AnanyaJha:
### What does this PR do?
This PR adds an alert for the user indicating that running the `Retrieve Source from Org` from the org browser might overwrite pre-existing local files.

@W-6350303@


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -125,6 +125,8 @@ export const messages = {
     'One or more %s files with the specified path already exist in your workspace. Do you want to overwrite them?',
   warning_prompt_dir_overwrite:
     'A folder with the specified project name already exists in the selected directory. Do you want to overwrite it?',
+  warning_prompt_cmp_file_overwrite:
```
AnanyaJha: @hbc-bindu @ntotten This is what the message currently looks like if the user wants to run a retrieve on a component and he/she already has a local version of that file present. 
Screen Shot 2019-07-22 at 5 42 58 PM
Thoughts?

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
```diff
@@ -0,0 +1,177 @@
...
+    return { type: 'CANCEL' };
+  }
+
+  private async createFilesGlob(
+    dirName: string,
+    fileName: string
+  ): Promise<string> {
+    const packageDirectories = await SfdxPackageDirectories.getPackageDirectoryPaths();
+    const basePath = path.join(
+      packageDirectories[0],
```
lcampos: `packageDirectories[0]` is always the path set as default in `sfdx-project.json`'s `packageDirectories` array . There is a problem in this method because it looks like the cli retrieve command is not respecting setting `default: true` to a path different than `force-app`.

Screen Shot 2019-07-22 at 7 33 54 PM

AnanyaJha: Do you have anything else in your `sfdx-project.json`? I'm trying to reproduce what you're seeing, but I haven't been able to yet.
Screen Shot 2019-07-22 at 8 08 57 PM

lcampos: I was testing why a brand new project I created using `SFDX: Create Project with Manifest`. I only added new `path` elements under `packageDirectories`. I basically edited the `sfdx-project.json` folder to add new paths and change which one is the default. After saving the change I went to the org browser and did a retrieve. It looks like the cli takes some time to catch that there was a change on which package directory is now set to default. Might be a caching issue at the cli level.

AnanyaJha: Ah okay, so I tried it again with a new project but I think the cli was able to catch the change on the package directory this time. If this seems like an intermittent issue we might want to address, I could add some telemetry for now. We could look at the frequency of alternate default `path`s and whether the file is actually created under the specified path after the user retrieves; this should let us see if there's any discrepancy between where the file was intended to be created and where it actually got created.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
hbc-bindu: Hi @AnanyaJha, you could rephrase the first sentence as:
"<Filename>" already exists in your workspace. Do you want to overwrite it?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -13,11 +13,11 @@ import { nls } from '../messages';
 import { telemetryService } from '../telemetry';
 import { getRootWorkspacePath, hasRootWorkspace, OrgAuthInfo } from '../util';
 
-type MetadataObject = {
+export type MetadataObject = {
   directoryName: string;
   inFolder: boolean;
   metaFile: boolean;
-  suffix: string;
+  suffix?: string;
```
brpowell: Is a suffix ever not returned in the response?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveCmp.test.ts
```diff
@@ -0,0 +1,283 @@
...
+      'ApexClass',
+      NodeType.MetadataType,
+      undefined,
+      metadataObject
+    );
+    const suffixes = generateSuffix(typeNode, 'ApexClass');
+    expect(suffixes).to.eql([`.${typeNode.metadataObject!.suffix}-meta.xml`]);
+  });
+
+  it('should generate appropriate suffix if lightning or aura type', async () => {
```
brpowell: Should we also aura?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
AnanyaJha: Screen Shot 2019-07-23 at 2 41 24 PM
 Hi @hbc-bindu I actually tweaked the message a bit, because of some inconsistent behavior we found in the CLI. When the user runs this command, they're expecting the file to be retrieved into their specified default directory. Initially, we were only checking their default directory to see if the file was present- and they would only see this message if it were present.

Occasionally, the file may actually get retrieved into some other package directory instead of the default, potentially overwriting the file from that other package. Now, the user will see this message if the file exists anywhere inside their workspace instead of just in the default package directory. Thoughts?

PR #1508, opened 2019-07-24
Org Browser labels

lcampos:
### What does this PR do?
Makes sure the Org Browser displays the object labels customers are familiar with instead of the object's metadata name.

@W-6354518@


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -264,5 +264,192 @@ export const messages = {
   error_fetching_metadata: 'Error fetching metadata for org.',
   error_org_browser_text:
     'Run "SFDX: Authorize an Org" to authorize your org again.',
-  force_list_metadata: 'SFDX: Force List Metadata'
+  force_list_metadata: 'SFDX: Force List Metadata',
+
+  AccessControlPolicy: 'Access Control Policies',
```
brpowell: So friendly 😎

PR #1521, opened 2019-07-31
[Org Browser] Handle errors not in JSON format

AnanyaJha:
### What does this PR do?
This PR adds functionality to 'sanitize' any errors that may get thrown before displaying them to the user. It also adds logging for the describeMetadata and listMetadata calls.

@W-6383707@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
@@ -36,10 +36,20 @@ export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<
 export async function forceDescribeMetadata(
   outputFolder: string
 ): Promise<string> {
+  const startTime = process.hrtime();
+  const forceDescribeMetadataExecutor = new ForceDescribeMetadataExecutor();
   const execution = new CliCommandExecutor(
-    new ForceDescribeMetadataExecutor().build({}),
+    forceDescribeMetadataExecutor.build({}),
     { cwd: getRootWorkspacePath() }
   ).execute();
+
+  execution.processExitSubject.subscribe(() => {
+    forceDescribeMetadataExecutor.logMetric(
+      execution.command.logName,
+      startTime
+    );
+  });
+
   if (!fs.existsSync(outputFolder)) {
     mkdir('-p', outputFolder);
   }
--- a/packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
@@ -28,7 +28,7 @@ export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<
     return new SfdxCommandBuilder()
       .withArg('force:mdapi:describemetadata')
       .withJson()
-      .withLogName('force_describe_metadata')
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -137,7 +137,11 @@ export class MetadataOutlineProvider
 }
 
 function parseErrors(error: string): Error {
-  const e = JSON.parse(error);
+  const sanitized = error.substring(
```
lcampos: let's add a test for this scenario.

packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
```diff
@@ -28,18 +28,28 @@ export class ForceDescribeMetadataExecutor extends SfdxCommandletExecutor<
...
 ): Promise<string> {
+  const startTime = process.hrtime();
+  const forceDescribeMetadataExecutor = new ForceDescribeMetadataExecutor();
   const execution = new CliCommandExecutor(
-    new ForceDescribeMetadataExecutor().build({}),
+    forceDescribeMetadataExecutor.build({}),
     { cwd: getRootWorkspacePath() }
   ).execute();
+
+  execution.processExitSubject.subscribe(() => {
```
brpowell: I don't believe this needs to be added since `execute()` isn't being overridden in the ForceDescribeMetadataExecutor, and the parent `execute()` takes care of logging the execution event:

in SfdxCommandletExecutor:
```
public execute(response: ContinueResponse<T>): void {
    const startTime = process.hrtime();
      cwd: getRootWorkspacePath()

    execution.processExitSubject.subscribe(() => {
      this.logMetric(execution.command.logName, startTime);
    });
}
```

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
brpowell: I might suggest putting this JSON sanitization snippet into a util since we're doing this anytime we collect JSON output from a CLI command. See `orgCreateResultParser.ts` and `deployResultParser.ts`

packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
brpowell: Oh you're not using the commandlet i see

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
AnanyaJha: Yup, we've logged another story to move this code into a util once Jonny finishes his change (which also uses the sanitization) as well.

packages/salesforcedx-vscode-core/src/commands/forceDescribeMetadata.ts
AnanyaJha: Yeah we can't use the SfdxCommandletExecutor when we need to collect info from the cli output, which means metrics aren't being reported. We've logged another bug to add this `logMetric` snippet to the other commands missing the telemetry as well.

lcampos:
@AnanyaJha you'll have to port it to the release branch

PR #1540, opened 2019-08-06
Changelog for release 46.9.0

praksb:
### What does this PR do?
Update changelog for relase 46.9.0

### DIFF
--- a/packages/salesforcedx-vscode/CHANGELOG.md
+++ b/packages/salesforcedx-vscode/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 #### salesforcedx-utils-vscode
 
-- Use VS Code api to open browser in host machine when operating in container mode ([PR #1448](https://github.com/forcedotcom/salesforcedx-vscode/pull/1448))
+- Use VS Code API to open browser in the host machine when operating in headless environment ([PR #1448](https://github.com/forcedotcom/salesforcedx-vscode/pull/1448))
 
 #### docs
 
-- Add Japanese translations for docs ([PR #1528](https://github.com/forcedotcom/salesforcedx-vscode/pull/1528))
+- Include Japanese translation for Salesforce Extensions for VS Code documentation ([PR #1528](https://github.com/forcedotcom/salesforcedx-vscode/pull/1528)) ([Japanese documentation](https://github.com/forcedotcom/salesforcedx-vscode/tree/develop/docs/articles-jp))
 
 ## 46.8.0 - August 1, 2019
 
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+## 46.9.0 - August 8, 2019
+
+## Fixed
+
+#### salesforcedx-vscode-core
+
+- Process errors not in JSON format to make them user friendly before displaying ([PR #1521](https://github.com/forcedotcom/salesforcedx-vscode/pull/1521))
```
lcampos: This PR was released in 46.8.0 so we can remove it.

praksb: will do.

packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 46.9.0 - August 8, 2019
+
+## Added
+
+#### salesforcedx-utils-vscode
+
+- Use VS Code api to open browser in host machine when operating in container mode ([PR #1448](https://github.com/forcedotcom/salesforcedx-vscode/pull/1448))
```
hbc-bindu: @praksb, while reviewing a different message Nate mentioned that container refers to a very specific situation and he would prefer headless instead. Let me know if this looks ok:
Use VC Code API to open a browser in the host machine when operating in headless environment.

packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+#### docs
+
+- Add Japanese translations for docs ([PR #1528](https://github.com/forcedotcom/salesforcedx-vscode/pull/1528))
```
hbc-bindu: @praksb, Do we have a location where the Japanese version of the docs will be uploaded? In which case, could we include that link here?
Include Japanese translation for Salesforce Extensions for VS Code documentation <add link>

packages/salesforcedx-vscode/CHANGELOG.md
praksb: that sounds good to me. i'll update the comment.

packages/salesforcedx-vscode/CHANGELOG.md
praksb: I think this is the link. let me add it.
https://github.com/forcedotcom/salesforcedx-vscode/tree/develop/docs/articles-jp

PR #1541, opened 2019-08-06
Change test placeholders for security

jonnyhork:
### What does this PR do?
Change the test mock URLs to `www.example.com` for security purposes.
Nate Totten has reminded me that the only safe URL to use are reserved ones such as `example.com, net, org` to avoid an accidental visit to a malware site.

More info can be found here: https://www.iana.org/domains/reserved
This is not related to specific work item.

### DIFF
--- a/packages/salesforcedx-utils-vscode/test/unit/cli/orgOpenContainerResultParser.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/cli/orgOpenContainerResultParser.test.ts
@@ -23,6 +23,7 @@ describe('force:org:open container parser', () => {
       username: 'test@example.com'
     }
   };
+
   it('should parse success info successfully', () => {
     const parser = new OrgOpenContainerResultParser(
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-utils-vscode/test/unit/cli/orgOpenContainerResultParser.test.ts
```diff
@@ -15,16 +15,15 @@ import {
...
 describe('force:org:open container parser', () => {
+  const orgOpenSuccessResult: OrgOpenSuccessResult = {
+    status: 0,
+    result: {
+      orgId: '00Dxx0000000123',
+      url: 'www.example.com',
+      username: 'test@example.com'
+    }
+  };
   it('should parse success info successfully', () => {
```
vazexqi: [Nitpick] Put a newline between the `it(..)` methods to make them consistent. The spacing between the `it` clauses are not consistent. Prettier cannot help fix these kinds of formatting issues.

PR #1545, opened 2019-08-07
[Tech Debt] Investigate use of 'category' attribute for command grouping

cademay:
### What does this PR do?

This PR uses the "category" attribute in package.json in order to append the "SFDX:" prefix onto command palette items. This comes as an alternative to hard-coding all of the "SFDX:" prefixes in package.nls.json.

However, there is one potential drawback to uses this attribute. Consider the following line from the VS Code API docs: "The editor context menu, on the other hand, shows disabled items but doesn't show the category label."

The "editor context menu" is the menu that shows up on right-click. Commands can be included in this menu by means of "editor/context" property in the package.json file

Since the category labels do not show up for items in this menu, if we replace all of the hard-coded "SFDX" prefixes with "category": "SFDX" attributes, then the commands in the right-click menu will be without "SFDX:" prefix.

Assuming that this is not the desired outcome, a compromise could be to only hard-code "SFDX:" into the titles for the 5 commands in question (which is what this pull request displays). However, because the category attribute is used in package.json, while the titles are hard-coded into package.nls.json, this approach would result in the application of the SFDX prefix being spread across 2 different files. This could possibly lead to confusion in the future.

This applies to the following 5 commands:

      SFDX: Retrieve Source in Manifest from Org
      SFDX: Deploy Source in Manifest to Org
      SFDX: Delete This from Project and Org
      SFDX: Deploy This Source to Org
      SFDX: Retrieve This Source from Org

With this in mind, please let me know your thoughts on the use of the category attribute. Thank you 

@W-4670375@


lcampos:
Thanks for the detailed summary @cademay . Having the `SFDX` prefix in the commands across vscode is important for our user's experience. Given the options, having a more complex development experience for naming commands based on where they are surfaced versus one way of naming them regardless of where the command is displayed, I think we should close this spike/PR and keep the current practice of adding the prefix in the label.

AnanyaJha:
I really like the summary! I agree with Luis, introducing this change will make it more complex to add new commands in the future. I think the information you've found is still valuable though, so maybe you can transfer it to a Quip doc and add it to our spikes folder in Quip?

Issue #1394, opened 2019-06-12
Don't modify eslint path automatically, don't automatically change user settings


AndrewRayCode:
Every time I open VSCode this extension automatically adds `  "eslint.nodePath": "/Users/myuser/.vscode/extensions/salesforce.salesforcedx-vscode-lwc-45.15.1/node_modules"
` to my `settings.json` file.

Please stop auto modifying config files, this affects any VSCode session I open regardless of it's a Salesforce project. It happens without prompting. Your extensions have already fully deleted my local config settings in the past from trying to do things like this. Now you're auto modifying them? Please don't add behavior like this to plugins!

This breaks any project with custom ESLint plugins, because it tries to use a randomly installed one in the sfdx plugin, not the correct locally installed on per project.

For anyone else who has their config mangled by these bad actor plugins and starts seeing errors like this because it's using a randomly installed eslint:
```
[Error - 11:43:40 AM] Cannot find module 'eslint-plugin-prettier' Referenced from: /Users/me/repo/.eslintrc.js
```

You can block the bad plugin by setting it manually:
```
  "eslint.nodePath": "./node_modules"
```

lcampos:
Hi @AndrewRayCode, can you provide more info on how you are setting up the Salesforce extensions ? The issue you reference only happens when the Salesforce extensions get activated which is specific to the project being a Salesforce project. The `settings.json` that gets the eslint value is exclusively the one for the project.
We're working on modifying this behavior but would like to make sure we're not missing something else.

AndrewRayCode:
yikes

this is a serious issue, please re-open

bk-dev:
Our developers are facing this same problem. Please address.

ntotten:
We are looking at how we can address this. The problem is there isn’t a user specific project config for VSCode. The way we can not write this setting is to add a package.json to projects that have LWC and require developers to use npm. This however means we have to require them to install node.

bk-dev:
> We are looking at how we can address this. The problem is there isn’t a user specific project config for VSCode. The way we can not write this setting is to add a package.json to projects that have LWC and require developers to use npm. This however means we have to require them to install node.

I imagine the nuances of LWCs can lead to inadvertent developer errors which a linter can help remedy. However, linters have always been an optional project add-on. The benefits of a linter in this day and age is pretty well known. Do you guys have a custom ESLint plugin for LWCs?

ntotten:
[...]

PR #1556, opened 2019-08-12
[FIX] Don't override eslint.nodePath, just modify it

Bartheleway:
### What does this PR do?
This modify how the salesforce extension modify the eslint.nodePath workspace config. This setting can also be modified by the user so we have to take this into account.

This is a quick fix. As suggested in the issue, I think a rework is needed around that subject.

#1394

### Side note
I'm not used to typescript so feel free to make any comment.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/index.ts
+++ b/packages/salesforcedx-vscode-lwc/src/index.ts
@@ -152,21 +152,31 @@ function startLWCLanguageServer(context: ExtensionContext) {
   context.subscriptions.push(client);
 }
 
+
+// User has not set one, use the eslint bundled with our extension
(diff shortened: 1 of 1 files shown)


ntotten:
Thanks for the PR. I will discuss with the team, but I don't think this is the approach we will take. We're looking at removing this configuration entirely. The hope is we can get a PR accepted on the ESLint extension that will allow us to set the config of that extension temporarily so that we are not writing or overriding any user settings.

Bartheleway:
Hi, I agree this is not a complete fix. As stated in the issue, this is only a quick fix to give the team more time to remove this configuration entirely. It would be awesome to have it included in a minor fix quickly as it's an issue for my project and for other people as well.

ntotten:
We will be fixing this another way soon.

PR #1557, opened 2019-08-12
telemetry exceptions

lcampos:
### What does this PR do?
Updates the telemetry library to handle exceptions as a separate event.
This is exclusive to the core extension, I'll update the rest of the extensions in a separate PR.

@W-6449855@


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -55,13 +55,19 @@ export class SourcePathChecker implements PostconditionChecker<string> {
           return inputs;
         }
       } catch (error) {
-        telemetryService.sendError(error.name);
+        telemetryService.sendException(
+          'force_source_retrieve_with_sourcepath',
+          `Error while parsing oackage directories. ${error.message}`
```
AnanyaJha: *package

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -26,13 +26,14 @@ export async function getWorkspaceOrgType(
...
-    telemetryService.sendError(
+    telemetryService.sendException(
+      'get_workspace_org_type',
       'workspaceOrgType.getWorkspaceOrgType ran into an undefined username.'
     );
   }
 
   const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username!).catch(err =>
-    telemetryService.sendError(err)
+    telemetryService.sendException('get_workspace_org_type_scratch_org', err)
```
AnanyaJha: Is this sending the error object or the message?

lcampos: Oh right, let me update it to just send a message.

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
```diff
@@ -157,6 +159,32 @@ export default class TelemetryReporter extends vscode.Disposable {
     }
   }
 
+  public sendExceptionEvent(
+    exceptionName: string,
+    exceptionMessage: string,
+    measurements?: { [key: string]: number }
```
brpowell: Do we also want the option to send properties in addition to measurements?

lcampos: Eventually yes, that's available in the latest version of application insights but I wanted to keep the changes as small as possible for now. I also didn't see a current need for sending properties so I decided to leave updating the application insights version in a future PR.

PR #1560, opened 2019-08-14
Add retrieved types when retrieving components with org browser 

brpowell:
### What does this PR do?

Adding telemetry for org browser to gain insight into most retrieved types.  

W-6455428

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -337,7 +337,7 @@ export abstract class SfdxCommandletExecutor<T>
       output += realData.toString();
     });
 
-    execution.processExitSubject.subscribe(async exitCode => {
+    execution.processExitSubject.subscribe(exitCode => {
       this.logMetric(
         execution.command.logName,
         startTime,
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -341,7 +341,7 @@ export abstract class SfdxCommandletExecutor<T>
       this.logMetric(
         execution.command.logName,
         startTime,
-        await this.getTelemetryData(exitCode === 0, response, output)
+       this.getTelemetryData(exitCode === 0, response, output)
       );
     });
     this.attachExecution(execution, cancellationTokenSource, cancellationToken);
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -328,7 +328,8 @@ export abstract class SfdxCommandletExecutor<T>
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
     const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: getRootWorkspacePath()
+      cwd: getRootWorkspacePath(),
+      env: { SFDX_JSON_TO_STDOUT: 'true' }
     }).execute(cancellationToken);
 
     let output = '';
--- a/packages/salesforcedx-vscode-core/src/commands/commands.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/commands.ts
@@ -27,7 +27,7 @@ import { notificationService, ProgressNotification } from '../notifications';
 import { isSfdxProjectOpened } from '../predicates';
 import { SfdxPackageDirectories } from '../sfdxProject';
 import { taskViewService } from '../statuses';
-import { telemetryService } from '../telemetry';
+import { TelemetryData, telemetryService } from '../telemetry';
 import { getRootWorkspacePath, hasRootWorkspace, OrgAuthInfo } from '../util';
 
 export class CompositePreconditionChecker implements PreconditionChecker {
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -331,12 +331,29 @@ export abstract class SfdxCommandletExecutor<T>
...
+      this.logMetric(
+        execution.command.logName,
+        startTime,
+        await this.getTelemetryData(exitCode === 0, response, output)
+      );
     });
     this.attachExecution(execution, cancellationTokenSource, cancellationToken);
   }
 
+  protected getTelemetryData(
```
brpowell: This method can be overridden in child classes to add additional data to a command execution

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
```diff
@@ -39,20 +39,33 @@ import {
...
+    this.typeName = typeName;
+    this.componentName = componentName;
   }
 
   public build(): Command {
     return new SfdxCommandBuilder()
       .withDescription(nls.localize('force_source_retrieve_text'))
       .withArg('force:source:retrieve')
-      .withFlag('-m', this.metadataArg)
+      .withFlag('-m', `${this.typeName}:${this.componentName}`)
```
brpowell: I changed this so I could access just the type name and not the whole string we use for the command

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
```diff
@@ -39,20 +39,33 @@ import {
...
       .withLogName('force_source_retrieve')
       .build();
   }
+
+  protected getTelemetryData(): TelemetryData {
+    const retrievedTypes: any = {};
+    retrievedTypes[this.typeName] = 1;
+    return {
+      properties: {
+        'org-browser/retrievedTypes': JSON.stringify(retrievedTypes)
```
brpowell: 1. I'm trying to future proof this so when we add the ability to retrieve multiple components, we know all the types and the number of components for each that was retrieved. This will require some query building kung fu to parse the json.

2. I'm thinking about adding a prefix (in this case **org-browser**) to properties so we have a better idea of what they are used for in dashboards, since we're starting to have a decent amount of them. Any strong feelings on this?

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -331,12 +331,29 @@ export abstract class SfdxCommandletExecutor<T>
       cwd: getRootWorkspacePath()
```
lcampos: You'll need to add `env: { SFDX_JSON_TO_STDOUT: 'true' }` here for the changes in line 335 to work in all scenarios.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -331,12 +331,29 @@ export abstract class SfdxCommandletExecutor<T>
...
+    let output = '';
+    execution.stdoutSubject.subscribe(realData => {
+      output += realData.toString();
+    });
+
+    execution.processExitSubject.subscribe(async exitCode => {
+      this.logMetric(
+        execution.command.logName,
+        startTime,
+        await this.getTelemetryData(exitCode === 0, response, output)
```
lcampos: This change will affect all commands that are not implementing their own `execute` method. Is the data for those commands going to look the same as it currently does ?

brpowell: Correct, it will look the same. This is because those commands will use the default getTelemetryData which returns undefined for additional data.

lcampos: I think we need to remove the `await` in `this.getTelemetryData` and `async` in line 340 since we're not calling any promises.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
AnanyaJha: Looks good to me, we should keep the prefix.

packages/salesforcedx-vscode-core/src/commands/commands.ts
brpowell: This has `await` in case an overridden version requires some asynchronous operation. Awaiting something that's not a promise will just convert the value to a resolved promise.

lcampos: I prefer having this as a sync operation for now and update it to be an async one whenever we have a need for it.

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -328,15 +328,33 @@ export abstract class SfdxCommandletExecutor<T>
...
     }).execute(cancellationToken);
 
-    execution.processExitSubject.subscribe(() => {
-      this.logMetric(execution.command.logName, startTime);
+    let output = '';
+    execution.stdoutSubject.subscribe(realData => {
+      output += realData.toString();
+    });
+
+    execution.processExitSubject.subscribe(async exitCode => {
```
lcampos:
```suggestion
    execution.processExitSubject.subscribe(exitCode => {
```
@brpowell

packages/salesforcedx-vscode-core/src/commands/commands.ts
```diff
@@ -328,15 +328,33 @@ export abstract class SfdxCommandletExecutor<T>
...
+      this.logMetric(
+        execution.command.logName,
+        startTime,
+        await this.getTelemetryData(exitCode === 0, response, output)
```
lcampos:
```suggestion
       this.getTelemetryData(exitCode === 0, response, output)
```
@brpowell

PR #1562, opened 2019-08-14
update auth web login and auth dev hub to conditionally use device lo…

dehru:
This PR changes how Authorize A Dev Hub and Authorize An Org behave when running in Docker ( or any environment with SFDX_CONTAINER_MODE set ).  Those commands will now generate force:auth:device:login instead of force:auth:web:login.  For most users there will be no change.  But for VSCode Remote Development, they will now be able to authenticate using Salesforce OAuth 2 Device Flow.

### What does this PR do?
This PR enables support for VSCode Remote Development in a container.  It aligns with recent changes to the sfdx CLI to make this new force:auth:device:login command available.

@W-6351129@

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ packages/dbaeumer*
 
 # Netlify
 .netlify
+
+# Vscode History Extension
+.history
--- a/packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
@@ -10,6 +10,7 @@ import {
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 
+import { isSFDXContainerMode } from '../util';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
@@ -23,14 +24,18 @@ import { isDemoMode } from '../modes/demo-mode';
 
 export class ForceAuthDevHubExecutor extends SfdxCommandletExecutor<{}> {
   public build(data: {}): Command {
-    return new SfdxCommandBuilder()
-      .withDescription(
-        nls.localize('force_auth_web_login_authorize_dev_hub_text')
-      )
-      .withArg('force:auth:web:login')
-      .withArg('--setdefaultdevhubusername')
-      .withLogName('force_auth_dev_hub')
-      .build();
+    const command = new SfdxCommandBuilder().withDescription(
+      nls.localize('force_auth_web_login_authorize_dev_hub_text')
+    );
+    if (isSFDXContainerMode()) {
+      command
+        .withArg('force:auth:device:login')
+        .withLogName('force_auth_device_dev_hub');
+    } else {
+      command.withArg('force:auth:web:login').withLogName('force_auth_dev_hub');
+    }
+    command.withArg('--setdefaultdevhubusername');
+    return command.build();
   }
 }
 
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
@@ -40,25 +40,57 @@ describe('Force Auth Web Login For Dev Hub in Demo  Mode', () => {
   });
 });
 
-describe('Command chosen is based on results of isDemoMode()', () => {
-  let originalValue: any;
+describe('Force Auth Dev Hub is based on environment variables', () => {
+  describe('in demo mode', () => {
+    let originalValue: any;
 
-  beforeEach(() => {
-    originalValue = process.env.SFDX_ENV;
+    beforeEach(() => {
+      originalValue = process.env.SFDX_ENV;
+    });
+
+    afterEach(() => {
+      process.env.SFXD_ENV = originalValue;
+    });
+
+    it('Should use ForceAuthDevHubDemoModeExecutor if demo mode is true', () => {
+      process.env.SFDX_ENV = 'DEMO';
+      expect(createExecutor() instanceof ForceAuthDevHubDemoModeExecutor).to.be
+        .true;
+    });
+
+    it('Should use ForceAuthDevHubExecutor if demo mode is false', () => {
+      process.env.SFDX_ENV = '';
+      expect(createExecutor() instanceof ForceAuthDevHubExecutor).to.be.true;
+    });
   });
 
-  afterEach(() => {
-    process.env.SFXD_ENV = originalValue;
-  });
+  describe('in container mode', () => {
+    afterEach(() => {
+      delete process.env.SFDX_CONTAINER_MODE;
+    });
+    it('Should use force:auth:web:login when container mode is not defined', () => {
+      const authWebLogin = new ForceAuthDevHubExecutor();
+      const authWebLoginCommand = authWebLogin.build({});
+      expect(authWebLoginCommand.toCommand()).to.equal(
+        'sfdx force:auth:web:login --setdefaultdevhubusername'
+      );
+    });
+    it('Should use force:auth:web:login when container mode is empty', () => {
+      process.env.SFDX_CONTAINER_MODE = '';
+      const authWebLogin = new ForceAuthDevHubExecutor();
+      const authWebLoginCommand = authWebLogin.build({});
+      expect(authWebLoginCommand.toCommand()).to.equal(
+        'sfdx force:auth:web:login --setdefaultdevhubusername'
+      );
+    });
 
-  it('Should use ForceAuthDevHubDemoModeExecutor if demo mode is true', () => {
-    process.env.SFDX_ENV = 'DEMO';
-    expect(createExecutor() instanceof ForceAuthDevHubDemoModeExecutor).to.be
-      .true;
-  });
-
-  it('Should use ForceAuthDevHubExecutor if demo mode is false', () => {
-    process.env.SFDX_ENV = '';
-    expect(createExecutor() instanceof ForceAuthDevHubExecutor).to.be.true;
+    it('Should use force:auth:device:login when container mode is defined', () => {
+      process.env.SFDX_CONTAINER_MODE = 'pickles';
+      const authWebLogin = new ForceAuthDevHubExecutor();
+      const authWebLoginCommand = authWebLogin.build({});
+      expect(authWebLoginCommand.toCommand()).to.equal(
+        'sfdx force:auth:device:login --setdefaultdevhubusername'
+      );
+    });
   });
 });
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthWebLogin.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthWebLogin.test.ts
@@ -9,6 +9,7 @@ import { expect } from 'chai';
 import * as sinon from 'sinon';
 import * as vscode from 'vscode';
 import {
+  AuthParams,
   AuthParamsGatherer,
   createExecutor,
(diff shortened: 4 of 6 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceOrgOpen.test.ts
```diff
@@ -38,17 +38,13 @@ describe('Force Org Open', () => {
   });
 
   describe('Executor is chosen based on environment', () => {
-    let originalEnvValue: string;
-
-    beforeEach(() => {
-      originalEnvValue = process.env.SFDX_CONTAINER_MODE as string;
```
dehru: Jonny, i found that there was some side-effects from us trying to restore the original value.  It took me a while to figure it out, but by the time my tests ran process.env.SFDX_CONTAINER_MODE became the string 'undefined' instead of the value undefined.  I don't know exactly how or why, but yeah, that happened...  So now we're just going to delete it so there are definitely no side-effects.

.gitignore
```diff
@@ -105,4 +105,7 @@ packages/dbaeumer*
 **/.DS_Store
 
 # Netlify
-.netlify
\ No newline at end of file
+.netlify
+
+# Vscode History Extension
+.history
```
vazexqi: Is this the app https://marketplace.visualstudio.com/items?itemName=xyz.local-history

Would you recommend it? Should it be part of the .vscode/extensions.json?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthWebLogin.test.ts
```diff
@@ -227,3 +228,40 @@ describe('Command chosen is based on results of isDemoMode()', () => {
...
+    const authWebLogin = new ForceAuthWebLoginExecutor();
+    const authWebLoginCommand = authWebLogin.build(
+      ({} as unknown) as AuthParams
+    );
+    expect(authWebLoginCommand.toCommand()).to.equal(
+      'sfdx force:auth:web:login --setalias  --instanceurl  --setdefaultusername'
+    );
+  });
+
+  it('and use force:auth:web:login when container mode is empty', () => {
```
vazexqi: I don't think we use `and` to denote the `it(...)`.

I'm open to discuss this but it's better to follow the established convention of a file.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthWebLogin.test.ts
```diff
@@ -227,3 +228,40 @@ describe('Command chosen is based on results of isDemoMode()', () => {
...
+    const authWebLogin = new ForceAuthWebLoginExecutor();
+    const authWebLoginCommand = authWebLogin.build(
+      ({} as unknown) as AuthParams
+    );
+    expect(authWebLoginCommand.toCommand()).to.equal(
+      'sfdx force:auth:web:login --setalias  --instanceurl  --setdefaultusername'
+    );
+  });
+
+  it('and use force:auth:device:login when container mode is defined', () => {
```
vazexqi: Ditto.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthWebLogin.test.ts
```diff
@@ -227,3 +228,40 @@ describe('Command chosen is based on results of isDemoMode()', () => {
     expect(createExecutor() instanceof ForceAuthWebLoginExecutor).to.be.true;
   });
 });
+
+describe('Should work correctly in container mode', () => {
+  afterEach(() => {
+    delete process.env.SFDX_CONTAINER_MODE;
+  });
+  it('and use force:auth:web:login when container mode is not defined', () => {
```
vazexqi: I don't think we use `and` to denote the `it(...)`.

I'm open to discuss this but it's better to follow the established convention of a file.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -61,4 +61,34 @@ describe('Command chosen is based on results of isDemoMode()', () => {
     process.env.SFDX_ENV = '';
     expect(createExecutor() instanceof ForceAuthDevHubExecutor).to.be.true;
   });
+
+  describe('Should work correctly in container mode', () => {
+    afterEach(() => {
+      delete process.env.SFDX_CONTAINER_MODE;
+    });
+    it('and use force:auth:web:login when container mode is not defined', () => {
```
vazexqi: I don't think we use `and` to denote the `it(...)`.

I'm open to discuss this but it's better to follow the established convention of a file.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -61,4 +61,34 @@ describe('Command chosen is based on results of isDemoMode()', () => {
...
+      const authWebLogin = new ForceAuthDevHubExecutor();
+      const authWebLoginCommand = authWebLogin.build({});
+      expect(authWebLoginCommand.toCommand()).to.equal(
+        'sfdx force:auth:web:login --setdefaultdevhubusername'
+      );
+    });
+    it('and use force:auth:web:login when container mode is empty', () => {
```
vazexqi: Ditto.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -61,4 +61,34 @@ describe('Command chosen is based on results of isDemoMode()', () => {
...
+      process.env.SFDX_CONTAINER_MODE = '';
+      const authWebLogin = new ForceAuthDevHubExecutor();
+      const authWebLoginCommand = authWebLogin.build({});
+      expect(authWebLoginCommand.toCommand()).to.equal(
+        'sfdx force:auth:web:login --setdefaultdevhubusername'
+      );
+    });
+
+    it('and use force:auth:device:login when container mode is defined', () => {
```
vazexqi: Ditto.

.gitignore
dehru: I'm still experimenting.  I loved WebStorm's built in local history.  It saved my butt quite a few times when I royally messed up git

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
dehru: If you see these nested statements when running npm test it makes more sense....we're kinda building up broken sentences.  At least that's what i've done with testing in the past like this.
```
Command chosen is based on results of isDemoMode()
    ✓ Should use ForceAuthDevHubDemoModeExecutor if demo mode is true
    ✓ Should use ForceAuthDevHubExecutor if demo mode is false
    Should work correctly in container mode
      ✓ and use force:auth:web:login when container mode is not defined
      ✓ and use force:auth:device:login when container mode is defined
```
I'll admit that I should adjust the top level statement.  So that the test output reads like.
```
The command that is run
    should handle isDemoMode()
    ✓ and use ForceAuthDevHubDemoModeExecutor if demo mode is true
    ✓ and use ForceAuthDevHubExecutor if demo mode is false
    should work correctly in container mode
      ✓ and use force:auth:web:login when container mode is not defined
      ✓ and use force:auth:device:login when container mode is defined
```

But to fit in, I'll try to match the current implementation.

vazexqi: @lcampos Thoughts?

jonnyhork: IMO, I can see how the `it(...)` statements are supposed to fit together to complete a sentence and I like the idea of continuity when running all the tests together. However I just don't often read the output of the mocha runner form top to bottom like a complete paragraph or sentence but more like a list of bullet points and zero in on a single line or test.
With that in mind I would like to be able to understand what failed to work in as few words as possible.

lcampos: The current naming convention follows Spec-style testing, which originated from Behavior Driven Development (BDD). When you write a test you want to specify the behavior of something.

The purpose of `describe` is to group tests that are common by certain criteria and also allows for nesting depending on how specific you want to get with the grouping strategy. In this case I'd say the use of `Should do x ...` in the describe does not provide as much grouping value as something like `In container mode`.

In the spirit of specifying the behavior of something and following BDD, we are using `it` over the `test` interface since we want to describe how the feature should work. If you are reading the test you read it as `it should ...`.

A good example of what I just mentioned are the telemetry tests. The describes are used as ways of grouping tests and the it clauses contain messages that correspond to the grouping they're part of and the expected outcome.

```
  Telemetry 
    in dev mode
      ✓ Should not initialize telemetry reporter
      ✓ Should show telemetry info message
      ✓ Should not show telemetry info message

    production mode
      ✓ Should show telemetry info message
      ✓ Should not show telemetry info message
      ✓ Should send telemetry data
```

This convention only makes sense if you are following BDD practices, which we've standardized on and that's why our mocha runner is configured to only allow bdd interface usage (https://mochajs.org/#interfaces). Let's keep using the current conventions.

packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
```diff
@@ -23,14 +24,18 @@ import { isDemoMode } from '../modes/demo-mode';
...
+    if (isSFDXContainerMode()) {
+      command.withArg('force:auth:device:login');
+    } else {
+      command.withArg('force:auth:web:login');
+    }
+    command
       .withArg('--setdefaultdevhubusername')
-      .withLogName('force_auth_dev_hub')
-      .build();
+      .withLogName('force_auth_dev_hub');
```
lcampos: We need to send a different log name based on mode or we won't be able to use the metrics to see which mode was used for the command's execution.

cc @ntotten

dehru: I'll update for both forceAuthDevHub and forceAuthWebLogin.  I'll leave the current ones unchanged, and invent 2 new ones.  

```
// Authorize a dev hub ( existing log name )
'force_auth_dev_hub'
// Authorize a dev hub ( new log name for device flow )
'force_auth_device_dev_hub'
```
and
```
// Authorize an org ( existing log name - no change )
'force_auth_web_login'
// Authorize an org via device flow ( new log name )
'force_auth_device_login'
```

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
dehru: Cool, thanks for that example @lcampos.  Here's where I ended up.

Force Auth Web Login is based on environment variables
    in demo mode
      ✓ Should use ForceAuthDevHubDemoModeExecutor if demo mode is true
      ✓ Should use ForceAuthDevHubExecutor if demo mode is false
    in container mode
      ✓ Should use force:auth:web:login when container mode is not defined
      ✓ Should use force:auth:web:login when container mode is empty
      ✓ Should use force:auth:device:login when container mode is defined

packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
vazexqi: @dehru - I am going to add you to the appinsights that we have for Platform Dev Tools so you can see how this works underneath. It's easier when you are able to explore the data and see how the names affect the counts.

PR #1568, opened 2019-08-16
Allow diffing local metadata against an org

lcampos:
### What does this PR do?
Exposes a new menu command to trigger a diff of the currently opened file against the org the user has authenticated to.

FYI: I'm still working out the tests but this part is ready for review.

diff_demo

@W-6393758@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { extractJsonObject } from '../helpers';
+
+export interface DiffSuccessResponse {
+  status: number;
+  result: {
+    remote: string;
+    local: string;
+    fileName: string;
+  };
+}
+
+export interface DiffErrorResponse {
+  commandName: string;
+  exitCode: number;
+  message: string;
+  name: string;
+  stack: string;
+  status: number;
+  warnings: any[];
+}
+
+export class DiffResultParser {
+  private response: any;
+
+  constructor(stdout: string) {
+    try {
+      this.response = extractJsonObject(stdout);
+    } catch (e) {
+      const err = new Error('Error parsing diff result');
+      err.name = 'DiffResultParserFail';
+      throw err;
+    }
+  }
+
+  public isSuccessful(): boolean {
+    return this.response && this.response.status === 0 ? true : false;
+  }
+
+  public getSuccessResponse(): DiffSuccessResponse | undefined {
+    if (this.isSuccessful()) {
+      return this.response as DiffSuccessResponse;
+    }
+  }
+
+  public getErrorResponse(): DiffErrorResponse | undefined {
+    if (!this.isSuccessful()) {
+      return this.response as DiffErrorResponse;
+    }
+  }
+}
--- a/packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
@@ -7,22 +7,64 @@
 
 import { extractJsonObject } from '../helpers';
 
+/**
+ * Interface that stores information about a successful diff operation
+ */
 export interface DiffSuccessResponse {
+  /**
+   * Request status e.g. 0 = Success
+   */
   status: number;
+  /**
+   * Object that contains the successful result
+   */
   result: {
+    /**
+     * Provides the location to the cached file retrieved from the org
+     */
     remote: string;
+    /**
+     * Provides the location to the file provided by the user in --sourcepath
+     */
     local: string;
+    /**
+     * Name of the file being diffed
+     */
     fileName: string;
   };
 }
 
+/**
+ * Interface that stores information about an unsuccessful diff operation
+ */
 export interface DiffErrorResponse {
+  /**
+   * Name of the command that was executed e.g. Diff
+   */
   commandName: string;
+  /**
+   * Exit code provided by the command e.g. 1
+   */
   exitCode: number;
+  /**
+   * Error message provided by the command e.g. The path could not be found.
+   */
   message: string;
+  /**
+   * Error name
+   */
   name: string;
+  /**
+   * Stack trace for the current error
+   */
   stack: string;
+  /**
+   * Request status e.g. 1 = Error
+   */
   status: number;
+  /**
+   * Array of warnings provided by the command
+   */
   warnings: any[];
 }
 
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -146,7 +146,15 @@
         },
         {
           "command": "sfdx.force.diff",
-          "when": "sfdx:project_opened"
+          "when": "sfdx:project_opened && resourceFilename =~ /.*(app|cmp|auradoc|evt|intf|css|design|Controller\\.js|Renderer\\.js|Helper\\.js)$/"
+        },
+        {
+          "command": "sfdx.force.diff",
+          "when": "sfdx:project_opened && resourceLangId == apex"
+        },
+        {
+          "command": "sfdx.force.diff",
+          "when": "sfdx:project_opened && resourceLangId == visualforce"
         }
       ],
       "explorer/context": [
@@ -673,7 +681,7 @@
       },
       {
         "command": "sfdx.force.diff",
-        "title": "SFDX: Diff File Against Org"
+        "title": "%force_diff_against_org%"
       }
     ],
     "configuration": {
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -143,6 +143,10 @@
         {
           "command": "sfdx.force.source.delete.current.file",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest'"
+        },
+        {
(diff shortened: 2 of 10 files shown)


packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -0,0 +1,159 @@
...
+    CancelResponse | ContinueResponse<{ filePath: string }>
+  > {
+    return { type: 'CONTINUE', data: { filePath: this.sourcePath } };
+  }
+}
+
+export async function forceSourceDiff(sourceUri: vscode.Uri) {
+  if (sourceUri) {
+    const editor = vscode.window.activeTextEditor;
+    if (
```
lcampos: We'll initially have a limited set of supported files, I'd like to know if there's a better way of limiting/checking for the supported types. I tried updating the `where` clause at the command level but that was getting a bit too complex and there's no easy way to write a test to verify the logic. Any ideas ?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -666,6 +670,10 @@
           "light": "resources/light/retrieve.svg",
           "dark": "resources/dark/retrieve.svg"
         }
+      },
+      {
+        "command": "sfdx.force.diff",
+        "title": "SFDX: Diff File Against Org"
```
lcampos: I'll move this to a label in `package.nls.json`

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    execution.processExitSubject.subscribe(async () => {
+      this.logMetric(execution.command.logName, startTime);
+
+      try {
+        const diffParser = new DiffResultParser(stdOut);
+        const diffParserSuccess = diffParser.getSuccessResponse();
+        const diffParserError = diffParser.getErrorResponse();
+
+        if (diffParser.isSuccessful() && diffParserSuccess) {
```
AnanyaJha: do we need to check whether it's successful and whether diffParserSuccess is defined?

lcampos: Yes because of how I've implemented the parser methods where `diffParserSuccess` can be `undefined`. I'm open to changing the parser if this seems like over doing checks.

brpowell: It could probably just be `if (diffParserSuccess)` since `isSuccessful()` isn't used elsewhere

packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
```diff
@@ -0,0 +1,57 @@
...
+export class DiffResultParser {
+  private response: any;
+
+  constructor(stdout: string) {
+    try {
+      this.response = extractJsonObject(stdout);
+    } catch (e) {
+      const err = new Error('Error parsing diff result');
+      err.name = 'DiffResultParserFail';
+      throw err;
```
brpowell: Do we want to throw an error that warns them if they don't have a more recent enough version of the CLI that bundles the diff plugin?

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -0,0 +1,159 @@
...
+      cancellationToken
+    );
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+
+export class SourcePathGatherer
```
brpowell: [FilePathGatherer](https://github.com/forcedotcom/salesforcedx-vscode/blob/b5af214c2421f6c916f5e11d2140ffdbec62f067/packages/salesforcedx-vscode-core/src/commands/commands.ts#L115) already exists in `commands.ts`

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
brpowell: Doesn't the CLI command already handle this?

AnanyaJha: Are you referring to when the diff command is exposed in the drop down menu?

packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
lcampos: Yes but maybe later, initially we will release it as an additional plugin that needs to be installed on top of the cli.

packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { extractJsonObject } from '../helpers';
+
+export interface DiffSuccessResponse {
```
sfsholden: I think it'd be good to add some documentation here, like what's currently in MetadataInfo.

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -0,0 +1,137 @@
...
+    const commandBuilder = new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_source_diff_text'))
+      .withArg('force:source:diff')
+      .withLogName('force_source_diff')
+      .withFlag('--sourcepath', filePath)
+      .withJson();
+    return commandBuilder.build();
+  }
+
+  public async execute(response: ContinueResponse<string>): Promise<void> {
```
sfsholden: Can we break this down a little bit to simplify this method?

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
sfsholden: Could we add this check to the diff plugin itself? We could add a new command for checking if the type is supported. That would save us from being in a situation where the two code bases get out of sync in case we add additional types or remove types.

packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
lcampos: That's a good idea, let me add something.

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
lcampos: I could move out the contents of `execution.processExitSubject.subscribe ...` but the rest I can't since we're overriding the `execute` provided by `SfdxCommandletExecutor`.

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
lcampos: I modified the first implementation of this. In the current version, I'm addressing the scenario where the user has added a keyboard shortcut to the diff command. In that scenario, sourceUri is always null so we need to see if we can get a file uri from the currently open file, if any.

I agree this could get out of sync with the diff plugin but that would be on us to make sure we coordinate the updates/releases instead of leaving a gap where the user cannot use the command in the keyboard shortcut scenario.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -143,6 +143,18 @@
         {
           "command": "sfdx.force.source.delete.current.file",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest'"
+        },
+        {
+          "command": "sfdx.force.diff",
```
lcampos: I added this so we show the diff command exclusively in the supported file types/languages.

brpowell: Nice. Another option later could be to set a context based on the open file whether it's diffable or not, if you don't want the large regex.

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
brpowell: Is this thread obsolete with these new changes? https://github.com/forcedotcom/salesforcedx-vscode/pull/1568/files/a7a1de7cc9154bc77538ec46ce6b6f3c86b477bb#r315284559

lcampos: Only the first 3 comments, @sfsholden 's comment is still relevant.

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export async function handleDiffResponse(stdOut: string) {
+  try {
+    const diffParser = new DiffResultParser(stdOut);
+    const diffParserSuccess = diffParser.getSuccessResponse();
+    const diffParserError = diffParser.getErrorResponse();
+
+    if (diffParserSuccess) {
+      const diffResult = diffParserSuccess.result;
+      const remote = vscode.Uri.file(diffResult.remote);
```
lcampos: Updated this from `vscode.Uri.parse` to `vscode.Uri.file` since it was throwing the following error `Error: BAD uri lacks scheme, falling back to file-scheme.`. This was because I was providing a file path as a string and without a scheme e.g. `User/.../fileName.cls` instead of `file:///User/.../fileName.cls`.

packages/salesforcedx-utils-vscode/src/cli/diffResultParser.ts
```diff
@@ -0,0 +1,99 @@
+/*
```
AnanyaJha: Descriptions look good!

Issue #1542, opened 2019-08-07
Retrieve Standard Objects in Org Browser


banderson5144:
### Summary

Retrieve standard objects for Org Browser

### Steps To Reproduce:

1.  Refresh list of CustomObject metadata type

### Expected result

Should see list of Standard and Custom objects

### Actual result

Just list of Custom Objects

### Additional information

**VS Code Version**: 1.36.1

**SFDX CLI Version**: 7.15.3

**OS and version**: Windows 10

suraj07081995:
Run this code to see the list of Standard Objects and Custom Objects
`Map<string,Schema.sObjectType> standardObj = Schema.getGlobalDescribe();`
`for(string str : standardObj.keyset() ){`
`if(!str.contains('__')){`
     `System.debug('Standard Object ==>>'+str);`
  `}`
`}`
`for(string str : standardObj.keyset() ){`
`if(str.contains('__')){`
  `System.debug('Custom Object ==>>'+str);`
`}`
`}`
I hope this helps you!

jvaloto:
If you want to retrieve this components in metadata, you have to add this to your package.xml (account e.g):

```
<types>
      <members>Account</members>
      <name>CustomObject</name>
   </types>
```

brpowell:
A fix for this was merged in #1571 and addressed in today's release 46.11.0

PR #1571, opened 2019-08-16
Fix not being able to retrieve standard objects in org browser

brpowell:
### What does this PR do?

Due to how we were filtering components based on their manageability, we were accidentally omitting standard objects from appearing under CustomObjects. This fixes the filtering condition.

#1542 
W-6467588


banderson5144:
When will this be released?

PR #1580, opened 2019-08-22
source diff docs & error handling

lcampos:
### What does this PR do?
Adds a source diff document to the public docs, enables diffing lwc components and handles the scenario in which a user triggers the source diff command and has not installed the `@salesforce/sfdx-diff` plugin.


docs/_data/sidebar.yml
```diff
@@ -32,6 +32,8 @@
       url: /articles/user-guide/default-org
     - title: Org Browser
       url: /articles/user-guide/org-browser
+      - title: Source Diff
```
sfsholden: Indentation is slightly off for this line

docs/articles/user-guide/source-diff.md
```diff
@@ -0,0 +1,33 @@
...
+1. In your terminal, run `sfdx plugins:install @salesforce/sfdx-diff`
+1. Once the install is complete you should see `@salesforce/sfdx-diff` as part of your installed plugins when running `sfdx plugins`
+
+## Usage
+
+Starting in version `46.11.0` you'll now get a new menu option `SFDX: Diff File Against Org` when right-clicking on an open metadata file.
+
+![Source Diff command](../../images/source_diff.png)
+
+Because source diff is in beta, there is a limited set of metadata supported. The list of supported metadata is the following:
```
sfsholden: I personally would update this 'The list of supported metadata is the following:' to something like this 'Please see below for the supported metadata list:'
@hbc-bindu What do you think?

hbc-bindu: I would suggest:
Source Diff currently supports the following list of metadata:
or
Source Diff currently supports a limited set of metadata:

PR #1606, opened 2019-08-27
Refactor sobject refresh

lcampos:
### What does this PR do?
Refactor sobject refresh code in order to fix some logic issues and write faster, more detailed tests.

I'm still sorting out some error handling and some tests but most of the work is ready to be reviewed.

@W-6425138@


packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -9,6 +9,7 @@
     "vscode": "^1.32.0"
   },
   "dependencies": {
+    "@salesforce/core": "2.0.1",
```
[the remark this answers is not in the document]
AnanyaJha: Were you able to understand why it breaks those extensions?

AnanyaJha:
The changes look good so far

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -244,50 +214,60 @@ export class SObjectDescribe {
...
-      this.getVersion(),
-      this.batchPart
-    ];
-    const batchRequestUrl = batchUrlElements.join('/');
-    const options: XHROptions = {
+
+    return batchRequest;
+  }
+
+  public buildXHROptions(types: string[], nextToProcess: number): XHROptions {
```
sfsholden: What does XHR stand for?

lcampos: XMLHttpRequest, we use these objects & library cause we're running describes directly against the org vs running it via the CLI. Here's more info on XHR https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest

packages/salesforcedx-sobjects-faux-generator/test/integration/fauxGenerate.test.ts
```diff
@@ -5,38 +5,31 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { SFDX_PROJECT_FILE } from '@salesforce/salesforcedx-utils-vscode/out/src';
+import { AuthInfo } from '@salesforce/core';
```
lcampos: With these changes, we do not create a repo, create a scratch org, run describes against it, etc. For now, these tests assume we always get the same response shape from the org and we make sure that our code works as expected.

I'll be adding a system test that runs queries against an org to make sure we catch any api breaking changes as part of a follow up story.

AnanyaJha: Good idea to add a system-test to catch any breaking changes

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -165,77 +170,42 @@ export enum SObjectCategory {
...
   public async describeGlobal(
     projectPath: string,
-    type: SObjectCategory,
-    username?: string
+    type: SObjectCategory
   ): Promise<string[]> {
-    const builder = new SfdxCommandBuilder()
-      .withArg('force:schema:sobject:list')
-      .withFlag('--sobjecttypecategory', type.toString());
-    if (username) {
```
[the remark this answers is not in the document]
lcampos: Yes, that's because I noticed that the username parameter was never being sent by the code calling the describe command and decided to update the interfaces in order to remove the `username` parameter.

packages/salesforcedx-sobjects-faux-generator/src/describe/configUtil.ts
```diff
@@ -0,0 +1,55 @@
...
+  ): Promise<string | undefined> {
+    const defaultUserName = (await this.getConfigValue(
+      projectPath,
+      defaultUserNameKey
+    )) as string;
+    const username = await Aliases.fetch(defaultUserName);
+    return Promise.resolve(username);
+  }
+
+  public static async getConfigValue(
```
brpowell: Is there an easy way to reuse [this](https://github.com/forcedotcom/salesforcedx-vscode/blob/3ff4f374e491c4d95f6c4b33b8c4fb5dc3930990/packages/salesforcedx-vscode-core/src/util/configUtil.ts#L38)?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -165,77 +170,42 @@ export enum SObjectCategory {
 
 type SubRequest = { method: string; url: string };
 type BatchRequest = { batchRequests: SubRequest[] };
-
 type SubResponse = { statusCode: number; result: SObject };
-
 type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
 
+export class ForceListSObjectSchemaExecutor {
```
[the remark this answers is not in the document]
brpowell: Do we want this to extend `SfdxCommandletExecutor`?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -165,77 +170,42 @@ export enum SObjectCategory {
...
+  }
+}
+
 export class SObjectDescribe {
   private accessToken?: string;
   private instanceUrl?: string;
   private readonly servicesPath: string = 'services/data';
   // the targetVersion should be consistent with the Cli even if only using REST calls
-  private targetVersion = '';
+  private readonly targetVersion = '46.0';
```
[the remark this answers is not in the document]
brpowell: We could do something similar to what the new templates plugin does with parsing the package json

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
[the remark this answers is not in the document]
lcampos: We do but we can't since `SfdxCommandletExecutor` has `vscode` dependencies and this module does not have access to it.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -314,48 +309,29 @@ export class SObjectDescribe {
...
+      const username = await ConfigUtil.getUsername(projectPath);
+      const authInfo = await AuthInfo.create({ username });
+      const opts = authInfo.getConnectionOptions();
+      this.accessToken = opts.accessToken;
+      this.instanceUrl = opts.instanceUrl;
+    } catch (err) {
+      const error = new Error();
+      error.name = 'Authentication Error';
+      error.message = err.message;
+      throw error;
```
brpowell: Can we log when there is a failure to authenticate?

lcampos: These errors bubble up and get logged by the error handling code in `forceGenerateFauxClasses.ts`

packages/salesforcedx-sobjects-faux-generator/src/describe/configUtil.ts
lcampos: It's not easy, I initially looked at having this code in the `salesforcedx-utils-vscode` module but that caused compilation issues in `salesforcedx-vscode-core` and `salesforcedx-vscode-apex`. The origin of the problems seems to be with importing more than one module that depends on `@salesforce/core`. Fixing/refactoring this would be a much bigger work item that was out of scope for this fix.

packages/salesforcedx-sobjects-faux-generator/package.json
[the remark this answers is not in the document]
lcampos: Yes, those problems are caused by importing more than one module that depends on `@salesforce/core` into the same package. I looked at refactoring all the code that depends on it into the `utils` module and it will be a significant refactor and testing effort.

PR #1610, opened 2019-08-28
Change log generator implementation

sfsholden:
### What does this PR do?
Automates the process of building the change log for releases.


scripts/change-log-generator.js
```diff
@@ -0,0 +1,249 @@
...
+  process.cwd(),
+  'packages',
+  'salesforcedx-vscode',
+  'CHANGELOG.md'
+);
+const CHANGE_LOG_BRANCH = 'changeLog-v';
+
+// Change log values
+const LOG_HEADER =
+  '%s - (INSERT RELEASE DATE [Month Day, Year])\n' +
```
AnanyaJha: Since we typically do releases weekly on Thursdays, would it be possible to have this be an auto-generated value that we can manually change if needed?

AnanyaJha:
These changes are very very cool!!

scripts/change-log-generator.js
sfsholden: Yes, I think so. I'm glad you mentioned it. I wasn't sure if that was something that people would think was overkill. I'll look into adding it.

scripts/change-log-generator.js
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/**
+ * Review the log and retrieve all of the commits specific to this
+ * release. Parse this info and return it as a list of hashmaps.
+ */
+function parseCommits() {
+  var results = [];
+  var commits = shell
+    .exec('git log --oneline', { silent: true })
```
sfsholden: I found that this approach isn't guaranteed to bring in all of the commits that might be part of a release. For example, if we are committing changes to develop between Tuesday-Thursday, the log of the new release branch shows that these commits appear before the SHA256 that we're checking for. 

This is the approach that I think will work better:
1) git log --left-right --graph --cherry-pick --oneline new_release_branch...previous_release_branch
2) For each of the commits listed, check if their PR# is already a part of the changelog. If they are not, then include them.

sfsholden: Note that I would include these changes as part of a separate PR.

lcampos:
@sfsholden this looks good, I'm not sure why I got a `Permission denied` error when trying to run the script directly but using the `package.json` scripts does work.

PR #1614, opened 2019-08-28
Cade may/auth dev hub configure local or global

cademay:
### What does this PR do?

updates the behavior of the 'Authorize a Dev Hub' command

new behavior: upon successful login, if there is no global default already, it sets the default dev hub globally

@W-6144190@


packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -27,6 +32,62 @@ describe('Force Auth Web Login for Dev Hub', () => {
...
+    expect(getDefaultDevHubUsernameStub.calledTwice).to.equal(true);
+    expect(getDefaultDevHubUsernameStub.calledWith(false, ConfigSource.Global)).to.equal(true);
+    expect(getDefaultDevHubUsernameStub.calledWith(false, ConfigSource.Local)).to.equal(true);
+  });
+
+  it('Should call set and write on the config file', async () => {
+
+    const configWrite = stub(ConfigFile.prototype, 'write');
+    const configSet = stub(ConfigFile.prototype, 'set');
+    setGlobalDefaultDevHubStub.restore();
```
brpowell: Probably don't need this line since it's covered in the `afterEach` already

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -27,6 +32,62 @@ describe('Force Auth Web Login for Dev Hub', () => {
...
+    expect(setGlobalDefaultDevHubStub.called).to.equal(false);
+    expect(getDefaultDevHubUsernameStub.calledTwice).to.equal(true);
+    expect(getDefaultDevHubUsernameStub.calledWith(false, ConfigSource.Global)).to.equal(true);
+    expect(getDefaultDevHubUsernameStub.calledWith(false, ConfigSource.Local)).to.equal(true);
+  });
+
+  it('Should call set and write on the config file', async () => {
+
+    const configWrite = stub(ConfigFile.prototype, 'write');
+    const configSet = stub(ConfigFile.prototype, 'set');
```
brpowell: Make sure to do `configWrite.restore()` and `configSet.restore()` at the end.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -27,6 +32,62 @@ describe('Force Auth Web Login for Dev Hub', () => {
...
+    setGlobalDefaultDevHubStub.restore();
+
+    const testUsername = "test@test.com"
+    await authWebLogin.setGlobalDefaultDevHub(testUsername);
+
+    expect(configWrite.calledOnce).to.equal(true);
+    expect(configSet.calledOnce).to.equal(true);
```
brpowell: These are good. Similar to your other cases, can we also verify that `configSet` was `calledWith(DEFAULT_DEV_HUB_USERNAME_KEY, testUsername)`? Also let's expect that `ConfigFile.create()` was passed an object with the `isGlobal` property set to true. This is confirming that we're modifying the global configuration and not a local one.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -5,13 +5,18 @@
...
+import { ConfigSource, OrgAuthInfo } from '../../../src/util/index';
 import {
   createExecutor,
   ForceAuthDevHubDemoModeExecutor,
   ForceAuthDevHubExecutor
 } from '../../../src/commands/forceAuthDevHub';
+import { DEFAULT_DEV_HUB_USERNAME_KEY, SFDX_PROJECT_FILE } from '../../../src/constants';
 import { nls } from '../../../src/messages';
+import { SinonStub, stub, spy } from 'sinon'; //
+import Sinon = require('sinon');
```
AnanyaJha: Extra slashes at the end, I don't think these imports are correctly ordered? Please remove imports that aren't being used as well 🙂

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
```diff
@@ -27,6 +32,62 @@ describe('Force Auth Web Login for Dev Hub', () => {
   });
 });
 
+describe('configureDefaultDevHubLocation on processExit of ForceAuthDevHubExecutor', () => {
+
+  let getDefaultDevHubUsernameStub: SinonStub;
+  let setGlobalDefaultDevHubStub: SinonStub;
+  const authWebLogin = ForceAuthDevHubExecutor.prototype
```
AnanyaJha: Missing semicolon

AnanyaJha:
Try running 'npm run lint' on your project because it looks like there's errors that should've been picked up by the linter!

packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
```diff
@@ -37,6 +48,51 @@ export class ForceAuthDevHubExecutor extends SfdxCommandletExecutor<{}> {
...
+  public async execute(response: ContinueResponse<any>): Promise<void> {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const execution = new CliCommandExecutor(this.build(response.data), {
+      cwd: getRootWorkspacePath()
+    }).execute(cancellationToken);
+
+    execution.processExitSubject.subscribe(this.configureDefaultDevHubLocation.bind(this));
+
```
AnanyaJha: I was wondering why we chose to bind this method? It looks like if we follow the usual pattern for this type of function we should achieve the same behavior ie.

```execution.processExitSubject.subscribe(() => this.configureDefaultDevHubLocation())```

File: packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts [hunk 7]
```diff
@@ -37,6 +48,51 @@ export class ForceAuthDevHubExecutor extends SfdxCommandletExecutor<{}> {
...
+      if (localDevHubName) {
+        await this.setGlobalDefaultDevHub(localDevHubName);
+      }
+    }
+  }
+
+  public async setGlobalDefaultDevHub(newUsername: string): Promise<void> {
+
+    const homeDirectory = require('os').homedir();
```
AnanyaJha: If we need to import 'os' let's include it with the rest of the imports at the top of the file

brpowell: ^ and either `import { homedir } from 'os';` or `import * as os from 'os'`

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
cademay: actually still needed because while the afterEach does indeed restore it, after 'afterEach' is called, 'beforeEach' is called again which instantiates it

packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extensions.autoUpdate": false
-}
+  "extensions.autoUpdate": false,
+}
```
AnanyaJha: Let's remove this change

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
AnanyaJha: Ah, so is this because you don't want the `setGlobalDefaultDevHub` method to be stubbed for this test case?

AnanyaJha:
Your changes look good! One last thing, can you reformat the files with Prettier? We can take a look at it together if the option isn't showing up for you in the command palette.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceAuthDevHub.test.ts
cademay: exactly

because if it's stubbed then none of the methods inside it get called, so that stub was getting in the way of calls to the ConfigFile method calls

PR #1629, opened 2019-09-03
FilePathExistsChecker refactor

brpowell:
### What does this PR do?

Reconciles the two existing FilePathExists postcondition checkers. These two essentially do the same thing - check for existing files in a workspace before performing a potential overwrite. We had two because of the required different behavior for how to glob for the files and the type of input being passed. This PR attempts to generalize the behavior by introducing GlobStrategies, as well as putting some now shared code into a more common module. The refactor helps before moving forward with the org browser multi-retrieve work.   

W-6526764


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
```diff
@@ -111,92 +97,20 @@ export async function forceSourceRetrieveCmp(componentNode: BrowserNode) {
...
-      dirName,
-      label
+      globStrategy,
+      nls.localize('warning_prompt_cmp_file_overwrite', label)
     )
   );
   await commandlet.run();
 }
-
-export class FilePathExistsChecker implements PostconditionChecker<{}> {
```
brpowell: The org browser specific FilePathExistsChecker

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -81,80 +79,3 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
...
-    dirPath: string,
-    fileName: string,
-    fileExt: string
-  ): string {
-    const bundleName = fileName;
-    return path.join(dirPath, bundleName, `${fileName}${fileExt}`);
-  }
-}
-
-export class FilePathExistsChecker
```
brpowell: The create commands specific FilePathExistsChecker

packages/salesforcedx-vscode-core/src/commands/util/globStrategies.ts
```diff
@@ -0,0 +1,103 @@
...
+  public abstract globs(
+    selection: DirFileNameSelection
+  ): Promise<GlobPattern[]>;
+}
+
+/**
+ * Create glob patterns that exactly match the given DirFileNameSelection.
+ * outuptdir in selection expected as a fully resolved path.
+ */
+class CheckGivenPath extends BaseGlobStrategy {
```
brpowell: This strategy is used by the create template commands.

packages/salesforcedx-vscode-core/src/commands/util/globStrategies.ts
```diff
@@ -0,0 +1,103 @@
...
+    return Promise.resolve([`{${filePaths.join(',')}}`]);
+  }
+}
+
+/**
+ * Create glob patterns for each package directory in the workspace. outputdir
+ * in selection expected as path that will be the suffix of the fully qualified
+ * package directory path. E.g. outputdir: main/default/classes
+ */
+class CheckAllPackages extends BaseGlobStrategy {
```
brpowell: This strategy is being used by the org browser retrieve command

packages/salesforcedx-vscode-core/src/commands/util/postCheckers.ts
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import {
```
brpowell: I've added a new `util` submodule in the `commands` module. In this PR it houses the globStrategies, sourcePathStrategies, and new location for the FilePathExistsChecker. `commands.ts` is becoming a rather large mixed bag, so I thought we could lay some groundwork for later to break it up a bit into more readable files. Perhaps something to do during our time allocation for this kind of work.

Let me know what you think of going this route, I am open to whatever.

packages/salesforcedx-vscode-core/src/commands/util/sourcePathStrategies.ts
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
```
brpowell: The new home for SourcePathStrategy related code since it is now shared between the template commands and retrieve component command.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveCmp.test.ts
```diff
@@ -1,285 +1,280 @@
-/*
```
brpowell: Ignore for now along with `baseTemplateCommand.test.ts`, will fix.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postCheckers.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import { Uri, workspace } from 'vscode';
+import { FilePathExistsChecker } from '../../../../src/commands/util';
+import { GlobStrategy } from '../../../../src/commands/util';
+import { nls } from '../../../../src/messages';
+import { notificationService } from '../../../../src/notifications';
+
+const env = sandbox.create();
+
+describe('Postcondition Checkers', () => {
+  describe('FilePathExistsChecker', () => {
```
brpowell: New tests for FilePathExistsChecker. It's now being tested with the glob strategies and independently of the commands that utilize it.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveCmp.test.ts
```diff
@@ -75,211 +65,3 @@ describe('Generate Appropriate Suffix', () => {
     expect(suffixes).to.eql(expected);
   });
 });
-
-describe('FilePathExistsChecker for Aura bundle', () => {
```
brpowell: These removed tests are handled in the new tests

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveCmp.ts
```diff
@@ -111,92 +97,20 @@ export async function forceSourceRetrieveCmp(componentNode: BrowserNode) {
...
+    ? GlobStrategyFactory.createCheckBundleInAllPackages(...fileExts)
+    : GlobStrategyFactory.createCheckFileInAllPackages(...fileExts);
 
   const executor = new ForceSourceRetrieveExecutor(typeName, componentName);
 
   const commandlet = new SfdxCommandlet(
-    workspaceChecker,
-    parameterGatherer,
+    new SfdxWorkspaceChecker(),
+    new DirFileNameGatherer(join('main', 'default', dirName), componentName),
```
lcampos: Very cool 👍

lcampos: The one thing I'd like to know is if we can/should avoid hardcoding `main/default` to the retrieve since I've seen people set up their projects with the following structure:

```
  - force-app
      - classes
      - triggers
```

instead of:

```
  -force-app
    - main
       - default
          - classes
          - triggers
```

brpowell: This hard coding used to be [here](https://github.com/forcedotcom/salesforcedx-vscode/pull/1629/files#diff-018c50cc4869276ca6a53b2421852471L191). The reason we're doing so at least for the retrieve component command is because the org browser calls `force:source:retrieve` with mostly default parameters, which will always retrieve to `main -> default -> somefolder`.

packages/salesforcedx-vscode-core/src/commands/util/postCheckers.ts
AnanyaJha: I think this is a good idea! I was going to ask whether we should include it in the utils folder directly in the `core` package, but I think this structure makes more sense because we don't need these outside the `commands` scope.

packages/salesforcedx-vscode-core/src/commands/util/index.ts
```diff
@@ -0,0 +1,7 @@
+export {
```
lcampos: Add copyright header

packages/salesforcedx-vscode-core/src/commands/util/parameterGatherers.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  ContinueResponse,
+  ParametersGatherer
+} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
```
lcampos: Add copyright header

packages/salesforcedx-vscode-core/src/commands/util/parameterGatherers.ts
```diff
@@ -0,0 +1,19 @@
+import {
```
lcampos: I like this approach to break the different command utility methods into files based on the interfaces they implement. Are we going to do this for the rest of the classes that are also implementing these interfaces. These changes now leave us with 2 main places where these type of classes are hosted, under the `commands/util` folder and on the `commands.ts` file.

brpowell: Correct, I will be filing work to finish breaking up `commands.ts` into respective files.

PR #1642, opened 2019-09-11
Updated docs site for DSC

ntotten:
* Updated docs design for DSC
* docs path
* updated paths
* netlify url rewrites & redirects
* change ruby version


docs/_includes/head.html
```diff
@@ -0,0 +1,170 @@
...
+  name="viewport"
+  content="width=device-width, initial-scale=1, maximum-scale=1"
+/>
+<meta
+  property="og:image"
+  content="https://res.cloudinary.com/hy4kyit2a/image/upload/og-image.jpg"
+/>
+<link
+  rel="icon"
+  href="//developer.salesforce.com/resources2/favicon.ico"
```
lcampos: Shouldn't these image's href use `https` ?

ntotten: Doesn't really matter, right? That is without a protocol and the site always loads as HTTPS.

lcampos: I think that it only matters for the development scenario if the page is accessed directly without spinning up the server. For production, you are correct in that it will use the protocol being used by the site.

ntotten: Oh, i just realized. These are part of the generated pages from the docs team. They shouldn't be edited directly.

lcampos: Ok, that's good to know 👍

PR #1652, opened 2019-09-17
Remove use of momentjs

lcampos:
### What does this PR do?
This PR replaces momentjs by using the native locale methods and other date formatting utilities. This is part of the changes needed to reduce the size of the extensions. By removing momentjs, we remove a dependency on a module that is not tree-shakeable.

@W-6587405@

### DIFF
--- a/packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
@@ -21,7 +21,7 @@ describe('Date format utility', () => {
     expect(doubleDigit).to.equal('09');
   });
 
-  it('Should return a the same two character string when provided a double digit number', () => {
+  it('Should return the same two character string when provided with a double digit number', () => {
     const singleDigit = 13;
     const doubleDigit = makeDoubleDigit(singleDigit);
     expect(doubleDigit).to.equal('13');
--- a/packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
@@ -30,7 +30,8 @@ describe('Date format utility', () => {
   it('Should return a YYYYMMddHHmmss formatted string', () => {
     const utcString = '2019-09-10T04:34:28+0000';
     const result = getYYYYMMddHHmmssDateFormat(new Date(utcString));
-    expect(result).to.equal('20190909213428');
+    expect(result.startsWith('2019')).to.equal(true);
+    expect(result.length).to.equal(14);
   });
 
   it('Should return a YYYYMMddHHmmss formatted string when using optionYYYYMMddHHmmss', () => {
@@ -42,7 +43,8 @@ describe('Date format utility', () => {
       optionYYYYMMddHHmmss
     );
 
-    expect(localDateFormatted).to.equal('09/09/2019, 9:34:28 PM');
+    const dateTimeFormatRegex = /([0-9]{2})(\/)([0-9]{2})(\/)([0-9]{4})(\,)\s([0-9])(:)([0-9]{2})(:)([0-9]{2})\s(AM|PM)/;
+    expect(dateTimeFormatRegex.test(localDateFormatted)).to.equal(true);
   });
 
   it('Should return a HHmm formatted string when using optionHHmm', () => {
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -90,9 +93,8 @@ export class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
         mkdir('-p', logDir);
       }
 
-      const date = moment(new Date(response.data.startTime)).format(
-        'YYYYMMDDhhmmss'
-      );
+      const localUTCDate = new Date(response.data.startTime);
+      const date = getYYYYMMddHHmmssDateFormat(localUTCDate);
```
lcampos: This is used in naming the debug log that gets downloaded when running `SFDX: Get Debug Logs`

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -135,16 +137,17 @@ export class LogFileSelector
     if (logInfos.length > 0) {
       const logItems = logInfos.map(logInfo => {
         const icon = '$(file-text) ';
+        const localUTCDate = new Date(logInfo.StartTime);
+        const localDateFormatted = localUTCDate.toLocaleDateString(
```
lcampos: This is used for listing the available debug log records when running `SFDX: Get Debug Logs`
debug_log_list

packages/salesforcedx-vscode-core/src/decorators/traceflag-time-decorator.ts
```diff
@@ -5,7 +5,10 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import * as moment from 'moment';
+import {
```
lcampos: The changes on this file control the formatting of time and date showed in the footer text added after running `SFDX: Turn On Apex Debug Log for Replay Debuger`
replay_debugger_footer

packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '../../../src/date';
+
+describe('Date format utility', () => {
+  it('Should return a string with two characters, zero being the first character', () => {
+    const singleDigit = 9;
+    const doubleDigit = makeDoubleDigit(singleDigit);
+    expect(doubleDigit).to.equal('09');
+  });
+
+  it('Should return a the same two character string when provided a double digit number', () => {
```
sfsholden: typo in test description

lcampos:
> Changes look good! What made the momentjs module not be compatible with tree-shaking?

@AnanyaJha this thread (https://github.com/moment/moment/issues/2373) goes a bit more in depth into the challenges of breaking it into smaller pieces using tools like webpack but it boils down to the way the library is structured.

packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('Should return a YYYYMMddHHmmss formatted string when using optionYYYYMMddHHmmss', () => {
+    const utcString = '2019-09-10T04:34:28+0000';
+
+    const localUTCDate = new Date(utcString);
+    const localDateFormatted = localUTCDate.toLocaleDateString(
+      'en-US',
+      optionYYYYMMddHHmmss
+    );
+
+    const dateTimeFormatRegex = /([0-9]{2})(\/)([0-9]{2})(\/)([0-9]{4})(\,)\s([0-9])(:)([0-9]{2})(:)([0-9]{2})\s(AM|PM)/;
```
sfsholden: @lcampos This test fails locally for me because the conversion from utc to mst causes the time stamp to be 10:34 instead of 9:34. The regex needs to be updated to:
([0-9]{2})(\/)([0-9]{2})(\/)([0-9]{4})(\,)\s([0-9]**{1,2}**)(:)([0-9]{2})(:)([0-9]{2})\s(AM|PM)

packages/salesforcedx-utils-vscode/test/unit/date/format.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('Should return a HHmm formatted string when using optionHHmm', () => {
+    const utcString = '2019-09-10T04:34:28+0000';
+
+    const localUTCDate = new Date(utcString);
+    const localTimeFormatted = localUTCDate.toLocaleTimeString(
+      'en-US',
+      optionHHmm
+    );
+
+    const timeFormatRegex = /([0-9])(:)([0-9]{2})\s(AM|PM)/;
```
sfsholden: If we update the last example, we should really update this one to ([0-9]**{1,2}**)(:)([0-9]{2})\s(AM|PM). This one passes for me locally, but only because we beginning the regex at the start of the string.

lcampos: Thanks for pointing this out, I'll open a follow up PR to take care of these failures.

PR #1662, opened 2019-09-17
Retrieve Multiple Components by Metadata Type

brpowell:
### What does this PR do?

Enables retrieving multiple components by metadata type. If there are local versions of the components to be retrieved, prompt the user the option to skip, overwrite or cancel the operation.

W-6501184


packages/system-tests/assets/sfdx-simple/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
lcampos: I think this needs to be removed

.vscode/snippets.code-snippets
```diff
@@ -0,0 +1,31 @@
+{
```
lcampos: Nice

lcampos:
@brpowell adding it here so everyone's aware of the conversation we had. I think we should split these changes into 2 reviews, one with the changes needed to the template commands and the other for the changes needed for the org browser.

PR #1667, opened 2019-09-18
Retrieve Multiple Components by Metadata Type

brpowell:
### What does this PR do?

Enables retrieving multiple components by metadata type. If there are local versions of the components to be retrieved, prompt the user the option to skip, overwrite or cancel the operation.

W-6501184


.vscode/snippets.code-snippets
```diff
@@ -0,0 +1,31 @@
...
+  // "Print to console": {
+  // 	"scope": "javascript,typescript",
+  // 	"prefix": "log",
+  // 	"body": [
+  // 		"console.log('$1');",
+  // 		"$2"
+  // 	],
+  // 	"description": "Log output to console"
+  // }
+  "Insert Copyright Header": {
```
brpowell: For convenience

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/describers.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  public abstract buildMetadataArg(): string;
+
+  public abstract gatherOutputLocations(): Promise<LocalComponent[]>;
+
+  protected async buildOutput(node: BrowserNode): Promise<LocalComponent[]> {
+    const typeNode = node.getAssociatedTypeNode();
+    // create cmp for every package directory to accommodate for the cli retrieve bug
+    const packageDirectories = await SfdxPackageDirectories.getPackageDirectoryPaths();
+    return packageDirectories.map(directory => ({
```
brpowell: Because of the CLI bug where sometimes a component can be retrieved to a non-default package directory, we need to check every one of them and add it as a separate component. Once the bug is fixed we can remove this

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/describers.ts
```diff
@@ -0,0 +1,86 @@
...
+      suffix: typeNode.suffix
+    }));
+  }
+}
+
+class TypeNodeDescriber extends NodeDescriber {
+  public buildMetadataArg(data?: LocalComponent[]): string {
+    if (data) {
+      const dedupe = new Set<string>(); // filter dupes caused by cli bug. See buildOutput in parent class
+      data.forEach(c => dedupe.add(`${c.type}:${c.fileName}`));
```
brpowell: Because there can be multiple components that actually represent one (just with different package directories), we need to dedupe them. Once the CLI bug is fixed we can remove this

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -8,13 +8,23 @@ import {
...
-import { GlobStrategy } from './globStrategies';
+import { getRootWorkspacePath } from '../../util';
+import { MetadataDictionary } from '../../util/metadataDictionary';
+import { PathStrategyFactory } from './sourcePathStrategies';
 
+type OneOrMany = LocalComponent | LocalComponent[];
+type ContinueOrCancel = ContinueResponse<OneOrMany> | CancelResponse;
+
+// TODO: Replace with ComponentOverwritePrompt in subsequent PR
 export class FilePathExistsChecker
```
brpowell: This technically isn't needed anymore. The create template commands can use the new OverwriteComponentPrompt checker, but that can be done in a separate PR to keep this one smaller.

docs/articles/user-guide/org-browser.md
```diff
@@ -37,6 +37,8 @@ You can refresh metadata at org level, for a type, for folders in a type, and fo
 
 ## Retrieving Components
 
-You can retrieve a component to your local project by clicking the retrieve button **(7)** next to the component name. Currently, the component is retrieved into your default package directory, which is defined in the [sfdx-project.json](../getting-started/first-project#the-sfdx-projectjson-file) file. If a local version of the component exists, you'll be prompted to confirm before overwriting it.
+You can retrieve a component to your local project by clicking the retrieve button **(7)** next to the component name. Currently, the component is retrieved into your default package directory, which is defined in the [sfdx-project.json](../getting-started/first-project#the-sfdx-projectjson-file) file.
 
-![Overwrite local component](../../images/org_browser_overwrite.png)
+You may also retrieve multiple components at a time by clicking the retrieve button next to a metadata type. This will fetch all of the components in the default org for that particular type. If any components will be overwritten by the operation, a prompt will appear allowing you to choose how to handle each component.
+
+![Overwrite components](../../images/overwrite-prompt.png)
```
brpowell: @hbc-bindu Doc updates

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/describers.ts
lcampos: We should probably add the work item number to the comment on the code

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.describer = describer;
+  }
+
+  public build(data?: LocalComponent[]): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_source_retrieve_text'))
+      .withLogName('force_source_retrieve')
+      .withArg('force:source:retrieve')
+      .withArg('-m')
+      .withArg(this.describer.buildMetadataArg(data))
```
lcampos: Should we evaluate if `data` has a value before setting a command argument ?

packages/salesforcedx-vscode-core/src/commands/util/parameterGatherers.ts
```diff
@@ -1,13 +1,10 @@
-/*
```
lcampos: Why did we remove the copyright header ?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
lcampos: Are we splitting it as a separate story or just a follow up PR under the same story ?

packages/salesforcedx-vscode-core/src/util/metadataDictionary.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { LocalComponent } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import { PathStrategyFactory, SourcePathStrategy } from '../commands/util';
+
+const DEFINITIONS: { [key: string]: MetadataInfo } = {
+  apexclass: {
```
lcampos: Why do we need the definitions for ApexClass, ApexTrigger, etc if not having a definition defaults to using `PathStrategyFactory.createDefaultStrategy()` in `OverwriteComponentPrompt.componentExists` ?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -55,3 +65,141 @@ export class FilePathExistsChecker
...
+      const path = join(
+        getRootWorkspacePath(),
+        pathStrategy.getPathToSource(outputdir, fileName, extension)
+      );
+      return existsSync(path);
+    });
+  }
+
+  private getFileExtensions(component: LocalComponent) {
+    const info = MetadataDictionary.getInfo(component.type);
```
lcampos: You can pass `info` as a parameter since you've already called it as part of `componentExists`.

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -55,3 +65,141 @@ export class FilePathExistsChecker
...
+    let metadataSuffix;
+    if (component.suffix) {
+      metadataSuffix = component.suffix;
+    } else if (info && info.suffix) {
+      metadataSuffix = info.suffix;
+    } else {
+      throw new Error(`Missing suffix for ${component.type}`);
```
lcampos: I think we might need to add some error handling in the code that ends up calling this method for the scenarios in which we might end up here. Below is a screenshot of the error I got when I broke this.
Screen Shot 2019-09-20 at 5 50 05 PM

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
brpowell: I have it as a separate story

packages/salesforcedx-vscode-core/src/util/metadataDictionary.ts
brpowell: The definitions aren't just for the path strategy, but to hold any meta information that's needed in the extensions. The template commands will be able to reference the extensions/directory information for these types.

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
brpowell: I would do that, although I need both the `suffix` and `type` properties off of `component`

packages/salesforcedx-vscode-core/src/commands/util/parameterGatherers.ts
brpowell: heh whoops, good question 🙃

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.ts
brpowell: I have `data` as an optional property in the `buildMetadataArg` interface so this should be okay. That responsibility will fall to the implementation.

PR #1679, opened 2019-09-23
Select sObject category for Refresh SObject Definitions command 

brpowell:
### What does this PR do?

Choose between All, Custom, or Standard sObject definitions to refresh

W-6558267


brpowell:
Screen Shot 2019-09-23 at 2 24 50 PM

UI example @hbc-bindu @ntotten

Issue #1678, opened 2019-09-23
Not activating and Taking commands


himanshu371:
Issue Type: <b>Bug</b>

CLI is not beeing activating automatically and taking commands also 

Extension version: 46.15.1
VS Code version: Code 1.38.1 (b37e54c98e1a74ba89e03073e5a3761284e3ffb0, 2019-09-11T13:35:15.005Z)
OS version: Windows_NT x64 10.0.17763

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz (4 x 2712)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>oop_rasterization: disabled_off<br>protected_video_decode: enabled<br>rasterization: enabled<br>skia_deferred_display_list: disabled_off<br>skia_renderer: disabled_off<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>viz_display_compositor: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|7.89GB (2.28GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>

frizz2310:
Had the same problem with the same version on MacOS.  Reinstalled everything, tried older versions, uninstalled all extensions, checked CLI, Java, etc.  Still no luck.  Finally, turned off Settings > Extensions > Salesforce Feature Previews > Internal-development.  Forgot I turned this on, but after turning it off and reloading, everything is working again.

bcattaneo:
Experiencing the same issue. It was apparently after the 9/19/2019 update.

bcattaneo:
@himanshu371 try updating/reinstalling the plugin. Mine updated automatically and now it's working again, I guess they fixed something.

lcampos:
@himanshu371 are you still running into this ? If so, what errors do you see when you open VSCode's Dev Tools (Help > Toggle Developer Tools then select the Console tab) ?

himanshu371:
[...]

Issue #1670, opened 2019-09-19
command 'sfdx.force.project.create' not found


kbadamane121:
this issue has started after i installed the vs studio code and i am not able to create create any project using sfdx commands

kbadamane121:
Thanks for your reply

   - Salesforce extensions for VSCode version:-46.15.1/previous was 46.14.0
   - VSCode version:-1.38.1
   - OS:-WIN 10

Cannot find en-us, defaulting to en

Activating extension 'salesforce.salesforcedx-vscode-apex' failed: Class
extends value undefined is not a constructor or null.
_logMessageInConsole @ abstractExtensionService.ts:391
console.ts:137 [Extension Host] Activating extension
`salesforce.salesforcedx-vscode-apex` failed:  Class extends value
undefined is not a constructor or null
Cannot activate extension 'Apex Interactive Debugger' because it depends on
extension 'salesforce.salesforcedx-vscode-apex', which failed to activate.
_logMessageInConsole @ abstractExtensionService.ts:391

abstractExtensionService.ts:391 Cannot activate extension 'Apex Replay
Debugger' because it depends on extension
'salesforce.salesforcedx-vscode-apex', which failed to activate.

lcampos:
@kbadamane121 this is caused by your configuration enabling a setting called `salesforcedx-vscode-core.internal-development` which was meant to be used only by our team. I'll remove it in the next release but you should be able to get past this errors by removing it from your settings.json file.

lcampos:
A fix to remove the setting and update the api so the Apex extension doesn't fail when that setting is in use has been released.

pathamallas:
I m trying to create a project on palette its showing 'sfdx.force.project.create' not found

alexiswilliams:
salesforcedx-vscode-core.internal-development is still showing up for me. I had to remove it as well from the settings JSON file.

shivanand11totar:
![Untitled](https://user-images.githubusercontent.com/65186811/81826268-3c9e7d80-9555-11ea-8ef6-2d8381efd3d1.png)
![Untitled](https://user-images.githubusercontent.com/65186811/81826413-5fc92d00-9555-11ea-9ece-575197a882a2.png)

shivanand11totar:
![Untitled](https://user-images.githubusercontent.com/65186811/81826510-7d969200-9555-11ea-91bb-f9074ecdd323.png)

shivanand11totar:
[...]

PR #1689, opened 2019-09-30
Fix internal development mode causing Apex extension issues

lcampos:
### What does this PR do?
Add missing api methods to the internal development mode which cause the Apex extension to fail during activation

@W-6695536@, #1670 , #1678

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -710,11 +710,6 @@
           "type": "boolean",
           "default": false,
           "description": "%push_or_deploy_on_save_enabled_description%"
-        },
-        "salesforcedx-vscode-core.internal-development": {
-          "type": "boolean",
-          "default": false,
-          "description": "%internal_dev_description%"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -710,11 +710,6 @@
           "type": "boolean",
           "default": false,
           "description": "%push_or_deploy_on_save_enabled_description%"
-        },
```
lcampos: Customers are enabling this setting by mistake so I'm removing it from the Settings menu.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -460,6 +460,16 @@ export async function activate(context: vscode.ExtensionContext) {
 
     // Api
     const internalApi: any = {
+      channelService,
```
lcampos: Missing methods needed by the Apex extension, Apex Replay Debugger and Apex Debugger.

packages/salesforcedx-vscode-core/package.json
brpowell: How is it enabled now?

lcampos: Like any of the other settings, via the settings.json file or the Settings UI.

brpowell: Gotcha, that's what I figured. Wasn't sure if it was "deregistered" or something by doing this.

Issue #1708, opened 2019-10-14
Org create/set fail when alias contains white space


maaaaarco:
### Summary
When creating a new Scratch Org I'm allowed to enter an alias that contains non alphanumeric characters like the whitespace. This has to implications:

1. Creation of the Scratch Org will fail on windows
1. Switching to a Scratch Org, whose alias contains a whitespace, will fail on both Mac and Windows

### Steps To Reproduce:
**Windows**
1.  Create a Default Scratch Org and give it an alias that contains a whitespace like _my scratch org_
1.  Command will fail with error message:
`11:43:25.483 sfdx force:org:create -f config\reporting-scratch-def.json --setalias my scratch org --durationdays 7 --setdefaultusername --json --loglevel fatal
11:44:04.160 sfdx force:org:create -f config\reporting-scratch-def.json --setalias my scratch org --durationdays 7 --setdefaultusername --json --loglevel fatal ended with exit code 1`

**Mac**
1.  Create a Default Scratch Org and give it an alias that contains a whitespace like _my scratch org_
1.  Create a Default Scratch Org and give it an alias that does **not** contain a whitespace like _myScratchOrg_
1.  Try to switch back to previous Scratch Org _my scratch org_
1.  Command will fail with this error message:
`=== Failures
NAME             MESSAGE
───────────────  ──────────────────────────────────────
defaultusername  No org configuration found for name my
16:10:56.227 sfdx force:config:set defaultusername=my ended with exit code 0`

### Expected result
Salesforce Extension should prevent me to create alias that contain non alphanumeric characters

### Actual result
I'm allowed to specify alias that contain non alphanumeric characters

maaaaarco:
Solved with last update that prevents users to enter Scratch Org names with whitespace

PR #1709, opened 2019-10-14
Scratch Org alias and Expiration days validations

maaaaarco:
### What does this PR do?
Prevents user to specify an alias that contains non alphanumeric characters and provides a useful error message:
Screenshot 2019-10-14 at 17 17 23

Prevents a user to specify a number that is not in the 1-30 range and provides a useful error message:
Screenshot 2019-10-14 at 17 31 10

#1708


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -270,6 +270,9 @@ export const messages = {
   empty_components: 'No components available',
   error_auth_token: 'Error refreshing authentication token.',
   error_no_org_found: 'No org authorization info found.',
+  error_invalid_org_alias:
+    'Alias can only contain underscores and alphanumeric characters',
+  error_invalid_expiration_days: 'Number of days should be between 1 and 30',
```
brpowell: @hbc-bindu can you review this text?

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -143,27 +149,20 @@ export class AliasGatherer implements ParametersGatherer<Alias> {
       prompt: nls.localize(
         'parameter_gatherer_enter_scratch_org_expiration_days'
       ),
-      value: defaultExpirationdate
+      value: defaultExpirationdate,
+      validateInput: value => {
```
brpowell: Cool! Thank you for contributing @maaaaarco. Can you add some tests for your validation functions?

maaaaarco: Sure, will do

maaaaarco: @brpowell do you think that it would make sense to group these kind of functions in a util file (under the commands folder) ?
In this way we might reuse them for other commands too.

maaaaarco: I moved those function is a separate file and add the specific tests... I'm a bit lost though when it comes to the _forceOrgCreate.test_...
Because of these
`beforeEach(() => {
    });`
I'm not sure how to test that the validation function is being called. Do you have any suggestions/example I can look at?

brpowell: I think the tests you have for the validations themselves are good as is, the vscode api takes care of triggering the validation function

maaaaarco: Yeah cool, I didn't see that package. I'll do as u suggest ;)

maaaaarco:
Refactored as suggested

brpowell:
This PR has been merged with #1729 and will be released next week. Thank you @maaaaarco for your contribution!

PR #1711, opened 2019-10-14
Break up commands.ts into separate utils

brpowell:
### What does this PR do?

Break up the commands.ts utils file into separate utils.

W-6548487


brpowell:
No added functionality, just reorganizing and import updates. Breaking up this file required many imports to be changed.

packages/salesforcedx-vscode-core/src/commands/util/preconditionCheckers.ts
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: Nitpick: this should be 2019 since the code has moved around significantly

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos: Same as in `preconditionCheckers.ts`

PR #1717, opened 2019-10-16
Remove analytics option from project create

lcampos:
### What does this PR do?
Removes the analytics option from project create since the cli piece will not go out in the next release.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
@@ -33,8 +33,7 @@ import {
 
 export enum projectTemplateEnum {
   standard = 'standard',
-  empty = 'empty',
-  analytics = 'analytics'
+  empty = 'empty'
 }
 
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
```diff
@@ -144,10 +143,6 @@ export class SelectProjectTemplate
       new ProjectTemplateItem(
         projectTemplateEnum.empty,
         'force_project_create_empty_template'
-      ),
-      new ProjectTemplateItem(
-        projectTemplateEnum.analytics,
-        'force_project_create_analytics_template'
       )
     ];
```
AnanyaJha: do we need to modify the build method to use the `--standard` flag for the template instead of `data.projectTemplate` again?

brpowell: It's okay as is. We're just passing `standard` or `empty` to the `--template` flag depending on the selection

PR #1729, opened 2019-10-22
Scratch Org alias and Expiration days validations (merge PR)

brpowell:
### What does this PR do?

Prevents user from specifying an alias that contains non alphanumeric characters and provides a useful error message. Also prevents specifying a number that is not in the 1-30 range for the scratch org expiration date.

This PR is for running the autobuilds and merging the contributions by @maaaaarco.

Original PR #1709 , @W-6819919@


lcampos:
@brpowell looks there's a couple of test failures with the tests added here.

brpowell:
@lcampos just pushed an update. Since input validations are being used now instead of automatically converting the input, I removed the tests that verified the latter.

Issue #1702, opened 2019-10-08
Refresh SObject definitions not working in 46.17.0 (also 47.1.0)


az-oolloow:
### Summary

When i run SFDX Refersh SObject definitions against a developer edition org (not scratch, not sandbox), it fails

### Steps To Reproduce:

1.  Connect SFDX cli and Salesforce Extension pack to a developer edition org
1.  Run SFDX Refresh SObject definitions (there is now a new menu for all objects, custom or standard - for me all fail)

### Expected result

The SObject definitions are refreshed

### Actual result

The command fails

### Additional information

Error log:
```
Starting SFDX: Refresh SObject Definitions

14:34:48.623 sfdx sobject definitions refresh
Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.
14:34:53.478 sfdx sobject definitions refresh ended with error Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.

14:34:53.479 sfdx sobject definitions refresh ended with exit code 1
```
Starting SFDX: Refresh SObject Definitions

14:34:48.623 sfdx sobject definitions refresh
Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.
14:34:53.478 sfdx sobject definitions refresh ended with error Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.

14:34:53.479 sfdx sobject definitions refresh ended with exit code 1
```
Starting SFDX: Refresh SObject Definitions

14:40:21.581 sfdx sobject definitions refresh
Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.
14:40:26.996 sfdx sobject definitions refresh ended with error Failure performing sObject describe Must pass a username and/or OAuth options when creating an AuthInfo instance.

14:40:26.997 sfdx sobject definitions refresh ended with exit code 1
```

**VS Code Version**:
Version: 1.38.1
Commit: b37e54c98e1a74ba89e03073e5a3761284e3ffb0
Date: 2019-09-11T13:31:32.854Z

**SFDX CLI Version**:
```
sfdx-cli/7.27.0-714f62f9c1 darwin-x64 node-v10.15.3
```

**OS and version**:
macOS 10.15 (19A583)

All other commands work fine, i can retrieve source, deploy it, run tests. So the Auth token is definitely valid.

az-oolloow:
[...]

PR #1739, opened 2019-10-23
Ensure valid session when fetching SObject definitions 

brpowell:
### What does this PR do?

Ensures there is a valid session token before fetching SObject definitions from the default org.

W-6803122, #1702

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { AuthInfo } from '@salesforce/core';
+import { Org } from '@salesforce/core';
 import {
   CliCommandExecution,
   CliCommandExecutor,
@@ -264,7 +264,6 @@ export class SObjectDescribe {
 
   public buildXHROptions(types: string[], nextToProcess: number): XHROptions {
     const batchRequest = this.buildBatchRequestBody(types, nextToProcess);
-
     return {
       type: 'POST',
       url: this.buildBatchRequestURL(),
@@ -319,10 +318,11 @@ export class SObjectDescribe {
   public async getConnectionData(projectPath: string) {
     try {
       const username = await ConfigUtil.getUsername(projectPath);
-      const authInfo = await AuthInfo.create({ username });
-      const opts = authInfo.getConnectionOptions();
-      this.accessToken = opts.accessToken;
-      this.instanceUrl = opts.instanceUrl;
+      const org = await Org.create({ aliasOrUsername: username });
+      await org.refreshAuth();
+      const { accessToken, instanceUrl } = org.getConnection();
+      this.accessToken = accessToken;
+      this.instanceUrl = instanceUrl;
     } catch (err) {
       const error = new Error();
       error.name = 'Authentication Error';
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -319,10 +318,11 @@ export class SObjectDescribe {
   public async getConnectionData(projectPath: string) {
     try {
       const username = await ConfigUtil.getUsername(projectPath);
-      const authInfo = await AuthInfo.create({ username });
-      const opts = authInfo.getConnectionOptions();
-      this.accessToken = opts.accessToken;
-      this.instanceUrl = opts.instanceUrl;
+      const org = await Org.create({ aliasOrUsername: username });
+      await org.refreshAuth();
```
brpowell: Will fetch a new token only if the current one is invalid

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -289,12 +288,28 @@ export class SObjectDescribe {
...
+        options = this.buildXHROptions(types, nextToProcess);
+        response = await this.runRequest(options);
+      } catch (e) {
+        if (e.status !== 401) {
+          throw e;
+        }
+        await this.getConnectionData(org, true);
+        options = this.buildXHROptions(types, nextToProcess);
+        response = await this.runRequest(options);
       }
```
brpowell: First attempt to run the request with the current token. If it's expired, refresh it.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -289,12 +288,28 @@ export class SObjectDescribe {
...
+      const org = await Org.create({
+        aliasOrUsername: await ConfigUtil.getUsername(projectPath)
+      });
+
       if (!this.accessToken || !this.instanceUrl) {
-        await this.getConnectionData(projectPath);
+        await this.getConnectionData(org);
+      }
+
+      let response: XHRResponse;
```
lcampos: With this code there's still the potential of ending up in the same invalid state if refreshing the connection doesn't work. We can change this to be in a "retry function" that will run the call outs two times, at most, and handles errors.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -316,19 +331,13 @@ export class SObjectDescribe {
...
-      const authInfo = await AuthInfo.create({ username });
-      const opts = authInfo.getConnectionOptions();
-      this.accessToken = opts.accessToken;
-      this.instanceUrl = opts.instanceUrl;
-    } catch (err) {
-      const error = new Error();
-      error.name = 'Authentication Error';
-      error.message = err.message;
-      throw error;
+  public async getConnectionData(org: Org, withRefresh?: boolean) {
```
lcampos: This looks good 👍

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
brpowell: If the refresh fails then something is likely wrong with the connection, or the refresh token is bad. We could handle the latter case to clearly ask to reauthorize the org in the error that gets thrown. This was meant to only handle the unauthorized status of the first call and not necessarily be a retry x times kind of thing.

lcampos: I see, let's make sure we handle any errors after doing the refresh then.

PR #1759, opened 2019-10-29
Reenable dependency cache in circle ci builds

brpowell:
### What does this PR do?

Reenable dependency cache in circle ci builds

W-6815766


brpowell:
It appears a new cache key is working. I have rerun the workflow multiple times to make sure that the cache is saved and restored with the new key correctly. Can verify here https://circleci.com/gh/forcedotcom/salesforcedx-vscode/5120

PR #1773, opened 2019-11-06
add commands for sfdx force:lightning:lwc:start

mysticflute:
### What does this PR do?

This adds several commands to the salesforcedx-vscode-lwc package:
- SFDX: Start Local Development Server
- SFDX: Stop Local Development Server
- SFDX: Open Local Development Server
- SFDX: Preview Component Locally

This uses `SfdxCommandlet` and related classes, but a few small changes were made since this command starts a server and the process will continue to run and show output, unlike the other commands that wait for and expect process to end once the command is successful.

W-6328653


packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -222,24 +225,28 @@ export class CliCommandExecution implements CommandExecution {
...
+            await this.killExecution();
           } catch (e) {
             console.log(e);
           }
         }
       });
     }
   }
+
+  public async killExecution(signal: string = 'SIGKILL') {
```
mysticflute: this is needed so that the process that is running the server can be stopped.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -53,7 +53,7 @@ export class ChannelService {
         this.getExecutionTime() + ' ' + execution.command.toCommand()
       );
       this.channel.append(' ');
-      if (data !== undefined) {
+      if (data !== undefined && data !== null) {
```
mysticflute: this can happen if the process was killed with a signal

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -67,7 +68,7 @@ export abstract class SfdxCommandletExecutor<T>
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
     const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: getRootWorkspacePath(),
+      cwd: this.executionCwd,
```
mysticflute: `getRootWorkspacePath` is not exposed to other packages outside of core, so this was the easiest way to have custom `execute` implementations still have access to this value.

packages/salesforcedx-vscode-core/src/notifications/progressNotification.ts
```diff
@@ -35,6 +36,19 @@ export class ProgressNotification {
           execution.processErrorSubject.subscribe(data => {
             return resolve();
           });
+
+          if (progressReporter) {
```
mysticflute: This enables reporting incremental progress updates and (more importantly for the new command) to be able to signify task completion while the child process itself continues to run.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -17,5 +17,27 @@
  */
 export const messages = {
```
mysticflute: @blythesheldon please take a quick look when you get a chance

packages/salesforcedx-vscode-lwc/typings/open/index.d.ts
```diff
@@ -0,0 +1,71 @@
+/// <reference types="node"/>
+import { ChildProcess } from 'child_process';
+
+// IMPORTANT NOTE:
```
mysticflute: basically this file can be removed and the tsconfig change can be undone once the version of typescript is updated.

packages/salesforcedx-vscode-lwc/src/commands/commandUtils.ts
```diff
@@ -0,0 +1,25 @@
+import * as open from 'open';
```
lcampos: Instead of using the `open` module we should use vscode's `openExternal` api.

mysticflute: nice

Issue #1703, opened 2019-10-08
Testing Support for LWC


xyc:
**Is your feature request related to a problem? Please describe.**
The experience of developers testing with LWC can be improved by enhancement to the extensions, as listed below.

**Describe the solution you'd like**
The following features are proposed:
- Run Test / Debug Test / Watch Test UI (W-6739766, W-6765079)
    - CodeLens #1706
    - `editor/title` menu bar actions #1752
- Test Explorer UI (W-6739768) #1752
- Code Coverage Reporting & UI (W-6739773)

Developer Docs: #1765

PR #1808, opened 2019-11-18
Testing Support for LWC: Add more automation coverage

xyc:
### What does this PR do?
Add more code coverage for #1703

Parent issue: #1703


xyc:
There's one test failing in `forceLwcTestRunAction`. Looking into why.

xyc:
Okay that test is fixed. There are some additional properties in task execution (like some private properties added by VS Code `_someProperties`). I updated the match condition and the test passes.

Issue #1626, opened 2019-09-03
Java heap space error caused by ApexIndexer


pibenett:
Hello,

Unfortunately the same issue as #1103 is still exiting. I'm unable to load the @isTest panel and the code completion is not working.
Here is the complete log generated when I tried to run a test from VSCode for example:

`10:16:27.791 [Apex Prelude Service STARTING] INFO  a.j.l.i.s.ApexPreludeManagedService - Patching standard library for System.Database.* methods
10:16:27.800 [Apex Prelude Service STARTING] INFO  a.j.l.i.s.ApexPreludeManagedService - Done patching standard library for System.Database.* methods
10:16:29.740 [Apex Prelude Service STARTING] INFO  a.j.l.i.s.ApexPreludeManagedService - Scanning built-in and system Apex types took 7367 ms.
10:16:29.744 [ApexIndexer STARTING] INFO  a.jorje.lsp.impl.index.ApexIndexer - ApexIndexer: Scanning user-defined types.
sept. 03, 2019 10:18:07 AM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
GRAVE: Internal error: java.lang.reflect.InvocationTargetException
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:192)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:99)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #1817, opened 2019-11-26
add java max heap setting

kenhuman:
### What does this PR do?
Add a setting for JVM max heap.

#1626 
#1103


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -226,6 +226,11 @@
           "default": null,
           "description": "%java_home_description%"
         },
+        "salesforcedx-vscode-apex.java.memory": {
+          "type": "integer",
+          "default": 4096,
```
lcampos: This default is a bit high if we consider the minimum system requirements for running VSCode https://code.visualstudio.com/docs/supporting/requirements. Based on what we've seen on our testing, small to medium projects do well with 1 or 2 GB. It's on larger projects where a higher value would help. @sfsholden @rcoringrato-sfdc what do you think ?

cc @ntotten

kenhuman: @lcampos I agree and am happy to update the PR. Would the following be a resolution:

```suggestion
          "default": 1024,
```

and a similar change at line 37 of requirements.ts?

packages/salesforcedx-vscode-apex/src/requirements.ts
```diff
@@ -31,8 +32,14 @@ export interface RequirementsData {
  */
 export async function resolveRequirements(): Promise<RequirementsData> {
   const javaHome = await checkJavaRuntime();
+  const javaMemory: number = workspace
+    .getConfiguration()
+    .get<number>(JAVA_MEMORY_KEY, 4096);
```
kenhuman:
```suggestion
    .get<number>(JAVA_MEMORY_KEY, 1024);
```

packages/salesforcedx-vscode-apex/package.json
lcampos: Sorry @kenhuman I take back my previous comment. After spending more time testing and going through the Java documentation I think the change should be to exclusively override the JVM defaults if the setting is set and not for us to provide a hardcoded default. This is because the JVM allocates 1/4th of your physical memory. For example, in a machine that has 8GB of RAM the JVM is already allocating 2 for this process.

The purpose of this change is to help users running into heap size issues override the default values and allow those users that are not running into those issues still use the JVM defaults. Let me know what you think.

kenhuman: @lcampos I have updated this PR. The java.memory configuration variable now defaults to null, causing the JVM to use its default value when allocating memory.

lcampos:
@kenhuman I've merged your change as part of #1869 in which I cleaned up the merges and added documentation under the Java Setup section of our doc site. Thanks for your contribution to the project.

Issue #1832, opened 2019-12-09
Indexing Lightning Components is stuck forever


sc5148:
After updating to latest release 47.9.0 every time I open VScode the output window pops up automatically showing a lot of errors under 'Aura Language Server' and 'indexing lightning components' in the status bar below goes on forever. This should be affecting all users, can you please prioritize this?

![image](https://user-images.githubusercontent.com/19307757/70422124-7ed60700-1a6b-11ea-9f21-f13666185194.png)

This is the content of the output window

`Starting *AURA* language server at c:\Users\ChakrabortyS\VSCode Projects\Verivox dev
[Info  - 09:59:16] Starting *AURA* language server at c:\Users\ChakrabortyS\VSCode Projects\Verivox dev
Loading support library: browser
Loading support library: browser
Loading support library: ecmascript
... language server started in 1116.33 ms
Loading support library: ecmascript
[Info  - 09:59:17] ... language server started in 1116.33 ms
Error queuing up indexing of static resources. Error details: TypeError: Cannot read property '0' of undefined
    at Object.indexStaticResources (c:\Users\ChakrabortyS\.vscode\extensions\salesforce.salesforcedx-vscode-lightning-47.9.0\node_modules\lwc-language-server\lib\metadata-utils\static-resources-util.js:55:85)
    at async Promise.all (index 2)
    at async Promise.all (index 0)
    at async Timeout._onTimeout (c:\Users\ChakrabortyS\.vscode\extensions\salesforce.salesforcedx-vscode-lightning-47.9.0\node_modules\aura-language-server\lib\server.js:39:9)
Error queuing up indexing of content resources. Error details: TypeError: Cannot read property '0' of undefined
    at Object.indexContentAssets (c:\Users\ChakrabortyS\.vscode\extensions\salesforce.salesforcedx-vscode-lightning-47.9.0\node_modules\lwc-language-server\lib\metadata-utils\content-assets-util.js:56:83)
    at async Promise.all (index 3)
    at async Promise.all (index 0)
    at async Timeout._onTimeout (c:\Users\ChakrabortyS\.vscode\extensions\salesforce.salesforcedx-vscode-lightning-47.9.0\node_modules\aura-language-server\lib\server.js:39:9)
Error queuing up indexing of labels. Error details: TypeError: Cannot read property '0' of undefined
[...]

PR #1836, opened 2019-12-10
set explicit version for lwc dev server

mysticflute:
### What does this PR do?
explicitly set the dependency for lwc-language-server in the salesforcedx-vscode-lightning project.

#1832


mysticflute:
@nrkruk we should discuss this

PR #1852, opened 2019-12-20
Implement Windows build/test for CircleCI.

rcoringrato-sfdc:
### What does this PR do?

* Add jobs to build and test on Windows using CircleCI

* W-6746995

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,41 +1,168 @@
 #
-version: 2
-jobs:
-  build:
-    docker:
-      - image: circleci/node:10.2.0
-    environment:
-      - SFDX_URL_LINUX=https://developer.salesforce.com/media/salesforce-cli/sfdx-linux-amd64.tar.xz
-      - SFDX_AUTOUPDATE_DISABLE=true
-      - SFDX_USE_GENERIC_UNIX_KEYCHAIN=true
-      - SFDX_DOMAIN_RETRY=300
+# CircleCI configuration for salesforcedx-vscode
+#
+# The following environment variables are required in the 'vscode-test' context
+# when running tests for Windows:
+#     SFDX_CI_DEVHUB_CLIENTID  -- connected app client ID for the dev hub
+#     SFDX_CI_DEVHUB_JWT_KEY   -- private key for the connected app (base64 encoded)
+#     SFDX_CI_DEVHUB_USERNAME  -- username of an authorized user of the dev hub
+#
+version: 2.1
 
+orbs:
+  win: circleci/windows@2.2.0
+
+commands:
+  build-artifacts:
+    description: 'Building and archiving extension artifacts.'
     steps:
       - checkout
       - restore_cache:
-          key: dependency-cache-v3-{{ checksum "package.json" }}
+          key: dependency-cache-v0-{{ checksum "package.json" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,41 +1,168 @@
...
+      - SFDX_AUTOUPDATE_DISABLE: true
+      - SFDX_DOMAIN_RETRY: 300
+      - NODE_VERSION: '10.2.0'
+    parameters:
+      run-unit-tests:
+        type: boolean
+        default: true
+    steps:
+      - build-win
+      - when:
```
lcampos: In which scenarios do we want to run unit tests and no integration tests ?

.circleci/config.yml
```diff
@@ -1,41 +1,168 @@
...
+    steps:
+      - build-artifacts
+
+  run-tests:
+    description: 'Run tests on Windows'
+    executor: win/default
+    environment:
+      - SFDX_AUTOUPDATE_DISABLE: true
+      - SFDX_DOMAIN_RETRY: 300
+      - NODE_VERSION: '10.2.0'
```
lcampos: We might want to add the NPM version too since the jobs are picking up an old version (5.x).

.circleci/config.yml
```diff
@@ -1,41 +1,168 @@
...
+      - store_test_results:
+          path: ./junit-aggregate/
+      - run:
+          name: 'Generate code coverage'
+          shell: bash.exe
+          command: |
+            codecov --disable=gcov
+      - run:
+          name: 'Fail build for test failures'
+          shell: powershell.exe
```
lcampos: Looks like we're using bash everywhere except here, should this be `bash.exe` instead of `powershell.exe` ?

.circleci/config.yml
rcoringrato-sfdc: Unit tests and integration tests are run in parallel jobs.

.circleci/config.yml
rcoringrato-sfdc: I tried to upgrade is using NVM and also with choco, but without success. Its a little more challenging under Windows. I'll open a separate work item to address this.

.circleci/config.yml
rcoringrato-sfdc: We had the powershell script from the AppVeyor configuration, so I just reused it. I'll convert it to bash for consistency.

rcoringrato-sfdc: On second thought the powershell implementation for parsing the XML is simple and reliable. I prefer to keep it as is. Do you have an objection to using PowerShell?

.circleci/config.yml
rcoringrato-sfdc: An alternative implementation would be to use the "parallelism" attribute and  CIRCLE_NODE_INDEX to separate the unit and integration tests. I felt the parameter approach was a little clearer.

.circleci/config.yml
lcampos: I have no preference, let's go with whatever is easier to maintain.

.circleci/config.yml
lcampos: Got it, I just wanted to understand how we would run all the tests. Let's keep it as is and see how it goes, we can make changes later if needed.

PR #1854, opened 2019-12-23
Update vscode-commands.md

garychenming:
### What does this PR do?
Correct the `Org Info` command


lcampos:
Thanks for contributing @garychenming

PR #1857, opened 2019-12-31
Wdxsampletemplate

reidaelliott:
### What does this PR do?
  utilizes the force:analytics:template:create command through vs code

W-6820310 - User Story - [SFDX] Create sample templates for use in projects

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
@@ -38,53 +38,41 @@ export class ForceAnalyticsTemplateCreateExecutor extends BaseTemplateCommand {
   public getFileExtension(): string {
     return '.json';
   }
-  public build(data: TemplateNameAndDir): Command {
+  public build(data: TemplateAndDir): Command {
     return new SfdxCommandBuilder()
       .withDescription(nls.localize('force_analytics_template_create_text'))
       .withArg('force:analytics:template:create')
       .withFlag('--outputdir', data.outputdir)
-      .withFlag('--templatename', data.templateName)
+      .withFlag('--templatename', data.fileName)
       .withLogName('force_analytics_template_create')
       .build();
   }
 
-  public sourcePathStrategy: SourcePathStrategy = PathStrategyFactory.createDefaultStrategy();
+  public sourcePathStrategy: SourcePathStrategy = PathStrategyFactory.createWaveTemplateBundleStrategy();
 
   public getDefaultDirectory() {
     return ANALYTICS_TEMPLATE_DIRECTORY;
   }
-
-  public getPathToSource(outputDir: string, fileName: string) {
-    return super.getPathToSource(outputDir, fileName ? fileName : '');
-  }
 }
 
-export type TemplateNameAndDir = DirFileNameSelection & TemplateName;
+export type TemplateAndDir = DirFileNameSelection & Template;
 
-export interface TemplateName {
-  templateName: string;
+export interface Template {
+  // fileName is the templateName
+  fileName: string;
 }
 
-export class SelectProjectTemplate implements ParametersGatherer<TemplateName> {
-  private readonly prefillValueProvider?: () => string;
-
-  constructor(prefillValueProvider?: () => string) {
-    this.prefillValueProvider = prefillValueProvider;
-  }
-  public async gather(): Promise<
-    CancelResponse | ContinueResponse<TemplateName>
-  > {
+export class SelectProjectTemplate implements ParametersGatherer<Template> {
+  public async gather(): Promise<CancelResponse | ContinueResponse<Template>> {
     const projectTemplateInputOptions = {
-      prompt: 'template name'
+      prompt: nls.localize('force_analytics_template_name_text')
     } as vscode.InputBoxOptions;
-    if (this.prefillValueProvider) {
-      projectTemplateInputOptions.value = this.prefillValueProvider();
-    }
-    const templateName = await vscode.window.showInputBox(
+    const fileName = await vscode.window.showInputBox(
       projectTemplateInputOptions
     );
-    return templateName
-      ? { type: 'CONTINUE', data: { templateName } }
+
+    return fileName
+      ? { type: 'CONTINUE', data: { fileName } }
       : { type: 'CANCEL' };
   }
 }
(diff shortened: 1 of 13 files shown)


packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+export interface TemplateName {
+  templateName: string;
+}
+
+export class SelectProjectTemplate implements ParametersGatherer<TemplateName> {
+  private readonly prefillValueProvider?: () => string;
+
+  constructor(prefillValueProvider?: () => string) {
+    this.prefillValueProvider = prefillValueProvider;
+  }
```
smithgp: Nit: If you want, you can shorten this by just doing the field in the constructor: e.g.:
```
constructor(private readonly prefillValueProvider?: () => string) {}
```

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const outputDirGatherer = new SelectOutputDir(ANALYTICS_TEMPLATE_DIRECTORY);
+
+const parameterGatherer = new CompositeParametersGatherer(
+  new SelectProjectTemplate(),
+  outputDirGatherer
+);
+
+export async function forceAnalyticsTemplateCreate() {
```
smithgp: Nit: you probably don't need the `async` nor the `await` if you just call 1 other async method and return that Promise.

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  public sourcePathStrategy: SourcePathStrategy = PathStrategyFactory.createDefaultStrategy();
+
+  public getDefaultDirectory() {
+    return ANALYTICS_TEMPLATE_DIRECTORY;
+  }
+
+  public getPathToSource(outputDir: string, fileName: string) {
+    return super.getPathToSource(outputDir, fileName ? fileName : '');
```
lcampos: I'm not sure this is needed, can you explain this scenario where filename needs to be set to an empty string ?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -116,6 +116,7 @@ export const messages = {
   force_apex_test_run_all_tests_description_text:
     'Runs all tests in the current project',
 
+  force_analytics_template_create_text: 'SFDX: Add Sample Analytics Template',
```
lcampos: For commands that create metadata files we always use `SFDX: Create ...`

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
reidaelliott: It’s because we don’t use fileName at this point . It wasn’t causing issues, but noticed error in the console.  Filene gets sent in from the framework with undefined otherwise and it types to string.  Suggest another way?

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+  private readonly prefillValueProvider?: () => string;
+
+  constructor(prefillValueProvider?: () => string) {
+    this.prefillValueProvider = prefillValueProvider;
+  }
+  public async gather(): Promise<
+    CancelResponse | ContinueResponse<TemplateName>
+  > {
+    const projectTemplateInputOptions = {
+      prompt: 'template name'
```
brpowell: This label should be placed in the [messages file](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/messages/i18n.ts)

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export type TemplateNameAndDir = DirFileNameSelection & TemplateName;
+
+export interface TemplateName {
+  templateName: string;
+}
+
+export class SelectProjectTemplate implements ParametersGatherer<TemplateName> {
+  private readonly prefillValueProvider?: () => string;
```
brpowell: Is this prefillValueProvider necessary?

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
brpowell: The fileName is the name of the file that's opened once the template is generated. In this case I'm guessing it's the JSON file in the root of the bundle. I'm not sure why this method needs to be changed to public and overridden.

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
```diff
@@ -0,0 +1,106 @@
...
+    return templateName
+      ? { type: 'CONTINUE', data: { templateName } }
+      : { type: 'CANCEL' };
+  }
+}
+
+const outputDirGatherer = new SelectOutputDir(ANALYTICS_TEMPLATE_DIRECTORY);
+
+const parameterGatherer = new CompositeParametersGatherer(
+  new SelectProjectTemplate(),
```
brpowell: Can we reuse the SelectFileName gatherer instead of having this new one? Lines 62-73 in the [lightning interface template](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/commands/templates/forceLightningInterfaceCreate.ts) provide an example.

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
reidaelliott: no its not the same thing as the user inputs, that just ends up being the directory name.  Each template has a template-info.json file maybe I could set the fileName to this file to avoid this.

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
reidaelliott: nope - good catch will remove

packages/salesforcedx-vscode-core/src/commands/templates/forceAnalyticsTemplateCreate.ts
reidaelliott: templatename is different than filename here, its the name of the directory not a file.  wavetemplates are different from other bundle types in that regard

smithgp:
Thanks for merging it, @lcampos !

PR #1862, opened 2020-01-07
Set context correctly when switching to org without alias

brpowell:
### What does this PR do?

Fixes the issue of selecting a scratch org without an alias through the org picker and the org context not being set correctly.

W-6236451


packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -88,8 +88,7 @@ export class OrgAuthInfo {
   public static async getUsername(
     usernameOrAlias: string
   ): Promise<string | undefined> {
-    const username = await Aliases.fetch(usernameOrAlias);
-    return Promise.resolve(username);
+    return (await Aliases.fetch(usernameOrAlias)) || usernameOrAlias;
```
brpowell: If a username doesn't have an alias, it now just returns `usernameOrAlias` which should be a username unless it's undefined.

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceOrgType.test.ts
```diff
@@ -18,22 +18,6 @@ import {
 } from '../../../src/context';
 import { OrgAuthInfo } from '../../../src/util';
 
-describe('getUsername', () => {
```
brpowell: Moved this to `authInfo.test.ts`

packages/salesforcedx-vscode-core/src/util/authInfo.ts
AnanyaJha: This change looks good! I think because this method was returning undefined before, we have this check being done in orgList, metadataCmp, and metadataType too. Should we get rid of the duplicated check?

brpowell: Ah good point, I can rework those a bit.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -22,17 +22,8 @@ export async function getWorkspaceOrgType(
     e.name = 'NoDefaultusernameSet';
     throw e;
   }
-
   const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
-
-  if (isNullOrUndefined(username)) {
```
brpowell: Given line 20, this block will never be entered.

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
```diff
@@ -137,7 +137,7 @@ describe('Filter Authorization Info', async () => {
       )
     ];
     defaultDevHubStub.returns('test-devhub1@gmail.com');
-    getUsernameStub.returns(undefined);
+    getUsernameStub.returns('test-devhub1@gmail.com');
```
brpowell: This emulates getUsername returning the username back to the execution when there isn't an alias.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -22,17 +22,8 @@ export async function getWorkspaceOrgType(
...
-    telemetryService.sendException(
-      'get_workspace_org_type',
-      'workspaceOrgType.getWorkspaceOrgType ran into an undefined username.'
-    );
-  }
-
-  const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username!).catch(err =>
+  const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username).catch(err =>
```
sfsholden: Why was "username!" changed to "username" here?

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceOrgType.test.ts
```diff
@@ -76,9 +60,6 @@ describe('getWorkspaceOrgType', () => {
     const orgType = await getWorkspaceOrgType(defaultUsername);
 
     expect(orgType).to.equal(OrgType.NonSourceTracked);
-    expect(authInfoCreateStub.getCall(0).args[0]).to.eql({
-      username: undefined
```
sfsholden: Should we now expect that the username here is no longer undefined? I'm not certain about removing this assertion altogether.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
brpowell: We don't need to force remove (`undefined|null`), since there isn't a way for it to be `undefined` here.

PR #1873, opened 2020-01-15
Add "inline help text" to generated definition classes

ChuckJonas:
### What does this PR do?

Adds the fields "Inline Help Text" to the generated SObject Definitions.  This provides additional context for a field.

72396067-bf731100-36f8-11ea-8b78-c18b15f25219_png__1704×800_

Screenshot_1_14_20__6_13_PM

Would be great if we could also use field/object descriptions, etc but unfortunately the SObjectDescribe API doesn't return them

None created.  I decided to do this on a whim. Can create if needed


lcampos:
Thanks for opening this PR @ChuckJonas. I think the changes look good, I'll just run some tests to verify something and will then start working on getting it merged for next week's release.

lcampos:
Closing this PR since the changes got merged as part of #1890. Thanks for the contribution @ChuckJonas

PR #1881, opened 2020-01-17
@W-6876798 Update extensions to use vscode-test

KrisGraySFDC:
### What does this PR do?
The vscode team has split up the vscode package into 2, with a new one (vscode-test) being specific for test usage. We should update our repository to use it since it provides some updates to the api's we use for testing.

Removes the 'vscode' dependency and adds 'vscode-test'. Also adds the types and removes the postinstall command as described in the migration steps here:
https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007d9DjIAI/view


KrisGraySFDC:
Probably more work than I can tackle at the moment. Good spike though.

PR #1896, opened 2020-01-22
typescript config change for es2017 support

mysticflute:
### What does this PR do?
update the target and lib configurations to es2017


mysticflute:
waiting on update to replace istanbul with nyc in the vscode integration tests testrunner file.

PR #1900, opened 2020-01-23
Update error message to support java8 and java11

AnanyaJha:
### What does this PR do?
This PR updates the error message that gets displayed when an unsupported version of Java is detected to include Java 8 and Java 11

@W-6267831@


AnanyaJha:
Updated error message:
Screen Shot 2020-01-23 at 1 49 16 PM

PR #1905, opened 2020-01-24
Update docs, publish scripts & nvm to nodejs 12.4.0

lcampos:
### What does this PR do?
Update documents and publishing scripts to use Node 12.4.0

@W-7151404@


scripts/publish.js
```diff
@@ -17,16 +17,31 @@ shell.set('+v');
  *
  */
 
-// Checks that you are running this with Node v10.2.0 and above
+// Checks that you are running this with Node v12.4.0 and above
 const [version, major, minor, patch] = process.version.match(
   /^v(\d+)\.?(\d+)\.?(\*|\d+)$/
 );
-if (parseInt(major) !== 10 || parseInt(minor) < 2) {
+if (parseInt(major) !== 12 || parseInt(minor) < 4) {
```
lcampos: Make sure you are using Node 12.4.0 or above to run the release.

scripts/publish.js
```diff
@@ -17,16 +17,31 @@ shell.set('+v');
...
   console.log(
-    'You do not have the right version of node. We require version 10.2.0.'
+    'You do not have the right version of node. We require version 12.4.0.'
   );
   process.exit(-1);
 }
+// Check if you have installed all the required tooling
```
lcampos: All these dependencies are installed globally but we didn't use to check for them. This will give the user a better error message when the script fails and is caused by one of these dependencies not being installed.

scripts/publish.js
```diff
@@ -17,16 +17,31 @@ shell.set('+v');
  *
```
brpowell: We might want to remove [Line 4](https://github.com/forcedotcom/salesforcedx-vscode/pull/1905/files#diff-e6889d205eee7cf0d8b917c5576051c8L4) because when we run `shell.exec` and if it fails, it will automatically exit without providing the reason why and makes debugging difficult (ran into this problem with Jason). We already check the exitCode and if it's non-zero we do `process.exit`

lcampos: Cool, I'll add that to these changes

PR #1908, opened 2020-01-24
Lightning LSP version update / add integration test suite for lightning and lwc

nrkruk:
### What does this PR do?
- Added integration tests for Lightning and LWC packages
- Update the LSP versions


scripts/download-vscode.js
```diff
@@ -0,0 +1,79 @@
...
+    let vscodeBasePath = path.dirname(path.dirname(vscodeFullPath));
+
+    console.log('Directory Name: ' + vscodeDirname);
+    console.log('Base Path: ' + vscodeBasePath);
+    console.log('Full Path: ' + vscodeFullPath);
+
+    // TODO make this configurable as per package or global?
+
+    // For each extension, copy over the vscode binary
+    for (let i = 0; i < extensionDirectories.length; i++) {
```
lcampos: I know that this resembles how running integration test work right now but did we look at just downloading vscode at a top level location and then just use that vscode to run all the tests ?

scripts/run-vscode-integration-tests-with-top-level-extensions.js
```diff
@@ -8,23 +8,24 @@ const path = require('path');
...
-    cwd,
-    'out',
-    'test',
-    'vscode-integration'
-  )}' node ./node_modules/vscode/bin/test`
-);
+  );
+  const CODE_TESTS_PATH = path.join(cwd, 'out', 'test', 'vscode-integration');
+
+  const CODE_VERSION = '1.41.1';
```
lcampos: Why are we hardcoding the version of vscode to download and run tests with ? Right now we always download the latest stable version. In the past we've hardcoded the vscode version only when the stable version has regressions that cause the tests to fail and remove it once vscode addresses the issue.

nrkruk: yeah were working through how to fix this still -- Its temporary but good catch

scripts/download-vscode.js
nrkruk: I didn't try having them all point to the top level directory, but this is better than what currently happens where we download the binary and extract it in every package. Now it just downloads it once and extracts it everywhere

packages/salesforcedx-vscode-lightning/test/vscode-integration/telemetry/telemetry.test.ts
```diff
@@ -1,71 +1,126 @@
...
-    await telemetryService.sendExtensionDeactivationEvent();
-    assert.calledOnce(sendEvent);
-
-    const expectedData = {
-      extensionName: 'salesforcedx-vscode-lightning'
-    };
-    assert.calledWith(sendEvent, 'deactivationEvent', expectedData);
-  });
-});
+// /*
```
nrkruk: TODO - this doesn't need commented out any longer as its fixed in develop

scripts/download-vscode.js
```diff
@@ -0,0 +1,79 @@
...
+    console.log('Executable Path: ' + executablePath);
+
+    let vscodeIndex = executablePath.indexOf('.vscode-test') + 13;
+
+    // 'vscode-1.41.1'
+    let vscodeDirname = executablePath.substring(
+      vscodeIndex,
+      executablePath.indexOf(path.sep, vscodeIndex)
+    );
+    process.env.CODE_VERSION = vscodeDirname;
```
nrkruk: This doesn't work, we'll have to check the copied directory name to get the version

scripts/download-vscode.js
lcampos: Ok, let's log a work item to try that approach later. No need to change it as part of this PR.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -0,0 +1,251 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { assert, expect } from 'chai';
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+let doc: vscode.TextDocument;
+let editor: vscode.TextEditor;
```
bpbuch: `doc` and `editor` don't need to be defined in the global scope.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    lwcDir = path.join(
+      vscode.workspace.workspaceFolders![0].uri.fsPath,
+      'force-app',
+      'main',
+      'default',
+      'lwc'
+    );
+    await coreExtension.activate();
```
bpbuch: We should try to run the tests without explicitly activating the extensions.

packages/salesforcedx-vscode-lightning/test/vscode-integration/intellisense.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * Test that aura markup intellisense contains aura, lightning, custom aura, custom lwc tags
+   */
+
+  it('Aura markup intellisense', async function() {
+    const docUri = vscode.Uri.file(
+      path.join(auraDir, 'DemoComponent', 'DemoComponent.cmp')
+    );
+    const doc = await vscode.workspace.openTextDocument(docUri);
+    const editor = await vscode.window.showTextDocument(doc);
```
bpbuch: Nit: doc and editor could be initialized in beforeEach.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -0,0 +1,242 @@
...
+ */
+
+import { assert, expect } from 'chai';
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+describe('LWC Intellisense Test Suite', function() {
+  let lwcDir: string;
+
+  before(async function() {
```
bpbuch: This doesn't need to be async.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -0,0 +1,242 @@
...
+          { label: 'c-hello-binding', kind: vscode.CompletionItemKind.Property }
+        ]
+      });
+    } catch (error) {
+      throw error;
+    }
+  });
+
+  it('LWC JS @Salesforce Import Intellisense', async function() {
+    this.timeout(10000);
```
bpbuch: A comment is needed to explain why the timeout is increased.

packages/salesforcedx-vscode-lwc/test/vscode-integration/jsconfig.test.ts
```diff
@@ -0,0 +1,218 @@
...
+const TEST_COMPONENT_NAME = 'testComponent';
+const CREATE_COMPONENT_NAME = 'createComponent';
+
+describe('jsconfig Test Suite', function() {
+  let lwcDir: string;
+
+  let configPath: string;
+  let config: object;
+
+  before(async function() {
```
bpbuch: This function doesn't need to be async.

PR #1848, opened 2019-12-14
Enable support for javadoc symbols

danielepiccone:
### What does this PR do?
This PR enables support for correctly indentation and symbold in javadoc blocks.

### What issues does this PR fix or reference?
https://github.com/forcedotcom/salesforcedx-vscode/issues/1469

Issue #1469, opened 2019-07-11
Multi-line block comments do not insert an asterisk in Apex


erocheleau:
### Summary

When trying to document an Apex class using Javadoc comment format, there is no proper multi-line formatting being triggered or a new asterisk being added on the new line after pressing enter.

What should happen:
```
/**
 *
 */
```

What is currently happening:
```
/**

 */
```

### Steps To Reproduce:

1.  Open an Apex class.
2. Start typing `/**`
3. It will auto-complete with `/**| */` ( | is the cursor)
4. Press Enter.

### Expected result

What should happen:
```
/**
 *
 */
```

### Actual result

What is currently happening:
```
/**

 */
```

### Additional information

I have tried with other extensions that claim to handle multi-line comments for you as well. But none of them did any difference for this issue.

**VS Code Version**: `1.36.0`

**SFDX CLI Version**: `sfdx-cli/7.13.0-27dbcb37d3 win32-x64 node-v10.15.3`

**OS and version**: `WIN 10 x64`

eclarke12:
+1 if we can also get ApexDoc/JavaDoc stub-creation (ie. auto-generate `@params` `@returns` etc that would be a bonus

erocheleau:
@eclarke12 oh yes for sure that would be useful too!

ralphcallaway:
fwiw, if someone is feeling like coding this could be done by cloning an existing vs code java doc extension and expanding the supporting language definitions to include apex. pre-sfdx when i used st3 i used the docblockr extension to this effect and it worked great, but with vs code you have to explicitly list which languages your extension supports which limits the ability to use plugins for this sort of things with languages they weren't explicitly designed for

danielepiccone:
@ralphcallaway thank you for the hint! 
This issue annoyed me for a log time and decided to give it a shot today.
I have opened a PR here with the changes https://github.com/forcedotcom/salesforcedx-vscode/pull/1847/files. 
The code is entirely cloned from the vscode-java extension.

eclarke12:
Thanks @danielepiccone ! Excited to see if this gets accepted.

lcampos:
Closing this since it got released as part of version 47.17.1

erocheleau:
Thank you very much!

PR #1910, opened 2020-01-25
Lcampos/danielepiccone support javadoc blocks

lcampos:
### What does this PR do?
Follow up to PR #1848, this moves the code to the language utility module and adds the OSSREADME file since this code was taken from the vscode repo.

#1469 , @W-7153501@

### DIFF
--- a/packages/salesforcedx-vscode-apex/OSSREADME.json
+++ b/packages/salesforcedx-vscode-apex/OSSREADME.json
@@ -1,0 +1,30 @@
+[
+  {
+    "name": "microsoft/vscode",
+    "license": "MIT",
+    "licenseDetail": [
+      "MIT License",
+      "",
+      "Copyright (c) Microsoft Corporation",
+      "",
+      "Permission is hereby granted, free of charge, to any person obtaining a copy",
+      "of this software and associated documentation files (the \"Software\"), to deal",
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/OSSREADME.json
```diff
@@ -0,0 +1,30 @@
+[
```
lcampos: @ntotten could you verify that I'm not missing something here ? The original code is on https://github.com/Microsoft/vscode/blob/9d611d4dfd5a4a101b5201b8c9e21af97f06e7a7/extensions/typescript/src/typescriptMain.ts#L186

packages/salesforcedx-vscode-apex/src/languageClientUtils/javaDocSymbols.ts
```diff
@@ -0,0 +1,48 @@
+/*---------------------------------------------------------------------------------------------
```
lcampos: Moved to a utility and added the OSS header we normally use in this type of scenario.

packages/salesforcedx-vscode-apex/OSSREADME.json
ntotten: Seems good to me. AFAIK, we just need to include a copy of the license.

PR #1915, opened 2020-01-28
feat: set envar for cli logging

amphro:
### What does this PR do?
Set SFDX_TOOL on all CLI commands invoked by vscode extensions. The CLI will record that envar to appinsights so we can see what percentage of commands are invoked by vscode.

@W-7162538@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
@@ -67,6 +67,7 @@ export class CliCommandExecutor {
           GlobalCliEnvironment.environmentVariables
         )
       : options;
+    this.options.env.SFDX_TOOL = 'salesforce-vscode-extensions';
   }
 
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -67,6 +67,7 @@ export class CliCommandExecutor {
           GlobalCliEnvironment.environmentVariables
         )
       : options;
+    this.options.env.SFDX_TOOL = 'salesforce-vscode-extensions';
```
lcampos: I think this should be enough. What happens when a user is not in the latest version of the CLI that consumes this change ?

amphro: Nothing. The CLI does nothing with this envar currently, so it is just a new op. Once they update to the CLI that has the analytic changes, probably next week, we will start tot get the data.

PR #1916, opened 2020-01-28
Log vscode.env.uiKind as common telemetry property

jgellin-sf:
### What does this PR do?
Logs the `vscode.env.uiKind` (current values are either "Desktop" or "Web") as a common telemetry property. This way we can filter in AppInsights based on whether a telemetry event comes from the Desktop client or Web client.

W-6984859

### DIFF
--- a/packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
+++ b/packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
@@ -130,6 +130,10 @@ export default class TelemetryReporter extends vscode.Disposable {
       commonProperties['common.vscodemachineid'] = vscode.env.machineId;
       commonProperties['common.vscodesessionid'] = vscode.env.sessionId;
       commonProperties['common.vscodeversion'] = vscode.version;
+      if (vscode.env.uiKind) {
+        commonProperties['common.vscodeuikind'] =
+          vscode.UIKind[vscode.env.uiKind];
+      }
     }
     return commonProperties;
   }


packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
```diff
@@ -130,6 +130,10 @@ export default class TelemetryReporter extends vscode.Disposable {
       commonProperties['common.vscodemachineid'] = vscode.env.machineId;
       commonProperties['common.vscodesessionid'] = vscode.env.sessionId;
       commonProperties['common.vscodeversion'] = vscode.version;
+      if (vscode.env.uiKind) {
```
lcampos: `vscode.env.uiKind` seems to be an enum based on the docs and the types, do we need this check ? Everything else looks good.

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
```diff
@@ -130,6 +130,10 @@ export default class TelemetryReporter extends vscode.Disposable {
...
+        commonProperties['common.vscodeuikind'] =
+          vscode.UIKind[vscode.env.uiKind];
+      }
```
jgellin-sf: Logging enum name ("Desktop" or "Web") rather than value (1 or 2) because it is more meaningful.

packages/salesforcedx-vscode-core/src/telemetry/telemetryReporter.ts
jgellin-sf: Probably not. It's an abundance of caution in case VSCode has somehow left this undefined, in which case using the value as a key in the next line might throw an exception. Want me to get rid of this check?

lcampos: No, I just wanted to understand why you added it. I agree with you, let's leave it as is.

jgellin-sf:
If people are happy with this PR, will someone with write access please merge it?

PR #1921, opened 2020-01-30
Conflict detection for manifest operations

rcoringrato-sfdc:
### What does this PR do?

- Beta implementation of manifest based conflict detection.

W-7115316


rcoringrato-sfdc:
More complete test coverage is in progress.

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -173,3 +177,96 @@ export class OverwriteComponentPrompt
...
+      return inputs;
+    }
+    return { type: 'CANCEL' };
+  }
+
+  public async getDefaultUsernameOrAlias(): Promise<string | undefined> {
+    return await OrgAuthInfo.getDefaultUsernameOrAlias(true);
+  }
+
+  public async getDefaultPackageDir(): Promise<string | undefined> {
```
brpowell: Seems like a useful helper that could live in the `SfdxPackageDirectories` util, in case future code needs the default package directory.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+  public static getInstance(): ConflictDetector {
+    if (!ConflictDetector.instance) {
+      ConflictDetector.instance = new ConflictDetector(
+        new CommonDirDirectoryDiffer()
+      );
+    }
+    return ConflictDetector.instance;
+  }
+
+  private readonly relativeMetdataTempPath = path.join(
```
brpowell: Do we want to do like what we did with the diff plugin and utilize `os.temp` so that it's periodically cleared?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+      shell.cp(data.manifest, retrievePackageXmlPath);
+    } catch (error) {
+      console.error(error);
+      channelService.appendLine(
+        nls.localize('error_creating_packagexml', error.toString())
+      );
+      notificationService.showErrorMessage(
+        nls.localize('error_creating_packagexml', error.toString())
+      );
+      return Promise.reject();
```
brpowell: Can we send telemetryExceptions in the different catch blocks of this method as well?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -734,6 +734,11 @@
           "type": "boolean",
           "default": false,
           "description": "%push_or_deploy_on_save_enabled_description%"
+        },
+        "salesforcedx-vscode-core.conflict-detection.enabled": {
```
brpowell: Micro-nit 😄, especially since other settings are doing the dash convention, but can we use camelCase for the property name? Capital letters (except the first letter) are interpreted by the settings UI as word breaks so "conflictDetection" will appear as "Conflict Detection: Enabled" in the UI instead of "conflict-detection: Enabled". Convention documented here: https://code.visualstudio.com/api/references/contribution-points#Configuration-schema

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postconditionCheckers.test.ts
```diff
@@ -283,4 +289,135 @@ describe('Postcondition Checkers', () => {
...
+      expect(response.type).to.equal('CANCEL');
+    });
+
+    it('Should return CancelResponse when a default package directory is not defined.', async () => {
+      const postChecker = new ConflictDetectionChecker(emptyMessages);
+      env.stub(postChecker, 'getDefaultPackageDir').returns(undefined);
+      settingsStub.returns(true);
+
+      const response = await postChecker.check(validInput);
+      expect(response.type).to.equal('CANCEL');
```
brpowell: Is it possible that this could be a CancelResponse because the username is undefined and not because of the undefined packageDir, since the former check happens first?

rcoringrato-sfdc: Yes, good catch. I've updated the tests to verify method calls.

packages/salesforcedx-vscode-core/package.json
rcoringrato-sfdc: I'll make the change. Seems like we can't change the others without effecting functionality?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: Will do.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: The directory is removed when the conflict call is finished.

packages/salesforcedx-vscode-core/package.json
lcampos: We should also remove the `enabled` word from it since we've gotten feedback from our users saying the way the settings UI displays the settings gives the impression the setting is already enabled when it isn't.
Screen Shot 2020-02-04 at 1 47 01 PM

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -49,10 +53,23 @@ export async function forceSourceDeployManifest(manifestUri: vscode.Uri) {
       return;
     }
   }
+
+  const messages: ConflictDetectionMessages = {
+    warningMessageKey: 'conflict_detect_conflicts_during_deploy',
+    commandHint: input => {
+      return new SfdxCommandBuilder()
+        .withArg('force:source:deploy')
+        .withFlag('--manifest', input)
```
lcampos: Need to add `.withLogName('<metric key>')` for this command to generate telemetry.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+      cancellationToken
+    );
+
+    // 5: diff project directory (local) and retrieved directory (remote)
+    // Assume there are consistent subdirs from each root i.e. 'main/default'
+    const diffs = this.differ.diff(localSourcePath, remoteSourcePath);
+
+    // 6: cleanup temp directory
+    try {
+      // shell.rm('-rf', projectMetadataTempPath);
```
lcampos: Is this supposed to be uncommented ?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+  ): Promise<string> {
+    const startTime = process.hrtime();
+    const execution = new CliCommandExecutor(command, options, true).execute(
+      cancellationToken
+    );
+
+    const result = new CommandOutput().getCmdResult(execution);
+    this.attachExecution(execution, cancellationTokenSource);
+    execution.processExitSubject.subscribe(() => {
+      this.logMetric(execution.command.logName, startTime);
```
lcampos: You can directly call `telemetryService.sendCommandEvent` here and remove the `logMetric` method on line 208.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+    const results = await this.checkForConflictsInternal(
+      getRootWorkspacePath(),
+      data,
+      cancellationTokenSource,
+      cancellationToken
+    );
+    return results;
+  }
+
+  private async checkForConflictsInternal(
```
lcampos: The logic here could be broke down into different methods like one that just creates a directory, another that unzips the metadata zip file, etc. This will allow you to write unit tests per method and have a simpler integration test. Similar to how we've done it in other areas like the org browser.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,248 @@
...
+      notificationService.showErrorMessage(
+        nls.localize('error_extracting_org_source', error.toString())
+      );
+      return Promise.reject();
+    }
+
+    // 4: convert org source to decomposed (source) format
+    await this.executeCommand(
+      this.buildMetadataApiConvertOrgSourceCommand(data),
+      { cwd: projectPath },
```
lcampos: We should add `env: { SFDX_JSON_TO_STDOUT: 'true' }` here so that all error messages are surfaced through stdout since them being surfaced through stderr will be deprecated. 
https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_cli_env_variables.htm

Changing this will mean reworking a bit what's on `attachExecution`.

lcampos: Let me correct myself, stderr surfacing errors won't be deprecated but it won't be the default behavior in the near future.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: Commented that out to trouble-shoot the failure on Windows.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -49,10 +53,23 @@ export async function forceSourceDeployManifest(manifestUri: vscode.Uri) {
       return;
     }
   }
+
+  const messages: ConflictDetectionMessages = {
+    warningMessageKey: 'conflict_detect_conflicts_during_deploy',
```
sfsholden: This doesn't matter right now since we don't have the Japanese translation of the message yet, but I think we should change this to:
```suggestion
    warningMessageKey: nls.localize('conflict_detect_conflicts_during_deploy'),
```

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveManifest.ts
```diff
@@ -53,10 +56,23 @@ export async function forceSourceRetrieveManifest(explorerPath: vscode.Uri) {
       return;
     }
   }
+
+  const messages: ConflictDetectionMessages = {
+    warningMessageKey: 'conflict_detect_conflicts_during_retrieve',
```
sfsholden:
```suggestion
    warningMessageKey: nls.localize('conflict_detect_conflicts_during_retrieve'),
```

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
sfsholden: Disregard - I see that we localize the warningMessageKey in postconditionCheckers.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveManifest.ts
sfsholden: Disregard - I see that we localize the warningMessageKey in postconditionCheckers.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
sfsholden: I know we're clearing out the temporary directory we're creating, but I think utilizing the os.temp directory might be safer.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: I tried breaking it out, but I ended up putting it back the way it was. I separated out the error handling to make it a little cleaner. Since I can intercept the calls to executeCommand the integration test isn't too bad. I feel better covering this with an integration test anyway. I'll break out the directory diff-ing into a separate test suite, too.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
rcoringrato-sfdc: This builder is used to generated output to the console, so telemetry is not necessary.

packages/salesforcedx-vscode-core/package.json
rcoringrato-sfdc: I changed it to just "conflictDetection" and the settings UI lost its mind. I'll need to work with this more. Will this setting change when the feature is GA?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: I'll look into using os.temp.

packages/salesforcedx-vscode-core/package.json
brpowell: I'd probably only keep the word enabled if there are other sub-settings under the conflictDetection category. Changing the name of the setting I think would affect people who already have it set.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,218 @@
...
+    notificationService.reportCommandExecutionStatus(
+      execution,
+      cancellationToken
+    );
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+
+  private reportError(messageKey: string, error: Error) {
+    console.error(error);
```
AnanyaJha: are we showing the error on the console here as well as the vs code notification?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: Done

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: Done

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: @AnanyaJha Yes, it shows up in both places. I'll add a test case to verify.

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
rcoringrato-sfdc: Done

packages/salesforcedx-vscode-core/test/integration/testdata/aura/PictureGalleryCard/PictureGalleryCard.cmp
```diff
@@ -0,0 +1,30 @@
+<aura:component implements="flexipage:availableForAllPageTypes"
```
lcampos: We should move these to a new folder/project under `system-tests/assets/`. That's where we already have a couple of test projects for scenarios where we need metadata.

packages/salesforcedx-vscode-core/package.json
lcampos: This setting will probably be around for a while so we should try and pick a name we can stick to for the long run. @hbc-bindu can you help us choosing a name for this ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
lcampos: Got it 👍

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,218 @@
...
+    if (!ConflictDetector.instance) {
+      ConflictDetector.instance = new ConflictDetector(
+        new CommonDirDirectoryDiffer()
+      );
+    }
+    return ConflictDetector.instance;
+  }
+
+  public getCachePath(username: string): string {
+    return path.join(os.tmpdir(), '.sfdx', 'tools', 'conflicts', username);
```
lcampos: We need to add the `username` or something that makes this path unique or this won't work for users running two or more separate projects running conflict detection at once. Or it could also cause incorrect conflicts when running different projects before the temp directory gets properly cleaned.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,218 @@
...
+    // Assume there are consistent subdirs from each root i.e. 'main/default'
+    const localSourcePath: string = path.join(projectPath, data.packageDir);
+    const diffs = this.differ.diff(localSourcePath, convertedSourcePath);
+
+    // 6: cleanup temp directory
+    try {
+      shell.rm('-rf', tempMetadataPath);
+    } catch (error) {
+      // Failed cleanup should not terminate the service
+      this.reportError('error_cleanup_temp_files', error);
```
lcampos: In case this fails to clean up the temp folder, is there a situation where the user could run into false positives caused by old files being around ? I'm thinking we might want to check if the tmp directory is present and delete it before unzipping & converting the code.

packages/salesforcedx-vscode-core/src/conflict/directoryDiffer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // process remote files to generate differences
+    let scannedRemote = 0;
+    this.walkFiles(remoteSourcePath, '', stats => {
+      scannedRemote++;
+      if (localSet.has(stats.relPath)) {
+        const file1 = path.join(localSourcePath, stats.relPath);
+        const file2 = path.join(remoteSourcePath, stats.relPath);
+        if (this.filesDiffer(file1, file2)) {
+          different.add(stats.relPath);
```
lcampos: Eventually we will need the full filepaths of the temp file and the local (workspace) file when there's conflicts so we can pipe them into the visual diff api on vscode. We might need to modify the results returned here but I don't think visual diff is part for the first phase, right ?

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,218 @@
...
+      ),
+      projectPath,
+      cancellationTokenSource,
+      cancellationToken
+    );
+
+    // 5: diff project directory (local) and retrieved directory (remote)
+    // Assume there are consistent subdirs from each root i.e. 'main/default'
+    const localSourcePath: string = path.join(projectPath, data.packageDir);
+    const diffs = this.differ.diff(localSourcePath, convertedSourcePath);
```
lcampos: It would be useful to log how many files where diffed and how many of those had conflicts.

packages/salesforcedx-vscode-core/test/integration/testdata/aura/PictureGalleryCard/PictureGalleryCard.cmp
rcoringrato-sfdc: Will do.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: It will produce false positives, I'll add the cleanup up stream.

packages/salesforcedx-vscode-core/src/conflict/directoryDiffer.ts
rcoringrato-sfdc: The root paths are available where these results are checked, so we do have the full path available there. Visualization is not part of the first phase, we have a meeting next week to discuss what the visualization might look like.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
rcoringrato-sfdc: That information is now available from the differ, I can log that in the postconditionchecker, where I log the filenames.

rcoringrato-sfdc: @lcampos Are you suggesting logging for the users benefit, or using telemetry for metrics?

packages/salesforcedx-vscode-core/package.json
hbc-bindu: Some suggestions:
Manage Conflicts at Sync – most generic – detect, resolve, collect metrics and more
Resolve Conflicts at Sync – detect conflicts and allow resolution of the differences
Detect Conflicts at Sync - detect conflicts of the local/remote differences

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
lcampos: For both, it'd be useful to understand how many files are diffed and how many of those have conflicts. From a user perspective you'd want to know these since the list of file we're currently printing can get to long. From our perspective this will help us tune the feature to make it more performant, etc.

rcoringrato-sfdc: Yes, good idea, I've added the telemetry as well.

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  private reportError(messageKey: string, error: Error) {
+    console.error(error);
+    channelService.appendLine(nls.localize(messageKey, error.toString()));
+    notificationService.showErrorMessage(
+      nls.localize(messageKey, error.toString())
+    );
+    telemetryService.sendException(
+      'ConflictDetectionException',
+      error.toString()
```
brpowell: Does this only include the error message?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -476,6 +476,25 @@ export const messages = {
...
   XOrgHub: 'X Org Hubs',
 
+  conflict_detect_error:
+    'An error was encountered during conflict detection. %s',
+  conflict_detect_retrieve_org_source:
+    'Conflict Detection: retrieving org source',
+  conflict_detect_convert_org_source:
+    'Conflict Detection: converting org source',
+  conflict_detect_conflicts_during_deploy:
+    'Conflicts were detected during the metadata deploy',
```
brpowell: Small thing but do you think the modal should say something about the conflicting files being shown in the output view?

PR #1927, opened 2020-02-02
[W-7174475] document linking integration tests

bpbuch:
### What does this PR do?
Added integration tests for document linking
Included viewSource and contactList from lwc-recipes
Updated sinon sandbox creation

@W-7174475@

### Functionality Before
N/A

### Functionality After
Increased test coverage


packages/salesforcedx-vscode-lightning/test/vscode-integration/hover.test.ts
```diff
@@ -81,16 +81,27 @@ describe('Aura Hovers', function() {
...
+/** Helper to find the expected hover content
+ *
+ * @param instances - hover instances
+ * @param expectedContent - content that is being searched for
+ */
+function findContentFromInstances(instances: Hover[], expectedContent: string) {
+  for (const instance of instances) {
+    const contents = instance!.contents as MarkdownString[];
+
+    return contents.find(content => content.value.includes(expectedContent));
```
mysticflute: not sure I completely understand, but if there is more than once `instance` and it's not in the order you expect, wouldn't this return undefined? Seems like you might want to do instances.find

packages/salesforcedx-vscode-lwc/test/vscode-integration/hover.test.ts
```diff
@@ -78,18 +78,31 @@ describe('LWC Hovers', function() {
       position
     )) as Hover[];
 
-    expect(hoverInstances).to.have.lengthOf.at.least(1);
+    console.log(JSON.stringify(hoverInstances));
```
mysticflute: assuming this is just for temporary testing

packages/salesforcedx-vscode-lightning/test/vscode-integration/navigation.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+
+    await new Promise(r => setTimeout(r, 1500));
+  });
+
+  afterEach(async function() {
+    await commands.executeCommand('workbench.action.closeActiveEditor');
+  });
+
+  it('Should provide navigation to a selected LWC tag', async function() {
```
mysticflute: are other test cases coming later? like navigation to a selected Aura tag?

packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testExplorer/testOutlineProvider.test.ts
```diff
@@ -158,8 +158,8 @@ describe('LWC Test Outline Provider', () => {
     });
 
     it('Should provide test file nodes and test cases nodes', async () => {
-      expect(actualFileNodes.length).to.equal(1);
-      const actualFileNode = actualFileNodes[0];
+      expect(actualFileNodes.length).to.equal(3);
```
mysticflute: why are these changing?

packages/system-tests/assets/lwc-recipes/force-app/main/default/lwc/contactList/__tests__/contactList.test.js
```diff
@@ -0,0 +1,105 @@
+import { createElement } from 'lwc';
```
mysticflute: these test files won't get picked up, right?

packages/system-tests/assets/lwc-recipes/force-app/main/default/lwc/contactList/contactList.html
```diff
@@ -0,0 +1,21 @@
+<template>
```
mysticflute: should we have our own version of test files instead of sharing with system-tests? just worried about depending on another package and potential noise later

bpbuch: I agree.  We definitely should change the way the integration tests pick up sample workspaces.  This change should be included with the multi-root refactor.

packages/system-tests/assets/lwc-recipes/force-app/main/default/lwc/contactList/__tests__/contactList.test.js
bpbuch: Correct.

packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testExplorer/testOutlineProvider.test.ts
bpbuch: These test cases are dependent on the order and amount of files in the sample workspace.  They need to be updated to look up the specific file that they are using.

packages/salesforcedx-vscode-lightning/test/vscode-integration/navigation.test.ts
bpbuch: Yes. I need to create/find a sample with Aura to Aura linking.

packages/salesforcedx-vscode-lightning/test/vscode-integration/hover.test.ts
bpbuch: There shouldn't be multiple instances with the content that is being searched for, so I just grab the first content match that is found.

mysticflute: maybe I'm missing something, but I meant if the first instance doesn't have the content then it wouldn't make it to the second instance

mysticflute: so the for loop here is a little confusing

bpbuch: ah. Nice catch.

PR #1936, opened 2020-02-06
Fix size of the vsix for the extension pack

lcampos:
### What does this PR do?
This addresses the vsix size issue noticed after including gifs on the current changelog. It also breaks up the changelog per version and only bundles the current one.

@W-7197900@


packages/salesforcedx-vscode/.vscodeignore
```diff
@@ -1 +1,4 @@
-images/overview.gif
\ No newline at end of file
+images/**/*
```
lcampos: Only ship `images/VSCodeBundle.png` and ignore all the previous changelogs

PR #1935, opened 2020-02-06
 Fix Command Palette not allowing empty alias

FabienTaillon:
### What does this PR do?
* Fix Command Palette not allowing empty alias
* Updated Japanese translation for #1876 (thanks @shunkosa for the updated translation)

### What issues does this PR fix or reference?
* #1929 
* @W-7147786@

Issue #1929, opened 2020-02-05
Creating Scratch Org via Command Palette won't let us use default alias


FabienTaillon:
### Summary

Using the Command Palette to create a Scratch Org will force us to write a Scratch Org alias, default alias doesn't work anymore.

### Steps To Reproduce:

1.  Open the Command Palette
2.  Select `SFDX: Create a Default Scratch Org...`
3. Press Enter on all questions to use default values

### Expected result

The command uses the default value if we don't write any Scratch Org alias.

### Actual result

We're getting the following error: `Alias can only contain underscores, spaces and alphanumeric characters`.

### Additional information

![fielddependencies](https://user-images.githubusercontent.com/2465122/73853168-b9bca700-4830-11ea-8fc2-91ef9a00ac67.png)
Most likely linked to this [recent change](https://github.com/forcedotcom/salesforcedx-vscode/commit/b8e23f62a869045c375b05de20827a238424e84d).

**VS Code Version**:
1.41.1

**Salesforce Extension Pack Version**:
47.17.1

**SFDX CLI Version**:
@oclif/plugin-autocomplete 0.1.5
@oclif/plugin-commands 1.2.3 (core)
@oclif/plugin-help 2.2.3 (core)
@oclif/plugin-not-found 1.2.3 (core)
@oclif/plugin-plugins 1.7.9 (core)
@oclif/plugin-update 1.3.9 (core)
@oclif/plugin-warn-if-update-available 1.7.0 (core)
@oclif/plugin-which 1.0.3 (core)
@salesforce/sfdx-trust 3.0.7 (core)
analytics 1.4.1 (core)
generator 1.1.2 (core)
salesforcedx 47.17.0 (core)
├─ salesforcedx-templates 47.18.0 (core)
└─ salesforce-alm 47.15.0 (core)
sfdx-cli 7.44.0 (core)

**OS and version**:
macOS Catalina
Windows 10

vazexqi:
Tracked internally as W-7147786

sfsholden:
Closing. This issue was resolved in #1953

PR #1953, opened 2020-02-08
Fix scratch org create command not allowing empty string during input

brpowell:
### What does this PR do?

Fixes the input validation when entering an org alias to not prevent a user from entering an empty string. This is a follow up PR to #1935 and contains commits contributed by @FabienTaillon in order to run the builds against the external contribution. It also adds tests for the fix and updates the expiration days input to behave the same way.

#1929, @W-7147786@

### Functionality Before
before

### Functionality After
after


packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
```diff
@@ -286,7 +286,7 @@ export const messages = {
   empty_components: '利用できるコンポーネントがありません',
   error_auth_token: '認証トークン更新中にエラーが発生しました。',
   error_no_org_found: '組織の認証情報が見つかりませんでした。',
-  error_invalid_org_alias: '別名にはアンダースコアと英数字のみを使用できます。',
+  error_invalid_org_alias: '別名にはアンダースコア、スペースと英数字のみを使用できます。',
```
sfsholden: Nice, katana for space?

FabienTaillon: Yes, @shunkosa did the translation :)

PR #1954, opened 2020-02-08
Publishing script that uses the artifacts from CircleCI

lcampos:
### What does this PR do?
Updates the publishing process to use the vsix files generated by the CircleCI builds

@W-7188372@


scripts/publish-circleci.js
```diff
@@ -0,0 +1,121 @@
...
+
+// Generate the SHA256 and append to the file
+shell.exec(`npm run vscode:sha256`);
+
+// Concatenate the contents to the proper SHA256.md
+shell.exec('./scripts/concatenate-sha256.js');
+
+// Remove the temp SHA256 file
+shell.rm('./SHA256');
+
```
rcoringrato-sfdc: What do you think about a **--dryrun** option that performs all of the local operations, but doesn't commit and publish?

lcampos: Is that because we want to be able to test the publish script or something different ?

rcoringrato-sfdc: Yes, for testing.

lcampos: Let's talk more about that and I can address it as a follow up PR. I think having a dryrun option in this script means to exclusively download the vsix files in the correct location, which I don't think would be a good way to test it and will make the script more complex.

scripts/download-vsix-from-circleci.js
```diff
@@ -0,0 +1,101 @@
...
+    {
+      silent: true
+    }
+  )
+  .stdout.trim();
+
+const buildArtifactsJSON = JSON.parse(cciArtifacts);
+if (buildArtifactsJSON && buildArtifactsJSON.length === 0) {
+  console.log(
+    `Looks like the CircleCI build number ${circleciBuild} did not generate any artifacts`
```
sfsholden: I think we should include here that the build the user wants for getting the artifacts is 'build-all'.

scripts/download-vsix-from-circleci.js
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env node
+
+const shell = require('shelljs');
+shell.set('-e');
+shell.set('+v');
+
+const { getVsixName, getLocalPathForDownload } = require('./publish-utils');
+const CIRCLECI_API_URI = 'https://circleci.com/api/v1.1';
+const circleciToken = process.env['CIRCLECI_TOKEN'];
```
sfsholden: We might want to include instructions on what these are and how to set them for new users that haven't set up circleci-to-vscode.

scripts/publish-circleci.js
```diff
@@ -0,0 +1,121 @@
...
+  const [version, major, minor, patch] = nextVersion.match(
+    /^(\d+)\.(\d+)\.(\d+)$/
+  );
+  const currentBranch = shell
+    .exec('git rev-parse --abbrev-ref HEAD', {
+      silent: true
+    })
+    .stdout.trim();
+
+  if (!currentBranch !== releaseBranchName) {
```
sfsholden: Is this supposed to be currentBranch !== releaseBranchName?

scripts/download-vsix-from-circleci.js
```diff
@@ -0,0 +1,101 @@
...
+const circleciBuild = process.env['CIRCLECI_BUILD'];
+const nextVersion = process.env['SALESFORCEDX_VSCODE_VERSION'];
```
brpowell: Is there a reason we use environment variables for these values? They change on a weekly basis (or even more if we need to hotfix). If we were to eventually utilize this script in an automated environment we'd need to change these in some settings portal every time.

sfsholden: That is a good point. Why do we need SALESFORCEDX_VSCODE_VERSION when circleci-to-vscode doesn't?

scripts/publish-utils.js
```diff
@@ -0,0 +1,13 @@
...
+const path = require('path');
+
+module.exports = {
+  getVsixName: data => {
+    return data.replace('extensions/', '');
+  },
+
+  getLocalPathForDownload: (fileName, pkgVersion) => {
+    const pkgName = fileName.replace(`-${pkgVersion}.vsix`, '');
+    return path.join('packages', pkgName, fileName);
```
brpowell:
```suggestion
    return path.join(__dirname, '..', 'packages', pkgName, fileName);
```

scripts/download-vsix-from-circleci.js
lcampos: It's true that this script is not optimized for running on an automation environment (e.g. GH Actions) but I think those updates should be done once we start working on that automation piece. Doing it right now would be to optimize for something we don't exactly know how we'll structure.

The reason we use `SALESFORCEDX_VSCODE_VERSION ` here and not on the circleci-to-vscode utility is because we want to validate this is executed in the correct branch since the publish process makes updates the files (e.g. sha generation) that need to be checked in and merged.

scripts/publish-circleci.js
lcampos: Yeah, good catch

scripts/publish-circleci.js
lcampos: I added a new npm script called `circleci:artifacts` that just downloads the vsix files in the correct location. Would that be enough to address this ?

rcoringrato-sfdc: Yes it does, thanks.

scripts/download-circleci-vsix.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  for (let i = 0; i < jsonData.length; i++) {
+    const vsixName = getVsixName(jsonData[i]);
+    const optsDownload = buildVsixUrl(jsonData[i]);
+    console.log('vsix ===> ', vsixName);
+    download(vsixName, optsDownload)
+      .then(resultArray => {
+        console.log('download was successful', resultArray);
+        // extract(resultArray[0], resultArray[1], isInsiders);
```
jag-sfdc: Can this commented code be removed?

lcampos: I ended up removing this file and replacing it with different code in the latest commits

scripts/download-vsix-from-circleci.js
sfsholden: Just an idea: We could turn this into a task that accepts the vscode version as an input. The default value could be the latest vscode version. This would still keep the check, but easily allow for an override.

PR #1969, opened 2020-02-11
Add custom labels to supported metadata

hbc-bindu:
### What does this PR do?

Add Custom Labels to the list of supported metadata for Source Diff.

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/docs/_articles/en/user-guide/source-diff.md
+++ b/docs/_articles/en/user-guide/source-diff.md
@@ -28,6 +28,8 @@ Source Diff currently supports the following metadata:
 - Aura Interface
 - Aura Token
 - Custom Labels
+- Layout
 - Lightning Web Component
+- Permission Sets
 - Visualforce Component
 - Visualforce Page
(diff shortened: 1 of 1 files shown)


docs/_articles/en/user-guide/source-diff.md
```diff
@@ -27,6 +27,7 @@ Source Diff currently supports the following metadata:
 - Aura Event
 - Aura Interface
 - Aura Token
+- Custom Labels
```
lcampos: Very timely PR since I just noticed we forgot to update this. We should also add Layouts and Permission sets to it.

smaddox-sf: oooo and this is a very timely comment @lcampos  because I'm working on a blog post.  I will add the Layouts & permission sets to the post.  (I had the custom labels, which is why I noticed the missing doc :) )

hbc-bindu: Updated the supported metadata list with Layout and Permission Sets.

PR #1973, opened 2020-02-12
Update apex-tmlanguage dependency

lcampos:
### What does this PR do?
Updates the apex-tmlanguage dependency to get the latest syntax highlighting updates. Below is a list of updates.
- SOQL when clause highlighting
- Switch statements with multiple string clauses
- JavaDoc support
- Missing apex class highlighting

@W-6532166@, @W-7210864@, @W-5973555@, 

### Functionality Before
SOQL
Screen Shot 2020-02-12 at 3 28 26 PM

JavaDoc
javadoc-before

Apex keyword
system-class-before

Strings on switch
switch-string-before

### Functionality After
SOQL
Screen Shot 2020-02-12 at 3 27 40 PM

JavaDoc
javadoc-after

Apex keyword
system-class-after

Strings on switch
Screen Shot 2020-02-12 at 3 09 17 PM

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -24,7 +24,7 @@
     "Programming Languages"
   ],
   "dependencies": {
-    "@salesforce/apex-tmlanguage": "1.2.0",
+    "@salesforce/apex-tmlanguage": "1.3.0",
     "@salesforce/core": "2.0.1",
     "@salesforce/salesforcedx-sobjects-faux-generator": "47.17.1",
     "@salesforce/salesforcedx-utils-vscode": "47.17.1",


sfsholden:
Preview of before and after on 'Strings on switch' look like they are the same.

lcampos:
> Preview of before and after on 'Strings on switch' look like they are the same.

Yeah, I should've probably have the token view for that scenario. If you take a closer look the quotes in the `Customer No Show` string are white because it is getting incorrectly tokenized.

sfsholden:
> > Preview of before and after on 'Strings on switch' look like they are the same.
> 
> Yeah, I should've probably have the token view for that scenario. If you take a closer look the quotes in the `Customer No Show` string are white because it is getting incorrectly tokenized.

Ah, I see that now. Thank you

PR #1980, opened 2020-02-13
Reenable Telemetry tests and fix issue with the suite not running

KrisGraySFDC:
### What does this PR do?
Brings back the Telemetry tests which were causing the whole suite to not run.

@W-7171718@

### Functionality Before
Tests were commented out.

### Functionality After
Tests are now running during the integration test pass.
Also removed the circular reference between the telemetry class and the instantiation of the singleton.


packages/salesforcedx-vscode-lightning/src/telemetry/telemetry.ts
```diff
@@ -8,7 +8,7 @@
 import * as util from 'util';
 import * as vscode from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
-import { telemetryService } from '.';
+// import { telemetryService } from '.';
```
nrkruk: No idea how this happened, I feel like this was a bad merge somewhere along the line - dunno how it was working before

PR #1987, opened 2020-02-14
Fix broken write-aura doc links

lcampos:
### What does this PR do?
Fixes sidebar and article aura links that are broke

@W-7226967@


AnanyaJha:
The numbering for the suggestions under`Quick Fix: Declare Missing Methods` is off, but links seem fine!

PR #2003, opened 2020-02-19
Update publish scripts

lcampos:
### What does this PR do?
This PR contains changes to the new publish scripts:
- Add validation for SALESFORCEDX_VSCODE_VERSION
- Update file perms
- Create & push a git tag
- Deprecate previous publish script

@W-7237842@


scripts/download-vsix-from-circleci.js
```diff
@@ -66,6 +66,14 @@ if (
   process.exit(-1);
 }
 
+// Check release version environment variable
+if (!nextVersion.match(/^(\d+)\.(\d+)\.(\d+)$/)) {
```
lcampos: This validation is also in the `publish-circleci.js` script, that's because we can call both scripts independently.

PR #2015, opened 2020-02-20
[W-7241114] Add GitHub Action to upgrade language servers

bpbuch:
### What does this PR do?
Adds a GitHub Action that checks daily for language server updates and opens a PR per extension(lwc or lightning) to upgrade if necessary. 

@W-7241114@

### Functionality Before
Dependabot would open a PR for each individual dependency, but this would not work when multiple dependencies needed to be updated in the same PR.

### Functionality After
Upgrade PRs for grouped language server packages per extension

### DIFF
--- a/.github/workflows/check-dependencies.yml
+++ b/.github/workflows/check-dependencies.yml
@@ -1,0 +1,32 @@
+name: Check Dependencies
+
+on:
+  schedule:
+    - cron: '00 0 * * *'
+
+jobs:
+  check-language-servers:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+          - salesforcedx-vscode-lwc
+          - salesforcedx-vscode-lightning
+    steps:
+      - uses: actions/checkout@master
+      - name: set remote url
+        run: git remote set-url --push origin https://$GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
+      - name: package-update
+        uses: taichi/actions-package-update@master
+        env:
+          AUTHOR_EMAIL: actions@github.com
+          AUTHOR_NAME: github-actions
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EXECUTE: true
+          BRANCH_PREFIX: ${{ matrix.package }}/
+          COMMIT_MESSAGE: Upgrade Langauge Servers for ${{ matrix.package }}
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-dependencies.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: set remote url
+        run: git remote set-url --push origin https://$GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
+      - name: package-update
+        uses: taichi/actions-package-update@master
+        env:
+          AUTHOR_EMAIL: actions@github.com
+          AUTHOR_NAME: github-actions
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
lcampos: The current github token provides more access than the one required by this action (creating pull requests). We should change this to use a token that only grants pull request creation access.

bpbuch: Would you be able to add a less powerful token to the repo secrets and provide me with the name?

.github/workflows/check-dependencies.yml
```diff
@@ -0,0 +1,32 @@
...
+      matrix:
+        package:
+          - salesforcedx-vscode-lwc
+          - salesforcedx-vscode-lightning
+    steps:
+      - uses: actions/checkout@master
+      - name: set remote url
+        run: git remote set-url --push origin https://$GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
+      - name: package-update
+        uses: taichi/actions-package-update@master
```
mysticflute: how much work is it to do something similar to this ourselves? I'd prefer not to use a 3rd party github action.

If we do use this we need to get a security review and we should also point to a specific tag and not @`master`, which I'm assuming is dynamic. If we could do something ourselves though I'd lean towards that. It doesn't have to be fancy.

lcampos:
@bpbuch @mysticflute let's talk to security before we make more changes to this

.github/workflows/check-dependencies.yml
bpbuch: I have the same concerns as you.  I don't think it would take more than a couple days to create a similar action.

If we decide to use this 3rd party action, I agree that we should set a specific version.

lcampos:
Closing this since we don't have approval to use actions yet. We will revisit this in the future.

PR #2031, opened 2020-02-24
Sinon types lwc

KrisGraySFDC:
### What does this PR do?
Upgrade the Sinon types for lwc and lightning

I only did the lightning packages as doing them all turned into a very long process. There's about 1200 that need converting and some require code changes. Overall, its probably 200+ files and 2000+ code line changes to make happen. 

So we should probably do it incrementally by package. 

@W-7247400@

### Functionality Before
@types/sinon was ^2.3.2

### Functionality After
Now @types/sinon is ^7.5.1 which is the latest.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testRunner/taskService.test.ts
+++ b/packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testRunner/taskService.test.ts
@@ -13,9 +13,22 @@ import {
 } from '../../../../src/testSupport/testRunner/taskService';
 
 describe('Task Service Unit Tests', () => {
-  let executeTaskStub: SinonStub;
-  let onDidStartTaskStub: SinonStub;
-  let onDidEndTaskStub: SinonStub;
+  let executeTaskStub: SinonStub<
+    [vscode.Task],
+    Thenable<vscode.TaskExecution | void>
+  >;
+  let onDidStartTaskStub: SinonStub<
+    [
+      (e: vscode.TaskStartEvent) => any,
+      any?,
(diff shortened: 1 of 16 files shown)


packages/system-tests/assets/lwc-recipes/force-app/main/default/lwc/jsconfig.json
```diff
@@ -2,7 +2,23 @@
     "compilerOptions": {
         "experimentalDecorators": true,
         "baseUrl": ".",
-        "paths": {}
```
KrisGraySFDC: Not sure why this got updated. I'll remove it.

packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testRunner/taskService.test.ts
```diff
@@ -13,9 +13,22 @@ import {
 } from '../../../../src/testSupport/testRunner/taskService';
 
 describe('Task Service Unit Tests', () => {
-  let executeTaskStub: SinonStub;
-  let onDidStartTaskStub: SinonStub;
-  let onDidEndTaskStub: SinonStub;
+  let executeTaskStub: SinonStub<
```
mysticflute: these types look a bit painful to maintain. wondering if we can simplify this somehow, especially if the editor isn't going to tell you when they are wrong

KrisGraySFDC: So when it reports the error, it does tell you what the value should become. I just couldn't get the quick fix to address it.

PR #2041, opened 2020-02-27
Create release branch action

brpowell:
### What does this PR do?

Adds a GitHub action for running the release branch creation script. Triggered with a [repository_dispatch](https://developer.github.com/v3/repos/#create-a-repository-dispatch-event)

### DIFF
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -1,0 +1,46 @@
+name: Create Release Branch
+
+on:
+  repository_dispatch:
+    types: create_release_branch
+
+jobs:
+  create_branch:
+    name: 'Create Branch'
+    
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+    - name: Checkout
+      uses: actions/checkout@v2.0.0
+      with:
+        ref: 'develop'
+    - name: Install Script Dependencies
+      run: npm install shelljs && npm install -g lerna
+    - name: Create and Push the Release Branch
+      run: |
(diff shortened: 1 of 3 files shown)


scripts/create-release-branch.js
```diff
@@ -17,7 +17,7 @@ const releaseBranchName = `release/v${nextVersion}`;
 // Check if release branch has already been created
 const isRemoteReleaseBranchExist = shell
   .exec(
-    `git ls-remote --heads git@github.com:forcedotcom/salesforcedx-vscode.git ${releaseBranchName}`,
+    `git ls-remote --heads origin ${releaseBranchName}`,
```
brpowell: Give the `origin` remote alias instead since that has all the authentication needed, important for the action and still works locally.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,46 @@
...
+        ref: 'develop'
+    - name: Install Script Dependencies
+      run: npm install shelljs && npm install -g lerna
+    - name: Create and Push the Release Branch
+      run: |
+        git config --local user.email "releaseaction@github.com"
+        git config --local user.name "Release Action"
+        node scripts/create-release-branch.js
+      env:
+        SALESFORCEDX_VSCODE_VERSION: ${{ github.event.client_payload.version }}
```
brpowell: When a repository_dispatch is created, the payload sent should include a `version` key in the format `major.minor.patch`

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,46 @@
+name: Create Release Branch
+
+on:
```
lcampos: Let's add documentation on how we will trigger this. Adding it somewhere under the `contributing` folder would be good.

PR #2049, opened 2020-03-02
removed beta setup requirements

mlregal:
since the feature is no longer in beta, the extra setup steps are not required, so I removed them from the document.

### What does this PR do?
Removed extra setup steps that are not required since this is no longer in beta

### Functionality Before
Users were directed to locate and check a box that no longer exists in VS Code

### Functionality After
Users are no longer directed to check a box that doesn't exist


lcampos:
Hi @mlregal thanks for opening this PR. At the moment we can't merge it because the Conflict Detection feature is still in beta behind the `salesforcedx-vscode-core.detectConflictsAtSync` setting. Are you referring to this setting in your comments or a different one ?

lcampos:
I'm closing this since the feature is still in beta. We'll update the docs once the status of the feature changes.

PR #2052, opened 2020-03-04
Single Apex Class file deploys through Tooling API

AnanyaJha:
### What does this PR do?
This covers the changes needed to use the Tooling API for single file Apex Class deploys through the editor/explorer context right click and command palette.

@W-6531270@


packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+  ToolingCreateResult,
+  ToolingRetrieveResult
+} from './index';
+// tslint:disable-next-line:no-var-requires
+const DOMParser = require('xmldom-sfdx-encoding').DOMParser;
+
+export class ToolingDeploy {
+  private static instance: ToolingDeploy;
+  private metadataType?: string;
+  private connection?: Connection;
```
AnanyaJha: thinking about making these public in case users don't want to call the deploy() method and want more granular control over the operations

packages/salesforcedx-vscode-core/package.json
```diff
@@ -53,7 +53,8 @@
     "nyc": "^13",
     "sinon": "^7.3.1",
     "typescript": "3.1.6",
-    "vscode": "^1.1.36"
+    "vscode": "1.1.17",
+    "xmldom-sfdx-encoding": "0.1.29"
```
AnanyaJha: This is the library I saw force-com-toolbelt using for xml parsing

packages/salesforcedx-vscode-core/package.json
```diff
@@ -53,7 +53,8 @@
     "nyc": "^13",
     "sinon": "^7.3.1",
     "typescript": "3.1.6",
-    "vscode": "^1.1.36"
+    "vscode": "1.1.17",
```
lcampos: This should still be `^1.1.36`

packages/salesforcedx-vscode-core/package.json
```diff
@@ -749,6 +750,11 @@
           "type": "boolean",
           "default": false,
           "description": "%conflict_detection_enabled_description%"
+        },
+        "salesforcedx-vscode-core.tooling-deploys": {
```
lcampos: We should rename this to something more generic that we can use for the rest of the changes we'll be doing so that everything fits under one setting

packages/salesforcedx-vscode-core/src/commands/forceSourceToolingDeploy.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  public async execute(response: ContinueResponse<string>): Promise<void> {
+    const startTime = process.hrtime();
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
+    const executionWrapper = new CompositeCliCommandExecutor(
+      this.build(response.data)
+    ).execute(cancellationToken);
```
brpowell: It looks like the CLI command is still being built and executed on top of running the tooling deploy logic

packages/salesforcedx-vscode-core/src/commands/forceSourceToolingDeploy.ts
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    try {
+      const deployLibrary = new ToolingDeploy();
+      const usernameOrAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(true);
+      let username: string | undefined;
+      if (usernameOrAlias) {
+        username = await OrgAuthInfo.getUsername(usernameOrAlias);
+      }
+      await deployLibrary.init(username!);
```
brpowell: Instead of having an `init` method, can we just pass the username to the constructor on line 55? The connection can lazily be created when trying to do the deploy if it doesn't exist yet. Removes the extra step of having to explicitly initialize.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  public async toolingRetrieve(): Promise<ToolingRetrieveResult> {
+    try {
+      let retrieveResult = (await this.connection!.tooling.retrieve(
+        'ContainerAsyncRequest',
+        this.asyncRequest!.id
+      )) as ToolingRetrieveResult;
+      let count = 0;
+      while (retrieveResult.State === 'Queued' && count <= 20) {
+        await this.sleep(100);
```
lcampos: We should take a closer look at why we need this sleep. Maybe we can replace this with a pattern that is more reliable.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+  private asyncRequest?: ToolingCreateResult;
+
+  public constructor() {}
+
+  public static getInstance(): ToolingDeploy {
+    if (!this.instance) {
+      this.instance = new ToolingDeploy();
+    }
+    return this.instance;
+  }
```
brpowell: I'm not sure we need this

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+  public async init(username: string, apiVersion?: string) {
+    try {
+      const connection = await Connection.create({
+        authInfo: await AuthInfo.create({ username })
+      });
+      this.connection = connection;
+      if (apiVersion) this.apiVersion = apiVersion;
+    } catch (e) {
+      throw e;
+    }
```
brpowell: This try-catch isn't needed, the error will chain up automatically.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        await this.createMetadataContainer();
+        await this.createContainerMember([sourcePath, metadataPath]);
+        await this.createContainerAsyncRequest();
+        const output = await this.toolingRetrieve();
+        return output;
+      }
+    } catch (e) {
+      throw e;
+    }
```
brpowell: [Ditto](https://github.com/forcedotcom/salesforcedx-vscode/pull/2052/files#r387847277)

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
brpowell: I think it's due to the fact that the deploy is being done with a ContainerAsyncRequest which requires polling the status

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -71,10 +75,22 @@ export async function forceSourceDeploySourcePath(sourceUri: vscode.Uri) {
...
+
+  const toolingDeployEnabled = sfdxCoreSettings.getToolingDeploys();
+  // this supported types logic is temporary until we have a way of generating the metadata type from the path
+  // once we have the metadata type we can check to see if it is a toolingsupportedtype from that util
+  const supportedType =
+    path.extname(sourceUri.fsPath) === APEX_CLASS_EXTENSION ||
+    sourceUri.fsPath.includes(`${APEX_CLASS_EXTENSION}-meta.xml`);
+  const executor =
+    toolingDeployEnabled && supportedType
+      ? new DeployRetrieveExecutor()
```
brpowell: I'm not sure we need an entirely new CommandletExecutor for this. Can we perform the tooling deploy logic in the `ForceSourceDeploySouthPathExecutor`? You might be able to just define the execute() method in `ForceSourceDeploySourcePathExecutor` where you check to see if the type is supported for tooling, and if it is run the tooling logic. Otherwise call `super.execute()` to do the fallback CLI way.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+      const deployFailed = new Error();
+      if (!containerMember.success) {
+        deployFailed.message = 'Unexpected error creating apex class member';
+        deployFailed.name = 'ApexClassMemberCreationFailed';
+        throw deployFailed;
+      }
+      this.containerMember = containerMember;
+    } catch (e) {
+      throw e;
+    }
```
brpowell: [Ditto](https://github.com/forcedotcom/salesforcedx-vscode/pull/2052/files#r387847277)

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+      if (!contAsyncRequest.success) {
+        deployFailed.message =
+          'Unexpected error creating container async request';
+        deployFailed.name = 'ContainerAsyncRequestFailed';
+        throw deployFailed;
+      }
+      this.asyncRequest = contAsyncRequest;
+    } catch (e) {
+      throw e;
+    }
```
brpowell: [Ditto](https://github.com/forcedotcom/salesforcedx-vscode/pull/2052/files#r387847277)

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+        retrieveResult = (await this.connection!.tooling.retrieve(
+          'ContainerAsyncRequest',
+          this.asyncRequest!.id
+        )) as ToolingRetrieveResult;
+        count++;
+      }
+      return retrieveResult;
+    } catch (e) {
+      throw e;
+    }
```
brpowell: [Ditto](https://github.com/forcedotcom/salesforcedx-vscode/pull/2052/files#r387847277)

packages/salesforcedx-vscode-core/src/commands/forceSourceToolingDeploy.ts
AnanyaJha: The CLI command is being built still so it gets surfaced in the output, but from what I understood the `execute()` method of the CompositeCliCommandExecutor doesn't actually spin up a shell process and run the command like the usual **CommandExecutors. It seems to act as more of a wrapper

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export class ToolingDeploy {
+  private static instance: ToolingDeploy;
+  private metadataType?: string;
+  private connection?: Connection;
+  private apiVersion?: string;
+
+  private container?: ToolingCreateResult;
+  private containerMember?: ToolingCreateResult;
+  private asyncRequest?: ToolingCreateResult;
```
brpowell: Is there a reason we can't pass these things (along with `metadataType`) as arguments to the methods that need them, instead of having them as properties? Keeping them as optional properties is forcing the use of the non-null assertion (`!` operator) quite a bit which is mitigating the usefulness of typescript keeping us safe.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+        authInfo: await AuthInfo.create({ username })
+      });
+      this.connection = connection;
+      if (apiVersion) this.apiVersion = apiVersion;
+    } catch (e) {
+      throw e;
+    }
+  }
+
+  public async deploy(options: any) {
```
brpowell: We should have the options parameter typed so consumers of the eventual library have useful autocompletion and there's type checking, considering it's a primary endpoint

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,190 @@
...
+      const deployFailed = new Error();
+      if (!metadataContainer.success) {
+        deployFailed.message = 'Unexpected error creating metadata container';
+        deployFailed.name = 'MetadataContainerCreationFailed';
+        throw deployFailed;
+      }
+      this.container = metadataContainer;
+    } catch (e) {
+      throw e;
+    }
```
brpowell: [Ditto](https://github.com/forcedotcom/salesforcedx-vscode/pull/2052/files#r387847277)

packages/salesforcedx-vscode-core/src/commands/forceSourceToolingDeploy.ts
AnanyaJha: I like the idea of lazily creating the connection, I'll change it to that structure

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
AnanyaJha: Yeah, it's because we're polling the status of the ContainerAsyncRequest to see when it changes from 'Queued' to anything else

packages/salesforcedx-vscode-core/package.json
```diff
@@ -749,6 +750,11 @@
...
+          "type": "boolean",
+          "default": true,
```
lcampos: Shouldn't this default to `false` ?

packages/salesforcedx-vscode-core/src/commands/forceSourceToolingDeploy.ts
```diff
@@ -0,0 +1,162 @@
...
+    executionWrapper: CompositeCliCommandExecution,
+    parser: ToolingDeployParser,
+    sourceUri?: string
+  ) {
+    const table = new Table();
+    let title: string;
+    switch (parser.result.State) {
+      case 'Completed':
+        title = nls.localize(`table_title_deployed_source`);
+        const successRows = parser.buildSuccesses(
```
lcampos: The first time I ran a deploy I got an issue from somewhere in here (Row is missing the key filePath) . I then ran it again without making any change to the file being deployed and it was successful.

Screen Shot 2020-03-04 at 2 25 26 PM

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
AnanyaJha: Oops yes, good point I thought I'd covered that 😅

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
AnanyaJha: Ah I see what you mean, let me see how I can consolidate

packages/salesforcedx-vscode-core/package.json
lcampos: @AnanyaJha this should be moved to `dependencies` or it will not be bundled in the vsix and the extension will fail to activate.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+  public metadataType?: string;
+  public connection: Connection;
+  private apiVersion?: string;
+
+  public constructor(connection: Connection, apiVersion?: string) {
+    this.connection = connection;
+    if (apiVersion) this.apiVersion = apiVersion;
+  }
+
+  public async deploy(filePath: string) {
```
lcampos: FYI: I simplified things a bit since this is the first version of the deploy functionality.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+const DOMParser = require('xmldom-sfdx-encoding').DOMParser;
+const CONTAINER_ASYNC_REQUEST = 'ContainerAsyncRequest';
+const METADATA_CONTAINER = 'MetadataContainer';
+
+export class ToolingDeploy {
+  public metadataType?: string;
+  public connection: Connection;
+  private apiVersion?: string;
+
+  public constructor(connection: Connection, apiVersion?: string) {
```
lcampos: Instead of receiving a username we now expect a connection object.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,62 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+    const supportedType =
+      path.extname(response.data) === APEX_CLASS_EXTENSION ||
+      response.data.includes(`${APEX_CLASS_EXTENSION}-meta.xml`);
+    if (betaPerfEnabled && supportedType) {
+      const startTime = process.hrtime();
+      const cancellationTokenSource = new vscode.CancellationTokenSource();
+      const cancellationToken = cancellationTokenSource.token;
+
+      const executionWrapper = new CompositeCliCommandExecutor(
+        this.build(response.data)
```
brpowell: Using the default `build()` here give the impression in the output panel that we're still running the CLI command

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,62 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+          true
+        );
+        if (!usernameOrAlias) {
+          throw new Error(nls.localize('error_no_default_username'));
+        }
+        const orgConnection = await OrgAuthInfo.getConnection(usernameOrAlias);
+        const deployLibrary = new ToolingDeploy(orgConnection);
+        const deployOutput = await deployLibrary.deploy(response.data);
+
+        const parser = new ToolingDeployParser(deployOutput!);
```
brpowell:
```suggestion
        const parser = new ToolingDeployParser(deployOutput);
```

Not needed

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,62 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+      try {
+        const usernameOrAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(
+          true
+        );
+        if (!usernameOrAlias) {
+          throw new Error(nls.localize('error_no_default_username'));
+        }
+        const orgConnection = await OrgAuthInfo.getConnection(usernameOrAlias);
+        const deployLibrary = new ToolingDeploy(orgConnection);
+        const deployOutput = await deployLibrary.deploy(response.data);
```
brpowell: This is only set up to handle one source path at a time. If deploy on save is enabled, a failure will be thrown for the next queued operation with multiple files (e.g. save all and quick saves).

packages/salesforcedx-vscode-core/src/commands/util/toolingDeployResultParser.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.result = deployResult;
+  }
+
+  public buildSuccesses(componentSuccess: DeployResult) {
+    const success = [
+      {
+        state: 'Add',
+        fullName: componentSuccess.fullName,
+        type: componentSuccess.componentType,
+        filePath: componentSuccess.fileName
```
brpowell: `componentSuccess.fileName` doesn't include the package directory part of the path, so you'll have `classes/test.cls` instead of the regular `my-package/main/default/classes/test.cls`

lcampos: Yeah, I also haven't tested this fully on an org & project with properly configured namespaces. I'll look into that next.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,62 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+
+        const parser = new ToolingDeployParser(deployOutput!);
+        await parser.outputResult(executionWrapper);
+        await DeployQueue.get().unlock();
```
brpowell: Can unlock the deploy queue in a `finally {}` block instead of having to do it in both the `try` and `catch`

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const containerMemberObject = {
+      MetadataContainerId: id,
+      FullName: fileName,
+      Body: body,
+      Metadata: metadataField,
+      ...(contentEntity ? { contentEntityId: contentEntity.Id } : {})
+    };
+    const containerMember = (await this.connection.tooling.create(
+      supportedToolingTypes.get(this.metadataType!)!,
```
brpowell: I know it's v1 so not terribly critical, but we should fix this to not use the not-null assertion. It's an operator that tends to make things difficult to debug.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -749,6 +750,11 @@
           "type": "boolean",
           "default": false,
           "description": "%conflict_detection_enabled_description%"
+        },
+        "salesforcedx-vscode-core.beta-perf-enhancements": {
```
brpowell: Screen Shot 2020-03-09 at 5 55 11 PM

Totally a nit but I suggest updating the setting name to be more descriptive - It's a bit odd to read. Examples: `"salesforcedx-vscode-core.performanceEnhancements"` or `"salesforcedx-vscode-core.deployAndRetrieveEnhancements"` and put "beta" in the description. It might also be nice to write that enabling the setting may improve the experience by speeding up deploy times. As a customer I might not know what the implication is behind "using the tooling api for single file deploys"

lcampos: I agree, I'm going to change the description to be more generic so we can use it to gate all the new perf enhancements

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: Yeah, but to avoid doing that we would need to update `CompositeCliCommandExecutor`'s constructor or use a new executor class. I think it's ok for now but we should address it in the next PRs.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: I'm taking care of this now on line 56 with the use of `multipleSourcePaths`

packages/salesforcedx-vscode-core/src/deploys/deployUtil.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface ToolingCreateResult {
+  id: string;
+  success: boolean;
+  errors: string[];
+  name: string;
+  message: string;
+}
+
+export interface ToolingRetrieveResult {
```
lcampos: Made some changes to the types to reflect what the api is returning e.g. some values are null when no value is provided

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,72 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+      response.data.includes(`${APEX_CLASS_EXTENSION}-meta.xml`);
+    const multipleSourcePaths = response.data.includes(',');
+
+    if (betaDeployRetrieve && supportedType && !multipleSourcePaths) {
+      const startTime = process.hrtime();
+      const cancellationTokenSource = new vscode.CancellationTokenSource();
+      const cancellationToken = cancellationTokenSource.token;
+
+      const executionWrapper = new CompositeCliCommandExecutor(
+        this.build(response.data)
```
brpowell:
```suggestion
        // TODO: Build command for non-cli execution like ForceGenerateFauxClassesExecutor 
        this.build(response.data)
```

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+const METADATA_CONTAINER = 'MetadataContainer';
+
+export class ToolingDeploy {
+  public metadataType?: string;
+  public connection: Connection;
+  private apiVersion?: string;
+
+  public constructor(connection: Connection, apiVersion?: string) {
+    this.connection = connection;
+    if (apiVersion) this.apiVersion = apiVersion;
```
brpowell:
```suggestion
    this.apiVersion = apiVersion;
```

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -58,5 +58,6 @@
   "refresh_components_text": "Refresh Components",
   "force_source_retrieve_display_text": "Retrieve Source from Org",
   "force_diff_against_org": "SFDX: Diff File Against Org",
-  "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template"
+  "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
+  "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience."
```
sfsholden: I like this wording to the user

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -33,6 +44,72 @@ export class ForceSourceDeploySourcePathExecutor extends BaseDeployExecutor {
...
+    const betaDeployRetrieve = sfdxCoreSettings.getBetaDeployRetrieve();
+    // this supported types logic is temporary until we have a way of generating the metadata type from the path
+    // once we have the metadata type we can check to see if it is a toolingsupportedtype from that util
+    const supportedType =
+      path.extname(response.data) === APEX_CLASS_EXTENSION ||
+      response.data.includes(`${APEX_CLASS_EXTENSION}-meta.xml`);
+    const multipleSourcePaths = response.data.includes(',');
+
+    if (betaDeployRetrieve && supportedType && !multipleSourcePaths) {
+      const startTime = process.hrtime();
```
sfsholden: I would recommend breaking some of this logic out into a few methods. I think it would make it easier to read and unit test.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+      deployFailed.name = 'MetadataContainerCreationFailed';
+      throw deployFailed;
+    }
+    return metadataContainer;
+  }
+
+  public buildMetadataField(metadataContent: string) {
+    const parser = new DOMParser();
+    const document = parser.parseFromString(metadataContent, 'text/xml');
+    const apiVersion =
```
sfsholden: Is it possible for these values to be different? And if so, would that cause any issues for us?

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
```diff
@@ -0,0 +1,159 @@
...
+      Metadata: metadataField,
+      ...(contentEntity ? { contentEntityId: contentEntity.Id } : {})
+    };
+    const containerMember = (await this.connection.tooling.create(
+      supportedToolingTypes.get(this.metadataType!)!,
+      containerMemberObject
+    )) as ToolingCreateResult;
+
+    if (!containerMember.success) {
+      const deployFailed = new Error();
```
sfsholden: Fairly similar code to above - I think this could be turned into a util method.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
lcampos: Yes, normally we want to use the apiVersion on the `-meta.xml` files (document.getElementsByTagName('apiVersion')...) but there might be scenarios in which we want to override it and that's when we would define `this.apiVersion`. Most of the time we want to use the api version defined in the `-meta.xml` file since some features might behave differently in the org depending on their api version, specially when it comes to Apex where some features are deprecated or introduced in specific api versions.

packages/salesforcedx-vscode-core/src/deploys/toolingDeploy.ts
lcampos: I agree, I didn't want to refactor too much on this PR cause we will have to generalize this code in order to support the rest of the types we'll use with Tooling API. I'm thinking that will naturally happen as part of introducing the second piece of metadata we'll support.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: Yeah, I think this will happen once we start using the new library and introduce a new executor for non-cli work. Those things should start happening fairly soon.

PR #2057, opened 2020-03-06
Update publish-vsix.js to use the global vsce module

lcampos:
### What does this PR do?
Updates publish-vsix.js script to use the global vsce module instead of hardcoding the path to a locally installed one. 

@W-7298800@

### DIFF
--- a/scripts/publish-vsix.js
+++ b/scripts/publish-vsix.js
@@ -12,16 +12,15 @@ if (!vsix.length) {
   shell.exit(1);
 }
 
-const vsce = '../../node_modules/.bin/vsce';
 const VSCE_PERSONAL_ACCESS_TOKEN = process.env['VSCE_PERSONAL_ACCESS_TOKEN'];
 let vscePublish = '';
 if (VSCE_PERSONAL_ACCESS_TOKEN) {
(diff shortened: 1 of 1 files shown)


scripts/publish-vsix.js
```diff
@@ -12,16 +12,15 @@ if (!vsix.length) {
   shell.exit(1);
 }
 
-const vsce = '../../node_modules/.bin/vsce';
 const VSCE_PERSONAL_ACCESS_TOKEN = process.env['VSCE_PERSONAL_ACCESS_TOKEN'];
 let vscePublish = '';
 if (VSCE_PERSONAL_ACCESS_TOKEN) {
   vscePublish = shell.exec(
-    `${vsce} publish --pat ${VSCE_PERSONAL_ACCESS_TOKEN} --packagePath ${vsix}`
+    `vsce publish --pat ${VSCE_PERSONAL_ACCESS_TOKEN} --packagePath ${vsix}`
```
lcampos: @sfsholden the `--pat` attribute takes care of using the token so we won't need to run `vsce login` in a clean environment, we'll only need to define the token as an env variable.

PR #2063, opened 2020-03-11
Publishing Logging Updates

sfsholden:
### What does this PR do?
Making some art out of the publishing logs. Wrapped these logging changes in a logger-util class in case someone ever needs to turn off the colors or we want to change them going forward.

@W-7250940@

### DIFF
--- a/scripts/download-vscode.js
+++ b/scripts/download-vscode.js
@@ -4,6 +4,7 @@ const ncp = require('ncp');
 const path = require('path');
 const fs = require('fs');
 const downloadAndUnzipVSCode = require('vscode-test').downloadAndUnzipVSCode;
+const logger = require('./logger-util');
 
 // Copy vscode to these directories
 const extensionDirectories = [
@@ -20,7 +21,7 @@ const extensionDirectories = [
 // ~/salesforcedx-vscode/packages/salesforcedx-vscode-lwc/.vscode-test/vscode-1.41.1/Visual Studio Code.app/Contents/MacOS/Electron
 downloadAndUnzipVSCode()
   .then(executablePath => {
-    console.log('Executable Path: ' + executablePath);
+    logger.debug('Executable Path: ' + executablePath);
 
     let vscodeIndex = executablePath.indexOf('.vscode-test') + 13;
 
@@ -39,9 +40,9 @@ downloadAndUnzipVSCode()
     // '~/salesforcedx-vscode/'
     let vscodeBasePath = path.dirname(path.dirname(vscodeFullPath));
 
-    console.log('Directory Name: ' + vscodeDirname);
-    console.log('Base Path: ' + vscodeBasePath);
-    console.log('Full Path: ' + vscodeFullPath);
+    logger.debug('Directory Name: ' + vscodeDirname);
+    logger.debug('Base Path: ' + vscodeBasePath);
+    logger.debug('Full Path: ' + vscodeFullPath);
 
     // If this script is run from an individual package, don't copy it around unnecessarily
     // Example:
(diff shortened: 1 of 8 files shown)


scripts/logger-util.js
```diff
@@ -0,0 +1,24 @@
+const RESET = "\x1b[0m";
+const BOLD = "\x1b[1m";
+const CYAN_BOLD = `${BOLD}\x1b[36m%s${RESET}`;
+const RED = `\x1b[31m%s${RESET}`;
+const YELLOW = `\x1b[33m%s${RESET}`;
+
+module.exports = {
```
sfsholden: Decided to move the color modifications into their own util. If we find that we want to change these colors going forward, this will be an easy way to do it.

scripts/validation-utils.js
```diff
@@ -33,90 +33,76 @@ module.exports = {
...
     }
   },
 
   checkAWSCliInstall: () => {
-    console.log(
-      `\nVerifying AWS CLI is installed for node version ${NODE_VERSION}.`
-    );
+    logger.header(`\nVerifying AWS CLI is installed for node version ${NODE_VERSION}.`);
     if (!shell.which('aws')) {
-      console.log('The AWS CLI is not installed or could not be found.');
```
sfsholden: Decided to remove all of these steps. Installing via pip3 seemed much more complicated (having to modify your path, dealing with python issues). From the release last week, installing via their install gui worked just fine.

scripts/logger-util.js
```diff
@@ -0,0 +1,24 @@
...
+  debug: str => {
+    console.log(str);
+  },
+
+  info: str => {
+    console.log(YELLOW, str);
+  },
+
+  error: str => {
+    console.log(RED, str);
```
lcampos: Is there any advantage of using console.log over console.error ?

scripts/download-vscode.js
```diff
@@ -59,24 +60,24 @@ downloadAndUnzipVSCode()
             '.vscode-test',
             vscodeDirname
           );
-          console.log('Creating Directories: ' + copyDestination);
+          logger.debug('Creating Directories: ' + copyDestination);
```
lcampos:
```suggestion
          logger.debug(`Creating Directories:  ${copyDestination}`);
```

scripts/download-vscode.js
```diff
@@ -59,24 +60,24 @@ downloadAndUnzipVSCode()
...
           if (!fs.existsSync(copyDestination)) {
             fs.mkdirSync(copyDestination, { recursive: true });
           }
-          console.log('Copying to: ' + copyDestination);
-          ncp(vscodeFullPath, copyDestination, function(err) {
+          logger.debug('Copying to: ' + copyDestination);
```
lcampos:
```suggestion
          logger.debug(`Copying to: ${copyDestination}`);
```

scripts/download-vsix-from-circleci.js
```diff
@@ -48,13 +49,13 @@ if (
...
 ) {
-  console.log(
+  logger.error(
     'Looks like your CircleCI Token does not grant you access to the salesforcedx-vscode builds.'
   );
   process.exit(-1);
 }
 
-// Get artifact names from CircleCI build
+logger.header("Get artifact names from CircleCI build.");
```
lcampos:
```suggestion
logger.header('Get artifact names from CircleCI build.');
```

scripts/download-vsix-from-circleci.js
```diff
@@ -66,31 +67,33 @@ const cciArtifacts = shell
...
+logger.info(
+  `There are ${buildArtifactsJSON.length} artifacts that will be processed...`
 );
 
 // Real-clean
 shell.exec('git clean -xfd -e node_modules');
 
-// Download vsix files from CircleCI and save vsix files in their correct package/salesforcedx-vscode-... location
+logger.header(
+  "Download vsix files from CircleCI and save vsix files in their correct package/salesforcedx-vscode-... location"
```
lcampos:
```suggestion
  'Download vsix files from CircleCI and save vsix files in their correct package/salesforcedx-vscode-... location'
```

scripts/logger-util.js
sfsholden: In this case I think the main advantage would be that we could always turn off the red, or change it to a different color. If we go with console.error it forces the color output to always be red.

PR #2064, opened 2020-03-11
Single Apex Class file deploys through Tooling API (#2052)

lcampos:
### What does this PR do?
Ports PR #2052 to the release branch for 48.6.0

@W-6531270@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/deploys/toolingDeployResultParser.ts
+++ b/packages/salesforcedx-vscode-core/src/deploys/toolingDeployResultParser.ts
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import {
+  Row,
+  Table
+} from '@salesforce/salesforcedx-utils-vscode/out/src/output';
+import { nls } from '../messages';
+import {
+  DeployResult,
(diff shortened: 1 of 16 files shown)


packages/salesforcedx-vscode-core/src/deploys/toolingDeployResultParser.ts
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import {
+  Row,
+  Table
+} from '@salesforce/salesforcedx-utils-vscode/out/src/output';
```
jag-sfdc: This is the dependency to salesforcedx-utils-vscode I was asking on Slack. The Row, Table classes itself are pretty simple and can have their own version here or is there any other option to remove the dependency?

jag-sfdc: To be clear, removing the dependency need not happen here - I'm thinking about it when we move it to the deploy-and-retrieve.

lcampos: I see, there are things we need to decide on the library side. For example, do we want it to return the results in this table format or is that the responsibility of the consumer. If we decide that the library should return results in table format then we would port the code that does that into the library as well.

We should talk about this more, the expectation is that anything needed to be in the library will not introduce any dependency to code that lives in the extensions or any other module that can't be bundled as part of the library.

Issue #2065, opened 2020-03-11
Cannot read property 'get' of undefined


rajesh-k:
I see the "Cannot read property 'get' of undefined" in vscode output anytime a javascript or xml file is saved. Uninstalling the Salesforce CLI Integration extension gets rid of this error. I need the Salesforce CLI Integration package along with the extension pack. Let me know how I can get rid of this error.
The vscode console log is attached.

[1583945690803.log](https://github.com/forcedotcom/salesforcedx-vscode/files/4319588/1583945690803.log)

lcampos:
Thanks for reporting this @rajesh-k . We've released a fix for this in version `48.6.0` of the extensions.

lgomezgwt:
This error is happening again in version 49.5.0

Lvkz:
This error is happening again in version 49.5.0

lcampos:
Hi @Lvkz I've checked and even when you might be getting the same error message I couldn't find it being a regression on the original issue. Please open a new issue and add details on how to replicate the problem you are seeing so we can work on a fix.

Eunix:
I also have the same error again now.

lohit398:
I also have the error again in VS code for non-salesforce projects.

Thanks!

alechko:
I'm having this issue as well on non-salesforce projects, even on non-react projects, this seems to be triggered simultaneously with saving a file.

Salesforce CLI - 49.9.0
Code - 1.49

Issue #1988, opened 2020-02-14
Causes PHP Syntax Highlighting Issue


westc:
### Summary

When this extension is enabled in Visual Studio Code the native syntax highlighting for PHP doesn't highlight other embedded languages (HTML, CSS and JavaScript) as expected.

### Steps To Reproduce:

1.  Enable this extension.
1.  Create a new file tab.
1.  Set the language to PHP.
1.  Enter the following code:
    ```php
    <?php if (true): ?>
      <div>Hello world!!!</div>
      <script type="text/javascript">
      var x = 3;
      document.write(x);
      </script>
      <style type="text/css">
      div {
        color: #00F;
      }
      </style>
    <?php endif; ?>
    ```

### Expected result

All of the code will be syntax highlighted depending on the language:
![image](https://user-images.githubusercontent.com/704518/74575285-c2ae2680-4f5c-11ea-85e5-38270cb74f10.png)

### Actual result

The PHP code will be highlighted as PHP but the inline HTML, CSS and JS will not be highlighted:
![image](https://user-images.githubusercontent.com/704518/74575338-fc7f2d00-4f5c-11ea-92d2-03da5527be40.png)

### Additional information

**VS Code Version**:  1.42.1

**SFDX CLI Version**:  sfdx-cli/7.36.0-b1c4e9c7c8 darwin-x64 node-v10.15.3

**OS and version**:  macOS Mojave 10.14.6

**OS According to VS Code**:  Darwin x64 18.7.0

lcampos:
Thanks for reporting this @westc , we've made changes in release `48.6.0` to make sure the extensions are only active in Salesforce projects which will take care of this.

xyc:
Looks like this is related to [html](https://github.com/forcedotcom/salesforcedx-vscode/blob/7184190b8659cf38733c31f2c355709945da5f7f/packages/salesforcedx-vscode-lightning/package.json#L77-L97) language contribution. While "Aura Components" extension is enabled (even if not activated) this issue would present. @mysticflute @nrkruk could we continue conversation here WRT modification for aura support?

ntotten:
@lcampos Is it even possible to disable the `language` contribution of an extension without deactivating it entirely?

jeffb-sfdc:
I'm still able to reproduce this.

michaelsoriano:
This is still happening. Messing up HTML color syntax in PHP files. 

[...]

PR #2094, opened 2020-03-21
Testing Support for LWC: Support internal workspace

xyc:
### What does this PR do?
Support internal workspace

Parent issue #1703, @W-6975995@

### Functionality Before
For LWC testing, SFDX workspace is supported

### Functionality After
For LWC testing, SFDX and internal workspaces are supported


packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -81,6 +81,7 @@
     "test:vscode-insiders-integration": "cross-env CODE_VERSION=insiders npm run test:vscode-integration"
   },
   "activationEvents": [
+    "workspaceContains:modules",
```
xyc: @lcampos this is needed to activate the extension in the subfolders as we discussed (not only for tests but general usage of LWC extension) - feel free to suggest any alternative

lcampos: This new event would still cause the activation of the extensions in projects that are not salesforce or core projects, causing a regression to these issues we've fixed (#1988 and #2065). 
I'd have to take a look at how we structure individual lwc modules in core and see if there's something we can leverage here. @mysticflute do you have any suggestions ?

xyc: @lcampos IMHO the regression risk is relatively low here since it was checking the _top-level_ folder `module`, which might not be popular among non-salesforce projects. VS Code doesn't seem to provide an [activation event](https://code.visualstudio.com/api/references/activation-events) to check outside of workspace folder either. Maybe it's possible to execute a [command](https://code.visualstudio.com/api/references/activation-events#onCommand) for activation, based on the parent folder check, but I'm not sure it's worth the effort. If we want to be very safe here, there need to be a null check somewhere regarding #2065.

xyc: Regarding #1988, looks like this issue still presents, as of 48.11.0. Briefly took a look and I think it might be because that `salesforcedx-vscode-lightning` contributes to `html` language. `activate` hasn't been called. Only when I disable "Aura Components" extension file syntax highlighting works for embedded html in PHP.

In my opinion there isn't much we can do here, except recommending users to disable "Aura Components" extension within the workspace.

mysticflute: regarding core workspaces, I guess it depends on the available options for `activationEvents`. Core modules should have a pom.xml file that extends from the salesforce pom. If possible that could be something we could check for, but it doesn't seem we can do a check like that here.

If we can only check for the presence of files, if it's possible to check the root folder then we can check that it contains the `modules` folder, `pom.xml` file, and `build` directory which should is pretty specific. In this case the better long term IMO if we add a configuration file that should be present and we check for that, e.g., `salesforce-vscode.config.json`. And make this file part of the template for component modules or have blt automatically add it.

That is more work though. While we have talked about possible needing a configuration file I don't think we need it yet. So what I'm leaning towards is letting the extension activate, and in the extension activation code we should have more checks to see if it should actually do anything there.

xyc: @mysticflute Good point. I believe the extension is already checking if the workspace [isLWC](https://github.com/forcedotcom/lightning-language-server/blob/0e7dbe94671a62991109cadc95f22aa0d948a0ae/packages/lightning-lsp-common/src/shared.ts#L35) in [activation code](https://github.com/forcedotcom/salesforcedx-vscode/blob/d88eb17761e96bbfdf27aa99ccfcdd2798bf34f2/packages/salesforcedx-vscode-lwc/src/index.ts#L72-L83), if activationMode config is set to `autodetect` (which is the default config). If activationMode config is set to `always` it always activates.

xyc: Moving the activation part to a new work item tracking it W-7509875, as discussed offline with @lcampos

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -81,6 +81,7 @@
   },
   "activationEvents": [
     "workspaceContains:sfdx-project.json",
+    "workspaceContains:../workspace-user.xml",
```
xyc, line 84: This ensures proper activation in partial workspace.

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
```diff
@@ -0,0 +1,45 @@
...
+  const workspaceType = workspaceService.getCurrentWorkspaceType();
+  if (testRunType === TestRunType.DEBUG) {
+    if (
+      workspaceService.isSFDXWorkspace(workspaceType) ||
+      workspaceService.isCoreWorkspace(workspaceType)
+    ) {
+      return ['--debug', '--', ...jestArgs];
+    }
+    // TODO: For LWC OSS
+    // Debug args are ['test:unit', '--debug', '--passthrough', ...jestArgs]
```
xyc, line 28: This is for potential`create-lwc-app` app support in the future

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -51,6 +51,8 @@ export const messages = {
     'SFDX: Navigate to Lightning Web Component Test',
   no_lwc_jest_found_text:
     'sfdx-lwc-jest is not installed. Install it from https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.unit_testing_using_jest_installation',
+  no_lwc_testrunner_found_text:
```
lcampos, line 54: This exposes a url to an internal repo, not sure if that's common practice. @hbc-bindu @ekapner do you have any suggestions here ?

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
```diff
@@ -0,0 +1,45 @@
...
+  jestArgs: string[],
+  testRunType: TestRunType
+) {
+  const workspaceType = workspaceService.getCurrentWorkspaceType();
+  if (testRunType === TestRunType.DEBUG) {
+    if (
+      workspaceService.isSFDXWorkspace(workspaceType) ||
+      workspaceService.isCoreWorkspace(workspaceType)
+    ) {
+      return ['--debug', '--', ...jestArgs];
```
lcampos, line 25: I don't think the order of the arguments matters. If that is true, you can restructure the code so you create an array of options depending on the workspace type and add the debug argument at the end if needed. That will remove a lot of the code duplication.

packages/salesforcedx-vscode-lwc/src/testSupport/index.ts
```diff
@@ -12,8 +13,26 @@ import { lwcTestIndexer } from './testIndexer';
 import { taskService } from './testRunner/taskService';
 import { testResultsWatcher } from './testRunner/testResultsWatcher';
 import { startWatchingEditorFocusChange } from './utils/context';
+import { workspaceService } from './workspace/workspaceService';
```
lcampos, line 16:
```suggestion
import { workspaceService } from './workspace';
```
Since workspaceService is already exported by the module's index.ts

packages/salesforcedx-vscode-lwc/src/testSupport/testRunner/testRunner.ts
```diff
@@ -11,7 +11,12 @@ import * as vscode from 'vscode';
 import { nls } from '../../messages';
 import { telemetryService } from '../../telemetry';
 import { TestExecutionInfo, TestInfoKind } from '../types';
-import { getLwcTestRunnerExecutable, getTestWorkspaceFolder } from './index';
+import {
+  getCliArgsFromJestArgs,
+  getLwcTestRunnerExecutable,
+  workspaceService
+} from '../workspace';
+import { getTestWorkspaceFolder } from './index';
```
lcampos, line 19: We should also move `getTestWorkspaceFolder` under the `workspace` module.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
ekapner: I would say we don't want to expose an internal repo. @lcampos -- Want to make sure I understand your question.  And then we can decide next steps. Can they follow the instrux and be successful? If so, we can update the topic to remove references to the internal repo. If they must have access to the repo to be successful, that's a problem. And I would need to discuss with Stephanie and @hbc-bindu what the alternatives are.

packages/salesforcedx-vscode-lwc/src/testSupport/testRunner/testRunner.ts
xyc: Good catch - updated.

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
xyc: This is actually intentional to account for the test runner differences. The order matters here and I chose to explicitly declare the parameters since the test runners are quite different (esp. `create-lwc-app` test runner)

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
xyc: I've removed the url for now (It's gated anyway). @ekapner @hbc-bindu Let me know what you think with the current text. We will provide them with documentation in internal channels.

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
lcampos: That's fine if the order matters, I think there are still ways in which you can rewrite the code so that there's no duplication of conditions if you want to enable/disable the debug flag.

xyc: Updated the function and simplified the current implementation. For reference [this](https://github.com/forcedotcom/salesforcedx-vscode/blob/4b194842168f5189ba26529d2fe340cbff64aac9/packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts) is the original commit in case we need to do it differently in the future.

PR #2096, opened 2020-03-24
Conflict visualization using a dedicated view.

rcoringrato-sfdc:
### What does this PR do?

- Provide a view to visualize org metadata differences

@W-7219630@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>


packages/salesforcedx-vscode-core/src/conflict/index.ts
```diff
@@ -15,4 +20,51 @@ export {
...
 export const conflictDetector = ConflictDetector.getInstance();
+
+export async function setupConflictView(
+  extensionContext: vscode.ExtensionContext
+): Promise<void> {
+  const view = conflictView;
+  await view.init(extensionContext);
+}
+
+export async function registerConflictView(): Promise<vscode.Disposable> {
```
brpowell: Does this need to be `async`?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -221,39 +226,72 @@ export class ConflictDetectionChecker implements PostconditionChecker<string> {
...
+      });
+      channelService.showChannelOutput();
+
+      const choice = await notificationService.showWarningModal(
+        nls.localize(this.messages.warningMessageKey),
+        nls.localize('conflict_detect_override'),
+        nls.localize('conflict_detect_show_conflicts')
+      );
+
+      if (choice !== nls.localize('conflict_detect_override')) {
```
brpowell: If we choose to override the conflicts, do we want to clear the conflict view? Maybe a question for @smaddox-sf

brpowell:
A nit, but do we have an SVG version of the tree view icon? It's a bit fuzzy with the PNG.

Screen Shot 2020-03-25 at 2 24 47 PM

ntotten:
We should just use one of the built in icons.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -720,6 +737,14 @@
       {
         "command": "sfdx.force.diff",
         "title": "%force_diff_against_org%"
+      },
+      {
+        "command": "sfdx.force.conflict.open",
```
lcampos: We need to hide this from the command palette or running it from there throws an error.
Screen Shot 2020-03-30 at 9 49 36 AM

packages/salesforcedx-vscode-core/src/conflict/conflictNode.ts
```diff
@@ -0,0 +1,87 @@
...
+export class ConflictFileNode extends ConflictNode {
+  constructor(conflict: ConflictFile, parent: ConflictNode) {
+    super(conflict.fileName, vscode.TreeItemCollapsibleState.None, parent);
+    this._conflict = conflict;
+  }
+
+  public attachCommands() {
+    this.contextValue = 'conflict-actions';
+    this.command = {
+      title: 'Compare',
```
lcampos: This should be a label.

packages/salesforcedx-vscode-core/src/conflict/index.ts
```diff
@@ -15,4 +20,51 @@ export {
...
+  const viewItems: vscode.Disposable[] = [];
+
+  viewItems.push(
+    vscode.commands.registerCommand('sfdx.force.conflict.diff', entry =>
+      conflictDiff(entry)
+    )
+  );
+
+  viewItems.push(
```
lcampos: I found this a bit confusing, clicking on the name of the file opens the conflict but clicking the icon exclusively opens the file. Shouldn't we always just open the conflict if the user is clicking the conflict view ?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
rcoringrato-sfdc: Yes, this was an oversight, the view should be cleared.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -107,6 +112,13 @@
           "name": "Metadata",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
         }
+      ],
+      "conflicts": [
+        {
+          "id": "conflicts",
+          "name": "Conflicts",
+          "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
```
lcampos: This means that the conflict view can only be seen/used by non-scratch orgs. I think we should change this condition to showing the icon/view when a dx project is open and the user has enabled conflict detection. Here's an example of how we get the setting value in package.json https://github.com/forcedotcom/salesforcedx-vscode/blob/f6d905a4d53ab7599ff94cd231c5e5228a046445/packages/salesforcedx-vscode-lightning/package.json#L112

packages/salesforcedx-vscode-core/src/conflict/conflictNode.ts
rcoringrato-sfdc: I can make it a label, but I never saw it displayed anywhere.

packages/salesforcedx-vscode-core/src/conflict/index.ts
rcoringrato-sfdc: I found this to be a useful capability. It allows direct navigation to just the project file. This is also how the VCS view works, so there is precedent for it.

packages/salesforcedx-vscode-core/package.json
rcoringrato-sfdc: @lcampos What is the benefit for displaying it for a scratch org context?

packages/salesforcedx-vscode-core/src/conflict/index.ts
rcoringrato-sfdc: I'll test it without, thanks.

smaddox-sf:
> A nit, but do we have an SVG version of the tree view icon? It's a bit fuzzy with the PNG.
> 
> Screen Shot 2020-03-25 at 2 24 47 PM

Let's switch to this standard icon - https://github.com/microsoft/vscode-codicons/blob/master/src/icons/issues.svg

packages/salesforcedx-vscode-core/package.json
lcampos: @rcoringrato-sfdc sorry I forgot some of the implementation details about the feature. This should only be exposed for non-scratch orgs and we should also check that the setting is enabled.

packages/salesforcedx-vscode-core/src/conflict/index.ts
lcampos: Got it, thanks for the context. Let's keep it as it is.

PR #2102, opened 2020-03-27
Perf changes for source retrieve

lcampos:
### What does this PR do?
Here's an early version of how consuming the new retrieve library endpoint will look like. These changes include a new small wrapper that will handle calling channelservice, notificationservice, showing the progress of the execution, etc.

@W-7500391@

### Functionality Before/After
retrieve_beta

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
@@ -27,6 +27,10 @@ import {
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './util';
+import { LibraryCommandletExecutor } from './util/libraryCommandlet';
+import { ToolingDeploy, ToolingDeployParser } from '../deploys';
+import { ToolingRetrieve } from 'deploy-and-retrieve/lib/src/retrieve';
+import { Connection } from '@salesforce/core';
 
 export class ForceSourceRetrieveSourcePathExecutor extends SfdxCommandletExecutor<
   string
@@ -95,11 +99,45 @@ export async function forceSourceRetrieveSourcePath(explorerPath: vscode.Uri) {
       return;
     }
   }
-  const commandlet = new SfdxCommandlet(
+  /*const commandlet = new SfdxCommandlet(
     new SfdxWorkspaceChecker(),
     new FilePathGatherer(explorerPath),
     new ForceSourceRetrieveSourcePathExecutor(),
     new SourcePathChecker()
+  ); */
+  const commandlet = new SfdxCommandlet(
+    new SfdxWorkspaceChecker(),
+    new FilePathGatherer(explorerPath),
+    new LibraryRetrieveSourcePathExecutor(),
+    new SourcePathChecker()
   );
   await commandlet.run();
+}
+
+export class LibraryRetrieveSourcePathExecutor extends LibraryCommandletExecutor<
+  string
+> {
+  public async execute(response: ContinueResponse<string>): Promise<void> {
+    const startTime = process.hrtime();
+    try {
+      console.log('--- lib request execute phase');
+      await this.build();
+      if (this.orgConnection === undefined) {
+        throw new Error('Connection is not established');
+      }
+      const toolingRetrieve = new ToolingRetrieve(this.orgConnection);
+      const deployOutput = await toolingRetrieve.getMetadata(response.data); // deployLibrary.deploy(response.data);
+
+      // const parser = new ToolingDeployParser(deployOutput);
+      // const outputResult = await parser.outputResult();
+      channelService.appendLine(deployOutput);
+      this.logMetric('beta_command', startTime);
+    } catch (e) {
+      telemetryService.sendException(
+        'force_source_deploy_with_sourcepath_beta',
+        e.message
+      );
+      channelService.appendLine(e.message);
+    }
+  }
 }
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
@@ -4,8 +4,6 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import * as path from 'path';
-import * as vscode from 'vscode';
 
 import {
   Command,
@@ -16,11 +14,16 @@ import {
   CancelResponse,
   ContinueResponse
 } from '@salesforce/salesforcedx-utils-vscode/out/src/types/index';
+import { Retrieve } from '@salesforce/source-deploy-retrieve/lib/tooling/retrieve';
+import * as path from 'path';
+import * as vscode from 'vscode';
 import { channelService } from '../channels';
 import { nls } from '../messages';
 import { notificationService } from '../notifications';
+import { sfdxCoreSettings } from '../settings';
 import { SfdxPackageDirectories } from '../sfdxProject';
 import { telemetryService } from '../telemetry';
+import { APEX_CLASS_EXTENSION } from './templates/metadataTypeConstants';
 import {
   FilePathGatherer,
   SfdxCommandlet,
@@ -28,9 +31,6 @@ import {
   SfdxWorkspaceChecker
 } from './util';
 import { LibraryCommandletExecutor } from './util/libraryCommandlet';
-import { ToolingDeploy, ToolingDeployParser } from '../deploys';
-import { ToolingRetrieve } from 'deploy-and-retrieve/lib/src/retrieve';
-import { Connection } from '@salesforce/core';
 
 export class ForceSourceRetrieveSourcePathExecutor extends SfdxCommandletExecutor<
   string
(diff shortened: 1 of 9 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@salesforce/core": "2.2.0",
     "@salesforce/salesforcedx-utils-vscode": "48.8.0",
+    "@salesforce/source-deploy-retrieve": "github:forcedotcom/deploy-and-retrieve#lcampos/retrieve",
```
lcampos: This will of course be replaced once we publish the library.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -85,6 +85,10 @@ export class ChannelService {
     });
   }
 
+  public showCommandWithTimestamp(commandName: string) {
```
lcampos: The next updates will have the rest of the code in this service using this new method.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -95,11 +99,59 @@ export async function forceSourceRetrieveSourcePath(explorerPath: vscode.Uri) {
...
+      : new ForceSourceRetrieveSourcePathExecutor(),
     new SourcePathChecker()
   );
   await commandlet.run();
 }
+
+// this supported types logic is temporary until we have a way of generating the metadata type from the path
+// once we have the metadata type we can check to see if it is a toolingsupportedtype from that util
+
+function useBetaRetrieve(explorerPath: vscode.Uri): boolean {
```
lcampos: Scoping the retrieve exclusively to Apex Class for now.

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.telemetryName = telemetryLogName;
+    // initialize connection
+    const usernameOrAlias = await OrgAuthInfo.getDefaultUsernameOrAlias(true);
+    if (!usernameOrAlias) {
+      throw new Error(nls.localize('error_no_default_username'));
+    }
+    this.orgConnection = await OrgAuthInfo.getConnection(usernameOrAlias);
+  }
+
+  public retrieveWrapper(
```
lcampos: This is where most of the visual aspects of the command execution are wired (publishing to output panel, showing progress message, etc). Initially I wanted to have it be generic enough to support calling the deploy, retrieve and other library methods but doing that wouldn't allow us to assume a response shape (line 58) so we could also have this format the results. After talking with Nick I opted to have a dedicated wrapper for the retrieve for now.

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ToolingRetrieveResult } from '@salesforce/source-deploy-retrieve/lib/tooling/retrieve';
+import { ProgressLocation, window } from 'vscode';
+import { channelService } from '../../channels';
+import { nls } from '../../messages';
+import { notificationService } from '../../notifications';
+import { TelemetryData, telemetryService } from '../../telemetry';
+import { OrgAuthInfo } from '../../util';
+import { CommandletExecutor } from './sfdxCommandlet';
+
+export abstract class LibraryCommandletExecutor<T>
```
lcampos: Light-weight executor for running the new library commands. @jag-sfdc @AnanyaJha once you have your deploy PRs ready we would just have to add a deploy wrapper here.

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -0,0 +1,84 @@
...
+          title: commandName,
+          location: ProgressLocation.Notification
+        },
+        async () => {
+          // @ts-ignore
+          return (await fn.call(this, ...args)) as ToolingRetrieveResult;
+        }
+      );
+
+      channelService.appendLine('Library result =>' + JSON.stringify(result));
```
lcampos: I'm still working on defining the shape of the json being return from the library. Once that's done I'll replace this with a call to the json-to-table utilities.

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -0,0 +1,84 @@
...
+  protected executionName: string = '';
+  protected startTime: [number, number] | undefined;
+  protected telemetryName: string | undefined;
+
+  public execute(response: ContinueResponse<T>): void {}
+
+  public async build(
+    execName: string,
+    telemetryLogName: string
+  ): Promise<void> {
```
lcampos: @brpowell this will change once we start working in having a Client in the library that shares the connection with the rest of the functionality.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
jag-sfdc: We can expose a method on deploy-retrieve that returns if a type is supported based on filePath, so here we don't even care about figuring out the metadatype.

lcampos: That's a good idea, I'll add it to my retrieve changes.

lcampos: That's a good idea, I'll add it to my retrieve changes.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -95,11 +103,62 @@ export async function forceSourceRetrieveSourcePath(explorerPath: vscode.Uri) {
...
   await commandlet.run();
 }
+
+// this supported types logic is temporary until we have a way of generating the metadata type from the path
+// once we have the metadata type we can check to see if it is a toolingsupportedtype from that util
+export function useBetaRetrieve(explorerPath: vscode.Uri): boolean {
+  const filePath = explorerPath.fsPath;
+  const betaDeployRetrieve = sfdxCoreSettings.getBetaDeployRetrieve();
+  const supportedType =
+    path.extname(filePath) === APEX_CLASS_EXTENSION ||
```
lcampos, line 124: I found a bug when using the `*-meta.xml` file path in the library so just enabling for the non-meta xml path for now.

packages/salesforcedx-vscode-core/src/commands/util/retrieveParser.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import {
```
lcampos, line 8: Decided to have the parser as a function vs a class since there was not a lot of benefits from using a class here.

packages/salesforcedx-vscode-core/src/commands/util/retrieveParser.ts
```diff
@@ -0,0 +1,71 @@
...
+        { key: 'type', label: nls.localize('table_header_type') },
+        {
+          key: 'filePath',
+          label: nls.localize('table_header_project_path')
+        }
+      ],
+      title
+    );
+  } catch (e) {
+    telemetryService.sendException(
```
lcampos, line 61: If we run into any parsing issue we'll return the raw apiresult so the user can at least get some input on the operation.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -517,5 +517,8 @@ export const messages = {
   beta_tapi_mdcontainer_error: 'Unexpected error creating metadata container',
   beta_tapi_membertype_error: 'Unexpected error creating %s member',
   beta_tapi_car_error: 'Unexpected error creating container async request',
-  beta_tapi_queue_status: 'The deploy is still in the Queue'
+  beta_tapi_queue_status: 'The deploy is still in the Queue',
+  lib_retrieve_result_title: 'Retrieved Source',
```
lcampos, line 521: @hbc-bindu let me know what you think about these messages.

packages/salesforcedx-vscode-core/src/commands/util/retrieveParser.ts
brpowell: We should do this at some point for deploys too 😆

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -95,11 +106,84 @@ export async function forceSourceRetrieveSourcePath(explorerPath: vscode.Uri) {
...
+    (path.extname(filePath) === APEX_TRIGGER_EXTENSION ||
+      filePath.includes(`${APEX_TRIGGER_EXTENSION}-meta.xml`)) ||
+    (path.extname(filePath) === VISUALFORCE_COMPONENT_EXTENSION ||
+      filePath.includes(`${VISUALFORCE_COMPONENT_EXTENSION}-meta.xml`)) ||
+    (path.extname(filePath) === VISUALFORCE_PAGE_EXTENSION ||
+      filePath.includes(`${VISUALFORCE_PAGE_EXTENSION}-meta.xml`));
+  return betaDeployRetrieve && supportedType;
+}
+
+export function isAuraOrLWC(filePath: string) {
```
lcampos, line 139: Is there a better way of doing this ?

brpowell: Maybe something like this?

```typescript
import { RegistryAccess, registryData } from '@salesforce/source-deploy-retrieve'

const registry = new RegistryAccess();
const component = registry.getComponentsFromPath(filePath)[0];
const typeName = component.type.name;
const { auradefinitionbundle, lightningcomponentbundle } = registryData.types;
return typeName === auradefinitionbundle.name ||
       typeName === lightningcomponentbundle.name;
```

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -95,11 +106,84 @@ export async function forceSourceRetrieveSourcePath(explorerPath: vscode.Uri) {
...
+  let res = false;
+  if (
+    !fs.statSync(filePath).isDirectory() ||
+    (!filePath.endsWith(AURA_DIRECTORY) && !filePath.endsWith(LWC_DIRECTORY))
+  ) {
+    const parent = path.dirname(filePath);
+    const grandParent = path.dirname(parent);
+    res =
+      parent.endsWith(AURA_DIRECTORY) ||
+      grandParent.endsWith(AURA_DIRECTORY) ||
```
brpowell, line 149: technically there could be x number of parents before reaching the component bundle itself (I believe this is the case but let me know otherwise) and it would be valid, so using the registry approach like i posted above might be more resilient

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
lcampos: I like this better since I could replace a lot of the logic in `useBetaRetrieve` that evaluates the paths

PR #2103, opened 2020-03-28
Fix LWC Test issue with default shell set to Git Bash

### LINKED ISSUE #2097: With default shell set to Git Bash, running Jest tests failed in Windows
### Summary

With default shell set to Git Bash, running Jest tests failed in Windows

### Steps To Reproduce:

1. in Windows, Install Git
2. Open Command Palette: Terminal: Select Default Shell
3. Set default shell to Git Bash
4. Run a jest test from VS Code

### Expected result

Should successfully run

### Actual result

Fails with:
node_modules.binlwc-jest: command not found
The terminal process terminated with exit code: 127

### Additional information

**VS Code Version**:

**SFDX CLI Version**:

**OS and version**:

xyc:
### What does this PR do?
Fixes issue where LWC tests fail in VS Code when default shell is set to Git Bash.
The fix is to always launch tasks in Windows with command prompt (cmd.exe).

Fixes #2097, @W-7372054@


packages/salesforcedx-vscode-lwc/src/testSupport/testRunner/taskService.ts
```diff
@@ -126,7 +126,22 @@ class TaskService {
       sfdxTaskId: taskId
     };
     const taskSource = 'SFDX';
-    const taskShellExecution = new vscode.ShellExecution(cmd, args);
+    // https://github.com/forcedotcom/salesforcedx-vscode/issues/2097
```
lcampos, line 129: I think this should be ok, might want to add posting a message to the output panel saying we're defaulting to cmd.exe so the user that's on git bash or other client is aware of it. 

@mysticflute, @smaddox-sf what do you think ?

mysticflute: If I understand the issue, this does not occur by default on windows but only if the user has specifically set the default shell to git bash? If that's correct then it sounds reasonable to output a message

xyc: @mysticflute That's correct. But we don't know if the user has set the default shell to git bash or cmd.exe. If it's fine outputting a message every time in Windows, I'll add it. Frankly I think this won't actually affect the experience, since after running a test the terminal would not be reusable for the user. (The terminal would say: "Terminal will be reused by tasks, press any key to close it." )

mysticflute: I see. Is it possible to print to the output panel without forcing it to display/pop up? If the message can be added in the background without disrupting the user each time that sounds better to me.

packages/salesforcedx-vscode-lwc/src/testSupport/testRunner/taskService.ts
```diff
@@ -126,7 +131,25 @@ class TaskService {
...
+    // Git Bash shell doesn't handle command paths correctly.
+    // Always launch with command prompt (cmd.exe) in Windows.
+    const isWin32 = /^win32/.test(process.platform);
+    let taskShellExecutionOptions: vscode.ShellExecutionOptions | undefined;
+    if (isWin32) {
+      channelService.appendLine(
+        nls.localize('task_windows_command_prompt_messaging')
```
xyc, line 141: Uses a generic message here - in case taskService could be reused by other extensions in the future.

xyc: @mysticflute @lcampos Added the output panel message here

PR #2115, opened 2020-04-08
Add display and error message to scratch def prompt

sfsholden:
### What does this PR do?

1. Adds a display message to the user when selecting the project-scratch-def file. This will help a user identify the exact file name format that we are matching on. This will be beneficial for users that must make use of multiple scratch definitions and aren't aware of our matching rules.
1. If no files are present for selection an error will be displayed.

@W-7412945@

### Functionality Before
_Previously, if a user did not have any files that matched our criteria we would leave them with a blank selection screen without any helpful info._
create_scratch_not_found

### Functionality After
_When no files are found, display an error._
error_message

_Display message with additional info when selecting the project scratch def._
new_display_message


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -57,6 +57,8 @@ export const messages = {
     'Enter an org alias or use default alias',
   parameter_gatherer_enter_custom_url:
     'Enter a custom login URL or use default URL',
+  parameter_gatherer_enter_scratch_org_def_files:
```
sfsholden, line 60: @hbc-bindu Please let me know if you'd like to update these messages to the user.

hbc-bindu: Hi @sfsholden, tiny change:
'Enter an org alias or use the default alias'
'Enter a custom login URL or use the default URL'

sfsholden: This piece is specifically for grabbing the scratch def file. The prompts for an org alias and custom login URL are later in the workflow

sfsholden: The last gif in the description has an example of that full workflow :)

AnanyaJha:
This seems really helpful! Quick question, from the gif it looks like the full message inside of the quickpick where files are present is getting a little hidden/crowded. Specifically the `files with format...` part, maybe we could try wit a shorter message? not sure, just a thought 🙂

sfsholden:
> This seems really helpful! Quick question, from the gif it looks like the full message inside of the quickpick where files are present is getting a little hidden/crowded. Specifically the `files with format...` part, maybe we could try with a shorter message? not sure, just a thought 🙂

For the gif I had to condense it so it would fit the normal size requirements for uploading. Normally you would be able to see the full message (it's size is about the same length as the prompt for the scratch org duration).  If you have any suggestions that would be shorter though, please let me know! This was the smallest I could come up with without losing context for the user.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -287,6 +289,8 @@ export const messages = {
     'Run "SFDX: Authorize an Org" to authorize your org again.',
   error_org_browser_init: 'Org Browser has not been initialized',
   error_overwrite_prompt: 'Error checking workspace for existing components',
+  error_no_scratch_def:
+    'No scratch defition files found. These files must be in the "config" folder and have file names that end with "-scratch-def.json". See [Scratch Org Definition File](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm) for help.',
```
lcampos, line 293:
```suggestion
    'No scratch definition files found. These files must be in the "config" folder and have file names that end with "-scratch-def.json". See [Scratch Org Definition File](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm) for help.',
```

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
```diff
@@ -192,7 +192,11 @@ const preconditionChecker = new CompositePreconditionChecker(
   new DevUsernameChecker()
 );
 const parameterGatherer = new CompositeParametersGatherer(
-  new FileSelector('config/**/*-scratch-def.json'),
+  new FileSelector(
```
lcampos, line 195: We create the scratch org definition files under the `config` folder because of the decisions made in the templates repo/plugin. However, the scratch org definition files can be anywhere in the project.
We should either allow the user to pick a file anywhere in the project (similar to how we pick where to create an apex class) or we should enforce the files to be under the `config` folder while also providing a command that creates a scratch org definition file for them.

cc @smaddox-sf

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sfsholden: Thanks, good catch

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
sfsholden: I like that idea of creating a scratch def file for a user in the event that one is not present. We could also add some generic templates (like having one that inserts a namespace param). We could make some of the more common scenarios easier on customers. 

@smaddox-sf What are your thoughts on enforcing the config folder be the default and the creation of the scratch def file?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sfsholden: @hbc-bindu Hey sorry I misunderstood - I will add your suggestions to those old messages. Does everything look good for the new messages?

packages/salesforcedx-vscode-core/src/commands/forceOrgCreate.ts
smaddox-sf: @sfsholden, @lcampos  - i like the idea of creating a scratch def file for them so they don't have to go to documentation and then come back.  I think it makes sense to keep enforcing the config folder since the template repo needs that location.  It is a logical location, so i don't see an issue with that.

sfsholden: Added W-7449375 to internal engineering for addressing this piece.

PR #2128, opened 2020-04-22
Add check for local dev server start failure

sfsholden:
### What does this PR do?
Checks for error logging in the event that the start of the local dev server failed. This output is printed in the event that the address for the server is invalid, or an error was caught during start up.

@W-7462993@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -114,26 +114,15 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
       }
     });
 
-    // handler errors
-    execution.processExitSubject.subscribe(async exitCode => {
-      DevServerService.instance.clearServerHandler(serverHandler);
-
-      if (!serverStarted && !cancellationToken.isCancellationRequested) {
-        let message = nls.localize('force_lightning_lwc_start_failed');
-
-        // TODO proper exit codes in lwc-dev-server for address in use, auth/org error, etc.
-        if (exitCode === 127) {
-          message = nls.localize('force_lightning_lwc_start_not_found');
-        }
-
-        showError(new Error(message), logName, commandName);
-      } else if (exitCode !== undefined && exitCode !== null && exitCode > 0) {
-        const message = nls.localize(
-          'force_lightning_lwc_start_exited',
-          exitCode
-        );
-        showError(new Error(message), logName, commandName);
+    execution.stderrSubject.subscribe(async data => {
+      if (data && data.toString().includes('Server start up failed')) {
+        this.handleErrors(cancellationToken, serverHandler, 1);
+        progress.complete();
       }
+    });
+
+    execution.processExitSubject.subscribe(async exitCode => {
+      this.handleErrors(cancellationToken, serverHandler, exitCode);
     });
 
     notificationService.reportExecutionError(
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -114,28 +114,17 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
...
-      } else if (exitCode !== undefined && exitCode !== null && exitCode > 0) {
-        const message = nls.localize(
-          'force_lightning_lwc_start_exited',
-          exitCode
-        );
-        showError(new Error(message), logName, commandName);
+    execution.stderrSubject.subscribe(async data => {
+      if (data && data.toString().includes('Server start up failed')) {
+        this.handleErrors(cancellationToken, serverHandler, 1);
+        progress.complete();
```
sfsholden, line 120: This piece is needed in order for ProgressNotification to close the open VSCode window. This piece isn't needed for processExitSubject because there is an observer for this event already registered in ProgressNotification.

lcampos: My understanding of using the CLI's `SFDX_JSON_TO_STDOUT: 'true'` environment variable (line 71) was that the command's output will always go through stdout even for error situations. Does that only apply to "core" cli commands and not plugins ?

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -148,6 +137,30 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
...
+  private handleErrors(
+    cancellationToken: vscode.CancellationToken,
+    serverHandler: ServerHandler,
+    exitCode: number | null | undefined
+  ) {
+    DevServerService.instance.clearServerHandler(serverHandler);
+    if (!this.serverStarted && !cancellationToken.isCancellationRequested) {
+      let message = nls.localize('force_lightning_lwc_start_failed');
+
+      // TODO proper exit codes in lwc-dev-server for address in use, auth/org error, etc.
```
lcampos, line 150: Agreed, do we have a work item for tracking that work ?

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
sfsholden: That appears to be the case. From local testing, this definitely doesn't hit the route for stdout.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
sfsholden: Thanks for pointing that out, I added the last two.
W-7148356 - address already in use
W-7485707 - expired org
W-7485723 - no default org set

jag-sfdc: As part of that can we also add a default_failure error code instead of "1".

sfsholden: In this particular scenario I think "1" is appropriate. We're printing this message in the event that a server start failed, but at this point we might not know the reason why. Once we have those specific use cases I think using a different error code would be wise.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
lcampos: @amphro is the use of `SFDX_JSON_TO_STDOUT: 'true'` exclusive to toolbelt or should plugins also honor that flag ?

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
mysticflute: yeah we didn't have anything created for this. my bad for leaving TODOs without work items... I try not to do that.

PR #2162, opened 2020-05-02
Create release branch using CircleCI

lcampos:
### What does this PR do?
Updates CircleCI config to have a scheduled job that creates the release branch every monday at 7pm.

@W-7504646@


.circleci/config.yml
```diff
@@ -154,6 +176,19 @@ jobs:
               Write-Host "Forcing build failure due to test failure(s) :'("
               $host.SetShouldExit(1)
             }
+  create-release-branch:
```
lcampos, line 179: New job that runs the create-release-branch script

.circleci/config.yml
```diff
@@ -187,3 +222,14 @@ workflows:
       - run-tests:
           run-unit-tests: false
           context: vscode-test
+
+  cut-release:
```
lcampos, line 226: Scheduled job that runs every monday at 7pm from the develop branch. This is a cool tool to verify the cron setting https://crontab.guru/

scripts/create-release-branch.js
```diff
@@ -1,28 +1,38 @@
...
+  checkVSCodeVersion,
+  checkLernaInstall,
+  checkBaseBranch
+} = require('./validation-utils');
 const logger = require('./logger-util');
 
 shell.set('-e');
 shell.set('+v');
 
+const currentVersion = require('../packages/salesforcedx-vscode/package.json')
```
lcampos, line 14: Had to make some changes in order to dynamically get the weekly version number. So we'll get the current version from one of the package.json's, bump the minor version and then set it as `SALESFORCEDX_VSCODE_VERSION`. The rest is the same logic as before.

sfsholden: I think we should modify this logic so that if the SALESFORCEDX_VSCODE_VERSION env has not already been set, we will fire this logic. Otherwise this prevents a user from overriding it without making script changes.

sfsholden: Also, how frequently do we want to bump the major version?

lcampos: > Otherwise this prevents a user from overriding it without making script changes

I don't want a user to override this value, at least not when running the build via CircleCI. 

> Also, how frequently do we want to bump the major version?

That will keep happening on every mayor platform release we'll update the mayor version (3 times a year).

jag-sfdc: Is there a way to pass in arguments to this script instead of env setting? If so, that will be even better so we don't end up reusing an old env setting.

lcampos: @jag-sfdc yes, we could have this code for extracting the value from a package.json as part of the circleci step in the config file that calls this script. I decided to have it here since it's easier to test outside of circleci.

The other option would be to have the value as a real build environment variable but that would mean someone would have to manually change that config on CircleCI everytime we decide to release (at least once a week).

PR #2168, opened 2020-05-04
Fix for serverStarted check

sfsholden:
### What does this PR do?
Fix for regression caused by initial PR. When stopping the local development server, the user would be presented with errors that they should not see.

@W-7462993@

### Functionality Before
stop_errors

### DIFF
--- a/packages/salesforcedx-vscode-lwc/test/vscode-integration/commands/forceLightningLwcStart.test.ts
+++ b/packages/salesforcedx-vscode-lwc/test/vscode-integration/commands/forceLightningLwcStart.test.ts
@@ -261,6 +261,19 @@ describe('forceLightningLwcStart', () => {
         );
       });
 
+      it('shows no error when server is stopping', () => {
+        const executor = new ForceLightningLwcStartExecutor();
+        const fakeExecution = new FakeExecution(executor.build());
+        cliCommandExecutorStub.returns(fakeExecution);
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-lwc/test/vscode-integration/commands/forceLightningLwcStart.test.ts
```diff
@@ -261,6 +261,19 @@ describe('forceLightningLwcStart', () => {
         );
       });
 
+      it('shows no error when server is stopping', () => {
```
sfsholden, line 264: Added a new test to ensure that no errors were printed by the forceLightningLwcStart command when shutting down the server normally.

packages/salesforcedx-vscode-lwc/test/vscode-integration/commands/forceLightningLwcStop.test.ts
```diff
@@ -32,4 +61,26 @@ describe('forceLightningLwcStop', () => {
     await forceLightningLwcStop();
     sinon.assert.calledOnce(stopStub);
   });
+
+  it('shows successful server stop', async () => {
```
sfsholden, line 65: Also added a test for the Stop command to ensure it prints the successful stop text during this scenario.

PR #2169, opened 2020-05-04
Release configuration

lcampos:
### What does this PR do?
Updates CircleCI config to release from master branch.

@W-7504646@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -298,13 +298,11 @@ workflows:
   commit-workflow:
     jobs:
       - build-all
-      - run-tests:
-          run-unit-tests: true
-          context: vscode-test
+      - unit-tests
       - run-tests:
           run-unit-tests: false
           context: vscode-test
-      - release:
+      - publish:
           filters:
             branches:
               only:
@@ -325,9 +323,7 @@ workflows:
                 - master
     jobs:
       - build-all
-      - run-tests:
-          run-unit-tests: true
-          context: vscode-test
+      - unit-tests
       - run-tests:
           run-unit-tests: false
           context: vscode-test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,6 +265,13 @@ jobs:
             git tag v${GUS_BUILD}
             git push origin v${GUS_BUILD}
             git push origin master
+      - run:
+          name: 'Store artifacts that were published'
+          command: |
+            mkdir extensions
+            find packages -name "*.vsix" -type f -exec cp {} ./extensions \;
+      - store_artifacts:
+          path: ./extensions
       - run: *gus-close-change-mgmt
 
   create-release-branch:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,20 @@ _refs:
       git config credential.helper 'cache --timeout=120'
       git config user.email "$GH_EMAIL"
       git config user.name "Release Bot"
+  gus-create-change-mgmt: &gus-create-change-mgmt
+    name: 'Gus create change case management'
+    command: |
+      npm install -g @salesforce/change-case-management
+      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
+      echo "Gus build is ${GUS_BUILD}"
+      sfchangecase create -i 500B0000005YGshIAG --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
+      sfchangecase check --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
+  gus-close-change-mgmt: &gus-close-change-mgmt
+    name: 'Gus close change case management'
+    command: |
+      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
+      echo "Gus build is ${GUS_BUILD}"
+      sfchangecase update --location ${REPO_LOCATION} --release offcore.tooling.${GUS_BUILD} --status 'Closed - Deploy Successful'
 
 commands:
   build-artifacts:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -154,6 +190,55 @@ jobs:
...
+          name: 'Concatenating the contents to the proper SHA256.md'
+          command: scripts/concatenate-sha256.js
+      - run:
+          name: 'Removing the temp SHA256 file.'
+          command: rm ./SHA256
+      - run:
+          name: 'Adding the SHA256 to git.'
+          command: |
+            git add SHA256.md
+            git commit -m "[skip ci] Updated SHA256"
```
lcampos, line 230: Adding `[skip ci]` to the commit message allows us to skip building/releasing again after pushing these updates back to the master branch.

.circleci/config.yml
```diff
@@ -154,6 +190,55 @@ jobs:
               Write-Host "Forcing build failure due to test failure(s) :'("
               $host.SetShouldExit(1)
             }
+  publish:
```
lcampos, line 193: Very similar sequence as on the current scripts. FYI, the current scripts stopped working because of a new redirect in the CircleCI api.

.circleci/config.yml
```diff
@@ -168,6 +253,19 @@ workflows:
       - run-tests:
           run-unit-tests: false
           context: vscode-test
+      - release:
```
lcampos, line 256: Only runs when updates are pushed to `master` and if all the tests are successful

.circleci/config.yml
```diff
@@ -168,6 +253,19 @@ workflows:
...
+          filters:
+            branches:
+              only:
+                - master
+          requires:
+            - build-all
+            - run-tests:
```
lcampos, line 263: Can't call these jobs with parameters, I'll refactor them

.circleci/config.yml
```diff
@@ -12,16 +12,52 @@ version: 2.1
...
+      git config credential.helper 'cache --timeout=120'
+      git config user.email "$GH_EMAIL"
+      git config user.name "Release Bot"
+  gus-create-change-mgmt: &gus-create-change-mgmt
+    name: 'Gus create change case management'
+    command: |
+      npm install -g @salesforce/change-case-management
+      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
+      echo "Gus build is ${GUS_BUILD}"
+      sfchangecase create -i 500B0000005YGshIAG --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
```
jag-sfdc, line 46: Is it ok to have a GUS Id checked in ?

.circleci/config.yml
sfsholden: Could you give more info on the redirect?

.circleci/config.yml
lcampos: I don't think it's a problem since we're already posting links to work items in some of the issues but I'll move this to an environment variable.

.circleci/config.yml
```diff
@@ -154,6 +227,55 @@ jobs:
...
+          command: scripts/concatenate-sha256.js
+      - run:
+          name: 'Removing the temp SHA256 file.'
+          command: rm ./SHA256
+      - run:
+          name: 'Adding the SHA256 to git.'
+          command: |
+            git add SHA256.md
+            git commit -m "[skip ci] Updated SHA256"
+      - run:
```
sfsholden, line 268: Does this mean we no longer have to worry about the aws publish piece?

.circleci/config.yml
lcampos: Sure, requesting the artifacts of a build via the api now returns a url to where they are hosted instead of immediately providing the artifact payload after using the details in the initial json response. You can see the issue by running `scripts/download-vsix-from-circleci.js` locally.

.circleci/config.yml
```diff
@@ -154,6 +227,55 @@ jobs:
...
+          name: 'Publishing to the VS Code Marketplace.'
+          command: npm run vscode:publish
+      - run:
+          name: 'Create git tag (e.g. v48.1.0) & push changes back to master branch'
```
sfsholden, line 272: In the publish-circleci script, this step comes after `npm run vscode:publish`. Does the order matter?

.circleci/config.yml
lcampos: That's correct, we're deprecating that step.

.circleci/config.yml
```diff
@@ -12,16 +12,52 @@ version: 2.1
 orbs:
   win: circleci/windows@2.2.0
 
+_refs:
```
sfsholden, line 15: Simple question, but I'd like to make sure my understanding is correct. Does this automatically kick off once we merge the release branch into master? Also, what will happen if we address minor doc changes by merging into master now?

.circleci/config.yml
lcampos: I think you meant before instead of after. I changed the order of those operations cause I do not want to generate a git tag or push any commit to the branch unless releasing to the marketplace is successful.

.circleci/config.yml
lcampos: > Does this automatically kick off once we merge the release branch into master?

That is correct, the release process will now be to just merge the release branch into the master branch.

> what will happen if we address minor doc changes by merging into master now?

We should talk about those on a case by case basis, hopefully we won't have to do many of those but if we do we can add `[skip ci]` to the commit message which will prevent circleci from running the release job.

.circleci/config.yml
sfsholden: As a separate work item - 
1) Could we add a new script that is doing most of this work here (specifically from 'Packaging extensions' -> 'Create git tag (e.g. v48.1.0) & push changes back to master branch')? 
2) I think we're going to need to do a lot of script clean up so it's not confusing for anyone that may have to do this process manually for whatever reason.

lcampos: @sfsholden agreed, I'll create a work item to follow up on these points.

.circleci/config.yml
```diff
@@ -37,6 +37,20 @@ _refs:
       git config credential.helper 'cache --timeout=120'
       git config user.email "$GH_EMAIL"
       git config user.name "Release Bot"
+  gus-create-change-mgmt: &gus-create-change-mgmt
+    name: 'Gus create change case management'
+    command: |
+      npm install -g @salesforce/change-case-management
+      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
+      echo "Gus build is ${GUS_BUILD}"
+      sfchangecase create -i ${GUS_TEMPLATE_ID} --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
```
jag-sfdc, line 46: How is targetusername set for this?

lcampos: By an environment variable that is consumed by the `sfchangecase` cli, here's the repo https://github.com/forcedotcom/change-case-management

PR #2172, opened 2020-05-05
Add check for server address already in use

sfsholden:
### What does this PR do?
Adds a check for the error text 'EADDRINUSE: address already in use'. When this happens, we will call the error handling logic with exit code 98.

@W-7148356@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -121,7 +121,7 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
         if (data.toString().includes('Server start up failed')) {
           errorCode = 1;
         }
-        if (data.toString().includes('EADDRINUSE: address already in use')) {
+        if (data.toString().includes('EADDRINUSE')) {
           errorCode = 98;
         }
         if (errorCode !== -1) {
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -84,6 +84,7 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     channelService.showChannelOutput();
 
     let serverStarted = false;
+    let printedError = false;
 
     const progress = new Subject();
     ProgressNotification.show(
@@ -115,7 +116,7 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     });
 
     execution.stderrSubject.subscribe(async data => {
-      if (data) {
+      if (!printedError && data) {
         let errorCode = -1;
         if (data.toString().includes('Server start up failed')) {
           errorCode = 1;
@@ -131,17 +132,21 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
             errorCode
           );
           progress.complete();
+          printedError = true;
         }
       }
     });
 
     execution.processExitSubject.subscribe(async exitCode => {
-      this.handleErrors(
-        cancellationToken,
-        serverHandler,
-        serverStarted,
-        exitCode
-      );
+      if (!printedError) {
+        this.handleErrors(
+          cancellationToken,
+          serverHandler,
+          serverStarted,
+          exitCode
+        );
+        printedError = true;
+      }
     });
 
     notificationService.reportExecutionError(
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -115,9 +115,23 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     });
 
     execution.stderrSubject.subscribe(async data => {
-      if (data && data.toString().includes('Server start up failed')) {
-        this.handleErrors(cancellationToken, serverHandler, serverStarted, 1);
-        progress.complete();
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -21,6 +21,8 @@ export const messages = {
   force_lightning_lwc_start_text: 'SFDX: Start Local Development Server',
   force_lightning_lwc_start_not_found:
     'To run this command, first install the @salesforce/lwc-dev-server plugin. For more info, see [https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev](https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev).',
+  force_lightning_lwc_start_addr_in_use:
```
sfsholden, line 24: @hbc-bindu Please let me know if you'd like to update this text. I was debating about adding more information on how to kill the process, but not sure if we should actually include that piece. It's also a long message already.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -84,6 +84,7 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     channelService.showChannelOutput();
 
     let serverStarted = false;
+    let printedError = false;
```
sfsholden, line 87: Currently we receive the same error ''EADDRINUSE: address already in use" multiple times. In addition, webruntime will kill the process with an exit code of 1. This helps prevent us from printing the same error to the user multiple times, along with the generic error from force_lightning_lwc_start_failed.

sfsholden:
This does not address the issue where the same error for EADDRINUSE is printed to the console multiple times. This must be fixed within the `start` method of `server.ts` within webruntime. When this error is seen, it does not throw it back to `LocalDevServer.ts`. It eventually exits with an exit code of 1 - meaning we don't have a way to handle this error handling on our end.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -115,19 +116,37 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
...
-      if (data && data.toString().includes('Server start up failed')) {
-        this.handleErrors(cancellationToken, serverHandler, serverStarted, 1);
-        progress.complete();
+      if (!printedError && data) {
+        let errorCode = -1;
+        if (data.toString().includes('Server start up failed')) {
+          errorCode = 1;
+        }
+        if (data.toString().includes('EADDRINUSE: address already in use')) {
+          errorCode = 98;
```
sfsholden, line 125: Referenced this site for determining error code: https://mariadb.com/kb/en/operating-system-error-codes/

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -115,19 +116,37 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
...
     execution.stderrSubject.subscribe(async data => {
-      if (data && data.toString().includes('Server start up failed')) {
-        this.handleErrors(cancellationToken, serverHandler, serverStarted, 1);
-        progress.complete();
+      if (!printedError && data) {
+        let errorCode = -1;
+        if (data.toString().includes('Server start up failed')) {
+          errorCode = 1;
+        }
+        if (data.toString().includes('EADDRINUSE: address already in use')) {
```
lcampos, line 124: I would scope this down to only checking for `EADDRINUSE` since that's the error code and the message `address already in use` could change.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
mysticflute: I think we should try to handle this on the webruntime side, to make sure it only prints once and sends the proper exit code. but this looks ok for now

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -153,10 +172,12 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     if (!serverStarted && !cancellationToken.isCancellationRequested) {
       let message = nls.localize('force_lightning_lwc_start_failed');
 
-      // TODO proper exit codes in lwc-dev-server for address in use, auth/org error, etc.
```
xyc, line 156: @sfsholden @mysticflute Is there an existing pattern in lwc-dev-server of process exit code? Are we going to keep using this pattern of generating an "exitCode" from std err?

mysticflute: No, we want webruntime to exit with the specific codes when applicable, and also lwc-dev-server itself to exit with the proper code. Then in the extension we simply check for the exit code. I think this solution is temporary until webruntime can be updated to exit with the right code.

mysticflute: But if a proper exit code doesn't exist then it should still just be `1`.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
hbc-bindu: @sfsholden, very minor changes:
The local development server couldn’t start as the address is already in use. To resolve, do one of these: \n1) Stop the local dev server running on any another instance.\n 2) Change the default port [Configuration for Projects (Optional)](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.get_started_local_dev_setup).\n 3) Kill the process running on the specified port.

PR #2174, opened 2020-05-06
Local Dev Scratch Org expire message @W-7485707@

xyc:
### What does this PR do?
Provide a better error message when scratch org expires

@W-7485707@

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -35,6 +35,16 @@ const SfdxCommandletExecutor = sfdxCoreExports.SfdxCommandletExecutor;
 const logName = 'force_lightning_lwc_start';
 const commandName = nls.localize(`force_lightning_lwc_start_text`);
 
+/**
+ * Hints for providing a user-friendly error message / action.
+ * Hints come from the stderr output of lwc-dev-server. (We should move this to lwc-dev-server later)
+ */
+export const enum errorHints {
+  SERVER_STARTUP_FALIED = 'Server start up failed',
+  ADDRESS_IN_USE = 'EADDRINUSE',
+  INACTIVE_SCRATCH_ORG = 'Error authenticating to your scratch org. Make sure that it is still active'
+}
+
 export interface ForceLightningLwcStartOptions {
   /** whether to automatically open the browser after server start */
   openBrowser: boolean;
@@ -44,6 +54,7 @@ export interface ForceLightningLwcStartOptions {
 
 export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
   private readonly options: ForceLightningLwcStartOptions;
+  private errorHint?: string;
 
   constructor(options: ForceLightningLwcStartOptions = { openBrowser: true }) {
     super();
@@ -124,6 +135,9 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
         if (data.toString().includes('EADDRINUSE')) {
           errorCode = 98;
         }
+        if (data.toString().includes(errorHints.INACTIVE_SCRATCH_ORG)) {
+          this.errorHint = errorHints.INACTIVE_SCRATCH_ORG;
+        }
         if (errorCode !== -1) {
           this.handleErrors(
             cancellationToken,
@@ -172,7 +186,10 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     if (!serverStarted && !cancellationToken.isCancellationRequested) {
       let message = nls.localize('force_lightning_lwc_start_failed');
 
-      if (exitCode === 112) {
+      if (
+        exitCode === 1 &&
+        this.errorHint === errorHints.INACTIVE_SCRATCH_ORG
+      ) {
         message = nls.localize('force_lightning_lwc_inactive_scratch_org');
       }
       if (exitCode === 127) {
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -129,10 +129,10 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     execution.stderrSubject.subscribe(async data => {
       if (!printedError && data) {
         let errorCode = -1;
-        if (data.toString().includes('Server start up failed')) {
+        if (data.toString().includes(errorHints.SERVER_STARTUP_FALIED)) {
           errorCode = 1;
         }
-        if (data.toString().includes('EADDRINUSE')) {
+        if (data.toString().includes(errorHints.ADDRESS_IN_USE)) {
           errorCode = 98;
         }
         if (data.toString().includes(errorHints.INACTIVE_SCRATCH_ORG)) {
--- a/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
+++ b/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
@@ -172,6 +172,9 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     if (!serverStarted && !cancellationToken.isCancellationRequested) {
       let message = nls.localize('force_lightning_lwc_start_failed');
 
+      if (exitCode === 112) {
+        message = nls.localize('force_lightning_lwc_inactive_scratch_org');
+      }
       if (exitCode === 127) {
         message = nls.localize('force_lightning_lwc_start_not_found');
       }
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -23,6 +23,8 @@ export const messages = {
     'To run this command, first install the @salesforce/lwc-dev-server plugin. For more info, see [https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev](https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev).',
   force_lightning_lwc_start_addr_in_use:
     'The local development server could not start because the address is already in use. To resolve either: \n1) Stop the local dev server if it is running in another instance.\n or 2) Change the default port [Configuration for Projects (Optional)](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.get_started_local_dev_setup).\n or 3) Kill the process running on the specified port.',
+  force_lightning_lwc_inactive_scratch_org:
+    'The local development server could not start because the scratch org is not active. To resolve, run "SFDX: Create a Default Scratch Org" to create a new scratch org.',
```
xyc, line 27: @hbc-bindu Please let me know if we need to update this message. Thanks!

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -172,6 +172,9 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
     if (!serverStarted && !cancellationToken.isCancellationRequested) {
       let message = nls.localize('force_lightning_lwc_start_failed');
 
+      if (exitCode === 112) {
+        message = nls.localize('force_lightning_lwc_inactive_scratch_org');
```
AnanyaJha, line 176: I noticed in the PR that shelby opened up she added some more error handling logic for the specific exit code in `execution.stderrSubject.subscribe` is that something we have to do here? Or is that exit code automatically returned when the scratch org is expired?

AnanyaJha:
PR looks good to me! Just checking, is this getting merged into shelby's branch rather than develop?

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
hbc-bindu: @xyc , consider rephrasing as:
The local development server couldn’t start as there’s no active scratch org. Run "SFDX: Create a Default Scratch Org" to create a scratch org.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
sfsholden: Good question.

My changes were a bit different - I was looking for specific text printed to stderr and then called this same logic with the new error code. In my case, the process was not exiting with the error code I wanted to see - which is why I was examining stderr. If in the future we received the correct error code via processExitSubject, we would still process it the same way in this handleErrors method. 

It sounds like this code execution is hitting processExitSubject with the exit code 112.

AnanyaJha: Ah okay that makes sense, thanks :)

sfsholden:
> PR looks good to me! Just checking, is this getting merged into shelby's branch rather than develop?

I'll be merging my changes in just a moment. We can rebase this branch off of develop shortly.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
xyc: Thanks @sfsholden . @AnanyaJha Yeah I was going to let lwc-dev-server return 112 when the scratch org expires. FYI the correspondent PR in lwc-dev-server: https://github.com/forcedotcom/lwc-dev-server/pull/189

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
xyc: Updated, thanks @hbc-bindu

xyc:
Rebased off develop. Looks like this needs a re-review.

sfsholden:
> Rebased off develop. Looks like this needs a re-review.

Looks like it's blocked waiting on a review from Lightning Tooling.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -23,6 +23,8 @@ export const messages = {
     'To run this command, first install the @salesforce/lwc-dev-server plugin. For more info, see [https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev](https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.get_started_local_dev).',
   force_lightning_lwc_start_addr_in_use:
     "The local development server couldn't start as the address is already in use. To resolve, do one of these\n 1) Stop the local dev server running on any another instance.\n or 2) Change the default port [Configuration for Projects (Optional)](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.get_started_local_dev_setup).\n 3) Kill the process running on the specified port.",
+  force_lightning_lwc_inactive_scratch_org:
+    'The local development server couldn’t start as your scratch org is not active. Run "SFDX: Create a Default Scratch Org" to create a scratch org, or run "SFDX: Set a Default Org" to switch to an active scratch org.',
```
xyc, line 27: @ekapner I updated the user messaging here. Would you have additional feedback?

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts
```diff
@@ -124,6 +135,9 @@ export class ForceLightningLwcStartExecutor extends SfdxCommandletExecutor<{}> {
         if (data.toString().includes('EADDRINUSE')) {
```
sfsholden, line 135:
```suggestion
        if (data.toString().includes(errorHints.ADDRESS_IN_USE)) {
```

sfsholden: Let's also do the same for the Server start up failed message.

sfsholden:
:shipit:

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
ekapner: I worry about providing a URL to docs that could become obsolete at some future point in time. So, let me see if we can be more generic.

lcampos: @ekapner it's a valid concern but this is better than not giving users a path forward and expecting them to search the internet for an answer. Most of the urls we provide are also under our control which helps in mitigating the problems. When we've ran into the situation of urls being out of date we're able to address it pretty fast.

xyc: Are you referring to the URL in `force_lightning_lwc_start_addr_in_use`?

ekapner: 'To run this command, install the @salesforce/lwc-dev-server plugin. For more info, see the Local Development Server topics in the Lightning Web Components documentation.,
  force_lightning_lwc_start_addr_in_use:
    "The local development server can't start because the address is already in use. To fix, try one of these options: \n 1) Stop the local dev server running on any another instance.\n or 2) Change the default port see the Local Development Server topics in the Lightning Web Components documentation.\n 3) Kill the process running on the specified port.",
  force_lightning_lwc_inactive_scratch_org:
    'The local development server can't start because your scratch org is not active. Run "SFDX: Create a Default Scratch Org" to create a scratch org, or run "SFDX: Set a Default Org" to select an active scratch org.',

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
```diff
@@ -23,6 +23,8 @@ export const messages = {
...
   force_lightning_lwc_start_failed:
     'The local development server was not able to start.',
```
ekapner, line 29: The local development server can't start.'

<ekapner>any other help we can provide?

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
ekapner: @lcampos @xyc - ok, our style standards are to avoid the direct URLs - but if you both feel like there's low risk here, let's keep them for now. I agree, it's always desirable to point users directly to the info they need.

xyc: Thanks @ekapner . I'll incorporate your comments & keep the URLs. In general I agree with @lcampos that providing a url is more actionable for users. I also wonder if it would be possible to provide a link redirect when such deprecation happens.

packages/salesforcedx-vscode-lwc/src/messages/i18n.ts
xyc: This is the default message when we can't determine a cause for server start failure

xyc: Not sure if we should do more around here. But the Runtime Resiliency epic (which this story is a part of) should mitigate this and locate the reason more precisely.

sfsholden: To expand, I don't think this will ever be presented to the user. Like Xiaoyi mentioned, this was added in the event that we couldn't determine the cause of a server start failure. This was added to ensure that the call to start the server would be halted from the extension side in the event that something unexpected occurred.

PR #2182, opened 2020-05-07
Update npm install for publish job

lcampos:
### What does this PR do?
Update install step on publish job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,10 +224,10 @@ jobs:
       - checkout
       - restore_cache: *restore_cache
       - run: *gh-config
-      - run: *install
       - run:
           name: 'Install build dependencies'
           command: |
+            npm install
             npm install -g lerna@3.13.1
             npm install -g vsce
             npm install -g shelljs@0.8.3


.circleci/config.yml
```diff
@@ -224,10 +224,10 @@ jobs:
       - checkout
       - restore_cache: *restore_cache
       - run: *gh-config
-      - run: *install
```
lcampos, line 227: Having this as an alias is causing install issues so I'm updating it to a normal `npm install` in the job.

.circleci/config.yml
```diff
@@ -224,10 +224,10 @@ jobs:
...
       - run:
           name: 'Install build dependencies'
           command: |
+            npm install
```
brpowell, line 230:
```suggestion
            npm install --only=prod
```
Is this needed, or no?

AnanyaJha: what does the `--only=prod` do?

lcampos: Not needed, setting node_env to production in the compile step will take care of what we need

brpowell: @AnanyaJha It avoids installing the dev dependencies for a production build

PR #2193, opened 2020-05-12
Honor CLI telemetry disablement setting

rcoringrato-sfdc:
### What does this PR do?
- Honor CLI telemetry setting for extension telemetry
- Disable telemetry for spawned CLI functionality when extension telemetry is disabled

@W-7263491@

### Functionality Before
Even when telemetry is opted-out in the extension telemetry is still enabled for any spawned CLI functionality. Extension did not honor the CLI disablement setting.

### Functionality After
Opting out of extension telemetry will disable telemetry for any spawned CLI functionality. Disabling CLI telemetry will disable extension telemetry.

### DIFF
--- a/contributing/telemetry.md
+++ b/contributing/telemetry.md
@@ -6,6 +6,8 @@ Similar to what vscode does, we are reporting on certain events that happen in t
 
 All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is resposible of showing an information message to users and properly initializing the service for other extensions in this repository to leverage. In the same line as vscode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
 
+Additionally, if telemetry is disabled for the Salesforce CLI via the `disableTelemetry` configuration setting, or the `SFDX_DISABLE_TELEMETRY` environment variable, the telemetryService will not send data.
+
 ## Adding telemetry to an extension
 
 - Add [vscode-extension-telemetry](https://github.com/Microsoft/vscode-extension-telemetry#readme) as a devDependency
--- a/contributing/telemetry.md
+++ b/contributing/telemetry.md
@@ -4,9 +4,9 @@ Similar to what vscode does, we are reporting on certain events that happen in t
 
 ## How it works
 
-All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is resposible of showing an information message to users and properly initializing the service for other extensions in this repository to leverage. In the same line as vscode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
+All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is responsible for showing an information message to users and properly initializing the service for other extensions in this repository to leverage. Similar to VSCode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
 
-Additionally, if telemetry is disabled for the Salesforce CLI via the `disableTelemetry` configuration setting, or the `SFDX_DISABLE_TELEMETRY` environment variable, the telemetryService will not send data.
+Additionally, if telemetry is disabled for Salesforce CLI using the `disableTelemetry` configuration setting or the `SFDX_DISABLE_TELEMETRY` environment variable, the telemetryService doesn't send data.
 
 ## Adding telemetry to an extension
 
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -77,8 +77,10 @@ import { registerPushOrDeployOnSave, sfdxCoreSettings } from './settings';
 import { taskViewService } from './statuses';
 import { telemetryService } from './telemetry';
 import {
+  getRootWorkspacePath,
   hasRootWorkspace,
   isCLIInstalled,
+  isCLITelemetryAllowed,
   showCLINotInstalledMessage
 } from './util';
 import { OrgAuthInfo } from './util/authInfo';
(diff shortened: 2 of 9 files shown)


brpowell:
Cool 👍 . I think the other extensions use the core extension's telemetry setting right? So only the core version of the module needs to have these changes?

rcoringrato-sfdc:
@brpowell Yes, the other extensions reference the core telemetry service to setup their own, so they should inherit the settings.

contributing/telemetry.md
```diff
@@ -6,6 +6,8 @@ Similar to what vscode does, we are reporting on certain events that happen in t
 
 All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is resposible of showing an information message to users and properly initializing the service for other extensions in this repository to leverage. In the same line as vscode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
 
+Additionally, if telemetry is disabled for the Salesforce CLI via the `disableTelemetry` configuration setting, or the `SFDX_DISABLE_TELEMETRY` environment variable, the telemetryService will not send data.
```
rcoringrato-sfdc, line 9: @ekapner Can you review these changes, please.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -34,3 +42,26 @@ export function showCLINotInstalledMessage() {
...
+): Promise<boolean> {
+  if (isCLIInstalled()) {
+    const forceConfig = await new ForceConfigGet().getConfig(
+      projectPath,
+      SFDX_CONFIG_DISABLE_TELEMETRY
+    );
+    const disabledConfig =
+      forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || 'false';
+    const isDisabled = disabledConfig === 'true' ? true : false;
+    return !isDisabled;
```
sfsholden, line 64: I find this piece a little hard to read. Would something like this be simpler?

return forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) ? !forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) : true;

contributing/telemetry.md
ekapner: Although not highlighted in green: The service in `salesforce-vscode-core` is **responsible**(typo) of showing an...

Instead of: In the same line as vscode,
Please change: Similar to VSCode, ...

Edits for green/highlighted text:
Additionally, if telemetry is disabled for Salesforce CLI using the `disableTelemetry` configuration setting or the `SFDX_DISABLE_TELEMETRY` environment variable, the telemetryService doesn't send data.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -441,9 +443,12 @@ async function setupOrgBrowser(
 export async function activate(context: vscode.ExtensionContext) {
   const extensionHRStart = process.hrtime();
   // Telemetry
+  const isCliTelemetryAllowed = await isCLITelemetryAllowed(
```
lcampos, line 446: I don't think we need to add this check here, we can just have everything be part of the internals of the telemetry service.

xyc: +1. I'm also wary of adding something like this to the activate function

rcoringrato-sfdc: @xyc There is precedent for async behavior in activate(). What specifically is your concern?

xyc: Performance primarily. This puts everything below on hold. Especially looks like it involves a CLI command. If this has to be in `activate` I would suggest to wrap in something like `Promise.all`, otherwise I would agree with what @lcampos suggests and read the value on demand.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -440,10 +442,9 @@ async function setupOrgBrowser(
 
 export async function activate(context: vscode.ExtensionContext) {
   const extensionHRStart = process.hrtime();
-  // Telemetry
   const machineId =
     vscode && vscode.env ? vscode.env.machineId : 'someValue.machineId';
-  telemetryService.initializeService(context, machineId);
+  await telemetryService.initializeService(context, machineId);
```
xyc, line 447: I think this is still a costly function to await on in `activate`.
1. One approach I could think of is make `isTelemetryEnabled` of `TelemetryService` an async function, and do something like:
```
  public sendExtensionActivationEvent(hrstart: [number, number]): void {
    this.isTelemetryEnabled().then(()=>{
        // ...
    });
```
This requires a bigger refactor though.

2. Another approach is to move this down further. My main concern is about the time for commands to become available (`registerCommands`).

cc @lcampos

lcampos: @xyc I've measured this and refactoring it is not a high priority. The highest impact item in the activation process are other checks which we are addressing as part of the perf work. Once we're done with the perf work a lot of the checks here will go away.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
```diff
@@ -62,14 +69,16 @@ export class TelemetryService {
       );
       this.context.subscriptions.push(this.reporter);
     }
+
+    this.setCliTelemetryEnabled(this.isTelemetryEnabled() && !isDevMode);
```
lcampos, line 73: Why do we need to consider `!isDevMode` here ?

rcoringrato-sfdc: @lcampos So that CLI telemetry is not captured for dev mode.

packages/salesforcedx-vscode-core/src/index.ts
lcampos: @rcoringrato-sfdc @xyc I want to clarify that my ask was not based on performance but about where the code belongs. Having the check be part of the params a consumer needs to provide did not make sense since that should be part of the internal checks the service runs. The service itself is checking for different environment variables, checking the CLI telemetry preference should be one of those and there is no value in having it be done outside of the service since it's not needed anywhere else.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
lcampos: I think that the CLI should be doing that internally, just like we are doing a few lines above. If this is not the case we should open a bug on it. It shouldn't be the consumer's responsibility to determine if it's a dev scenario or not.

packages/salesforcedx-vscode-core/src/index.ts
rcoringrato-sfdc: @lcampos Understood with regards to the placement of the check. My most recent update places the check in TelemetryService initialization.

packages/salesforcedx-vscode-core/src/telemetry/telemetry.ts
rcoringrato-sfdc: @lcampos I'll remove that check and open a bug if appropriate.

Issue #2195, opened 2020-05-13
Show Output Channel for Auth Commands in Container


dehru:
# Summary
When the user is running vscode in a container ( like remote development with docker ), the commands for authorize a devhub and authorize an org need to display the channel output since the browser can't be launched by the CLI.  The output channel includes the information the user needs to complete their device flow login.

Currently, the output panel is there, but it is not focused like it used to be.

### Steps To Reproduce:

1.  Run vscode ( using remote development ) with the salesforce extensions installed.
1.  Use the command palette to launch Authorize a Dev Hub or Authorize an Org.

### Expected result

Salesforce CLI channel output should be focused for the user so they know there are steps they need to take.  ( copy TOKEN, Go to URL )

### Actual result
Salesforce CLI channel is not focused.  User has no idea they need to take action.

### Additional information

**VS Code Version**: Version: 1.46.0-insider
**SFDX CLI Version**: sfdx-cli/7.54.4-8ff9ba9cc5
**OS and version**: OS: Darwin x64 18.7.0

dehru:
I'll be submitting a PR to fix this issue.

lcampos:
Closing since this was fixed and released

PR #2196, opened 2020-05-13
Show the Salesforce CLI channel output when executing auth commands in a container

dehru:
### What does this PR do?
This PR flips the showChannelOutput boolean based on the environment variable SFDX_CONTAINER_MODE.  When the extensions are running in a container, like visual studio remote development in docker, where the browser is not available to the sfdx cli, the user can see the instructions to manually launch the browser to a url with a token.

### Functionality Before
I need to manually switch to the output channel
auth-devhub-vsonline

### Functionality After
The command automatically shows the output channel
hub-auth-device

#2195, @W-7491069@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceAuthDevHub.ts
@@ -33,7 +33,7 @@ import { ConfigSource, OrgAuthInfo } from '../util/index';
 import { ForceAuthDemoModeExecutor } from './forceAuthWebLogin';
 
 export class ForceAuthDevHubExecutor extends SfdxCommandletExecutor<{}> {
-  protected showChannelOutput = false;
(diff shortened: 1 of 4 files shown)


praksb:
@dehru Will the output panel also show up when the entire bottom window is closed and the user runs "Authorize a Dev Hub" or "Authorize an Org" from the command palette? I think we also want to fix that.

dehru:
> @dehru Will the output panel also show up when the entire bottom window is closed and the user runs "Authorize a Dev Hub" or "Authorize an Org" from the command palette? I think we also want to fix that.

@praksb - yes, this little boolean somehow magically does that!

jonnyhork:
I QA'd this PR successfully and observed the output channel opening when `SFDX_CONTAINER_MODE` was set in the env. The extension appropriately used web auth and did not display the output channel when the env variable was not set to true.

Issue #1823, opened 2019-12-03
Add setting to set --wait parameter for Apex test run


nufsty2:
**Is your feature request related to a problem? Please describe.**
With large Apex tests, the default --wait parameter is not long enough and the tests time out before coverage can be generated.

**Describe the solution you'd like**
Just a simple entry in the settings menu to set the --wait parameter value.

**Describe alternatives you've considered**
I can run "sfdx force:apex:test:report -i <ID> --wait <# minutes> -c" from the command line, but that doesn't generate a coverage report that is viewable in VSCode (to see exactly which lines are covered, etc). This is still viewable from the Developer Console, but who wants to use the Developer Console?

**Additional context**
Not a huge issue, would just be more complete and convenient to have the setting.

Andrews-Consulting:
Are you actively building the code to address this issue?  If not, do you mind if I put some thought into it?

PR #2215, opened 2020-05-26
Add wait option for force:apex:test:run

Andrews-Consulting:
### What does this PR do?
Add support for the wait option on the force:apex:test:run cli command, when the force:apex:test:run command is invoked from the extension.

#1823 

### Functionality Before
The default value for "--wait" was always used because no wait value was able to be specified for the force:apex:test:run command. 

### Functionality After
Provides a new feature setting to define the --wait value for the force:apex:test:run command.
If the value provided is less than 2, the value is ignored and the --wait option IS NOT added to the force:apex:test:run command.  If the value provided is 2 or greater, then the --wait option IS added to the force:apex:test:run command.

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
@@ -30,7 +30,7 @@ export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{
   protected shouldGetCodeCoverage: boolean = false;
   protected builder: SfdxCommandBuilder = new SfdxCommandBuilder();
   private outputToJson: string;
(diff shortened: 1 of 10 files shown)


AnanyaJha:
Hi @Andrews-Consulting thanks for the contribution, this is a very helpful feature. We're currently working on rewriting this command and I believe it'll have some impact on your current implementation of the `wait` option. I think it'd be worthwhile to revisit adding this option after we've got the command in a stable condition with the rewrites- and we'll be sure to come back to this PR then 🙂

AnanyaJha:
Hi @Andrews-Consulting! Wanted to give you an update, we've made significant progress on the rewrite of the Apex test run command so we'll have to approach adding in the `wait` functionality a bit differently. I'm going to go ahead and close this PR for now and I'll log something to add this option in using our new implementation. Thanks for opening this!

PR #2218, opened 2020-05-27
Add guard for unsupported cli telemetry setting

rcoringrato-sfdc:
### What does this PR do?
Guards the check for _disableTelemetry_, which is unsupported by earlier versions of Salesforce CLI

@W-7605599@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
+++ b/packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
@@ -53,13 +53,18 @@ export function disableCLITelemetry() {
 export async function isCLITelemetryAllowed(
   projectPath: string
 ): Promise<boolean> {
-  // if (isCLIInstalled()) {
-  //   const forceConfig = await new ForceConfigGet().getConfig(
-  //     projectPath,
-  //     SFDX_CONFIG_DISABLE_TELEMETRY
-  //   );
-  //   const disabledConfig = forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || '';
-  //   return disabledConfig !== 'true';
-  // }
+  if (isCLIInstalled()) {
+    try {
+      const forceConfig = await new ForceConfigGet().getConfig(
+        projectPath,
+        SFDX_CONFIG_DISABLE_TELEMETRY
+      );
+      const disabledConfig =
+        forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || '';
+      return disabledConfig !== 'true';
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -53,13 +53,18 @@ export function disableCLITelemetry() {
...
+        SFDX_CONFIG_DISABLE_TELEMETRY
+      );
+      const disabledConfig =
+        forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || '';
+      return disabledConfig !== 'true';
+    } catch (e) {
+      console.log('Error checking cli settings: ' + e);
+    }
+  }
   return true;
```
xyc, line 69: Do we want to err on this side if an exception happens?

rcoringrato-sfdc: @xyc That mimics the prior behavior. Do you feel it should be otherwise?

xyc: Yes, especially since this default to true when a generic error happens, not only when the CLI version is earlier.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -53,13 +53,18 @@ export function disableCLITelemetry() {
...
 ): Promise<boolean> {
-  // if (isCLIInstalled()) {
-  //   const forceConfig = await new ForceConfigGet().getConfig(
-  //     projectPath,
-  //     SFDX_CONFIG_DISABLE_TELEMETRY
-  //   );
-  //   const disabledConfig = forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || '';
-  //   return disabledConfig !== 'true';
-  // }
+  if (isCLIInstalled()) {
```
lcampos, line 56: What happens if we don't check for the CLI being present ? I'm asking cause this will add a second `isCLIInstalled()` check to the activation of core and that call is not the most efficient

rcoringrato-sfdc: @lcampos Based on my testing the settings check will hang indefinitely if the cli is not installed. This case is difficult to trap without more extensive changes.

lcampos: Ok, let's keep it as is and we can have a follow up story to address the perf hit.

PR #2225, opened 2020-05-29
Add preference for completion statistics

sfsholden:
### What does this PR do?
Adds a configurable preference to pass to the Apex LSP for printing completion statistics.

### To Enable
Add the following value to the settings.json:
`"salesforcedx-vscode-apex.advanced.enable-completion-statistics": "true"`

@W-7637588@

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -269,7 +269,7 @@
           "default": false,
           "description": "%enable_sobject_refresh_on_startup_description%"
         },
-        "salesforcedx-vscode-apex.enable-completion-statistics": {
+        "salesforcedx-vscode-apex.advanced.enable-completion-statistics": {
           "type": "boolean",
           "default": false,
           "description": "%apex_completion_statistics_description%"
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -268,11 +268,6 @@
           "type": "boolean",
           "default": false,
           "description": "%enable_sobject_refresh_on_startup_description%"
-        },
-        "salesforcedx-vscode-apex.advanced.enable-completion-statistics": {
-          "type": "boolean",
-          "default": false,
-          "description": "%apex_completion_statistics_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -268,6 +268,11 @@
           "type": "boolean",
           "default": false,
           "description": "%enable_sobject_refresh_on_startup_description%"
+        },
+        "salesforcedx-vscode-apex.enable-completion-statistics": {
+          "type": "boolean",
+          "default": false,
+          "description": "%apex_completion_statistics_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-apex/package.nls.json
+++ b/packages/salesforcedx-vscode-apex/package.nls.json
@@ -8,7 +8,7 @@
   "show_error_title": "Display Error",
   "go_to_definition_title": "Go to Definition",
   "java_home_description": "Specifies the folder path to the Java 8 or Java 11 runtime used to launch the Apex Language Server (for example, /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home).",
-  "java_memory_description": "Specifies the amount of memory allocation to the Apex Language Server in MB or null to use the default value.",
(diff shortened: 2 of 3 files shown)


packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -10,10 +10,11 @@
   "java_home_description": "Specifies the folder path to the Java 8 or Java 11 runtime used to launch the Apex Language Server (for example, /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home).",
   "java_memory_description": "Specifies the amount of memory allocation to the Apex Language Server in MB or null to use the default value.",
   "apex_semantic_errors_description": "Allow Apex Language Server to surface semantic errors",
+  "apex_completion_statistics_description": "List statistics for Apex Completions",
```
sfsholden, line 13: @brpowell I could not find a way to have this be hidden from the settings UI. Do you happen to have more info on this?

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -268,6 +268,11 @@
           "type": "boolean",
           "default": false,
           "description": "%enable_sobject_refresh_on_startup_description%"
+        },
+        "salesforcedx-vscode-apex.enable-completion-statistics": {
```
sfsholden, line 272: @brpowell I could not find a way to have this be hidden from the settings UI. Do you happen to have more info on this?

packages/salesforcedx-vscode-apex/package.nls.json
sfsholden, line 13: @ekapner If I can, I want to hide this from customers. In the event that I cannot, does this wording look okay to you?

packages/salesforcedx-vscode-apex/package.json
xyc: Could you talk a little bit about the reasoning why this need to be hidden from UI?

sfsholden: I wouldn't say it needs to be hidden. It sounded like this was the preference during standup since we already have a lot of settings that users see on a regular basis. The only times we'd really want this to be enabled are 1) we are doing development and want to see the differences with what we make or 2) a customer is seeing an issue with completion and we ask them to enable this preference to get more info.

xyc: Okay in that case I might add something like "debugOption" or "advanced" in the namespace: (So that it becomes `salesforcedx-vscode-apex.advanced.enable-completion-statistics`)

Similar to `salesforcedx-vscode-apex.java.home`
Screen Shot 2020-05-29 at 7 06 53 PM

Reference:
- https://code.visualstudio.com/api/references/contribution-points#Configuration-schema - Search for "hierarchy".

packages/salesforcedx-vscode-apex/package.nls.json
ekapner: @sfsholden - Lists the statistics for Apex completions 

lower case "completions"

Regarding the memory description, do we know the default value, or does it differ?

packages/salesforcedx-vscode-apex/package.json
brpowell: Yeah I was saying it might be a possible option. You might be able to get the configuration like you're doing so in the languageServer.ts file without this entry here but I'm not totally sure. As for whether it should be in the UI maybe @smaddox-sf should chime in.

packages/salesforcedx-vscode-apex/package.nls.json
sfsholden: @ekapner It looks like the value will differ for the java memory, as it is calculated at runtime and based on system configuration. This has some good info: https://stackoverflow.com/questions/4667483/how-is-the-default-max-java-heap-size-determined

ekapner: Thanks. I would edit the description to say (although not highlighted in green):
"...Apex Language Server in MB, or null to use the system default value."

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -268,6 +268,11 @@
           "type": "boolean",
           "default": false,
           "description": "%enable_sobject_refresh_on_startup_description%"
+        },
+        "salesforcedx-vscode-apex.advanced.enable-completion-statistics": {
```
lcampos, line 272: Bryan mentioned this already, if you don't want this to be surfaced in the UI just remove it completely from the `package.json`. That will make the extension rely on code to know how to handle the setting but we've done it before, take a look at `internal-development` in the core extension.

sfsholden:
> looks good just needs the new jar version which can be part of this or on a follow up PR

I'll open a new PR for the jar once everything has been merged into master for apex-jorje.

PR #2226, opened 2020-05-30
LWC Preview on Mobile

brandonpage:
### What does this PR do?
This PR adds the ability to preview a Lightning Web Component on an Android emulator or iOS simulator.  

The code here is largely a dumb wrapper to expose the functionality of Mobile Platform Enablement team's SFDX plugin.  The plugin has been approved to be open sourced, but until it is publicly available all features introduced in this PR are locked behind a (default off) feature flag.  VSCode extension features:
- Ability to run preview from command palette.
- Success and failure notifications for emulator and simulator.
- Remembers last used emulator/simulator (can be turned off in settings).
- Setting to change log level of SFDX Plugin.
- Prompts the user to install the SFDX Plugin if necessary.   
- Many tests added. 

### Functionality Before
Preview opens the LWC in the browser.

### Functionality After
By default, there is no change in functionality.  

If the mobile preview feature is enabled the user is prompted to select between Desktop, Android or iOS.  If Android or iOS are selected the user will be prompted to supply the name of the device they wish to use.  The SFDX Plugin will create and launch the device if necessary.  The plugin will also supply detailed environment setup feedback if something goes wrong.  

mpe_launch_ios

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -787,6 +787,11 @@
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
         }
+      },
+      "salesforcedx-vscode-core.previewOnMobile": {
+        "type": "boolean",
+        "default": false,
+        "description": "%force_lightning_lwc_enable_mobile%"
       }
     },
     "jsonValidation": [
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -786,12 +786,12 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
+        },
+        "salesforcedx-vscode-core.previewOnMobile": {
+          "type": "boolean",
+          "default": false,
+          "description": "%force_lightning_lwc_enable_mobile%"
         }
-      },
-      "salesforcedx-vscode-core.previewOnMobile": {
-        "type": "boolean",
-        "default": false,
-        "description": "%force_lightning_lwc_enable_mobile%"
       }
     },
     "jsonValidation": [
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -786,11 +786,6 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
-        },
-        "salesforcedx-vscode-core.previewOnMobile": {
-          "type": "boolean",
-          "default": false,
-          "description": "%force_lightning_lwc_enable_mobile%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
@@ -10,6 +10,7 @@ import {
   BETA_DEPLOY_RETRIEVE,
   CONFLICT_DETECTION_ENABLED,
   INTERNAL_DEVELOPMENT_FLAG,
+  LWC_MOBILE_PREVIEW_ENABLED,
   PUSH_OR_DEPLOY_ON_SAVE_ENABLED,
   RETRIEVE_TEST_CODE_COVERAGE,
   SFDX_CORE_CONFIGURATION_NAME,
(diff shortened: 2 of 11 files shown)


packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -223,3 +226,11 @@ function startLWCLanguageServer(context: ExtensionContext) {
   // client can be deactivated on extension deactivation
   context.subscriptions.push(client);
 }
+
+export function getGlobalStore(): vscode.Memento {
```
lcampos, line 230: Let's move this to a separate class/util and not part of the main index. Same for `getWorkspaceSettings`

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -195,7 +195,7 @@
         },
         {
           "command": "sfdx.force.lightning.lwc.preview",
-          "when": "false"
+          "when": "sfdx:project_opened && resource =~ /.*/lwc/[^/]+(/[^/]+\\.(html|css|js))?$/"
```
lcampos, line 198: Is the expectation that a user can trigger a preview from the Command Palette ? If so, this won't work since the resource will be empty and the evaluation will always be `false`

packages/salesforcedx-vscode-core/package.json
```diff
@@ -786,6 +786,11 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
+        },
```
lcampos, line 789: I think this should be moved to `salesforcedx-vscode-lwc/package.json` since all the local dev configs live there.

packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -74,6 +75,10 @@ export class SfdxCoreSettings {
     return this.getConfigValue(BETA_DEPLOY_RETRIEVE, false);
   }
 
+  public getLwcPreviewOnMobileEnabled(): boolean {
```
lcampos, line 78: Same as with the settings, this should be part of the lwc extension and not the core one.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -98,5 +167,143 @@ export async function forceLightningLwcPreview(sourceUri: vscode.Uri) {
...
+  execution.processExitSubject.subscribe(async exitCode => {
+    if (exitCode !== 0) {
+      const message =
+        platformSelection.id === PreviewPlatformType.Android
+          ? nls.localize('force_lightning_lwc_android_failure', targetUsed)
+          : nls.localize('force_lightning_lwc_ios_failure', targetUsed);
+      showError(new Error(message), logName, commandName);
+
+      // Error code 127 means the lwc on mobile sfdx plugin is not installed.
+      if (exitCode === 127) {
```
lcampos, line 271: It would be good to also push this message to the error message surfaced on the lower right corner similar to how we do it for the Diff plugin. You can look at that code here https://github.com/forcedotcom/salesforcedx-vscode/blob/0e506ab11331b4345d792562456be0daa8b61e25/packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts#L76

mobile-plugin-missing-error

diff-plugin-missing-error

packages/salesforcedx-vscode-lwc/package.json
brandonpage: I don't understand what you mean.  This line allows the preview command to show in the command palette if editor is currently focused on a valid file in an lwc project.  This causes source uri to be undefined when the command is called but that is why I get the uri from the document currently open in the editor windows in that scenario specifically.  Looking at other packages I thought this common practice.

lcampos: I see what you mean, I wasn't opening any component that why I didn't see it in the command palette. We might want to change the label in the future to reflect it only works when opening the file or add it to the docs somewhere.

packages/salesforcedx-vscode-lwc/src/util/workspaceUtils.ts
```diff
@@ -0,0 +1,32 @@
+import * as vscode from 'vscode';
```
lcampos, line 1:
```suggestion
/*
 * Copyright (c) 2020, salesforce.com, inc.
 * All rights reserved.
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 */
import * as vscode from 'vscode';
```

PR #2232, opened 2020-06-02
Smit/cancel command error

smitshah2531:
### What does this PR do?

Fixed "Cancelling command execution throws errors" 

@W-5757070@

### Functionality Before For "Cancelling Command Execution Throws Error"

Before

### Functionality After For "Cancelling Command Execution Throws Error"

After
 
### Functionality Before For "Parsing Error" when running  "SFDX: Deploy Source to Org"

Before_Parsing

### Functionality After  For "Parsing Error" when running  "SFDX: Deploy Source to Org"

After_Parsing

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
@@ -59,20 +59,22 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
     execution.processExitSubject.subscribe(async exitCode => {
       this.logMetric(execution.command.logName, startTime);
       try {
-        const deployParser = new ForceDeployResultParser(stdOut);
-        const errors = deployParser.getErrors();
-        if (errors && !deployParser.hasConflicts()) {
-          channelService.showChannelOutput();
-          handleDiagnosticErrors(
-            errors,
-            workspacePath,
-            execFilePathOrPaths,
-            BaseDeployExecutor.errorCollection
-          );
-        } else {
-          BaseDeployExecutor.errorCollection.clear();
+        if (stdOut !== '') {
+          const deployParser = new ForceDeployResultParser(stdOut);
+          const errors = deployParser.getErrors();
+          if (errors && !deployParser.hasConflicts()) {
+            channelService.showChannelOutput();
+            handleDiagnosticErrors(
+              errors,
+              workspacePath,
+              execFilePathOrPaths,
+              BaseDeployExecutor.errorCollection
+            );
+          } else {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/notifications/notificationService.ts
```diff
@@ -84,13 +84,12 @@ export class NotificationService {
     cancellationToken?: vscode.CancellationToken
   ) {
     observable.subscribe(async data => {
-      if (data !== undefined && data.toString() === '0') {
+      if (data !== undefined && data !== null && data.toString() === '0') {
```
xyc, line 87: To simply, you can just use `String(data)`

Reference: https://exploringjs.com/impatient-js/ch_strings.html#converting-to-string

packages/salesforcedx-vscode-core/src/notifications/notificationService.ts
```diff
@@ -12,6 +12,7 @@ import { DEFAULT_SFDX_CHANNEL } from '../channels';
 import { STATUS_BAR_MSG_TIMEOUT_MS } from '../constants';
 import { nls } from '../messages';
 import { sfdxCoreSettings } from '../settings';
+// import { stringify } from 'querystring';
```
AnanyaJha, line 15: let's remove this import if it's not being used

smitshah2531: Sounds good. I wasn't sure should I keep it or not but sure I will update it.

packages/salesforcedx-vscode-core/src/notifications/notificationService.ts
smitshah2531: sounds good. Thanks!!!

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -59,20 +59,22 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
-          channelService.showChannelOutput();
-          handleDiagnosticErrors(
-            errors,
-            workspacePath,
-            execFilePathOrPaths,
-            BaseDeployExecutor.errorCollection
-          );
-        } else {
-          BaseDeployExecutor.errorCollection.clear();
+        if (stdOut !== '') {
```
AnanyaJha, line 62: can we make this a positive check instead?
ie. `if (stdOut) { doSomething(); }`

smitshah2531: Yes, we can make it a positive check. Solved!!! Thanks!!!

PR #2236, opened 2020-06-04
Feature/#2140 package install

gabrielbarker:
### What does this PR do?
This PR adds a command to install a package into the default org.

### What issues does this PR fix or reference?
#2140 

### Functionality Before
This had to be done through the org or CLI

### Functionality After
There is a command, SFDX: Install Package, to the vscode command palette.

PR #2244, opened 2020-06-08
2140 package install

lcampos:
### What does this PR do?
Port PR #2236 to enable tests runs. It also adds package id validation to the command.

@W-7660006@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/validations.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/validations.ts
@@ -28,7 +28,7 @@ export function isAlphaNumString(value: string | undefined): boolean {
   return value !== undefined && value !== '' && !/\W/.test(value);
 }
 
-export function isRecordId(value: string = '', prefix: string): boolean {
+export function isRecordIdFormat(value: string = '', prefix: string): boolean {
   return (
     isAlphaNumString(value) &&
     value.startsWith(prefix) &&
--- a/packages/salesforcedx-utils-vscode/src/helpers/validations.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/validations.ts
@@ -28,6 +28,14 @@ export function isAlphaNumString(value: string | undefined): boolean {
   return value !== undefined && value !== '' && !/\W/.test(value);
 }
 
+export function isRecordId(value: string = '', prefix: string): boolean {
+  return (
+    isAlphaNumString(value) &&
+    value.startsWith(prefix) &&
+    (value.length === 15 || value.length === 18)
+  );
+}
+
 export function isAlphaNumSpaceString(value: string | undefined): boolean {
   return value !== undefined && /^\w+( *\w*)*$/.test(value);
 }
--- a/packages/salesforcedx-utils-vscode/test/unit/helpers/validations.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/helpers/validations.test.ts
@@ -10,7 +10,7 @@ import {
   isAlphaNumString,
   isInteger,
   isIntegerInRange,
-  isRecordId
+  isRecordIdFormat
 } from '../../../src/helpers/validations';
 
 describe('Input Box Validations', () => {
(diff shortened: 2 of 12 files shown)


packages/salesforcedx-utils-vscode/test/unit/helpers/validations.test.ts
```diff
@@ -140,4 +141,26 @@ describe('Input Box Validations', () => {
...
+      expect(res).to.equal(false);
+    });
+
+    it('Should return false if value has incorrect length', async () => {
+      const res = isRecordId('123xx0000000', '123');
+      expect(res).to.equal(false);
+    });
+
+    it('Should return true if value is a valid record id', async () => {
+      const res = isRecordId('123xx0000000123', '123');
```
brpowell, line 162: Maybe one for 18 too?

packages/salesforcedx-vscode-core/package.json
```diff
@@ -281,6 +281,10 @@
         }
       ],
       "commandPalette": [
+        {
+          "command": "sfdx.force.package.install",
+          "when": "sfdx:project_opened"
```
brpowell, line 286: Does it make sense to also check for the default username state?

packages/salesforcedx-vscode-core/src/commands/forcePackageInstall.ts
```diff
@@ -0,0 +1,156 @@
...
+    }).execute(cancellationToken);
+
+    notificationService.reportExecutionError(
+      execution.command.toString(),
+      (execution.stderrSubject as any) as Observable<Error | undefined>
+    );
+    channelService.streamCommandOutput(execution);
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
```
brpowell, line 82: Might be able to remove this and just rely on the `execute()` from the base class.

packages/salesforcedx-vscode-core/src/commands/forcePackageInstall.ts
```diff
@@ -0,0 +1,156 @@
...
+  installationKey: string;
+}
+
+export class SelectPackageID implements ParametersGatherer<PackageID> {
+  public async gather(): Promise<CancelResponse | ContinueResponse<PackageID>> {
+    const packageIdInputOptions = {
+      prompt: nls.localize('parameter_gatherer_enter_package_id'),
+      placeHolder: nls.localize('package_id_gatherer_placeholder'),
+      validateInput: value => {
+        return isRecordId(value, PKG_ID_PREFIX) || value === ''
```
brpowell, line 101: Will `value === ''` permit entering a blank package id?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forcePackageInstall.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    it('Should return cancel if package ID is undefined', async () => {
+      inputBoxSpy.onCall(0).returns(undefined);
+      const gatherer = new SelectPackageID();
+      const response = await gatherer.gather();
+      expect(inputBoxSpy.callCount).to.equal(1);
+      expect(response.type).to.equal(EVENT_CANCEL);
+    });
+
+    it('Should return cancel if packageId is empty string', async () => {
```
brpowell, line 42: Instead of cancelling maybe this should be caught by the validation

packages/salesforcedx-utils-vscode/src/helpers/validations.ts
```diff
@@ -28,6 +28,14 @@ export function isAlphaNumString(value: string | undefined): boolean {
   return value !== undefined && value !== '' && !/\W/.test(value);
 }
 
+export function isRecordId(value: string = '', prefix: string): boolean {
```
xyc, line 31: I'd rather name this `isRecordIdFormat` since a string satisfying these conditions is not necessarily a record id in the org

packages/salesforcedx-vscode-core/src/commands/forcePackageInstall.ts
lcampos: No, the command validation with catch that and not let the user continue

packages/salesforcedx-vscode-core/package.json
lcampos: I don't think we need that, we do those checks at the command level

packages/salesforcedx-vscode-core/src/commands/forcePackageInstall.ts
lcampos: True, let me do that

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forcePackageInstall.test.ts
lcampos: Yeah, I think I need to update this test now that I've added the command validation

lcampos: Nevermind, this test is still valid.

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -27,3 +27,4 @@ export const DEFAULT_USERNAME_KEY = 'defaultusername';
 export const DEFAULT_DEV_HUB_USERNAME_KEY = 'defaultdevhubusername';
 export const CONFLICT_DETECTION_ENABLED = 'detectConflictsAtSync';
 export const BETA_DEPLOY_RETRIEVE = 'experimental.deployRetrieve';
+export const PKG_ID_PREFIX = '04t';
```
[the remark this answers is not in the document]
lcampos: I think we'll use this in the future if we need to support more than one id in the extensions. For now we'll stick with the constant.

PR #2265, opened 2020-06-15
Remove LWC Preview on Mobile Feature Flag, add Documentation. 

brandonpage:
### What does this PR do?
Removes the feature flag needed to preview Lightning Web Components on a mobile device from within VSCode.  Updates public documentation to cover this functionality.  Removes tests for previous desktop only experience (which are already duplicated for new quick pick ui). 

### Functionality Before
Feature flag needed to be enabled to have the option of previewing lwc on mobile.

### Functionality After
Option of previewing on desktop browser, Android or iOS by default.  
new_default


khawkins:
@xyc @lcampos JFYI it turns out we will *not* be ready to release this in this week's cycle—we still have outstanding tasks for our release documentation, before we can go "public public" (beta) with everything.

So review away, but please don't merge in this week's cycle. 🙂 Thanks!

xyc:
@khawkins Thanks for letting us know. We are freezing the release next week for TDX so the earliest next release date is going to be Jul 2.

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -366,11 +366,6 @@
             "fatal"
           ],
           "default": "warn"
-        },
-        "salesforcedx-vscode-lwc.preview.enableMobilePreviews": {
-          "type": "boolean",
-          "default": false,
-          "description": "%force_lightning_lwc_enable_mobile%"
```
lcampos, line 373: We need to remove this label from `package.nls.json` too

khawkins:
Regarding the [comment above](https://github.com/forcedotcom/salesforcedx-vscode/pull/2265#issuecomment-644381204), it turns out we *are* good to go whenever—we don't need to hold off as originally expected.  Sorry for the confusion.

PR #2267, opened 2020-06-16
Use the change-case-management-orb

vazexqi:
### What does this PR do?

Streamlines the .circleci/config.yml to use our orb.

Part of internal safety checks for releasing.

### What needs to happen **before** this can be merged?

- [x] This flow only runs on publish. I need someone to work with me to test it out - I think we need to create some "fake" release records but I don't want to do it without consulting the team first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.2.0
-  sfchangecase: salesforce/change-case-management@2.0.1
+  sfchangecase: salesforce/change-case-management@3
 
 _refs:
   defaults: &defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.2.0
+  sfchangecase: salesforce/change-case-management@2.0.1
 
 _refs:
   defaults: &defaults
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -37,28 +38,11 @@ _refs:
...
-    command: |
-      if [ "$GUS_SKIP" == "false" ]; then
-        export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
-        echo "Gus build is ${GUS_BUILD}"
-        sfchangecase update --location ${REPO_LOCATION} --release offcore.tooling.${GUS_BUILD} --status 'Closed - Deploy Successful'
-      else
-        echo "Gus change case management was skipped because GUS_SKIP is true"
-      fi
+      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
+      echo "export GUS_RELEASE=offcore.tooling.$GUS_BUILD" >> $BASH_ENV
```
vazexqi, line 45: This was mentioned at https://circleci.com/docs/2.0/env-vars/#using-parameters-and-bash-environment as a way to share ENV VARIABLES between steps. This needs to be verified with a sample run.

.circleci/config.yml
```diff
@@ -37,28 +38,11 @@ _refs:
...
-        echo "Gus build is ${GUS_BUILD}"
-        sfchangecase create -i ${GUS_TEMPLATE_ID} --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
-        sfchangecase check --release offcore.tooling.${GUS_BUILD} --location ${REPO_LOCATION}
-      else
-        echo "Gus change case management was skipped because GUS_SKIP is true"
-      fi
-  gus-close-change-mgmt: &gus-close-change-mgmt
-    name: 'Gus close change case management'
-    command: |
-      if [ "$GUS_SKIP" == "false" ]; then
```
vazexqi, line 55: We can still skip. This is taken care of in the orb itself. See the [internal repo](https://github.com/forcedotcom/change-case-management-orb/blob/master/src/commands/check.yml#L16)

.circleci/config.yml
```diff
@@ -274,7 +265,9 @@ jobs:
             git tag v${GUS_BUILD}
             git push origin v${GUS_BUILD}
             git push origin master
-      - run: *gus-close-change-mgmt
+      - sfchangecase/update:
```
vazexqi, line 268: The default status is to close it. See the [internal repo](https://github.com/forcedotcom/change-case-management-orb/blob/master/src/commands/update.yml#L13)

vazexqi:
The codecov/project — 74.23% (+-0.02%) compared to e75898f is not accurate I believe since this is just a config change.

vazexqi:
Based on https://github.com/forcedotcom/source-deploy-retrieve/pull/61 this PR is also ready to be merged.

As a reminder,  to bypass this, set `SF_CHANGE_CASE_BYPASS` to true.

CC: @lcampos

PR #2272, opened 2020-06-17
As/tool tip feature

amandastern99:
Adds tooltip features for org picker and show org status bar items.


packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -20,6 +20,7 @@ let statusBarItem: StatusBarItem;
 export function showOrg() {
   if (!statusBarItem) {
     statusBarItem = window.createStatusBarItem(StatusBarAlignment.Left, 50);
+    statusBarItem.tooltip = 'Open Org';
```
lcampos, line 23: Lets move the label to `src/messages/i18n.ts` so it's available for translation

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -33,6 +33,7 @@ export class OrgList implements vscode.Disposable {
       49
     );
     this.statusBarItem.command = 'sfdx.force.set.default.org';
+    this.statusBarItem.tooltip = 'Change Default Org';
```
lcampos, line 36: Same as above, lets move it to the messages module and use `nls.localize(...)` here

PR #2279, opened 2020-06-18
Port changes from release v48.21.0 to develop

sfsholden:
### What does this PR do?
Ports changes that were part of the release, but not develop.


sfsholden:
Would we prefer to just do a normal merge of the release into develop like we used to?

PR #1884, opened 2020-01-17
10.2.0 has an issue with compiling node-gyp

KrisGraySFDC:
### What does this PR do?
Addresses a compile error during npm install.
This PR updates the following dependencies:
- Use of nodejs v12.4.0
- npm v6.9.0
- vscode versions 1.40 and above
- mocha 5
- sinon 7.3.1

### What issues does this PR fix or reference?
See this issue relating to 10.2.0
https://github.com/nodejs/node/issues/20921

@W-7138332@

PR #2286, opened 2020-06-23
Update @types/node

xyc:
### What does this PR do?
Update @types/node to 12.0.12

@W-7143011@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -29,7 +29,7 @@
     "@salesforce/salesforcedx-utils-vscode": "48.21.0",
     "@salesforce/source-deploy-retrieve": "1.0.10",
     "adm-zip": "0.4.13",
-    "applicationinsights": "1.0.6",
```
xyc, line 32: There's an [issue](https://github.com/microsoft/ApplicationInsights-node.js/issues/433) with applicationinsights version 1.0.6 and a newer `@types/node` version.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -43,7 +43,7 @@
     "@types/chai": "^4.0.0",
     "@types/glob": "^5.0.30",
     "@types/mocha": "^5",
-    "@types/node": "8.9.3",
+    "@types/node": "^13.11.0",
```
brpowell, line 46: Do we want to use a version 12 flavor of these types since VS Code stable and insiders is on 12.8.1 and 12.13.0 respectively?

xyc: > VS Code stable and insiders is on 12.8.1 and 12.13.0 respectively?
Where could I find this?

`^13.11.0` comes from the version of `@types/node` in a new projected scaffolded by yo https://code.visualstudio.com/api/get-started/your-first-extension

brpowell: Ah interesting. If you go to the "About" section in the menu bar this shows up:

Screen Shot 2020-06-23 at 3 52 24 PM

And on the master branch of the vscode repo they're using ^12.11.7

xyc: Yeah that's interesting. I think we should use 12 for now. Not sure about the specific version though.
Our node dependency is [on 12.4.0](https://github.com/forcedotcom/salesforcedx-vscode/pull/1884/files). We could use that or update both node dependency and `@types/node` to `12.8.1`.

@lcampos thoughts?

xyc:
@jag-sfdc sorry I made a typo in the title. This will be the first PR for that story - more incoming.

packages/salesforcedx-vscode-core/package.json
lcampos: @brpowell @xyc the node version decision comes from the Electron version being shipped by the min VSCode version (engine in package.json) that we decide to support. VSCode 1.43 is the one that started shipping with NodeJS version 12.8.1 (https://code.visualstudio.com/updates/v1_43#_electron-70-update). We should only merge these changes if we can merge a change to the extension's engine version to be 1.43 and above in the same week/PR.

If we can't do that then let's stick to @types/node being 12.0.12 (I don't see a version equal to 12.4.0).

xyc: Makes sense. Let's address engine update separately. Changed `@types/node` version to `12.0.12`.

PR #2291, opened 2020-06-25
Consume library version of execute anonymous apex command in VS Code

AnanyaJha:
### What does this PR do?
With this PR, we are now using the Apex library instead of the CLI to run the `force:apex:execute` command

 @W-7561179@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -25,9 +25,10 @@
     "Other"
   ],
   "dependencies": {
-    "@salesforce/core": "2.4.1",
+    "@salesforce/core": "2.5.1",
```
AnanyaJha, line 28: The Apex library is currently using version `2.5.1` of `@salesforce/core` and it seems like there were changes between versions `2.4.1` and `2.5.1` that are making the two versions incompatible w each other. If we update the dependency in VS Code to be `2.5.1`, we'll also have to update the version of the SDR library we're currently using to be the latest (currently unreleased) version since that has the correct version of `@salesforce/core`.

AnanyaJha: Or, we can move the version of `@salesforce/core` used in the Apex library to `2.4.1` until the rest of these changes can made. thoughts?

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -4,115 +4,23 @@
...
-    const conn = await OrgAuthInfo.getConnection(usernameOrAlias);
-    this.sourceClient = new SourceClient(conn);
-  }
-
-  public retrieveWrapper(fn: (...args: any[]) => Promise<ApiResult>) {
-    const commandName = this.executionName;
-
-    return async function(...args: any[]): Promise<ApiResult> {
-      channelService.showCommandWithTimestamp(`Starting ${commandName}`);
+  public build(execName: string, telemetryLogName: string) {}
```
AnanyaJha, line 21: needs a return type

packages/salesforcedx-vscode-core/package.json
```diff
@@ -25,9 +25,10 @@
...
     "@salesforce/salesforcedx-utils-vscode": "48.21.0",
-    "@salesforce/source-deploy-retrieve": "1.0.10",
+    "@salesforce/source-deploy-retrieve": "1.0.13",
+    "@salesforce/salesforcedx-apex": "github:forcedotcom/salesforcedx-apex#develop",
```
AnanyaJha, line 31: Will be updated once the library is released

packages/salesforcedx-vscode-core/package.json
xyc: If salesforcedx-apex is not released yet I'd recommend changing the version to `2.4.1` so that it doesn't block this PR, unless we release new version of SDR library sooner

lcampos: Keep version 2.5.1, we'll be releasing a new version of deploy retrieve tomorrow, we'll just have to hold off on merging this PR so it doesn't break the build

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,158 @@
...
+
+export class ApexLibraryExecuteExecutor extends ApexLibraryExecutor {
+  public async execute(
+    response: ContinueResponse<{ fileName: string }>
+  ): Promise<void> {
+    this.setStartTime();
+
+    try {
+      await this.build(
+        'Apex Execute (via library)',
```
lcampos, line 69: I would just keep this message as `Apex Execute`

packages/salesforcedx-vscode-core/src/commands/util/apexCommandlet.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ApexService } from '@salesforce/salesforcedx-apex/packages/apex/lib';
```
lcampos, line 7: We should be able to import library stuff by just doing `@salesforce/salesforcedx-apex`

packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
+import { ExecuteAnonymousResponse } from '@salesforce/salesforcedx-apex/packages/apex/lib';
```
lcampos, line 7: Same as above, the import need to be `@salesforce/apex`

packages/salesforcedx-vscode-core/src/commands/util/apexCommandlet.ts
AnanyaJha: Once we publish the library package, we'll be able to import the library in files by just doing `@salesforce/apex`. 
But if I've understood correctly, since we're using the entire github repository, not just the library package, as the "package" in our package.json right now, we need to reference the top level `salesforcedx-apex` repository in our imports.
I haven't been able to import the library package directly using the github url, but let me know if there's something different i should try

AnanyaJha: Or did you mean we should change the name of the repo itself to `apex`?

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,158 @@
...
-      let writeFile;
       const document = editor.document;
-
-      if (editor.selection.isEmpty) {
-        writeFile = await writeFileAsync(fileName, document.getText());
-      } else {
-        writeFile = await writeFileAsync(
-          fileName,
-          document.getText(editor.selection)
+      let fileName = document.uri.fsPath;
```
lcampos, line 41: Why do we need to store this as a file first ? What if we update the library to accept a stream or a buffer ?

packages/salesforcedx-vscode-core/src/commands/util/apexCommandlet.ts
lcampos: I'm not sure if the github url approach supports a mono repo structure, but you are right in that once it's published we can change to `@salesforce/apex`. However, I want to make sure that we don't have to do `@salesforce/apex/lib` once we publish.

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
AnanyaJha: Yeah we could definitely update the library to accept a buffer or stream as well. I think it would be worth it to address this in a separate PR, once we've added the functionality into the library, thoughts?

packages/salesforcedx-vscode-core/src/commands/util/apexCommandlet.ts
AnanyaJha: Ah okay I see what you mean, I think we'll need to specify the `main` and `files` fields in the library's package.json so that we don't need to specify the `lib` folder. I'll open a PR in the library to address that

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
brpowell: The update in the library is simple, I think it would be better to do it correctly now than add debt.

packages/salesforcedx-vscode-core/src/commands/util/apexCommandlet.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { languages } from 'vscode';
+import { LibraryCommandletExecutor } from './libraryCommandlet';
+
+export class ApexLibraryExecutor extends LibraryCommandletExecutor<{
```
AnanyaJha, line 10: This portion will be refactored later on based on updates in the Apex library

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,175 @@
...
-type TempFile = {
-  fileName: string;
-};
+const workspaceChecker = new SfdxWorkspaceChecker();
+const fileNameGatherer = new AnonApexGatherer();
+
+export class ApexLibraryExecuteExecutor extends ApexLibraryExecutor {
+  protected executeService: ExecuteService | undefined;
+
+  public async build(
```
lcampos, line 60: This method is generic enough that it should be moved to `ApexLibraryExecutor`

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,175 @@
...
+        ...(response.data.apexCode && { apexCode: response.data.apexCode })
+      });
+    } catch (e) {
+      telemetryService.sendException('force_apex_execute_library', e.message);
+      notificationService.showFailedExecution(this.executionName);
+      channelService.appendLine(e.message);
+    }
+  }
+
+  public executeWrapper(
```
lcampos, line 105: This should be moved to `ApexLibraryExecutor`

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,175 @@
...
-  });
+
+      return result;
+    };
+  }
 }
 
-const workspaceChecker = new SfdxWorkspaceChecker();
-const fileNameGatherer = new CreateApexTempFile();
+export function formatResult(
```
lcampos, line 149: We should have a formatter only class/util and move this there.

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,175 @@
...
+  return outputText;
+}
 
-export async function forceApexExecute(withSelection?: any) {
+export async function forceApexExecute() {
   const commandlet = new SfdxCommandlet(
     workspaceChecker,
     fileNameGatherer,
-    new ForceApexExecuteExecutor()
+    new ApexLibraryExecuteExecutor()
```
lcampos, line 175: This might not be the right place for it but we should have a setting that would allow users to use the old CLI implementation, at least while we roll out the library changes.

packages/salesforcedx-vscode-core/src/commands/util/deployRetrieveCommandlet.ts
```diff
@@ -0,0 +1,109 @@
...
+import { channelService } from '../../channels';
+import { handleDeployRetrieveLibraryDiagnostics } from '../../diagnostics';
+import { nls } from '../../messages';
+import { notificationService } from '../../notifications';
+import { OrgAuthInfo } from '../../util';
+import { LibraryCommandletExecutor } from './libraryCommandlet';
+import { LibraryDeployResultParser } from './libraryDeployResultParser';
+import { outputRetrieveTable } from './retrieveParser';
+
+export class DeployRetrieveLibraryExecutor extends LibraryCommandletExecutor<
```
lcampos, line 23: If this only hosts `DeployRetrieveLibraryExecutor` then let's rename the file to `deployRetrieveLibraryExecutor`

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -303,7 +302,10 @@ export const messages = {
   error_no_scratch_def:
     'No scratch definition files found. These files must be in the "config" folder and have file names that end with "-scratch-def.json". See [Scratch Org Definition File](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm) for help.',
   force_list_metadata: 'SFDX: Force List Metadata',
-
+  apex_execute_compile_success: 'Compiled successfully.',
```
lcampos, line 305: Are these labels the same used in the CLI command ? cc @ekapner

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
AnanyaJha: Yeah I'll open up a follow up PR for that

packages/salesforcedx-vscode-core/src/messages/i18n.ts
AnanyaJha: Yes the `apex_execute_compile_success` and `apex_execute_runtime_success` are the same as the CLI command, but would be great if you could take a look @ekapner !
`apex_execute_text` is used for telemetry and `force_apex_execute_library` is the label we'll display to the user when running the command.

ekapner: @lcampos @AnanyaJha - a bit concerned about referencing the SFDX Dev Guide in the error message (mostly because the doc team is thinking about rearchitecting it - and I worry that this link will break - realistically we're talking 6-12 months out) - but could happen. We can take the approach to fix it once it's reported.

ekapner: @AnanyaJha - I need more context - I'll set up a meeting

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -303,7 +302,10 @@ export const messages = {
...
+  apex_execute_runtime_success: 'Executed successully.',
```
ekapner, line 306: typo: successfully

packages/salesforcedx-vscode-core/src/messages/i18n.ts
AnanyaJha: @ekapner are you referring to the `error_no_scratch_def` message? That message is actually not part of this change, it's already in production. But maybe we can create a separate work item if that link will be updated in the future

lcampos: @ekapner that's good to know. We can keep up with the status of that effort and update our links once they have a release date. From a user perspective it's valuable to have a link that takes you directly to where you need to consume the info from instead of having them search for it, huge time saver and a modern developer experience.

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -4,99 +4,96 @@
...
-type TempFile = {
-  fileName: string;
-};
+const workspaceChecker = new SfdxWorkspaceChecker();
+const fileNameGatherer = new AnonApexGatherer();
+
+export class ApexLibraryExecuteExecutor extends ApexLibraryExecutor {
+  protected executeService: ExecuteService | undefined;
+
+  public createService(conn: Connection): void {
```
lcampos, line 57: This part highlights a current limitation with the library in that we need to individually initialize every service in the library before using it. We have flows in which we'll need to run more than one service from the library sequentially (e.g. the new Replay Debugger) and we should only have to initialize (pass a connection object) once and be able to use as many services as needed.

We don't need to solve it for this PR but we should have a work item to track those changes.

AnanyaJha: Agreed, we should have a single entry point for the consumer to access any methods in the library. Here's the work item to track that: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008HWFgIAO/view

PR #2297, opened 2020-06-26
Update telemetry to accept properties and measurements separately

brpowell:
### What does this PR do?

Updates telemetry service to be able to send properties and measurements separately for sendCommandEvent.

@W-7751287@


packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -21,6 +21,7 @@ import { channelService } from '../../channels';
 import { notificationService, ProgressNotification } from '../../notifications';
 import { taskViewService } from '../../statuses';
 import { TelemetryData, telemetryService } from '../../telemetry';
+import { Measurements, Properties } from '../../telemetry/telemetry';
```
xyc, line 24: Do we want to expose Measurements/Properties in `index.ts` of the telemetry module?

lcampos:
@brpowell we should also update the `telemetry.ts` methods in lightning and lwc packages.

PR #2299, opened 2020-06-29
Telemetry for metadata type/component count for beta deploy/retrieve @W-7759892@

brpowell:
### What does this PR do?

Adds telemetry for the number and types of components being deployed and retrieved.

@W-7759892@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -25,9 +25,9 @@
     "Other"
   ],
   "dependencies": {
-    "@salesforce/core": "2.4.1",
+    "@salesforce/core": "2.5.1",
     "@salesforce/salesforcedx-utils-vscode": "48.23.0",
-    "@salesforce/source-deploy-retrieve": "1.0.10",
+    "@salesforce/source-deploy-retrieve": "1.0.14",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
@@ -33,8 +33,11 @@ import {
   VISUALFORCE_PAGE_EXTENSION
 } from './templates/metadataTypeConstants';
 import { FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
+import {
+  createComponentCount,
+  useBetaDeployRetrieve
+} from './util/betaDeployRetrieve';
 import { LibraryCommandletExecutor } from './util/libraryCommandlet';
-import { useBetaDeployRetrieve } from './util/useBetaDeployRetrieve';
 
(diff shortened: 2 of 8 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "@salesforce/core": "2.4.1",
```
lcampos, line 28: You need to update to `@salesforce/core` version 2.5.1 here and on the rest of the modules that use it

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
```diff
@@ -40,3 +40,16 @@ export function useBetaDeployRetrieve(explorerPath: vscode.Uri[]): boolean {
     typeName === apextrigger.name;
   return betaDeployRetrieve && supportedType;
 }
+
+export function createComponentCount(components: MetadataComponent[]) {
```
xyc, line 44: Can we add a unit test for this method?

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -48,6 +53,7 @@ export abstract class LibraryCommandletExecutor<T>
       throw new Error(nls.localize('error_no_default_username'));
     }
     const conn = await OrgAuthInfo.getConnection(usernameOrAlias);
+    // @ts-ignore private logger mismatch
```
lcampos, line 56: Isn't the private logger problem throwing issues at runtime ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -123,14 +126,17 @@ export class LibraryDeploySourcePathExecutor extends LibraryCommandletExecutor<
...
+        this.sourceClient.tooling.deploy
       );
 
-      await this.sourceClient.tooling.deployWithPaths({
-        paths: [response.data]
-      });
-      this.logMetric();
+      const registryAccess = new RegistryAccess();
+      const components = registryAccess.getComponentsFromPath(response.data);
+      const deployPromise = this.sourceClient.tooling.deploy({ components });
```
lcampos, line 135: What are we gaining by doing this instead of `await this.sourceClient.tooling.deploy(...` ?

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
brpowell: Yup

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
brpowell: It doesn't appear to be

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
brpowell: I suppose it's a minuscule benefit but it's to kick off the request before building the telemetry data.

PR #2309, opened 2020-07-01
CircleCi Config Improvements @W-7721208@

sfsholden:
#### Refactorings
1. Changes YAML anchors into commands. Eventually we will want to move some of these commands into an orb for easy reuse across repos.
1. Adds a new executor to replace the "defaults" YAML anchor.
#### New Features
1. Adds Slack integration for visibility into job status, errors, and successes for publish and release branch creation workflows.
1. Changes the release strategy from publishing when a merge happens into master to publishing via a pipeline strategy using parameters and API calls.
1. On publish and release branch creation, adds a "hold" job that requires approval in CircleCi. This helps prevent any accidental or malicious API calls.
1. Adds tasks for easy API calls in expected formats.
1. Adds a workflow for creating the release branch via an API call. This will allow us to easily run through the same circleci workflow outside of the scheduled cron job.

@W-7721208@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,10 +200,9 @@ executors:
 jobs:
   build-all:
     description: 'Build artifacts in a non-Windows environment'
-    parameters:
-      e:
-        type: executor
-    executor: << parameters.e >>
+    executor:
+      name: default-executor
+      working-dir: ~/project
     environment:
       - SFDX_URL_LINUX: https://developer.salesforce.com/media/salesforce-cli/sfdx-linux-amd64.tar.xz
       - SFDX_AUTOUPDATE_DISABLE: true
@@ -377,8 +376,7 @@ workflows:
   # run a build and a windows test on every commit
   commit-workflow:
     jobs:
-      - build-all:
-          e: default-executor
+      - build-all
       - unit-tests
       - integration-tests
       - publish:
@@ -403,8 +401,7 @@ workflows:
                 - develop
                 - master
     jobs:
-      - build-all:
-          e: default-executor
+      - build-all
       - unit-tests
       - integration-tests
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,12 +78,17 @@ commands:
             echo "export GUS_RELEASE=<< parameters.release-tag-name >>.$GUS_BUILD" >> $BASH_ENV
 
   install:
-    description: 'Installing node'
+    description: 'Installing dependencies'
     steps:
-      - node/install:
-          node-version: << pipeline.parameters.node-version >>
+      - run: npm install
       - run: npm --version
       - run: node --version
+
+  package-extensions:
+    steps:
+      - run:
+          name: 'Packaging extensions'
+          command: npm run vscode:package
 
   restore_dependency_cache:
     # This logic cannot be done within the node orb currently. This is due to the fact that we use
@@ -119,6 +124,20 @@ commands:
           fingerprints:
             - << pipeline.parameters.fingerprint >>
 
+  stage-and-store-artifacts:
+    parameters:
+      location:
+        type: string
+        default: 'extensions'
+    steps:
+      - run:
+          name: 'Staging artifacts'
+          command: |
+            mkdir << parameters.location >>
+            find packages -name "*.vsix" -type f -exec cp {} ./<< parameters.location >> \;
+      - store_artifacts:
+          path: ./<< parameters.location >>
+
   build-artifacts:
     description: 'Building and archiving extension artifacts.'
     steps:
@@ -126,16 +145,8 @@ commands:
       - restore_dependency_cache
       - install
       - compile
-      - run:
-          name: 'Packaging extensions'
-          command: npm run vscode:package
-      - run:
-          name: 'Staging artifacts'
-          command: |
-            mkdir extensions
-            find packages -name "*.vsix" -type f -exec cp {} ./extensions \;
-      - store_artifacts:
-          path: ./extensions
+      - package-extensions
+      - stage-and-store-artifacts
       - save_dependency_cache
 
   build-win:
@@ -310,9 +321,7 @@ jobs:
       - gh-config
       - install
       - compile
-      - run:
-          name: 'Packaging extensions'
-          command: npm run vscode:package
+      - package-extensions
       - run:
           name: 'Generating the SHA256 and appending to the file.'
           command: npm run vscode:sha256
@@ -330,13 +339,7 @@ jobs:
       - run:
           name: 'Publishing to the VS Code Marketplace.'
           command: npm run vscode:publish
-      - run:
-          name: 'Store artifacts that were published'
-          command: |
-            mkdir extensions
-            find packages -name "*.vsix" -type f -exec cp {} ./extensions \;
-      - store_artifacts:
-          path: ./extensions
+      - stage-and-store-artifacts
       - run:
           name: 'Create git tag (e.g. v48.1.0) & push changes back to master branch'
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,7 @@ jobs:
       - slack/notify:
           channel: 'pdt_releases'
           color: '#FFDEAD'
-          message: 'TESTING - Starting publish for VSCode Extensions'
+          message: 'Starting publish for VSCode Extensions'
       - ssh-config
       - checkout
       - gus-prepare-environment-variables
@@ -376,7 +376,7 @@ jobs:
       - slack/notify:
           channel: 'pdt_releases'
           color: '#9bcd9b'
-          message: 'TESTING - VSCode Extensions v${RELEASE_VERSION} have been published to the marketplace'
+          message: 'VSCode Extensions v${RELEASE_VERSION} have been published to the marketplace'
       - slack/status:
           channel: 'pdt_releases'
           fail_only: true
@@ -390,7 +390,7 @@ jobs:
       - slack/notify:
           channel: 'pdt_releases'
           color: '#FFDEAD'
-          message: 'TESTING - Starting creation of release branch for VSCode Extensions'
+          message: 'Starting creation of release branch for VSCode Extensions'
       - ssh-config
       - checkout
       - restore_dependency_cache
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -12,76 +12,183 @@ version: 2.1
 orbs:
   win: circleci/windows@2.2.0
   sfchangecase: salesforce/change-case-management@3
+  slack: circleci/slack@3.4.2
 
-_refs:
```
sfsholden, line 16: Moved YAML anchors into commands as per circleci recommendations. These have been made to be flexible depending on how we need to use them. Some of the new commands will be candidates for a new salesforce orb that we and other teams can call in the future.

.circleci/config.yml
```diff
@@ -12,76 +12,183 @@ version: 2.1
...
-    command: |
-      git config credential.helper 'cache --timeout=120'
-      git config user.email "$GH_EMAIL"
-      git config user.name "Release Bot"
-  gus-prepare-environment-variables: &gus-prepare-environment-variables
-    name: 'Prepare environment variables for GUS Change Case Management'
-    command: |
-      export GUS_BUILD="$(node -pe "require('./packages/salesforcedx-vscode/package.json').version")"
-      echo "export GUS_RELEASE=offcore.tooling.$GUS_BUILD" >> $BASH_ENV
+parameters:
```
sfsholden, line 17: These are circleci pipeline parameters. What makes this special is that we can change how the workflow behaves depending on the parameters that we pass in to the circleci API call. What this means is that we can easily run with a different version of node, the cli, trigger a creation of the release, and publish with more flexibility and control through circle ci.

.circleci/config.yml
```diff
@@ -12,76 +12,183 @@ version: 2.1
...
           name: 'Staging artifacts'
           command: |
-            mkdir extensions
-            find packages -name "*.vsix" -type f -exec cp {} ./extensions \;
+            mkdir << parameters.location >>
+            find packages -name "*.vsix" -type f -exec cp {} ./<< parameters.location >> \;
       - store_artifacts:
-          path: ./extensions
-      - save_cache:
-          key: dependency-cache-v2-{{ checksum "package.json" }}
```
sfsholden, line 71: Just a pointer - caching should be better now for the build-all step. We weren't using the same key when restoring the cache as we were for saving in the build-artifacts command. So we were always rebuilding the cache for every build-all job.

.circleci/config.yml
```diff
@@ -95,33 +202,46 @@ commands:
...
       - run:
           name: 'Run lint'
           shell: bash.exe
           command: |
             npm run lint
 
+executors:
+  default-executor:
+    parameters:
+      # We are unable to use the recommended cimg docker images since they use the root path instead of home
```
sfsholden, line 223: ### Recommended Change from CircleCi - update to cimg docker image

### Conclusion: Not incorporating recommendation
I do think that there is a way that we could still upgrade this image but I'm not convinced it's worth pursuing. The motivation towards changing the docker image is that it should be faster according to [here](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/). However from doing testing the jobs that it was running all took longer.

### Additional Info from Testing
For using the cimg docker images - [here](https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/1440/workflows/d27b1179-5686-4639-8cec-33cebd0e3b1d/jobs/10373) is an example run showing the path that the image uses. This causes compilation problems because of the path. Changing the working directory doesn't make any difference.

.circleci/config.yml
```diff
@@ -12,76 +12,183 @@ version: 2.1
 orbs:
   win: circleci/windows@2.2.0
```
sfsholden, line 13: ### Recommended Change from CircleCi
Incorporate node orb

### Conclusion: Not incorporating at this time
The node orb is reliant on package versions being generated using the package-lock.json. Since we rely on the package.json we won't be incorporating this for now. However, we can revisit at a later date if we change to package-locks.

.circleci/config.yml
```diff
@@ -95,33 +202,46 @@ commands:
...
 jobs:
   build-all:
     description: 'Build artifacts in a non-Windows environment'
-    docker:
-      - image: circleci/node:12.4.0
+    executor:
+      name: default-executor
+      working-dir: ~/project
     environment:
       - SFDX_URL_LINUX: https://developer.salesforce.com/media/salesforce-cli/sfdx-linux-amd64.tar.xz
```
sfsholden, line 241: I opted to keep a lot of our environment variables as is instead of changing them to pipeline parameters. If anyone has any suggestions on ones that would be good to easily control via the pipeline please feel free to make them.

.circleci/config.yml
```diff
@@ -209,104 +329,150 @@ jobs:
...
+            - build-all
+            - unit-tests
+            - integration-tests
+            - hold
           filters:
             branches:
               only:
                 - master
+
+  create-release-branch-workflow:
```
sfsholden, line 453: Allows us to create a release branch easily with circleci outside of the scheduled cron job.

scripts/create-release-branch-workflow.sh
```diff
@@ -0,0 +1,7 @@
+CircleCIToken=$1
+curl -v -u ${CircleCIToken}: -X POST --header "Content-Type: application/json" -d '{
+  "branch": "develop",
+  "parameters": {
+    "create-release": true
```
sfsholden, line 5: These are the pipeline parameters passed in via the API call to CircleCi. We can easily add a parameter here to change how circleci handles various workflows, jobs, and commands.

.circleci/config.yml
```diff
@@ -130,8 +250,8 @@ jobs:
     executor: win/default
```
sfsholden, line 250: ### Recommendation from CircleCi
Use parallelism for faster test runs

### Conclusion
Skipping for now. It doesn't seem like there is a lot of benefit for time investment and circleci resources. Right now on average our unit-tests take about ~9-10 minutes. Setting up the VM to run the unit-tests on takes a little over 7 minutes. When enabling parallelism this time for setup actually increases (since we are using additional VMs for each executor that run simultaneously). As we continue to add more tests, this might be something that we want to revisit in the future.

.circleci/config.yml
```diff
@@ -12,76 +12,183 @@ version: 2.1
...
+          keys:
+            - << parameters.base-cache-path >>{{ checksum "package.json" }}
+            # if cache for exact version of 'package.json' is not present, then load the most recent cache
+            - << parameters.base-cache-path >>
+
+  save_dependency_cache:
+    parameters:
+      base-cache-path:
+        type: string
+        default: 'dependency-cache-v2-'
```
lcampos, line 127: Let's update the default to `dependency-cache-v3-`, since we updated this last week because we wanted to flush the cache. Which going forward will be easier now that it's a param.

.circleci/config.yml
lcampos: Good catch 👍

.circleci/config.yml
```diff
@@ -209,104 +329,150 @@ jobs:
...
       - checkout
-      - restore_cache: *restore_cache
+      - gus-prepare-environment-variables
+      - restore_dependency_cache
       - run:
           name: 'Setup gus case management'
           command: |
             sudo npm install -g @salesforce/change-case-management
-      - run: *gus-prepare-environment-variables
+      - sfchangecase/install
```
lcampos, line 351: We need to remove this until we merge a fix to the case management orb. We currently replaced this line with "manually" installing `@salesforce/change-case-management` above.

.circleci/config.yml
```diff
@@ -209,104 +329,150 @@ jobs:
...
+    when: << pipeline.parameters.create-release >>
+    jobs:
+      - slack/approval-notification:
+          message: Pending Approval for Creation of Release Branch
+          channel: 'pdt_releases'
+          color: '#0E1111'
+      - build-all
+      - unit-tests
+      - integration-tests
+      - hold: # Requires manual approval in Circle Ci
```
vazexqi, line 463: @sfsholden - Good job on this. Luis asked me to take a look and things look 👍 

Quick question on the approval.

Can I approve my own workflow? Or must it be a different individual?

scripts/create-release-branch-workflow.sh
```diff
@@ -0,0 +1,7 @@
+CircleCIToken=$1
```
lcampos, line 1: Do we need to add `#!/usr/bin/env bash` to the bash scripts ?

scripts/publish-workflow.sh
```diff
@@ -0,0 +1,7 @@
+CircleCIToken=$1
```
vazexqi, line 1: @sfsholden - Any reason for not creating a script/approve-workflow as well? You already have the slack notification so you have the workflow id?

I _think_ you could also do https://circleci.com/docs/api/v2/#approve-a-job?

.circleci/config.yml
vazexqi: While not necessarily faster, I think they might be better maintained moving forward. So might be something to consider moving forward.

sfsholden: @vazexqi Agreed. From their documentation it looks like node images are in the process of being deprecated. We're going to have to play around with the image more before we can use it do to those path issues. There was probably something small that I was missing

.circleci/config.yml
sfsholden: @vazexqi 
Thanks for taking a look!

Anyone with push access to the repo by default will be able to approve the workflow. [Here](https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval) is more info on what goes into the approval process.

To restrict the approval process further you can add [contexts](https://circleci.com/docs/2.0/contexts/#approving-jobs-that-use-restricted-contexts).

scripts/create-release-branch-workflow.sh
sfsholden: VSCode is still able to launch it via a task because it's specified as a shell script, but I will go ahead and it incase we need to launch it manually. Good catch

scripts/publish-workflow.sh
sfsholden: @vazexqi I personally think it might be better to have an official "approve" process in circleci for security reasons. This would help prevent instances where someone may have accidentally launched the script without intending to publish. It also leaves a history of that individual who approved the workflow.

I'll send you a copy of a demo using my forked branch. Feel free to let me know what you think of the flow

.circleci/config.yml
```diff
@@ -37,6 +37,9 @@ parameters:
   create-release:
     type: boolean
     default: false
+  cache-version:
+    type: string
+    default: 'dependency-cache-v3-'
```
sfsholden, line 42: Decided to change this into a pipeline parameter so we can clear the cache with an api call as needed.

PR #2313, opened 2020-07-02
Lock Snyk version 1.358.0

brpowell:
### What does this PR do?

Lock Snyk version 1.358.0 since the previous version introduced a regression that causes the extensions to not compile in development mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "prettier": "1.16.4",
     "remap-istanbul": "^0.9.5",
     "shx": "0.2.2",
-    "snyk": "^1.82.1",
+    "snyk": "1.358.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "prettier": "1.16.4",
     "remap-istanbul": "^0.9.5",
     "shx": "0.2.2",
-    "snyk": "^1.82.1",
+    "snyk": "1.358.0",
```
xyc, line 15: Wonder why CI hasn't caught it

brpowell: Probably 2 things depending on the job - circleci dependency cache, and installing non dev dependencies in the build-all phase.

PR #2318, opened 2020-07-02
Pin @salesforce/kit

brpowell:
### What does this PR do?

Pin @salesforce/kit since newer version causing compile issues until we update typescript.


lcampos:
I think you also need to add it to `salesforcedx-apex`

Issue #2295, opened 2020-06-26
Unable to Create Project in VS Code due to 'Salesforce CLI Integration' extension taking too long and preventing others extensions from opening


adaslal:
### Summary

Unable to Create Project in VS Code due to 'Salesforce CLI Integration' extension taking too long and preventing others extensions from opening.

I have installed SFDX CLI and VS Code with the 'Salesforce Extension Pack'. Even then,I am not able to create a Salesforce Project in VS code

### Steps To Reproduce:

1.  Open VS Code
2.  Press Ctrl + Shift + P.
3.  Select 'Create SFDX Project'

### Expected result

You should be prompted to Select between types of Project(Eg: 'Standard')

### Actual result

I see the notification 'Activating Extensions' at the bottom left and nothing happens afterwards. When I checked the Issue I saw the following which says

'Did you attach the CPU Profile'

'This is a reminder to make sure that you have not forgotten to attach 'C:\Users\myName\salesforce.salesforcedx-vscode-core-unresponsive.cpuprofile.txt' to an existing performance issue.
![image](https://user-images.githubusercontent.com/57103745/85872379-45899c80-b7ed-11ea-8dd6-6d8df85af0ec.png)

### Additional information

_Feel free to attach a screenshot_.
![image](https://user-images.githubusercontent.com/57103745/85872866-fabc5480-b7ed-11ea-9460-bdc2df0989be.png)

**VS Code Version**: 
Version: 1.46.1 (user setup)
Commit: cd9ea6488829f560dc949a8b2fb789f3cdc05f5d
Date: 2020-06-17T21:13:20.174Z
Electron: 7.3.1
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

**SFDX CLI Version**: sfdx-cli/7.63.0-c897e8bb1a win32-x64 node-v10.20.1

**OS and version**:  OS - Windows 10 Enterprise. Version - 1909

smaddox-sf:
Hi @adaslal - Thanks for reaching out.  It sounds like the CLI setup might not be quite right.  That is usually the case when you see the 'activating' message but nothing happens.  Can you verify if the CLI configuration is [setup as described here](https://developer.salesforce.com/tools/vscode/en/troubleshooting/#set-salesforce-cli-path-windows)?

adaslal:
[...]

PR #2324, opened 2020-07-06
Core activation perf changes

lcampos:
### What does this PR do?
Removes the CLI checks done during core activation, adds ignoring `node_modules` in globs used for searching files and updates telemetry checks to use `@salesforce/core` libraries

#2295

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/forceConfigGet.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/forceConfigGet.ts
@@ -8,7 +8,10 @@
 import { SfdxCommandBuilder } from './commandBuilder';
 import { CliCommandExecutor } from './commandExecutor';
 import { CommandOutput } from './commandOutput';
-
+/**
+ * @deprecated
+ * NOTE: This code is deprecated in favor of using ConfigUtil.ts
+ */
 export class ForceConfigGet {
   public async getConfig(
     projectPath: string,
--- a/packages/salesforcedx-vscode-core/src/channels/channelService.ts
+++ b/packages/salesforcedx-vscode-core/src/channels/channelService.ts
@@ -64,13 +64,13 @@ export class ChannelService {
 
       this.channel.append(' ');
       if (data !== undefined) {
-        this.channel.appendLine(
-          nls.localize('channel_end_with_error', data.message)
-        );
-
         if (/sfdx.*ENOENT/.test(data.message)) {
           this.channel.appendLine(
             nls.localize('channel_end_with_sfdx_not_found')
+          );
+        } else {
+          this.channel.appendLine(
+            nls.localize('channel_end_with_error', data.message)
           );
         }
       } else {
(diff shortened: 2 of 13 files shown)


packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -64,14 +64,14 @@ export class ChannelService {
...
       if (data !== undefined) {
-        this.channel.appendLine(
-          nls.localize('channel_end_with_error', data.message)
-        );
-
         if (/sfdx.*ENOENT/.test(data.message)) {
           this.channel.appendLine(
             nls.localize('channel_end_with_sfdx_not_found')
           );
+        } else {
```
lcampos, line 71: Since we're removing the CLI installed checks during startup I'm also modifying this so that is we receive a `ENOENT` message from attempting to run a CLI command we only print the message on how to install/configure it instead of also printing the error message from the CLI.

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -50,21 +48,13 @@ export function disableCLITelemetry() {
...
-        projectPath,
-        SFDX_CONFIG_DISABLE_TELEMETRY
-      );
-      const disabledConfig =
-        forceConfig.get(SFDX_CONFIG_DISABLE_TELEMETRY) || '';
-      return disabledConfig !== 'true';
-    } catch (e) {
-      console.log('Error checking cli settings: ' + e);
-    }
+export async function isCLITelemetryAllowed(): Promise<boolean> {
```
lcampos, line 51: Updating this method to use `ConfigAggregator` which is much faster than running `force:config:get` command.

rcoringrato-sfdc: @lcampos Does it provide the same functionality?

lcampos: @rcoringrato-sfdc yes, it behaves the same way as the previous version of the code

packages/salesforcedx-utils-vscode/src/cli/forceConfigGet.ts
```diff
@@ -8,7 +8,10 @@
 import { SfdxCommandBuilder } from './commandBuilder';
 import { CliCommandExecutor } from './commandExecutor';
 import { CommandOutput } from './commandOutput';
-
+/**
+ * @deprecated
```
sfsholden, line 12: If this is deprecated is there a reason why we can't just remove it? We could always retrieve the logic from the history if we really need it

lcampos: Great question, we can't remove it right now cause it's being used in the ISV debugger (mainly). We need to refactor some of that code before we can delete this.

PR #2327, opened 2020-07-07
Webpack core extension

lcampos:
### What does this PR do?
Introduces the changes needed to run webpack on the core extension build.

### DIFF
--- a/packages/salesforcedx-vscode-core/webpack.config.js
+++ b/packages/salesforcedx-vscode-core/webpack.config.js
@@ -1,0 +1,93 @@
+const path = require('path');
+const glob = require('glob');
+const DIST = path.resolve(__dirname);
+
+const getEntryObject = () => {
+  const entryArray = glob.sync('src/**/*.ts');
+  const srcObj = entryArray.reduce((acc, item) => {
+    const modulePath = item.replace(/\/[\.A-Za-z0-9_-]*\.ts/g, '');
+    const outputModulePath = path.join('out', modulePath, 'index');
+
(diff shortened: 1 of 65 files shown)


packages/salesforcedx-vscode-core/webpack.config.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    return acc;
+  }, {});
+
+  if (getMode() !== 'development') {
+    return srcObj;
+  }
+
+  const entryTestArray = glob.sync('test/**/*.ts');
```
lcampos, line 24: There is currently a problem with this part of the configuration. Including it means we'll transpile the tests using the ts-loader plugin below, but our test suite is big enough to make webpack run out of memory. This only happens when running compile in `development` mode since `production` mode will not try to process test code.

I think we could either change the configuration so that we use webpack only for code in `src` and have typescript compile the code under `test`. Or we can bump the memory allocation config to something that allows us to run everything through webpack but that will have an impact on how we run our builds, etc.

brpowell: > ...we could either change the configuration so that we use webpack only for code in src and have typescript compile the code under test.

Sounds like the move 👍

sfsholden: I agree, that sounds like the better long term approach.

smithgp:
Sorry, pushed the un-Draft button on the wrong PR, just switched this back to draft.

gbockus-sf:
This is so old it's probably no longer useful.  Closing.

Issue #2281, opened 2020-06-20
Deploy on Save fails with no Error Message!


reddydev2018:
Deploy on Save used to work all fine, until I started getting this issue where it says Deploy failed but with no Error message. This seems to be weird, not sure how to get through this as I can't edit the code from any other place except VS Code! Any help is much appreciated - Thanks!

<img width="1120" alt="Screen Shot 2020-06-20 at 5 01 38 PM" src="https://user-images.githubusercontent.com/39275926/85211511-18824900-b318-11ea-988c-a91b53f008a5.png">

brpowell:
Hi @reddydev2018. Can you try disabling the deploy/retrieve enhancement setting and see if you get the errors for the component?

reddydev2018:
Hi Bryan - Sorry for the delay in response. Yeah I tried with that option is well be I am still seeing the same issue (Attached the screenshot) and even with the RightClick-->Deploy This Source to Org I am getting same issue. I am currently using 'SFDX force:source:deploy -p path to file' to make my life easy but I personally don't like to use CLI and in this case I don't have any other options left to choose :)

<img width="1438" alt="Screen Shot 2020-06-23 at 1 59 07 AM" src="https://user-images.githubusercontent.com/39275926/85366944-5210ab00-b4f6-11ea-916f-6ccb4d44c006.png">

lcampos:
Hi @reddydev2018, we found a scenario where the errors are not shown and have a fix for it but we're not sure it's the same issue you're running into. When you switched back to using the CLI were you able to get any details on the deploy error for that component ?

reddydev2018:
Hi @lcampos - Yes when I tried using CLI I was able to view the detailed error for the Component that I am trying to deploy on save. Also I observed that its not common with all the Components, it is just happening on random ones and some of them are working fine with deploy on save and I have no clue why is it behaving like that!

taalgaard:
@lcampos I wonder if your issue and my issue are the same?

https://github.com/forcedotcom/salesforcedx-vscode/issues/2256#issuecomment-648544065

lcampos:
@reddydev2018 in version 48.23.0 we released a fix for showing error messages that we were missing in certain scenarios. If you are still able to replicate the deploy failure on your side can you let us know if the new version addresses the error messages not showing ?

lcampos:
@taalgaard did you turn on the performance enhancements beta feature ?

taalgaard:
[...]

Issue #2188, opened 2020-05-11
Deploy (experimental) apex class with a duplicate name as one in a managed package 


mnunezdm:
### Summary

I'm trying to deploy an Apex Class with a name `TestClass.cls` into an org which has a managed package installed that also has a `TestClass.cls`

### Steps To Reproduce:

1. Activate experimental deploy
1. Install a managed package
1. Try to deploy an Apex Class which has a duplicate name with one in that managed package

### Expected result

Deployed correctly

### Actual result

In case the class is not created

```
11:35:15.287 Starting Deploy (Beta)
=== Deploy Errors
11:35:17.766 Finished Deploy (Beta)
```

In case the class is already created

```
11:52:42.619 Starting Deploy (Beta)
When retrieving results with Metadata or FullName fields, the query qualificatio​ns must specify no more than one row for retrieval. Result size: 2
```

### Additional information

_Feel free to attach a screenshot_.

**VS Code Version**: v1.45

**SFDX CLI Version**:

```
sfdx plugins --core                                                                                                                                                                                       
@oclif/plugin-autocomplete 0.1.5 (core)
@oclif/plugin-commands 1.2.3 (core)
@oclif/plugin-help 2.2.3 (core)
@oclif/plugin-not-found 1.2.3 (core)
@oclif/plugin-plugins 1.7.9 (core)
@oclif/plugin-update 1.3.9 (core)
@oclif/plugin-warn-if-update-available 1.7.0 (core)
@oclif/plugin-which 1.0.3 (core)
@salesforce/sfdx-trust 3.0.7 (core)
analytics 1.7.1 (core)
generator 1.1.2 (core)
salesforcedx 48.12.1 (core)
├─ @salesforce/sfdx-plugin-lwc-test 0.1.5 (core)
├─ salesforcedx-templates 48.10.0 (core)
└─ salesforce-alm 48.12.0 (core)

sfdx-cli 7.56.1 (core)
```

**OS and version**: Ubuntu 18.04.4 LTS in WSL 2

benjaminloerincz:
Also, on retrieving the class the class of the managed package will be retrieved instead of the class of the non-managed package.

sfdx-cli/7.60.0-e2bb2907cb darwin-x64 node-v10.15.3

lcampos:
The fix for both deploy and retrieve (experimental) has been released as part of 48.24.0. Thanks for providing feedback.

PR #2330, opened 2020-07-07
@W-7793005@ Support namespaces on beta deploy/retrieve

lcampos:
### What does this PR do?
Add support for namespaces on deploy/retrieve operations for the beta perf enhancements. This also fixes an error with parsing results in specific scenarios.

#2188, #2281 , @W-7793005@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
@@ -13,25 +13,16 @@ import {
   ContinueResponse,
   ParametersGatherer
 } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
-import {
-  RegistryAccess,
-  registryData
-} from '@salesforce/source-deploy-retrieve';
-import * as path from 'path';
+import { RegistryAccess } from '@salesforce/source-deploy-retrieve';
 import * as vscode from 'vscode';
 import { channelService } from '../channels';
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -130,9 +121,15 @@ export class LibraryDeploySourcePathExecutor extends LibraryCommandletExecutor<
         this.sourceClient.tooling.deploy
       );
 
+      const projectNamespace = (await SfdxProjectConfig.getValue(
+        'namespace'
+      )) as string;
       const registryAccess = new RegistryAccess();
       const components = registryAccess.getComponentsFromPath(response.data);
-      const deployPromise = this.sourceClient.tooling.deploy({ components });
+      const deployPromise = this.sourceClient.tooling.deploy({
```
AnanyaJha, line 129: can we create this `deployPromise` after getting the `metadataCount`? Or do we need to create this promise before `await`ing it later?

lcampos: We could, I kept it as it was and just added the `namespace` option to the existing call.

brpowell: The point of it was to have the request running as the map was being built, but if it's confusing you can change it.

PR #2345, opened 2020-07-14
Support for new JavaScript debugger in VS Code 1.47.0

xyc:
### What does this PR do?
Fixes LWC test debugging in VS Code 1.47.0, which introduced [new JavaScript debugger](https://github.com/microsoft/vscode-js-debug) by default on 07/10/2020 

@W-7771814@

### Functionality Before
Debugging LWC tests doesn't work if `"debug.javascript.usePreview"` VS Code setting is set to true (default value) in VS Code 1.47.0

### Functionality After
Debugging LWC tests works if `"debug.javascript.usePreview"` VS Code setting is set to true (default value) in VS Code 1.47.0

### DIFF
--- a/packages/salesforcedx-vscode-lwc/src/testSupport/commands/forceLwcTestDebugAction.ts
+++ b/packages/salesforcedx-vscode-lwc/src/testSupport/commands/forceLwcTestDebugAction.ts
@@ -42,6 +42,7 @@ export function getDebugConfiguration(
     cwd,
     runtimeExecutable: command,
     args,
+    resolveSourceMapLocations: ['**', '!**/node_modules/**'],
     console: 'integratedTerminal',
     internalConsoleOptions: 'openOnSessionStart',
     port: 9229,
--- a/packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
+++ b/packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
@@ -4,6 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { workspace } from 'vscode';
 import { TestRunType } from '../testRunner/testRunner';
 
 /**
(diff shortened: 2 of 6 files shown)


packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
```diff
@@ -16,7 +17,12 @@ export function getCliArgsFromJestArgs(
...
 ) {
   const cliArgs = ['--', ...jestArgs];
-  if (testRunType === TestRunType.DEBUG) {
+
+  const usePreviewJavaScriptDebugger = workspace
+    .getConfiguration('debug')
+    .get('javascript.usePreview');
+
+  if (testRunType === TestRunType.DEBUG && !usePreviewJavaScriptDebugger) {
     cliArgs.unshift('--debug');
```
xyc, line 26: This is the gist of the fix. `--inspect-brk` is no longer necessary - https://github.com/microsoft/vscode/issues/99526#issuecomment-657114179 for the [new VS Code JavaScript debugger](https://github.com/microsoft/vscode-js-debug). (Note that the comment mentions a fallback. However it doesn't apply here because it looks up package.json scripts for `--inspect-brk` args).

Therefore we don't need the `--debug` args in `sfdx-lwc-jest` or `lwc-test` if users are using the new debugger. Launch configuration without `--debug` args would work with the new debugger. For user who has specified using the old debugger, we keep the same behavior.

cc @trevor-bliss @muenzpraeger

packages/salesforcedx-vscode-lwc/src/testSupport/commands/forceLwcTestDebugAction.ts
```diff
@@ -42,6 +42,7 @@ export function getDebugConfiguration(
     cwd,
     runtimeExecutable: command,
     args,
+    resolveSourceMapLocations: ['**', '!**/node_modules/**'],
```
xyc, line 45: The new debugger needs this arg so it doesn't show a warning message about missing source map for rollup (warning only, doesn't affect debugging). This doesn't affect the old debugger.

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getCliArgsFromJestArgs.ts
muenzpraeger: Nice!

packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/mocks/vscodeConfiguration.ts
```diff
@@ -0,0 +1,38 @@
+import * as vscode from 'vscode';
```
lcampos, line 1:
```suggestion
/*
 * Copyright (c) 2020, salesforce.com, inc.
 * All rights reserved.
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 */
import * as vscode from 'vscode';
```

PR #2349, opened 2020-07-15
Added a flag to control usage of Apex library

AnanyaJha:
### What does this PR do?
This PR adds a flag so users can revert back to the original CLI implementation of Apex commands if needed

 @W-7826660@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -803,6 +803,11 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
+        },
+        "salesforcedx-vscode-core.experimental.cliCommand": {
+          "type": "boolean",
+          "default": false,
+          "description": "%experimental_cli_command_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -804,10 +804,10 @@
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
         },
-        "salesforcedx-vscode-core.experimental.cliCommand": {
+        "salesforcedx-vscode-core.experimental.useApexLibrary": {
           "type": "boolean",
-          "default": false,
-          "description": "%experimental_cli_command_description%"
+          "default": true,
+          "description": "%experimental_apex_library_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -63,5 +63,6 @@
   "force_source_retrieve_display_text": "Retrieve Source from Org",
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
-  "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience."
+  "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
+  "experimental_cli_command_description": "Revert to CLI implementation of Apex commands"
 }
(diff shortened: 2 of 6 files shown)


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -63,5 +63,6 @@
   "force_source_retrieve_display_text": "Retrieve Source from Org",
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
-  "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience."
-}
\ No newline at end of file
+  "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
+  "experimental_cli_command_description": "Revert to CLI implementation of Apex commands"
```
AnanyaJha, line 67: @ekapner For this PR we're adding a new flag to control switching between the CLI implementation for Apex commands and the new library that we're replacing that with. Right now we've named the flag as `cliCommand` under the `experimental` category and the description is above ^^. Please let me know if you have any suggestions for the name or description for the flag, thanks!

AnanyaJha: Here's what it would look like for the user 
Screen Shot 2020-07-15 at 11 09 08 AM

packages/salesforcedx-vscode-core/package.json
```diff
@@ -803,6 +803,11 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
+        },
+        "salesforcedx-vscode-core.experimental.cliCommand": {
```
lcampos, line 807: We can have this under experimental, change `cliCommand` to `apexlibrary` or something apex specific and default it to true. 

Or we can change the name to be nothing under experimental and keep it as something along the lines of default to cli command use.

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -6,23 +6,97 @@
...
+import { getRootWorkspacePath, hasRootWorkspace } from '../util';
 import {
   ApexLibraryExecutor,
+  CommandletExecutor,
   SfdxCommandlet,
+  SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './util';
 
+export class ForceApexExecuteExecutor extends SfdxCommandletExecutor<{}> {
```
lcampos, line 35: Is this shown as new because we removed it in a previous PR ?

AnanyaJha: Yeah we removed it in the one that introduced the apex library changes

packages/salesforcedx-vscode-core/package.nls.json
ekapner: @AnanyaJha - pinged you in the work item - need some background info before I can provide feedback.

ekapner: How about Experimental: Use Apex Library

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -28,3 +28,4 @@ export const DEFAULT_DEV_HUB_USERNAME_KEY = 'defaultdevhubusername';
 export const CONFLICT_DETECTION_ENABLED = 'detectConflictsAtSync';
 export const BETA_DEPLOY_RETRIEVE = 'experimental.deployRetrieve';
 export const PKG_ID_PREFIX = '04t';
+export const USE_APEX_LIBRARY = 'experimental.useApexLibrary';
```
xyc, line 31: Place config values next to each other? Ideally i'd like to see them grouped under an object but that's too much refactoring for now

AnanyaJha: True I think it's rather confusing to have the config values spread out across multiple files, so it could be worth looking into refactoring. For now, I think it's best to keep this value within the `constants` file so we're at least following the current pattern to minimize more confusion

xyc: Yeah, I mean in this file group the values together and maybe add a comment :)

The underlying idea is that these values are implicitly typed (getConfigValue would expect a config value rather than `string`)

PR #2355, opened 2020-07-16
Enhance Telemetry Data for Single File Deploys @W-7827398@

amandastern99:
Added telemetry data to collect metadata component counts for single file deploys to compare beta deploy times with the CLI.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
@@ -73,7 +73,7 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
       } catch (e) {
         telemetryService.sendException(
           e.name,
-          'error detecting deploy components'
+          'error detecting deployed components'
         );
       }
       this.logMetric(execution.command.logName, startTime, properties);
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
@@ -14,10 +14,7 @@ import {
   Table
 } from '@salesforce/salesforcedx-utils-vscode/out/src/output';
 import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
-import {
-  RegistryAccess,
-  SourceClient
-} from '@salesforce/source-deploy-retrieve';
+import { RegistryAccess } from '@salesforce/source-deploy-retrieve';
 import * as vscode from 'vscode';
 import { channelService } from '../channels';
 import { handleDiagnosticErrors } from '../diagnostics';
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
@@ -14,7 +14,10 @@ import {
   Table
 } from '@salesforce/salesforcedx-utils-vscode/out/src/output';
 import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
-import { RegistryAccess } from '@salesforce/source-deploy-retrieve';
+import {
+  RegistryAccess,
+  SourceClient
+} from '@salesforce/source-deploy-retrieve';
 import * as vscode from 'vscode';
 import { channelService } from '../channels';
 import { handleDiagnosticErrors } from '../diagnostics';
@@ -35,11 +38,13 @@ export enum DeployType {
 export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
   string
 > {
+  protected sourceClient: SourceClient | undefined;
   public static errorCollection = vscode.languages.createDiagnosticCollection(
     'deploy-errors'
   );
 
   public execute(response: ContinueResponse<string>): void {
+    const source = this.sourceClient;
     const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
@@ -66,9 +71,10 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
         );
         const metadataCount = JSON.stringify(createComponentCount(components));
         properties = { metadataCount };
-      } catch (e) {}
-      this.logMetric(execution.command.logName, startTime, properties);
-
+        this.logMetric(execution.command.logName, startTime, properties);
+      } catch (e) {
+        properties = '';
+      }
       try {
         if (stdOut) {
           const deployParser = new ForceDeployResultParser(stdOut);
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -57,7 +63,18 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
+      const registryAccess = new RegistryAccess();
+      try {
+        const components = registryAccess.getComponentsFromPath(
+          execFilePathOrPaths
+        );
+        const metadataCount = JSON.stringify(createComponentCount(components));
+        properties = { metadataCount };
+        this.logMetric(execution.command.logName, startTime, properties);
+      } catch (e) {
+        properties = '';
```
xyc, line 76: is this line necessary?

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -57,7 +63,18 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
-      this.logMetric(execution.command.logName, startTime);
+      let properties;
+      const registryAccess = new RegistryAccess();
+      try {
+        const components = registryAccess.getComponentsFromPath(
+          execFilePathOrPaths
+        );
+        const metadataCount = JSON.stringify(createComponentCount(components));
+        properties = { metadataCount };
+        this.logMetric(execution.command.logName, startTime, properties);
```
brpowell, line 74: This needs to be moved below the try/catch block. If there's an error, we won't log anything at all.

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -33,11 +38,13 @@ export enum DeployType {
 export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
   string
 > {
+  protected sourceClient: SourceClient | undefined;
```
brpowell, line 41:
```suggestion
```

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -33,11 +38,13 @@ export enum DeployType {
...
   public static errorCollection = vscode.languages.createDiagnosticCollection(
     'deploy-errors'
   );
 
   public execute(response: ContinueResponse<string>): void {
+    const source = this.sourceClient;
```
brpowell, line 47:
```suggestion
```

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -57,7 +61,17 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
-      this.logMetric(execution.command.logName, startTime);
+      let properties;
+      const registryAccess = new RegistryAccess();
+      try {
+        const components = registryAccess.getComponentsFromPath(
+          execFilePathOrPaths
+        );
+        const metadataCount = JSON.stringify(createComponentCount(components));
+        properties = { metadataCount };
+      } catch (e) {}
```
brpowell, line 72: Since we're silently swallowing this error (generally we shouldn't but I think it's alright here) let's put a comment saying something like "registry does not handle multiple paths. only log component count for single paths".

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -14,6 +14,10 @@ import {
   Table
 } from '@salesforce/salesforcedx-utils-vscode/out/src/output';
 import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import {
+  RegistryAccess,
+  SourceClient
```
brpowell, line 19: Remove the `SourceClient` import

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -128,8 +128,7 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
       )) as string;
       const registryAccess = new RegistryAccess();
       const components = registryAccess.getComponentsFromPath(response.data);
-      const deployPromise = this.sourceClient.tooling.deploy({
-        components,
+      const deployPromise = this.sourceClient.tooling.deploy(components, {
```
brpowell, line 131: Should this be changed? I don't think we're using the new version of the library yet.

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
lcampos: We should still log this error too. Outside of not supporting multiple paths this could help us find other errors with the registry being used on metadata types with different shapes that might cause problems.

amandastern99: What do you think would be the best thing to log for this?

AnanyaJha: I think you could log something similar to what Bryan suggested above, "Registry does not handle multiple paths yet"

lcampos: You should log it as an exception, similar to how we do it a few lines below in this same file (https://github.com/forcedotcom/salesforcedx-vscode/blob/dc1e637a517d70b6b199cc5a0ae08d96e4580bb1/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts#L83)

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -57,7 +58,23 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
+        const components = registryAccess.getComponentsFromPath(
+          execFilePathOrPaths
+        );
+        const metadataCount = JSON.stringify(createComponentCount(components));
+        properties = { metadataCount };
+        // registry does not handle multiple paths. only log component count for single paths
+      } catch (e) {
+        telemetryService.sendException(
+          e.name,
+          'error detecting deploy components'
```
brpowell, line 73:
```suggestion
          'error detecting deployed components'
```

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -57,7 +58,23 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
+        const components = registryAccess.getComponentsFromPath(
+          execFilePathOrPaths
+        );
+        const metadataCount = JSON.stringify(createComponentCount(components));
+        properties = { metadataCount };
+        // registry does not handle multiple paths. only log component count for single paths
+      } catch (e) {
+        telemetryService.sendException(
+          e.name,
+          'error detecting deployed components'
```
lcampos, line 73:
```suggestion
          `error detecting deployed components: ${e.message}`
```

lcampos: Please check if this compiles first :)

PR #2366, opened 2020-07-20
Module changes needed to introduce webpack use in core extension

lcampos:
### What does this PR do?
This PR updates some code in the core package so it follows the module import/export pattern that will allow us to introduce webpack (follow up PR). These changes are taken from #2327 with the exception of the webpack configuration.


xyc:
@lcampos Could you comment more on what module import/export pattern is required for webpack

PR #2367, opened 2020-07-21
Update snyk to 1.364.0

xyc:
### What does this PR do?
Update snyk version

@W-7854057@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,8 +120,6 @@ commands:
       - restore_cache:
           keys:
             - << parameters.base-cache-path >>{{ checksum "package.json" }}
-            # if cache for exact version of 'package.json' is not present, then load the most recent cache
-            - << parameters.base-cache-path >>
 
   save_dependency_cache:
     parameters:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -120,8 +120,6 @@ commands:
       - restore_cache:
           keys:
             - << parameters.base-cache-path >>{{ checksum "package.json" }}
-            # if cache for exact version of 'package.json' is not present, then load the most recent cache
```
sfsholden, line 123: Xiaoyi and I synced on this - I am in agreement with removing this fallback key. It seems to be causing more issues that it really helps with. In addition, when we have a new checksum for our package.json it will be better for us to regenerate the cache. We think that part of the errors that were being thrown (perm issues) could be due to the fact that the windows jobs are generating a new cache and then the linux job is trying to reference it.

xyc: Thanks @sfsholden For reference there's a circle CI article if we were to see the perm issue later: 
https://support.circleci.com/hc/en-us/articles/360004250693-Restoring-cache-fails-with-Permission-Denied-
Section "Architecture Permissions Issues"

lcampos: Go for it

Issue #2369, opened 2020-07-22
"Execute Anonymous Apex with Editor Contents" no longer working in temporary editors


klouf-mp:
### Summary

When running the command "SFDX: Execute Anonymous Apex with Editor Contents" on a temporary editor (an editor with contents that have never been saved to a file), the command fails with the error "File not found at the specified path: Untitled-1".

Saving the contents of the editor to a file and then running the command results in expected behavior.

### Steps To Reproduce:

1.  Open a new editor
2. Enter some valid Anonymous Apex code to be executed
3. Run SFDX Extension command "SFDX: Execute Anonymous Apex with Editor Contents"
4. Command fails with error "File not found at the specified path: Untitled-1".

### Expected result

The Apex code should be executed and the results displayed in the Output panel. (I think previously it was saved to a temporary file and that was what was used as input, but I could be wrong on that? Seems like maybe that's not happening anymore?)

### Actual result

Command fails with error "File not found at the specified path: Untitled-1".

### Additional information

**VS Code Version**: 1.47.2

**SFDX CLI Version**: sfdx-cli/7.66.2-4f159a1d07 darwin-x64 node-v12.13.0
**SFDX Plugins**: 
@oclif/plugin-autocomplete 0.2.0
@oclif/plugin-commands 1.3.0 (core)
@oclif/plugin-help 3.0.1 (core)
@oclif/plugin-not-found 1.2.4 (core)
@oclif/plugin-plugins 1.7.10 (core)
@oclif/plugin-update 1.3.10 (core)
@oclif/plugin-warn-if-update-available 1.7.0 (core)
@oclif/plugin-which 1.0.3 (core)
@salesforce/sfdx-diff 0.0.6
@salesforce/sfdx-trust 3.0.7 (core)
@zervice/zervice-sfdx-plugins 2.0.3
analytics 1.12.0 (core)
generator 1.1.3 (core)
salesforcedx 49.2.3 (core)
├─ custom-metadata 1.0.7 (core)
├─ salesforcedx-templates 48.32.0 (core)
├─ @salesforce/sfdx-plugin-lwc-test 0.1.7 (core)
└─ salesforce-alm 49.3.0 (core)
sfdmu 3.5.8
sfdx-cli 7.66.2 (core)

**OS and version**: macOS Catalina 10.15.6

AnanyaJha:
[...]

PR #2370, opened 2020-07-22
Execute anonymous apex if file is not saved

AnanyaJha:
### What does this PR do?
This PR addresses the issue where an error is thrown if the user tries to run `SFDX: Execute Anonymous Apex with Editor Contents` when the file has not been saved yet.

#2369, @W-7860461@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -109,10 +109,14 @@ export class AnonApexGatherer
       }
 
       const document = editor.document;
-      if (!editor.selection.isEmpty) {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -109,10 +109,14 @@ export class AnonApexGatherer
       }
 
       const document = editor.document;
-      if (!editor.selection.isEmpty) {
+      if (!editor.selection.isEmpty || !fs.existsSync(document.uri.fsPath)) {
```
xyc, line 112: can we check for `document.isUntitled` here instead?

AnanyaJha: Oh so cool, I haven't seen that api before

PR #2383, opened 2020-07-29
Consume new deploy result in VSCode @W-7815625@

amandastern99:
This pull request consumes the new metadata deploy result in vscode. If a single file deploy for Apex, Visualforce, LWC or Aura does not require the use of a namespace it will be deployed using the metadata deploy.

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -26,7 +26,6 @@
   ],
   "dependencies": {
     "@salesforce/core": "2.5.1",
-    "@salesforce/kit": "1.2.2",
     "@salesforce/salesforcedx-utils-vscode": "49.0.0",
     "@salesforce/source-deploy-retrieve": "1.0.15",
     "@salesforce/apex-node": "0.0.2",
@@ -58,7 +57,7 @@
     "mock-spawn": "0.2.6",
     "nyc": "^13",
     "sinon": "^7.3.1",
-    "typescript": "3.1.6"
+    "typescript": "3.7.5"
   },
   "scripts": {
     "vscode:prepublish": "npm prune --production",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -29,6 +29,7 @@
     "@salesforce/kit": "1.2.2",
     "@salesforce/salesforcedx-utils-vscode": "49.0.0",
     "@salesforce/source-deploy-retrieve": "1.0.15",
+    "@salesforce/apex-node": "0.0.2",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -27,7 +27,7 @@
   "dependencies": {
     "@salesforce/apex-node": "0.0.4",
     "@salesforce/core": "2.5.1",
-    "@salesforce/source-deploy-retrieve": "1.0.17",
+    "@salesforce/source-deploy-retrieve": "1.0.18",
     "@salesforce/salesforcedx-utils-vscode": "49.5.0",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -803,6 +803,11 @@
           "type": "boolean",
           "default": false,
           "description": "%experimental_deploy_retrieve_description%"
+        },
+        "salesforcedx-vscode-core.experimental.useApexLibrary": {
+          "type": "boolean",
+          "default": true,
+          "description": "%experimental_apex_library_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -15,7 +15,7 @@
     "theme": "light"
   },
   "aiKey": "ec3632a4-df47-47a4-98dc-8134cacbaf7e",
-  "version": "49.0.0",
+  "version": "49.5.0",
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
@@ -25,10 +25,10 @@
     "Other"
   ],
   "dependencies": {
+    "@salesforce/apex-node": "0.0.4",
     "@salesforce/core": "2.5.1",
-    "@salesforce/salesforcedx-utils-vscode": "49.0.0",
     "@salesforce/source-deploy-retrieve": "1.0.17",
-    "@salesforce/apex-node": "0.0.2",
+    "@salesforce/salesforcedx-utils-vscode": "49.5.0",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
@@ -38,7 +38,7 @@
     "xmldom-sfdx-encoding": "0.1.29"
   },
   "devDependencies": {
-    "@salesforce/salesforcedx-test-utils-vscode": "49.0.0",
+    "@salesforce/salesforcedx-test-utils-vscode": "49.5.0",
     "@salesforce/ts-sinon": "^1.0.0",
     "@types/adm-zip": "^0.4.31",
     "@types/chai": "^4.0.0",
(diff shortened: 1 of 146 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "@salesforce/core": "2.5.1",
     "@salesforce/salesforcedx-utils-vscode": "49.0.0",
-    "@salesforce/source-deploy-retrieve": "1.0.15",
+    "@salesforce/source-deploy-retrieve": "1.0.16",
```
lcampos, line 30: Should we be using version `1.0.17` since it's the latest one ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -13,9 +13,17 @@ import {
...
+  RegistryAccess,
+  ToolingDeployStatus
+} from '@salesforce/source-deploy-retrieve';
+import {
+  DeployStatus,
+  SourceDeployResult
+} from '@salesforce/source-deploy-retrieve/lib/types/newClient';
 import * as vscode from 'vscode';
 import { channelService } from '../channels';
+import { handleDeployRetrieveLibraryDiagnostics } from '../diagnostics/diagnostics';
```
lcampos, line 26:
```suggestion
import { handleDeployRetrieveLibraryDiagnostics } from '../diagnostics';
```

lcampos:
The changes are looking good but you might want to rebase soon since some of the files you're updating have had changes in the last days.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -114,28 +119,53 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+      )) as SourceDeployResult;
+      const parser = new LibraryDeployResultParser(result);
+      const outputResult = parser.resultParser(result);
       this.logMetric({ metadataCount });
+      channelService.appendLine(outputResult);
+      channelService.showCommandWithTimestamp(`Finished ${this.executionName}`);
+      if (
+        result.status === DeployStatus.Succeeded ||
+        // @ts-ignore
+        result.status === ToolingDeployStatus.Completed
```
brpowell, line 157: Does this have to be ts-ignored? `status` should support both enum values

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -114,28 +119,53 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+      ) {
+        DeployRetrieveLibraryExecutor.errorCollection.clear();
+        await notificationService.showSuccessfulExecution(this.executionName);
```
brpowell, line 160:
```suggestion
        notificationService.showSuccessfulExecution(this.executionName);
```

Can remove since further execution isn't dependent on this exiting

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -114,28 +119,53 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+      } else {
+        handleDeployRetrieveLibraryDiagnostics(
+          // @ts-ignore
+          result,
+          DeployRetrieveLibraryExecutor.errorCollection
```
brpowell, line 165: Does this need to be ts-ignored?

packages/salesforcedx-vscode-core/src/commands/util/libraryDeployResultParser.ts
```diff
@@ -20,72 +20,73 @@ type ComponentSuccess = {
...
-      success = listOfFiles.map(file => ({
-        state: mdState,
-        fullName: componentSuccess.fullName!,
-        type: componentSuccess.componentType,
-        filePath: file
-      }));
+    for (const component of componentSuccess.components!) {
+      const listOfFiles = component.component.walkContent();
+      listOfFiles.push(component.component.xml);
+      if (listOfFiles) {
```
brpowell, line 40: `walkContent` always returns an array so this check isn't needed.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -114,28 +119,53 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+        );
```
AnanyaJha, line 166: are we moving away from using the deploy wrapper to run the command & contain all the output processing?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/betaDeployRetrieve.test.ts
```diff
@@ -40,18 +40,22 @@ describe('Force Source Deploy with Sourcepath Beta', () => {
       sandboxStub
         .stub(SfdxCoreSettings.prototype, 'getBetaDeployRetrieve')
         .returns(true);
-      const components: MetadataComponent[] = [
+      const components: SourceComponent[] = [
+        // @ts-ignore
```
AnanyaJha, line 44: do we need the `@ts-ignore` statements in this file?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
amandastern99: This causes the warning "Promises must be handled appropriately (no-floating-promises)." Should I ignore this warning?

xyc: you can append a catch block to get rid of the warning: `notificationService.showSuccessfulExecution(this.executionName).catch(error => {})`;

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/betaDeployRetrieve.test.ts
```diff
@@ -20,6 +19,23 @@ import {
 } from '../../../../src/commands/util/betaDeployRetrieve';
 import { SfdxCoreSettings } from '../../../../src/settings/sfdxCoreSettings';
 
+function createComponent(type: any, ext: string, extrafile?: string) {
```
lcampos, line 22: We can replace `any` with `MetadataType`

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -114,28 +120,53 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+          location: vscode.ProgressLocation.Notification
+        },
+        () => deployPromise
+      )) as SourceDeployResult;
+      const parser = new LibraryDeployResultParser(result);
+      const outputResult = parser.resultParser(result);
       this.logMetric({ metadataCount });
+      channelService.appendLine(outputResult);
+      channelService.showCommandWithTimestamp(`Finished ${this.executionName}`);
+      if (
```
lcampos, line 155: I think this needs to be updated to `if (result.status === DeployStatus.Succeeded ||
        result.status === ToolingDeployStatus.Completed){ `

packages/salesforcedx-vscode-core/package.json
```diff
@@ -27,8 +27,8 @@
   "dependencies": {
     "@salesforce/apex-node": "0.0.4",
     "@salesforce/core": "2.5.1",
+    "@salesforce/source-deploy-retrieve": "1.0.17",
```
lcampos, line 30: Shouldn't we use `1.0.18` ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: @brpowell same question. a lot of the things added here can be moved to the wrapper code

brpowell: If i remember correctly we updated it to be more straightforward since we were having issues debugging/conditionally setting the api to use based on the namespace, but it was awhile ago so can't remember exactly. And since this command is the only one essentially using a wrapper it's not repeating any logic anyway. But it can be changed back if wanted.

PR #2388, opened 2020-07-31
Added docs for remote dev using WSL2

ntotten:
### What does this PR do?
Adds docs for using VS Code remote dev with WSL2.

### DIFF
--- a/docs/_articles/en/user-guide/remote-development.md
+++ b/docs/_articles/en/user-guide/remote-development.md
@@ -1,21 +1,26 @@
 ---
-title: Remote Development - Containers
+title: Remote Development
 lang: en
 ---
 
 ## Overview
 
-Salesforce Extension for VS Code supports remote development and allows you to use a container as a full-featured development environment. You can open a project mounted into the container and edit with full code completions, code navigation, debugging, and more.
+Salesforce Extension for VS Code supports remote development and allows you to use a container as a full-featured development environment. You can open a project mounted into a Docker container or using WSL2 and edit with full code completions, code navigation, debugging, and more.
 
 > NOTICE: The remote development feature is currently in beta. If you find any bugs or have feedback, [open a GitHub issue](./bugs-and-feedback).
 
 Refer to VS Code documentation if you want to understand more about remote development:
 
 - Use a [Docker container](https://www.docker.com/) as your [development container](https://code.visualstudio.com/docs/remote/containers#_indepth-setting-up-a-folder-to-run-in-a-container)
+- Use [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) as your [development container](https://code.visualstudio.com/docs/remote/wsl)
 - Provide [full-featured development environment](https://code.visualstudio.com/docs/remote/remote-overview)
 - Switch your development environment by [connecting to a container](https://code.visualstudio.com/docs/remote/containers)
 
-## Install
+## Using a Docker Container
+
+For remote development using a Docker Container follow the instructions below.
+
+### Install
 
 To start remote development in dev container, install:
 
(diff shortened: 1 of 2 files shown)


docs/_articles/en/user-guide/remote-development.md
```diff
@@ -30,35 +35,90 @@ To start remote development in dev container, install:
...
 
+## Using Windows Subsystem for Linux
+
+For remote development using the Windows Subsytem for Linux follow the instructions below.
+
+## Install
+
+To start developing in a container with the Windows Subsystem for Linux (WSL), first you must set up WSL on your machine. To do so, following the [instructions provided by Microsoft](https://docs.microsoft.com/en-us/windows/wsl/install-win10). **Note, you should only use WSL2. There are known issues with WSL1. The instructions below assume Ubuntu. Other Linux distros should work, but the steps may be slightly different.**
+
+After you have installed WSL2, you will need to setup your environment for Salesforce Development. The two main things you will need are Node.js, the Salesforce CLI, and the JDK.
```
lcampos, line 75: The problem I see here is that the Salesforce CLI is not officially supported for running in WSL2. Has that changed ?

ntotten: WSL2 is - it is actually just Linux, so I really don't think there should be an issue. WSL1 was more problematic. In fact, we should probably change what we say about WSL not being supported and say that WSL1 isn't supported, but WSL2 is. //cc @clairebianchi

docs/_articles/en/user-guide/remote-development.md
```diff
@@ -1,21 +1,26 @@
 ---
-title: Remote Development - Containers
+title: Remote Development
 lang: en
 ---
 
 ## Overview
 
-Salesforce Extension for VS Code supports remote development and allows you to use a container as a full-featured development environment. You can open a project mounted into the container and edit with full code completions, code navigation, debugging, and more.
+Salesforce Extension for VS Code supports remote development and allows you to use a container as a full-featured development environment. You can open a project mounted into a Docker container or using WSL2 and edit with full code completions, code navigation, debugging, and more.
```
ekapner, line 8: would spell out like this per my web search: Windows Subsystem for Linux (WSL) 2

docs/_articles/en/user-guide/remote-development.md
```diff
@@ -1,21 +1,26 @@
...
 
 > NOTICE: The remote development feature is currently in beta. If you find any bugs or have feedback, [open a GitHub issue](./bugs-and-feedback).
 
 Refer to VS Code documentation if you want to understand more about remote development:
 
 - Use a [Docker container](https://www.docker.com/) as your [development container](https://code.visualstudio.com/docs/remote/containers#_indepth-setting-up-a-folder-to-run-in-a-container)
+- Use [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) as your [development container](https://code.visualstudio.com/docs/remote/wsl)
```
ekapner, line 15: Do we need to specify the "2" for WSL ?

docs/_articles/en/user-guide/remote-development.md
```diff
@@ -30,35 +35,90 @@ To start remote development in dev container, install:
...
+![Authorize Success Message](./images/authorize_message.png)
+
+1. Run **SFDX: Create a Default Scratch Org**.
 1. Click **Reopen in Container** to build a dev container. Now the side bar shows `Dev Container: Salesforce Project`.
 
    You’ve set up a dev container for an existing project to use as your full-time development environment.
 
+## Using Windows Subsystem for Linux
+
+For remote development using the Windows Subsytem for Linux follow the instructions below.
```
ekapner, line 69: ...for Linux, follow the ...

ekapner: @lcampos - should I go ahead and make a branch off this branch and update the docs that way?

docs/_articles/en/user-guide/remote-development.md
lcampos: Yes, there is a previous version which is known as WSL or WSL 1

docs/_articles/en/user-guide/remote-development.md
lcampos: @ekapner you can just branch out of the current `develop` branch since this has been merged to it.

PR #2393, opened 2020-08-04
Fix snippets for Assertions

PawelWozniak:
### What does this PR do?
Added missing "s" in assertion snippets as per documentation of system method.
https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_methods_system_system.htm

### Functionality Before
Snippets were resolved to "System.AssertEqual" and "System.AssertNotEqual" which gives error: method do not exists.

### Functionality After
Snippets are now resolved to "System.AssertEquals" and "System.AssertNotEquals" which are correct method names as per documentation of System class.

### DIFF
--- a/packages/salesforcedx-vscode-apex/snippets/apex.json
+++ b/packages/salesforcedx-vscode-apex/snippets/apex.json
@@ -40,33 +40,33 @@
         ],
         "description": "System.assert(false, message);"
     },
-    "System Assert Equal": {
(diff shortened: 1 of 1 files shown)


lcampos:
@PawelWozniak thanks for contributing to the project. We've merged your PR and it will be part of version 49.5.0 of the extensions.

PR #2396, opened 2020-08-04
@W-7904975@ Do not show SObject refresh pop up when a project does not have a defaultusername

jag-sfdc:
### What does this PR do?
Checks if an org has been authorized before showing a pop up to refresh sobjects

@W-7904975@

### Functionality Before
Screen Shot 2020-07-20 at 2 34 46 PM

### Functionality After
The pop up to refresh sobject is not shown if no org is authorized.

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
@@ -190,7 +190,12 @@ export async function forceGenerateFauxClassesCreate(
 export async function initSObjectDefinitions(projectPath: string) {
   const hasDefaultUsernameSet =
     (await getDefaultUsernameOrAlias()) !== undefined;
-  initSObjectDefinitionsWithUserName(projectPath, hasDefaultUsernameSet);
+  initSObjectDefinitionsWithUserName(projectPath, hasDefaultUsernameSet).catch(e =>
+    telemetryService.sendErrorEvent({
+      message: e.message,
+      stack: e.stack
+    })
+  );
 }
 
 export async function initSObjectDefinitionsWithUserName(projectPath: string, hasDefaultUsernameSet: boolean) {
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -190,6 +190,15 @@ export async function forceGenerateFauxClassesCreate(
 export async function initSObjectDefinitions(projectPath: string) {
   const hasDefaultUsernameSet =
```
lcampos, line 191: Isn't the root of the problem in this evaluation ? In the past we were already checking for this and the projectPath before triggering the refresh.

packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -190,6 +190,15 @@ export async function forceGenerateFauxClassesCreate(
...
     (await getDefaultUsernameOrAlias()) !== undefined;
+  initSObjectDefinitionsWithUserName(projectPath, hasDefaultUsernameSet).catch(e =>
+    telemetryService.sendErrorEvent({
+      message: e.message,
+      stack: e.stack
+    })
+  );
+}
+
+export async function initSObjectDefinitionsWithUserName(projectPath: string, hasDefaultUsernameSet: boolean) {
```
lcampos, line 201: `initSObjectDefinitions` and `initSObjectDefinitionsWithUserName` should be merged into one method, we don't have a scenario where we want to trigger a refresh operation when the project does not have a default username assigned.

packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
jag-sfdc: Yes, this is the root of the problem. We still want to throw the error as usual when someone has no authorized org and runs the "Refresh SObjects" manually. But, when we prompt them with an option to refresh sobjects, we make sure org is authorized and hence call initSObjectDefinitionsWithUsername.
Ideally I hate this name and would prefer simple function overloading, but typescript doesn't allow it.

Thinking about it now, I think the better naming will be 
initSObjectDefinitions(projectPath: string) changed to verifyUsernameAndInitSObjectDefinitions(projectPath: string) -> Used by the command
initSObjectDefinitionsWithUserName(projectPath: string, hasDefaultUsernameSet: boolean) changed to initSObjectDefinitions(projectPath: string) -> user by our new refresh button and we have made sure username has been set.

PR #2407, opened 2020-08-10
ISV Debugger - Clear project prior to new creation (@W-5620191)

sfsholden:
### What does this PR do?
Fixes an issue with the ISV Debugger that leaves stale unmanaged source and installed packages. Upon creation of the ISV project, we will remove the current project after prompting the user with an overwrite prompt.

@W-5620191@

### Functionality Before
_This screen recording shows an old 'Subscriber' project being overwritten by a new 'Subscriber' project with the ISV Debugger. Upon completion, there are still old installed-packages at `.sfdx/tools/installed-packages`, along with old unmanaged source code in `force-app/main/default/classes`._
debugger_before_editted

### Functionality After
_This screen recording shows an old 'Subscriber' project being completely overwritten by the new 'Subscriber' project. Only the new installed-packages are seen in `.sfdx/tools/installed-packages` and old unmanaged source code is no longer present._
debugging_after_edited

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
@@ -250,6 +250,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       this.relativeInstalledPackagesPath
     );
 
+    // remove any previous project at this path location
+    shell.rm('-rf', projectPath);
+
     // 1: create project
     await this.executeCommand(
       this.buildCreateProjectCommand(response.data),
@@ -265,6 +268,8 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       cancellationTokenSource,
       cancellationToken
     );
+
+    console.log('It got here at least');
 
     // 2b: update sfdx-project.json with namespace
     const orgNamespaceInfoResponseJson = await this.executeCommand(
@@ -383,6 +388,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
     try {
       const packagesTempPath = path.join(projectMetadataTempPath, 'packages');
       shell.mkdir('-p', packagesTempPath);
+
+      // Remove any previous installed packages at this location.
+      // shell.rm('-rf', projectInstalledPackagesPath);
       shell.mkdir('-p', projectInstalledPackagesPath);
       const zip = new AdmZip(
         path.join(projectMetadataTempPath, 'unpackaged.zip')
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -202,8 +202,7 @@ export const messages = {
     'SFDX: ISV Debugger Setup, Step 7 of 7: Converting package: %s',
   isv_debug_bootstrap_processing_package: 'Processing package: %s',
   isv_debug_bootstrap_generate_launchjson: 'Creating launch configuration',
-  isv_debug_bootstrap_open_project:
-    'Opening project in new Visual Studio Code window',
+  isv_debug_bootstrap_open_project: 'Opening project in Visual Studio Code',
```
sfsholden, line 205: We can't guarantee that we will open the project in a new VSCode window if the user already has that project open. This is working as designed and controlled on VSCode's side.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/isvdebugger/bootstrapCmd.test.ts
```diff
@@ -19,7 +19,8 @@ import { getRootWorkspacePath } from '../../../../src/util';
 describe('ISV Debugging Project Bootstrap Command', () => {
   const LOGIN_URL = 'a.b.c';
   const SESSION_ID = '0x123';
-  const PROJECT_NAME = 'sfdx-simple';
```
sfsholden, line 22: Since we are now wiping out the project prior to executing the ISV Debugger, we will copy the contents of 'sfdx-simple' into a new clone with the stub project setup call.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sfsholden: @ekapner Minor wording update for ISV Debugger upon completion in VSCode.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -250,6 +250,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
       this.relativeInstalledPackagesPath
     );
 
+    // remove any previous project at this path location
+    shell.rm('-rf', projectPath);
```
xyc, line 254: Do we want to warn user that this actually removes everything vs. merging and overwrite the files?

sfsholden: This is the text presented to the user when the project already exists. An overwrite does imply that we will be wiping out the previous logic. However, since this has behaved as a merge for quite some time it might be worth having an additional informative message that this logic has changed to a full overwrite. What do you think?
Screen Shot 2020-08-11 at 8 57 56 AM

packages/salesforcedx-vscode-core/src/messages/i18n.ts
lcampos: I think she's on PTO this week but we can always do a follow up PR once she's back

PR #2412, opened 2020-08-12
@W-7905048@ Refresh ignores sobjects ending with Share, History, Feed and Event

jag-sfdc:
### What does this PR do?
Refresh ignores sobjects ending with Share, History, Feed and Event

@W-7905048@

### Functionality Before
All sobjects definitions are fetched ( about 568)

### Functionality After
SObjects ending with Share, History, Feed and Event are filtered out while refreshing SObjects either at the project loading or when "Refresh SObject Definition" command is run (386 sobjects in all).

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -215,10 +215,8 @@ export class FauxClassGenerator {
 
   // VisibleForTesting
   public isRequiredSObject(sobject: string): boolean {
-    if (sobject.endsWith('Share') || sobject.endsWith('History') || sobject.endsWith('Feed') || sobject.endsWith('Event')) {
-      return false;
-    }
-    return true;
+    // Ignore all sobjects that end with Share or History or Feed or Event
+    return !/Share$|History$|Feed$|Event$/.test(sobject);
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -212,6 +213,14 @@ export class FauxClassGenerator {
     return this.successExit();
   }
 
+  // VisibleForTesting
+  public isRequiredSObject(sobject: string): boolean {
+    if (sobject.endsWith('Share') || sobject.endsWith('History') || sobject.endsWith('Feed') || sobject.endsWith('Event')) {
+      return false;
+    }
+    return true;
```
xyc, line 221:
```suggestion
    return !/Share$|History$|Feed$|Event$/.test(sobject);
```

jag-sfdc: I generally avoid regex - it makes readability harder and difficult to maintain. Any thoughts others?

jag-sfdc: @xyc - I've updated to the regex with some description about what we are doing. Let me know if it looks ok.

PR #2423, opened 2020-08-17
@W-7958729@ Feature gate for functions support

jag-sfdc:
### What does this PR do?
Adds a feature gate for functions related work

@W-7958729@ 

### Functionality Before
N/A

### Functionality After
Future functions related implementations will be exposed based on this flag.


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -64,5 +64,6 @@
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
   "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
-  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands"
-}
+  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands",
+  "experimental_support_functions": "Enable building and managing Salesforce functions from VSCode"
```
xyc, line 68: @ekapner

xyc: `VSCode` should be `VS Code`

ekapner: Yep, just adding a space is my only feedback

packages/salesforcedx-vscode-core/package.json
```diff
@@ -807,6 +807,11 @@
           "type": "boolean",
           "default": true,
           "description": "%experimental_apex_library_description%"
+        },
+        "salesforcedx-vscode-core.experimental.supportFunctions": {
```
lcampos, line 811: The only "problem" I see with defining the flag like this right now is that users will be able to see it and enable it. We do want them to do that once we have all the changes ready but for now we might want to "hide" this by just defining the flag as part of `sfdxCoreSettings.ts`. This is what we do with the `internal` flag.

https://github.com/forcedotcom/salesforcedx-vscode/blob/6e97314cbc8492de0defe220dadfa4417d4b5b3e/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts#L66

jag-sfdc:
I'm making the change @lcampos suggested and will be part of my next PR on salesforce functions.

PR #2428, opened 2020-08-20
Use @salesforce/templates library for template creation

xyc:
### What does this PR do?
Use `@salesforce/templates` library for creating templates. This functionality is behind the flag `salesforcedx-vscode-core.experimental.useTemplatesLibrary`

- Analytics
- Apex Class
- Apex Trigger
- Project

See remaining templates in #2437 (for an easier review)
- Lightning App
- Lightning Component
- Lightning Event
- Lightning Interface
- Lightning LWC
- Visualforce Component
- Visualforce Page

@W-7747009@

### Functionality Before
apex-class-cli

### Functionality After
We should cancel the progress bar :) task is finished before it knows
apex-class-library

Measurement:
```
ApexClassCreate Elapsed: 4205.001875013113 
ApexClassCreate Elapsed: 2854.296660989523 
ApexClassCreate Elapsed: 2824.861781001091 
ApexClassCreate Elapsed: 2846.4673389792442 
ApexClassCreate Elapsed: 2850.0102190077305

ApexClassCreate (Library) Elapsed: 601.2615059912205 
ApexClassCreate (Library) Elapsed: 35.476530998945236 
ApexClassCreate (Library) Elapsed: 24.221484005451202 
ApexClassCreate (Library) Elapsed: 26.848446995019913 
ApexClassCreate (Library) Elapsed: 24.462306022644043
```
- 600% faster first run
- 10148% faster subsequent runs in the same session (with the optimization for multiple runs in library API)

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -29,6 +29,7 @@
     "@salesforce/core": "2.5.1",
     "@salesforce/salesforcedx-utils-vscode": "49.6.0",
     "@salesforce/source-deploy-retrieve": "1.0.18",
+    "@salesforce/templates": "49.4.3",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
@@ -39,7 +40,6 @@
   },
   "devDependencies": {
     "@salesforce/salesforcedx-test-utils-vscode": "49.6.0",
-    "@salesforce/templates": "49.4.3",
     "@salesforce/ts-sinon": "^1.0.0",
     "@types/adm-zip": "^0.4.31",
     "@types/chai": "^4.0.0",
@@ -50,6 +50,7 @@
     "@types/shelljs": "^0.7.8",
     "@types/sinon": "^2.3.2",
     "@types/vscode": "1.40.0",
+    "@types/yeoman-assert": "^3.1.1",
     "chai": "^4.0.2",
     "cross-env": "5.2.0",
     "mocha": "^5",
@@ -58,7 +59,8 @@
     "mock-spawn": "0.2.6",
     "nyc": "^13",
     "sinon": "^7.3.1",
-    "typescript": "3.7.5"
+    "typescript": "3.7.5",
+    "yeoman-assert": "^3.1.1"
   },
   "scripts": {
     "vscode:prepublish": "npm prune --production",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -29,7 +29,7 @@
     "@salesforce/core": "2.5.1",
     "@salesforce/salesforcedx-utils-vscode": "49.6.0",
     "@salesforce/source-deploy-retrieve": "1.0.18",
-    "@salesforce/templates": "49.4.3",
+    "@salesforce/templates": "49.4.4",
     "adm-zip": "0.4.13",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -39,6 +39,7 @@
   },
   "devDependencies": {
     "@salesforce/salesforcedx-test-utils-vscode": "49.6.0",
+    "@salesforce/templates": "49.4.3",
     "@salesforce/ts-sinon": "^1.0.0",
     "@types/adm-zip": "^0.4.31",
     "@types/chai": "^4.0.0",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -813,7 +813,7 @@
         },
         "salesforcedx-vscode-core.experimental.useTemplatesLibrary": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "%experimental_templates_library_description%"
         }
       }
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -808,6 +808,11 @@
           "type": "boolean",
           "default": true,
           "description": "%experimental_apex_library_description%"
+        },
+        "salesforcedx-vscode-core.experimental.useTemplatesLibrary": {
+          "type": "boolean",
+          "default": false,
+          "description": "%experimental_templates_library_description%"
         }
       }
     },
--- a/packages/salesforcedx-vscode-core/package.nls.json
+++ b/packages/salesforcedx-vscode-core/package.nls.json
@@ -64,5 +64,6 @@
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
   "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
-  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands"
+  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands",
+  "experimental_templates_library_description": "Use Templates Library for Templates CLI commands"
 }
--- a/packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
@@ -5,6 +5,8 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import { ProjectOptions, TemplateType } from '@salesforce/templates';
+
 import {
   CliCommandExecutor,
   Command,
@@ -23,7 +25,9 @@ import * as vscode from 'vscode';
 import { channelService } from '../channels';
 import { nls } from '../messages';
 import { notificationService, ProgressNotification } from '../notifications';
+import { sfdxCoreSettings } from '../settings';
 import { taskViewService } from '../statuses';
+import { LibraryBaseTemplateCommand } from './templates/libraryBaseTemplateCommand';
 import {
   CompositeParametersGatherer,
   EmptyPreChecker,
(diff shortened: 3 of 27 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceApexClassCreate.test.ts
```diff
@@ -7,31 +7,122 @@
...
+      assert.file([apexClassPath, apexClassMetaPath]);
+      assert.fileContent(
+        apexClassPath,
+        'public with sharing class TestApexClass'
+      );
+      assert.fileContent(
+        apexClassMetaPath,
+        `<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>49.0</apiVersion>
```
xyc, line 122: We should also expose the default API version in library API, so that we don't have to update every release in tests.

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
```diff
@@ -0,0 +1,182 @@
...
+      });
+    }
+    if (result.error) {
+      channelService.appendLine(result.error.message);
+      notificationService.showFailedExecution(commandName);
+    }
+  };
+}
+
+export abstract class LibraryBaseTemplateCommand<T>
```
xyc, line 74: Base class for all templates commands

packages/salesforcedx-vscode-core/package.json
```diff
@@ -807,6 +810,11 @@
           "type": "boolean",
           "default": true,
           "description": "%experimental_apex_library_description%"
+        },
+        "salesforcedx-vscode-core.experimental.useTemplatesLibrary": {
+          "type": "boolean",
+          "default": false,
```
lcampos, line 816: Lets change this to default `true`

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -64,5 +64,6 @@
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
   "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
-  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands"
+  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands",
```
lcampos, line 67: cc @ekapner

packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
```diff
@@ -246,22 +290,28 @@ const parameterGatherer = new CompositeParametersGatherer(
...
 }
 
-const projectWithManifestCreateCommandlet = new SfdxCommandlet(
-  workspaceChecker,
-  parameterGatherer,
-  new ForceProjectCreateExecutor({ isProjectWithManifest: true }),
-  pathExistsChecker
-);
 export async function forceProjectWithManifestCreate() {
+  const createTemplateExecutor = sfdxCoreSettings.getTemplatesLibrary()
```
lcampos, line 307: I'm not sure what's happening but the project create commands fails whether the flag is on or off. I think it might be something close to here.

lcampos: The flow stops/fails when it should open the file explorer window so a user can pick where they want the project to be created.

xyc: Anything logged on the extension host output? I can create with flag on/off and with/without manifest.

lcampos: I didn't see any error on the output panel or the dev tools console. I didn't looked too close into the issue since I was going through the rest of the commands in this PR which they did work as expected.

I also did a clean install/compile of your branch so you might want to see if there's something in your cache that didn't make it into this PR. Or you can test with the vsix files built in CircleCI, I haven't test with those yet.

xyc: Not sure what's different about our setup. I've tested the vsix file (Core extension) in https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/1874/workflows/3afc06b0-006f-45f4-84ea-e8a2beb2ff63/jobs/11661/artifacts and creating standard/empty/analytics projects are fine with/without manifest.

lcampos: I just ran some tests with the vsix files and looks like I have some issue on my local setup since everything worked properly. Thanks for checking

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
```diff
@@ -0,0 +1,204 @@
...
+        const fileName = this.getOutputFileName(response.data);
+        telemetryService.sendCommandEvent(this.telemetryName, startTime, {
+          dirType: this.identifyDirType(result.outputDir),
+          commandExecutor: 'library'
+        });
+        await this.openCreatedTemplateInVSCode(result.outputDir, fileName);
+        return {
+          output: result.rawOutput
+        };
+      } catch (error) {
```
lcampos, line 118: Let's add telemetry here

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
```diff
@@ -0,0 +1,204 @@
...
+    const startTime = process.hrtime();
+    await wrapExecute(this.executionName, async () => {
+      try {
+        const templateOptions = this.constructTemplateOptions(response.data);
+        const result = await this.createTemplate(
+          this.templateType,
+          templateOptions
+        );
+        const fileName = this.getOutputFileName(response.data);
+        telemetryService.sendCommandEvent(this.telemetryName, startTime, {
```
lcampos, line 110: You can move everything using `result` inside of `wrapEecute`

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
```diff
@@ -0,0 +1,204 @@
...
+  public abstract get templateType(): TemplateType;
+  /**
+   * Construct template creation options from user input
+   * @param data data from Continue response
+   */
+  public abstract constructTemplateOptions(data: T): TemplateOptions;
+
+  public async execute(response: ContinueResponse<T>): Promise<void> {
+    const startTime = process.hrtime();
+    await wrapExecute(this.executionName, async () => {
```
lcampos, line 102: The idea here is that the call back function is the function in the library you want to execute instead of creating another callback function. That way the wrapExecute is generic enough that it can be used by several library calls.

```
await wrapExecute(this.createTemplate);
...
await this.createTemplate(this.templateType, templateOptions);
```

Here's how we're doing it for the retrieve command https://github.com/forcedotcom/salesforcedx-vscode/blob/810aba681b2cfabf55a9019ff7f4e6910f57ced1/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts#L127

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
```diff
@@ -0,0 +1,204 @@
...
+  output?: string;
+  error?: Error;
+}
+
+/**
+ * Warpping function exeuction with VS Code interface, such as Output channel and progress
+ * @param commandName command name
+ * @param fn function to execute
+ */
+function wrapExecute(
```
lcampos, line 45: We could move this to be part of `LibraryBaseTemplateCommand` and would make it easier reference some of it's data and functionality

xyc: The reason I have it as a separate function is I hope it could be reusable for other commands and not limited to `LibraryBaseTemplateCommand`

lcampos: Sure, but it's a bit early to know if that will happen since there's still more work to be done in the other libraries. It's better if we generalize the code once we have the other libraries ready. For now I want to make sure it's easy to maintain and finish the rest of the work for templates

xyc: Well I've extracted the function out precisely because I couldn't reuse similar `wrapExecute` function from other libraries. I'm not sure if it's still worth wrapping with similar functions if it's only used once. (vs. just inlining what's in the function).

The rest of the work for templates doesn't require more changes from `LibraryBaseTemplateCommand`, since they are all extending the base class, except for error telemetry.

xyc: For example `getLogsWrapper`, `retrieveWrapper` are only used once in apex/sdr.

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
xyc: The approach I go with templates each individual command/execute doesn't need to call `createTemplate` explicitly.  It's all done in `LibraryBaseTemplateCommand`. For generalization, `wrapExecute` is designed so that the function it wraps needs to return a raw output to get appended to the output channel, and doesn't require knowledge about business logic.

packages/salesforcedx-vscode-core/src/commands/templates/libraryBaseTemplateCommand.ts
lcampos: Got it, those libraries are still going to change which is why it might not be a good moment to generalize. 

If there's not much to re-use I think we can get rid of the wrapper function here and just keep everything inside of `execute`. It would also make the code easier to read.

xyc: Sounds good, if you don't think it can be used elsewhere for now, I can move what's inside the wrapper function into `execute`.

lcampos: I don't see that happening in the next weeks so it's safe to make the update.

xyc: Cool commit https://github.com/forcedotcom/salesforcedx-vscode/pull/2428/commits/5a6e09d49ff584df8aa098da06288bb9e3db4dc5 removes `wrapExecute`.

packages/salesforcedx-vscode-core/src/commands/templates/forceApexClassCreate.ts
```diff
@@ -22,8 +25,29 @@ import {
...
 import { OverwriteComponentPrompt } from '../util/postconditionCheckers';
 import { BaseTemplateCommand } from './baseTemplateCommand';
+import { LibraryBaseTemplateCommand } from './libraryBaseTemplateCommand';
 import { APEX_CLASS_DIRECTORY, APEX_CLASS_TYPE } from './metadataTypeConstants';
 
+export class LibraryForceApexClassCreateExecutor extends LibraryBaseTemplateCommand<
+  DirFileNameSelection
+> {
+  public executionName = nls.localize('force_apex_class_create_text');
+  public telemetryName = 'force_apex_class_create';
```
lcampos, line 35: @xyc last change, we need to use a different telemetry name for the library or we won't be able to compare events between library executions or normal cli ones.

xyc: That was my initial thought, but then I wasn't feeling like using a different name for every single template, so I added a custom dimension `commandExecutor` and set it to `library` for all library commands. https://github.com/forcedotcom/salesforcedx-vscode/pull/2428/files#diff-81cf0fdec768a9dede6cce86cb33ffc6R85

lcampos: cool, that should do it

PR #2432, opened 2020-08-24
Dehru/soql editor instance communication

dehru:
### What does this PR do?
This PR handles the back and forth communication between the soql text document and the custom editor ( soql builder ui ).  

TESTING: The salesforcedx-vscode-soql package is not setup yet for tests.  I will enter a separate WI (W-7999492) and communicate with the salesforcedx-vscode team to follow suggested patterns for tests and coverage ( looks like mocha is the preferred framework )

@W-7739435@

### Functionality Before
Minimal

### Functionality After
communication


packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -40,35 +45,32 @@ export class SOQLEditorProvider implements vscode.CustomTextEditorProvider {
...
-
-    html = html.replace('<!-- CSP TAG -->', cspMetaTag);
-    html = html.replace('./0.app.js', `${zeroDotScriptUri}`);
-    html = html.replace('./app.js', `${scriptUri}`);
-
+    html = EditorUtils.transformHtml(html, pathToLwcDist, webview);
     return html;
   }
+  private disposeInstance(instance: SOQLEditorInstance) {
+    const found = this.instances.findIndex(storedInstance => {
```
xyc, line 69: `WeakSet` or `WeakMap` might be a more suitable data structure for `instances`

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -40,35 +45,32 @@ export class SOQLEditorProvider implements vscode.CustomTextEditorProvider {
...
-      img-src ${webview.cspSource} https:;
-      script-src ${webview.cspSource};
-      style-src 'unsafe-inline' ${webview.cspSource};"
-    />`;
-
-    html = html.replace('<!-- CSP TAG -->', cspMetaTag);
-    html = html.replace('./0.app.js', `${zeroDotScriptUri}`);
-    html = html.replace('./app.js', `${scriptUri}`);
-
+    html = EditorUtils.transformHtml(html, pathToLwcDist, webview);
```
jonnyhork, line 65: Is there any way this could fail? If it were in a try catch is there a way we could gracefully fail? I wonder if we want some static Error webview of default page like a custom 404 error page.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -40,35 +45,32 @@ export class SOQLEditorProvider implements vscode.CustomTextEditorProvider {
...
-  }
+    const instance = new SOQLEditorInstance(document, webviewPanel, _token);
+    this.instances.push(instance);
+    instance.onDispose(this.disposeInstance.bind(this));
 
+    Array.prototype.push.apply(
+      this.context.subscriptions,
+      instance.subscriptions
+    );
+  }
```
jonnyhork, line 57: Its not clear to me why we need this? Why are we calling the prototype too?

xyc: It's for pushing everything in that array.
I'd recommend this tho for easier readability: `this.context.subscriptions.push(...instance.subscriptions)`

jonnyhork: Yeah, I assumed we are pushing into an array. I didn't understand why we need to call `apply` on the Array prototype. Why can't we use a simple `push()` on an array. I guess I wanted to know if there is some weirdness going on that required us to use `Array.prototype.push.apply`?

xyc: simple push pushes the array as an item, while `Array.prototype.push.apply` pushes each individual item

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
dehru: It would be disasterous, but i guess a bad dist directory could would be un-recoverable.     Let's keep it in mind, but won't implement it here.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
dehru: Oh yeah, cool, using the ...rest params feature.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -35,6 +37,7 @@
     "typescript": "3.7.5"
   },
   "scripts": {
+    "build": "yarn compile",
```
lcampos, line 40: The repo uses `npm` not `yarn`

packages/salesforcedx-vscode-soql/yarn.lock
```diff
@@ -0,0 +1,1063 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
lcampos, line 1: Remove this since the repo uses npm to build, release, etc

packages/salesforcedx-vscode-soql/package.json
lcampos: That's because lerna, vsce (vscode cli used for publishing) and yarn don't play well yet.

dehru: ok.  I'll adjust.

dehru: habit

PR #2433, opened 2020-08-24
Implement Commitizen for commit format enforcement

sfsholden:
### What does this PR do?
1. Adds task for initializing commitizen after installation
2. Adds task for running commit with commitizen
3. Adds commitizen as a dependency

@W-8001201@

### How do these changes affect me?
1. When submitting a commit, the commit format must now match `type: commitMessage`. Where type must be one of the following: feat, fix, improvement, docs, style, refactor, perf, test, build, ci, chore or revert
2. Do I have to use commitizen when committing? No. So long as the format and type are valid, you do not have to use commitizen.
3. When committing, if you'd like commitizen to help format the message these are all the ways it can be invoked: `git-cz`, `npm run commit`, or via the new VS Code Task `Commit`.


package.json
```diff
@@ -1,7 +1,10 @@
 {
   "devDependencies": {
     "@salesforce/dev-config": "1.4.1",
+    "@salesforce/dev-scripts": "0.3.14",
```
sfsholden, line 4: Needed for importing the husky hook used by the cli

sfsholden:
Please note this history is a bit of mess. One issue I have found with correcting it is that I'm not able to drop old commits. This is because of the linting and husky changes that were currently in flight at the time. Note that this will be fixed with the squashed commit.

package.json
```diff
@@ -1,7 +1,10 @@
...
     "ajv": "6.1.1",
+    "commitizen": "^3.0.5",
+    "cz-conventional-changelog": "^3.2.0",
```
sfsholden, line 7: Note that the default cz-commitizen does not have capabilities for some of the customizations I think we'll want. There is another project, cz-customizable, that I was experimenting with for enforcing scopes and different types. However, this is not approved by security. It sounds like the CLI team will be creating their own adaptor (off-set of cz-commitizen) at which point we might be able to add some of these customizations.

sfsholden: Also note, that for adding the work item as a required field, I think we might want to go down the route that the apex-jorje project took. I believe the enforcement of the rev and work item in the final squashed commit comes from a bot named [TOK_sfci12](https://git.soma.salesforce.com/tok-sfci12).

package.json
```diff
@@ -64,7 +69,13 @@
   },
   "husky": {
     "hooks": {
+      "commit-msg": "yarn sfdx-husky-commit-msg",
```
brpowell, line 72: Should this be `npm run` since the extensions don't use Yarn? And it doesn't look like `sfdx-husky-commit-msg` is a script in this package (or I might be misunderstanding how this runs).

sfsholden: `sfdx-husky-commit-msg` comes from the new import '@salesforce/dev-scripts'. And I will change it to npm run, thanks for the callout :)

package.json
```diff
@@ -69,7 +68,7 @@
   },
   "husky": {
     "hooks": {
-      "commit-msg": "yarn sfdx-husky-commit-msg",
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
```
sfsholden, line 71: Updated to use the commitlint command being called by `sfdx-husky-commit-msg` that is part of the dev-scripts project. Removed the import since we are no longer calling the script.

lcampos: I think we're missing adding a dependency since I'm getting `/bin/sh: commitlint: command not found` error when running a test commit on this branch.

package.json
```diff
@@ -25,6 +27,8 @@
     "postinstall": "lerna bootstrap --no-ci -- --no-package-lock && node scripts/reformat-with-prettier",
     "bootstrap": "lerna bootstrap --no-ci -- --no-package-lock && node scripts/reformat-with-prettier",
     "circleci:artifacts": "node scripts/download-vsix-from-circleci.js",
+    "commit-init": "commitizen init cz-conventional-changelog --save-dev --save-exact --force",
```
xyc, line 30: Is this necessary for someone cloning new project? If yes can we make it part of postinstall?

sfsholden: That is a good idea, I will add it.

package.json
sfsholden: Thanks for the catch, I didn't clean properly before testing it again

package.json
```diff
@@ -1,7 +1,11 @@
 {
   "devDependencies": {
+    "@commitlint/cli": "^7",
+    "@commitlint/config-conventional": "^7",
     "@salesforce/dev-config": "1.4.1",
     "ajv": "6.1.1",
+    "commitizen": "^3.0.5",
```
sfsholden, line 7: Versions are based on what the CLI is using:
    [commitlint/cli](https://github.com/forcedotcom/sfdx-core/blob/3ef407b667cdd65dc2bbd8259924e0b8b7ecb284/yarn.lock#L99) 
    [commitlint/config-conventional](https://github.com/forcedotcom/sfdx-core/blob/3ef407b667cdd65dc2bbd8259924e0b8b7ecb284/yarn.lock#L116)
    [commitizen](https://github.com/forcedotcom/sfdx-core/blob/3ef407b667cdd65dc2bbd8259924e0b8b7ecb284/yarn.lock#L1444)
    [cz-conventional-changelog](https://github.com/forcedotcom/sfdx-core/blob/3ef407b667cdd65dc2bbd8259924e0b8b7ecb284/yarn.lock#L1450)

package.json
sfsholden: I decided to undo the change to postinstall. It's not required and would need a dev to install commitizen prior to init. These steps have been added to the contributing doc.

CONTRIBUTING.md
```diff
@@ -13,6 +13,12 @@
 1. Send us a pull request when you are done. We'll review your code, suggest any
    needed changes, and merge it in.
 
+### Committing
```
sfsholden, line 16: @ekapner Please review once you are back from PTO! Thank you in advance.

package.json
xyc: We could do `npx commitizen ...` and dev doesn't need to install it, or add it to dev dependency. And why is this not required?

xyc: but I agree to the point that `postinstall` is not the best place, since it might be excessive every time we add a dependency. Git doesn't have a `post-clone`, but there's a way to simulate that. https://github.com/git-hook/post-clone. The goal is to have a setup as simple as possible, but let's not worry about it now

sfsholden: So it has been added as a dev dependency, but by default when you run `npm install` it is not installed. A user would need to use `npm install --dev`. 

This is not actually required because commitizen is meant more as a guide. Someone doesn't have to use it so long as their commit format matches "type: commit message", where type is a value that the husky hook expects.

xyc: Sounds good! thanks for the explanation @sfsholden .

sfsholden: @xyc You bet, thanks for the great questions

CONTRIBUTING.md
```diff
@@ -13,6 +13,12 @@
...
+
+1. We enforce commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `npm install -g commitizen` and running `npm run commit-init`. When you commit, we recommend that you use `npm run commit` which will prompt you with a series of questions to format the commit message. Or you can use our VS Code Task `Commit`.
```
ekapner, line 18: ...`npm run commit`, which prompts you with a series...

CONTRIBUTING.md
```diff
@@ -13,6 +13,12 @@
...
+1. The commit message format that we expect is: `type: commit message`. Valid types are: feat, fix, improvement, docs, style, refactor, perf, test, build, ci, chore and revert.
+1. Before commit and push, husky will run several hooks to ensure the commit message is in the correct format and that everything lints and compiles properly.
```
ekapner, line 20: Before commit and push, Husky runs several hooks to ensure...

CONTRIBUTING.md
```diff
@@ -43,15 +49,15 @@ Agreement. You can do so by going to https://cla.salesforce.com/sign-cla.
 
 ### Merging `develop` into `main`
 
-- When a development cycle finishes, the content of the `develop` branch will become the `master` branch
+- When a development cycle finishes, the content of the `develop` branch will become the `main` branch
```
ekapner, line 52: ...finishes, the content of the `develop` branch becomes the `main` branch. (end with period?)

PR #2435, opened 2020-08-24
@W-7958764@ Create function support

jag-sfdc:
### What does this PR do?
Adds support to create function from VS Code

@W-7958764@

### Functionality Before
N/A

### Functionality After
Aug-24-2020 14-29-49


packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -0,0 +1,88 @@
...
+      this.setFileExtension('js');
+    } else if (data.language === 'typescript') {
+      this.setFileExtension('ts');
+    }
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_function_create_text'))
+      .withArg('evergreen:function:create')
+      .withArg(data.fileName)
+      .withFlag('--language', data.language)
+      .withLogName('force_org_create_function')
```
xyc, line 44:
```suggestion
      .withLogName('force_create_function')
```

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -496,6 +502,14 @@ export async function activate(context: vscode.ExtensionContext) {
     return internalApi;
   }
 
+  // Set functions enabled context
+  const functionsEnabled = sfdxCoreSettings.getFunctionsEnabled();
```
xyc, line 506: If user updates the setting, it won't be applied unless restarted. We could use`workspace.onDidChangeConfiguration` to listen and updates the context

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (language === undefined) {
+      return { type: 'CANCEL' };
+    }
+    return {
+      type: 'CONTINUE',
+      data: {
+        fileName: name,
+        language,
+        outputdir: ''
```
xyc, line 71: will output dir be used? If not let's update `FunctionInfo` to just include `fileName` and `language`

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Good catch. Will change this.

packages/salesforcedx-vscode-core/src/index.ts
jag-sfdc: I think the purpose of this config is to deter users from turning it on. Once we have reached the beta stage, we will move this to package.json like the other experimental configs. So we are good with it as is I think.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Agree! that was my first thought too. But that required far more refactoring since every other template has outputdir and extends DirFileNameSelection (that has outputdir) and by not doing that, I wouldn't be able to use all the template related existing code. I can add a comment here why we are doing this here. Happy to discuss on hangout if there are other options.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -74,6 +74,7 @@ export const messages = {
     'forceide:// URL from Setup',
   parameter_gatherer_invalid_forceide_url:
     "The forceide:// URL is invalid. From your subscriber's org, copy and paste the forceide:// URL shown on the Apex Debugger page in Setup.",
+  parameter_gatherer_enter_function: 'Enter function details',
```
jag-sfdc, line 77: FYI @ekapner

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -64,5 +64,6 @@
   "force_diff_against_org": "SFDX: Diff File Against Org",
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
   "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
-  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands"
-}
+  "experimental_apex_library_description": "Use Apex Library for Apex CLI commands",
+  "force_function_create": "SFDX: Create Function"
```
lcampos, line 68: cc @ekapner

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
lcampos, line 2:
```suggestion
 * Copyright (c) 2020, salesforce.com, inc.
```

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
xyc: Yeah, I agree. Thanks for providing this background. It would be helpful to have a comment though. IMHO strategy pattern and the way our base class BaseTemplateCommand is structured are a bit overengineering for the purpose of processing output folder and file names - Let's not worry about it now.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -0,0 +1,88 @@
...
+import { nls } from '../../messages';
+import {
+  CompositeParametersGatherer,
+  SfdxCommandlet,
+  SfdxWorkspaceChecker
+} from '../util';
+import { BaseTemplateCommand } from './baseTemplateCommand';
+import { FUNCTION_TYPE } from './metadataTypeConstants';
+
+export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
```
lcampos, line 28: Are we going to deal with the plugin not being installed in this PR or that will be on a different one ?

jag-sfdc: I was considering this story as a happy path scenario. The pre-checks and post function creation tasks will be the next story.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -302,6 +304,7 @@ export const messages = {
   error_overwrite_prompt: 'Error checking workspace for existing components',
   error_no_scratch_def:
     'No scratch definition files found. These files must be in the "config" folder and have file names that end with "-scratch-def.json". See [Scratch Org Definition File](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm) for help.',
```
ekapner, line 306: No scratch org definition found. The file must reside in the "config" folder and end with "-scratch-def.json". See...

@jag-sfdc - any reason to use plural? What happens if you have mult scratch defs that end in -scratch-def.json?Does VS Code prompt you to select one?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -302,6 +304,7 @@ export const messages = {
...
+  error_function_error: 'Function creation error',
```
ekapner, line 307: Can we provide more details about how to get out of the jam?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: Yes VS Code prompts to select one. I can change this.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: I'm handling error handling as my next story. This message can be removed for now. The next story will have some pre checks and how to resolve them.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
lcampos: sounds good 👍

PR #2437, opened 2020-08-25
Use @salesforce/templates library (Remaining templates)

xyc:
### What does this PR do?
Use `@salesforce/templates` library for creating templates. Follow up of #2428
- Lightning App
- Lightning Component
- Lightning Event
- Lightning Interface
- Lightning LWC
- Visualforce Component
- Visualforce Page

@W-7747009@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
@@ -5,6 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import { LightningAppOptions, TemplateType } from '@salesforce/templates';
+import { LibraryBaseTemplateCommand } from './libraryBaseTemplateCommand';
+
 import {
   Command,
   SfdxCommandBuilder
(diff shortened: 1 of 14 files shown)


packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
```diff
@@ -64,27 +92,33 @@ const outputDirGatherer = new SelectOutputDir(AURA_DIRECTORY, true);
 const metadataTypeGatherer = new MetadataTypeGatherer(AURA_TYPE);
 
 export async function forceLightningAppCreate() {
+  const createTemplateExecutor = sfdxCoreSettings.getTemplatesLibrary()
```
jag-sfdc, line 95: This piece of code is repeating across all templates. Can we refactor this out and have a unit test for this? We could reduce how much we test in 'Library Create' tests in all templates. Right now we are testing the templates library itself instead of the interaction.

packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
```diff
@@ -34,6 +37,31 @@ import {
   AURA_TYPE
 } from './metadataTypeConstants';
 
+export class LibraryForceLightningAppCreateExecutor extends LibraryBaseTemplateCommand<
+  DirFileNameSelection
```
brpowell, line 41: How come a new base class is being used instead of BaseTemplateCommand?

packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
xyc: The 'Library Create' tests are actually adding a lot of code coverage (about +10% for each template classes). The previous tests are only covering commands construction, while the new tests are actually doing template creation and asserting that files are created, file's content etc. This is an unintended benefit but it adds 1% to the overall coverage in the vscode repo.

As for the experimental settings, I don't think there's much benefits in extracting them out at this moment, since the eventual goal is to remove the setting and it'll be a similar amount of work if it's in the form right now, or it's in the form of a small utility function.

packages/salesforcedx-vscode-core/src/commands/templates/forceLightningAppCreate.ts
xyc: This is a follow up of #2428 where [`LibraryBaseTemplateCommand`](https://github.com/forcedotcom/salesforcedx-vscode/pull/2428/files#diff-81cf0fdec768a9dede6cce86cb33ffc6) is defined. It's designed to substitute `BaseTemplateCommand` - so we can remove `BaseTemplateCommand` in the future. Please take a look at the original PR.

PR #2442, opened 2020-08-25
Make Refresh SObject Definitions more resilient when handling expired access tokens 

brpowell:
### What does this PR do?

Update the Refresh SObject Definitions command to use an sfdx-core connection when making the sObject describe request. A core connection will automatically take care of refreshing an access token, reducing the chances of the command failing due to authentication issues.

@W-7709580@


packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -13,7 +13,6 @@
     "@salesforce/salesforcedx-utils-vscode": "49.6.0",
     "cross-spawn": "6.0.5",
     "path-exists": "3.0.0",
-    "request-light": "0.2.4",
```
brpowell, line 16: Using `Connection.request` now

PR #2447, opened 2020-08-26
Replacing non-inclusive content from the repository (W-7953407)

LisandroFernandezSF:
Identify and replace non-inclusive content:
blacklist -> blocklist
whitelist -> allowlist

ID: W-7953407

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/package.nls.json
+++ b/packages/salesforcedx-vscode-apex-debugger/package.nls.json
@@ -1,7 +1,7 @@
 {
-  "user_id_filter_text": "A list of user IDs with requests allowlisted for debugging",
-  "request_type_filter_text": "A list of Apex request types allowlisted for debugging",
-  "entry_point_filter_text": "A regular expression matching entry points allowlisted for debugging",
+  "user_id_filter_text": "A list of user IDs with requests allowed for debugging",
+  "request_type_filter_text": "A list of Apex request types allowed for debugging",
+  "entry_point_filter_text": "A regular expression matching entry points allowed for debugging",
   "sfdx_project_text": "Location of SFDX project",
   "launch_snippet_label_text": "Apex Debugger: Launch",
   "launch_snippet_description_text": "A new configuration for launching Apex Debugger",
--- a/packages/salesforcedx-vscode-apex-debugger/package.nls.json
+++ b/packages/salesforcedx-vscode-apex-debugger/package.nls.json
@@ -1,7 +1,7 @@
 {
-  "user_id_filter_text": "A list of user IDs with requests whitelisted for debugging",
-  "request_type_filter_text": "A list of Apex request types whitelisted for debugging",
-  "entry_point_filter_text": "A regular expression matching entry points whitelisted for debugging",
+  "user_id_filter_text": "A list of user IDs with requests allowlisted for debugging",
+  "request_type_filter_text": "A list of Apex request types allowlisted for debugging",
+  "entry_point_filter_text": "A regular expression matching entry points allowlisted for debugging",
   "sfdx_project_text": "Location of SFDX project",
   "launch_snippet_label_text": "Apex Debugger: Launch",
   "launch_snippet_description_text": "A new configuration for launching Apex Debugger",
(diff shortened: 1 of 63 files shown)


packages/salesforcedx-vscode/CHANGELOG.md
```diff
@@ -64,7 +64,7 @@
 
 #### docs
 
-- Update whitelist to allowlist ([PR #2323](https://github.com/forcedotcom/salesforcedx-vscode/pull/2323))
+- Update allowlist to allowlist ([PR #2323](https://github.com/forcedotcom/salesforcedx-vscode/pull/2323))
```
xyc, line 67: This is a bit confusing

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
```diff
@@ -326,7 +326,7 @@ export const messages = {
   CompactLayout: 'Compact Layouts',
   ConnectedApp: 'Connected Apps',
   ContentAsset: 'Content Assets',
-  CorsWhitelistOrigin: 'Cors Whitelist Origins',
+  CorsAllowlistOrigin: 'Cors Allowlist Origins',
```
xyc, line 329: I think this is an metadata `xmlName`. Will this need to be changed in sdr lib too? https://github.com/forcedotcom/source-deploy-retrieve/blob/6eedb130caaec33de64111b99f4d7edcf880265e/src/metadata-registry/data/registry.json#L1174 @brpowell

rcoringrato-sfdc: @xyc @LisandroFernandezSF Agreed, I think this should remain as it was.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -340,7 +340,7 @@ export const messages = {
   CompactLayout: 'Compact Layouts',
   ConnectedApp: 'Connected Apps',
   ContentAsset: 'Content Assets',
-  CorsWhitelistOrigin: 'Cors Whitelist Origins',
+  CorsAllowlistOrigin: 'Cors Allowlist Origins',
   CspTrustedSite: 'Csp Trusted Sites',
```
rcoringrato-sfdc, line 344: @LisandroFernandezSF Since this is the name of an entity, and therefore part of the public api I think this change will need to happen at a later date.

packages/salesforcedx-vscode/CHANGELOG.md
rcoringrato-sfdc: @xyc Would it be better to rephrase as "replaced non-inclusive content"?

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
brpowell: Yup, should stay the same but still good find. On the documentation page for this type they say allowlist, but the type name itself stays for API reasons.

packages/salesforcedx-vscode/CHANGELOG.md
xyc: Yes that rephrasing would be good

LisandroFernandezSF: Done!

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
LisandroFernandezSF: I understand the property  `CorsWhitelistOrigin` should stay, but if I change only the value, like:  `CorsWhitelistOrigin: 'Cors Allowlist Origins'`, is that OK?

packages/salesforcedx-vscode-apex-debugger/package.nls.json
```diff
@@ -1,7 +1,7 @@
 {
-  "user_id_filter_text": "A list of user IDs with requests whitelisted for debugging",
-  "request_type_filter_text": "A list of Apex request types whitelisted for debugging",
-  "entry_point_filter_text": "A regular expression matching entry points whitelisted for debugging",
+  "user_id_filter_text": "A list of user IDs with requests allowlisted for debugging",
```
lcampos, line 2: We could replace the use of `allowlisted` with just `allowed`

LisandroFernandezSF: Done!

PR #2450, opened 2020-08-26
@W-7963183@ Deploy multiple components via SDR metadata api

AnanyaJha:
### What does this PR do?
This PR supports deploying multiple components by either deploying multiple filepaths or a folder with the deploy-retrieve library through the metadata api. There were also some version upgrades that required some (slightly unrelated) updates as well.

@W-7963183@


packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -71,6 +72,20 @@ export class MultipleSourcePathsGatherer implements ParametersGatherer<string> {
...
+    this.uris = uris;
+  }
+  public async gather(): Promise<ContinueResponse<string[]>> {
+    const sourcePaths = this.uris.map(uri => uri.fsPath);
+    return {
+      type: 'CONTINUE',
+      data: sourcePaths
+    };
+  }
+}
```
AnanyaJha, line 87: I haven't done it in this PR but we should explore having a single gatherer no matter the number of files being deployed

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -122,8 +142,21 @@ export class LibraryDeploySourcePathExecutor extends DeployRetrieveLibraryExecut
...
+        components = registryAccess.getComponentsFromPath(response.data);
+      } else {
+        const allComponents: SourceComponent[] = [];
+
+        for (const filepath of response.data) {
+          allComponents.push(...registryAccess.getComponentsFromPath(filepath));
+        }
+        components = allComponents.filter(
+          (cmp, index) => allComponents.indexOf(cmp) === index
+        );
```
AnanyaJha, line 157: re: earlier convo with bryan
we can look into including this logic at the registry level instead -> you can pass in multiple paths to the registry and it will create a unique list of all the components for you

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/libraryDeployResultParser.test.ts
```diff
@@ -151,102 +184,50 @@ describe('Deploy Parser', () => {
...
-            filePath: apexClassPath,
+            filePath: apexClassPathOne,
             message: 'Unexpected error happened during deploy'
           }
         ]
       }
     ]
   };
 
-  // it('should create array of success info for updated class', async () => {
```
AnanyaJha, line 167: these comments seem to be leftover from a previous merge, not sure if we kept them for something

lcampos: I'm not sure we've replaced these tests with new one and some of them seem valid

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: Agreed, we should open a bug/work item to add this to the library

brpowell:
```typescript
components = allComponents.filter(
    (cmp, index) => allComponents.indexOf(cmp) === index
);
```

I'm not sure I'm following how this dedupes the components. Don't we need to make some unique key based off of the type and fullname?

AnanyaJha: Yeah good catch, I realized I was testing with a shallow copy of an existing element from the array. I'll update this filter to specifically compare the type & full name of each object

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/libraryDeployResultParser.test.ts
AnanyaJha: I think these tests were replaced by the ones currently present- they test a level higher than `buildSuccesses` and `buildErrors` through the `resultParser` method

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
xyc: Could we use a better indexing mechanism? This has O(n**2) complexity, where n is number of components

AnanyaJha: updated!

PR #2457, opened 2020-08-27
Update Publishing to use Main & Update Docs

sfsholden:
### What does this PR do?
1. Update circleci configs to use main instead of master.
2. Update publishing script to use main.
3. Update some docs to use main instead of master.

### What PR does this fix or reference?
@W-8017900@

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 #
-# https://github.com/github/gitignore/blob/master/Node.gitignore
+# https://github.com/github/gitignore/blob/main/Node.gitignore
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 #
-# https://github.com/github/gitignore/blob/master/Node.gitignore
+# https://github.com/github/gitignore/blob/main/Node.gitignore
```
sfsholden, line 2: Not sure if this change is needed.

PR #2464, opened 2020-08-29
docs: quick fix doc fix

xyc:
### What does this PR do?
Document `salesforcedx-vscode-apex.enable-semantic-errors` setting as a prerequisite for Quick Fix section.

@W-8022975@

### DIFF
--- a/docs/_articles/en/apex/writing.md
+++ b/docs/_articles/en/apex/writing.md
@@ -82,7 +82,7 @@ Any syntax errors such as a missing semi-colon or a bracket are marked with a re
 
 ## Quick Fix
 
-When you reference a method that isn’t declared in your source, use the Quick Fix widget to automatically declare the method.
+When you reference a method that isn’t declared in your source, use the Quick Fix widget to automatically declare the method. To enable the Quick Fix widget, turn on the `salesforcedx-vscode-apex.enable-semantic-errors` setting.
 
 Declare Missing Methods quick fix can be invoked in the following ways:
 


docs/_articles/en/apex/writing.md
```diff
@@ -82,7 +82,7 @@ Any syntax errors such as a missing semi-colon or a bracket are marked with a re
 
 ## Quick Fix
 
-When you reference a method that isn’t declared in your source, use the Quick Fix widget to automatically declare the method.
+When you reference a method that isn’t declared in your source, use the Quick Fix widget to automatically declare the method. To enable the Quick Fix widget, turn on the `salesforcedx-vscode-apex.enable-semantic-errors` setting.
```
xyc, line 85: @ekapner could you take a look at this? thanks

ekapner: @xyc - I'm looking in Settings and don't see this. Is it a setting or do you need to install an extension?

lcampos: This setting is not ready to be used and shouldn't be documented. Why do we think we need it for quick fixes @xyc ?

xyc: @lcampos Quick fix features were not available for me without this setting on.

xyc: @ekapner this setting was part of Apex Extension

sfsholden: @xyc I just did a sanity check in my local with the setting disabled and could still get quick fix suggestions. Let me know if you'd like to sync in case this still isn't working for you with the setting set to false.

xyc: Thanks @sfsholden . Here's a quick summary of the offline discussion for reference:

With the setting disabled, there's no yellow squiggly line under the code with the quick fix code action, which makes it harder to discover the code action, but the code action still exists and it shows up when clicking or hovering over it.

The reason we don't want to recommend the semantic errors setting in documentation, is that the semantics errors provided are sometimes wrong.

Issue #2410, opened 2020-08-11
Apex extension High CPU -  Zulu Platform x64 Architecture


ReviverAce:
### Summary

Every time I open VS Code, CPU usage goes to 100%. Starts normally but as each second pasts Zulu proccess is raising the percentage of cpu use. I figured out that it happens because of the apex extension. It started after clicking some vs code communicate that appeared in bottom right corner. I think it was something with sObjects. There was button "enable" and after that it's working like this. I can end task in manager which stops the problem but then I'm not able to run tests (option is missing) from vs code. I tried to wait for the proccess to finish but it seems nothing else happens. 

### Steps To Reproduce:

1.  Open VS code
1.  Wait for CPU to run high

### Expected result

Works fine, cpu usage is normal, you can run tests after some time

### Actual result

cpu usage is 100%, you cannot run tests when ending proccess or waiting like 20 minutes.

### Additional information

_Feel free to attach a screenshot_.
![image](https://user-images.githubusercontent.com/46908367/89931713-dfc26a00-dc0c-11ea-9b27-375d74434d43.png)
![image](https://user-images.githubusercontent.com/46908367/89932610-311f2900-dc0e-11ea-9d5c-a62a5379c3ea.png)

**VS Code Version**:
1.47.3
**SFDX CLI Version**:
APEX 49.4.0 also checked on 49.2.0 but same reason

**OS and version**:
windows 10 education 1909

johnpanos:
I am also experiencing super high CPU usage. I can vouch for this bug. I'm posting here as well as the previous issue I commented on was closed.

![image](https://user-images.githubusercontent.com/25217527/89861150-ee037e00-db59-11ea-8c94-b02ebc179570.png)

_Originally posted by @johnpanos in https://github.com/forcedotcom/salesforcedx-vscode/issues/968#issuecomment-671737141_

romnus:
@johnpanos I've been dealing with this Java issue too. At first I disabled the Apex Language Server extension, which fixes the issue, but then you don't have any syntax highlighting and let's face it, we all need that.

So whenever I start VS Code or change workspaces (which is rare now because of this issue), I kill the Java process. It restarts, but if you kill it five times in less than three minutes, it won't restart. Then you'll have your syntax highlighting and your processing power, but you'll sacrifice other features like clicking the Run Test button to run tests directly from the code.

FAR FAR from ideal, but it's getting me through.

smaddox-sf:
[...]

PR #2467, opened 2020-09-01
fix: create minimal sobject refresh on activation

lcampos:
### What does this PR do?
New type of sobject refresh that gets the definition of a reduced set of SObjects. This is meant to be used through the activation message. Reducing the amount of sobject definitions helps avoid the performance spikes seen lately while still allowing the user to have a better apex lang server experience.

NOTE: i'm still working on updating the test but the changes are ready to start getting feedback

#2410 , @W-8027533@

### DIFF
--- a/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
+++ b/packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
@@ -31,14 +31,32 @@ import { nls } from '../messages';
 
 export const INDENT = '    ';
 const MODIFIER = 'global';
-
+const startupMinSObjects = [
+  'Account',
+  'Attachment',
+  'Case',
+  'Contact',
+  'Contract',
+  'Lead',
+  'Note',
+  'Opportunity',
+  'Order',
+  'Pricebook2',
+  'PricebookEntry',
+  'Product2',
+  'RecordType',
+  'Report',
+  'Task',
+  'User'
+];
 export interface CancellationToken {
   isCancellationRequested: boolean;
 }
 
 export enum SObjectRefreshSource {
   Manual = 'manual',
-  Startup = 'startup'
+  Startup = 'startup',
+  StartupMin = 'startupmin'
 }
 
 export interface FieldDeclaration {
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -221,6 +239,82 @@ export class FauxClassGenerator {
     return this.successExit();
   }
 
+  public async generateMin(
```
lcampos, line 242: I created this reduced method since we'll keep iterative over this approach as we get more feedback and didn't want to make `generate` harder to maintain.

packages/salesforcedx-vscode-apex/src/commands/forceGenerateFauxClasses.ts
```diff
@@ -212,44 +223,62 @@ export async function initSObjectDefinitions(projectPath: string) {
...
-    SOBJECTS_DIR
+    SOBJECTS_DIR,
+    STANDARDOBJECTS_DIR
   );
 }
 
 export async function checkSObjectsAndRefresh(projectPath: string) {
-  const hasDefaultUsernameSet =
-    await getDefaultUsernameOrAlias();
+  const hasDefaultUsernameSet = await getDefaultUsernameOrAlias();
```
lcampos, line 240: Most of the changes here simplify the previous flow.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -221,6 +239,82 @@ export class FauxClassGenerator {
...
+    while (j < startupMinSObjects.length) {
+      try {
+        if (
+          this.cancellationToken &&
+          this.cancellationToken.isCancellationRequested
+        ) {
+          return this.cancelExit();
+        }
+        fetchedSObjects = fetchedSObjects.concat(
+          await describe.describeSObjectBatch(startupMinSObjects, j)
```
xyc, line 289: I'm not quite getting how batching works here. In `describeSObjectBatch`, it calls `runRequest` which makes an xhr request, and this block is running `startupMinSObjects.length` times.
https://github.com/forcedotcom/salesforcedx-vscode/blob/1cfa48b0b45da0c47aaa0e466115042ce0f1026c/packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts#L284

brpowell: This hasn't changed from the previous implementation. The describe calls are in essence paginated with `j` tracking how many objects have been fetched. [This line right below](https://github.com/forcedotcom/salesforcedx-vscode/pull/2467/files#diff-b663fb174d801c41abb1ad0fa66fedc3R291) sets `j` to how many were fetched in the current "page"

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
brpowell: Makes sense. Although it seems like it'd be a relatively small change in the current generate to support this - is cloning the implementation any easier to maintain/less confusing?

lcampos: I think we'll keep iterating over this and I want to be able to do it fast enough without breaking the other commands that use the `generate` piece. If it turns out that the changes are not very specific then I'll definitely merge `generate` and `generateMin`.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
xyc: Got it, so `j` is always equal to the current count of `fetchedSObjects` here. I think from an API perspective this should all be part of `describeSObjectBatch`

xyc: Since it hasn't changed from original, fine keeping as is

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
xyc: @lcampos @brpowell Not specific to this PR - What do you think if we parallelize batch fetching? For example in my example org there are 461 standard sobjects, and batch comes in size of 25. There are 19 requests made sequentially.

Here's a piece of sample code in `generate`:
```
const filteredSObjects = sobjects.filter(this.isRequiredSObject);
    let j = 0;
    const requests = [];
    while (j < filteredSObjects.length) {
      requests.push(describe.describeSObjectBatch(filteredSObjects, j));
      j += 25;
    }

    try {
      if (
        this.cancellationToken &&
        this.cancellationToken.isCancellationRequested
      ) {
        return this.cancelExit();
      }
      const results = await Promise.all(requests);
      fetchedSObjects = ([] as SObject[]).concat(...results);
    } catch (errorMessage) {
      return this.errorExit(
        nls.localize('failure_in_sobject_describe_text', errorMessage)
      );
    }

    // tslint:disable-next-line:prefer-for-of
    for (let i = 0; i < fetchedSObjects.length; i++) {
      if (fetchedSObjects[i].custom) {
        customSObjects.push(fetchedSObjects[i]);
      } else {
        standardSObjects.push(fetchedSObjects[i]);
      }
    }
```

Measurement for parallelization tested on refreshing standard sobjects (3 runs): 
sequential: 26864.92 ms
parallelized: 3584.46 ms

brpowell: We should definitely look into changing this in another PR - I don't see any reason why this can't be parallelized

xyc: @brpowell There's concurrent API request limit enforced on requests with 20s or longer, which would hardly be a case here, but we should have some error handling around that. https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_api.htm

Discussed with @lcampos offline - we could make this enhancement on PWT's work on `@salesforce/core` which will replace this code

PR #2468, opened 2020-09-01
Added docs on soql and anonymous apex

ntotten:
### What does this PR do?
Adds docs for executing SOQL queries and anonymous apex.


docs/_articles/en/apex/writing.md
```diff
@@ -5,11 +5,11 @@ lang: en
 
 Use the Apex extension for VS Code to access code-editing features such as code completion, go to definition, view outlines of Apex classes and triggers, refactoring, and find syntactic errors in your code.
 
-## Code-Completion
+## Code Completion
```
ntotten, line 8: I changed these because it isn't normal to use them hyphenated even if it might be technically correct. See: https://code.visualstudio.com/docs/editor/intellisense

ekapner: Yeah, I think the dehyphenation is just fine.

PR #2470, opened 2020-09-01
feat: Pull dependencies after Function creation @W-7958774@

jag-sfdc:
### What does this PR do?
Optionally runs 'npm install' at the end of Functions creations

@W-7958774@

### Functionality Before
N/A

### Functionality After
Sep-01-2020 09-58-39


packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -72,6 +75,10 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
     taskViewService.addCommandExecution(execution, cancellationTokenSource);
   }
 
+  protected runPostCommandTasks(outputDir: string) {
```
jag-sfdc, line 78: I started to use baseTemplateCommand since I got everything I needed for Function creation. Now I see functions being more and more different - running post command task for example is only for functions. I've tried to add this empty method to base class to still use baseTemplateCommand. The other option is to not use baseTemplates and have execute and build implemented for functions separately.

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -60,6 +60,7 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
           this.getPathToSource(response.data.outputdir, response.data.fileName)
         );
         vscode.window.showTextDocument(document);
+        this.runPostCommandTasks(path.dirname(this.getPathToSource(response.data.outputdir, response.data.fileName)));
```
xyc, line 63: Could we use a variable for outputDir? it's the same one as line #60

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
xyc: I think this is fine for now. When it's moved to templates lib, we could probably move this to yeoman generator https://yeoman.io/authoring/dependencies.html

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -541,5 +541,7 @@ export const messages = {
     'Package ID should be a 15 or 18 character Id that starts with 04t',
   package_id_gatherer_placeholder: '04t...',
   force_function_enter_function: 'Enter a name for the function',
-  force_function_enter_language: 'Select a language for your function'
```
jag-sfdc, line 544: FYI @ekapner

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -66,5 +66,6 @@
   "experimental_deploy_retrieve_description": "Enable performance enhancements to the deploy/retrieve experience.",
   "experimental_apex_library_description": "Use Apex Library for Apex CLI commands",
   "force_function_create": "SFDX: Create Function",
-  "experimental_templates_library_description": "Use Templates Library for Templates CLI commands"
```
jag-sfdc, line 69: fyi @ekapner

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -66,5 +66,6 @@
...
+  "experimental_templates_library_description": "Use Templates Library for Templates CLI commands",
+  "functions_pull_dependencies": "Run 'npm install' to pull dependencies after creating a function"
```
ekapner, line 70: @danyusf - do you think we should say: Run npm install to pull dependencies from `package.json` after creating a function

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -541,5 +541,7 @@ export const messages = {
...
+  force_function_enter_language: 'Select a language for your function',
```
ekapner, line 544: @jag-sfdc - when this pops up will we provide a list of valid languages?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -541,5 +541,7 @@ export const messages = {
...
+  force_function_pull_dependencies_error: "Error pulling dependencies. Try running 'npm install' for the function yourself",
```
ekapner, line 545: @jag-sfdc - thinking this error could occur because npm isn't installed or no dependencies are defined in package.json. If so, would like to present both. cc: @danyusf

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -541,5 +541,7 @@ export const messages = {
...
+  force_function_should_run_post_install: 'Should pull dependencies after function creation'
```
ekapner, line 546: @jag-sfdc When does user see this message?

packages/salesforcedx-vscode-core/package.nls.json
danyusf: I guess it's usually assumed dependencies are pulled from the package.json of a Node project, although if you wanted to hedge bets on using something other than npm in the future (e.g. yarn), you could simply say "Pull dependencies from package.json"

packages/salesforcedx-vscode-core/src/messages/i18n.ts
danyusf: Will the error info be captured in the Output pane in VSCode? That should help determine the cause of the failure for most devs (e.g. "npm: command not found")

Note that npm is a requirement for developing Node functions (although this may or may not be true for the upcoming "Apex functions", which isn't in this extension's scope).

packages/salesforcedx-vscode-core/package.nls.json
ekapner: @danyusf - if I'm interpreting YOU and the dev guide, sounds like Node is required for JavaScript/TypeScript but something else might be required once we support additional languages?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: Yes, this shows a list of languages - javascript and typescript.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: This is when user goes to settings and want to turn off this feature of pulling dependencies after function creation.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: Yes, there is a way to surface the error (like "npm: command not found"). I'll make that change. So, can I change the text to "Error pulling dependencies - <actual error text here>" ?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: Good catch, we don't need this anymore. I'll remove this message.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: @ekapner - This how the error message would look now
Screen Shot 2020-09-03 at 9 24 33 AM

packages/salesforcedx-vscode-core/package.json
```diff
@@ -823,6 +823,11 @@
           "type": "boolean",
           "default": true,
           "description": "%experimental_templates_library_description%"
+        },
+        "salesforcedx-vscode-core.functions_pull_dependencies": {
```
lcampos, line 827: We should change the use of `pull` for `install` since that's the term used by NodeJS

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -44,6 +47,16 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
       .withLogName('force_create_function')
       .build();
   }
+
+  public runPostCommandTasks(outputDir: string) {
+    if (sfdxCoreSettings.getFunctionsPullDependencies()) {
+      cp.exec('npm install', { cwd: outputDir }, err => {
+        if (err) {
+          notificationService.showWarningMessage(nls.localize('force_function_pull_dependencies_error', err.message));
```
xyc, line 55: could we have a test case covering this scenario?

packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -85,6 +86,14 @@ export class SfdxCoreSettings {
     return this.getConfigValue(FUNCTIONS_ENABLED, false);
   }
 
+  public async updateFunctionsPullDependencies(value: boolean) {
+    await this.setConfigValue(FUNCTIONS_PULL_DEPENDENCIES, value);
+  }
+
+  public getFunctionsPullDependencies(): boolean {
+    return this.getConfigValue(FUNCTIONS_PULL_DEPENDENCIES, true);
```
xyc, line 94: Not in the scope of this PR - but the abstractions this file provides seem to be very limited.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -44,6 +47,16 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
       .withLogName('force_create_function')
       .build();
   }
+
+  public runPostCommandTasks(outputDir: string) {
+    if (sfdxCoreSettings.getFunctionsPullDependencies()) {
+      cp.exec('npm install', { cwd: outputDir }, err => {
```
lcampos, line 53: We should also handle the scenario where NodeJS is not present to show the correct error message

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -44,6 +47,16 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
       .withLogName('force_create_function')
       .build();
   }
+
+  public runPostCommandTasks(outputDir: string) {
```
xyc, line 51: The outputDir here could potentially different from the outputDir input by the user (`response.data.outputDir`), it's the directory of the file that VS Code is going to open. Could we add a comment to clarify?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: What do you think, @danyusf? Sounds like we may swap out "pull" for "install" - based on feedback from Luis.

packages/salesforcedx-vscode-core/package.json
ekapner: @lcampos - do you think we should use "install" instead of "pull" in the error msg text?

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -33,3 +33,4 @@ export const TELEMETRY_ENABLED = 'telemetry.enabled';
 export const USE_APEX_LIBRARY = 'experimental.useApexLibrary';
 export const FUNCTIONS_ENABLED = 'experimental.supportFunctions';
 export const USE_TEMPLATES_LIBRARY = 'experimental.useTemplatesLibrary';
+export const FUNCTIONS_PULL_DEPENDENCIES = 'functions-pull-dependencies';
```
lcampos, line 36: Let's have this be `experimental.functions...` since the other functions flag is also has the `experimental` prefix

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
lcampos: Why can't we accomplish this by using a `PostconditionChecker` ?

packages/salesforcedx-vscode-core/package.json
lcampos: Yes, `pull` is a git term and NodeJS uses `install`. Everything related to NodeJS Functions should use `install` if we have it using `pull`

packages/salesforcedx-vscode-core/src/messages/i18n.ts
danyusf: @ekapner yes, let's use install

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
jag-sfdc: OverwriteComponentPrompt and ConflicDetection should kick in before the command is executed, but in this case we want the task to executed after the command execution is fully done.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: I've changed it to targetDir to avoid that confusion.

lcampos (approved):

PR #2488, opened 2020-09-04
SOQL SObject Integration

jgellin-sf:
### What does this PR do?
This PR establishes a messaging protocol between the VSCode extension and the SOQL builder iframe for the UI to get retrieve a list of SObjects and SObject metadata.

#@W-7977055@

### Functionality Before
No SObject integration.

### Functionality After
UI can post `sobjects_request` to retrieve a list of sobject names and `sobject_metadata_request` to retrieve the metadata for a particular sobject.


packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -16,29 +26,35 @@ interface SoqlEditorEvent {
 export enum MessageType {
   ACTIVATED = 'activated',
   QUERY = 'query',
+  SOBJECT_METADATA_REQUEST = 'sobject_metadata_request',
+  SOBJECT_METADATA_RESPONSE = 'sobject_metadata_response',
+  SOBJECTS_REQUEST = 'sobjects_request',
+  SOBJECTS_RESPONSE = 'sobjects_response',
```
jgellin-sf, line 32: Protocol for UI to request list of SObject names or particular SObject metadata from VSCode extension.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -16,29 +26,35 @@ interface SoqlEditorEvent {
...
   UPDATE = 'update'
 }
 
 export class SOQLEditorInstance {
```
jgellin-sf, line 36: Changed privacy scope of most members from `private` to `protected` for visibility to tests via subclassing.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -101,15 +143,60 @@ export class SOQLEditorInstance {
...
+          break;
+        }
+        case MessageType.SOBJECTS_REQUEST: {
+          this.retrieveSObjects().catch(() => {
+            channelService.appendLine(
+              `An error occurred while handling a request for object names.`
+            );
+          });
+          break;
+        }
```
jgellin-sf, line 163: These are asynchronous calls. A nested web service request may be necessary so we don't `await`. Otherwise may block unnecessarily.

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/soqlEditorInstance.test.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+class TestSoqlEditorInstance extends SOQLEditorInstance {
+  public sendEvent(event: any) {
+    this.onDidRecieveMessageHandler({
+      type: event.type,
+      message: event.message
+    });
+  }
+}
```
jgellin-sf, line 157: This subclass allows us to simulate message passing from the UI with minimal setup.

jgellin-sf:
For the time being errors are handled by simply writing to the `channelService` from `salesforcedx-vscode-core`.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dobladez: Not for this PR:  How about moving `MessageType` and `SoqlEditorEvent` to `soql-tooling` ? We need them on that side, and vscode-soql depends on soql-tooling.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -70,28 +90,50 @@ export class SOQLEditorInstance {
...
     };
   }
 
-  private createDocumentChangeHandler(document: vscode.TextDocument) {
-    return (e: vscode.TextDocumentChangeEvent) => {
+  protected createSObjectsUpdater(
+    webview: vscode.Webview
+  ): (sobjectNames: string[]) => void {
+    return (sobjectNames: string[]) => {
+      webview.postMessage({
```
xyc, line 109: Not for this PR, just something we can take in mind when designing webview applications. We could consider RPC mechanism for abstracting over `postMessage`, so that communication between extension and webview is over simple method calls.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -19,28 +19,45 @@
...
+    "cross-env": "5.2.0",
     "eslint": "6.0.1",
     "eslint-plugin-header": "^3.0.0",
     "eslint-plugin-jsdoc": "^30.0.3",
     "eslint-plugin-prettier": "^3.1.4",
+    "mocha": "^5",
+    "sinon": "^7.3.1",
     "typescript": "3.7.5"
   },
+  "extensionDependencies": [
```
lcampos, line 50: This means a user won't be able to run this extension without having `salesforcedx-vscode-core` installed. I thought we wanted the soql extension to run independent to everything else.

praksb: Good catch. @jgellin-sf you think we can get remove this extensionDependency?

praksb: It looks like it is needed for OrgAuthInfo and ChannelService. So if we do take this dependency then when someone installs the SOQL extension then salesforcedx-vscode-core would automatically be also installed if it is not present right? That sounds ok to me. Is there any downside to that?

jgellin-sf: We did talk about running this extension independent to the others. On the other hand, we probably want to do things the same way in these various extensions.

The only downside to taking this dependency is that if the user only wants SOQL tooling and doesn't want all of the commands, etc. in salesforcedx-vscode-core. 

This code uses OrgAuthInfo, channelService, and telemetryService from the core extension. I can duplicate or work around these functions if I need to. OrgAuthInfo is a utility class that (maybe?) could be duplicated or moved to it's own library and imported separately. The services are singletons so they don't really work as libraries imported separately. Working around those means doing things in a different way to the rest of our extensions.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dobladez: Good idea. Do we have a pattern/library already in use?  How about https://github.com/SAP/vscode-webview-rpc-lib ?

We might still need an async channel. I think "RPC"-style calls might not cover all use cases.

packages/salesforcedx-vscode-soql/package.json
praksb: @jgellin-sf It should be ok to take this dependency. I wouldn't do the extra work now to do any additional refactoring. We need those commands from core to provide org authentication anyway.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
jgellin-sf: Yeah I'm right with you on the fence on this one. These async calls actually have try-catch blocks in the called function, and these catch blocks are mostly called to satisfy the linter.

packages/salesforcedx-vscode-soql/package.json
xyc: > The only downside to taking this dependency is that if the user only wants SOQL tooling and doesn't want all of the commands, etc. in salesforcedx-vscode-core.

I'd add that if this extension depends on the core extension, it would add an additional time for extension activation, since the extension cannot be activated until all dependent extensions are activated. (On average my measurement indicates that core extension takes 2 seconds for activation on macOS)

jonnyhork: @xyc Does the activation of a one extension trigger the activation of the dependent extensions?

Would that also mean that we cannot use the SOQL extension unless there is an sfdx-project.json in the root directory, because the core ext would not be activted?

xyc: > Does the activation of a one extension trigger the activation of the dependent extensions?

Yes, the dependent extensions will be triggered.

> Would that also mean that we cannot use the SOQL extension unless there is an sfdx-project.json in the root directory, because the core ext would not be activted?

If `activate` function of core extension returns without throwing, it would be considered as successfully activated. I think last time there were some issues activating core extension in non-sfdx project context when `push-or-deploy-on-save.enabled` setting is enabled. It's worth checking out the status now if we are going with this approach.

lcampos: We are still sorting out issues when other extensions force `salesforcedx-vscode-core` to activate outside a salesforce project. If we need the soql extension to activate outside of a salesforce project (sfdx-project.json being present) please do not add a dependency to `salesforcedx-vscode-core`

packages/salesforcedx-vscode-soql/src/index.ts
```diff
@@ -8,12 +8,38 @@
...
   startLanguageClient(context);
+  startTelemetry(extensionHRStart);
 }
 
 export function deactivate(): Thenable<void> | undefined {
+  stopTelemetry();
   return stopLanguageClient();
 }
+
+function startTelemetry(hrstart: [number, number]): void {
```
lcampos, line 25: This could be moved to the telemetry service code

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -19,28 +19,45 @@
     "Programming Languages"
   ],
   "dependencies": {
+    "@salesforce/sobject-metadata": "0.1.4",
```
lcampos, line 22: I thought we agreed on having this be part of `@salesforce/core`, did something change ?

packages/salesforcedx-vscode-soql/package.json
praksb: It is ok to have the sfdx-project.json restriction for now for Beta. The CodeBuilder launch experience for a SOQL only user is not clearly defined yet, so it would be premature to optimize for that now. Even then since we control the launch we could definitely create a sfdx-project.json if needed for the user.

For desktop user this is not much of a change since the current SOQL experience requires an sfdx-project.json.

packages/salesforcedx-vscode-soql/package.json
jgellin-sf: Nothing changed, it’s just not there yet.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -19,28 +19,45 @@
...
     "@salesforce/soql-builder-ui": "0.0.2",
     "@salesforce/soql-language-server": "0.2.1",
     "debounce": "^1.2.0",
+    "vscode-extension-telemetry": "0.1.6",
     "vscode-languageclient": "^6.1.3"
   },
   "devDependencies": {
+    "@salesforce/salesforcedx-test-utils-vscode": "*",
```
dehru, line 30: @lcampos - our PRs broke because we didn't pick up the bump change to 49.9.0 that happened in develop.  Would it be helpful to just (star) these and control the version elsewhere in a single place, like in salesforcedx-vscode-core?

lcampos: No, just rebase the PR to the latest changes in develop. We need to keep this around or lerna won't be able to create the symlinks in development or run the build tasks in circleci

packages/salesforcedx-vscode-soql/package.json
lcampos: sounds good 👍

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
xyc: @dobladez It might be a bit late to make this change for soql builder. RPC should support async calls too. I don't think we currently have this pattern now. The lib you mentioned seems a good choice. I've also built upon https://github.com/GoogleChromeLabs/comlink in the past.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -16,29 +26,35 @@ interface SoqlEditorEvent {
...
   UPDATE = 'update'
 }
 
 export class SOQLEditorInstance {
   // handlers assigned in constructor
-  private updateWebview: () => void;
-  private onDidRecieveMessageHandler: (e: SoqlEditorEvent) => void;
-  private onTextDocumentChangeHandler: (
+  protected updateWebview: () => void;
+  protected onDidRecieveMessageHandler: (e: SoqlEditorEvent) => void;
```
praksb, line 39: spelling error onDidRec**_ei_**veMessageHandler instead.

dehru: i before e except after c.
( english sucks )

dehru (approved):
We are going to merge this in now.  There are 2 other PRs stuck behind this.  I think all items were either addressed or set aside as future refactoring.

xyc (approved):

PR #2496, opened 2020-09-08
fix: function command palette changes

xyc:
### What does this PR do?
Adding function start command to command palette; Remove function invoke command from command palette.

@W-7958829@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -521,6 +521,14 @@
           "when": "sfdx:functions_enabled"
         },
         {
+          "command": "sfdx.force.function.start",
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
+        },
+        {
+          "command": "sfdx.force.function.invoke",
+          "when": "false"
+        },
+        {
           "command": "sfdx.force.function.stop",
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -520,6 +520,14 @@
           "command": "sfdx.force.function.create",
           "when": "sfdx:functions_enabled"
         },
+        {
+          "command": "sfdx.force.function.start",
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
```
lcampos, line 525: Are we trying to limit this command in the command palette to only be available if the user has a functions file already open ?

xyc: Yes, that would be the limitation for now. We could expand it to be anywhere, but we'd need to implement a picker since there could be multiple functions.

lcampos: Sounds good to me, why don't we want to do the same for Invoke and Stop ?

xyc: For invoke, there's already CodeLens where user can invoke the command. If we wish, we could add it back to command palette when user has json files open.
For stop, there can be at most one container running at this moment, so there's no ambiguity.

lcampos (approved):

PR #2498, opened 2020-09-09
Function start/stop commands

xyc:
### What does this PR do?
Start / Stop functions locally

@W-7958829@

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -158,6 +158,10 @@
       ],
       "editor/context": [
         {
+          "command": "sfdx.force.function.start",
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
+        },
+        {
           "command": "sfdx.force.source.retrieve.current.source.file",
           "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking && resourceLangId != 'forcesourcemanifest'"
         },
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -159,7 +159,7 @@
       "editor/context": [
         {
           "command": "sfdx.force.function.start",
-          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
         },
         {
           "command": "sfdx.force.source.retrieve.current.source.file",
@@ -289,7 +289,7 @@
         },
         {
           "command": "sfdx.force.function.start",
-          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
         }
       ],
       "commandPalette": [
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -36,6 +36,7 @@
     "rxjs": "^5.4.1",
     "sanitize-filename": "^1.6.1",
     "shelljs": "0.8.3",
+    "strip-ansi": "^5.2.0",
     "xmldom-sfdx-encoding": "0.1.29"
   },
   "devDependencies": {
--- a/packages/salesforcedx-vscode-core/src/channels/channelService.ts
+++ b/packages/salesforcedx-vscode-core/src/channels/channelService.ts
@@ -33,16 +33,6 @@ export class ChannelService {
   public streamCommandOutput(execution: CommandExecution) {
     this.streamCommandStartStop(execution);
     execution.stderrSubject.subscribe(data =>
-      this.channel.append(data.toString())
-    );
-    execution.stdoutSubject.subscribe(data =>
-      this.channel.append(data.toString())
-    );
-  }
-
-  public streamCommandOutputWithoutColor(execution: CommandExecution) {
-    this.streamCommandStartStop(execution);
-    execution.stderrSubject.subscribe(data =>
       this.channel.append(stripAnsi(data.toString()))
     );
     execution.stdoutSubject.subscribe(data =>
--- a/packages/salesforcedx-vscode-core/src/channels/channelService.ts
+++ b/packages/salesforcedx-vscode-core/src/channels/channelService.ts
@@ -5,6 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import stripAnsi from 'strip-ansi';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
 
@@ -36,6 +37,16 @@ export class ChannelService {
     );
     execution.stdoutSubject.subscribe(data =>
       this.channel.append(data.toString())
+    );
+  }
+
+  public streamCommandOutputWithoutColor(execution: CommandExecution) {
+    this.streamCommandStartStop(execution);
+    execution.stderrSubject.subscribe(data =>
+      this.channel.append(stripAnsi(data.toString()))
+    );
+    execution.stdoutSubject.subscribe(data =>
+      this.channel.append(stripAnsi(data.toString()))
     );
   }
 
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
@@ -27,7 +27,7 @@ import {
   SfdxWorkspaceChecker
 } from '../util';
 
-import { Uri } from 'vscode';
+import { Uri, window } from 'vscode';
 import { FunctionService } from './functionService';
 
 export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
@@ -130,6 +130,17 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
  * @param sourceUri
  */
 export async function forceFunctionStart(sourceUri: Uri) {
+  if (!sourceUri) {
+    // Try to start function from current active editor, if running SFDX: start function from command palette
+    sourceUri = window.activeTextEditor?.document.uri!;
+  }
+  if (!sourceUri) {
+    notificationService.showWarningMessage(
+      nls.localize('force_function_start_warning_not_in_function_folder')
+    );
+    return;
+  }
+
   const commandlet = new SfdxCommandlet(
     new SfdxWorkspaceChecker(),
     new FilePathGatherer(sourceUri),
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
@@ -10,6 +10,8 @@ import {
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
 import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import * as fs from 'fs';
+import * as path from 'path';
 import { Observable } from 'rxjs/Observable';
 import { Subject } from 'rxjs/Subject';
 import * as vscode from 'vscode';
@@ -17,6 +19,7 @@ import { channelService } from '../../channels';
 import { nls } from '../../messages';
 import { notificationService, ProgressNotification } from '../../notifications';
 import { taskViewService } from '../../statuses';
+import { telemetryService } from '../../telemetry';
 import {
   FilePathGatherer,
   SfdxCommandlet,
@@ -27,9 +30,9 @@ import {
 import { Uri } from 'vscode';
 import { FunctionService } from './functionService';
 
-export class ForceFunctionStart extends SfdxCommandletExecutor<string> {
-  public build(sourceFsPath: string): Command {
-    this.executionCwd = sourceFsPath;
+export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
+  public build(functionDirPath: string): Command {
+    this.executionCwd = functionDirPath;
     return new SfdxCommandBuilder()
       .withDescription(nls.localize('force_function_start_text'))
       .withArg('evergreen:function:start')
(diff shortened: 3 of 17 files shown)


packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -5,40 +5,150 @@
...
 }
 
 /**
  * Executes sfdx evergreen:function:start --verbose
  * @param sourceUri
  */
-export async function forceFunctionStart(sourceUri: Uri) {
+export async function forceFunctionStart(sourceUri?: Uri) {
+  if (!sourceUri) {
+    // Try to start function from current active editor, if running SFDX: start function from command palette
```
xyc, line 134: We'll start function from the current active text editor if running "SFDX: Start Function" from command palette. We could potentially add a picker for multiple functions later

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -138,8 +138,14 @@ export const messages = {
   force_lightning_interface_create_text: 'SFDX: Create Aura Interface',
   force_function_create_text: 'SFDX: Create Function',
   force_function_start_text: 'SFDX: Start Function',
+  force_function_start_warning_no_toml:
+    'No function.toml found. Create function.toml or intialize a new function folder using SFDX: Create Function.',
```
xyc, line 142: This is an edge case that only happens when you have created a function folder, deleted function.toml and try to start function there

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -138,8 +138,14 @@ export const messages = {
...
+  force_function_start_warning_not_in_function_folder:
```
xyc, line 143: This happens when you run "SFDX: Start Function" in command palette, but haven't open a file function folder. See the comments in https://github.com/forcedotcom/salesforcedx-vscode/pull/2498/files#r488314698

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -138,8 +138,14 @@ export const messages = {
   force_lightning_interface_create_text: 'SFDX: Create Aura Interface',
   force_function_create_text: 'SFDX: Create Function',
   force_function_start_text: 'SFDX: Start Function',
+  force_function_start_warning_no_toml:
```
xyc, line 141: @ekapner Please take a look at the text here

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: You can also start function from right click in a file (for example `index.ts`), or right click in the function folder or file in explorer view

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -5,40 +5,150 @@
...
+        progress.complete();
+        taskViewService.removeTask(task);
+        notificationService
+          .showSuccessfulExecution(executionName)
+          .catch(() => {});
+        this.logMetric(execution.command.logName, startTime);
+      }
+    });
+
+    execution.processExitSubject.subscribe(async exitCode => {});
```
xyc, line 119: We'll address this in the prerequisites story

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -5,40 +5,150 @@
...
  * @param sourceUri
  */
-export async function forceFunctionStart(sourceUri: Uri) {
+export async function forceFunctionStart(sourceUri?: Uri) {
+  if (!sourceUri) {
+    // Try to start function from current active editor, if running SFDX: start function from command palette
+    sourceUri = window.activeTextEditor?.document.uri!;
+  }
+  if (!sourceUri) {
+    notificationService.showWarningMessage(
```
jag-sfdc, line 138: Would we hit this at all? Since the command palate itself shows up only for files in functions folder.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -5,40 +5,150 @@
...
+   * Locate the directory that has function.toml
+   * @param sourceFsPath path to start function from
+   */
+  public static getFunctionDir(sourceFsPath: string) {
+    let current = fs.lstatSync(sourceFsPath).isDirectory()
+      ? sourceFsPath
+      : path.dirname(sourceFsPath);
+    const { root } = path.parse(sourceFsPath);
+    while (current !== root) {
+      const tomlPath = path.join(current, 'function.toml');
```
jag-sfdc, line 54: If a customer has custom directories within function and wants to start function from there, we should still be allowing that? In that case, we need to recursively search for the function's root directory. Is this something we take care now?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
jag-sfdc: Never mind, your comment below explained it. So command palate does show up when files are not open, it is the right-click start function that opens up only for files under functions.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStop.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  it('Should stop function, show notification and send telemetry', async () => {
+    const srcUri = Uri.file(
+      path.join(getRootWorkspacePath(), 'functions', 'demoJavaScriptFunction')
+    );
+    const executor = new ForceFunctionStartExecutor();
+    const mockExecution = new MockExecution(executor.build(srcUri.fsPath));
+    const killExecutionStub = sandbox.stub(mockExecution, 'killExecution');
+    cliCommandExecutorStub.returns(mockExecution);
+    await forceFunctionStart(srcUri);
```
jag-sfdc, line 77: Why don't we just stub FunctionService instead?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+/**
+ * A running task that can be terminated
+ */
+interface Terminable {
+  terminate: () => Promise<void>;
+}
+
+export class FunctionService {
```
jag-sfdc, line 15: Can we have a test for this class?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: Yes, it looks up by `path.dirname` until it couldn't find it

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
xyc: FunctionService is just an implementation detail for stopping the server, and is suspect to change when we employed VS Code tasks later. All methods here are already covered. (the diff coverage is 95.23%)

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStop.test.ts
xyc: Same reason as https://github.com/forcedotcom/salesforcedx-vscode/pull/2498#discussion_r488852960. We'd like to test the actual behavior of starting/stopping container over how we implement them.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: I'll add some more comments to the jsdocs here

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: Yes, command palette shows up, but we don't know what function to start. That's where this message is shown. And this is where I refer to when/if we want to show a picker for functions

xyc:
@jag-sfdc https://github.com/forcedotcom/salesforcedx-vscode/pull/2498/commits/6447f60ecc4d9a46e66d9e6378ea0375e0417718 this commit updates invoke tests to use the actual mock function folder here

packages/salesforcedx-vscode-core/src/channels/channelService.ts
```diff
@@ -39,6 +40,16 @@ export class ChannelService {
     );
   }
 
+  public streamCommandOutputWithoutColor(execution: CommandExecution) {
```
lcampos, line 43: These changes should probably be done to `streamCommandOutput` and have it work the same way for every command that outputs using the channel service. Does adding `stripAnsi` to `streamCommandOutput` break other commands ?

xyc: I haven't tried that out because I think adding it as an additional method is less risky since a lot of commands use `streamCommandOutput`. I could add it to `streamCommandOutput` if you like.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -156,6 +157,10 @@
         }
       ],
       "editor/context": [
+        {
+          "command": "sfdx.force.function.start",
+          "when": "sfdx:functions_enabled && sfdx:project_opened && resource =~ /.*/functions/[^/]+(/[^/]+\\.(ts|js|json|toml))?$/"
```
lcampos, line 162: I think this when clause needs to be tweaked a bit more. It currently shows the command if you right click on the top level `functions` folder, ideally it should only be shown on the child folders that represent an individual function.

We should try and limit the command to be shown when right clicking only on those file that do represent the actual function code (index, .toml, etc). It is currently shown when right-clicking on files like package-lock.json or tsconfig.json

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
lcampos: We need to change this code to stop iterating when it reaches the project's `functions` directory or the root of the project. Right now it will iterate over all the folders in the file path initially provided if it doesn't find a `function.toml`.

For example, if I trigger this from a function with no `.toml` or by right clicking the `functions` folder the path could look like this `'/Users/myuser/Documents/myDXProject/functions/littleBot'.
This code will keep iterating until reaching `Users` in the scenario described here.

lcampos (changes requested):
Left a couple of comments, the rest of the code looks good.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
lcampos: I think it should be safe to add it since the other commands process cli output for things like deploy, queries, etc which do not contain ansi characters unless explicitly asked for. @AnanyaJha might know more about this since she had to add it to one of the rewritten Apex commands.

packages/salesforcedx-vscode-core/package.json
xyc: > It currently shows the command if you right click on the top level functions folder, ideally it should only be shown on the child folders that represent an individual function.

It doesn't show up in the top level folder
Screen Shot 2020-09-15 at 6 03 21 PM

>We should try and limit the command to be shown when right clicking only on those file that do represent the actual function code (index, .toml, etc). It is currently shown when right-clicking on files like package-lock.json or tsconfig.json

I added some limits in https://github.com/forcedotcom/salesforcedx-vscode/pull/2498/commits/49e287e2380861c76b66118964ce5afd0cd8ee46 I think this is less concerning than similar limits in CodeLens though, since it actually makes sense to start function from _any_ file inside the function folder, and currently the command is available only in right click menu.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: I could add this to stop at project's workspace root. We could not exclude the possibility where there is another folder named `functions` inside the function folder.

packages/salesforcedx-vscode-core/src/channels/channelService.ts
xyc: Updated - integration tests and sanity check looks good

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
xyc: Updated

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: Project is missing the function.toml file. Create function.toml or create a new function using SFDX: Create Function.

@danyusf - assuming we don't doc this TOML file and how to create it - if we do, then we can leave this message as-is. But if not, would it be better to say to add it back? Or just say create a new function?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: Drop the "please" - we usually don't use please unless we're asking customer to do something VERY unpleasant. So: Open a function file to run SFDX: Start Function.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -138,9 +138,15 @@ export const messages = {
...
+    'No function.toml found. Create function.toml or intialize a new function folder using SFDX: Create Function.',
+  force_function_start_warning_not_in_function_folder:
+    'Please open a function file to run SFDX: Start Function',
   force_function_invoke_text: 'SFDX: Invoke Function',
   force_function_invoke_tooltip: 'Send Request',
   force_function_stop_text: 'SFDX: Stop Function',
+  force_function_stop_in_progress: 'Stopping Function',
+  force_function_stop_not_started: 'No Function is running locally',
```
ekapner, line 149: How does customer recover from No function is running locally? Is it SFDX: Start Function? Right-click on a Functions file and then select it?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
danyusf: We will doc the TOML file (we already have it doc'd in the dev preview docs, and at GA will have it in a DSC dev guide). Going forward the scenarios where they have a Function project but no TOML file will be rare (if they use DX to create the project it creates a default TOML for you). I would say _depending on the situation_, a dev might want to just add the TOML file to an existing project they've been working on previously, or start over, so the proposed error message seems reasonable.

ekapner: No function.toml found. Create a function.toml, or create a new function using SFDX: Create Function.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: @xyc ^^

xyc: Usually this is something that user doesn't need to recover from, since if their intention was to stop the function container, their goal is already completed here. And this is a mild informational message to indicate that. Similar message can be found here:
https://github.com/forcedotcom/salesforcedx-vscode/blob/e9e90b928028123342143518cac1c26188965eb3/packages/salesforcedx-vscode-lwc/src/messages/i18n.ts#L35-L36

ekapner: Ok, thanks.

lcampos (approved):

PR #2501, opened 2020-09-09
Soql Builder - Transform soql to ui and back to soql

dehru:
### What does this PR do?

_( this PR was branched off of an unmerged PR jg/sobjectIntegration.  So for diff simplification, I'm using that as a base.  It needs to be switched to develop before final merge )_

This PR integrates the soql model from @salesforce/soql-model with the @salesforce/soql-builder-ui so that soql queries in the .soql file are translated into the ui model, and that changes to the ui are translated into actual soql.  

@W-8027744@

### Functionality After
soql-to-ui-to-soql


dehru:
( I've got some CI build issues i will address.  I'm not seeing those locally. )

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -56,25 +50,9 @@ export class SOQLEditorInstance {
     protected webviewPanel: vscode.WebviewPanel,
     protected _token: vscode.CancellationToken
   ) {
-    this.updateWebview = this.createWebviewUpdater(
```
dehru, line 59: Simplifying this unnecessary nonsense that i introduced before.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -90,82 +68,67 @@ export class SOQLEditorInstance {
...
-          console.log('message type is not supported');
-        }
+  protected onDidRecieveMessageHandler(e: SoqlEditorEvent): void {
+    switch (e.type) {
+      case MessageType.ACTIVATED: {
+        this.updateWebview(this.document);
+        break;
+      }
+      case MessageType.QUERY: {
+        const soql = SoqlUtils.convertUiModelToSoql(JSON.parse(e.message));
```
dehru, line 106: This is the meat of the integration

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -90,82 +68,67 @@ export class SOQLEditorInstance {
...
-    document: vscode.TextDocument
-  ) {
-    return function updateWebview(): void {
-      webview.postMessage({
-        type: MessageType.UPDATE,
-        message: document.getText()
-      });
-    };
+  protected updateWebview(document: vscode.TextDocument): void {
+    const uimodel = SoqlUtils.convertSoqlToUiModel(document.getText());
```
dehru, line 72: This is the meat of the PR.

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/soqlEditorInstance.test.ts
```diff
@@ -136,22 +146,64 @@ describe('SoqlEditorInstance should', () => {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     uri: vscode.Uri,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     token: vscode.CancellationToken
   ): string {
     return 'SELECT A FROM B';
   }
 }
 
 class TestSoqlEditorInstance extends SOQLEditorInstance {
```
dehru, line 195: Thanks @jgellin-sf for this pattern.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
xyc: nit: camel case `uiModel`

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+    queryModel: Soql.Query
+  ): ToolingModelJson {
+    const fields =
+      queryModel.select &&
+      (queryModel.select as Soql.SelectExprs).selectExpressions
+        ? (queryModel.select as Soql.SelectExprs).selectExpressions
+            .filter(expr => !SoqlModelUtils.containsUnmodeledSyntax(expr))
+            .map(expr => ((expr as unknown) as Soql.FieldRef).fieldName)
+        : undefined;
+    const sObject = queryModel.from ? queryModel.from.sobjectName : undefined;
```
xyc, line 34:
```suggestion
    const sObject = queryModel.from?.sobjectName;
```

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ModelDeserializer, ModelSerializer } from '@salesforce/soql-model';
+import { Impl, Soql, SoqlModelUtils } from '@salesforce/soql-model/lib';
```
xyc, line 9: For readability I'd recommend a more specific name than `Impl`, or at least in this file (by using import as).

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/soqlEditorInstance.test.ts
```diff
@@ -136,22 +145,64 @@ describe('SoqlEditorInstance should', () => {
...
+      } but does not include ${uiModelOne.sObject}`
+    );
+  });
+
+  it('handles activation event and updates the webview', async () => {
+    const updateWebviewSpy = sandbox.spy(instance, 'updateWebview');
+    instance.sendEvent({
+      type: MessageType.ACTIVATED
+    });
+    assert(
```
praksb, line 174: can we use `expect` instead of `assert` to make this validation more readable? similar to what we have in soqlUtils.test.ts.

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
dehru: I'm undecided a bit here as the soql-model is written that way, and exports those members...I guess I'm used to seeing that.  @jgellin-sf - what's your thoughts?

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/soqlEditorInstance.test.ts
dehru: Done

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
dehru: Interestingly, I had to shut prettier off on this line after the change.  Have you seen it complain before about using the elvis operator?

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
lcampos: Additional to that, you shouldn't have to import using `/lib`, you can change the library's package.json so that everything you need to consume can be imported by just using `@salesforce/soql-model`

jgellin-sf: Naming is hard. I like the interfaces being defined separately from the implementations because that makes things in the model really flexible independent of implementation. There are a couple of conventions for implementations of interfaces, non-prefixed interfaces and Impl-postfixed implementations which is what I went with, or I-prefixed interfaces and non-prefixed implementations, which seems like an anti-convention in typescript for some reason. Grouping the various interfaces and the various implementations with namespaces felt right as a way to make it clear what we are dealing with in certain situations, but the namespaces themselves are maybe not named that well. 'Soql' for the SOQL model namespace is maybe ok, but 'Impl' for the model implementation namespace lacks specificity. 

@xyc I am onboard with the `import { Impl as SoqlModelImpl }` or some such renaming for readability, but maybe it's better to rename the entire namespace at the source. Out of curiosity, what conventions do you favor for interface-implementation naming?

jgellin-sf (approved):

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
jgellin-sf: Regarding @lcampos comments, the soql-model lib is set up as he describes, so all of those imports can be in one statement.

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
xyc: @dehru Most likely an update to prettier version would solve it

packages/salesforcedx-vscode-soql/src/editor/soqlUtils.ts
xyc: > Out of curiosity, what conventions do you favor for interface-implementation naming
I don't think we have specific naming to separate interface vs implementation, but `SoqlModelImpl` seems fine in this case

dehru: We are going to change this at the source.  and then we will update this.

dehru:
This was branched off of jg/sobjectIntegration branch.  And when that was squash merged, i getting all his commits.  The file diffs are correct.  And once this is squash merged, the 30 commits won't matter.  But let me know if you'd prefer I rebase.

xyc (approved):

PR #2502, opened 2020-09-09
fix: switch to json schemas from schemas module

smithgp:
### What does this PR do?

Switch to the json schema files in the @salesforce/schema module.

### Functionality Before

scratch def and sfdx-project json schemas work 

### Functionality After

scratch def and sfdx-project json schemas still work


smithgp:
@lcampos @amphro Any comments? Thanks.

AnanyaJha (approved):

packages/salesforcedx-vscode-core/package.json
```diff
@@ -895,11 +895,11 @@
     "jsonValidation": [
       {
         "fileMatch": "sfdx-project.json",
-        "url": "https://forcedotcom.github.io/schemas/sfdx-project.json/sfdx-project.schema.json"
```
lcampos, line 898: This should be changed to use the npm module being published. We have something on our roadmap to do it or you can do it in this PR

lcampos (changes requested):

packages/salesforcedx-vscode-core/package.json
smithgp: @lcampos Thomas did bring that up when I talked to him about this, and it's straightforward enough. It does have a side affect of not immediately picking up changes to the schema files in the schema repo until there's been a publish there and here, but does have the advantage of reliable delivery schedule and offline access.
Is that repo's module published on a schedule or on each PR? Or is it manual? Or is that still in the works? I don't see any circleci or github actions config in there, and it was last published in May which means the module is missing a couple of the most recent changes, which would make switching to it now have regressions.

lcampos: Yeah, we're aware of those changes when it comes to using a module, this is still something we want to do so the experience across tools is consistent (CLI, VSCode, etc).

Specific to releases we've been doing them manually since we just started moving to that new model. We'll be publishing it again in the next days since we have the v50 updates coming up. That release will be done using CircleCI. After that I think we'll only publish it on every major platform release (spring, summer, winter) or when there are significant changes that must go out.

smithgp: Ok, that makes sense. I should have some time to see about switching this over to the npm module, and we can wait to merge until a newer module is published.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -28,6 +28,7 @@
     "@salesforce/apex-node": "0.0.9",
     "@salesforce/core": "2.9.0",
     "@salesforce/salesforcedx-utils-vscode": "49.12.0",
+    "@salesforce/schemas": "^1",
```
smithgp, line 31: This isn't technically needed since salesforce/core brings it in too, but I like to spell out dependencies that the code directly references.

smithgp:
This needs to wait to be merged until a newer salesforce/schemas module is published.

packages/salesforcedx-vscode-core/package.json
lcampos: Yes, I agree on being explicit about that dependency

lcampos (approved):
This looks good and ready to merge. I'll get the schema automation sorted out in the next days, but that shouldn't stop us from merging this.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -908,11 +909,11 @@
     "jsonValidation": [
       {
         "fileMatch": "sfdx-project.json",
-        "url": "https://forcedotcom.github.io/schemas/sfdx-project.json/sfdx-project.schema.json"
+        "url": "./node_modules/@salesforce/schemas/sfdx-project.schema.json"
```
amphro, line 912: I just want to confirm that this will always be the path right? i.e. does vscode currently or will do any module hoisting or linking (like yarn2) in the future? If so these paths would break.

amphro (approved):

packages/salesforcedx-vscode-core/package.json
smithgp: @amphro It doesn't appear that vscode does anything like that right now, it seems to just basically unzip the .vsix as-is. I have seen other extensions that reference files in ./node_modules directly programmatically; I thought I had seen this way in package.json, too, which is why I figured this would work (which it does).
I suppose vscode could change something in the future, but also, as far as I know, there is no programatic api to dynamically add schemas to the default json langserver; you have to do it through this package.json, which has to point to urls or relative file paths.

smithgp:
The fixed 1.0.3 version of `@salesforce/schemas` works. I'll merge once tests pass.

PR #2515, opened 2020-09-11
feat: adding codelens to invoke function from payload @W-8046500@

jag-sfdc:
### What does this PR do?
Add codelens "Send Request" to invoke function from a payload json

@W-8046500@

### Functionality Before
N/A

### Functionality After
Sep-11-2020 07-41-10


packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
```diff
@@ -6,6 +6,38 @@
...
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker
+} from '../util';
+
+export class ForceFunctionInvoke extends SfdxCommandletExecutor<string> {
+  public build(payloadUri: string): Command {
+    return new SfdxCommandBuilder()
+      .withDescription(nls.localize('force_function_invoke_text'))
+      .withArg('evergreen:function:invoke')
+      .withArg('http://localhost:8080')
```
jag-sfdc, line 29: Ideally we want this port from start. We have a dependency on the functions cli to expose that. I have it logged as a separate work item - W-8070455

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvokeCodeLensProvider.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        'sfdx.force.function.invoke'
+      );
+    });
+  });
+
+  it('should not create codelens for package.json', async () => {
+    const files = await workspace.findFiles(
+      new RelativePattern(
+        getRootWorkspacePath(),
+        'package.json'
```
jag-sfdc, line 37: The previous test just used an in-memory text document. For this test, we need a dummy package.json, so I've created an empty package.json in sfdx-simple (below). We can change this test to use the functions' package.json which will go as start/stop function tests. @xyc - Can you confirm if this is ok?

xyc: Yeah sounds good. Could we place this dummy `package.json` under `sfdx-simple/functions/demoFunction`? I could add some additional files in there for start/stop

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvoke.test.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
+import { ForceFunctionInvoke as Unit } from '../../../../src/commands/functions/forceFunctionInvoke';
```
xyc, line 9: I don't see we use this pattern elsewhere. I don't have a very strong preference against this pattern, but I feel if we just use the original names the code readability would still be on par or slightly better

xyc (approved):
Looking good with some code style comments

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvokeCodeLensProvider.test.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
+import { CancellationTokenSource, RelativePattern, workspace } from 'vscode';
+import { provideFunctionInvokeCodeLens as unit } from '../../../../src/commands/functions';
+import { getRootWorkspacePath } from '../../../../src/util';
+
+describe('Force Apex Function Invoke Codelens', () => {
```
xyc, line 13:
```suggestion
describe('Force Function Invoke Codelens', () => {
```

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvokeCodeLensProvider.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should create codelens for a json payload file', async () => {
+    const cancellationTokenSource = new CancellationTokenSource();
+    workspace.openTextDocument({
```
xyc, line 17: I suppose this creates an untitled document. Would codelens show up in untitled JSON? (You can create an untitled doc and go to command palette -> change language mode)

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvoke.test.ts
jag-sfdc: I thought this actually improves readability by clearly showing what you are testing. I can revert it if not everyone is onboard and is something we will follow going forward.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvokeCodeLensProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+export const functionInvokeCodeLensProvider = new ForceFunctionInvokeCodeLensProvider();
+
+/**
+ * Register Code Lens Provider with the extension context
+ * @param context Extension context
+ */
+export function registerFunctionInvokeCodeLensProvider(context: ExtensionContext) {
```
lcampos, line 56: This part of the code fails to register the code lens in the correct file if more than one *.json file is present. If a user builds the function it will generate a package-lock.json file and the code lens will get assigned to that one instead of the function's `package.json`

lcampos (changes requested):

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvokeCodeLensProvider.ts
lcampos: send-request

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvokeCodeLensProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+      functionInvokeCodeLensProvider
+    )
+  );
+}
+
+export async function provideFunctionInvokeCodeLens(
+  document: TextDocument,
+  token: CancellationToken
+): Promise<CodeLens[]> {
+  if (path.basename(document.uri.fsPath) === 'package.json') {
```
lcampos, line 69: Following the comment left above, we probably want to expand this to lock files (npm & yarn) and other tooling config files we recommend users to have on their salesforce projects like eslint, prettier, etc.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvoke.test.ts
xyc: When someone reads the code, a full name gives the person more context at a place where it's tested. Whereas if using a generic unit name, you'll have to refer to somewhere else. If you look at a single line where `Unit` was involved, is it easier to tell what the statement is doing than using a full name?

In addition, we are in an integration test suite here although we are using it for unit tests.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvokeCodeLensProvider.test.ts
jag-sfdc: This created an untitled.json doc and hence gets picked up to show the codelens. I tried to change the language mode of a different file type from command palate -> change language mode to JSON, but codelens don't appear in that case.

xyc: I was wanting to confirm here that if you create an untitled doc (cmd + N) in VS Code and change language mode to JSON it should not show any CodeLens, since document selector should have limited it to functions folder. If that doesn't show up it's fine

xyc: Otherwise, you might want to add `scheme: 'file'` to document selector

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvokeCodeLensProvider.ts
jag-sfdc: I've now included 'package.json', 'package-lock.json', 'tslint.json', 'lerna.json', 'tsconfig.json' as the list of non-payload files. Let me know I should include any more.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionInvokeCodeLensProvider.test.ts
jag-sfdc: Ah! That's a such an edge case to take care :)
Thankfully it does not show the codelens. That's because the filename is still ~Untitled even when we change the language mode.

xyc: Sounds good, thanks for checking! I was wondering how this test was able to generate codelens

jag-sfdc: Good question. For some reason, the behavior is different here. When you set language as json, it changes the filename to Untitled.json.

lcampos (approved):

PR #2517, opened 2020-09-11
Added expanded extension pack

ntotten:
### What does this PR do?
Adds an extension pack that contains all of the sf extension as well as recommended third-party extensions


smithgp:
@ntotten Would this be an appropriate place to also include the salesforce analytics extensions? (https://marketplace.visualstudio.com/items?itemName=salesforce.analyticsdx-vscode)

lcampos (approved):

ntotten:
@smithgp do you think there is a lot of overlap in people using the core ones and the analytics ones? I am not opposed to including the analytics extensions as well, but at the same time we probably dont want to force install extensions that aren't needed. The intent of this was to have a single pack that includes our core extensions and a few recommended third-party extensions.

smithgp:
@ntotten Well, 100% of folks using the analytics extensions will be using the core extensions; in the other direction, the analytics extension install number is in the low 1000's right now, so it's definitely a smallish subset of the platform vscode extensions installs. We do want that to increase, though, particularly as we include more analytics functionality in base/freemium licenses and as we look to move some analytics functionality into the base platform.
But, if the intent here is primarily for core platform devs to bring in the recommended 3rd part extension, then, yeah, this might not be the right place. Would another extension pack (e.g. 'Salesforce Add-ons' or something) that had the analytics and commerce extensions (and anything else that comes along from other clouds) make sense?

PR #2524, opened 2020-09-14
chore(soql): add missing type for mkdirp

dehru:
### What does this PR do?
This fixes the compile error:

```salesforcedx-vscode-soql: node_modules/@salesforce/core/lib/util/fs.d.ts(4,28): error TS7016: Could not find a declaration file for module 'mkdirp'. '/Users/tcromer/workspace/vscode/salesforcedx-vscode/packages/salesforcedx-vscode-soql/node_modules/mkdirp/index.js'```

This appears to be a downstream dependency from us as we don't use this module directly.

### Functionality After
Compile and Lint pass.

### DIFF
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -32,6 +32,7 @@
     "@salesforce/ts-sinon": "1.2.2",
     "@types/chai": "^4.0.0",
     "@types/debounce": "^1.2.0",
+    "@types/mkdirp": "^1.0.1",
     "@types/mocha": "^5",
     "@types/node": "12.0.12",
     "@types/sinon": "^2.3.2",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/salesforcedx-vscode-soql/package.json [hunk 1]
```diff
@@ -32,6 +32,7 @@
     "@salesforce/ts-sinon": "1.2.2",
     "@types/chai": "^4.0.0",
     "@types/debounce": "^1.2.0",
+    "@types/mkdirp": "0.5.2",
```
dehru, line 35: there is version 1.0.1 out now, but npm list mkdirp shows several hits on the 0.5.x versions.  This seem to be coming from the system-tests package.

dehru: Here's the only reference i see to it.

https://github.com/forcedotcom/salesforcedx-vscode/blob/987031482aeaa0da4a53d0e1e285d9278dd0ce39/packages/system-tests/package.json#L28

lcampos (approved):

PR #2527, opened 2020-09-15
Xyc/run current integration test

xyc:
### What does this PR do?
Adding the following launch config: 
- Launch Current Integration Test with sfdx-simple (Extensions other than LWC extension)
- Launch Current Integration Test with lwc-recipes (LWC extension only)

@W-8087484@

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -106,7 +106,7 @@
       "internalConsoleOptions": "openOnSessionStart"
     },
     {
-      "name": "Launch Current Integration Test with lwc-recipes",
+      "name": "Launch Current LWC Integration Test with lwc-recipes",
       "type": "extensionHost",
       "request": "launch",
       "runtimeExecutable": "${execPath}",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -87,6 +87,25 @@
       "preLaunchTask": "Compile"
     },
     {
+      "name": "Launch Current Integration Test with sfdx-simple",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "env": {
+        "SFDX_TEST_FILE_PATH": "${file}"
+      },
+      "args": [
+        "${workspaceFolder}/packages/system-tests/assets/sfdx-simple",
+        "--extensionDevelopmentPath=${workspaceFolder}/packages",
+        "--extensionTestsPath=${workspaceFolder}/packages/salesforcedx-vscode-core/out/test/vscode-integration"
+      ],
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -104,34 +87,25 @@
...
-        "${input:circleCiId}"
-      ]
+      "args": ["${input:circleCiId}"]
     },
     {
       "label": "Watch",
       "command": "npm",
       "type": "shell",
       "presentation": {
-        "reveal": "silent",
```
xyc, line 126: This is what's actually changed: watch task is revealed in the terminal tab. The other changes are code formatting changes from prettier

jag-sfdc (approved):

.vscode/launch.json
```diff
@@ -86,6 +86,44 @@
       },
       "preLaunchTask": "Compile"
     },
+    {
+      "name": "Launch Current Integration Test with sfdx-simple",
```
lcampos, line 90: Lets change the name to `Launch Current Core Integration Test with sfdx-simple` since these launch configs are specific to tests in the core module.

AnanyaJha: Let's do the same thing for the lwc recipes launch config- `Launch Current LWC Integration Test with lwc-recipes`

dehru:
This is very helpful.  I just tried it out.  I'm kinda new to testing in this repo, and so i might have done something wrong at first because it didn't work right off the bat.  Here's my steps:

1. ( install, compile )
2. Ran this launch config with sfdx-simple in Code ( not insiders )
3. Got an error about access to one of the vscodeIntegrationTests.xml.
    ( wondered if i needed to run all the tests first to maybe setup these xml files )
4. Closed Code. Ran all the tests from the command line.
5. Got same error.
6. Opened Code-Insiders and ran this launch config and it worked.
7. Re-opened Code and ran this launch config and it worked this time.

It might be environment specific for me, but now that I have it working, I'm super happy.

.vscode/launch.json
dehru: This worked for me running a test in the soql package.  Is it supposed to only work in core?

dehru (approved):
Worked for me after a couple hiccups.  See previous comment.

.vscode/launch.json
xyc: @lcampos This one works for all but LWC extension. @AnanyaJha I can update the second one so it's more informative

xyc:
@dehru I got issues with vscodeIntegrationTests.xml too in the past, but when I tried this from a fresh cloned repo there're no issues

PR #2538, opened 2020-09-17
Function start prerequisites

xyc:
### What does this PR do?
Check the prerequisites are met for "SFDX: Start Function":
- Plugin installed
- Docker installed and started

@W-8046471@


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -142,6 +142,10 @@ export const messages = {
     'No function.toml found. Create a function.toml, or create a new function using SFDX: Create Function.',
   force_function_start_warning_not_in_function_folder:
     'Open a function file to run SFDX: Start Function',
+  force_function_start_warning_plugin_not_installed:
+    'To run this command, install the evergreen plugin. For more info, see [Getting Started with Salesforce Functions](https://github.com/forcedotcom/evergreen-docs/blob/master/Getting%20Started%20Guide/Evergreen%20Getting%20Started%20Guide.md#installing-the-salesforce-functions-sfdx-plugin).',
+  force_function_start_warning_docker_not_installed_or_not_started:
+    'It looks like Docker is not installed or running. To run this command, install and start Docker Desktop from [https://www.docker.com/products/docker-desktop](https://www.docker.com/products/docker-desktop)',
```
xyc, line 148: @ekapner @danyusf These are the messages shown to users in VS Code when evergreen plugin is not installed or docker is not installed or started.

- Is there a better doc website URL for installing the evergreen plugin?
- Does the official docker desktop link https://www.docker.com/products/docker-desktop work?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -142,6 +142,10 @@ export const messages = {
     'No function.toml found. Create a function.toml, or create a new function using SFDX: Create Function.',
   force_function_start_warning_not_in_function_folder:
     'Open a function file to run SFDX: Start Function',
+  force_function_start_warning_plugin_not_installed:
+    'To run this command, install the evergreen plugin. For more info, see [Getting Started with Salesforce Functions](https://github.com/forcedotcom/evergreen-docs/blob/master/Getting%20Started%20Guide/Evergreen%20Getting%20Started%20Guide.md#installing-the-salesforce-functions-sfdx-plugin).',
```
xyc, line 146: Ideally, we'd like to install the missing plugin automatically when user is prompted, but we don't have that pattern now. We could track that as an enhancement.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -140,12 +215,13 @@ export async function forceFunctionStart(sourceUri?: Uri) {
...
-    notificationService.showWarningMessage(
-      nls.localize('force_function_start_warning_not_in_function_folder')
+    const warningMessage = nls.localize(
+      'force_function_start_warning_not_in_function_folder'
     );
+    notificationService.showWarningMessage(warningMessage);
     telemetryService.sendException(
-      'force_function_start',
-      'force_function_start_not_in_function_folder'
+      'force_function_start_not_in_function_folder',
```
xyc, line 223: Updated so that the telemetry log names are consistent

packages/salesforcedx-vscode-core/src/messages/i18n.ts
danyusf: Because Evergreen isn't GA yet (and won't be until roughly the 232 time-frame), that Getting Started doc link is correct. Note that that link is in a private github repo that we add pilot customers to, so anyone getting this message who's not in the pilot program will get a 404 for that link (but, truthfully, it would be unexpected to have a customer get this message who's not already working with the pilot).

I believe that Docker link should work (it's the same link I use in the Evergreen dev guide).

One request (which I probably should have mentioned earlier) -- avoid using the term "Evergreen", as product marketing has decided the product name will be "Salesforce Functions", and the Evergreen term will be mostly an internal term. @ekapner I believe previous PRs I've looked at were not using "Evergreen", but if I missed any uses, my apologies.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
danyusf: Note "evergreen" should be replaced with "Salesforce Functions" (e.g. "To run this command, install the Salesforce Functions plugin.")

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: @danyusf - thanks for the reminder about Evergreen - and I believe you're correct that I haven't seen it used in previous messages we've reviewed.

xyc: @danyusf @ekapner  Thank you for the reminder - Yes I haven't mentioned evergreen other than this one. I'll update it to "Salesforce Functions"

ekapner: @xyc - I installed Docker from that link

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: Our doc style guide says to hyphenate "plug-in" but unsure if our messages consistently hyphenate it. So if you don't see it hyphenated elsewhere, let's not introduce an inconsistency

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -31,6 +31,37 @@ import {
...
+  | 'force_function_start_docker_plugin_not_installed_or_started';
+
+const forceFunctionStartErrorInfo: {
+  [key in ForceFunctionStartErrorType]: {
+    cliMessage: string;
+    cliExitCode: number;
+    errorNotificationMessage: string;
+  }
+} = {
+  force_function_start_plugin_not_installed: {
```
jag-sfdc, line 49: Based on a slack discussion with @xyc , starting the same function multiple times kills the first one and starting a different function when another function has started throws a port-bind error. These are edge cases and we can take it up later.

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -146,6 +146,8 @@ export const messages = {
     'To run this command, install the Salesforce Functions plugin. For more info, see [Getting Started with Salesforce Functions](https://github.com/forcedotcom/evergreen-docs/blob/master/Getting%20Started%20Guide/Evergreen%20Getting%20Started%20Guide.md#installing-the-salesforce-functions-sfdx-plugin).',
   force_function_start_warning_docker_not_installed_or_not_started:
     'It looks like Docker is not installed or running. To run this command, install and start Docker Desktop from [https://www.docker.com/products/docker-desktop](https://www.docker.com/products/docker-desktop)',
+  force_function_start_unexpected_error:
+    'SFDX: Start Function exited unexpectedly with code %s',
```
xyc, line 150: @ekapner This is the message when an error code is unexpected, similar to what we have in local dev: 

https://github.com/forcedotcom/salesforcedx-vscode/blob/20266dc8760222b290f48df369bae9220a433d9a/packages/salesforcedx-vscode-lwc/src/messages/i18n.ts#L31

fyi @jag-sfdc https://github.com/forcedotcom/salesforcedx-vscode/pull/2538/commits/0f16aa0948a96ca54a0fcf2f69f5a6fcc478758c adds handling for unexpected error

lcampos (approved):

packages/salesforcedx-vscode-core/src/messages/i18n.ts
xyc: @ekapner Yeah most of our mentions of "plug-in" are hyphenated. I'm going to leave as is

packages/salesforcedx-vscode-core/src/messages/i18n.ts
xyc: @ekapner They will have to see what the output panel says. We have handled docker installation/ plugin installation, this is a precaution to show a message where CLI starts and throw an unexpected error. For recovering from an unexpected error, generally user would need to look at the error message that CLI (Functions plugin) outputs. It would be more helpful for the user if the message is more user friendly, but there's not much we can do in VS Code.

PR #2539, opened 2020-09-17
Port release 49.10.0 commits

lcampos:
Port release commits back to develop branch


sfsholden:
Let's do the merge instead just to be consistent. I'll kick it off now

Issue #2429, opened 2020-08-20
Repeatedly get "Cannot read property 'get' of undefined" for non Salesforce projects


robcowell:
### Summary

VS Code repeatedly outputs "Cannot read property 'get' of undefined" in the Salesforce CLI window of the Output tab/pane

### Steps To Reproduce:

1.  Ensure Salesforce extension pack for VS Code is installed
1.  Open a non-Salesforce project - in my case, it's a HTML/JS/CSS project with WebGL/ThreeJS

### Expected result

Salesforce extensions should not produce any output

### Actual result

VS Code repeatedly outputs "Cannot read property 'get' of undefined" in the Salesforce CLI window of the Output tab/pane

### Additional information

**VS Code Version**: 1.47.3

**SFDX CLI Version**: sfdx-cli/7.68.6-d37008df83 win32-x64 node-v12.18.3

**OS and version**: Windows 10 Pro version 2004

lohit398:
This is happening with 1.48.2 as well.

Thanks!

lcampos:
Thanks for reporting this, the fix for it has been released as part of version 49.13.0.

PR #2547, opened 2020-09-18
fix: cannot read property get of undefined

maaaaarco:
### What does this PR do?
Prevents the _core_ extension to initialise certain features when a non-Salesforce project is opened.

For reason beyond my comprehension the _salesforcedx-vscode-core_ extension gets activated even when there isn't a _sfdx-project.json_ file in the project folder.

It seems that this _workspaceContains_ clause is ignored:
```json
"activationEvents": [
    "onCommand:sfdx.force.project.create",
    "onCommand:sfdx.force.project.with.manifest.create",
    "onCommand:sfdx.debug.isv.bootstrap"
  ],
```

When the _core_ extension activates then this code runs:
```ts
if (hasRootWorkspace()) {
    const files = await vscode.workspace.findFiles(
      '**/sfdx-project.json',
      '**/{node_modules,out}/**'
    );
  }
```

Which cause the extension's commands and features like _registerPushOrDeployOnSave_, _orgPicker_ to be enabled also for non-Salesforce project. It's sufficient to have the `node_modules` folder.

The _core_ extension now relies on the predicate _IsSfdxProjectOpen_ in order to activate Salesforce specific features.

#2429


maaaaarco:
@lcampos btw, I think what cause the _core_ extension to be activated... it's the _lightning_ package that is wrongly detecting the workspace type.
I think there's an issue in the `activate` method of  the main `index.ts`... Specifically with this line:
```ts
// 3) If activationMode is autodetect or always, check workspaceType before startup
  const workspaceType = lspCommon.detectWorkspaceType(workspaceUris);
```
It returns 4 as workspaceType which maps to `SFDX = 4,`.

lcampos:
> @lcampos btw, I think I got what cause the _core_ extension to be activated... it's the _lightning_ package that is wrongly detecting the workspace type.
> I think there's an issue in the `activate` method of the main `index.ts`... Specifically with this line:
> 
> ```ts
> // 3) If activationMode is autodetect or always, check workspaceType before startup
>   const workspaceType = lspCommon.detectWorkspaceType(workspaceUris);
> ```
> 
> It returns 4 as workspaceType which maps to `SFDX = 4,`.

@maaaaarco yeah, that extension and lwc activate core in scenarios outside of salesforce dx projects. This looks like a good initial fix but we do need to rework the activation dependencies across extensions (which we will start working on soon).

Again thanks for the PR, I'll run some tests and merge it if everything looks good.

lcampos:
Closing since the changes here were merged in PR #https://github.com/forcedotcom/salesforcedx-vscode/pull/2601. @maaaaarco thanks again for your contributions.

PR #2549, opened 2020-09-18
perf(workspace): Manage org context through WorkspaceContext object

brpowell:
### What does this PR do?

Introduces `WorkspaceContext` - a central manager for handling all things regarding the context of a workspace (eventually). This PR is focused on having it manage the default org context for now, but later can be expanded to include things like setting the VS Code org type context and reading project configuration.

Objects can either subscribe to when the default org changes:

```typescript
//// orgPrinter.ts
class OrgPrinter implements OrgSubscriber {
  public onOrgChange(username?: string, alias?: string) {
    console.log(`The current username is now ${username}, with an alias of ${alias}`)
  }
}

//// index.ts
await WorkspaceContext.initialize(vscodeContext); // call before virtually everything in activation

const orgPrinter = new OrgPrinter();
WorkspaceContext.get().subscribe(orgPrinter)
```

or directly access the default org's username, alias, and connection:
```typescript
const { orgAlias, orgUsername } = WorkspaceContext.get();
const connection = await WorkspaceContext.get().getConnection();
const client = new SourceClient(connection)

await client.metadata.deploy('/path/to/metadata')

console.log(`deployed metadata to ${orgAlias || orgUsername}`)
```

Connections are cached by org per session after they are initially created to speed up subsequent uses of them. This was primarily the reason for creating this to speed up beta deploy/retrieve commands, and will be useful for other future non-cli based commands. 

@W-8108485@


packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -0,0 +1,125 @@
...
+      this._orgUsername = await OrgAuthInfo.getUsername(usernameOrAlias);
+      this._orgAlias =
+        usernameOrAlias !== this._orgUsername ? usernameOrAlias : undefined;
+    } else {
+      this._orgUsername = undefined;
+      this._orgAlias = undefined;
+    }
+
+    setupWorkspaceOrgType(this._orgUsername)
+      .catch(e => console.error('Error setting VS Code org type context'));
```
brpowell, line 100: This function already logs exceptions. Eventually the logic in this function should be reworked to live in here.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    for (const subscriber of this.subscribers) {
+      this.notifySubscriber(subscriber);
```
xyc, line 101: It's your call, but we could use `vscode.EventEmitter` or node event emitter instead of a custom implementation. You'll get things like `dispose` for free.

brpowell: I briefly thought about using the vscode EventEmitter but wasn't sure if there was a major benefit to doing one way over the other. I'll take a closer look tomorrow to see what it'd be like.

What's the preference of using vscode.EventEmitter vs. NodeJS build-in?

xyc: @brpowell https://code.visualstudio.com/api/references/vscode-api#EventEmitter%3CT%3E I haven't tried if it works across extensions, but the document says 
>Use this class if you want to provide event from within your extension, for instance inside a TextDocumentContentProvider or when providing API to other extensions.

also, `vscode.EventEmitter` is used across a lot of VS Code APIs, so it would be beneficial if we are providing it as an interface in our APIs for consistency.

brpowell: @xyc thanks for the suggestion, I went ahead and updated it to use the vscode.EventEmitter impl.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceDeploySourcePath.test.ts
```diff
@@ -95,33 +95,5 @@ describe('Force Source Deploy Using Sourcepath Option', () => {
...
-      );
-      const executor = new LibraryDeploySourcePathExecutor();
-      const filePath = path.join(
-        'test',
-        'file',
-        'path',
-        'classes',
-        'apexTest.cls'
-      );
-      const mockToolingDeploy = sb
```
AnanyaJha, line 117: do we not need this test anymore?

brpowell: It was failing with the changes and decided not to "fix" it. Seemed like a pointless test since we don't even support setting the namespace when deploying with metadata api, and the test wasn't testing that in the first place.

AnanyaJha (approved):

PR #2560, opened 2020-09-21
feat: show message if no default org is set when starting function

xyc:
### What does this PR do?
If org is not authenticated, show user an informational message if user has not auth’ed with an org and start function container.

@W-7958844@

### Functionality Before
No message is shown when running "SFDX: Start Function" without default org set.

### Functionality After
Following message is shown when running "SFDX: Start Function" without default org set.
Screen Shot 2020-09-21 at 3 48 08 PM


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -146,6 +146,8 @@ export const messages = {
   force_lightning_interface_create_text: 'SFDX: Create Aura Interface',
   force_function_create_text: 'SFDX: Create Function',
   force_function_start_text: 'SFDX: Start Function',
+  force_function_start_no_org_auth:
+    'No default org is set. SFDX: Start Function will continue without org features. Run "SFDX: Set a Default Org" to select an active scratch org, or run "SFDX: Authorize a Dev Hub" and "SFDX: Create a Default Scratch Org" to authenticate to a Dev Hub and create a default scratch org.',
```
xyc, line 150: @ekapner @danyusf also see the screenshot attached to description

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -121,6 +121,19 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
     }).execute(cancellationToken);
     const executionName = execution.command.toString();
 
+    OrgAuthInfo.getDefaultUsernameOrAlias(false)
+      .then(defaultUsernameorAlias => {
+        if (!defaultUsernameorAlias) {
+          const message = nls.localize('force_function_start_no_org_auth');
+          channelService.appendLine(message);
+          channelService.showChannelOutput();
+          notificationService.showInformationMessage(message);
```
brpowell, line 130: Do we want actionable buttons on the notification message to execute suggested commands?

xyc: We could but I'm not aware of similar action buttons elsewhere (for example in other CLI commands, or in local dev)

@smaddox-sf @jag-sfdc We could add buttons like below:
- "Continue without Org" does nothing
- "Set a Default Org" shows the org picker - which corresponds to the first suggestion.

Note that the informational message only appears for a few seconds and disappear, unless it's displayed as a modal prompt. Thoughts?

Screen Shot 2020-09-21 at 3 43 14 PM

jag-sfdc: For uniform experience and also to keep it simple for beta, I think we should skip this.

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
smaddox-sf: Agreed.  I thought we were just going to output a message in the 'output' panel telling them no default org set.  (just as an FYI, no prompt)

packages/salesforcedx-vscode-core/src/messages/i18n.ts
ekapner: @xyc - chatting with Dan to understand limitations for Start without an org. Seems like you can start and test locally. And only need an org when you want to test in one. So, Dan and I will provide feedback once I understand what "without org features" really means.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
ekapner: @smaddox-sf - tell me what you decide in case I don't need to edit this message.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
xyc: If the function uses org features, org auth is needed - For example if it uses `context.org.data.query`. If it doesn't use org features, it's not necessary to auth with an org.

xyc: `without org features` meaning functions that don't use org features would work fine. For functions that use org features, "SFDX: Start Function" runs fine with the message shown, but it will encounter an error when running "SFDX: Invoke Function" about org auth.

ekapner: No default org is set. We recommend that you select an active scratch org (SFDX: Set a Default Org) or create a new scratch org (SFDX: Authorize a Dev Hub, then SFDX: Create a Default Scratch Org). 

@danyusf - do you think it's obvious that we're telling them to run the commands in parens?

ekapner: And @danyusf - should we write this to be generic enough to stand test of time when sandboxes and prod orgs are supported?

danyusf: @ekapner I think that's fine, and yes, I think a Functions dev will understand the next steps. Regarding making the language work for future support for prod/sandbox orgs, this is tricky (just because there's a lot to do to make things "production ready" and things may get changed, or maybe we'll have a separate workflow for deploying Functions to prod/sandbox orgs). I would say the best we can do for now is let the text reflect the current state of Functions.

xyc: Thanks @ekapner @danyusf I'll update the text according to your recommendation for now

PR #2565, opened 2020-09-23
J Hork: Run and View SOQL Query Data

jonnyhork:
### What does this PR do?

This PR will allow the customer to run the SOQL query they have built using the custom editor and view the tabular results in a new webview. 
- The data of the SOQL query will only be stored in memory, no files are saved at this time.
- This work is intended to meet our goal for MVP of the SOQL Query Builder, there will be additional work items created to address any new functionality or changes in UX when viewing the query results.
- A 3PP has been created for the JS library used to render the table: [Tabulator](http://tabulator.info/)
- The Webview's state is preserved when moved into a background tab.

@W-7989627@
@W-8048824@
### Functionality Before
😴😴😴
CLI - low

### Functionality After
🤯🤯🤯🤘📊
Query Data Webview - low


packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+  we will need to flatten the results of nested values
+  in order to be parsed and diplayed correctly
+  */
+  private flattenQueryData(rawqQueryData: QueryResult<JsonMap>) {
+    const records = rawqQueryData.records;
+    // filter out the attributes key
+    records.forEach(result => delete result.attributes);
+    return records;
+  }
+}
```
jonnyhork, line 44: The structure of the response from the API will contain attributes that we want to flatten in order to render the data correctly. For example, we currently support `Select field1, field2 FROM sObject` which yields the following data structure:
```
{
...,
  records: [
    {
      attributes: {
        type: 'sObject',
        url: ''
      },
      field1: '',
      field2: false
    }
  ]
}
```
we will need to remove the `attributes` property in order to render the table of data correctly. This is the same process that the CLI currently does, but with support for more nested data structures.

As the Query Builder supports more complex queries we will need to handle the response in a more robust way.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -130,12 +134,39 @@ export class SOQLEditorInstance {
         });
         break;
       }
+      case MessageType.RUN_SOQL_QUERY: {
```
jonnyhork, line 137: I still need to add a test for this event, coming soon.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* 
+This file can be used to override Tabulator CSS
+for custom styling http://tabulator.info/docs/4.8/style
+*/
```
jonnyhork, line 11: Once we get to styling the extension we will likely want to customize webview and the table a bit. This file will be used for that in future work.

Table styles can be overridden:
http://tabulator.info/docs/4.8/style

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataViewController.js
```diff
@@ -0,0 +1,52 @@
...
+  function renderTableWith(tableData) {
+    new Tabulator('#data-table', {
+      data: tableData,
+      autoColumns: true,
+      pagination: 'local',
+      paginationSize: 10,
+      layout: 'fitColumns',
+      maxHeight: '100%'
+    });
+  }
```
jonnyhork, line 36: There are a quite few ways to customize and control behavior of the rendered table. We can leverage these as need presents itself. For now, I started with a basic set up.

http://tabulator.info/docs/4.8/options

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataViewController.js
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+(function() {
+  const vscode = acquireVsCodeApi();
```
jonnyhork, line 9: This is the JS file that runs in the webview. Any logic to communicate with the extension or control table behavior happens here.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataViewController.js
```diff
@@ -0,0 +1,52 @@
...
+
+  function updateUIWith(queryData, documentName) {
+    const title = document.getElementById('webview-title');
+    title.innerText = documentName;
```
jonnyhork, line 13: File Name Title

packages/salesforcedx-vscode-soql/src/queryResultsView/queryDataViewService.ts
```diff
@@ -0,0 +1,128 @@
...
+  public static register(context: vscode.ExtensionContext): void {
+    QueryDataViewService.extensionPath = context.extensionPath;
+  }
+
+  private updateWebviewWith(webview: vscode.Webview, queryData: JsonMap[]) {
+    webview.postMessage({
+      type: 'update',
+      data: queryData,
+      documentName: getDocumentName(this.document)
+    });
```
jonnyhork, line 44: This is different than a `CustomEditorEvent` and we only have one post message to the webview, so I left this as a string literal. I figured we can add some constants or enumerations if the messaging grows in the future.

jonnyhork (commented):
[Corresponding PR ](https://github.com/forcedotcom/soql-tooling/pull/25)on the tooling side, its small but might as well  have a link here.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -18,26 +21,27 @@ const sfdxCoreExports = sfdxCoreExtension
...
   SOBJECTS_RESPONSE = 'sobjects_response',
-  TEXT_SOQL_CHANGED = 'text_soql_changed'
+  TEXT_SOQL_CHANGED = 'text_soql_changed',
+  RUN_SOQL_QUERY = 'run_query'
 }
 
 async function withSFConnection(f: (conn: Connection) => void): Promise<void> {
-  const conn = await OrgAuthInfo.getConnection();
   try {
+    const conn = await OrgAuthInfo.getConnection();
```
AnanyaJha, line 44: would the new workspace context changes to preserve the connection across a session be useful here? cc @brpowell

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+      const rawQueryData = (await this.connection.query(
+        queryText
+      )) as QueryResult<JsonMap>;
+      const cleanQueryRecords = this.flattenQueryData(rawQueryData);
+
+      return cleanQueryRecords;
+    } catch (error) {
+      // TODO: Doc Review, i18n
+      vscode.window.showErrorMessage(`Could not run the Query ${error}`);
+      throw error;
```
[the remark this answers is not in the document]
ekapner: @jonnyhork - how about something like this? 

Your query contains invalid or incomplete syntax. Fix the syntax errors and try again.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
brpowell: Indeed - @jonnyhork check out the new (like yesterday new) `WorkspaceContext` service in core, it takes care of caching the connection for subsequent requests and allows you to listen when the default org changes. You might need to export it from the core api in `index.ts` to use it first.

packages/salesforcedx-vscode-soql/src/commonUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { TextDocument } from 'vscode';
+
+export function getDocumentName(document: TextDocument): string {
+  const documentPath = document.uri.fsPath;
+  return documentPath.split('/').pop() || '';
```
brpowell, line 12: You can use `path.basename` to get the end part of the filepath.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+  /*
+  As query complexity grows
+  we will need to flatten the results of nested values
+  in order to be parsed and diplayed correctly
+  */
+  private flattenQueryData(rawqQueryData: QueryResult<JsonMap>) {
+    const records = rawqQueryData.records;
```
xyc, line 39:
```suggestion
    const records = rawQueryData.records;
```

brpowell:
Would it be easier to add tabulator as a dependency with NPM?

packages/salesforcedx-vscode-soql/src/queryResultsView/queryDataViewService.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  public createOrShowWebView(): vscode.Webview {
+    this.currentPanel = vscode.window.createWebviewPanel(
+      this.viewType,
+      QUERY_DATA_VIEW_PANEL_TITLE,
+      vscode.ViewColumn.Two,
```
xyc, line 51: Would this open in the expected column when soql file is already in a split editor?

packages/salesforcedx-vscode-soql/src/commonUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+}
```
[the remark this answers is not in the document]
xyc: Do you support untitled document?

packages/salesforcedx-vscode-soql/src/commonUtils.ts
jonnyhork: ooo cool, as in `document.uri.path.baseName`? Ha, if thats the case it seems silly to have a separate file, method and tests for something I can access directly =(

jonnyhork: got it, like this `path.basename(this.document.uri.fsPath)`

jonnyhork: Looks good to me, thanks.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
jonnyhork: @dobladez You were refactoring the way we use the connection object based on previous feedback right? Does this look right to you?

jonnyhork: ok, if its that new then, we probably missed it, I'll check it out.

packages/salesforcedx-vscode-soql/src/commonUtils.ts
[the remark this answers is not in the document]
praksb: @xyc it has to be a .soql file extension for SOQL VS Code extension to activate.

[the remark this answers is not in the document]
xyc: @praksb what i mean for untitled document is you can do Command Palette -> Change Language Mode -> soql

Screen Shot 2020-09-23 at 2 04 22 PM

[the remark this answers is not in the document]
xyc: activation event has `onLanguage:soql` so I think it would activate in this case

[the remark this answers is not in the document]
praksb: That would be a good test case to try.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: fixed, thanks.

packages/salesforcedx-vscode-soql/src/queryResultsView/queryDataViewService.ts
jonnyhork: It opens in a new tab if the editor is already split. We plan to create a WI to handle the webview visibility and moving better.
image

packages/salesforcedx-vscode-soql/src/commonUtils.ts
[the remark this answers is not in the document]
jonnyhork: Yes the extension activates, if not already active, if you change the language mode to soql. However, the customer can not open the soql builder unless the file extension is `.soql`.
untitled file

If however I have a `.soql` file ext but I change the language association to HTML, then I can still open the Query Builder UI.
html file

Lastly, if I have actual HTML in the file with the `.soql` ext and open the UI then build a query, it will write over the contents of the file and WAD.

So bottom line I don't think there is anything handle differently there. If the user goes out of their way to mess things up and overwrite their file I'm not sure thats on us.

@xyc Are you thinking of a different scenario?

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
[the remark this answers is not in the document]
jonnyhork: Done, thanks

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+    // filter out the attributes key
+    records.forEach(result => delete result.attributes);
+    return records;
```
praksb, line 42: how about  getting rid of an intermediate variable records and just return rawqQueryData.forEach(result => delete result.attributes);

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export class QueryRunner {
+  constructor(
+    private connection: Connection,
+    private document?: vscode.TextDocument
+  ) {}
+
+  public async runQuery(queryText: string): Promise<JsonMap[]> {
+    try {
+      const rawQueryData = (await this.connection.query(
```
praksb, line 21: should we have a check before even attempting to run the query to see if there are any known syntax errors? I think that check has to be in the caller.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
praksb: Should this and the other parts of the webview be in soql-tooling? Should we structure this the same way we are handling the builder UI by decoupling it from the extension?

packages/salesforcedx-vscode-soql/src/commonUtils.ts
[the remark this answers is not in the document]
xyc: I looked up custom editor API and it doesn't seem to support untitled doc scenario, so I'll take that soql builder is not available in untitled docs for now. I'm curious if you could still run a query in untitled docs and _utilize the query results webview_, using "Execute SOQL Query with Selected Text".

I initially had some concerns about `document.uri.fsPath`, it turned out to be just `Untitled` in untitled documents so it won't be an issue here.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: I was hesitant to do that, forEach() has got me in the past because it [does not return any value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). I think I need to mutate a value in memory and then return that value.

>forEach() executes the callback function once for each array element; unlike map() or reduce() it always returns the value undefined and is not chainable. The typical use case is to execute side effects at the end of a chain.

xyc: @jonnyhork another way to do it is `return rawQueryData.map(({ attributes, ...props }) => props)` , see [here](https://stackoverflow.com/a/46839399)

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
jonnyhork: After thinking about it, at this point I don't see much benefit to doing that now. The webview currently doesn't make sense as a stand alone app like the LWC app we import and it much much simpler as far as html and js goes. At this point I think the only benefit it organizational. I'm not sure we want to spend the time on moving it over.
Perhaps we can take care of that later as tech debt?
@dehru Thoughts?

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: I'm not sure how we would check for syntax error based on the string text from the document. Wouldn't that require a parser and models that now live the UI?

If the query has bad syntax or can't be run then at least the user is notified thats the case and directed to fix it by the error message.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
praksb: my bad. i didn't read the code properly.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
praksb: sounds good.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
praksb: I think in the current design, you may have to call the SOQL parser directly to validate that. With that being said I think we shouldn't have to worry about it now.
It is something that we could think about later once we have the language server for SOQL working. The language server should be able to tell us at the minimum whether a given soql text has any syntax errors are not. It is an optimization we could consider in the future and consider disabling the RUN Query button in those scenarios with errors.

packages/salesforcedx-vscode-soql/src/commonUtils.ts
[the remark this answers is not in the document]
jonnyhork: >I'm curious if you could still run a query in untitled docs and utilize the query results webview, using "Execute SOQL Query with Selected Text".

I don't see why not, we will probably have to make a few changes to the expose the right stuff for the webviewService to be consumed by other extensions or commands, but ultimately the service just needs the records from an API response to render the table.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: Thanks @xyc, I like the looks of that! Feel like I've been told mutating state is bad juju anyway.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataViewController.js
```diff
@@ -0,0 +1,61 @@
...
+  const vscode = acquireVsCodeApi();
+
+  function updateUIWith(queryData, documentName) {
+    // Display the .soql file name as the title
+    const titleEl = document.getElementById('webview-title');
+    titleEl.innerText = documentName;
+
+    const totalRecordsSizeEl = document.getElementById('total-records-size');
+    totalRecordsSizeEl.innerText = `Returned ${queryData.records.length} of ${queryData.totalSize} total records`;
+
```
jonnyhork, line 24: @ekapner I added some new text to the table webview, this is modeled after the text currently in Workbench. Do you want to make any changes?

The total number represents the total number of records that match the query in the org, where as the "returned" value is the number of the total records returned by the API, capped at 2000.

So it's possible they would get "Returned 2000 of 3750 total records" etc.
image

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
jonnyhork: @brpowell We are now using the new hotness, see the `fix(soql): use the WorkspaceContext for connection` commit

packages/salesforcedx-vscode-soql/src/constants.ts
```diff
@@ -5,9 +5,22 @@
...
+
 export const WEBVIEW_RESOURCE_ROOTS_PATH =
   'node_modules/@salesforce/soql-builder-ui';
 export const SOQL_BUILDER_UI_PATH =
   'node_modules/@salesforce/soql-builder-ui/dist';
 export const HTML_FILE = 'index.html';
-export const VIEW_TYPE = 'soqlCustom.soql';
+export const EDITOR_VIEW_TYPE = 'soqlCustom.soql';
+export const QUERY_DATA_VIEW_TYPE = 'queryDataView';
+export const QUERY_DATA_VIEW_PANEL_TITLE = 'SOQL Query Results';
```
jgellin-sf, line 17: //TODO: i18n

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -130,12 +138,39 @@ export class SOQLEditorInstance {
...
+        break;
+      }
       default: {
         console.log('message type is not supported');
       }
     }
   }
 
+  protected handleRunQuery(): Promise<void> {
+    const queryText = this.document.getText();
```
jgellin-sf, line 156: I think this code that runs the query and opens in the webview should be moved to it's own class so it can be accessed in a general way, since it will be useful outside the context of the SOQL Builder editor. If not moved in this PR, a comment to that effect should be added for future movement.

jgellin-sf (commented):
I have a few comments about shuffling some of this code around. I also wonder whether the right place for the HTML/CSS/JS webview artifacts belong in this extension package, or whether, like the SOQL Builder UI webview artifacts, belongs in a separate packages so that we can potentially access them outside the context of VS Code.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
jonnyhork: @jgellin-sf Please see the comments above

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
jonnyhork: There are two classes `QueryRunner` and `QueryDataViewService'. The runner is really a wrapper for the connection that also cleans the results for display. The query view service can be used to generate a table if given the data etc.
Is there something more general that your are looking for?
Eventually our extension might want to export some of these services to be used by other extensions.

packages/salesforcedx-vscode-soql/src/queryResultsView/media/queryDataView.css
jgellin-sf: Aha. Thanks.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
jgellin-sf: It's probably ok. I was thinking that the run-and-show code that you're handling here would be valuable in some independent class... RunAndShowQuery or some such thing. We can always circle back if we need to.

dobladez: I think we should soon extract the calls to the Salesforce platform from `soqlEditorInstance` into a separate module/class, including the calls to retrieve sobjects metadata, run query, etc.

dobladez (approved):
Great work!

packages/salesforcedx-vscode-soql/test/vscode-integration/queryDataView/queryDataViewService.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const queryRecords = mockQueryData;
+    const dataViewService = new QueryDataViewService(
+      mockSubscription,
+      queryRecords,
+      mockTextDocument
+    );
+
+    const postMessageSpy = sandbox.spy(mockWebviewPanel.webview, 'postMessage');
+    QueryDataViewService.extensionPath = '';
+    sandbox.stub(vscode.window, 'createWebviewPanel').returns(mockWebviewPanel);
```
brpowell, line 54: It doesn't look like `sandbox.restore()` is called to account for this stub.

packages/salesforcedx-vscode-soql/test/vscode-integration/testUtilities.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  ]
+};
+
+export function getMockConnection(
+  sandbox: SinonSandbox,
+  testUserName = 'test@test.com'
+) {
+  const mockAuthInfo = { test: 'test' };
+  const mockConnection = {
```
brpowell, line 62: Not necessary now but you could explicitly type this object as a Connection and ts-ignore it in just this function so that you don't have to do it in individual test cases when you use it.

brpowell (approved):
Looks good, just had two minor comments

PR #2569, opened 2020-09-23
refactor: move shared functionality between library based commands to commandlet

brpowell:
### What does this PR do?

Moves common logic between library based commands up to the `LibraryCommandletExecutor` base class. Motivation for doing this now is to address the issue of a delay in reporting progress when the command starts to give users faster feedback.

@W-8112824@

### Functionality Before
Executing a library based command with an org connection had a delay in progress feedback.

### Functionality After
Delay is gone.


packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -5,46 +5,64 @@
...
-  }
+  protected readonly telemetry = new TelemetryBuilder();
+  protected readonly revealChannelOutput = false;
+  protected abstract readonly logName: string;
+  protected abstract readonly executionName: string;
 
-  public setStartTime() {
-    this.startTime = process.hrtime();
-  }
+  protected abstract run(response: ContinueResponse<T>): Promise<boolean>;
```
brpowell, line 25: Non-CLI commands implement the `run` abstract method and return a `boolean` on whether or not the execution was successful.

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -5,46 +5,64 @@
...
+
+  public async execute(response: ContinueResponse<T>): Promise<void> {
```
brpowell, line 27: The base `execute` takes care of all the logistics around running a library based command:
- Recording execution time
- Force opening the output channel or not
- Showing the progress notification
- Sending any telemetry data set by the command
- Catching errors and logging/reporting/outputting them

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -130,100 +133,102 @@ export async function forceSourceDeployMultipleSourcePaths(uris: vscode.Uri[]) {
...
+      const deploy = this.doDeploy(await getConnection, components, namespace);
       const metadataCount = JSON.stringify(createComponentCount(components));
-      const result = await vscode.window.withProgress(
-        {
-          title: this.executionName,
-          location: vscode.ProgressLocation.Notification
-        },
-        () => deploy
-      );
+      this.telemetry.addProperty('metadataCount', metadataCount);
```
brpowell, line 157: If commands want to add extra telemetry data, call `addProperty`/`addMeasurement` using the base class' `TelemetryBuilder`

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -5,46 +5,64 @@
...
-      this.telemetryName,
-      this.startTime,
-      properties,
-      measurements
-    );
-  }
+  protected readonly telemetry = new TelemetryBuilder();
+  protected readonly revealChannelOutput = false;
+  protected abstract readonly logName: string;
+  protected abstract readonly executionName: string;
```
brpowell, line 23: Required properties to be set by commands.

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -126,41 +131,40 @@ export class AnonApexGatherer
...
-      await this.build(
-        nls.localize('apex_execute_text'),
-        nls.localize('force_apex_execute_library')
-      );
+    const result = await executeService.executeAnonymous({ apexFilePath, apexCode });
 
-      if (this.executeService === undefined) {
-        throw new Error('ExecuteService is not established');
-      }
+    const success = result.compiled && result.success;
```
AnanyaJha, line 149: do we need to check for both `compiled` and `success`? i think `success` is only true if there's no compile or runtime issues

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
AnanyaJha: i don't think we need to do this as part of this story, but we should also explore allowing the user to cancel the operation like we have with the other traditional commands

AnanyaJha (approved):

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
brpowell: Ah nice, yeah probably not. This was the logic taken from ApexLibraryExecutor

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
brpowell: Yeah cancellation is something we need to explore for running library commands in general. The problem is we need to make the promises cancellable at the library level

PR #2585, opened 2020-09-28
Update preview command to launch a native app or the browser

maliroteh-sf:
### What does this PR do?
Add ability to locally preview an LWC in a native mobile app or the mobile browser.

@W-8079566@

### Functionality Before
No way for the user to provide a native app for previewing an LWC (only option was to preview in mobile browser)

### Functionality After
User is presented with a dialog where entries are populated from a user-provided config file
Screen Shot 2020-09-28 at 12 45 14 PM


packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -155,7 +156,81 @@ export async function forceLightningLwcPreview(sourceUri: vscode.Uri) {
     return;
   }
 
-  await selectPlatformAndExecute(startTime, componentName);
```
maliroteh-sf, line 158: Rather than having a giant `selectPlatformAndExecute` function I've broken it down and refactored it into multiple functions (each responsible for doing part of the overall action)

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -407,9 +574,48 @@ async function selectPlatformAndExecute(
...
+}
+
+/**
+ * Given a path to a directory, returns a path that is one level up.
+ *
+ * @param directory path to a directory
+ * @returns path to a directory that is one level up, or undefined if cannot go one level up.
+ */
+function directoryLevelUp(directory: string): string | undefined {
+  const idx = directory.lastIndexOf(path.sep);
```
xyc, line 615: You could just use `path.dirname(directory)` here.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -347,28 +413,129 @@ async function selectPlatformAndExecute(
...
+): Promise<string | undefined> {
+  let targetApp: string | undefined = 'browser';
+  const options: vscode.QuickPickItem[] = [];
+
+  if (configFile === undefined) {
+    return targetApp;
+  }
+
+  try {
+    const fileContent = fs.readFileSync(configFile, 'utf8');
```
xyc, line 453: Can we check for file existence first? Although there's a try catch block, it's better to check first.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -407,9 +574,50 @@ async function selectPlatformAndExecute(
...
 }
+
+/**
+ * Given a path, it recursively goes through that directory and upwards, until it finds
+ * a config file named sfdx-project.json and returns the path to the folder containg it.
+ *
+ * @param startPath starting path to search for the config file.
+ * @returns the path to the folder containing the config file, or undefined if config file not found
+ */
+function getProjectRootDirectory(startPath: string): string | undefined {
```
xyc, line 591: We are currently only supporting SFDX project in a single workspace folder. So for most of our existing implementation look up the first folder in the workspace, but this is fine to me too. https://github.com/forcedotcom/salesforcedx-vscode/blob/1a3987f593eb053c5d461ef38beb73c28ae37c04/packages/salesforcedx-vscode-core/src/util/rootWorkspace.ts#L21

maliroteh-sf: Yes, we discussed in our team and thought that if we go about doing it the way I'm doing here, then in the future when we support multiple workspaces then this code should work too. So if this is fine by you then we'll leave it as is.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -155,7 +156,81 @@ export async function forceLightningLwcPreview(sourceUri: vscode.Uri) {
...
   }
 
-  await selectPlatformAndExecute(startTime, componentName);
+  await executePreview(startTime, componentName, resourcePath);
+}
+
+/**
+ * Performs the action of previewing the LWC. It takes care of prompting the user
+ * and gathering all info needed to preview the LWC. This includes prompting the user
+ * to select a platform, a target device, a target native app (or borwser), etc.
```
khawkins, line 165: Nit: s/borwser/browser

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -155,7 +156,81 @@ export async function forceLightningLwcPreview(sourceUri: vscode.Uri) {
...
+    );
+    return;
+  }
+
+  if (platformSelection.id === PreviewPlatformType.Desktop) {
+    await startServer(true, componentName, startTime);
+    return;
+  }
+
+  const commandCancelledMessage = nls.localize(
```
khawkins, line 191: This may as well move somewhere above [line 179](https://github.com/forcedotcom/salesforcedx-vscode/pull/2585/files#diff-785bdec339e64f40207b4e9dcae14dd4R179), since it's dereferenced there too.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -407,9 +574,50 @@ async function selectPlatformAndExecute(
...
+ * Given a path, it recursively goes through that directory and upwards, until it finds
+ * a config file named sfdx-project.json and returns the path to the folder containg it.
+ *
+ * @param startPath starting path to search for the config file.
+ * @returns the path to the folder containing the config file, or undefined if config file not found
+ */
+function getProjectRootDirectory(startPath: string): string | undefined {
+  const searchingForFile = 'sfdx-project.json';
+
+  let dir: string | undefined = path.parse(startPath).dir;
```
khawkins, line 594: This isn't going to be fully accurate if `startPath` itself is a folder—it fails in the case where you're in the project folder already. It works because we're typically going to be buried down a multi-folder path from `sfdx-project.json`, but technically speaking:
```javascript
console.log(`${path.parse('/my/component/folder').dir)}`);
// '/my/component'
```

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -347,28 +413,129 @@ async function selectPlatformAndExecute(
...
+  const options: vscode.QuickPickItem[] = [];
+
+  if (configFile === undefined || fs.existsSync(configFile) === false) {
+    return targetApp;
+  }
+
+  try {
+    const fileContent = fs.readFileSync(configFile, 'utf8');
+    const json = JSON.parse(fileContent);
+    const appDefinitionsForSelectedPlatform =
```
khawkins, line 455: Not necessarily in this PR, but we need to look at how we can expose the underlying functionality we have in the `lwc-dev-mobile` plugin to make management of the app configuration more robust here, like we do there.

@xyc maybe you know, is there a standard mechanism plugin developers use to share code that isn't otherwise exposed through the plugin interface, between their plugins and the Extension? My first thought is to have a shared Node module dependency between the two, e.g. `@salesforce/lwc-dev-mobile-core`, but I'm curious if there's another way.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -347,28 +413,129 @@ async function selectPlatformAndExecute(
...
+ * @param platformSelection the selected platform
+ * @param configFile path to a config file
+ * @returns the name of the selected device or undefined if user cancels selection.
+ */
+async function selectTargetApp(
+  platformSelection: PreviewQuickPickItem,
+  configFile: string | undefined
+): Promise<string | undefined> {
+  let targetApp: string | undefined = 'browser';
+  const options: vscode.QuickPickItem[] = [];
```
khawkins, line 446: Can we call this `items`? There's actually an optional `QuickPickOptions` argument to `showQuickPick` (below), and calling this `options` makes things kind of confusing.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -347,28 +413,129 @@ async function selectPlatformAndExecute(
...
+      const detail: string = app.id;
+      options.push({ label, detail });
+    });
+  } catch {
+    // silengtly fail and default to previewing on browser
+  }
+
+  // if there are any devices available, show a pick list.
+  if (options.length > 0) {
+    options.unshift({
```
khawkins, line 473: Can we make an explicit `browserItem` variable out of this, and then check its equality to `selectedItem` [below](https://github.com/forcedotcom/salesforcedx-vscode/pull/2585/files#diff-785bdec339e64f40207b4e9dcae14dd4R486), instead of comparing the label? Identifying a component by its label feels fragile.

khawkins (approved):
Just the minor suggestions (and future considerations) from me. This looks great over all!

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
maliroteh-sf: Nice catch

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -317,58 +380,158 @@ async function selectPlatformAndExecute(
...
-    targetName = selectedItem && selectedItem.label;
-
-    if (targetName === undefined) {
-      vscode.window.showInformationMessage(commandCancelledMessage);
-      return;
+    if (selectedItem === undefined) {
+      // user cancelled operation
+      return undefined;
+    } else {
+      targetName = selectedItem.label;
```
khawkins, line 402: This is incorrect, at least for Android. You need the actual `name` of the device. For example, when I select my Nexus 6 emulator, it passes `Nexus 6 API 29` as the `target` to the preview command, but `avd` is going to be looking for `Nexus_6_API_29`, and so it's always going to think that the device does not exist.

You'll have to look up the original `devices` entry, based on one or more of the selected fields, and grab the actual name of the device.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
xyc: @khawkins Yes. In fact, Node module dependencies are the directions we are moving towards to replace CLI commands. We have and are moving a lot of our dependencies on SFDX CLI to node modules, and make the extension solely dependent on node modules. This has brought us these benefits:
- Users don't have to install CLI and additional plugins. Quite a few users have configuration issues while installing CLI
- Significant performance improvement in commands. There is a heavy overhead when oclif looks up plugins.

Ideally if the mobile preview relies on a node module and not CLI plugin that would be amazing, but if part of the functionalities are moved to node module that's fine by me too.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
maliroteh-sf: ooooh... very good catch. In my local testing I was using an Android emulator named 'emu2' so user friendly display name and device name were the same so I never caught this.

maliroteh-sf:
creating a generic `QuickPickItem` with a comparison token might be a bit overkill here but I like the idea of extending `QuickPickItem` and adding another field. Will update my PR with that change.

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -70,6 +71,12 @@ interface PreviewQuickPickItem extends vscode.QuickPickItem {
   platformName: keyof typeof PlatformName;
 }
 
+export interface DeviceQuickPickItem extends vscode.QuickPickItem {
+  label: string;
```
khawkins, line 75: You don't need the properties that already exist in `QuickPickItem`. Just anything additional (i.e. `name`).

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -285,21 +357,24 @@ async function selectPlatformAndExecute(
...
 
+    // populate quick pick list of devices from the parsed JSON data
     if (isAndroid) {
       const devices: AndroidVirtualDevice[] = JSON.parse(jsonString)
         .result as AndroidVirtualDevice[];
       devices.forEach(device => {
         const label: string = device.displayName;
         const detail: string = `${device.target}, ${device.api}`;
-        options.push({ label, detail });
+        const name: string = device.name;
```
khawkins, line 367: 💯 Thanks! It feels cleaner and more deterministic. 👍

packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
maliroteh-sf: 🤦  classic newbie

maliroteh-sf:
@xyc I think I'm finally done making changes to this PR 😃 
Kevin has signed off on this so now we just need your blessings (whenever you get a chance).

xyc (approved):
@maliroteh-sf Looks good! Also thanks for adding more code coverage to the PR :) It meets our V2MOM goal of having 80% coverage on new features.

PR #2589, opened 2020-09-29
Merge Automation for Releases via CircleCi (@W-7938778)

sfsholden:
### What does this PR do?
1. Adds a pre-publish workflow for merging in the changes from the release branch into main.
1. Validates that the release branch provided by the user is not older than the version for main.
1. Adds a task and script for calling this pre-publish workflow.
1. Adds a merge step after a successful publish of main back into develop.

@W-7938778@


.circleci/config.yml
```diff
@@ -15,33 +15,85 @@ orbs:
   slack: circleci/slack@3.4.2
 
 parameters:
+  # Configurable Versions
```
sfsholden, line 18: Refactored - grouped some parameters together for better readability. The only new parameter is 'pre-publish'.

.circleci/config.yml
```diff
@@ -15,33 +15,85 @@ orbs:
...
-  create-release:
+  pre-publish:
     type: boolean
     default: false
-  cache-version:
-    type: string
-    default: 'dependency-cache-v4-'
 
 commands:
+  build-artifacts:
```
sfsholden, line 51: Refactor - reordered commands alphabetically.

.circleci/config.yml
```diff
@@ -77,7 +129,42 @@ commands:
...
+              exit 1;
+            fi
+
+  package-extensions:
+    steps:
+      - run:
+          name: 'Packaging extensions'
+          command: npm run vscode:package
+
+  prepare-gus-environment-variables:
```
sfsholden, line 167: Refactor - rename command.

.vscode/tasks.json
```diff
@@ -125,13 +163,21 @@
...
+      ],
       "default": "-v"
     },
     {
       "id": "releaseOverride",
       "type": "promptString",
       "description": "Release Version: Leave blank to grab the latest release, or provide the release manually (Ex 49.11.0)"
+    },
+    {
+      "id": "releaseBranch",
```
sfsholden, line 178: Future Improvement: Merge this argument and the one above together. Eventually I want to be able to grab the latest release branch like we do for the change log generator when a user doesn't explicitly specify it.

.circleci/config.yml
```diff
@@ -419,12 +487,37 @@ workflows:
...
-        - not: << pipeline.parameters.publish >>
         - not: << pipeline.parameters.create-release >>
+        - not: << pipeline.parameters.publish >>
+        - not: << pipeline.parameters.pre-publish >>
+    jobs:
+      - build-all
+      - unit-tests
+      - integration-tests
+
+  pre-publish-workflow:
```
sfsholden, line 498: For now this is going to be its own workflow separate from publish. I'd like to see if we run into any conflicts going forward prior to publishing. Right now this is very happy path.

.circleci/config.yml
```diff
@@ -77,7 +129,42 @@ commands:
...
 
-  gus-prepare-environment-variables:
+  install:
+    description: 'Installing dependencies'
+    steps:
+      - run: npm install
+      - run: npm --version
+      - run: node --version
+
+  merge-validation:
```
sfsholden, line 139: New command: merge-validation

.circleci/config.yml
```diff
@@ -331,6 +370,35 @@ jobs:
...
+      - ssh-config
+      - checkout
+      - prepare-release-environment-variables
+      - merge-validation
+      - run:
+          name: 'Rebasing main off of release branch'
+          command: |
+            git checkout main
+            echo "Rebasing main off of release branch ${RELEASE_VERSION}"
+            git rebase release/v${RELEASE_VERSION}
```
sfsholden, line 392: Explanation for using rebase: see [this](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) article

We want to use rebase when pulling in commits from the release branch because we do want to rewrite the git history. Everything that is new from the release branch should be put on "top" of what is currently in main.

.circleci/config.yml
```diff
@@ -90,24 +177,22 @@ commands:
           name: 'Prepare environment variables for GUS Change Case Management'
           command: |
             export GUS_BUILD="$(node -pe "require('<< parameters.package-json-path >>').version")"
-            echo "export RELEASE_VERSION=$GUS_BUILD" >> $BASH_ENV
```
sfsholden, line 93: This logic has moved to a new command 'prepare-release-environment-variables'.

.circleci/config.yml
```diff
@@ -331,6 +370,36 @@ jobs:
...
+      - checkout
+      - gh-config
+      - prepare-release-environment-variables
+      - merge-validation
+      - run:
+          name: 'Rebasing main off of release branch'
+          command: |
+            echo "Rebasing main off of release branch ${RELEASE_VERSION}"
+            git checkout main
+            git rebase -Xtheirs release/v${RELEASE_VERSION}
```
sfsholden, line 393: Originally I had this command as `git rebase release/v${RELEASE_VERSION}`. But I was able to reproduce a conflict that was causing issues yesterday evening. This appears to work, though I'm wondering if we want to handle conflicts manually for now - so we understand what is happening behind the scenes. I haven't been able to reproduce the conflict again, so I'm not 100% sure this is the resolution we need.

.circleci/config.yml
```diff
@@ -331,6 +370,36 @@ jobs:
...
+            git push -f
```
sfsholden, line 394: Unsure if we should really have this set as a force push too. I hadn't seen any issues with just `git push` until yesterday evening. The error I was coming across in my forked repo was:
`To github.com:sfsholden/salesforcedx-vscode-sholden.git
 ! [rejected]          main -> main (non-fast-forward)
error: failed to push some refs to 'git@github.com:sfsholden/salesforcedx-vscode-sholden.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart.`

This was the only way I found to get around it.

.circleci/config.yml
sfsholden: Also to address a potential question - why rebase vs. merge?

With rebase we are ensuring that all new commits are rewriting the commit history for the main branch. Everything specific to the release will be on "top" of everything in main. This will keep the history linear to reflect our releases.

I did also explore using "merge" as an option here. However, when it comes to this specific step it seems like there's a problem with the actual merge commit message that follows. There's no issue with merging main into develop (likely because of the rebase that happens here), it's specific to this phase of the process.

Example Error that was Seen:
`Auto-merging README.md
error: Terminal is dumb, but EDITOR unset
Not committing merge; use 'git commit' to complete the merge.`

There is likely a good way to resolve this if we still want to consider using merge instead of rebase here to be consistent. Since the rebase seems to work more reliably, I decided to just continue with what worked best.

.vscode/tasks.json
```diff
@@ -10,7 +10,10 @@
         "focus": false,
         "panel": "shared"
       },
-      "args": ["run", "bootstrap"],
+      "args": [
```
sfsholden, line 13: Prettier results.

@xyc I was noticing that your prettier results were different from mine. Do you know if this still conflicts with the results you were seeing? I don't think my prettier has any custom setup. I'm using the out of the box extension 'esbenp.prettier-vscode'.

.circleci/config.yml
AnanyaJha, line 393: what does `-Xtheirs` do here?

.circleci/config.yml
AnanyaJha: Makes sense to me, i've seen that force push is usually necessary for rebases

.circleci/config.yml
sfsholden: In the event of a conflict, -Xthiers should take the changes from the incoming branch (in this case the release branch) to override the conflict automatically.

.circleci/config.yml
lcampos: We should figure out a way of accomplishing this without a force push. We've disabled force pushes for the `develop` and `main` branches because of how destructive it could be. There are some ways of recovering from those type of issues but that relies on having access to the git info from the machine that ran the force push or having a backup somewhere else. Because of this I do not think force pushing from CircleCI is viable.

lcampos (changes requested):
We need to replace force push before merging this PR.

lcampos (approved):

.vscode/tasks.json
xyc: I don't have custom setup either. Would it be the same results if you format it from command palette: "Format Document With..." then select prettier? I noticed that if I manually format it with "JSON Language Features" it produces the same result as your change here. In the "Format Document With..." you can also select "Configure Default Formatter" to use prettier as the default

sfsholden: @xyc Thank you for pointing this out. I wasn't actually seeing that command palette option before. Reinstalling the plugin fixed the issue for me.

PR #2610, opened 2020-10-08
chore: add context restriction for circle ci workflows

sfsholden:
### What does this PR do?
Restricts approval access for circle ci workflows. Only members of the GitHub Team 'PDT' should be able to approve workflows for publishing, creating release branches, or running pre-publish steps (merging).

@W-8165486@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -516,6 +516,7 @@ workflows:
           type: approval
       - merge-release-branch:
           e: default-executor
+          context: pdt-publish-restricted-context
           requires:
             - build-all
             - unit-tests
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -564,6 +566,7 @@ workflows:
           type: approval
       - create-release-branch:
           e: default-executor
+          context: pdt-publish-restricted-context
```
lcampos, line 569: How does this work for scheduled jobs ?

sfsholden: It shouldn't block scheduled jobs. 

This would only add restrictions to the scheduled job if we included this context line in the workflow 'cut-release' under 'create-release-branch'. That isn't something we want.

AnanyaJha (approved):

lcampos (approved):

PR #2612, opened 2020-10-08
test: fix core integration tests

xyc:
### What does this PR do?
Fix core integration test suite.

@LisandroFernandezSF @lcampos fyi, I noticed that core integration tests in circleci builds in `develop` branch are failing starting from https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/2476/workflows/562cc1d5-4d3f-4fbe-a7d8-03460f15c9ae/jobs/13506/steps. However CircleCI is not reporting the test failures. Instead it's showing fewer tests (335 tests, which should be 800+) passing.

@<Insert GUS WI>@

### DIFF
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
@@ -47,20 +47,14 @@ describe('Force Source Retrieve', () => {
 });
 
 describe('Force Source Retrieve and open', () => {
-  const sb: SinonSandbox = createSandbox();
+  let sb: SinonSandbox;
   const openAfterRetrieve: boolean = true;
   const forceSourceRetrieveExec = new ForceSourceRetrieveExecutor(
     new TestDescriber(),
     openAfterRetrieve
   );
-  const openTextDocumentStub: SinonStub = sb.stub(
-    vscode.workspace,
-    'openTextDocument'
-  );
-  const showTextDocumentStub: SinonStub = sb.stub(
-    vscode.window,
-    'showTextDocument'
-  );
+  let openTextDocumentStub: SinonStub;
+  let showTextDocumentStub: SinonStub;
   const resultData = `{
     "status": 0,
     "result": {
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
```diff
@@ -71,11 +65,18 @@ describe('Force Source Retrieve and open', () => {
...
           "filePath": "force-app/main/default/classes/TestClass.cls"}
   ] }}`;
-  const getCmdResultStub: SinonStub = sb
-    .stub(CommandOutput.prototype, 'getCmdResult')
-    .returns(resultData);
+  let getCmdResultStub: SinonStub;
 
-  afterEach(async () => {
+  beforeEach(() => {
+    sb = createSandbox();
```
xyc, line 71: @LisandroFernandezSF I think the caveat is sandbox and stubs should be created **inside** beforeEach.

lcampos (approved):

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
brpowell: I would be surprised if this was the issue, I've created sandboxes outside of these contexts without a problem. Sandboxes only need to be created once.

brpowell: Oh but the actual stubbing i think you're right 👍

xyc:
Created W-8190289 for test failures not surfacing in CircleCI

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
xyc: yeah, I think the bigger issue here is why core integration test failures are not surfaced in Circle CI. I created W-8190289 for it. @brpowell I remember you also mentioned false positive in core tests not long ago.

LisandroFernandezSF: @brpowell @xyc Do you want me to move the `let sb: SinonSandbox;` inside the beforeEach function?

brpowell: @xyc yeah that is alarming especially in this case when there were very clearly test failures. In the case that I experienced the test runner didn't even initialize so that made a little more sense. @LisandroFernandezSF not essential just a tip that it can be initialized once, up to you.

PR #2617, opened 2020-10-09
Jhork: Save Query Results JSON/CSV

jonnyhork:
### What does this PR do?
This PR will allow users to save the soql query results displayed in the webview as either csv or json format.

A file will be saved with the same name as the `.soql` file with the corresponding format extension added.
**Example:**
`zooAnimals.soql`---> `zooAnimals.soql.csv`

- The new files will be saved in a `soql-results` directory at `scripts/soql` in an sfdx-project.
- If there is already results from a given .soql file saved, the data will be overwritten unless the user changes the name of the previously created results file. So only one file per query of each format will be created unless the user changes the name.
- When the user clicks to save a the data, the generated file will be revealed in the file explorer.

@W-8136966@

### Functionality Before

Query Data Webview - low

### Functionality After
Save File


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -628,6 +628,7 @@ export async function activate(context: vscode.ExtensionContext) {
     EmptyParametersGatherer,
     getDefaultUsernameOrAlias,
     getUserId,
+    getRootWorkspacePath,
```
jonnyhork, line 631: I originally wrote almost the same function in our extension but decided it would be more dry to reuse the one in core.

packages/salesforcedx-vscode-soql/src/editor/htmlUtils.ts
```diff
@@ -84,7 +84,7 @@ export class HtmlUtils {
   ): string {
     const cspMetaTag = `<meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'none';
+      content="default-src 'self';
```
jonnyhork, line 87: I thought the CSP was preventing scripts for the table view form being loaded during testing, it may have just been a race condition but I think adding `self` is more secure anyway. I'm not a CSP expert though.

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -11,10 +11,7 @@ import { QueryResult } from 'jsforce';
 import * as vscode from 'vscode';
 
 export class QueryRunner {
-  constructor(
-    private connection: Connection,
-    private document?: vscode.TextDocument
```
jonnyhork, line 16: This was left over from previous PR when I thought the Query runner would also save files.

packages/salesforcedx-vscode-soql/src/queryDataView/dataProviders/csvDataProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class CsvDataProvider implements DataProvider {
+  public readonly fileExtension = DATA_CSV_EXT;
+  constructor(public readonly documentName: string) {}
+
+  public getFileContent(data: JsonMap[]): string {
+    const queryRecordsCsv = Papa.unparse(data, {
+      header: true,
+      delimiter: ','
+    });
```
jonnyhork, line 21: [PapaParse](https://www.papaparse.com/docs#json-to-csv)! (love the name)

If an error is thrown due to bad JSON format it will be caught by the caller.
This will produce something like:
```
Column 1,Column 2
1-1,1-2
2-1,2-2
3-1,3-2
4,5
```

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -0,0 +1,105 @@
...
+  protected getDataProvider(): DataProvider {
+    switch (this.format) {
+      case FileFormat.CSV:
+        return new CsvDataProvider(this.documentName);
+      case FileFormat.JSON:
+        return new JsonDataProvider(this.documentName);
+      default:
+        throw new Error('No DataProvider Found');
+    }
+  }
```
jonnyhork, line 53: @jgellin-sf I tried to do it the cool way as hash look-up but TS was fighting me on it so I moved on. The switch seems much simpler. I would love to get your help later maybe to see if I was on the right track for the alternative.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -0,0 +1,105 @@
...
+      // Save query results to disk
+      fs.writeFileSync(queryDataFilePath, fileContent);
+      this.showSaveSuccessMessage(savedFileName);
+      this.showFileInExplorer(queryDataFilePath);
+
+      return queryDataFilePath;
+    } catch (error) {
+      // TODO: i18n, CCX
+      vscode.window.showErrorMessage(
+        `Your data could not be saved. Run the query and try again.`
```
jonnyhork, line 76: @ekapner This message will be shown to the users if something goes wrong either creating the directory to store files or converting the json data to csv or Json data to a string.
Either way there is not much the customer can do about it except verify there is no error or something weird with the SOQL query itself or so API change/issue.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private getResultsDirectoryPath() {
+    return path.join(getRootWorkspacePath(), QUERY_RESULTS_DIR_PATH);
+  }
+
+  private createResultsDirectoryIfDoesNotExist() {
+    fs.mkdirSync(this.getResultsDirectoryPath(), {
+      recursive: true
+    });
```
jonnyhork, line 89: Using the recursive option will create the directory if it does not exist already. So there is no need to check before creating it. Also, node [does not recommend](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) you check for existence anyway.

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -71,20 +80,49 @@ export class QueryDataViewService {
...
-    return webview;
+    return this.currentPanel.webview;
   }
 
-  private getWebViewContent(webview: vscode.Webview): string {
-    let _html: string;
+  protected onDidRecieveMessageHandler(message: DataViewEvent) {
+    const { type, format } = message;
+    switch (type) {
+      case 'activate':
```
jonnyhork, line 99: This will address a race condition when loading the webview and the table ui, similar to how we do it with the soql builder UI.

packages/salesforcedx-vscode-core/src/index.ts
xyc: We are going to move this function to utils module soon, so it should be fine if you use a copy of that.

xyc: The goal is not to depend on the core extension

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -37,11 +46,8 @@ export class QueryDataViewService {
     QueryDataViewService.extensionPath = context.extensionPath;
   }
 
-  private updateWebviewWith(
-    webview: vscode.Webview,
-    queryData: QueryResult<JsonMap>
-  ) {
-    webview.postMessage({
+  private updateWebviewWith(queryData: QueryResult<JsonMap>) {
+    this.currentPanel!.webview.postMessage({
```
xyc, line 50: is `!` necessary here?

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -42,6 +39,7 @@ export class QueryRunner {
   private flattenQueryRecords(rawQueryRecords: JsonMap[]) {
     // filter out the attributes key
     return rawQueryRecords.map(
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
AnanyaJha, line 42: nit: do we need the eslint disable here?

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
jonnyhork: Yeah, just because of the way the webview lifecycle works, we can't instantiate a webview with out displaying it. So the current panel is assigned with `createOrShowWebview()` [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/8e4b0955b7cc6bb108baa6e9bae0fbb8006c0cd0/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts#L58)

Instead of when the new dataView class is created

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: Ts lint did not like that we were ignoring the `attributes` variable that is declared but not used. I think its just a symptom of the fancy destructuring syntax, where the intention _is_ to ignore the attributes variable.
`({ attributes, ...cleanRecords }) => cleanRecords`

packages/salesforcedx-vscode-core/src/index.ts
jonnyhork: ok, I can refactor a bit to not depend on the core extension.

jonnyhork: @xyc  So to clarify are you saying I should write my own getRootWorkspacePath function in our extension, or that it is fine now to use core but we should use the Utils module when its created?

It would be nice to not duplicate functions but in this case maybe best if we just have our own so we don't depend on core or have to make changes later.

xyc: I think if you copied the `getRootWorkspacePath` it would be fine. Otherwise, we'd still need to remove the exports in core here. In terms of amount of changes we need to make it's similar for both approaches, but I would like us not to export more utilities methods like this from core.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
xyc: Got it, but do you want to change it to optional chaining to avoid NPEs?
```suggestion
    this.currentPanel?.webview.postMessage({
```

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
xyc: You could configure a `argsIgnorePattern` for this rule to ignore args started with underscore https://eslint.org/docs/rules/no-unused-vars#argsignorepattern

Also I don't think we enforce this in other extensions.

packages/salesforcedx-vscode-core/src/index.ts
jonnyhork: got it, I'll fix it =)

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: cool, I'll try that

jonnyhork: or rather I may remove the rule, the underscore option wont work for me in this case as the variable needs to match the key name

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
ekapner: How about this, @jonnyhork? 

We couldn't save your data. Verify the query syntax, then run the query again.

jonnyhork: Sounds good to me

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
jonnyhork: yeah sounds good to me.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -0,0 +1,105 @@
...
+    vscode.commands.executeCommand(
+      'revealInExplorer',
+      vscode.Uri.parse(targetPath)
+    );
+  }
+
+  private showSaveSuccessMessage(savedFileName: string) {
+    vscode.window.showInformationMessage(
+      // TODO: i18n and CCX
+      `Your data has been saved in this workspace as: ${savedFileName}`
```
jonnyhork, line 102: @ekapner Sorry I forgot to mention you here.

This message will be shown to the user when the file is successfully saved in the `query-results` directory.
I was thinking to at least change from "data" to "results" to be consistant.
> `Your results have been saved in this workspace as: ${savedFileName}`

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
ekapner: Per our convo in Slack:

We couldn't save your results. Verify the query syntax, then run the query again.

ekapner: I'm thinking "the" sounds better than "your"

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
ekapner: @jonnyhork - should we say "query-results" folder instead of "workspace"?

We saved the results in the query-results folder as: ${savedFileName}

jonnyhork: yeah thats fine with me

packages/salesforcedx-vscode-core/src/index.ts
jonnyhork: fixed in the `fix(soql): remove dependency from core extension in fileservice` commit.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
jonnyhork: okey dokey

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
ekapner: @jonnyhork - changed my mind. :)
The results were saved in the query-results folder as: $...

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
jonnyhork: Addressed in the `fix(soql): update eslint rules` commit

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -0,0 +1,98 @@
...
+      );
+      const savedFileName = this.dataProvider.getFileName();
+      const queryDataFilePath = path.join(
+        this.getResultsDirectoryPath(),
+        savedFileName
+      );
+
+      this.createResultsDirectoryIfDoesNotExist();
+      // Save query results to disk
+      fs.writeFileSync(queryDataFilePath, fileContent);
```
brpowell, line 61: Consider possibly using `vscode.workspace.fs` methods for fs operations.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
```diff
@@ -0,0 +1,40 @@
...
+  We need to replace the hrefs with webviewUris,
+  this will need to change once we need a standalone data view.
+   */
+  html = HtmlUtils.replaceCspMetaTag(html, webview);
+  html = html.replace('${tabulatorStyleUri}', tabulatorStyleUri.toString());
+  html = html.replace('${baseStyleUri}', baseStyleUri.toString());
+  html = html.replace('${tabulatorUri}', tabulatorUri.toString());
+  html = html.replace('${viewControllerUri}', viewControllerUri.toString());
+
+  return html;
```
brpowell, line 39: Do we need to take care of sanitizing the document to ensure it's safe to load?

packages/salesforcedx-vscode-soql/test/vscode-integration/queryDataView/queryDataFileService.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+describe('Query Data File Service', () => {
+  const documentName = 'example.soql';
+  const workspacePath = vscode.workspace.workspaceFolders![0].uri.fsPath;
+  const testResultsDirPath = path.join(workspacePath, QUERY_RESULTS_DIR_PATH);
+
+  afterEach(() => {
+    // delete the query-results directory and its files.
+    // @ts-ignore
+    fs.rmdirSync(testResultsDirPath, { recursive: true });
```
brpowell, line 28: Does this need to be ts-ignored?

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
jonnyhork: I was trying to decide wether to use the VS code fs API actually, but I did not see a distinct advantage, in fact it seemed more verbose because I had to convert everything to vs code URI etc.

Is there an advantage to using `vscode.workspace.fs`?

Ultimately though, I decided to use the regular file system module because our goal is to build this as both a stand alone web app and VS code extension, so we want to limit any dependency on VS code APIs

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
jonnyhork: Well there are 
Content Security Protocols in place on the document to only load scripts from `self` 
```
const cspMetaTag = `<meta
      http-equiv="Content-Security-Policy"
      content="default-src 'self';
      img-src ${webview.cspSource} https:;
      script-src ${webview.cspSource};
      style-src 'unsafe-inline' ${webview.cspSource};"
    />`;
```
As well as restrictions on the webview to only load assets from a specified source
```
localResourceRoots: [
          vscode.Uri.file(
            path.join(
              QueryDataViewService.extensionPath,
              DATA_VIEW_RESOURCE_ROOTS_PATH
            )
          )
        ],
```
There is no user input in the webview.
I'm not sure if that covers our concerns or not?

packages/salesforcedx-vscode-soql/test/vscode-integration/queryDataView/queryDataFileService.test.ts
jonnyhork: Yeah for some reason it does, maybe the version of TS we run doesn't have the correct types for `fs`?

[fs.rmdirSync() has an options property](https://nodejs.org/api/fs.html#fs_fs_rmdirsync_path_options) but TS thinks it only take one argument. It worked as expected and I did't dig much deeper into why TS was yelling at me in the test.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
brpowell: Yeah not entirely sure either, might be worth reaching out to Nick. Like in this case even if there is no user input, what if the document at `const pathToHtml = path.join(pathToDataViewDist, HTML_FILE);` was replaced with a malicious one and then loaded here `let html = fs.readFileSync(pathToHtml).toString()`

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
brpowell: The main advantage is that it's flexible with the file system being used, whether it's local or some remote container depending on the context the extensions are loaded in. If it's going to be a standalone app at some point the NodeJS fs dependency will have to be removed as well. I don't think it matters either way at this point, but worth considering.

packages/salesforcedx-vscode-soql/test/vscode-integration/queryDataView/queryDataFileService.test.ts
brpowell: Gotcah. Might be because the @types/node version is out of sync with the target node version

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
xyc: @brpowell I think that should be less of our concerns. If a malicious actor has access to the html file, it also has access to the extension executable JavaScript files, basically we couldn't guarantee anything we are checking here. CSP and no user input should be fine for our use cases here.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
jonnyhork: Oh yeah, good point about the node dependency, node does not equal browser, ha. We aren't sure what the standalone app will look like yet, we'll cross that bridge when we get there.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
jonnyhork: Thanks for mentioning it Bryan and for the input @xyc, a lot of the consideration for vs code webview stuff is new to me as well.

dehru (approved):

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -89,7 +89,7 @@ import { isSfdxProjectOpened } from './predicates';
 import { registerPushOrDeployOnSave, sfdxCoreSettings } from './settings';
 import { taskViewService } from './statuses';
 import { telemetryService } from './telemetry';
-import { isCLIInstalled } from './util';
+import { getRootWorkspacePath, isCLIInstalled } from './util';
```
xyc, line 92: Could we undo the change of this file since we don't use `getRootWorkspacePath` from core now?

brpowell (approved):

xyc (approved):

packages/salesforcedx-vscode-core/src/index.ts
jonnyhork: yes, thanks, I forgot to do that when I refactored the core dependency stuff, thanks for reminding me. See the `fix(soql): clean up unused exports and comment` commit

PR #2625, opened 2020-10-13
channel service in utils

lcampos:
### What does this PR do?
Move channel service into utils module. Use the utils channel service in lwc extension

FYI, utils tests are still being worked on

@W-8183064@


packages/salesforcedx-utils-vscode/src/channels/channelService.ts
```diff
@@ -0,0 +1,108 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import stripAnsi from 'strip-ansi';
+import * as vscode from 'vscode';
+import { nls } from '../messages';
+
+import { CommandExecution } from '../cli';
+
+export const DEFAULT_SFDX_CHANNEL = vscode.window.createOutputChannel(
```
xyc, line 14: Per discussions with @lcampos offline, we need to include this in an init function to create the output channel in extension init, since extensions using the utils channel service will have its own channel.

xyc (approved):

PR #2635, opened 2020-10-15
Lf/retrieve and open doesnt handle types with xml only files

LisandroFernandezSF:
### What does this PR do?
When the 'Retrieve and open' feature is performed and there is no file extension defined, it opens the '-meta.xml' file by default.

@W-8193920@

### Functionality Before
When there wasn't a definition on MetadataDictionary the feature didn't open any file 

### Functionality After
When there isn't a definition on MetadataDictionary,  the feature opens '-meta.xml' file by default.


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.ts
```diff
@@ -104,15 +107,27 @@ export class ForceSourceRetrieveExecutor extends SfdxCommandletExecutor<
...
-          vscode.window.showTextDocument(document, showOptions);
+      const filesToOpen = [];
+      if (extensions) {
+        for (const ext of extensions) {
+          const tmpFile = resultJson.result.inboundFiles.find(
+            ({ filePath }: { filePath: string }) => filePath.endsWith(ext)
+          );
+          filesToOpen.push(path.join(getRootWorkspacePath(), tmpFile.filePath));
         }
+      } else {
```
lcampos, line 118: Do we have any tests covering this ?

LisandroFernandezSF: Done!

lcampos (approved):
🚢

PR #2640, opened 2020-10-16
Improve @AuraEnabled for better error handling.

PawelWozniak:
### What does this PR do?
Improved @AuraEnabled snippet by adding try catch block which catch apex exception and throw AuraHandledException which is properly handled by LWC and show error to user instead of ugly internal server error message.
Inspired by https://developer.salesforce.com/blogs/2017/09/error-handling-best-practices-lightning-apex.html

@W-8235555@

### Functionality Before
In case of apex exception thrown it will show ugly internal error to user. User don't know what to do.

### Functionality After
In case of apex exception thrown it is catch and re thrown as AuraHandledException which will show popup to user.


lcampos:
This is awesome, thanks for the contribution @PawelWozniak

PR #2643, opened 2020-10-19
fix: code changes needed for webpack on utils

lcampos:
### What does this PR do?
Updating code references (mainly) before re-introducing the use of webpack on the utils module

@W-8183039@


packages/salesforcedx-utils-vscode/src/cli/cliConfiguration.ts
```diff
@@ -28,15 +26,6 @@ export function isCLIInstalled(): boolean {
   return isInstalled;
 }
 
-export function showCLINotInstalledMessage() {
```
lcampos, line 31: This code is not being used so I'm removing it for now.

packages/salesforcedx-utils-vscode/src/workspaces/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
```
lcampos, line 1: With webpack we make the assumption that a module's entry point will be its index.ts file.

packages/salesforcedx-utils-vscode/src/telemetry/telemetryReporter.ts
```diff
@@ -8,12 +8,12 @@ import * as appInsights from 'applicationinsights';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import * as vscode from 'vscode';
```
xyc, line 11: so this doesn't work for webpack? `import * as vscode from 'vscode'`

xyc (approved):

AnanyaJha (approved):

packages/salesforcedx-utils-vscode/src/telemetry/telemetryReporter.ts
lcampos: This change wasn't specifically for webpack but for finding which things need to be mocked when writing the tests.

PR #2644, opened 2020-10-19
feat: Consume the new force:apex:log:list command in VS Code @W-7561156@

AnanyaJha:
### What does this PR do?
This PR introduces listing logs using the Apex Library if the setting is turned on. 

@W-7561156@

### Functionality Before
CLI text
Screen Shot 2020-10-19 at 2 34 16 PM

### Functionality After
Apex library text
Screen Shot 2020-10-26 at 3 37 39 PM

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -141,7 +141,6 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<ApexEx
 
   protected async run(response: ContinueResponse<ApexExecuteParameters>): Promise<boolean> {
     const connection = await workspaceContext.getConnection();
-    // @ts-ignore
     const executeService = new ExecuteService(connection);
     const { apexCode, fileName: apexFilePath } = response.data;
 
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -26,7 +26,6 @@ import { getRootWorkspacePath, hasRootWorkspace } from '../util';
 import {
   CommandletExecutor,
   LibraryCommandletExecutor,
-  LibraryExecution,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
@@ -140,7 +139,7 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<ApexEx
     'apex-errors'
   );
 
-  protected async run(response: ContinueResponse<ApexExecuteParameters>): Promise<LibraryExecution> {
+  protected async run(response: ContinueResponse<ApexExecuteParameters>): Promise<boolean> {
     const connection = await workspaceContext.getConnection();
     // @ts-ignore
     const executeService = new ExecuteService(connection);
@@ -165,7 +164,7 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<ApexEx
       handleApexLibraryDiagnostics(result, this.diagnostics, filePath);
     }
 
-    return { success };
+    return success;
   }
 }
 
--- a/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
@@ -26,6 +26,7 @@ import { getRootWorkspacePath, hasRootWorkspace } from '../util';
 import {
   CommandletExecutor,
   LibraryCommandletExecutor,
+  LibraryExecution,
   SfdxCommandlet,
   SfdxCommandletExecutor,
   SfdxWorkspaceChecker
@@ -131,7 +132,9 @@ export class AnonApexGatherer
   }
 }
 
-export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<ApexExecuteParameters> {
+export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<
+  ApexExecuteParameters
+> {
   protected executionName = nls.localize('apex_execute_text');
   protected logName = 'force_apex_execute_library';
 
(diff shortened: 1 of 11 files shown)


AnanyaJha:
The PR still needs to be updated with the new version of the library once it gets published, ignore the `@ts-ignores` for now

packages/salesforcedx-vscode-core/src/commands/forceApexExecute.ts
```diff
@@ -139,12 +140,16 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<ApexEx
     'apex-errors'
   );
 
-  protected async run(response: ContinueResponse<ApexExecuteParameters>): Promise<boolean> {
+  protected async run(response: ContinueResponse<ApexExecuteParameters>): Promise<LibraryExecution> {
     const connection = await workspaceContext.getConnection();
+    // @ts-ignore
```
AnanyaJha, line 145: the `@ts-ignore`s will be removed

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -101,7 +103,8 @@ export class ForceApexLogGetExecutor extends SfdxCommandletExecutor<
       const localUTCDate = new Date(response.data.startTime);
       const date = getYYYYMMddHHmmssDateFormat(localUTCDate);
       const logPath = path.join(logDir, `${response.data.id}_${date}.log`);
-      fs.writeFileSync(logPath, resultJson.result.log);
+      // this needs to be result[0] for this to work correctly but idk if that used to be necessary
+      fs.writeFileSync(logPath, resultJson.result[0].log);
```
AnanyaJha, line 107: will remove this, related to a bug we discovered

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
```diff
@@ -9,22 +9,28 @@ import * as vscode from 'vscode';
...
+
+export abstract class LibraryCommandletExecutor<T, U = void>
   implements CommandletExecutor<T> {
   protected readonly telemetry = new TelemetryBuilder();
   protected readonly revealChannelOutput = false;
   protected abstract readonly logName: string;
   protected abstract readonly executionName: string;
 
-  protected abstract run(response: ContinueResponse<T>): Promise<boolean>;
+  protected abstract run(
```
AnanyaJha, line 27: Library commands will now return a `LibraryExecution` object that will only include the `success` property by default. If the result object is needed from a command, then that can be specified when the individual executor is declared and it will be added as a property to the `LibraryExecution` result

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -243,8 +243,8 @@ export const messages = {
   force_apex_log_get_no_logs_text: 'No Apex debug logs were found',
   force_apex_log_get_pick_log_text: 'Pick an Apex debug log to get',
   force_apex_log_list_text: 'Getting Apex debug logs',
-  force_apex_log_get_library: 'Apex Library: Get Logs',
```
AnanyaJha, line 246: wasn't being used

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
brpowell: Sorry I'm not following, where is a result needed outside of the context of `run`?

AnanyaJha: For example, we use the response from the `force:apex:log:list` command in the `parameterGatherer` for the `force:apex:log:get` command [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/2757f2a94a3e8f24efd7ab9c967df62700e1b084/packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts#L144)

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
```diff
@@ -213,23 +223,48 @@ export class ForceApexLogList {
...
+  protected executionName: string = nls.localize('apex_log_list_text');
+  protected logName: string = 'force_apex_log_list_library';
+
+  protected async run(
+    response: ContinueResponse<{}>
+  ): Promise<LibraryExecution<LogRecord[]>> {
+    const connection = await workspaceContext.getConnection();
+    // @ts-ignore
+    const logService = new LogService(connection);
+    const logs = await logService.getLogRecords();
```
lcampos, line 265: This is part of the changes that will be released this week on the apex library. Which means we also need to update the version of it on the package.json and hold on merging the PR until that release happens

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
lcampos: @AnanyaJha is this the fix for the `undefined` error we've been discussing ?

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
brpowell: Since the libraries give us freedom from CLI shelling now, what if we used `logService.getLogRecords()` directly in the gatherer?

packages/salesforcedx-vscode-core/src/commands/forceApexLogGet.ts
AnanyaJha: Yeah that's the fix, but I'm still looking into what might have caused it

packages/salesforcedx-vscode-core/src/commands/util/libraryCommandlet.ts
AnanyaJha: Yeah we totally could do that too! I was thinking the main benefit with using the executors is that in case running `getLogRecords()` takes a long time, the user at least sees the progress bar letting them know that the command is running. Otherwise if we call `getLogRecords()` directly, they may just have a blank screen for a bit

brpowell: What about adding this to the gatherer and calling it in `gather()`?

```typescript
private async getLogRecords(): Promise<LogRecord[]> {
  const connection = await workspaceContext.getConnection();
  const logService = new LogService(connection);
  return vscode.window.withProgress({
    location: vscode.ProgressLocation.Notification,
    title: "Fetching log list"
  }, () => logService.getLogRecords();
}
```

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -243,8 +243,8 @@ export const messages = {
...
   apex_log_get_text: 'Get Apex Logs',
+  apex_log_list_text: 'Fetching Apex debug log list',
```
AnanyaJha, line 247: @ekapner let me know what you think of this text! This will be displayed when the user is waiting for a list of their debug logs to be displayed

ekapner: @AnanyaJha - yeah, I think we've used "fetching" before - so that would be consistent (if I'm remembering correctly). So, this looks good.

rcoringrato-sfdc (approved):

lcampos (approved):

PR #2655, opened 2020-10-22
Use webpack for salesforcedx-utils-vscode module

lcampos:
### What does this PR do?
Reintroduces the use of webpack on the utils module. Some code had to be moved around so the result of bundling would allow us to correctly compile it.

@W-8252102@


packages/salesforcedx-utils-vscode/package.json
```diff
@@ -43,7 +43,7 @@
     "vscode": "^1.40.0"
   },
   "scripts": {
-    "compile": "tsc -p ./",
+    "compile": "tsc -p ./ && webpack",
```
lcampos, line 46: Doing this so we only webpack the code under the src folder.

packages/salesforcedx-utils-vscode/src/telemetry/cliConfiguration.ts
```diff
@@ -5,31 +5,12 @@
...
-import { GlobalCliEnvironment } from './commandExecutor';
+import { GlobalCliEnvironment } from '../cli';
 import { ConfigUtil } from './configUtil';
 
 export const ENV_SFDX_DISABLE_TELEMETRY = 'SFDX_DISABLE_TELEMETRY';
-export const SFDX_CLI_DOWNLOAD_LINK =
-  'https://developer.salesforce.com/tools/sfdxcli';
 export const SFDX_CONFIG_DISABLE_TELEMETRY = 'disableTelemetry';
 
-export function isCLIInstalled(): boolean {
```
lcampos, line 17: This code was not being used right now so I've removed it, we might add it in the future if needed.

packages/salesforcedx-utils-vscode/src/telemetry/cliConfiguration.ts
```diff
@@ -5,31 +5,12 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { which } from 'shelljs';
```
lcampos, line 8: Moved this to the telemetry module cause it depends on `ConfigUtil` which has a `vscode` dependency. When bundling the code, it was causing unit tests in other modules to fail since they couldn't resolve that dependency. We might move this and configUtil back to the cli module once we sort out those type of problems.

packages/salesforcedx-utils-vscode/src/telemetry/index.ts
```diff
@@ -4,6 +4,5 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { TelemetryService } from './telemetry';
-export const telemetryService = TelemetryService.getInstance();
-export { TelemetryBuilder, TelemetryData, Properties, Measurements } from './telemetry';
+
+export { TelemetryService, TelemetryBuilder, TelemetryData, Properties, Measurements } from './telemetry';
```
lcampos, line 8: I had to remove `TelemetryService.getInstance()` from here cause it was causing webpack to produce some duplicate imports. We'll take a second look at this in the next days but for now I've made the necessary updates for this to work.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -79,10 +80,14 @@ export class TelemetryService {
    */
   public async initializeService(
     context: ExtensionContext,
-    extensionName: string
+    extensionName: string,
+    aiKey: string,
```
lcampos, line 84: Updated this so the consumer is the one providing they key and version data since bundling a dynamic require that reads the `package.json` file was a problem and couldn't find a better solution.

packages/salesforcedx-utils-vscode/test/unit/telemetry/telemetry.test.ts
```diff
@@ -95,15 +96,15 @@ describe('Telemetry production mode', () => {
     sb.restore();
   });
 
-  it('Should send telemetry data', async () => {
+  xit('Should send telemetry data', async () => {
```
lcampos, line 99: I had to ignore several of these tests because I couldn't make proxyquire replace the TelemetryReported code with a mock once the code is bundled. I think this will be easier to do if we start using `@salesforce/telemetry` but I haven't tried that yet.

packages/salesforcedx-utils-vscode/webpack.config.js
```diff
@@ -64,6 +45,8 @@ module.exports = {
   // excluding dependencies from getting bundled
   externals: {
     '@salesforce/core': 'commonjs @salesforce/core',
+    applicationinsights: 'commonjs applicationinsights',
```
lcampos, line 48: These two dependencies cannot be bundled so we're ignoring them and expecting them to be available in `node_modules`

packages/salesforcedx-utils-vscode/package.json
brpowell: Does there need to be both a `tsc` and `webpack` call if webpack uses the ts-loader?

lcampos: Yes because I couldn't find a way to configure webpack to not bundle the code under `test` since tsloader uses the tsconfig.json file too. So this allows us to compile the tests with the normal typescript compiler and the code under `src` with tsloader.

brpowell: Gotcha 👍  sounds good.

brpowell (approved):

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
xyc: do we want to make `aiKey` a constant?

packages/salesforcedx-utils-vscode/package.json
xyc: Did we try excluding the test folder? https://stackoverflow.com/questions/52059366/how-to-exclude-directory-from-getting-bundled-by-webpack

Although I think if webpack doesn't bundle the code we still need to get tests compiled

lcampos: Yeah, if you look at the current webpack config I'm not mapping the test code anymore. However, since tsloader uses tsconfig.json it still picks up the test code. I played with having two tsconfigs; one for `src` which would be used by webpack, the other for `test` and used by typescript. That started to get complex real fast so I opted for going this route.

xyc (approved):

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
lcampos: We can, I just moved it to the extension's package.json since we are already reading it to get the version data. I don't have a strong opinion on either option.

Issue #2608, opened 2020-10-08
Problems tab does not refresh when issues are addressed


robcowell:
### Summary

Problems tab in VS code reports issues that have since been addressed and only restarting VS Code clears them down

### Steps To Reproduce:

1.  Create some Apex code with issues ¯\_(ツ)_/¯
1.  Fix the issues
1.  Save the code

### Expected result

_Problems list should clear down_.

### Actual result

_Problems list persists with those issues, file remains highlighted in sidebar (red in my theme)_.

### Additional information

_See screenshot_.

**VS Code Version**: 1.49.3

**SFDX CLI Version**: 7.75.0-5521f40d11 Win32-x64 node-v12.18.3

**OS and version**: Windows Server 2012 R2 Standard (but also seen on my local machine Win10 Pro)
![image](https://user-images.githubusercontent.com/2143167/95427483-8d1cda00-093f-11eb-833e-632868498799.png)

BrettMN:
I have the same... _Problems_ 


VS Code Version: 1.50.0-insider

SFDX CLI Version: sfdx-cli/7.75.0 darwin-x64 node-v12.18.4

OS and version: MacOS Catalina 10.15.7

RPDevJesco:
During those situations, reloading VS Code should fix the issue. Has something to do with the VSCode cache.

BrettMN:
That workaround would mean restarting VS Code ever couple of minutes.   

I'd end up spending more time waiting for things to save, not refresh the problems and re-opening VSCode than doing work. 

This is only happening in an SFDX Project for me as well.   Other projects, where the SFDX plugins are disabled, do not have this issue.

lcampos:
Hi thanks for opening this issue. Can you confirm if these lines in the `Problems` tab are a result of a failed deploy ? If they are then this is working as designed since the list of errors is kept around until the next time a deploy happens. That is because it's currently not possible to evaluate if the error was actually fixed without re-deploying.

robcowell:
I have the deploy on save feature enables so no, a subsequent deploy 
doesn't clear them down. Not sure if the beta quick deploy feature is a 
factor here, I use that.

KoninosV2:
For me, the procedure to recreate the problem is:

1. Write code with problems (that don't get picked up on save but on deploy e.g. wrong sobject name).
2. Save the file
3. Try to deploy the file (and fail)
   3.1. Now the problems get the red underline and shown in the Problems tab.
[...]

PR #2671, opened 2020-10-26
fix: clear deploy diagnostics

lcampos:
### What does this PR do?
Fixes different issues with clearing diagnostics for deploy operations:
- Clear diagnostics before parsing deploy results
- Clear diagnostics in case there was an error parsing deploy results
- Use the same diagnostics collection between CLI and Library implementation so diagnostics can be cleared correctly regardless of which path created the diagnostics

#2608, @W-8266770@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
@@ -76,6 +76,7 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
       this.logMetric(execution.command.logName, startTime, properties);
 
       try {
+        BaseDeployExecutor.errorCollection.clear();
         if (stdOut) {
           const deployParser = new ForceDeployResultParser(stdOut);
           const errors = deployParser.getErrors();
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -76,6 +76,7 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
       this.logMetric(execution.command.logName, startTime, properties);
 
       try {
+        BaseDeployExecutor.errorCollection.clear();
         if (stdOut) {
```
AnanyaJha, line 80: do we want to clear the diagnostics even if we don't parse the deploy results (when there is no stdOut) ?

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -161,18 +158,17 @@ export class LibraryDeploySourcePathExecutor extends LibraryCommandletExecutor<
       const parser = new LibraryDeployResultParser(result);
       const outputResult = parser.resultParser(result);
       channelService.appendLine(outputResult);
-
+      BaseDeployExecutor.errorCollection.clear();
```
AnanyaJha, line 161: it looks like the diagnostics get cleared at the beginning of `handleDeployRetrieveLibraryDiagnostics()` anyway, so do we need to clear them here as well?

packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
AnanyaJha: jk this makes sense to me

brpowell (approved):

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
lcampos: Yes because `handleDeployRetrieveLibraryDiagnostics()` is only called when the status is not Succeeded or Completed

AnanyaJha (approved):

PR #2675, opened 2020-10-27
JHork: Save Button Icons

jonnyhork:
### What does this PR do?
This PR will set webview URI's for SLDS icons we use for our save buttons in the SOQL Builder.

@W-8253104@

### Functionality Before
no icons

### Functionality After
Save Icons

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
+++ b/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
@@ -20,7 +20,8 @@ export function getHtml(
     baseStyleUri,
     tabulatorStyleUri,
     viewControllerUri,
-    tabulatorUri
+    tabulatorUri,
+    saveIconUri
   } = assets;
 
   const pathToDataViewDist = path.join(extensionPath, DATA_VIEW_UI_PATH);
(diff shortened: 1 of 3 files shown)


ekapner (commented):
@jonnyhork - the save buttons look GOOD!

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
```diff
@@ -35,6 +36,10 @@ export function getHtml(
   html = html.replace('${baseStyleUri}', baseStyleUri.toString());
   html = html.replace('${tabulatorUri}', tabulatorUri.toString());
   html = html.replace('${viewControllerUri}', viewControllerUri.toString());
+  // There are multiple buttons that require this icon
+  while (html.match(/\${iconSave}/)) {
+    html = html.replace('${iconSave}', saveIconUri.toString());
+  }
```
brpowell, line 42: Just for more context - this section is to help "normalize" the html depending on the platform it's running on?

jonnyhork: This section is to satisfy the constraints around [loading local resources](https://code.visualstudio.com/api/extension-guides/webview#loading-local-content) in a webview.

The [examples](https://github.com/microsoft/vscode-extension-samples/blob/master/webview-sample/src/extension.ts#L194) you'll see from vs code use a template string for all of their html, so interpolation is a bit cleaner. However, out markup is more complex and we want to be able to write HTML in an html file instead of a single string.
This file is replacing all the relevant `src` attributes with `webviewUri`s.

dehru (approved):
QA'd and looks good!

brpowell (approved):

AnanyaJha (approved):

PR #2682, opened 2020-10-29
perf(deploybeta): allow multiple component retrieve for performance beta

rcoringrato-sfdc:
### What does this PR do?

Enables multiple component retrieve when using the beta retrieve

@W-7963272@

### Functionality Before
Only individual components could be retrieved when beta retrieve is enabled

### Functionality After
Multiple components can be retrieved with beta retrieve

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
@@ -143,10 +143,20 @@ export class LibraryRetrieveSourcePathExecutor extends LibraryCommandletExecutor
       'namespace'
     )) as string;
     const client = new SourceClient(await getConnection);
-    const retrieve = client.tooling.retrieve({
-      components,
-      namespace: projectNamespace
-    });
+    let retrieve;
+    if (components.length === 1) {
+      retrieve = client.tooling.retrieve({
+        components,
+        namespace: projectNamespace
+      });
+    } else {
+      retrieve = client.metadata.retrieve({
+        components,
+        namespace: projectNamespace,
+        merge: true,
+        output: response.data
+      });
+    }
     const metadataCount = JSON.stringify(createComponentCount(components));
     this.telemetry.addProperty('metadataCount', metadataCount);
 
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -143,10 +143,20 @@ export class LibraryRetrieveSourcePathExecutor extends LibraryCommandletExecutor
       'namespace'
     )) as string;
     const client = new SourceClient(await getConnection);
-    const retrieve = client.tooling.retrieve({
-      components,
-      namespace: projectNamespace
-    });
+    let retrieve;
+    if (components.length === 1) {
```
lcampos, line 147: This works for now because we have the metadata type checks in useBetaDeployRetrieve above. What is the plan for when we stop checking for them since Tooling API does not support every metadata type ?

rcoringrato-sfdc: Ah, now that 1.1.9 is available I'm planning on removing the type filter in useBetaDeployRetrieve. I'll need to move it here instead. Shouldn't the supported types be exported for the tooling api?

brpowell: That would be nice if it self reported the supported types.

lcampos: @rcoringrato-sfdc we could move those checks here or have them be exported from the tooling api code in the library. However, we've talked about eventually just having one endpoint per operation in the library which internally would be responsible for this routing depending on the metadata type and other operation details.

The next changes we do should be in line with that vision of having the library deal with this.

@brpowell I don't think that is part of the current planned work but we should keep it in mind.

brpowell: Yeah eventually having an "smart route" might be the goto. The other thing you could do is surround this in a try catch to try retrieving with tooling if we have one component. If it's an unsupported type, the library already throws an error so you could catch that and then route it to metadata api instead.

lcampos (approved):
:shipit:

AnanyaJha (approved):

PR #2683, opened 2020-10-29
chore: add additional slack messages for post publish steps

sfsholden:
### What does this PR do?
In order to clarify when there is a post merge issue vs. a publishing issue, this change adds additional messaging to the slack output.

@W-8290541@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -443,6 +443,14 @@ jobs:
       - sfchangecase/update:
           location: REPO_LOCATION
           release: GUS_RELEASE
+      - slack/notify:
+          channel: 'pdt_releases'
+          color: '#9bcd9b'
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -443,6 +443,14 @@ jobs:
       - sfchangecase/update:
           location: REPO_LOCATION
           release: GUS_RELEASE
+      - slack/notify:
```
sfsholden, line 446: I might go ahead and remove the post merge steps altogether (depending on the results of the automation in the next few weeks) - instead having a reminder posted to slack to perform the post merge. This might just not be worthwhile spending much more time on to have circle ci automate for us.

lcampos: Sounds good, the post merge is a difficult thing to automate but shouldn't be an issue to keep it as a manual step

lcampos (approved):

PR #2685, opened 2020-10-29
Jhork/check default username

jonnyhork:
### What does this PR do?
The SOQL Builder requires an auth info and a default org set in order to use the connection to get metadata. Because the SOQL extension is activated and usable outside an sfdx project we wan to notify the user that there is no default or set when they open the custom editor.
@W-8253055@

### Functionality Before
The UI would hang in a loading state and the user would not know why.

### Functionality After
SOQL - No Org Set


packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -52,7 +60,16 @@ export class SOQLEditorProvider implements vscode.CustomTextEditorProvider {
     this.instances.push(instance);
     instance.onDispose(this.disposeInstance.bind(this));
     this.context.subscriptions.push(...instance.subscriptions);
+
+    // Check to see if a default org is set.
+    if (!workspaceContext.username) {
```
jonnyhork, line 65: `workspaceContext.username` will refer to the _default_ username.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -17,6 +17,14 @@ import {
 import { HtmlUtils } from './htmlUtils';
 import { SOQLEditorInstance } from './soqlEditorInstance';
 
+const sfdxCoreExtension = vscode.extensions.getExtension(
```
lcampos, line 20: Please don't use this pattern, we are working on removing this type of usage of the core extension since it's causing performance problems.

lcampos (changes requested):

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
xyc: @jonnyhork  for channelService, you can take a look at https://github.com/forcedotcom/salesforcedx-vscode/blob/f4515a439f737546e740ce8db749ffb6bfca5eec/packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcStart.ts#L14 and use the utils package

jonnyhork: @lcampos  Is there another way to get the `workspaceContext` or check for the default user name you would recommend?

If not, then the plan was for us to have to refactor our extension anyway once the work to expose this stuff in a utility is done on your end. So the core exports would be addressed then (we already use this pattern to get the connection).

We are hoping to have a releasable `vsix` by Nov 12th. Will the utilities be available before then?

jonnyhork: @xyc Looks like channelService is no longer available in the `out` dir when I add it as a dependency, did something change?
`'@salesforce/salesforcedx-utils-vscode/out/src/commands';`, commands does not exist.
I am honestly not sure, can I just use
`'@salesforce/salesforcedx-utils-vscode/src/commands';`?

xyc: @jonnyhork actually let's wait a little bit for that, I think @lcampos still has some changes to be made on channelService

jonnyhork: Sounds good to me, I already gave up on it =)

jonnyhork: We will still need to know how to move forward from here. I realize we are trying to avoid using core as a dependency but my preference would be to move forward with whats accessable in core exports now so we can get the extension out the door. Then refactor this once the utility work is done.

We can also just stick this in waiting until the utils are available as long as its before Nov 12th.

lcampos: @jonnyhork you should be able to use channel service, notification service and telemetry from the utils module. Your branch is outdated so you should rebase first and so you can use those changes.

Specific to `workspaceContext`, that's not something we were thinking on immediately moving but should be simple to do now that we've done the heavy lifting needed in the utils module. We welcome you guys opening a PR for moving that to the utils module.

jonnyhork: Just getting caught up on this, looks like there is a work item on our board to submit a PR to pull out the worksapceContext to the utils, I think Jon Gellin will be doing this work. I may just put this work in waiting until that work is done. Still discussing with the team.

dehru:
This passed my QA.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -194,6 +194,15 @@ export class SOQLEditorInstance {
   }
 
   protected handleRunQuery(): Promise<void> {
+    // Check to see if a default org is set.
```
jonnyhork, line 197: We decided it would be good to check for the default username before running a query as well. Without this, if there was no default org set and the user clicks "Run Query" then nothing would happen and there would be no feedback to the user. The messaging was hidden in the output channel.
Especially since we always allow our users to at least run a query regardless if it has unsupported syntax, we wanted to provide feedback as to why nothing was happening.

lcampos (approved):
Approving since we've agreed to address refactoring the dependency with `salesforcedx-vscode-core` next week

PR #2686, opened 2020-10-29
refactor: refactor commandlet

xyc:
### What does this PR do?
Move commandlet to utils package

@W-8183100@


packages/salesforcedx-utils-vscode/src/index.ts
```diff
@@ -1,9 +1,29 @@
...
+export { EmptyPostChecker } from './commands/postconditionCheckers';
+export {
+  EmptyPreChecker,
+  SfdxWorkspaceChecker
+} from './commands/preconditionCheckers';
+export {
+  CommandletExecutor,
+  SfdxCommandlet,
+  SfdxCommandletExecutor
+} from './commands/sfdxCommandlet';
```
xyc, line 21: Export commandlet and telemetryService from index.ts since commandlet depends on telemetryService.

packages/salesforcedx-utils-vscode/src/index.ts
```diff
@@ -1,9 +1,29 @@
...
+
 export {
-  SFDX_PROJECT_FILE,
-  ENV_SFDX_DEFAULTUSERNAME,
-  ENV_SFDX_INSTANCE_URL,
-  SFDX_CONFIG_ISV_DEBUGGER_SID,
-  SFDX_CONFIG_ISV_DEBUGGER_URL,
-  DEFAULT_CONNECTION_TIMEOUT_MS,
-  CLIENT_ID
-} from './constants';
```
xyc, line 9: Moving these exports to `src/types`. Tests of `salesforcedx-sobjects-faux-generator` depends on it, and it doesn't have vscode dependency. Move them so we don't have to create similar mocks in `salesforcedx-sobjects-faux-generator`.

packages/salesforcedx-utils-vscode/src/commands/preconditionCheckers.ts
```diff
@@ -0,0 +1,57 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import { workspace } from 'vscode';
+import { nls } from '../messages';
+import { Predicate, PredicateResponse } from '../predicates';
+import { PreconditionChecker, SFDX_PROJECT_FILE } from '../types';
+import {
```
lcampos, line 14: We should export these functions so we can import with just having `../workspaces`

packages/salesforcedx-utils-vscode/src/index.ts
lcampos: Does this mean we should remove `index.ts` from the telemetry folder and probably have webpack not process it directly ?

xyc: I've removed the contents of `telemetry/index.ts`: https://github.com/forcedotcom/salesforcedx-vscode/blob/8605f1f540c4a6bb24dbdb4a96e4a24a81159c04/packages/salesforcedx-utils-vscode/src/telemetry/index.ts 

Unit tests would need some adjustments too.

packages/salesforcedx-utils-vscode/src/commands/preconditionCheckers.ts
xyc: Updated, but since workspace is in a different folder, we'd still have multiple copies of workspaces code

packages/salesforcedx-utils-vscode/src/commands/sfdxCommandlet.ts
```diff
@@ -0,0 +1,159 @@
...
+  protected attachExecution(
+    execution: CommandExecution,
+    cancellationTokenSource: vscode.CancellationTokenSource,
+    cancellationToken: vscode.CancellationToken
+  ) {
+    /* channelService.streamCommandOutput(execution);
+
+    if (this.showChannelOutput) {
+      channelService.showChannelOutput();
+    } */
```
AnanyaJha, line 54: are we planning on moving the channelService and adding in this piece in a future PR?

packages/salesforcedx-utils-vscode/src/commands/sfdxCommandlet.ts
```diff
@@ -0,0 +1,159 @@
...
+  >();
+  public readonly onDidFinishExecution: vscode.Event<[number, number]> = this
+    .onDidFinishExecutionEventEmitter.event;
+
+  protected attachExecution(
+    execution: CommandExecution,
+    cancellationTokenSource: vscode.CancellationTokenSource,
+    cancellationToken: vscode.CancellationToken
+  ) {
+    /* channelService.streamCommandOutput(execution);
```
xyc, line 50: This piece is handled by the consumer of `SfdxCommandlet`. It requires a named instance of `ChannelService`.

lcampos: Let's remove the commented out code

packages/salesforcedx-vscode-lightning/src/index.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { shared as lspCommon } from '@salesforce/lightning-lsp-common';
-import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode/out/src/telemetry';
+import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode/out/src';
```
lcampos, line 9: I think we'll need to update the soql extension too https://github.com/forcedotcom/salesforcedx-vscode/blob/8c1348493cb62410ff5a39d132b5f00eea564a7b/packages/salesforcedx-vscode-soql/src/telemetry/index.ts#L7

packages/salesforcedx-utils-vscode/src/commands/sfdxCommandlet.ts
```diff
@@ -0,0 +1,159 @@
...
+    ProgressNotification.show(execution, cancellationTokenSource);
+  }
+
+  public logMetric(
+    logName: string | undefined,
+    hrstart: [number, number],
+    properties?: Properties,
+    measurements?: Measurements
+  ) {
+    TelemetryService.getInstance()
```
lcampos, line 69: Not sure but I think we'll have to eventually change this or move it to the extension code. If every extension has it's own telemetry instance I think the current way we handle getting the telemetry instance might have to change to be more like what we have for channel service.

xyc: That could be a story for another day, but I won't be hurrying to address it in this PR, since it requires more thoughtfulness in the design of it. It could be that we pass ChannelService and TelemetryService instances into SfdxCommandlet.

As of current state, as long as TelemetryService is initialized correctly in the extension, telemetry will be logged.

sfsholden (approved):

packages/salesforcedx-utils-vscode/src/telemetry/index.ts
```diff
@@ -1,8 +1,17 @@
 /*
- * Copyright (c) 2018, salesforce.com, inc.
+ * Copyright (c) 2020, salesforce.com, inc.
  * All rights reserved.
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-export { TelemetryService, TelemetryBuilder, TelemetryData, Properties, Measurements } from './telemetry';
+/**
```
xyc, line 8: Re-removing the telemetry service index. (the exports were there because of a merge conflict.) See W-8680810 for the follow up work item.

PR #2687, opened 2020-10-29
Channel service refactor

lcampos:
### What does this PR do?
This PR update channel service so we can create unique channels by extension.

@W-8183064@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/index.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/index.ts
@@ -5,9 +5,6 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { ChannelService } from './channelService';
-export const channelService = ChannelService.getInstance();
-export { DEFAULT_SFDX_CHANNEL } from './channelService';
 export { ChannelService } from './channelService';
 
 import { NotificationService } from './notificationService';
--- a/packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
@@ -7,10 +7,8 @@
 
 import { Observable } from 'rxjs/Observable';
 import * as vscode from 'vscode';
-
 import { CommandExecution } from '../cli';
 import { nls } from '../messages';
-import { channelService } from './index';
 
 export const STATUS_BAR_MSG_TIMEOUT_MS = 5000;
 
@@ -97,14 +95,12 @@ export class NotificationService {
     this.showErrorMessage(
       nls.localize('notification_unsuccessful_execution_text', executionName)
     );
-    this.showChannelOutput();
   }
 
   private showCanceledExecution(executionName: string) {
     this.showWarningMessage(
       nls.localize('notification_canceled_execution_text', executionName)
     );
-    this.showChannelOutput();
   }
 
   public async showSuccessfulExecution(executionName: string) {
@@ -125,9 +121,7 @@ export class NotificationService {
         showButtonText,
         showOnlyStatusBarButtonText
       );
-      if (selection && selection === showButtonText) {
-        this.showChannelOutput();
-      }
+
       if (selection && selection === showOnlyStatusBarButtonText) {
         await vscode.workspace
           .getConfiguration('salesforcedx-vscode-core')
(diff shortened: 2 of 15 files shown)


packages/salesforcedx-utils-vscode/src/commands/index.ts
```diff
@@ -5,9 +5,6 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { ChannelService } from './channelService';
-export const channelService = ChannelService.getInstance();
```
lcampos, line 9: Exposing the channel service like this prevents us from initializing the service with a specific name. Instead, this gets initialized at startup as a channel with the default name.

packages/salesforcedx-vscode-lwc/src/commands/commandUtils.ts
```diff
@@ -6,12 +6,13 @@
...
 
 import {
-  channelService,
+  ChannelService,
   notificationService
 } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
 
+const channelService = ChannelService.getInstance('lwc');
```
lcampos, line 15: I didn't find a better way of initializing the service per extension so we can have a unique channel name per extension. I could move this into its own module in the extension that does this initialization but other than that I think we'll have to do something like this per extension. Let me know if you have any suggestions

packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
```diff
@@ -7,10 +7,8 @@
 
 import { Observable } from 'rxjs/Observable';
 import * as vscode from 'vscode';
-
 import { CommandExecution } from '../cli';
 import { nls } from '../messages';
-import { channelService } from './index';
```
lcampos, line 13: Similar to the comment on `commands/index.ts` this gets loaded at startup, which forces us to initialize the channel service with the default name. 
I also opted to remove channel service from here since we could still call showing the channel output directly from the command implementation in the extension side.

packages/salesforcedx-vscode-lwc/src/commands/commandUtils.ts
xyc: I was thinking putting `window.createOutputChannel` in an `ChannelService.prototype.initialize` function, and put initialize in the extension activation function. That way we don't have to call `ChannelService.getInstance('lwc')` every time. (Also `lwc` should be an i18n string)

lcampos: Yeah that's cleaner, let me look at making those changes

packages/salesforcedx-vscode-lwc/src/channel/index.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ChannelService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
```
lcampos, line 8: @xyc I ended up doing it this way instead of having it as part of the extension activation since this allows us to initialize the channel service when the extension code is loaded which happens before activation.

packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
```diff
@@ -125,9 +121,7 @@ export class NotificationService {
         showButtonText,
         showOnlyStatusBarButtonText
       );
-      if (selection && selection === showButtonText) {
```
xyc, line 128: Do we intend to remove this? show button won't have an effect then

packages/salesforcedx-vscode-lwc/src/channel/index.ts
xyc: Sounds good this works too.

xyc (approved):

packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
lcampos: Yes, I'm working on another PR which will modify some of the notification service code and I'll address the show button there. Just wanted to remove the channel service dependency for now.

PR #2695, opened 2020-11-02
perf: prune and ignore files when packaging vsix

dehru:
### What does this PR do?
This PR attempts to address the package size of the salesforce-vscode-soql vsix file.  Before this, the vsix file is great than 10mb.  This PR adds 2 stpes to reduce the vsix file size, each reduces the vsix file as seen below: 

- original vsix: 7084 files, 10.07MB
- after node-prune:  5482 files, 8.34MB
- after .vscodeignore: 3789 files, 7.06MB

Additionally, the UI bits of this ( Webview / Custom Editor ) have these sizes:
- soql-builder-ui: 532k
- soql-data-view: 393k

// None yet, hoping to talk with luis about further steps and then create work item.

### Functionality Before
original vsix size: 7084 files, 10.07MB

### Functionality After
new vsix size: 3789 files, 7.06MB


dehru:
@lcampos - I pushed this PR up as reference during our meeting on Wednesday afternoon.  I also want to discuss with you about webpack this extension.  https://code.visualstudio.com/api/working-with-extensions/bundling-extension

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -67,6 +67,7 @@
     "test": "npm run test:vscode-integration",
     "test:vscode-integration": "node ../../scripts/run-vscode-integration-tests-with-top-level-extensions",
     "test:vscode-insiders-integration": "cross-env CODE_VERSION=insiders npm run test:vscode-integration",
+    "prevscode:package": "npx node-prune",
```
dehru, line 70: I found that `npm prune --production` made no change in the size of the vsix.  So i didn't add it to the pre-package script.  

Below is my output of vsix sizes with 2 separate vsix's.  One created after npm prune and one before npm prune.  Note the sizes are the same.

```
-rw-r--r--  1 tcromer  wheel  7407284 Nov  2 11:59 salesforcedx-vscode-soql-50.3.0.after-prune.vsix
-rw-r--r--  1 tcromer  wheel  7407284 Nov  2 11:58 salesforcedx-vscode-soql-50.3.0.before-prune.vsix
```

packages/salesforcedx-vscode-soql/.vscodeignore
```diff
@@ -5,4 +5,16 @@ test/**
...
-tsconfig.json
\ No newline at end of file
+tsconfig.json
+.eslint.json
+**/*.ts
+**/tsconfig.json
+*.xml
+**/*.md
+*.vsix
+node_modules/rxjs/src/**
```
lcampos, line 15: Why are you explicitly ignoring these modules ?

packages/salesforcedx-vscode-soql/package.json
lcampos: Posting so others can see this. We synced offline and the reason why this happens is because `npm prune --production` only removes devdependencies. I think you should still leave the prune step as part of building the vsix. Take a look at how it's done for the other extensions.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -21,9 +21,9 @@
   ],
   "dependencies": {
     "@salesforce/core": "2.12.1",
-    "@salesforce/soql-builder-ui": "0.0.21",
-    "@salesforce/soql-data-view": "0.0.5",
-    "@salesforce/soql-language-server": "0.2.5",
+    "@salesforce/soql-builder-ui": "0.0.22",
+    "@salesforce/soql-data-view": "0.0.6",
+    "@salesforce/soql-language-server": "0.2.7",
     "@types/papaparse": "^5.2.3",
```
lcampos, line 27: This `@types/papaparse` should probably be moved to `devDependencies`

packages/salesforcedx-vscode-soql/.vscodeignore
dehru: rxjs ships a bunch of pre-baked bundles of it's library that we don't use. Like esm bundles and it's src directories.  I found this article about it and stole that little tidbit.  Looks like it removes 463 files = ~500k

https://tsh.io/blog/reduce-node-modules-for-better-performance/

jonnyhork (approved):

packages/salesforcedx-vscode-soql/.vscodeignore
lcampos: Awesome, we'll use this for the other extensions :)

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -52,6 +52,7 @@
     "eslint-plugin-jsdoc": "^30.0.3",
     "eslint-plugin-prettier": "^3.1.4",
     "mocha": "^5",
+    "node-prune": "^1.0.2",
```
lcampos, line 55: If you are only using this for the `prevscode:package` script I think you can remove it since you are running it using `npx`

lcampos (approved):

packages/salesforcedx-vscode-soql/package.json
dehru: Yes, thanks.

dehru: I wasn't sure the protocol here.  If i use it in a script, it's kinda sneaking in dependencies.

PR #2719, opened 2020-11-09
chore: update to vscode engine 1.46.0

lcampos:
### What does this PR do?
Update the project to use version 1.46.0 and above for VSCode.

@W-8374828@

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -6,7 +6,7 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.40.0"
+    "vscode": "^1.46.0"
   },
   "categories": [
     "Debuggers"
--- a/packages/salesforcedx-utils-vscode/test/unit/telemetry/MockContext.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/telemetry/MockContext.ts
@@ -5,7 +5,13 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import * as path from 'path';
-import { ExtensionContext, Memento } from 'vscode';
+import {
+  EnvironmentVariableCollection,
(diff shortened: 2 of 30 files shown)


packages/salesforcedx-apex-debugger/package.json
```diff
@@ -6,7 +6,7 @@
   "publisher": "salesforce",
   "license": "BSD-3-Clause",
   "engines": {
-    "vscode": "^1.40.0"
+    "vscode": "^1.46.0"
```
lcampos, line 9: Most users are on the latest 2-3 versions of VSCode but we do have some still running on versions prior to this. I picked this version since it will be a good way to move users to newer versions of VSCode and it's also not a big change for the project going from 1.40 to 1.51.

packages/salesforcedx-utils-vscode/test/unit/telemetry/MockContext.ts
```diff
@@ -26,10 +32,45 @@ class MockMemento implements Memento {
   }
 }
 
+class MockEnvironmentVariableCollection
```
lcampos, line 35: Nothing inside of this Mock class is directly used so we can get away with not implementing all the logic.

jgellin-sf (approved):
Looks good

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -34,7 +34,7 @@
     "@types/mocha": "^5",
     "@types/node": "12.0.12",
     "@types/sinon": "^2.3.2",
-    "@types/vscode": "1.40.0",
+    "@types/vscode": "1.46.0",
```
brpowell, line 37: should we have the caret for @types as well since it's there for the engine version?

brpowell (approved):

packages/salesforcedx-vscode-apex-debugger/package.json
lcampos: I didn't add it for the types since those can change and I want us to use the lowest supported version types, in this case version 1.46.0.

PR #2722, opened 2020-11-11
chore: update post publish job for merging (@W-8290541)

sfsholden:
### What does this PR do?
1. Changes color for start of the merge of main back into develop. This is to help indicate that it is in progress.
1. Added suggestion from CircleCi support to resolve the error:

> error: Terminal is dumb, but EDITOR unset
> Not committing merge; use 'git commit' to complete the merge.
> 

@W-8290541@

Example Failure:
https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/2842/workflows/d69adcc2-687a-418f-8bf6-1c26014d0b73/jobs/14623

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,7 @@ commands:
             git config credential.helper 'cache --timeout=<< parameters.cache_timeout >>'
             git config user.email "<< parameters.gh_email >>"
             git config user.name "<< parameters.gh_username >>"
+            git config --global pull.rebase true
 
   install:
     description: 'Installing dependencies'
@@ -455,7 +456,7 @@ jobs:
           name: 'Merge changes from main back into develop'
           command: |
             git checkout develop
-            git pull --rebase
+            git pull
             git merge main --commit --no-edit
             git push origin develop
       - slack/notify:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -449,8 +449,8 @@ jobs:
           message: 'VSCode Extensions v${RELEASE_VERSION} have been published to the marketplace'
       - slack/notify:
           channel: 'pdt_releases'
-          color: '#9bcd9b'
-          message: 'VSCode Extensions v${RELEASE_VERSION} Post Publish - Merging main into develop'
+          color: '#FFDEAD'
+          message: 'VSCode Extensions v${RELEASE_VERSION} Post Publish - Starting merge of main into develop'
       - run:
           name: 'Merge changes from main back into develop'
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -455,7 +455,7 @@ jobs:
           name: 'Merge changes from main back into develop'
           command: |
             git checkout develop
-            git pull
+            git pull --rebase
             git merge main --commit --no-edit
             git push origin develop
       - slack/notify:


.circleci/config.yml
```diff
@@ -449,13 +449,13 @@ jobs:
...
-          color: '#9bcd9b'
-          message: 'VSCode Extensions v${RELEASE_VERSION} Post Publish - Merging main into develop'
+          color: '#FFDEAD'
+          message: 'VSCode Extensions v${RELEASE_VERSION} Post Publish - Starting merge of main into develop'
       - run:
           name: 'Merge changes from main back into develop'
           command: |
             git checkout develop
-            git pull
+            git pull --rebase
```
AnanyaJha, line 458: why do we want to rebase these changes?

sfsholden: It's to hopefully get around this error:
> error: Terminal is dumb, but EDITOR unset
> Not committing merge; use 'git commit' to complete the merge.

I was under the impression the issue was happening from the `git merge` command. But after reaching out to circleci support, they pointed out the issue looked like it was actually the pull command. 

Here's their explanation:

> Looks like git pull is trying to open an editor to ask for a merge commit's message and failed.
> You can try git pull --rebase to avoid merge or git pull --no-commit && git commit -m "Merge" to merge with commit message.
> 
> By default, when you do a merge, Git tries to invoke an editor; however, in a non-interactive setting, Git checks EDITOR environment variable, but it's unset. That's why you are getting that message.

AnanyaJha: Ah okay that seems annoying. Is there any potential issue to develop's history if we switch to the `git pull --rebase` strategy instead of a pull/merge ?

jag-sfdc: Yes, I was also typing the same question @AnanyaJha has. Any potential downsides to this? Can we instead use git pull/merge and pass in a templated merge commit message to maintain the existing behavior?

sfsholden: Good questions!

We are only doing a `git pull` here because without it we hit the below error. So at this point in time, we have to update develop before we merge main into it. From talking with circleci support, they think this is likely due to a few differences in hashes for some commits.

Error from the past:
> To github.com:forcedotcom/salesforcedx-vscode.git
>  ! [rejected]          develop -> develop (non-fast-forward)
> error: failed to push some refs to 'git@github.com:forcedotcom/salesforcedx-vscode.git'
> hint: Updates were rejected because the tip of your current branch is behind
> hint: its remote counterpart. Integrate the remote changes (e.g.
> hint: 'git pull ...') before pushing again.
Example Failure Run in Circle Ci: 
https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/2496/workflows/be1f9e4b-b96f-4a8a-970a-bd35a1da275a/jobs/13576

Using `git pull --rebase`:
This shouldn't change how the merge of main into develop itself occurs. Here's some documentation from the manual for `git pull` that I think will help alleviate any concerns.
> Incorporates changes from a remote repository into the current branch. In its default mode, git pull is shorthand for git fetch followed by git merge FETCH_HEAD.
> More precisely, git pull runs git fetch with the given parameters and calls git merge to merge the retrieved branch heads into the current branch. With --rebase, it runs git rebase instead of git merge.

For our use case, updating `develop` to match `develop` upstream shouldn't have any impact whether we use the default merge functionality or rebase. This is purely to get around the terminal error.

.circleci/config.yml
```diff
@@ -449,13 +449,13 @@ jobs:
           message: 'VSCode Extensions v${RELEASE_VERSION} have been published to the marketplace'
       - slack/notify:
           channel: 'pdt_releases'
-          color: '#9bcd9b'
-          message: 'VSCode Extensions v${RELEASE_VERSION} Post Publish - Merging main into develop'
+          color: '#FFDEAD'
```
brpowell, line 452: You of all people Shelby, I can't believe you'd declare final fantasy dead.

sfsholden: 🤣  oh that is perfect

.circleci/config.yml
lcampos: The change looks good to me, we should always be using rebase when doing a pull so any local changes end up being on top of the ones already committed to the remote. 

However, if CircleCI thinks this is the solution to the issue I would prefer to have the global git config set to always rebase when we run git pull e.g. `git config --global pull.rebase true`

lcampos (approved):

PR #2732, opened 2020-11-13
fix(soql): parse uri for saved results as file

mysticflute:
### What does this PR do?
Fixes an issue where saved soql query result files are not revealed in the file explorer in Windows.

@W-8383294@

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
+++ b/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
@@ -85,7 +85,7 @@ export class QueryDataFileService {
   private showFileInExplorer(targetPath: string) {
     vscode.commands.executeCommand(
       'revealInExplorer',
-      vscode.Uri.parse(targetPath)
+      vscode.Uri.file(targetPath)
     );
   }
 


praksb (approved):

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -85,7 +85,7 @@ export class QueryDataFileService {
   private showFileInExplorer(targetPath: string) {
     vscode.commands.executeCommand(
       'revealInExplorer',
-      vscode.Uri.parse(targetPath)
+      vscode.Uri.file(targetPath)
```
jonnyhork, line 88: DOH! That was silly of me, we use file everywhere else.

I wonder why this works on macOS regardless?

jonnyhork (approved):
Thanks for taking care of this Nathan!

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
lcampos: I think it works on MacOS because of the format of the string being passed to the parse method. It could be that on Windows the `targetPath` does not include the scheme or has some characters that are not correctly handled. You can add breakpoints in the code and look at the differences in `targetPath` across OS. Here's more info about the differences in methods https://code.visualstudio.com/api/references/vscode-api#Uri.

lcampos (approved):

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
jonnyhork: Thanks for the link, they example they have in the docs helps to see.

PR #2734, opened 2020-11-13
Remote SOQL Errors - Language Client Part

code0100fun:
[W-7977000](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07B0000008ULb1IAG)

Dependent on https://github.com/forcedotcom/soql-tooling/pull/93

### What does this PR do?

Handles `runQuery` requests from the SOQL language server. This will run the query unmodified and return the result to the LS. The LS may then validate and add diagnostic messages.

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/client/client.ts
+++ b/packages/salesforcedx-vscode-soql/src/client/client.ts
@@ -27,7 +27,7 @@ const workspaceContext = sfdxCoreExtension?.exports?.workspaceContext;
 let client: LanguageClient;
 
 export function clearDiagnostics(): void {
-  client.diagnostics?.clear();
+  client?.diagnostics?.clear();
 }
 
 export async function startLanguageClient(
--- a/packages/salesforcedx-vscode-soql/src/client/client.ts
+++ b/packages/salesforcedx-vscode-soql/src/client/client.ts
@@ -5,19 +5,34 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import {
+  QueryValidationFeature,
+  RequestTypes
+} from '@salesforce/soql-language-server';
 import * as path from 'path';
-import { ExtensionContext, workspace } from 'vscode';
-
+import { ExtensionContext, extensions, workspace } from 'vscode';
 import {
   LanguageClient,
   LanguageClientOptions,
   ServerOptions,
   TransportKind
 } from 'vscode-languageclient';
+import { QueryRunner } from '../editor/queryRunner';
+
+const sfdxCoreExtension = extensions.getExtension(
+  'salesforce.salesforcedx-vscode-core'
+);
+const workspaceContext = sfdxCoreExtension?.exports?.workspaceContext;
 
 let client: LanguageClient;
 
-export function startLanguageClient(context: ExtensionContext): void {
+export function clearDiagnostics(): void {
+  client.diagnostics?.clear();
+}
+
+export async function startLanguageClient(
+  context: ExtensionContext
+): Promise<void> {
   // path to language server module
   const serverModule = context.asAbsolutePath(
     path.join(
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-soql/src/client/client.ts
```diff
@@ -55,9 +70,29 @@ export function startLanguageClient(context: ExtensionContext): void {
     serverOptions,
     clientOptions
   );
+  client.registerFeature(new QueryValidationFeature());
```
code0100fun, line 73: This feature sets `capabilities.soql.runQuery` so the server knows it's ok to send the `RunQuery` request for validation

packages/salesforcedx-vscode-soql/src/client/client.ts
```diff
@@ -55,9 +70,29 @@ export function startLanguageClient(context: ExtensionContext): void {
...
+      } catch (e) {
+        const error = {
+          name: e.name,
+          errorCode: e.errorCode,
+          message: e.message
+        };
+        return { error };
+      }
+    });
+  });
```
brpowell, line 95: Might be able to make this a little more readable:

```typescript
await client.onReady()
client.onRequest(RequestTypes.RunQuery, async (queryText: string) => {
  try {
    const conn = await workspaceContext.getConnection();
    const queryData = await new QueryRunner(conn).runQuery(queryText, {
      showErrors: false
    });
    return { result: queryData };
  } catch (e) {
    // simplify, unless you're specifically trying to omit other properties
    return { error: e };
  }
});
```

Not sure if you want to await client.onRequest or not, it seems more like a subscribe call but could be wrong.

packages/salesforcedx-vscode-soql/test/vscode-integration/client/client.test.ts
```diff
@@ -0,0 +1,114 @@
...
+function waitUntil(predicate: () => boolean) {
+  return new Promise(async resolve => {
+    let notFound = true;
+    let tries = 5;
+    while (notFound || tries-- > 0) {
+      await sleep(50);
+      notFound = !predicate();
+    }
+    resolve();
+  });
```
brpowell, line 37: Could this possibly cause flapping tests? What's the consequence of it not succeeding?

packages/salesforcedx-vscode-soql/src/client/client.ts
dobladez: Thanks. I'll change it to use `await`.
The caught exception cannot be returned as-is because the return value has to be a plain JS object (to be serialized as a JSON-RPC response). I'll add a comment to make it clear.

packages/salesforcedx-vscode-soql/test/vscode-integration/client/client.test.ts
dobladez: Hopefully not. LSP error diagnostics message arrive asynchronously, and it takes a few milliseconds on the test.
See usage below (line 99):
```
    await waitUntil(() => {
      return languages.getDiagnostics(soqlFileUri).length > 0;
    });
```

brpowell: Gotcha. I don't think this is blocking btw but if this becomes a problem or you'd like to address it earlier maybe you could mock the lsp response and make it less of an integration test.

dobladez: 👍  understood. Thanks.

This case is weird:

1. The client changes the file contents
1. The server gets notified, and then it's the server that sends a request to the client to run the query
1. The client runs it and responds (`client.onRequest( ... ) `)
1. If the query returned errors, we expect the server to push diagnostics back to the client

That's the flow under test here. IMHO, the main point of this test is the integration... if the test mocks it, then it'd not be testing much at all.

Unfortunately, VScode integration tests are apparently not fully real either: I tried the "serialization" issue I mentioned above and it is not caught during automated tests (may be it uses shared memory instead of serializing the objects?).

brpowell (approved):

PR #2742, opened 2020-11-16
fix(soql): include the readme in the vsix

jonnyhork:
### What does this PR do?
This PR will include the readme.md file in the packaged vsix so that it can be rendered on the marketplace.
@W-8214787@

### DIFF
--- a/packages/salesforcedx-vscode-soql/.vscodeignore
+++ b/packages/salesforcedx-vscode-soql/.vscodeignore
@@ -10,7 +10,7 @@ tsconfig.json
 **/*.ts
 **/tsconfig.json
 *.xml
-**/*.md
+**/*[!README].md
 *.vsix
 node_modules/rxjs/src/**
 node_modules/rxjs/bundles/**


dehru (approved):
This gets the job done for now.  I know we need to get the vsix cut today and this is a drop in the bucket of the file size.  I'm gonna play with the glob pattern so it just rips all .md files outa node_modules and leaves any we have alone.

tmock12 (approved):

jonnyhork:
> I'm gonna play with the glob pattern so it just rips all .md files outa node_modules and leaves any we have alone.

I was looking through the node_modules and could not find any .md, obviously I did not look at every one, but I'm curious @dehru did you see a bunch of .md files somewhere in the node_modules?

jgellin-sf (approved):

lcampos (approved):

dehru:
> > I'm gonna play with the glob pattern so it just rips all .md files outa node_modules and leaves any we have alone.
> 
> I was looking through the node_modules and could not find any .md, obviously I did not look at every one, but I'm curious @dehru did you see a bunch of .md files somewhere in the node_modules?

No.  It's just a standard glob pattern of common files that people publish but aren't necessary.  Luckly, it looks like all our dependent libraries are culling these when they publish.  But doesn't hurt to leave it in.

PR #2743, opened 2020-11-17
Remove core extension as dependency on SOQL extension part 1

jgellin-sf:
### What does this PR do?
This is the first of 2 PR's that will remove the core extension as a dependency on the SOQL extension. This PR covers telemetry and channel services. 

@W-8295301@

### Functionality Before
N/A

### Functionality After
N/A


packages/salesforcedx-vscode-soql/src/channel/index.ts
```diff
@@ -0,0 +1,13 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ChannelService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
+
+export const channelService = ChannelService.getInstance(
+  // i18n
+  'SOQL Extension'
```
lcampos, line 12: Update this to use a label that can be localized

lcampos (approved):
This can be merged after making the label change

packages/salesforcedx-vscode-soql/src/channel/index.ts
jgellin-sf: @lcampos See the `feat(soql): nls-enable channel name` commit.

lcampos: Looks good 👍

packages/salesforcedx-vscode-soql/src/messages/i18n.ja.ts
```diff
@@ -0,0 +1,20 @@
+/*
```
dehru, line 1: Japanese?

jgellin-sf: I just copied what LWC had.

lcampos: Yes, we support Japanese in the other extensions. We work with a team that helps us keeping them up to date. Leave it as is and next time we engage with them they'll take care of updating this file too.

packages/salesforcedx-utils-vscode/src/commands/channelService.ts
```diff
@@ -12,18 +12,18 @@ import { nls } from '../messages';
 
 export class ChannelService {
   private readonly channel: OutputChannel;
-  private static instance: ChannelService;
+  private static instances: { [key: string]: ChannelService } = {};
```
dehru, line 15: @lcampos - You should probably see this.  Was this the intent?  Currently, when we try to output to the SOQL Extension channel it prints to the LWC Extension Channel because this is a singleton with the channel stored on it...and that extension sets the channel name first.

dehru: This change fixes that and i verified we have output in SOQL Extension channel.

lcampos: Thanks for catching that. Please add a test to verify you get the correct channel when there's existing ones https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-utils-vscode/test/unit/commands/channelService.test.ts

lcampos (changes requested):
Just needs a test for the changes to `salesforcedx-utils-vscode`

lcampos (approved):
:shipit:

PR #2745, opened 2020-11-17
Update sidebar.yml

ekapner:
Author soql builder topic and update sidebar

@W-8205618@


docs/_data/sidebar.yml
```diff
@@ -145,6 +145,9 @@
     ja: SOQL
   baseUrl: /soql/
   children:
+     - title:
+        en: SOQL Builder
```
lcampos, line 149: We need to add the japanese entry too or the site breaks

lcampos (changes requested):

docs/_data/sidebar.yml
ekapner: @lcampos - added it back. :)

lcampos (approved):

PR #2762, opened 2020-11-25
feat: deploy with manifest in deploy/retrieve performance beta

brpowell:
### What does this PR do?

Enables deploying with a manifest in deploy/retrieve performance beta.

@W-8246812@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
@@ -8,17 +8,25 @@ import {
   Command,
   SfdxCommandBuilder
 } from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import { DeployStatus, WorkingSet } from '@salesforce/source-deploy-retrieve';
+import { join } from 'path';
 import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -36,6 +44,40 @@ export class ForceSourceDeployManifestExecutor extends BaseDeployExecutor {
...
 }
 
+export class LibraryDeployManifestExecutor extends LibraryCommandletExecutor<string> {
+  protected logName = 'force_source_deploy_with_manifest';
+  protected executionName = 'Deploy With Manifest (beta)';
+
+  protected async run(response: ContinueResponse<string>): Promise<boolean> {
+    try {
+      const ws = await WorkingSet.fromManifestFile(response.data, {
+        resolve: join(getRootWorkspacePath(), await SfdxPackageDirectories.getDefaultPackageDir() ?? '')
```
brpowell, line 54: The behavior of the CLI is that it will scan all package directories for matching components in the manifest - I wrongly assumed it was just the default. Instead we can set the resolve path to be just the root workspace path to get the same behavior. However, if there's a node_modules folder this will delay the operation since it'll wastefully iterate through this.

One way around this is to add node_modules to the forceignore file. Another is that we could enhance the library api to give a set of directories to explicitly check against. I think we should do the latter since that's akin to CLI, but should we check that in to the library before going forward with this?

PR #2768, opened 2020-11-30
chore: add git clean to clean script

sfsholden:
### What does this PR do?
Adds `git clean -xfd` to the clean script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "circleci:artifacts": "node scripts/download-vsix-from-circleci.js",
     "commit-init": "commitizen init cz-conventional-changelog --save-dev --save-exact --force",
     "commit": "git-cz",
-    "clean": "lerna run clean",
+    "clean": "lerna run clean && git clean -xfd",
     "compile": "lerna run --stream compile",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "circleci:artifacts": "node scripts/download-vsix-from-circleci.js",
     "commit-init": "commitizen init cz-conventional-changelog --save-dev --save-exact --force",
     "commit": "git-cz",
-    "clean": "lerna run clean",
+    "clean": "lerna run clean && git clean -xfd",
```
lcampos, line 34: Do we run the `clean` script on any of our automation flows ? I'm asking because this will also remove `node_modules` which might break something.

sfsholden: It does not look like it. We do call `git clean -xfd` directly within `cleanup-release-branch-after-release-issues.js`, `create-release-branch.js`, and `download-vsix-from-circle.js`. I'm not seeing a reference directly to this clean script though.

lcampos: Sounds good 👍

lcampos (approved):

PR #2772, opened 2020-12-01
Remove core extension as dependency on SOQL extension part 2

jgellin-sf:
### What does this PR do?
This is the second of ~~two~~ three PRs (see [previous](https://github.com/forcedotcom/salesforcedx-vscode/pull/2743)) that will remove the core extension as a dependency on the SOQL extension. This PR moves the WorkspaceContext singleton to the utilities package for consumption by other extensions. It also maintains the existing API in the core extension so that other users of this class can continue to do so. This PR does not yet actually remove the core extension dependency on the SOQL extension due to some churn in that extension's code. That change will follow in a third PR.

@W-8295301@

### Functionality Before
N/A

### Functionality After
N/A


packages/salesforcedx-utils-vscode/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/chai": "^4.0.0",
     "@types/cross-spawn": "6.0.0",
     "@types/mocha": "^5",
-    "@types/node": "12.0.12",
+    "@types/node": "14.14.10",
```
jgellin-sf, line 26: This change to do upstream dependency changes.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -20,7 +20,7 @@
   "devDependencies": {
     "@types/chai": "^4.0.0",
     "@types/mocha": "^5",
-    "@types/node": "12.0.12",
+    "@types/node": "14.14.10",
```
jgellin-sf, line 23: This change to do upstream dependency changes.

packages/salesforcedx-utils-vscode/src/auth/authUtil.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import { Aliases } from '@salesforce/core';
+import { ConfigUtil } from '../config';
+import {
+  DEFAULT_USERNAME_KEY
+} from '../constants';
+import { TelemetryService } from '../telemetry';
+
+export class AuthUtil {
```
jgellin-sf, line 15: Minimal auth function necessary to maintain current functioning of workspace context,

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -0,0 +1,72 @@
...
+  Global,
+  None
+}
+
+// This class should be reworked or removed once the ConfigAggregator correctly checks
+// local as well as global configs. It's also worth noting that ConfigAggregator, according
+// to its docs checks local, global and environment and, for our purposes, environment may
+// not be viable.
+
+export class ConfigUtil {
```
jgellin-sf, line 24: Minimal config function necessary to maintain current functioning of workspace context.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export interface OrgInfo {
+  username?: string;
+  alias?: string;
+}
+
+/**
+ * Manages the context of a workspace during a session with an open SFDX project.
+ */
+export class WorkspaceContextUtil {
```
jgellin-sf, line 24: Same as old WorkspaceContext singleton without org type handling.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -1,47 +1,29 @@
...
-
-export interface OrgInfo {
-  username?: string;
-  alias?: string;
-}
 
 /**
  * Manages the context of a workspace during a session with an open SFDX project.
  */
 export class WorkspaceContext {
```
jgellin-sf, line 15: WorkspaceContext is now basically a wrapper around WorkspaceContextUtil, with org type handling now done in the handleCliConfigChange listener.

packages/salesforcedx-sobjects-faux-generator/package.json
lcampos: The latest versions of VSCode still run on NodeJS 12.x.x so I would advice against changing this, you could accidentally start using new functionality that is not available.

packages/salesforcedx-utils-vscode/package.json
lcampos: We can't change this in the utils module without affecting the rest of the extensions, let's keep it as it is.

lcampos (changes requested):

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -10,7 +10,7 @@
     "Other"
   ],
   "dependencies": {
-    "@salesforce/core": "2.9.0",
+    "@salesforce/core": "2.11.0",
```
lcampos, line 13: We changed this yesterday, just rebase to the latest on develop

packages/salesforcedx-utils-vscode/src/auth/authUtil.ts
```diff
@@ -0,0 +1,47 @@
...
+        DEFAULT_USERNAME_KEY
+      );
+      if (defaultUserName === undefined) {
+        return undefined;
+      }
+
+      return JSON.stringify(defaultUserName).replace(/\"/g, '');
+    } catch (err) {
+      console.error(err);
+      await TelemetryService.getInstance().sendException('get_default_username_alias', err.message);
```
lcampos, line 39: I remember running into issues when using the telemetry service in other code in the utils module where it wouldn't get initialized for the correct extension since the utils code ran before the extension initializes. Did you see something like that ?

packages/salesforcedx-utils-vscode/package.json
jgellin-sf: @lcampos I starting having trouble compiling just a few hours ago. The reason was that some dependency of a dependency that was included via ^ had updated to node v14 and this was the only way to make typescript happy. That, or maybe I just got myself into a weird state. Either way. I've reverted these changes and re-installed everything and things are working again. ¯\_(ツ)_/¯

packages/salesforcedx-sobjects-faux-generator/package.json
jgellin-sf: Fixed in '...revert...' commit

packages/salesforcedx-utils-vscode/src/auth/authUtil.ts
jgellin-sf: I'll have to think about this. At the moment, this code is only called by WorkspaceContextUtil code that would only be called after WorkspaceContextUtil has been initialized to an extension context. So the extension should be initialized. But this code could be invoked in different ways in the future. We could test for telemetry activation (`if (TelemetryService.getInstance().getReporter())`) or allow the exception to go uncaught and catch/handle telemetry at a higher level. 

Note: In looking at the above problem, I noticed that the telemetry service suffers from the same singleton problem as the channel service had (fixed by changing from a singleton channel service to a singleton-per-channel channel service), so whichever extension initializes first will get all the telemetry assigned to it. This should be changed to a singleton-per-extension telemetry service.

lcampos (approved):

packages/salesforcedx-utils-vscode/src/auth/authUtil.ts
jgellin-sf: FWIW it turns out the reporter check is done in the telemetry service itself. So we won't get exceptions in any case.

lcampos: very cool 👍

dehru:
I've merged in develop but looks like CI be failing now.

PR #2781, opened 2020-12-02
chore: improve . vscodeignore in utils package

dehru:
### What does this PR do?
This PR reduces the size of the soql-biulder extension by 9mb.

@W-8529985@

### Functionality Before
21.5 mb

### Functionality After
12.5 mb

Screen Shot 2020-12-02 at 3 49 54 PM

### DIFF
--- a/packages/salesforcedx-vscode-soql/.vscodeignore
+++ b/packages/salesforcedx-vscode-soql/.vscodeignore
@@ -40,10 +40,6 @@ node_modules/rxjs/src/**
 node_modules/rxjs/bundles/**
 node_modules/rxjs/_esm5/**
 node_modules/rxjs/_esm2015/**
-node_modules/inquirer/node_modules/rxjs/src/**
-node_modules/inquirer/node_modules/rxjs/bundles/**
-node_modules/inquirer/node_modules/rxjs/_esm5/**
-node_modules/inquirer/node_modules/rxjs/_esm2015/**
 
 # ignore unused jsforce code
 node_modules/jsforce/test/**
--- a/packages/salesforcedx-vscode-soql/.vscodeignore
+++ b/packages/salesforcedx-vscode-soql/.vscodeignore
@@ -12,7 +12,47 @@ tsconfig.json
 *.xml
 **/*[!README].md
 *.vsix
+
+# local rxjs node modules
 node_modules/rxjs/src/**
 node_modules/rxjs/bundles/**
 node_modules/rxjs/_esm5/**
 node_modules/rxjs/_esm2015/**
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-soql/.vscodeignore
```diff
@@ -12,7 +12,43 @@ tsconfig.json
...
+node_modules/@salesforce/salesforcedx-utils-vscode/node_modules/**
+node_modules/@salesforce/salesforcedx-utils-vscode/src/**
+node_modules/@salesforce/salesforcedx-utils-vscode/test/**
+node_modules/@salesforce/salesforcedx-utils-vscode/.vscode
+node_modules/@salesforce/salesforcedx-utils-vscode/.vscodeignore
+node_modules/@salesforce/salesforcedx-utils-vscode/package-lock.json
+node_modules/@salesforce/salesforcedx-utils-vscode/tsconfig.json
+node_modules/@salesforce/salesforcedx-utils-vscode/webpack.config.js
+node_modules/@salesforce/salesforcedx-utils-vscode/out/test/**
+node_modules/@salesforce/salesforcedx-utils-vscode/node_modules/rxjs/src/**
```
lcampos, line 33: I'm stealing this and adding it to the other extensions

lcampos (approved):

packages/salesforcedx-vscode-soql/.vscodeignore
```diff
@@ -12,7 +12,43 @@ tsconfig.json
...
 *.vsix
+
+# local rxjs node modules
 node_modules/rxjs/src/**
 node_modules/rxjs/bundles/**
 node_modules/rxjs/_esm5/**
-node_modules/rxjs/_esm2015/**
\ No newline at end of file
+node_modules/rxjs/_esm2015/**
+node_modules/typescript/**
```
lcampos, line 21: Why was typescript getting added to the vsix ?

dehru: I'm not sure. I'd like to spend more time digging into the dependency chain with the work item.  But for now, i'm gonna merge this for todays release.

 I used this to find the biggest directories in node modules.  
`du -a . | sort -n -r | head -n 20`

lcampos: The lwc extension has a similar problem where `typescript` is getting imported as a child dependency via `jest-editor-support` and we can't remove it (yet) since it breaks all the testing capabilities it provides for LWC users. 

You should check if you didn't break anything in the extension by using the vsix, unfortunately the automation won't catch this for you.

PR #2783, opened 2020-12-03
fix: remove unused rxjs code imported by salesforcedx-utils-vscode

lcampos:
### What does this PR do?
Remove more unused dependency code from the vsix files


lcampos:
Closing since this is only useful for extensions where different versions of `@salesforce/core` get bundled.

PR #2787, opened 2020-12-03
feat: deploy with manifest in performance beta

brpowell:
### What does this PR do?

Enables beta performance enhancements for SFDX: Deploy Source in Manifest from Org

@W-8246812@


packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -36,6 +42,40 @@ export class ForceSourceDeployManifestExecutor extends BaseDeployExecutor {
   }
 }
 
+export class LibrarySourceDeployManifestExecutor extends LibraryCommandletExecutor<string> {
+  protected logName = 'force_source_deploy_with_manifest';
+  protected executionName = 'Deploy With Manifest (beta)';
+
+  protected async run(response: ContinueResponse<string>): Promise<boolean> {
+    try {
```
lcampos, line 50: We should add a catch condition so we can post the error to the output panel, etc

lcampos: nevermind, you just reminded me we handle that in the `execute` method of `LibraryCommandletExecutor`

lcampos (approved):

PR #2791, opened 2020-12-07
chore: support telemetry for multi-component cli deploy

brpowell:
### What does this PR do?

Add telemetry support for multi-component deploys using the CLI

@W-8541106@


packages/salesforcedx-vscode-core/src/commands/baseDeployCommand.ts
```diff
@@ -60,20 +60,17 @@ export abstract class BaseDeployExecutor extends SfdxCommandletExecutor<
...
-        const metadataCount = JSON.stringify(createComponentCount([...ws]));
-        properties = { metadataCount };
-        // registry does not handle multiple paths. only log component count for single paths
-      } catch (e) {
-        telemetryService.sendException(
-          e.name,
-          'error detecting deploy components'
-        );
+      const telemetry = new TelemetryBuilder();
+      if (execFilePathOrPaths) {
```
lcampos, line 64: Let's add this on a try/catch. Looking at the telemetry this code generates the most common exception.

brpowell: Yeah I'll add that back in. That exception count should drop a decent amount after this change since we're handling multiple paths now.

lcampos: @brpowell should we tweak the exception info a bit ? Right now it's impossible to know what caused it.

lcampos (approved):

PR #2792, opened 2020-12-07
Initial code completion for SOQL

dobladez:
### What does this PR do?

Implements code-completion for SOQL.
* It supports the most common keywords, completion of `FROM` clause with SObject names and completion of `SELECT` fields with SObject fields.
* Basic support for nested SELECTs

Also:
 * Extract interaction with other Salesforce extensions into separate `sfdx.ts`
 * On SOQL builder and editor only propose "queryable" objects (fixes forcedotcom/soql-tooling#83)

Depends on https://github.com/forcedotcom/soql-tooling/pull/99

@W-8431809@, @W-8432064@

### Functionality After
soql_completion_pr


packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -153,31 +138,47 @@ export class SOQLEditorInstance {
...
       case MessageType.SOBJECTS_REQUEST: {
-        this.retrieveSObjects().catch(() => {
-          channelService.appendLine(
-            `An error occurred while handling a request for object names.`
-          );
-        });
+        retrieveSObjects()
+          .then(sobjectNames => this.updateSObjects(sobjectNames))
+          .catch(e => {
+            if (e.errorCode !== 'NOT_FOUND') {
```
dehru, line 173: Is there a reason you are only including 'NOT_FOUND'?

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
```diff
@@ -0,0 +1,84 @@
...
+export const channelService = ChannelService.getInstance(
+  nls.localize('soql_channel_name')
+);
+
+const sfdxCoreExtension = vscode.extensions.getExtension(
+  'salesforce.salesforcedx-vscode-core'
+);
+const sfdxCoreExports = sfdxCoreExtension
+  ? sfdxCoreExtension.exports
+  : undefined;
```
dehru, line 23: Why is this necessary?  Wouldn't this be fatal?

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
```diff
@@ -0,0 +1,84 @@
...
+): Promise<void> {
+  try {
+    const conn = await workspaceContext.getConnection();
+    return f(conn);
+  } catch (e) {
+    channelService.appendLine(e);
+  }
+}
+export async function retrieveSObjects(): Promise<string[]> {
+  return new Promise<string[]>((resolve, reject) => {
```
dehru, line 37: Thank you for this.

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    return withSFConnection(async conn => {
+      conn.describeGlobal$((err, describeGlobalResult) => {
+        if (err) {
+          reject(err);
+        } else if (describeGlobalResult) {
+          const sobjectNames: string[] = describeGlobalResult.sobjects
+            .filter(o => o.queryable)
```
dehru, line 44: Oh, this is helpful.  What kinds of objects did this filter out?

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
dobladez: Good point. I simply moved the existing code without giving it much thought.

I'll add an assertion/safety check. The Extension should fail at activation time if that's `undefined`.

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
dobladez: Some SObjects cannot be queried for.  See: https://github.com/forcedotcom/soql-tooling/issues/83

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dobladez: Hmm... for `SOBJECTS_REQUEST ` it probably shouldn't be done.

For the case of `SOBJECT_METADATA_REQUEST` however, what happens is that the query text might refer to a non-existent SObject. IMHO, we shouldn't show an error toast here... it's an error we should expect (say, while the user is typing a FROM clause on the editor) and should handle explicitly, for example: disabling the Fields selection dropdown, and may be show some explanation text (within the SOQL UI) that an object with that name could not be found).

dehru: Gotcha.  I won't send 'SOBJECT_METADATA_REQUEST' to telemetry either or print it in the channel.  I added a story to handle the case in the UI.

And if they do run the query, they will see the error that the field doesn't exist anyway.  The error information coming back from run query is informative.

packages/salesforcedx-vscode-soql/src/sfdx/index.ts
dobladez: done

dehru (approved):

jgellin-sf:
This PR is QA complete. We are ready to merge once we get a codeowner approval. @forcedotcom/pdt

.vscode/launch.json
```diff
@@ -595,6 +595,29 @@
...
+      "sourceMaps": true,
+      "smartStep": true,
+      "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
+      "sourceMapPathOverrides": {
+        "../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
+        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///*": "*"
+      },
+      "XXXXXpreLaunchTask": "Compile",
```
lcampos, line 619:
```suggestion
      "preLaunchTask": "Compile",
```

packages/salesforcedx-vscode-soql/src/client/client.ts
```diff
@@ -103,3 +125,73 @@ export function stopLanguageClient(): Thenable<void> | undefined {
...
+    }
+  }
+
+  //  Expand SObject fields
+  const SOBJECT_FIELDS_PLACEHOLDER = /__SOBJECT_FIELDS_PLACEHOLDER:(\w+)/;
+  const sobjectFieldsIdx = completionItems.findIndex(item =>
+    SOBJECT_FIELDS_PLACEHOLDER.test(item.label)
+  );
+  if (sobjectFieldsIdx >= 0) {
+    const m = completionItems[sobjectFieldsIdx].label.match(
```
lcampos, line 155: Let's use a more meaningful name instead of just calling this `m`

packages/salesforcedx-vscode-soql/src/sfdx.ts
```diff
@@ -0,0 +1,87 @@
...
+import { ChannelService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
+import { DescribeSObjectResult } from 'jsforce';
+import * as vscode from 'vscode';
+import { nls } from './messages';
+
+export const channelService = ChannelService.getInstance(
+  nls.localize('soql_channel_name')
+);
+
+const sfdxCoreExtension = vscode.extensions.getExtension(
```
lcampos, line 18: We shouldn't be using this pattern anymore, replace it with the modules on `salesforcedx-utils-vscode` e.g. https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts

packages/salesforcedx-vscode-soql/src/sfdx.ts
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { Connection } from '@salesforce/core';
```
lcampos, line 8: What is the advantage of having all the services in one file like this ?

packages/salesforcedx-vscode-soql/test/vscode-integration/testUtilities.ts
```diff
@@ -24,10 +24,21 @@ import {
   QueryDataViewService
 } from '../../src/queryDataView/queryDataViewService';
 
+const sfdxCoreExtension = vscode.extensions.getExtension(
```
lcampos, line 27: Same as above, let's stop using this pattern for depending on the core extension.

packages/salesforcedx-vscode-soql/src/client/client.ts
```diff
@@ -60,6 +59,28 @@ export async function startLanguageClient(
     synchronize: {
       configurationSection: 'soql',
       fileEvents: workspace.createFileSystemWatcher('**/*.soql')
+    },
+    middleware: {
```
lcampos, line 63: Can you provide an example on when this code is needed ? Which query and completions would I have to write to use this ?

lcampos (changes requested):
Update the code to use salesforcedx-utils-vscode instead of depending on the core extension

jgellin-sf:
@lcampos I have merged some preparatory PR's refactoring core/utils stuff so that we can finally sever the core dependency from the soql extension. Since this PR refactors the client, I was waiting for this PR to merge before finally removing the core dependency from this extension.

packages/salesforcedx-vscode-soql/src/sfdx.ts
dobladez: My idea was to minimize and put together all integration points from the SOQL extensions to the other Salesforce extensions. To reduce code duplication, simplify testing and reduce coupling from the SOQL extension to the other salesforce extensions.

For instance, the logic to load the core extension (which we have to change), is now on a single place instead of four.

packages/salesforcedx-vscode-soql/src/sfdx.ts
dobladez: Yes.  @jgellin-sf  has this covered on another PR on top of this one

packages/salesforcedx-vscode-soql/src/client/client.ts
dobladez: Not sure I understand the question... but let me try to give some context.

The connection to the Salesforce Org (to retrieve SObjects metadata, needed for completion) is already handled by the Extensions, where the LSP client runs.

So, we thought of some options:

1. Don't use LSP at all and implement completion inside the VSCode extension. Problem gone.
2. The LSP server could use its own separate connection. This would add some complexity to pass on the connection info/credentials from the extension to the server, making sure the right Org is hit (in sync with the extension) and prevent us from sharing caching.
3. The LSP server asks the client for metadata when needed. This would make each code-completion requests very chatty:
image
4. The LSP server returns special completion items that the client must interpret and replace ("expand") with sobjects metadata:
image

So, we went with option 4, which allows us to continue using LSP while reusing the connections/caching of the extensions. What you see on this PR is the client-side of that.

More info here: https://salesforce.quip.com/JxYxAoa17YxO#NMUACAGfgxx

I hope my explanation makes sense. I would be happy to have a live discussion and go over more details. We already have improvements and refactorings coming on another PR on top of this one. Thanks!

lcampos: @dobladez could you give me access to that document ?

We should talk about how you are approaching the LSP functionality. Right now you are asking the client to massage/process the data returned by it. This might be ok for the SOQL extension but will make the SOQL LSP difficult to use in other scenarios/tools. At the moment, we know of at least two other tools outside of the SOQL extension where we want to use the language server and I don't believe the current approach is the correct one.

@praksb @jgellin-sf @dobladez lets talk about this once we're back from the holidays.

dobladez: Sure. Quip tells me you already have access to the quip doc....  anyway, the document doesn't say much more regarding connection-handling.

packages/salesforcedx-vscode-soql/src/sfdx.ts
jgellin-sf: @lcampos FYI here is the PR for reference: https://github.com/forcedotcom/salesforcedx-vscode/pull/2837

packages/salesforcedx-vscode-soql/test/vscode-integration/testUtilities.ts
```diff
@@ -64,18 +75,109 @@ export const mockQueryData: QueryResult<JsonMap> = {
...
-    describeGlobal$: (callback: (err: Error | undefined, resp: any) => void) => callback(undefined, mockDescribeGlobalResponse),
-    describe$: (name: string, callback: (err: Error | undefined, resp: any) => void) => callback(undefined, mockSObject),
+    describeGlobal$: (callback: (err: Error | undefined, resp: any) => void) =>
+      callback(undefined, mockDescribeGlobalResponse),
+    describe$: (
+      name: string,
+      callback: (err: Error | undefined, resp: any) => void
+    ) => {
+      const sobjectMetadata = mockSObjects.find(s => s.name === name);
+      // if (!sobjectMetadata) {
```
lcampos, line 176: nit: remove commented out code

lcampos (approved):

packages/salesforcedx-vscode-soql/src/client/client.ts
lcampos: circling back, we met and decided we'll change this in a future PR since it requires changes to the language server.

PR #2794, opened 2020-12-07
Jhork- Update issue and Q&A url, join publishing automation.

jonnyhork:
### What does this PR do?
The [soql-tooling](https://github.com/forcedotcom/soql-tooling) repo is going to be open source. We would like to direct user issues and q&a there instead of this repo.
We also want to add the two scripts that will allow the tooling repo to be published with more regularity like the rest of the extensions.
@W-8437912@


jonnyhork:
@lcampos I think you mentioned that the soql extension is specifically excluded from being published somewhere. Besides the scripts in this PR, is there some other change I should make to join the publishing party with other extensions?

lcampos (approved):

Issue #2490, opened 2020-09-06
SObject "Event" is not generated by the SFDX: Refresh SObject Definitions command


khjuan:
### Summary

After executing the SFDX: Refresh SObject Definitions command, the standard "Event" sobject was not created.

### Steps To Reproduce:

1.  In VSCode for windows, Ctrl-Shift P to bring up the command palette. Run the SFDX: Refresh SObject Definitions for the connected ORG
2. Check the .sfdx/tools/sobjects/standardObjects folder, and the Event.cls is not there. 

### Expected result

Expected Event.cls to be created as it is a standard SObject.

### Actual result

Event.cls cannot be located in .sfdx/tools/sobjects/standardObjects folder.

### Additional information

Executing the **sfdx force:schema:sobject:list -c standard** in the command line yielded the Event Sobject in the result.
Suspect the fauxClassGenerator.isRequiredSObject has inadvertedly excluded a legimate object name.

  public isRequiredSObject(sobject: string): boolean {
    // Ignore all sobjects that end with Share or History or Feed or Event
    return !/Share$|History$|Feed$|**Event$**/.test(sobject);
  }

**VS Code Version**: 1.48.2

**SFDX CLI Version**:  7.73.1

**OS and version**: Linux xxxxx 4.19.104-microsoft-standard #1 SMP

mohanraj-r:
@khjuan This issue is now fixed. I tried (with v54.6.1) SFDX: Refresh SObject Definitions and verified that Event.cls is created under .sfdx/tools/sobjects/standardObjects folder.

Issue #2480, opened 2020-09-02
Refresh sObject Definition does not work for Custom Sharing Rule Objects


monirkohi2006:
### Summary

Refresh sObject Definition does not work for Custom Sharing Rule Objects

### Steps To Reproduce:

1.  Define a custom sharing rule object. Ex: MyObject__Share shareRule = new MyObject__Share();
2. You should see the object and the instance with an underline stating "Variable does not exist: MyObject__Share".

### Expected result

SFDX should understand the share rule object as a valid object or should not underline it.

### Actual result

SFDX syntax checker doesn't understand the share rule object syntax.

### Additional information

![image](https://user-images.githubusercontent.com/32991357/92038424-33b20180-ed28-11ea-9548-e7eab44af759.png)

**VS Code Version**: Version: 1.48.2

**SFDX CLI Version**: sfdx-cli/7.68.6-d37008df83 darwin-x64 node-v12.18.3

**OS and version**: macOS Catalina & Version: 10.15.5

khjuan:
I believe this is by design. See related (but different) issue i raised https://github.com/forcedotcom/salesforcedx-vscode/issues/2490.

Though i am not sure why it wanted to exclude all the __Share sobjects, I do see some legitimate uses of such sobjects in apex codes, especially the test classes. Perhaps this could be made a configurable feature of some sort?

PR #2806, opened 2020-12-09
fix(refresh-sobject): manual refresh does not filter sobjects

maaaaarco:
### What does this PR do?
When manually run, the command _Refresh SObject Definitions_ does not exclude any of the SObject retrieved. 

#2490
#2480

### Functionality Before
SObjects with name ending in _Share|History|Feed|Event_ were excluded. This behaviour was introduced with #2412.

### Functionality After
When manually run none of the object retrieved is excluded.

### Personal Note
I do not understand why excluding those sobjects should be the desired behaviour. The only benefit I can think of is that when the refresh happens on startup it's going to be faster.
There are many valid use cases in Apex to work with objects ending in  `Share|History|Feed|Event`. In my specific case I'm implementing a batch for custom apex sharing, so I have to work with objects like `MyCustomObject__Share`.
Not having the faux generated cause the autocompletion engine to stop working **completely** in all classes that reference those objects... and that's really annoying as you can imagine 😭
Please consider removing that filter completely.


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -189,7 +189,13 @@ export class FauxClassGenerator {
         err.stack
       );
     }
-    const filteredSObjects = sobjects.filter(this.isRequiredSObject);
+
+    // when run manually does not exclude any object
+    const filteredSObjects =
```
lcampos, line 194: Since we have 3 different options when running `SFDX: Refresh SObject Definitions` (All SObjects, Custom SObjects & Standard SObjects), I think we should update this to be exclusive to `All SObjects` and for everything else to still use the filter.

To provide some context, we introduced the use of an sobject filter to reduce the amount of synced objects since an org without any customizations can have between 400-600 sobjects. This also helps the Apex language server since the faux classes generated for the synced sobjects are seen by the language server as another Apex class. If a user has a small project and syncs 500ish sobjects the project now becomes a larger project to process by the Apex language server and the user might start running into performance problems (depending on their hardware). 

@maaaaarco let me know what you think

maaaaarco: @lcampos I really appreciate the time for sharing the rationale behind it. I do understand the performance concerns you have and I do agree with them. When working with a newly created project or a small ones, loading those objects is most likely an overkill... On the other hand, I think that forcing these decisions on your users, without giving them the option to opt out, is a bad design pattern. And _I believe_ this to be true for whatever kind of application we build as developers. In my 6 years career as a Salesforce Developer working in consulting business I've seen many projects relying on those objects (especially the custom _Share_ objects).
Long story short, I agree with keeping the filter for most use cases but I do want the option to load all of them... I'm totally happy with the solution you suggested (disable the filter only when the _All SObjects_ option is selected) and I'm keen to update this PR if you would let it go through... We could even add a fourth option one named _Standard and Customs_ that would behave as the _All SObjects_ is doing now (keeping the filter active).
Let me know what do you think ;)

lcampos: @maaaaarco making that small tweak to keep the change exclusive to All SObjects would be enough for us to merge this PR. Adding another option to the menu might not be needed at the moment, we could discuss adding that as a follow up PR but I think it would be almost the same as the existing All SObjects option.

I do agree with you on giving users a way to opt out of certain tooling behavior. Specific to this command, it is one of the least used commands so we are still iterating over it based on the limited feedback and metrics we have. As people use it more we'll keep making changes to it so we get to a point where we have sensible defaults and settings to customize it if needed.

maaaaarco: @lcampos works for me, I'll have a look at this later next year after 4th of Jan. Again thanks for the background info 👍

lcampos: @maaaaarco sounds good, thanks for contributing to the project. Happy holidays and talk to you next year.

maaaaarco:
@lcampos I updated the PR as discussed ;)
Only when user manually runs the _Refresh All SObjects_ the extension does not filter any object.

lcampos (approved):

lcampos:
> @lcampos I updated the PR as discussed ;)
> Only when a user manually runs _Refresh All SObjects_ the extension does not filter any object.

The change looks good so I've merged it. Thanks again for the contribution @maaaaarco

PR #2809, opened 2020-12-10
feat(soql): new query command

jgellin-sf:
### What does this PR do?
This PR implements the command "SOQL: New Query", which opens an untitled, unsaved, empty query in the SOQL Builder.  This command is enabled in the command palette when an SFDX project is open.

@W-8516491@

### Functionality Before
User must create .soql file, then "Open With" SOQL Builder to open a new query in the SOQL Builder.

### Functionality After
User uses command "SOQL: New Query" to open a new query in the SOQL Builder.
createquery

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -173,6 +173,29 @@
       "internalConsoleOptions": "openOnSessionStart"
     },
     {
+      "name": "Launch Tests - SOQL - VSCode Integration Tests",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "${workspaceFolder}/packages/system-tests/assets/sfdx-simple",
+        "--extensionDevelopmentPath=${workspaceFolder}/packages",
+        "--extensionTestsPath=${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration"
+      ],
+      "stopOnEntry": false,
+      "sourceMaps": true,
+      "smartStep": true,
+      "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
+      "sourceMapPathOverrides": {
+        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///salesforcedx-sobjects-faux-generator/./*": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/*",
+        "webpack:///*": "*"
+      },
+      "preLaunchTask": "Compile",
+      "internalConsoleOptions": "openOnSessionStart"
+    },
+    {
       "name": "Launch Tests - Apex - VSCode Integration Tests",
       "type": "extensionHost",
       "request": "launch",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -189,7 +189,6 @@
       "sourceMapPathOverrides": {
         "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
         "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
-        "webpack:///salesforcedx-sobjects-faux-generator/./*": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/*",
         "webpack:///*": "*"
       },
       "preLaunchTask": "Compile",
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -125,7 +125,7 @@
     "commands": [
       {
         "command": "soql.builder.open.new",
-        "title": "SOQL: New Query"
+        "title": "SOQL: Create Query in SOQL Builder"
       }
     ],
     "menus": {
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -93,7 +93,7 @@
   "contributes": {
     "customEditors": [
       {
-        "viewType": "soqlCustom.soql",
+        "viewType": "%soqlCustom_soql%",
         "displayName": "SOQL Builder",
         "selector": [
           {
@@ -125,7 +125,7 @@
     "commands": [
       {
         "command": "soql.builder.open.new",
-        "title": "SOQL: Create Query in SOQL Builder"
+        "title": "%soql_builder_open_new%"
       }
     ],
     "menus": {
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -121,6 +121,20 @@
         "scopeName": "source.soql",
         "path": "./node_modules/@salesforce/apex-tmlanguage/grammars/soql.tmLanguage"
       }
-    ]
+    ],
+    "commands": [
+      {
+        "command": "soql.builder.open.new",
+        "title": "SOQL: New Query"
+      }
+    ],
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "soql.builder.open.new",
+          "when": "sfdx:project_opened"
+        }
+      ]
+    }
   }
 }
--- a/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
+++ b/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { EDITOR_VIEW_TYPE } from '../constants';
+import { telemetryService } from '../telemetry';
+
+export async function soqlOpenNew() {
+  telemetryService.sendCommandEvent('soql_builder_open_new', process.hrtime());
+
+  if (vscode.workspace) {
+    // create untitled file
+    const newFileUri = vscode.Uri.file(
+      path.join(vscode.workspace.rootPath!, 'untitled.soql')
+    ).with({ scheme: 'untitled' });
+    await vscode.workspace.openTextDocument(newFileUri);
+
+    // open with SOQL builder
+    vscode.commands.executeCommand(
+      'vscode.openWith',
+      newFileUri,
+      EDITOR_VIEW_TYPE
+    );
+  }
+}
--- a/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
+++ b/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
@@ -15,15 +15,15 @@ export async function soqlOpenNew() {
 
   if (vscode.workspace) {
     // create untitled file
-    const newFileUri = vscode.Uri.file(
-      path.join(vscode.workspace.rootPath!, 'untitled.soql')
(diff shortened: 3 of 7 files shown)


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -121,6 +121,20 @@
         "scopeName": "source.soql",
         "path": "./node_modules/@salesforce/apex-tmlanguage/grammars/soql.tmLanguage"
       }
-    ]
+    ],
+    "commands": [
+      {
+        "command": "soql.builder.open.new",
+        "title": "SOQL: New Query"
```
dobladez, line 128: Should the command title include the word "File" ? 🤔

dobladez:
Nice 👍. We should have done this earlier!

packages/salesforcedx-vscode-soql/package.json
ekapner: Throwing this out as a discussion point. Should the emphasis be on creating the file or the fact that we're launching SOQL Builder. This will be the new entry point, yes? So should it be SOQL: Open Builder or SOQL: Open SOQL Builder (I don't know if any of our commands use Launch or if that's an ok option based on our UI standards.

ekapner: When I open command palette and do a search, "Open" seems to be the most popular choice.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export async function soqlOpenNew() {
+  telemetryService.sendCommandEvent('soql_builder_open_new', process.hrtime());
+
+  if (vscode.workspace) {
+    // create untitled file
+    const newFileUri = vscode.Uri.file(
+      path.join(vscode.workspace.rootPath!, 'untitled.soql')
+    ).with({ scheme: 'untitled' });
+    await vscode.workspace.openTextDocument(newFileUri);
```
brpowell, line 21: Creating a new soql file feels like a similar UX to creating a new metadata template (create apex class, create lwc, etc). In that flow we prompt for a filename with an input box. If y'all don't think it makes sense here that's cool, just putting the thought out there for discussion.

ekapner: +1 to prompting for a file name

jgellin-sf: Good point. I think there are arguments to be made for both flows.  

Personally, I feel that since .soql files are ephemeral artifacts (they don't actually get deployed anywhere), the persistence of the file is not as important. The 'untitled' user flow allows a user to create and test their query, copy-paste to Apex if necessary, and then save as desired. 

@clairebianchi Which flow do you prefer for SOQL files?

packages/salesforcedx-vscode-soql/package.json
jgellin-sf: I'm ready to defer to @ekapner on this one.

jonnyhork: @ekapner This functionality is one of two(or more) ways to get to the builder but not necessarily a primary entry point. The intention is to create an easy pathway for customers to use the SOQL Builder if they don't have an existing `.soql` file.
So there is a distinction between
1. Opening and _existing_ file with the builder(current functionality) `SOQL: Open file with SOQL Builder` `SOQL: Open current editor with SOQL Builder`
2. Creating a new .soql file and using the builder create a query (this PR) `SOQL: Start a new query with SOQL Builder` `SOQL: Create a new Query with SOQL Builder`
3. Just a single command to "open" the builder (does not make sense) ' SOQL: Open Builder'

Because the builder is really an editor it always needs to be open with the context of a file, whether its new or existing.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jonnyhork: Keep in mind that at this point when they save the query results it will use the title of the document, so they will be saving `untitled.soql.csv`. In this case the persistence of the query may not be as important as the persistence of the results but that will still depend on the file being named.

jonnyhork: I think there is work in the backlog or at least discussion to prompt the user for a filename and location when they save the data. If that work is done and they are promoted with a save dialog then we may not have to be so concerned with asking on the front end with this command.

jonnyhork, line 21: I can't remember what this does exactly, it returns a text document right? IIRC opening a doc was a two step process or something. They separate "opening" a text document with viewing it in the editor or something like that?

Does this open the text editor ` vscode.workspace.openTextDocument(newFileUri);` 
then we quickly run ` vscode.commands.executeCommand(
      'vscode.openWith',
      newFileUri,
      EDITOR_VIEW_TYPE
    );` to switch it to the builder UI?

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -0,0 +1,30 @@
...
+import { EDITOR_VIEW_TYPE } from '../constants';
+import { telemetryService } from '../telemetry';
+
+export async function soqlOpenNew() {
+  telemetryService.sendCommandEvent('soql_builder_open_new', process.hrtime());
+
+  if (vscode.workspace) {
+    // create untitled file
+    const newFileUri = vscode.Uri.file(
+      path.join(vscode.workspace.rootPath!, 'untitled.soql')
```
jonnyhork, line 19: I wonder if until we give the customers the option to save results where they want we should be sticking new `.soql` files in the same place we expect them to be in an sfdx-project? 
`scripts/soql` as a default? Just to keep things in the same place.
Though if we are opinionated now, that might mean more work to undo it in the future...

Just putting it out there, either way.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jgellin-sf: `openTextDocument` enables a `TextDocument` object to be manipulated. It is the model for a file, the same model the an editor acts upon, but the model can exist independent of an editor as well. So that instruction does not open an editor.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jgellin-sf: Most of the artifacts we deal with in a project have designated locations, which makes means that creating an Apex class or an LWC component requires only a name. If we are _not_ opinionated, that means we have to gather a folder _and_ a filename from the customer. This is one of the reasons I like the untitled flow. It gets them in the editor quickly, and they may not even care to have a file at the end if they're just there to build and test.

packages/salesforcedx-vscode-soql/package.json
ekapner: Thanks, @jonnyhork, for the thorough explanation. Makes sense. 

So...how many menu items do we need? Throwing these out:
SOQL: Open File in SOQL Builder
SOQL: Open File in Text Editor
SOQL: Create a Query in SOQL Builder (unsure the distinction re: Start a new query vs Create a new...)

jonnyhork: Sorry for the confusion, the commands in my comment were just examples of what wording might look like for each option. Not to be taken as unique commands that we will have.
For this PR we only need one for option 2, whatever wording you want =)

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jonnyhork: yeah, thats generally what I thought. I remember finding the name method `openTextDocument` to be misleading.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jonnyhork: That makes sense, I also like this part too
>  It gets them in the editor quickly, and they may not even care to have a file at the end if they're just there to build and test.

packages/salesforcedx-vscode-soql/package.json
ekapner: What do you all think about these? I'm torn. But I like clarifying that we're talking soql builder. Looking at other VS code commands, let's get rid of the articles. Leaning toward Create.
SOQL: Create Query in SOQL Builder
SOQL: New Query in SOQL Builder

jgellin-sf: I like 'Create'

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
praksb: What would happen if I am building more than one query at a time and I launch SOQL builder more than once simultaneously? I think if we go the 'untitled' file route then we need to have some logic to handle that case. VS Code file name format for untitled is "Untitled-1", "Untitled-2" etc. so we need to check if an untitled file name already exists before generating it.

If we let the user pick/name their file then we may not have to worry about this case.

jgellin-sf:
The latest commits update the behavior to match the default VS Code Untitled behavior, including numbering, etc, and also update the command name to "SOQL: Create Query in SOQL Builder".

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
jonnyhork: This was not an issue for me during testing. The files incremented as you'd expect, VS Code handled it for us.

jonnyhork (approved):
QA Passed!

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -121,6 +121,20 @@
         "scopeName": "source.soql",
         "path": "./node_modules/@salesforce/apex-tmlanguage/grammars/soql.tmLanguage"
       }
-    ]
+    ],
+    "commands": [
+      {
+        "command": "soql.builder.open.new",
+        "title": "SOQL: Create Query in SOQL Builder"
```
lcampos, line 128: Why do we want to prefix this command with `SOQL: ` instead of using `SFDX:` like the other commands in the other extensions ?

lcampos: Move this label to a `package.nls.json` e.g. https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.nls.json

.vscode/launch.json
```diff
@@ -172,6 +172,29 @@
...
+        "--extensionTestsPath=${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration"
+      ],
+      "stopOnEntry": false,
+      "sourceMaps": true,
+      "smartStep": true,
+      "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
+      "sourceMapPathOverrides": {
+        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
+        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
+        "webpack:///salesforcedx-sobjects-faux-generator/./*": "${workspaceFolder}/packages/salesforcedx-sobjects-faux-generator/*",
```
lcampos, line 192: You can remove this line on `salesforcedx-sobjects-faux-generator`

jgellin-sf: Thanks. Updated in `chore(soql): i18n in package.json and cleanup` commit.

packages/salesforcedx-vscode-soql/package.json
jgellin-sf: I've moved this label and also the editor display name to `package.nls.json` in the `chore(soql): i18n in package.json and cleanup` commit.

Also, I've changed the command title to have an 'SFDX' prefix. The story on that was that at first I created a command title to be as brief as possible but I wanted to make sure it was searchable on both "SOQL and "Query" ("SOQL: New Query"). We went back and forth and landed on "Create Query in SOQL Builder" but never changed the prefix (until now!).

lcampos (approved):
:shipit:

PR #2818, opened 2020-12-15
feat: do not emit message from custom editor to webview if soql has not changed

dehru:
### What does this PR do?
This PR captures any incoming soql statement from the soql-builder-ui and caches it.  Which it laters uses to compare with text changes in the text editor's soql statement, preventing unnecessary messaging back to the soql-builder-ui if the soql statement hasn't changed.

This is part of a fix to tackle some unwanted "flickering" behavior in the soql-builder-ui when typing into input fields.  It's also specifically mentioned in the VS Code Custom Editor docs here under heading "From TextDocument to webviews" of [this anchor of the docs](https://code.visualstudio.com/api/extension-guides/custom-editors#synchronizing-changes-with-the-textdocument)

@W-8555599@

### Functionality Before

_Notice as i'm typing into the input, adding or removing characters, multiple characters keep getting added/removed suddenly._

flickr-before

### Functionality After

_Notice as I'm typing into the input, adding or removing characters is smooth._

flickr-after


jgellin-sf (approved):
LGTM

jonnyhork (approved):
Thanks for taking care of this, it makes a huge difference in the UX!

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/soqlEditorInstance.test.ts
```diff
@@ -135,6 +135,42 @@ describe('SoqlEditorInstance should', () => {
...
+      payload: aQuery
+    });
+    // attempt to update webview with unchanged soql statement
+    instance.updateWebview(mockTextDocument);
+    expect(
+      postMessageSpy.callCount === 0,
+      `postMessageSpy callcount expected 0, but got ${postMessageSpy.callCount}`
+    );
+    // a second update with the same statement will send
+    instance.updateWebview(mockTextDocument);
```
lcampos, line 152: Not sure I completely understand this test. Shouldn't postMessageSpy call count still be 0 given there was no update to the query in the doc ?

dehru: We only trap the message to updateWebview once.  So the call to sendEvent ( which mocks an incoming event and I just renamed mockReceiveEvent() to make more sense ) sets the internal cache of the soql statement to `select a,b,c from somewhere`.  

Then the first call to updateWebview is muffled, but not the second.  Even if the soql statement is unchanged, i just need to trap the first.  As a safety measure, i compare the statements and clear the cache after the first call.   

This potential gotcha with the change event is ( [warned by the docs](https://code.visualstudio.com/api/extension-guides/custom-editors#synchronizing-changes-with-the-textdocument) ).

lcampos (approved):

PR #2821, opened 2020-12-16
fix: allow Event SObj generation (#2490)

XVRick:
### What does this PR do?
Allows generation of faux Event object but still prevents objects like __ChangeEvent

#2490

### Functionality Before
Event SObject was being filtered out of generation code.

### Functionality After
Event SObject no longer filtered out but original intent of filtering our objects that end in Event is still working


lcampos (approved):

lcampos:
@XVRick thanks for your contribution to the project. We've merged the change and it will be going out in the next release of the extensions.

PR #2810, opened 2020-12-10
Capture errors, localized messages to user, send to telemetry

dehru:
### What does this PR do?
This PR cleans up a [previous PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/2795) and incorporates that feedback.

This PR captures various errors, conditionally messages the user, and optionally sends them to telemetry as exceptions. 

### What issues does this PR fix or reference?
@W-8086238@

### Functionality Before
No exceptions sent to telemetry, inconsistent messages to the user, no i18n

### Functionality After
Some exceptions are available in app insights.  Emily has reviewed messaging to user.  I've incorporated feedback on which messaging options should be used for which types of messages.   Messages have been localized.

PR #2822, opened 2020-12-16
Running query indicator

jgellin-sf:
### What does this PR do?
This PR provides the VSCode-side function for indication that a query is running. When the `Run Query` button is pressed, it changes to a disabled `Running...` button and a notification progress indicator called `Running query` is shown until the query results are shown, at which point the `Run Query` button is restored.

@W-8207738@

### Functionality Before
When `Run Query` is pressed, there is no visual indication that it was pressed or that anything is happening.

### Functionality After
runquerywithprogress


packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -181,11 +182,23 @@ export class SOQLEditorInstance {
...
+            },
+            () => {
+              return this.handleRunQuery();
+            }
+          )
+          .then(undefined, () => {
+            channelService.appendLine(
+              `An error occurred while running the SOQL query.`
+            );
+            this.runQueryDone();
```
jonnyhork, line 200: I haven't seen this pattern before with `then` instead of `catch`. What is the `undefined` represent? The return prevents the `then ` from running in the happy path?

Does this block only run if `handleRunQuery` throws? or some other situation?

jgellin-sf: Yeah this is because of VSCode's goofy Thenable wrapper around standard Promise chains. For Thenable the `then` function's first argument is the `resolve` callback (here it's `undefined` because there's no more work to do) and the second argument is the `reject` callback.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -181,11 +182,23 @@ export class SOQLEditorInstance {
...
-          channelService.appendLine(
-            `An error occurred while running the SOQL query.`
-          );
-        });
+        vscode.window
+          .withProgress(
+            {
+              cancellable: false,
+              location: vscode.ProgressLocation.Notification,
+              title: 'Running query'
```
lcampos, line 190: Add this to `src/messages/i18n.ts`

cc @ekapner

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -181,11 +182,23 @@ export class SOQLEditorInstance {
...
+            },
+            () => {
+              return this.handleRunQuery();
+            }
+          )
+          .then(undefined, () => {
+            channelService.appendLine(
+              `An error occurred while running the SOQL query.`
```
lcampos, line 198: Same, let's add this to `src/messages/i18n.ts`.
cc @ekapner

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -201,13 +214,21 @@ export class SOQLEditorInstance {
       const message = `No default org found. Set a default org to use SOQL Builder. Run "SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org" to set one.`;
```
lcampos, line 214: Same, move this message to `src/messages/i18n.ts`

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dehru: I've got these error messages defined in i18n with another PR.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dehru: I've got this one covered here: https://github.com/forcedotcom/salesforcedx-vscode/pull/2810

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
dehru: @jgellin-sf - I do not have this one in my other PR.  https://github.com/forcedotcom/salesforcedx-vscode/pull/2810

jgellin-sf: Thanks. I covered it now in the `chore(soql): i18n for new message` commit.

dehru:
I'm finding this Running notification goes away really quickly.  Before i can read it.  Is it supposed to stay up for the duration of the query running?  


https://user-images.githubusercontent.com/599418/102811974-82d97c00-4383-11eb-96c9-4732fb215664.mov

jgellin-sf:
@dehru Yes it should stay running until the query has returned data (or has failed to return data). I think this may be because of some of the async nature of `handleRunQuery` and such. They are asynchronous but not declared `async`, so they can't be `await`ed. So I think the notification goes away as soon as the un-awaited call to `handleRunQuery` completes.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
```diff
@@ -194,20 +206,27 @@ export class SOQLEditorInstance {
...
 
     const queryText = this.document.getText();
-    return withSFConnection(async conn => {
-      const queryData = await new QueryRunner(conn).runQuery(queryText);
-      this.openQueryDataView(queryData);
+    const conn = await workspaceContext.getConnection();
+    const queryData = await new QueryRunner(conn).runQuery(queryText);
+    this.openQueryDataView(queryData);
+    this.runQueryDone();
+  }
```
jgellin-sf, line 225: @dehru It turns out that VSCode's `withProgress` can handle a single Promise chain, but nested callbacks mess it up. To fix the notification progress bar disappearing I had to unwind the call to `withSFConnection`.

jgellin-sf: I think this one is ready for another QA attempt.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorInstance.ts
lcampos: @jgellin-sf @dehru let's not keep doing this thing of breaking up work so that we have follow up PRs exclusively for cleaning up previously unfinished work. Let's update this PR to use the labels it needs in the pattern we already have (i18n.ts file).

lcampos (changes requested):
Need to add the labels to the i18n.ts file

dehru (approved):
The fix worked for me fine.  It correctly displayed the progress message until the results returned.

jgellin-sf:
@dehru Will you give the QA another spin on this one after recent merges?

@lcampos The updates you requested are committed.

lcampos (approved):

PR #2825, opened 2020-12-21
Add toggle icon to switch between builder editor and text editor

dehru:
### What does this PR do?
This PR adds a contribution point to the soql builder so that when a user has a soql file open, they can toggle between the soql builder ui and the text editor.  This is loosely based on the way preferences have a toggle between the preference editor and the json file that drives it.

@W-8296859@

### Functionality Before
No button

### Functionality After
https://user-images.githubusercontent.com/599418/102729380-c84b6a00-42ed-11eb-97ed-d36a764a597e.mov


packages/salesforcedx-vscode-soql/package.nls.json
```diff
@@ -1,4 +1,5 @@
 {
   "soqlCustom_soql": "SOQL Builder",
-  "soql_builder_open_new": "SFDX: Create Query in SOQL Builder"
-}
\ No newline at end of file
+  "soql_builder_open_new": "SFDX: Create Query in SOQL Builder",
+  "soql_builder_toggle": "Toggle SOQL Builder"
```
dehru, line 4: @ekapner - This is a screen shot of where this shows when the user hovers over the icon for a moment.

Screen Shot 2020-12-20 at 6 07 01 PM

jgellin-sf (approved):
Looks good.

praksb (approved):

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
```diff
@@ -0,0 +1,28 @@
...
+  OPEN_WITH_COMMAND
+} from '../constants';
+import { telemetryService } from '../telemetry';
+
+export async function soqlBuilderToggle(doc: vscode.Uri): Promise<void> {
+  telemetryService.sendCommandEvent('soql_builder_toggle', process.hrtime());
+
+  const viewType = vscode.window.activeTextEditor
+    ? BUILDER_VIEW_TYPE
+    : EDITOR_VIEW_TYPE;
```
jonnyhork, line 21: It's not clear to me how this works for us, is `vscode.window.activeTextEditor` saying, 
"if the active editor is of type `text editor` then return true." ?

Is the there a `vscode.window.activeCustomEditor`?

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -127,6 +127,11 @@
       {
         "command": "soql.builder.open.new",
         "title": "%soql_builder_open_new%"
+      },
+      {
+        "command": "soql.builder.toggle",
+        "title": "%soql_builder_toggle%",
+        "icon": "$(go-to-file)"
```
jonnyhork, line 134: How did you find this icon? Does VS Code give us access to all the default icons they use?

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -94,8 +94,8 @@
   "contributes": {
     "customEditors": [
       {
-        "viewType": "%soqlCustom_soql%",
-        "displayName": "SOQL Builder",
+        "viewType": "soqlCustom.soql",
+        "displayName": "%soqlCustom_soql%",
```
jonnyhork, line 98: Are the `%` symbols for i18n?

jonnyhork (approved):

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
dehru: Basically, vscode.window.activeTextEditor is either undefined or an object.  It's an object if the soql source is displayed in the text editor.  It's undefined if the custom editor is being displayed.  So I'm basically setting viewType to either our custom editor viewType `soqlCustom.soql` or `default` ( which is always the text editor ).

There is no such thing as a `vscode.window.activeCustomEditor`...sure wish they had one.

packages/salesforcedx-vscode-soql/package.json
dehru: Yes, i found it here:  https://code.visualstudio.com/api/references/icons-in-labels

Apparently, they process this with something kinda like handebars and look for icons.  They don't specifically mention it, but based on some stack overflow questions and github issues, I was able to piece this implementation together.  And it worked!

packages/salesforcedx-vscode-soql/package.json
dehru: Yeah, these look up the values in package-i18n.json

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
jonnyhork: > viewType to either our custom editor viewType soqlCustom.soql or default ( which is always the text editor ).

@dehru @praksb The text editor is the default view for a `.soql` file because we have `"priority": "option"` in the package.json. However, the user does have the option to change that in the settings. If they due then I think `EDITOR_VIEW_TYPE` might also be `soqlCustom.soql`.

Did you text the case here when the "default" is no longer the text editor?

jonnyhork: image

jonnyhork: I'm not sure if the `default` value used in the ternary points to whatever the default this setting refers too or `default` is just the VS Code default editor as if there is no alternative.

packages/salesforcedx-vscode-soql/package.json
jonnyhork: thats pretty neat!

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
jonnyhork: I was following up on the issue with tabs I made in the VS Code repo and the MS liked it to a [duplicate](https://github.com/microsoft/vscode/issues/111474) that might be useful here. Did you see the command `View: Toggle editor type` in the command palette? Maybe that will handle the view type for us regardless of any settings and defaults.

praksb: @jonnyhork 
I just now tested the case by switching the SOQL Builder view to be default and it works fine. 
I also tried  using View: Toggle editor type in the command palette and it looks like it should work but currently when I use it I am running into the VS Code bug that you mentioned above.

praksb (approved):
@dehru I QA'ed this PR and it looks good to me.

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
jonnyhork: Thanks for testing it, I'm glad it works as is!

jonnyhork (approved):

packages/salesforcedx-vscode-soql/package.nls.json
ekapner: @dehru - can we say "Switch to SOQL Builder" or "Open in SOQL Builder" ?

lcampos (approved):

PR #2827, opened 2020-12-21
bug: fix bug causing soql builder does not show

dehru:
### What does this PR do?
Looks like we accidentally localized the viewType property of the custom editor.  Which has caused a misconfiguration of the custom editor and it does not render.

### Functionality Before
Builder does not render

### Functionality After
Builder renders


jgellin-sf (approved):

lcampos (changes requested):
Please add a test to make sure we can catch this type of regression in the future.

packages/salesforcedx-vscode-soql/test/vscode-integration/commands/soqlFileCreate.test.ts
```diff
@@ -32,6 +35,6 @@ describe('soqlOpenNew should', () => {
 
     expect(telemetryStub.called).is.true;
     expect(editorOpened.called).is.true;
+    expect(executeCommandSpy.getCall(0).args[2]).contains(EDITOR_VIEW_TYPE);
```
dehru, line 38: @lcampos - this is the simplist way i could verify that the command is executed with the correct viewType, which was what caused this bug.  

I'm wondering in the future about more functional tests, to verify the soql builder custom editor's html/js behavior within vscode.  Something more selenium / playwright.  I noticed that there is an existing `packages/system-tests`, which seems like a functional style test ( maybe similar to Selenium ) but uses `spectron`.

What's your opinion on those tests?

lcampos: We used to do that but `spectron` became a maintenance burden since it would break every time VSCode updated the Electron version they ship with. We stopped using it more than a year ago because there was not a compatible version for the VSCode versions we were in. We ended up replacing those tests with `unit` or `vscode-integration` tests in their specific extensions. 

That module, `system-tests`, is still used as a way to setup a lot of the `vscode-integration` tests we have. In those tests we open projects with specific metadata and use it to test things like the results from a language server, here's an example of that https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lightning/test/vscode-integration/hover.test.ts.
You could try exploring if that could be a way for you to test a webview.

lcampos (approved):

PR #2828, opened 2020-12-21
feat: run apex tests using the library

AnanyaJha:
### What does this PR do?
This PR adds the ability to run apex tests from the sidebar, code lens, and command palette using the Apex library

@W-7561367@


packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -23,21 +25,81 @@ const SfdxCommandlet = sfdxCoreExports.SfdxCommandlet;
 const SfdxWorkspaceChecker = sfdxCoreExports.SfdxWorkspaceChecker;
 const SfdxCommandletExecutor = sfdxCoreExports.SfdxCommandletExecutor;
 const notificationService = sfdxCoreExports.notificationService;
+const LibraryCommandletExecutor = sfdxCoreExports.LibraryCommandletExecutor;
+const channelService = sfdxCoreExports.channelService;
```
AnanyaJha, line 29: Since the `LibraryCommandlet` executor currently lives in the core package, the channelService it uses is the default `Salesforce CLI` channel. If we switch to using the new `channelService` we have in the utils package right now, the experience is wonky because the `Starting command.../Ending command...` text appears in the default `Salesforce CLI` channel, while the output from the test run appears in the new channel that we've created. And since the command palette portion of the `test:run` functionality still lives in the core package as well, all the output from that command is displayed via the `Salesforce CLI` channel too

In a follow up PR, I think we should migrate the LibraryCommandlet to the utils package, the Apex commands (test:run command) to the apex extension, and then switch this command to use the new channelService so it doesn't cause customers any confusion on which channel they should be looking at for apex test results

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -22,22 +25,81 @@ const sfdxCoreSettings = sfdxCoreExports.sfdxCoreSettings;
...
+      this.codeCoverage
+    );
+    const humanOutput = new HumanReporter().format(result, this.codeCoverage);
+    channelService.appendLine(humanOutput);
+    return true;
+  }
+}
 
 // build force:apex:test:run w/ given test class or test method
 export class ForceApexTestRunCodeActionExecutor extends SfdxCommandletExecutor<{}> {
```
AnanyaJha, line 90: reworked this executor a bit so that it could be used for the sidebar as well

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -224,6 +225,36 @@ export class ApexTestOutlineProvider
...
   }
+
+  private updateTestsFromLibrary(testResult: TestResult) {
+    const groups = new Set<ApexTestGroupNode>();
+    for (const test of testResult.tests) {
+      const apexGroupName = test.apexClass.fullName;
+      const apexGroupNode = this.apexTestMap.get(
+        apexGroupName
+      ) as ApexTestGroupNode;
+      // why u add the group node if it does exist?
```
AnanyaJha, line 236:
```suggestion
```

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "@salesforce/apex-tmlanguage": "1.4.0",
+    "@salesforce/apex-node": "0.1.9",
```
AnanyaJha, line 28: I'd like to wait until the new version of the Apex library is released so we can take advantage of the auth issue fix - so this version number will be updated

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
lcampos: Yes, that is part of the work we'll be doing in January to address some perf issues in the Apex extension.

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -4,12 +4,15 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
+import { HumanReporter, TestItem, TestService } from '@salesforce/apex-node';
+import { TestLevel } from '@salesforce/apex-node/lib/src/tests/types';
```
lcampos, line 8: Let's make sure we surface this type at the top of the library so we can just import it by using `@salesforce/apex-node`.

brpowell (approved):
Ran through the various avenues for running tests with and without code coverage, library setting on and off:

- test sidebar
- codelens
- command palette

Looks good

PR #2837, opened 2021-01-04
Remove core extension as dependency on SOQL extension part 3

jgellin-sf:
### What does this PR do?
This PR completes the removal of the core extension as a dependency on the SOQL extension. It is the SOQL Extension-side work that builds on the previous PRs ([2743](https://github.com/forcedotcom/salesforcedx-vscode/pull/2743) and [2772](https://github.com/forcedotcom/salesforcedx-vscode/pull/2772)).

@W-8295301@

### Functionality Before
N/A

### Functionality After
N/A


packages/salesforcedx-vscode-soql/src/sfdx.ts
```diff
@@ -7,6 +7,7 @@
 
 import { Connection } from '@salesforce/core';
 import { ChannelService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
+import { WorkspaceContextUtil } from '@salesforce/salesforcedx-utils-vscode/out/src/context';
 import { DescribeSObjectResult } from 'jsforce';
 import * as vscode from 'vscode';
```
lcampos, line 12: This import can now be removed

jgellin-sf: Fixed in `fix(soql): remove unused import` commit.

packages/salesforcedx-vscode-soql/src/index.ts
```diff
@@ -5,21 +5,21 @@
...
-import { checkDependencies } from './sfdx';
 
 export async function activate(
   context: vscode.ExtensionContext
 ): Promise<void> {
   const extensionHRStart = process.hrtime();
-  checkDependencies();
   context.subscriptions.push(SOQLEditorProvider.register(context));
   QueryDataViewService.register(context);
+  WorkspaceContextUtil.getInstance().initialize(context);
```
lcampos, line 22: I think we should add an `await` here. If we don't then we run the risk of initializing the language server with an empty connection.

lcampos (approved):

packages/salesforcedx-vscode-soql/src/index.ts
jgellin-sf: Good idea. I made the change in `feat(soql): await workspace context in activation`.

dehru:
This passes QA for me.

PR #2839, opened 2021-01-04
SOQL code-completion on WHERE clause expressions

dobladez:
### What does this PR do?

Code completion of WHERE clause expressions
    
  * Complete operators based on type of field
  * Complete literal values based on type of field
  * Show type name (and icon) on completion items for fields
  * Offer NULL as a completion value only if field is nillable
    
Also: refactoring of "placeholder" completion items: receive JSON data instead
of parseable values inside item label string.

@W-8597428@

Related LSP-server PR: https://github.com/forcedotcom/soql-tooling/pull/112

### Functionality Before
n/a

### Functionality After
https://user-images.githubusercontent.com/152839/103246284-b139f700-4941-11eb-9228-2a56fc06dd87.mov

https://user-images.githubusercontent.com/152839/103816838-60fb1d00-5044-11eb-9c42-8336da6bb86b.mov

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
+++ b/packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
@@ -7,10 +7,12 @@
 
 import { CompletionItemKind, SnippetString } from 'vscode';
 import ProtocolCompletionItem from 'vscode-languageclient/lib/protocolCompletionItem';
-import { retrieveSObject, retrieveSObjects } from '../sfdx';
+import { retrieveSObject, retrieveSObjects, channelService } from '../sfdx';
 
 import { Middleware } from 'vscode-languageclient';
 import { telemetryService } from '../telemetry';
+import { nls } from '../messages';
+import { DescribeSObjectResult } from 'jsforce';
 
 const EXPANDABLE_ITEM_PATTERN = /__([A-Z_]+)/;
 
@@ -48,14 +50,21 @@ async function filterByContext(
       !EXPANDABLE_ITEM_PATTERN.test(item.label) &&
       item?.data?.soqlContext?.sobjectName
     ) {
-      const objMetadata = await retrieveSObject(
+      const objMetadata = await safeRetrieveSObjectMetadata(
         item.data.soqlContext.sobjectName
       );
-      const fieldMeta = objMetadata.fields.find(
-        field => field.name === item.data.soqlContext.fieldName
-      );
-      if (fieldMeta && !operatorsByType[fieldMeta.type].includes(item.label)) {
+      if (!objMetadata) {
         continue;
+      } else {
+        const fieldMeta = objMetadata.fields.find(
+          field => field.name === item.data.soqlContext.fieldName
+        );
+        if (
+          fieldMeta &&
+          !operatorsByType[fieldMeta.type].includes(item.label)
+        ) {
+          continue;
+        }
       }
     }
 
@@ -103,7 +112,7 @@ const expandFunctions: {
     soqlContext?: SoqlItemContext
   ): Promise<ProtocolCompletionItem[]> => {
     try {
-      const sobjectItems = (await retrieveSObjects()).map(objName => {
+      const sobjectItems = (await safeRetrieveSObjectsList()).map(objName => {
         const item = new ProtocolCompletionItem(objName);
         item.kind = CompletionItemKind.Class;
         return item;
@@ -126,34 +135,35 @@ const expandFunctions: {
       return [];
     }
 
-    try {
-      const objMetadata = await retrieveSObject(soqlContext.sobjectName);
-      const sobjectFields = objMetadata.fields.reduce((fieldItems, field) => {
-        fieldItems.push(
-          newCompletionItem(
-            field.name,
-            field.name,
-            CompletionItemKind.Field,
-            field.type
-          )
-        );
-        if (field.relationshipName) {
-          fieldItems.push(
-            newCompletionItem(
-              `${field.relationshipName}`,
-              field.relationshipName + '.',
-              CompletionItemKind.Class,
-              'Ref. to ' + field.referenceTo
-            )
-          );
-        }
-
-        return fieldItems;
-      }, [] as ProtocolCompletionItem[]);
-      return sobjectFields;
-    } catch (metadataError) {
+    const objMetadata = await safeRetrieveSObjectMetadata(
+      soqlContext.sobjectName
+    );
+    if (!objMetadata) {
       return [];
     }
+
+    const sobjectFields = objMetadata.fields.reduce((fieldItems, field) => {
+      fieldItems.push(
+        newCompletionItem(
+          field.name,
+          field.name,
+          CompletionItemKind.Field,
+          field.type
+        )
+      );
+      if (field.relationshipName) {
+        fieldItems.push(
+          newCompletionItem(
+            `${field.relationshipName}`,
+            field.relationshipName + '.',
+            CompletionItemKind.Class,
+            'Ref. to ' + field.referenceTo
+          )
+        );
+      }
+      return fieldItems;
+    }, [] as ProtocolCompletionItem[]);
+    return sobjectFields;
   },
   LITERAL_VALUES_FOR_FIELD: async (
     soqlContext?: SoqlItemContext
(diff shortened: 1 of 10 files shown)


packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,317 @@
...
+  const expandedItems = [...items];
+
+  for (const [index, item] of items.entries()) {
+    const parsedCommand = item.label.match(EXPANDABLE_ITEM_PATTERN);
+    if (parsedCommand) {
+      const commandName = parsedCommand[1];
+
+      const handler = expandFunctions[commandName];
+      if (!handler) {
+        console.log(`Unknown SOQL completion command ${commandName}!`);
```
lcampos, line 79: Not sure if this needs to be surfaced to the customer but we should at least track it with the telemetry service. Same for the rest of the `console.log`s in the rest of this file.

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,317 @@
...
+  ): Promise<ProtocolCompletionItem[]> => {
+    if (!soqlContext?.sobjectName) {
+      console.log('SOBJECT_FIELDS_PLACEHOLDER missing `sobjectName`!');
+      return [];
+    }
+
+    try {
+      const objMetadata = await retrieveSObject(soqlContext.sobjectName);
+      const sobjectFields = objMetadata.fields.reduce((fieldItems, field) => {
+        const icon = iconsByFieldType[field.type];
```
lcampos, line 125: Not sure I understand the purpose of these icons. Is it just for decoration or is there a functional purpose for using them ?

packages/salesforcedx-vscode-soql/src/client/queryValidation.ts
```diff
@@ -0,0 +1,42 @@
...
+  client.onRequest(RequestTypes.RunQuery, async (queryText: string) => {
+    try {
+      return await withSFConnection(async conn => {
+        const queryData = await new QueryRunner(conn).runQuery(queryText, {
+          showErrors: false
+        });
+        return { result: queryData };
+      });
+    } catch (e) {
+      // NOTE: The return value must be serializable, for JSON-RPC.
```
lcampos, line 29: Should we be logging (telemetry) this too or is there a place where this eventually gets logged already ?

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  return expandedItems;
+}
+
+const expandFunctions: {
+  [key: string]: (
+    soqlContext?: SoqlItemContext
+  ) => Promise<ProtocolCompletionItem[]>;
+} = {
+  SOBJECTS_PLACEHOLDER: async (
```
praksb, line 98: expandFunctions is now doing lot more things that I originally thought it would do (which is just expand sobjects and its fields). 

I feel all this logic should move over to the language server instead of the client. I think we just go ahead and implement what we discussed in the last meeting to have the language server take in something like an sObjectProvider and move all the code here to the language server and keep the client simple.

praksb: Also, SOQL completion in Apex would rely on the SOQL language server and we cannot have this logic duplicated over there as well in the Apex extension.

praksb (changes requested):
Let's talk during our dev hangout. I feel that the placeholder expansion logic is not simple anymore and it needs to be switch over to the language server.

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: Good point. I'll add telemetry

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: It shows the user the type of each field.  I personally don't like the icons, but do find the information useful. It could be text instead.

In addition to the icons (which are currently displayed on all items of the list), I've included the textual name of the type to the right of the selected item only. You can see it on the videos above, but here's a screenshot example:
image

packages/salesforcedx-vscode-soql/src/client/queryValidation.ts
dobladez: That particular error is displayed to the user.
The comment is there to avoid the temptation of replacing that return call with `return {error: e}`, which seems natural but fails due to `e` not guaranteed to be JSON-serializable.

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: Agreed.  We have to define how to do it though.... it's not trivial and we need to agree on an approach.

BTW: as an intermediate solution, some of this logic can be moved to the server-side. For example, for this new logic for type-aware fields, the server could send all completion items with extra metadata indicating for which types the item applies... so the client would only (and optionally) filter them based on the field type. Still not a solution, but wanted to mentioning it.

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
brpowell: I've never seen a language server use emojis and it could be confusing to decode for users (as much as i love using emojis 🎉 ). I think we should just stick with displaying the type via the CompletionItem API like you have already.

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,320 @@
...
+    try {
+      const sobjectItems = (await retrieveSObjects()).map(objName => {
+        const item = new ProtocolCompletionItem(objName);
+        item.kind = CompletionItemKind.Class;
+        return item;
+      });
+
+      return sobjectItems;
+    } catch (metadataErrors) {
+      return [];
```
brpowell, line 114: If `retrieveSObjects()` throws an error, seems like something we might want to report

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,320 @@
...
+              'Ref. to ' + field.referenceTo
+            )
+          );
+        }
+
+        return fieldItems;
+      }, [] as ProtocolCompletionItem[]);
+      return sobjectFields;
+    } catch (metadataError) {
+      return [];
```
brpowell, line 157: Similar thing here, how do we get more insight into what the thrown issue is?

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
lcampos: @dobladez let's get rid of them, as @brpowell pointed out this is not how the rest of the language servers display information and we want to be consistent with the rest since this will eventually be used in Apex code.

dobladez: done

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: Done

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: Thanks. Done.

dobladez:
Ready to re-review:
* Removed type icons
* Added telemetry
* Added better error handling when fetching objects metadata

In the future, we will move all logic to the LSP server-side. First we need to define an approach in coordination with the other modules and extensions that fetch and cache org data.

praksb (approved):

packages/salesforcedx-vscode-soql/test/vscode-integration/client/completion.test.ts
```diff
@@ -5,18 +5,23 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { expect } from 'chai';
+import { Connection } from 'jsforce';
```
AnanyaJha, line 8: Do we need to use the 'Connection' from jsforce specifically? We typically access the 'Connection' object from `@salesforce/core`

dobladez: Changed. Thanks!

AnanyaJha (approved):

PR #2844, opened 2021-01-06
Prevent false positives in builds when test results aren't generated

brpowell:
### What does this PR do?

The autobuilds aggregate all of the generated test results across the packages and then test that the merged result output does not have any failures. However, If an unexpected test runner failure occurs during a run where results for a particular test category and package are not generated, the builds will not fail. This is because we aren't asserting that results were generated, just that the final output doesn't have any failures.

This change enhances the `aggregate-junit-xml.js` script to assert that if a package has tests for the specified test categories, we should expect output to be generated for it. If they don't, the build should fail.

@W-8190289@

### Functionality Before

Builds falsely pass when JUnit test results were not generated for a package.

### Functionality After

Builds fail when JUnit test results were not generated for a package.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,6 +303,7 @@ jobs:
           name: 'Aggregate test results'
           shell: bash.exe
           command: |
+            set -e
             npm run aggregateJUnit --unit
             junit-merge -d junit-aggregate -o junit-aggregate.xml
       - store_test_results:
@@ -351,6 +352,7 @@ jobs:
           name: 'Aggregate test results'
           shell: bash.exe
           command: |
+            set -e
             npm run aggregateJUnit --integration --vscode-integration
             junit-merge -d junit-aggregate -o junit-aggregate.xml
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,7 +304,7 @@ jobs:
           shell: bash.exe
           command: |
             set -e
-            npm run aggregateJUnit --unit
+            npm run aggregateJUnit unit
             junit-merge -d junit-aggregate -o junit-aggregate.xml
       - store_test_results:
           path: ./junit-aggregate/
@@ -353,7 +353,7 @@ jobs:
           shell: bash.exe
           command: |
             set -e
-            npm run aggregateJUnit --integration --vscode-integration
+            npm run aggregateJUnit integration vscode-integration
             junit-merge -d junit-aggregate -o junit-aggregate.xml
       - store_test_results:
           path: ./junit-aggregate/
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -351,7 +352,8 @@ jobs:
           name: 'Aggregate test results'
           shell: bash.exe
           command: |
-            npm run aggregateJUnit
+            set -e
+            npm run aggregateJUnit -- --integration --vscode-integration
```
brpowell, line 356: Specify the test categories we want to aggregate and assert the results for

packages/salesforcedx-apex-replay-debugger/package.json
```diff
@@ -41,7 +41,7 @@
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "test": "npm run test:unit && npm run test:integration",
     "test:unit": "node ./node_modules/nyc/bin/nyc.js ./node_modules/mocha/bin/_mocha --recursive out/test/unit --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaUnitTestsConfig.json",
-    "test:integration": "node ./node_modules/cross-env/dist/bin/cross-env.js VSCODE_NLS_CONFIG={} ./node_modules/nyc/bin/nyc.js ./node_modules/mocha/bin/_mocha --recursive out/test/integration config=../../config/mochaIntegrationTestsConfig.json"
+    "test:integration": "node ./node_modules/cross-env/dist/bin/cross-env.js VSCODE_NLS_CONFIG={} ./node_modules/nyc/bin/nyc.js ./node_modules/mocha/bin/_mocha --recursive out/test/integration --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaIntegrationTestsConfig.json"
```
brpowell, line 44: Funny enough the update script caught an issue where we haven't been running the apex-replay-debugger integration tests due to this misconfiguration.

.circleci/config.yml
```diff
@@ -303,7 +303,8 @@ jobs:
           name: 'Aggregate test results'
           shell: bash.exe
           command: |
-            npm run aggregateJUnit
+            set -e
```
lcampos, line 306: Do we need `set -e` if the new junit script does exit with an error ?

brpowell: Yeah, CircleCI doesn't fail the step unless we set this even if the script exits with a non-zero exit code. https://support.circleci.com/hc/en-us/articles/115015733328-Step-should-fail-but-job-finished-successfully

lcampos: Got it, thanks for clarifying

lcampos (approved):

AnanyaJha (approved):

PR #2847, opened 2021-01-06
fix(soql): disable remote query checks. Make it an experimental option

dobladez:
@W-8611581@

### Functionality Before
SOQL editor would always call the remote API to validate queries

### Functionality After
This "experimental" feature is now disabled by default. Added new configuration option for the user to enable it.

https://user-images.githubusercontent.com/152839/103817675-e501d480-5045-11eb-8f54-36ac25cec8b3.mov

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -595,29 +595,6 @@
       },
       "preLaunchTask": "Compile",
       "internalConsoleOptions": "openOnSessionStart"
-    },
-    {
-      "name": "Launch Tests - SOQL - VSCode Integration Tests",
-      "type": "extensionHost",
-      "request": "launch",
-      "runtimeExecutable": "${execPath}",
-      "args": [
-        "${workspaceFolder}/packages/system-tests/assets/sfdx-simple",
-        "--extensionDevelopmentPath=${workspaceFolder}/packages",
-        "--extensionTestsPath=${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration"
-      ],
-      "stopOnEntry": false,
-      "sourceMaps": true,
-      "smartStep": true,
-      "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
-      "sourceMapPathOverrides": {
-        "../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
-        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
-        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
-        "webpack:///*": "*"
-      },
-      "preLaunchTask": "Compile",
-      "internalConsoleOptions": "openOnSessionStart"
     }
   ],
   "compounds": [
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -143,6 +143,17 @@
           "group": "navigation"
         }
       ]
+    },
+    "configuration": {
+      "type": "object",
+      "title": "%configuration_title%",
+      "properties": {
+        "salesforcedx-vscode-soql.experimental.validateQueries": {
+          "type": "boolean",
+          "default": false,
+          "description": "%soql_validation_flag%"
+        }
+      }
     }
   }
 }
--- a/packages/salesforcedx-vscode-soql/package.nls.json
+++ b/packages/salesforcedx-vscode-soql/package.nls.json
@@ -1,5 +1,7 @@
 {
+  "configuration_title": "Salesforce SOQL Configuration",
   "soqlCustom_soql": "SOQL Builder",
   "soql_builder_open_new": "SFDX: Create Query in SOQL Builder",
-  "soql_builder_toggle": "Toggle SOQL Builder"
+  "soql_builder_toggle": "Toggle SOQL Builder",
+  "soql_validation_flag": "Use REST API to validate SOQL queries and to identify errors."
 }
--- a/packages/salesforcedx-vscode-soql/src/client/client.ts
+++ b/packages/salesforcedx-vscode-soql/src/client/client.ts
@@ -69,7 +69,6 @@ export async function startLanguageClient(
   // Start the client. This will also launch the server
   client.start();
   await client.onReady();
-
   client = queryValidation.afterStart(client);
 }
 
--- a/packages/salesforcedx-vscode-soql/src/client/client.ts
+++ b/packages/salesforcedx-vscode-soql/src/client/client.ts
@@ -4,13 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
-import {
-  QueryValidationFeature,
-  RequestTypes
-} from '@salesforce/soql-language-server';
 import * as path from 'path';
-import { CompletionItemKind, ExtensionContext, workspace } from 'vscode';
+import { ExtensionContext, workspace } from 'vscode';
+import * as queryValidation from './queryValidation';
+import * as codeCompletion from './codeCompletion';
 
 import {
   LanguageClient,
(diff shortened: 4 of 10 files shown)


packages/salesforcedx-vscode-soql/package.nls.json
```diff
@@ -1,4 +1,6 @@
 {
+  "configuration_title": "Salesforce SOQL Configuration",
   "soqlCustom_soql": "SOQL Builder",
-  "soql_builder_open_new": "SFDX: Create Query in SOQL Builder"
+  "soql_builder_open_new": "SFDX: Create Query in SOQL Builder",
+  "soql_editor_remote_checks_flag": "Validate SOQL editor queries against your Org. When enabled, the SOQL editor calls the remote API to validate the query on your default Org environment and provide better syntax checks."
```
ekapner, line 5:
```suggestion
  "soql_editor_remote_checks_flag": "Validate SOQL editor queries against your org. When enabled, the SOQL editor calls the remote API to validate the query on your default org and provide better syntax checks."
```

ekapner, line 5: @praksb @dobladez - do we need to say SOQL "editor" queries - or can we just say "SOQL queries"? Is this to distinguish a query built in a text editor vs. SOQL Builder?

ekapner: Also, couldn't figure out how to put my comment closer to the .mov. Can we all cap SOQL "Experimental: SOQL Editor Remote Checks" - but still want to make sure I understand the context for 'editor' here as well

dobladez: Good point. Yes, this flag only affects the behavior of the text editor... Should we use the full "SOQL text editor queries" ?

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -136,6 +136,17 @@
           "when": "sfdx:project_opened"
         }
       ]
+    },
+    "configuration": {
+      "type": "object",
+      "title": "%configuration_title%",
+      "properties": {
+        "salesforcedx-vscode-soql.experimental.soqlEditorRemoteChecks": {
```
praksb, line 144: can we just call it soqlValidation instead of soqlEditorRemoteChecks?

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -136,6 +136,17 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "%soql_editor_remote_checks_flag%"
```
praksb, line 147: soql_validation_flag instead?

packages/salesforcedx-vscode-soql/package.nls.json
praksb: Even though its applicable only to queries in the text editor, I don't think that distinction needs to be explicitly called out in the description text.

praksb: I would make this description a little more compact. Maybe something like "Enable SOQL query validation using the REST API and display errors".

packages/salesforcedx-vscode-soql/test/vscode-integration/client/client.test.ts
```diff
@@ -69,16 +61,57 @@ describe('SOQL language client', () => {
...
+  });
+
+  it('should not create diagnostics based off of remote query validation when disabled', async () => {
+    soqlFileUri = await writeSOQLFile(
+      'testSemanticErrors_remoteRunDefault.soql',
+      'SELECT Ids FROM Account'
+    );
+
+    stubSOQLExtensionConfiguration(sandbox, {
+      'experimental.soqlEditorRemoteChecks': true
```
praksb, line 89: I think this test needs to be fixed.  'experimental.soqlEditorRemoteChecks': true and the test is passing.

1) The test should be fixed so that the mock connection throws an error
2) experimental.soqlEditorRemoteChecks should be set to false

praksb (changes requested):
I just had a few small comments on the text description.

Also, Can you take a look at my comment on one of the tests?

packages/salesforcedx-vscode-soql/package.nls.json
ekapner: @praksb - If I'm understanding your point, are you saying that we should just use "SOQL queries"?

packages/salesforcedx-vscode-soql/test/vscode-integration/client/client.test.ts
dobladez: Great catch!!!  Fixed now. Very tricky:

Seems like the integration test framework runs a failed test multiple times.  And in the case of language diagnostics, the second time it run for the exact same file and input, no diagnostics check happens (I guess this is an optimization on VSCode. That's also how it haves when testing manually).

Here's what happened:
1. Assertion `expect(querySpy.notCalled).to.be.true` failed
2. Test re-run
3. VSCode didn't run the diagnostic check again (optimization)
4. Thus, assertion `expect(querySpy.notCalled).to.be.true` and the rest passed because no validation happened

Thanks!

packages/salesforcedx-vscode-soql/package.nls.json
praksb: @ekapner yes. i am suggesting using just soql queries.

praksb (commented):
Test fix looks good. Let me know what you think about the changes to the description text. Then we can move this to review and try to push it through by Monday.

packages/salesforcedx-vscode-soql/package.nls.json
ekapner: I don't think we need to mention better syntax checks. Is the fact that it's REST API important? Validate SOQL queries and identify errors. (?)

praksb: The reason I want to put "REST API" is that it will tell the customers that the checks happens on the server side. That way they would (hopefully) know that this validation wouldn't work if they are offline or having an auth issue and/or not connected to an org.

dobladez:
@praksb  @ekapner Is the text below what we agreed on? Thanks!

image

dobladez:
Final ;-)

image

.vscode/launch.json
```diff
@@ -595,29 +595,6 @@
...
-      "smartStep": true,
-      "outFiles": ["${workspaceFolder}/packages/*/out/**/*.js"],
-      "sourceMapPathOverrides": {
-        "../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
-        "../../../../test/vscode-integration": "${workspaceFolder}/packages/salesforcedx-vscode-soql/out/test/vscode-integration/*",
-        "webpack:///salesforcedx-utils-vscode/./*": "${workspaceFolder}/packages/salesforcedx-utils-vscode/*",
-        "webpack:///*": "*"
-      },
-      "preLaunchTask": "Compile",
-      "internalConsoleOptions": "openOnSessionStart"
```
praksb, line 620: Did you remove this by mistake?

packages/salesforcedx-vscode-soql/src/client/client.ts
```diff
@@ -60,28 +53,7 @@ export async function startLanguageClient(
...
-        const items = (await next(
-          document,
-          position,
-          context,
-          token
-        )) as ProtocolCompletionItem[];
-
-        return expandPlaceholders(items);
-      }
-    }
```
praksb, line 84: this shouldn't be part of this PR right?

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
```diff
@@ -0,0 +1,111 @@
...
+        return fieldItems;
+      }, [] as ProtocolCompletionItem[]);
+      return sobjectFields;
+    } catch (metadataError) {
+      return [];
+    }
+  }
+};
+
+function newFieldCompletionItem(
```
praksb, line 102: I think these changes should be part of your completion PR.

.vscode/launch.json
dobladez: No mistake. It was duplicated.

packages/salesforcedx-vscode-soql/src/client/client.ts
dobladez: It is... I've separated the completion vs. query validation into different files. `client.ts` was getting too big and with many responsibilities (initialization, completion, query validation)

packages/salesforcedx-vscode-soql/src/client/codeCompletion.ts
dobladez: It's just code moved from client.ts into a separate file. See commit dc95f508c5f116b5b39e3d08251b62ab4a8a33cb

praksb (approved):

jag-sfdc (approved):

PR #2864, opened 2021-01-14
refactor: copy library commandlet to utils module

brpowell:
### What does this PR do?

Simply copies the library commandlet to the utils module. Commands have not been changed to use this version yet because there are other things that need to be updated first. E.g. the utils version of the telemetry service needs to be used in the extensions packages first, otherwise telemetry will stop working.

@W-8659966@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -5,7 +5,6 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import * as vscode from 'vscode';
-import { notificationService, ProgressNotification } from './index';
 import { CliCommandExecutor, Command, CommandExecution } from '../cli';
 import {
   Measurements,
@@ -18,6 +17,7 @@ import { nls } from '../messages';
 import { ContinueResponse } from '../types';
 import { getRootWorkspacePath } from '../workspaces';
 import { ChannelService } from './channelService';
+import { notificationService, ProgressNotification } from './index';
 
 export interface FlagParameter<T> {
   flag: T;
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -0,0 +1,178 @@
+/*
```
brpowell, line 1: Consolidated the executors to this file.

xyc (approved):

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as vscode from 'vscode';
+import { notificationService, ProgressNotification } from '.';
```
xyc, line 8: let's use `from './index'` here for consistency

brpowell, line 8:
```suggestion
import { notificationService, ProgressNotification } from './index';
```

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
AnanyaJha: Can we move the tests for the `LibraryCommandletExecutor` over as well?

brpowell: No because they are integration tests. I've been trying to convert them into unit tests like some of the others in this package but the telemetry service dependency is making it very difficult. The telemetry service tests themselves are skipped in this package. We need to have a larger discussion about that module because it can cause a lot of problems with these refactors

PR #2867, opened 2021-01-15
fix: do not show toggle command in palette

dehru:
### What does this PR do?

The command to toggle between Soql Builder and Soql Editor has snuck into the Command Palette, even though it is not explicitly mentioned that it contributes to the command palette.  Fortunately, it's an easy fix.

@W-8728692@

### Functionality Before
Screen Shot 2021-01-15 at 12 42 48 PM

### Functionality After
Screen Shot 2021-01-15 at 12 41 05 PM

### DIFF
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -134,6 +134,10 @@
         {
           "command": "soql.builder.open.new",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "soql.builder.toggle",
+          "when": "false"
         }
       ],
       "editor/title": [


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -134,6 +134,10 @@
         {
           "command": "soql.builder.open.new",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "soql.builder.toggle",
+          "when": "false"
```
dehru, line 140: @lcampos - Is there any way to write a test for a change like this?

jgellin-sf (approved):

praksb (approved):

packages/salesforcedx-vscode-soql/package.json
brpowell: I think you'd have to create a system test with spectron but we've disabled those tests for the time being.

brpowell (approved):

packages/salesforcedx-vscode-soql/package.json
dehru: OK. Thanks that's what i was thinking.. And i understand that is flappy.

PR #2870, opened 2021-01-18
Adds OpenJDK path on Linux

renatoliveira:
This PR adds the snippet for the Java path on Linux.

I specifically added the name of the Linux distribution because the path can differ between distributions, but Pop! OS is pretty popular these days (almost like Ubuntu, according to DistroWatch). I also specified that the path is for installations using Debian's apt package manager because Java can be downloaded to another folder.


lcampos (approved):

lcampos:
Thanks for your contribution @renatoliveira, the change will go live as part of the next release of the extensions.

PR #2878, opened 2021-01-20
refactor: use telemetry service from utils package in core extension

brpowell:
### What does this PR do?

Updates core to use the telemetry service in the utils package.

@W-8720190@


packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -233,4 +266,38 @@ export class TelemetryService {
...
+   *
+   * @param callback function to call if telemetry is enabled
+   */
+  private validateTelemetry(callback: () => void): void {
+    if (this.reporter !== undefined) {
+      this.isTelemetryEnabled()
+        .then(enabled => Promise.resolve(enabled ? callback() : undefined))
+        .catch(err => console.error(err));
+    }
+  }
```
brpowell, line 302: In order to conform to the "no unhandled promise rejection" rule, i could either have A) updated the dozens of places in core that call telemetry methods with an empty `.catch()`, or B) change the telemetry methods to not return promises. I can't think of a reason why we'd want to wait for a telemetry method to resolve, so I decided to go with B and handle all the calls using this method. Let me know if you see an issue with doing it this way.

AnanyaJha: This sounds good to me - you might need to update again bc of the changes I merged for the Apex package yesterday :/

packages/salesforcedx-vscode-core/src/telemetry/index.ts
```diff
@@ -4,6 +4,12 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { TelemetryService } from './telemetry';
-export { TelemetryBuilder, TelemetryData, Properties, Measurements } from './telemetry';
+// import { TelemetryService } from './telemetry';
```
AnanyaJha, line 7:
```suggestion
```

AnanyaJha (approved):
looks good other than a small comment

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
xyc: Don't we need to update the current places using the async methods (`await telemetryService.method`) using this approach?

brpowell: Using await shouldn't adversely affect it but it doesn't make sense to have it anymore, I'll update those places

packages/salesforcedx-utils-vscode/test/unit/telemetry/telemetry.test.ts
```diff
@@ -279,14 +273,39 @@ describe('Telemetry production mode', () => {
...
-      0,
-      123
-    ]);
+    telemetryService.sendCommandEvent('create_apex_class_command', [0, 123]);
     assert.notCalled(reporter);
     expect(teleStub.firstCall.args).to.eql([false]);
   });
 
+  xit('Should show telemetry info message', async () => {
+    // create vscode extensionContext in which telemetry msg has never been previously shown
```
AnanyaJha, line 282: do these tests need to be skipped?

brpowell: They have the same issue as the tests above them https://github.com/forcedotcom/salesforcedx-vscode/pull/2655/files#r510284491

AnanyaJha: Ah okay gotcha, maybe we can leave comment with that info

AnanyaJha (approved):

xyc (approved):

PR #2886, opened 2021-01-22
refactor: use utils package library commandlet

brpowell:
### What does this PR do?

Updates the library based commands in core to use the library commandlet in the utils package.

@W-8749976@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -114,20 +114,16 @@ export abstract class SfdxCommandletExecutor<T>
 
 export abstract class LibraryCommandletExecutor<T>
   implements CommandletExecutor<T> {
-  /**
-   * Command name visible to user while executing.
-   */
   private readonly executionName: string;
-  /**
-   * Command name for logging purposes such as telemetry
-   */
   private readonly logName: string;
-  /**
-   * Output channel to report execution status to.
-   */
   private readonly outputChannel: vscode.OutputChannel;
   protected readonly telemetry = new TelemetryBuilder();
 
+  /**
+   * @param name Name visible to user while executing.
+   * @param logName Name for logging purposes such as telemetry.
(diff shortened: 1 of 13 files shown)


packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -114,27 +114,33 @@ export abstract class SfdxCommandletExecutor<T>
...
-  protected abstract readonly outputChannel: vscode.OutputChannel;
-  protected readonly telemetry = new TelemetryBuilder();
+  constructor(
+    executionName: string,
+    logName: string,
+    outputChannel: vscode.OutputChannel
+  ) {
+    this.executionName = executionName;
+    this.logName = logName;
+    this.outputChannel = outputChannel;
```
brpowell, line 134: Now provide the vs code output channel to report execution status with, since extensions could be using their own channels with the commandlet.

lcampos: Why did we choose to provide the OutputChannel instead of just the name of the channel ? We could do `ChannelService.getInstance(<provided channel name>)` instead of doing `new ChannelService(this.outputChannel)`

What happens if we try to create an output channel has already been created ?

brpowell: If it's done by name with `getInstance`, we run into the possibility of creating multiple channels with the same label. I was curious about that as well and tried it out. We need to pass the same channel object to commands in order for them to send output to the same window. An implication of this is as long as we want commands across different extensions to do that, the core extension must export it and the others have to add core as a dependency.

AnanyaJha:
Double checking that this PR only covers updating the library based commands in _Core_ to use the utils package library commandlet. If yes, then do we want to create a separate work item to make the same change for the library based commands in the Apex extension & also remove the librarycommandlet from the Core package ?

brpowell:
@AnanyaJha yes to all of those things :)

AnanyaJha (approved):

lcampos (approved):

PR #2894, opened 2021-01-26
fix: show telemetry message extension source

brpowell:
### What does this PR do?

Fixes a small issue where the telemetry message prompt shows a null extension source.

@W-8792554@

### Functionality Before

NullExtensionDescription

### Functionality After

Screen Shot 2021-01-26 at 10 01 10 AM


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -513,13 +515,40 @@ async function setupOrgBrowser(
...
+        if (selection && selection === showButtonText) {
+          vscode.commands.executeCommand(
+            'vscode.open',
+            vscode.Uri.parse(TELEMETRY_OPT_OUT_LINK)
+          );
+        }
+      });
+    context.globalState.update(TELEMETRY_GLOBAL_VALUE, true);
+  }
+}
```
brpowell, line 542: Calling the showInformationMessage api in the utils package is what caused vscode to be confused about what extension was showing the message, so I moved that logic to the core extension. Lmk if you think this function belongs somewhere else in the src.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -513,13 +515,40 @@ async function setupOrgBrowser(
...
+
 export async function activate(context: vscode.ExtensionContext) {
   const extensionHRStart = process.hrtime();
   const { name, aiKey, version } = require(context.asAbsolutePath(
     './package.json'
   ));
   await telemetryService.initializeService(context, name, aiKey, version);
-  telemetryService.showTelemetryMessage();
+  showTelemetryMessage(context);
+  // telemetryService.showTelemetryMessage();
```
brpowell, line 551:
```suggestion
```

packages/salesforcedx-vscode-core/src/index.ts
xyc: Looks good - Just a reminder that this won't show a "null" extension in production, only in dev.

brpowell: 😞

lcampos (approved):

PR #2905, opened 2021-01-29
refactor: migrate execute anonymous commands to apex package

brpowell:
### What does this PR do?

Moves the execute anonymous Apex commands to the apex package. 

@W-8660053@


packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
```diff
@@ -0,0 +1,232 @@
...
+import { nls } from '../messages';
+
+type TempFile = { fileName: string };
+
+function getRange(lineNumber: string, columnNumber: string): vscode.Range {
+  const ln = Number(lineNumber);
+  const col = Number(columnNumber);
+  const pos = new vscode.Position(ln > 0 ? ln - 1 : 0, col > 0 ? col - 1 : 0);
+  return new vscode.Range(pos, pos);
+}
```
brpowell, line 43: Needed this to process diagnostics, it's shared by the deploy/retrieve commands in the core package. We could also move this to utils...thoughts?

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
```diff
@@ -0,0 +1,232 @@
...
+        outputText += `${nls.localize('apex_execute_compile_success')}\n`;
+        outputText += `Error: ${diagnostic.exceptionMessage}\n`;
+        outputText += `Error: ${diagnostic.exceptionStackTrace}\n`;
+        outputText += `\n${response.logs}`;
+      }
+    }
+    OUTPUT_CHANNEL.appendLine(outputText);
+  }
+
+  private handleDiagnostics(
```
brpowell, line 177: Took the opportunity to make the result formatting and diagnostics logic private methods of this executor because they are internal details of the command. More encapsulated than have them scattered in separate files.

packages/salesforcedx-vscode-apex/src/constants.ts
```diff
@@ -96,3 +97,6 @@ export const ENABLE_SOBJECT_REFRESH_ON_STARTUP =
   'enable-sobject-refresh-on-startup';
 export const APEX_EXTENSION_NAME = 'salesforcedx-vscode-apex';
 export const LSP_ERR = 'apexLSPError';
+export const OUTPUT_CHANNEL = vscode.window.createOutputChannel(
+  nls.localize('channel_name')
+);
```
brpowell, line 102: Main question i have for the PR: do we want to start having the command output to a separate Apex channel now or later? We could wait until we get all the apex commands in here and switch them all at the same time instead of trickling them one by one into the channel.

packages/salesforcedx-vscode-apex/test/vscode-integration/commands/forceApexExecute.test.ts
```diff
@@ -0,0 +1,257 @@
...
+      settingStub.returns(false);
+      await forceApexExecute();
+      expect(cliExecutorStub.calledOnce).to.be.true;
+      expect(apexTempFile.calledOnce).to.be.true;
+      expect(apexExecutorStub.called).to.be.false;
+      expect(anonGather.called).to.be.false;
+    });
+  });
+
+  describe('Format Execute Anonymous Response', () => {
```
brpowell, line 152: Brought these tests into the command's test file since formatting results are part of the command.

brpowell:
I think I ported everything over/deleted the old stuff for this command. If i could get a double check that'd be helpful 🙇

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
AnanyaJha: If it's shared by the SDR commands too then I think we could move it over to utils

packages/salesforcedx-vscode-apex/src/constants.ts
AnanyaJha: Agreed, we're planning on waiting to move the test:run functionality over to the Apex channel at the end, so I think we could switch the rest of the Apex commands then too

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
lcampos: Cool, as long as this doesn't break testing since sometimes we've left things as public so they can be tested.

packages/salesforcedx-vscode-apex/src/constants.ts
lcampos: I think we should do it now, no need to wait for every command to be ported.

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
brpowell: Having trouble moving it over with the tests since it requires the vscode api ugh. not having luck with proxyquire or anything either.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -285,18 +285,6 @@ function registerCommands(
     'sfdx.force.data.soql.query.selection',
     forceDataSoqlQuery
   );
-
-  const forceApexExecuteDocumentCmd = vscode.commands.registerCommand(
-    'sfdx.force.apex.execute.document',
-    forceApexExecute,
```
lcampos, line 291: Forgot to remove the import for `forceApexExecute`

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import {
```
lcampos, line 7: Looks like all the code was moved to `salesforcedx-vscode-apex`. Can we delete the `forceApexExecute.ts` file from `salesforcedx-vscode-core` ?

brpowell: Ah whoops, yeah i stashed deleting it from core by accident. good catch

lcampos (approved):

PR #2916, opened 2021-02-02
feat(soql): extract comments out of soql text before running query

dobladez:
### What does this PR do?
On SOQL Builder: when there are comments at the top of `.soql` files, strip them out before calling the "Run Query" API.

@W-8611529@

### Functionality Before
image

### Functionality After
image


praksb (commented):
I think the line number in the error message returned from the SOQL API would be off. It would be ok not to address this as part of this PR and track it separately.

jonnyhork:
In conjunction with this [PR](https://github.com/forcedotcom/soql-tooling/pull/132) in the tooling repo, these changes passed QA for me.
Once the new version of the language server is published and we have PDT approval we can merge this.

brpowell (approved):

Issue #2892, opened 2021-01-26
Apex Test showing null ms for execution time


smaddox-sf:
https://twitter.com/robsalesforce/status/1353812301460221952?s=21

Tweet text: "If my Test Execution Time is "null ms", can I tell my client the code runs infinitely faster?"

screenshot:
![image](https://user-images.githubusercontent.com/59572234/105785479-d8c7b380-5f2f-11eb-876f-13bb951ea83e.png)

smaddox-sf:
This has been fixed and was released last week

Issue #2890, opened 2021-01-25
Apex Commands - Shows as successful when there was an issue and missing line numbers on error


smaddox-sf:
Reported from customers here:
- https://twitter.com/brianmfear/status/1352680280859971584
- https://salesforce.stackexchange.com/questions/332122/vs-code-extension-no-longer-including-line-number-in-apex-test-failure-output

**Details of the issue:**. 2 problems
1 - When I run Apex tests from VS Code (leveraging Salesforce CLI), I'm accustomed to the output telling me what line my tests failed on in the event of failure.

 2 - The plugin now reports the test as "ran successfully" instead of an error, no Problems appear in the Problems tab

ryu-727:
I'm facing same issue.

fredrikstaahl:
Same issue here.

pavantiruppur:
Same issue here.

lcampos:
This issue has been fixed in release v51.2.0 of the Salesforce extensions for VSCode. Thanks for providing feedback.

<img width="1560" alt="Screen Shot 2021-02-24 at 5 41 42 PM" src="https://user-images.githubusercontent.com/1041842/109090352-aee8e480-76c7-11eb-9700-ca07b84ab6ae.png">

PR #2922, opened 2021-02-04
feat: add diagnostics to Apex test run failures

AnanyaJha:
### What does this PR do?
Add diagnostics to Apex test run failures. This also updates the version of the templates library, apex library, SDR, and @salesforce/core.

#2890, #2892, @W-8845619@


packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
```diff
@@ -86,7 +91,50 @@ export class ApexLibraryTestRunExecutor extends LibraryCommandletExecutor<{
...
+
+    result.tests.forEach(test => {
+      if (test.diagnostic) {
+        const diagnostic = test.diagnostic;
+        const filepath = join(
+          defaultPackage,
+          'main',
+          'default',
+          'classes',
+          `${diagnostic.className}.cls` ?? ''
```
AnanyaJha, line 113: It seems less than ideal to be hardcoding the path to the Apex test class `main/default/classes` - any ideas on how else we can get this information?

packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
```diff
@@ -126,6 +126,7 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<
     response: ContinueResponse<ApexExecuteParameters>
   ): Promise<boolean> {
     const connection = await workspaceContext.getConnection();
+    // @ts-ignore
```
brpowell, line 129: How come this is ts-ignored?

AnanyaJha: Oh it's just temporary until we bump the version of @salesforce/core across the other repos

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRunCodeAction.ts
brpowell: Yeah this won't work because the class might not be in the default directory or even in main/default/classes. Unless we indexed the project ahead of time, the only option is to resolve the components of the project and find it. Resolution with the library is decently fast.

I have work in progress to let you do this:

```typescript
const class = resolveSource(packageDirectories).find(c => c.fullName === fullName && c.type === type);
```

but for now the API is kind of bad, so try this:

```typescript
let testClass: SourceComponent;
for (const packageDir of packageDirectories) {
  const components = ComponentSet.fromSource(packageDir);
  testClass = components.getSourceComponents({ fullName: 'TestClass', type: 'ApexClass' }).next().value
  if (testClass) {
    break;
  }
}
```

AnanyaJha: Exactly, I wanted to handle the scenarios where the project structure is different than the default main/default/classes. This solution worked for me, thank you! :)

PR #2927, opened 2021-02-08
fix: apex execute diagnostic reporting

brpowell:
### What does this PR do?

Fixes an issue where the diagnostic message wasn't set properly

@W-8857272@

### Functionality Before

Screen Shot 2021-02-08 at 1 58 19 PM

### Functionality After

Screen Shot 2021-02-08 at 2 01 08 PM

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
@@ -182,9 +182,9 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<
         columnNumber
       } = response.diagnostic[0];
       let message;
-      if (compileProblem !== '') {
+      if (compileProblem && compileProblem !== '') {
         message = compileProblem;
-      } else if (exceptionMessage !== '') {
+      } else if (exceptionMessage && exceptionMessage !== '') {
         message = exceptionMessage;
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-apex/src/commands/forceApexExecute.ts
```diff
@@ -175,17 +175,27 @@ export class ApexLibraryExecuteExecutor extends LibraryCommandletExecutor<
...
-            ? diagnostic.compileProblem
-            : diagnostic.exceptionMessage,
+        message,
         severity: vscode.DiagnosticSeverity.Error,
         source: filePath,
         range: this.getZeroBasedRange(
-          diagnostic.lineNumber ?? 1,
-          diagnostic.columnNumber ?? 1
+          Number(lineNumber) || 1,
+          Number(columnNumber) || 1
```
brpowell, line 198: The typings for `lineNumber` and `columnNumber` aren't right - they are returned as strings not numbers. Although the wsdl does say `int` for these fields, but seems to not be serialized that way.

lcampos: Yeah, it depends on which WSDL you are using, the strong typed one might do `int` but the others usually return strings

lcampos (approved):

PR #2932, opened 2021-02-09
refactor: remove core extension dependency from apex extension

brpowell:
### What does this PR do?

Removes the core extension dependency from the apex extension. This allows the Apex extension to start activating before the core extension finishes.

@W-8659996@

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -53,9 +53,6 @@
     "typescript": "3.8.3",
     "vscode-extension-telemetry": "0.0.17"
   },
-  "extensionDependencies": [
-    "salesforce.salesforcedx-vscode-core"
-  ],
   "scripts": {
     "vscode:prepublish": "npm prune --production",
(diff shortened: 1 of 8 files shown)


brpowell:
The apex library setting is still contributed by the core extension (doesn't require extension dependency to be read). Should we remove it and have the setting contributed by Apex extension now?

AnanyaJha:
> The apex library setting is still contributed by the core extension (doesn't require extension dependency to be read). Should we remove it and have the setting contributed by Apex extension now?

Yeah I think it makes sense to move the setting over to the Apex extension now

lcampos:
> > The apex library setting is still contributed by the core extension (doesn't require extension dependency to be read). Should we remove it and have the setting contributed by Apex extension now?
> 
> Yeah I think it makes sense to move the setting over to the Apex extension now

No, let's keep it as it is. Soon we'll be deprecating it and exclusively using the library so no need to move, rename, handle the old setting name, etc.

lcampos (approved):

PR #2934, opened 2021-02-10
fix: activate redhat XML extension only if it is version 0.14.0 @W-8852786@

jag-sfdc:
### What does this PR do?
Activates XML extension only if it is version 0.14.0

W-8852786

### Functionality Before
LWC Extension activation error for some users

### Functionality After
Fix for the LWC Extension activation error for some users


packages/salesforcedx-vscode-lwc/src/metasupport/metaSupport.ts
```diff
@@ -80,8 +80,14 @@ export class MetaSupport {
...
-      // checks plugin version greater than 0.13.0, might need to change.
-      if (parseInt(pluginVersionNumber.split('.')[1], 10) >= 14) {
+      // checks if the installed plugin version is exactly 0.14.0
+      // 0.15.0 has a regression and 0.13.0 or earlier versions are not supported
+
+      if (parseInt(pluginVersionNumber.split('.')[1], 10) === 15) {
+        channelService.appendLine(
+          nls.localize('force_lightning_lwc_redhat_extension_regression')
+        );
+      } else if (parseInt(pluginVersionNumber.split('.')[1], 10) >= 14) {
```
lcampos, line 90: I think we should just pin this down to exactly version 0.14.0 since previous versions don't work and the latest version is 0.15.0

lcampos (approved):

PR #2975, opened 2021-02-23
fix: support defer attribute from webpack 5

jonnyhork:
### What does this PR do?
We upgraded to Webpack 5  to generate the output of the `soql-builder-ui` module. The script tags now include a `defer="defer"` attribute, so we need to update out regex to match appropriately.

Corresponding [PR](https://github.com/forcedotcom/soql-tooling/pull/206)


lcampos (approved):

lcampos:
@jonnyhork if you are now webpacking `soql-builder-ui` does that mean you also modified the `package.json` for that module to move/remove those dependencies that are now bundled ?

packages/salesforcedx-vscode-soql/test/vscode-integration/editor/htmlUtils.test.ts
```diff
@@ -16,7 +16,7 @@ describe('html utilities', () => {
   const pathToLwcDist = SOQL_BUILDER_UI_PATH;
   const html = `
   <!-- CSP TAG -->'
-  <script src="./0.app.js"></script><script src="./app.js"></script>
+  <script defer="defer" src="./0.app.js"></script><script defer="defer" src="./app.js"></script>
```
xyc, line 19: I think defer is a boolean attribute, so `<script defer src="./0.app.js">` should be sufficient.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

jonnyhork: You are correct, however the output in the html is in the format seen here `defer="defer"` so we need to match for that.

jonnyhork:
@lcampos Thats a good point! Actually we have always been webpacking the `soql-builder-ui` however we did not consider the "double dependency" issue I think you're asking about.
One thought we had was to move all our [`dependencies`](https://github.com/forcedotcom/soql-tooling/blob/f30f845902f5f4838d391c37df8e24c783466285/packages/soql-builder-ui/package.json#L10) to `devDependencies`. 
Do you have another recommendation? That seemed like the simplest fix.

lcampos:
> @lcampos Thats a good point! Actually we have always been webpacking the `soql-builder-ui` however we did not consider the "double dependency" issue I think you're asking about.
> One thought we had was to move all our [`dependencies`](https://github.com/forcedotcom/soql-tooling/blob/f30f845902f5f4838d391c37df8e24c783466285/packages/soql-builder-ui/package.json#L10) to `devDependencies`.
> Do you have another recommendation? That seemed like the simplest fix.

That's one option, I know @xyc has looked at removing dependencies that are bundled by using a post-bundle process.

xyc:
We should move (all of) them to `devDependencies` imo.

PR #2991, opened 2021-02-25
Moving sobject-refresh command & automatic sobject refresh code to core

jag-sfdc:
### What does this PR do?
Moves sobject-refresh command & automatic sobject refresh code to core. No functionality change for the end user.

@W-8912259@

### Functionality Before
N/A

### Functionality After
Same as before. No functionality change.


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -619,6 +629,21 @@ export async function activate(context: vscode.ExtensionContext) {
   telemetryService.sendExtensionActivationEvent(extensionHRStart);
   console.log('SFDX CLI Extension Activated');
   return api;
```
rcoringrato-sfdc, line 631: @jag-sfdc Is this return in the proper place?

jag-sfdc: @rcoringrato-sfdc - Definitely not. Fixing it now.

rcoringrato-sfdc (approved):

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -157,10 +157,6 @@
           "command": "sfdx.force.apex.test.run",
```
lcampos, line 157: You should be able to also remove `salesforce/salesforcedx-sobjects-faux-generator` as a dependency in this extension

jag-sfdc: Sure, will do that.

PR #2997, opened 2021-02-28
feat: perform concurrent definition retrieval

rcoringrato-sfdc:
### What does this PR do?

Perform concurrent sobject definition retrieval to improve refresh performance.

@W-8912274@


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -317,6 +307,22 @@ export class FauxClassGenerator {
     return this.successExit();
   }
 
+  private async fetchObjects(
+    describe: SObjectDescribe,
+    types: string[]
+  ): Promise<SObject[]> {
+    const batchSize = 25;
```
lcampos, line 314: Let's add a note here that explains that `25` is the limit of sobjects supported by the describe api.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -274,30 +269,25 @@ export class FauxClassGenerator {
         username: await ConfigUtil.getUsername(projectPath)
       })
     });
-
     const describe = new SObjectDescribe(connection);
-    const standardSObjects: SObject[] = [];
+
+    if (
```
jag-sfdc, line 274: Is it possible to move this cancellation logic inside fetchObjects() ?

jag-sfdc (changes requested):

lcampos (approved):

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
rcoringrato-sfdc: @lcampos I added a constant instead. I thought that would be clearer.

lcampos: @rcoringrato-sfdc that works too 👍

lcampos (approved):

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
rcoringrato-sfdc: @jag-sfdc I think this code will continue to simplify with each work item, so I would like to defer this change until later. What do you think?

jag-sfdc: @rcoringrato-sfdc  - Sounds good to me. I think eventually the cancellation logic should be in describeSObjectBatch or in fetchObjects(). It doesn't help having it here. Am I right in assuming cancellation won't work until then?

rcoringrato-sfdc: @jag-sfdc As long as the operation is cancelled before describeGlobal() completes then the refresh can be cancelled. You are correct that after that point it will not cancel and is not useful. At a minimum the logic should be in describeSObjectBatch, but I'm not sure what state the in-progress jsforce connections will be in if we cancel at that point. I wonder if it is necessary to figure out how to cancel the connection?

jag-sfdc: I think this is fine as is. As you said, let's come back once all the work is complete and look into this.

jag-sfdc (approved):

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
xyc: It's not impossible, you can cancel an XHR from `jsforce` but imo it might not be worth the effort to trickle down the cancellation token from here to `@salesforce/core` to `jsforce`. This part (VS Code side cancellation) is still useful since this is in-between calls
.
(We are looking into similar issues regarding various cancellations: fyi @AnanyaJha )

PR #3008, opened 2021-03-02
refactor: consolidate sobject describe code

lcampos:
### What does this PR do?
Refactor sobject describe code in `salesforcedx-sobject-faux-describe`.
Removes circular dependencies between `describe` and `generator` modules

@W-8964667@


packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -29,24 +36,59 @@ export class SObjectDescribe {
    * @param type SObjectCategory
    * @returns string[] containing the sobject names
    */
-  public async describeGlobal(type: SObjectCategory): Promise<string[]> {
+  public async describeGlobal(
```
lcampos, line 39: Moved the filter logic to this call. Only if we execute the complete flow (query for sobject list, query for all sobject describes, create faux classes, etc) should we filter the sobjects. In the following weeks we'll start introducing flows where we already have the list of sobjects and in those scenarios we don't want to filter things out.

This keeps backwards compatibility with existing usage.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -115,7 +159,16 @@ export class SObjectDescribe {
     }
   }
 
-  public getVersion(): string {
-    return `${this.versionPrefix}${this.targetVersion}`;
+  public async fetchObjects(types: string[]): Promise<SObject[]> {
```
lcampos, line 162: Moving this function here allows us to do all sobject describe requests from this class.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -24,16 +24,16 @@ import {
...
-  MAX_BATCH_REQUEST_SIZE,
   SObject,
   SObjectCategory,
-  SObjectDescribe
-} from '../describe';
-import { ConfigUtil } from '../describe/configUtil';
-import { nls } from '../messages';
+  SObjectRefreshSource
+} from '../types';
+import { ConfigUtil } from './configUtil';
```
lcampos, line 36: Was only being used by this module so move it

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -29,24 +36,59 @@ export class SObjectDescribe {
    * @param type SObjectCategory
    * @returns string[] containing the sobject names
    */
-  public async describeGlobal(type: SObjectCategory): Promise<string[]> {
+  public async describeGlobal(
+    type: SObjectCategory,
```
rcoringrato-sfdc, line 40: @lcampos nit: It might be a little clearer if this is name "category" instead of "type".

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
rcoringrato-sfdc: @lcampos How about using **const connection = await workspaceContext.getConnection()** to get all connections?

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
lcampos: I agree that's a better name, I'll change it.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
lcampos: That is an option but I'm leaning more towards removing the code that gets the connection object (in following PRs) and have that be something the vscode extension provides to the `salesforcedx-sobject-faux-generator` module. This would be similar to what we already do on SDR and the Apex library.

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -29,24 +36,59 @@ export class SObjectDescribe {
...
+    source: SObjectRefreshSource
+  ): Promise<string[]> {
     const requestedDescriptions: string[] = [];
     const allDescriptions: DescribeGlobalResult = await this.connection.describeGlobal();
 
-    allDescriptions.sobjects.forEach((sobject: DescribeGlobalSObjectResult) => {
-      const isCustom = sobject.custom === true;
+    for (const sobject of allDescriptions.sobjects) {
```
xyc, line 46: nit: can use `Array.prototype.filter` here

rcoringrato-sfdc (approved):

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
lcampos: I wasn't able to cause we also need to return a string (sobject.name) instead of the complete object

xyc: `map` + `filter` or `reduce` could do it in one pass (Just a style preference. Fine with this too)

lcampos: I always forget about `reduce`, thanks for the suggestion I've change it to use that.

PR #3018, opened 2021-03-04
feat: generate typings for sobjects

rcoringrato-sfdc:
### What does this PR do?

Added type generation to the sobject faux object flow.
@W-8912288@


packages/salesforcedx-sobjects-faux-generator/src/generator/declarationGenerator.ts
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ChildRelationship, Field, SObject } from '../types';
+
+export const MODIFIER = 'global';
+export interface FieldDeclaration {
```
rcoringrato-sfdc, line 10: @jag-sfdc The static representation could contain a list of these objects with some additional information. That's sufficient to generate apex and type files.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -26,17 +26,13 @@ import {
...
-  SObject,
-  SObjectCategory,
-  SObjectRefreshSource
-} from '../types';
+import { SObject, SObjectCategory, SObjectRefreshSource } from '../types';
 import { ConfigUtil } from './configUtil';
+import { DeclarationGenerator, MODIFIER } from './declarationGenerator';
+import { TypingGenerator } from './typingGenerator';
 
+const TYPING_PATH = ['typings', 'lwc', 'sobjects2'];
```
rcoringrato-sfdc, line 34: I haven't figured out how to do the transition yet. ALS will collide with this typing generation until it is disabled.

lcampos: We can have a "local" variable/setting to gate this being enabled but still allow us to write tests, merge it, etc.

packages/salesforcedx-sobjects-faux-generator/src/generator/declarationGenerator.ts
```diff
@@ -0,0 +1,159 @@
...
+  name: string;
+  comment?: string;
+}
+
+export interface SObjectDefinition {
+  name: string;
+  fields: FieldDeclaration[];
+}
+
+export class DeclarationGenerator {
```
lcampos, line 22: Eventually we need to switch the main entry point to `salesforcedx-sobject-faux-generator` to not be `FauxClassGenerator` and be something that ties/calls all the potential format outputs. Do you think this could be it or would we have a different class/code calling this, FauxClassGenerator and TypingsGenerator ?

rcoringrato-sfdc: I'm leaning toward a different class as the orchestrator - DeclarationGenerator would just convert the data and FauxClassGenerator would just create the CLS output.

lcampos: Sounds good 👍

packages/salesforcedx-sobjects-faux-generator/src/generator/typingGenerator.ts
```diff
@@ -0,0 +1,105 @@
...
+    for (const sobject of sobjects) {
+      if (sobject.name) {
+        this.generateTypingForObject(targetFolder, sobject);
+      }
+    }
+  }
+
+  public generateTypingForObject(folderPath: string, sobject: SObject): string {
+    const typingPath = path.join(folderPath, sobject.name + '.d.ts');
+    if (fs.existsSync(typingPath)) {
```
xyc, line 35: this isn't needed as `fs.writeFileSync` overwrites the file if it exists

rcoringrato-sfdc: I was getting a permission error before I added this, probably since the perms were set to readonly when the previous contents were written?

xyc: Interesting, my definition files have permission `644`. Yeah it makes sense to remove first if we are writing with permission `444`.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -258,25 +235,43 @@ export class FauxClassGenerator {
...
     // This method is different from generateFauxClasses -  generateFauxClasses takes SObject array as input
     // and to generate that we would need a large definition file. If we go one level more specific, as what
     // generateAndWriteFauxClasses here requires, simpler declarations we have in the minSObjects.json is good.
     for (const sobject of sobjectDecl) {
-      const fauxClassPath = path.join(standardSObjectsFolderPath, sobject.name + '.cls');
-      sobject.fields.forEach(field => { field.modifier = 'global'; });
-      fs.writeFileSync(fauxClassPath, this.generateFauxClassTextFromDecls(sobject.name, sobject.fields), {
+      const fauxClassPath = path.join(
+        standardSObjectsFolderPath,
+        sobject.name + APEX_CLASS_EXTENSION
```
lcampos, line 269:
```suggestion
        `${sobject.name}${APEX_CLASS_EXTENSION}`
```

lcampos: I did this change directly in GitHub so please check that it does work before merging

packages/salesforcedx-sobjects-faux-generator/src/generator/declarationGenerator.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export const MODIFIER = 'global';
+export interface FieldDeclaration {
+  modifier: string;
+  type: string;
+  name: string;
+  comment?: string;
+}
+
+export interface SObjectDefinition {
```
lcampos, line 17: We could move this and `FieldDeclaration` to the `types` folder

packages/salesforcedx-sobjects-faux-generator/src/generator/declarationGenerator.ts
lcampos: Re-surfacing this thread, is it correct to assume this would happen on a follow up PR given `FauxClassGenerator` is still the entry point for sobject creation ?

lcampos:
@rcoringrato-sfdc if we merge these changes we'll be creating the lwc typings twice (here and from the Apex lang server). Are we going to add something to disable lwc typings from the faux module while we make the changes to the lang server or are we going to wait for the lang server changes before merging this PR ?

rcoringrato-sfdc:
@lcampos I added a class level flag to disable type declarations by default. As part of the story to add the updated ALS jar this flag should be removed. This should prevent the collision that you referred to. This PR could be committed separate from the ALS changes.

packages/salesforcedx-sobjects-faux-generator/src/generator/declarationGenerator.ts
rcoringrato-sfdc: Will do,

lcampos:
> @lcampos I added a class level flag to disable type declarations by default. As part of the story to add the updated ALS jar this flag should be removed. This should prevent the collision that you referred to. This PR could be committed separate from the ALS changes.

Sounds good, I totally missed `shouldGenerateTypes` being used in the faux class generator.

lcampos (approved):

jag-sfdc (approved):

Issue #3009, opened 2021-03-02
Switch to Apex output channel automatically after running Apex tests


AnanyaJha:
### Summary

After running Apex tests from the sidebar, command palette, or code action, the output should automatically be displayed via the `Apex` output channel. Currently to display the output, the user needs to manually select the `Apex` output channel from the drop down list provided. VS Code should automatically switch to the appropriate output channel as needed.

### Steps To Reproduce:
Select a test to run via the sidebar
Run test
Look for results in output window at the bottom of the screen

### Expected result
When the command finishes running, user should immediately see the results in the output window below - regardless of what channel the window was originally on.

Actual result
Nothing happens.

**VS Code Version**: 50.17.0 and up

robcowell:
I think this may be a broader change than just Apex tests - example scenario :-

- Ran tests, now watching in the Apex view
- Tests fail, requiring code change
- Edit code, save it - push on save enabled in VS Code
-     Expected : switch to Salesforce CLI view automatically so I can see confirmation of save/compile cycle to Org
-     Actual : it stayed on the Apex tab

So while I understand the "separation of concerns" for the plugin pack, following context and moving to appropriate views would greatly support this change

smaddox-sf:
Thanks for the input @robcowell.  You're right that it should consistently handle the output view based on the action.  We'll get this addressed in an upcoming release.

AnanyaJha:
Displaying the Apex output channel automatically was addressed by release v51.4.0, closing this issue now. Feel free to open a new issue if there are any other inconsistencies.

PR #3027, opened 2021-03-06
fix: pop up output channel upon completion of Apex commands

AnanyaJha:
### What does this PR do?
This PR fixes the issue where the output channel was not popping up automatically when an Apex-related command finishes running. These changes will cause the output channel to show automatically for 
- SFDX: Execute Anonymous with Currently Selected Text
- SFDX: Execute Anonymous with Editor Contents
- SFDX: Get Apex Debug Logs
- SFDX: Run Apex Tests
- Running Apex tests through the sidebar
- Running Apex tests through the code lens

This PR also fixes issues with SFDX: Get Debug Logs. Previously:
1) With the Apex Library setting turned off, the log output was previously displayed twice
2) With the Apex library setting turned on, the log output was not displayed at all. We made changes on the Apex library side to return the logs correctly to fix this.

The output is now consistent across both scenarios - the log is displayed once in the output channel

#3009 @W-8957541@ @W-8874346@

### Functionality Before

noOutputChannel

### Functionality After

ApexOutputChannel


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -25,7 +25,7 @@
     "Programming Languages"
   ],
   "dependencies": {
-    "@salesforce/apex-node": "0.1.15",
+    "@salesforce/apex-node": "0.1.16",
```
AnanyaJha, line 28: This will need to be updated again once the next version of the library is released

packages/salesforcedx-vscode-apex/test/vscode-integration/commands/forceApexExecute.test.ts
```diff
@@ -159,11 +159,9 @@ describe('Force Apex Execute', () => {
 
   describe('Format Execute Anonymous Response', () => {
     let outputStub: SinonStub;
-    let showOutputChannel: SinonStub;
```
AnanyaJha, line 162: any ideas on how to test the channel service? since showChannelOutput is called by the base class in the utils package now (commandletExecutors.ts), it doesn't seem to be stubbing/testing as expected :/

lcampos: I think I ran into a similar issue with the telemetry service. Have you tried stubbing or spying on the channel output before calling the code in utils ? Do we have "generic" tests for the channel service when it's called by the commandlet code as part of the utils module ?

AnanyaJha: Yeah I tried stubbing/spying on `channelService.showChannelOutput()` but it thinks that the method never gets called so the stub/spy is ineffective. It looks like in the utils module we're using proxyquire to do any testing around channel service, but we also aren't actually testing  `showChannelOutput` directly by stubbing/spying on it- probably for the same reason lol

xyc: Have you tried stubbing `ChannelService.prototype.showChannelOutput`? Also I think it's not called in `executor.run` here, rather `executor.execute`.

packages/salesforcedx-vscode-apex/test/vscode-integration/commands/forceApexExecute.test.ts
```diff
@@ -193,9 +194,8 @@ describe('Force Apex Execute', () => {
       );
 
       await executor.run({ type: 'CONTINUE', data: {} });
-
+      expect(showChannelOutputStub.notCalled).to.be.true;
```
lcampos, line 197: Shouldn't this be `calledOnce` instead of `notCalled` now that showing the channel output is set to true at the commandlet level ?

AnanyaJha: We're also calling `showChannelOutput` at the commandlet level now as part of the `execute()` method. This test calls the `run()` method directly instead of calling `execute()`, so `showChannelOutput` shouldn't get called at this level

lcampos (approved):

PR #3037, opened 2021-03-08
 Generate a subset of SObject definitions at the start @W-8912349@

jag-sfdc:
### What does this PR do?
 Generate a subset of SObject definitions at the start from a static definition file instead of querying the org.

@W-8912349@


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -72,6 +73,11 @@ export interface FieldDeclaration {
   comment?: string;
 }
 
+export interface SObjectDefinition {
```
jag-sfdc, line 76: Once @rcoringrato-sfdc 's changes go in, I'll remove this.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -10,6 +10,7 @@ import * as fs from 'fs';
 import { EOL } from 'os';
 import * as path from 'path';
 import { mkdir, rm } from 'shelljs';
+import * as minSObjectsFromFile from '../../minSObjects.json';
```
lcampos, line 13: Does this work when using the vsix files ? I think we need to change it to something like we do here https://github.com/forcedotcom/salesforcedx-vscode/blob/93ca07ba5ef8f2a20f6431a5f7075a6637658d43/packages/salesforcedx-vscode-lwc/src/metasupport/metaSupport.ts#L32

jag-sfdc: I haven't tested with vsix yet. Will try this.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -30,7 +30,7 @@
   },
   "scripts": {
     "compile": "webpack",
-    "lint": "tslint --project .",
+    "lint": "tslint --project . -e **/minSObjects.json",
```
xyc, line 33: can we use a tslint config file?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
lcampos: Actually I'm wrong, this should work cause you're reading it on the faux-generator module side not from the code directly in the `salesforcedx-vscode-*` modules. However, we might want to move it somewhere under the `src` folder.

packages/salesforcedx-sobjects-faux-generator/package.json
jag-sfdc: @xyc - That was my first choice and couldn't get it to work.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
jag-sfdc: @lcampos - How about moving it to a new "data" directory? It will align with where we have registry.json in SDR.

lcampos: That works too 👍

xyc: +1 for under `src`. `src/data` is fine too :)

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -267,47 +268,38 @@ export class FauxClassGenerator {
...
+      sobject.fields.forEach(field => { field.modifier = 'global'; });
+      fs.writeFileSync(fauxClassPath, this.generateFauxClassTextFromDecls(sobject.name, sobject.fields), {
+        mode: 0o444
+      });
     }
+  }
 
-    return this.successExit();
+  // VisibleForTesting
+  public getSObjectSubsetDefinitions(): SObjectDefinition[] {
```
lcampos, line 301: We can delete `startupMinSObjects` on line 41

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
```diff
@@ -145,7 +145,7 @@ export class ForceRefreshSObjectsExecutor extends SfdxCommandletExecutor<{}> {
     const commandName = execution.command.logName;
     try {
       let result;
-      if (response.data.source === SObjectRefreshSource.StartupMin) {
+      if (response.data.source === SObjectRefreshSource.Startup) {
```
lcampos, line 148: We have two different sobject refresh startup flows;`SObjectRefreshSource.Startup` which uses the `enable-sobject-refresh-on-startup` setting and the expectation is to download all custom objects and sobjects, the other flow is `SObjectRefreshSource.StartupMin` which is triggered by the user clicking on a pop-up message on the bottom right corner when there are no sobjects present in the project and it only creates 16 sobjects.

I don't think we should do this since that would force the flow that uses `enable-sobject-refresh-on-startup` setting to just create sobjects from cache and the pop-up message flow to sync all the objects from the org.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -267,47 +268,38 @@ export class FauxClassGenerator {
...
+    this.logSObjects('Standard', sobjectDecl.length);
+    return this.successExit();
+  }
 
-    try {
-      this.generateFauxClasses(standardSObjects, standardSObjectsFolderPath);
-    } catch (errorMessage) {
-      return this.errorExit(errorMessage);
+  // VisibleForTesting
+  public generateAndWriteFauxClasses(sobjectDecl: SObjectDefinition[], standardSObjectsFolderPath: string) {
```
lcampos, line 290: What's the benefit of adding this new method instead of using the existing `generateFauxClasses` used in the normal generator flow ?

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
jag-sfdc: Agree! We want both these to only download a list of min. For some reason when I earlier searched for StartupMin usage, I didn't find the pop-up message code path. I see it now. I'll take care of this.

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
jag-sfdc: The generateFauxClasses takes SObject array as input and to generate that we would need the earlier 2M definition file.  If we go one level more specific, as what generateAndWriteFauxClasses here requires, we should be fine with the simpler declarations we have in the json now.

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
jag-sfdc: One question - If we are going to make the enable-sobject-refresh-on-startup to be the default behavior as part of ther next story, do we even need the pop-up route to generate sobjects?

jag-sfdc: I have made the change to make both the code paths work the same. As part of the next story, I can get rid of the pop-up route to generate sobjects.

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
```diff
@@ -232,46 +237,17 @@ function getStandardSObjectsDirectory(projectPath: string) {
...
-      }
-    } else {
-      telemetryService.sendEventData(
-        'sObjectRefreshNotification',
-        { type: 'SObjects exist' },
-        undefined
-      );
+  if (projectPath && !fs.existsSync(getStandardSObjectsDirectory(projectPath))) {
+    telemetryService.sendEventData(
+      'Refresh SObjects',
```
lcampos, line 242: I think we had a conversation on a PR some time ago about starting to use a naming convention for telemetry events. I don't recall what we decided on, @sfsholden, @AnanyaJha  or @xyc do you remember if we decided on a naming convention ?

packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
lcampos: Got it, can you add a comment about this at the top of the method ?

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
sfsholden: I am not sure - I do not think I was a part of this discussion

lcampos: Since no one remembers let's just keep the telemetry name to what it used to be (`sObjectRefreshNotification`) so  it is still usable in current dashboards and queries.

lcampos (approved):
Just need to change the telemetry event name, everything else looks good to go.

PR #3049, opened 2021-03-09
feat: display error to user when auth issue

dehru:
### What does this PR do?
This PR address the issue https://github.com/forcedotcom/soql-tooling/issues/69 reported on soql-tooling.  Some of the solution was provided by earlier work in [this PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/2810).  I decided to add an actual vscode poptart to display the error much more obviously to the user.  Additionally, when auth issues happen, our extension tends to flood the channel with several of the same errors.  So I added a debounce to prevent that.  We just need 1 message.

@W-8580593@

### Functionality Before

Way too many messages in the channel.  And no real error message to see unless the user happens to be looking at the channel output.

Screen Shot 2021-03-09 at 4 38 01 PM

### Functionality After

I used a manipulated accessToken and refreshToken to mock an expired auth token in my `.sfdx/user.json` file.

Screen Shot 2021-03-09 at 4 13 29 PM


packages/salesforcedx-vscode-soql/test/vscode-integration/sfdx.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('displays an error if can not get connection to org', async () => {
+    sandbox.stub(sfdx.workspaceContext, 'getConnection').throws();
+    const vscodeErrorMessageSpy = sandbox.spy(
+      vscode.window,
+      'showErrorMessage'
+    );
+    await sfdx.withSFConnection(async () => {});
+    sfdx.debouncedShowChannelAndErrorMessage.flush();
```
dehru, line 24: i found out that the debounce module adds a` flush() `method thankfully!

xyc: i haven't tried this, but `@sinonjs/fake-timers` https://sinonjs.org/releases/v1.17.6/fake-timers/ seems to be the idiomatic way of doing this test

jonnyhork (approved):

packages/salesforcedx-vscode-soql/src/messages/i18n.ja.ts
```diff
@@ -31,5 +31,7 @@ export const messages = {
   error_unknown_error:
     'ERROR: %s. Unknown error. Open an issue and provide the error message details: https://github.com/forcedotcom/soql-tooling/issues/new/choose.',
   error_data_view_save:
-    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.'
+    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.',
+  error_connection:
+    'ERROR: We can’t query your org. Make sure that you’re connected to an authorized org and have permissions to view the object and fields.'
```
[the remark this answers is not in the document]
ekapner:
```suggestion
    'ERROR: We can’t query your org. Make sure that you’re connected to this org and have permissions to view the object and fields.'
```

packages/salesforcedx-vscode-soql/src/messages/i18n.ts
```diff
@@ -32,5 +32,7 @@ export const messages = {
   error_unknown_error:
     'ERROR: %s. Unknown error. Open an issue and provide the error message details: https://github.com/forcedotcom/soql-tooling/issues/new/choose.',
   error_data_view_save:
-    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.'
+    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.',
+  error_connection:
+    'ERROR: We can’t query your org. Make sure that you’re connected to an authorized org and have permissions to view the object and fields.'
```
ekapner, line 37:
```suggestion
    'ERROR: We can’t query your org. Make sure that you’re connected to this org and have permissions to view the object and fields.'
```

packages/salesforcedx-vscode-soql/src/messages/i18n.ja.ts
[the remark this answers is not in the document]
ekapner: Leaning toward "this org" because it's likely an expired auth token issue - and I'm assuming they selected the org they want to use.

ekapner (commented):
slight wording change for error message

packages/salesforcedx-vscode-soql/test/vscode-integration/sfdx.test.ts
dehru: @xyc - i tried that one but couldn't get it to work.  I tried with the whole fake clock and .tick(1000) and that api.  I have used that in the past.

xyc: you might also try `import { debounce } from "debounce";` (so that it's not default export), and mock `debounce` to be `f` => f.call()`. Then we don't have to export `debouncedShowChannelAndErrorMessage` just for testing

jonnyhork: @xyc are you thinking something like [this](https://github.com/forcedotcom/soql-tooling/blob/198fe5596d3414f744ba82ec0168e59bfaa953b1/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts#L15)?

xyc: yeah, in sinon it's easier. Take a look at [fs stubs](https://github.com/forcedotcom/salesforcedx-vscode/blob/d1913dffa24eadc0e7d611ae675fa5b7704650e4/packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testIndexer/lwcTestIndexer.test.ts#L75-L78)

dehru: Man, i battled this for a couple hours, having previously had success with sinon and setTimeouts.  I'll take another whack at it.

dehru: So i revisited this again.  But sorry, I've spent time yesterday and today trying to use sinon FakeTimers and/or overwrite the prototype debounce to get this to work.  I just can't get that test to pass without using the provided flush() function of the debounce library.

Once difference between fs stubs and this module is it only exports a single default function, not an object.  I can't seem to intercept and/or overwrite that before it's cached by node.

packages/salesforcedx-vscode-soql/src/messages/i18n.ja.ts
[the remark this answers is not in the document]
dehru: Updated

packages/salesforcedx-vscode-soql/src/messages/i18n.ts
```diff
@@ -32,5 +32,7 @@ export const messages = {
   error_unknown_error:
     'ERROR: %s. Unknown error. Open an issue and provide the error message details: https://github.com/forcedotcom/soql-tooling/issues/new/choose.',
   error_data_view_save:
-    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.'
+    'ERROR: We can’t save the file to the specified directory. Make sure you have write permissions for the directory.',
+  error_connection:
+    'ERROR: We can’t query your org. Make sure that you’re connected to this org and have permissions to view the object and fields.'
```
lcampos, line 37: cc @ekapner

packages/salesforcedx-vscode-soql/src/sfdx.ts
```diff
@@ -17,14 +19,25 @@ export const channelService = ChannelService.getInstance(
 
 export const workspaceContext = WorkspaceContextUtil.getInstance();
 
+function showChannelAndErrorMessage(e: any) {
```
lcampos, line 22: You should add telemetry to this error too.

packages/salesforcedx-vscode-soql/test/vscode-integration/sfdx.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  beforeEach(() => {
+    sandbox = sinon.createSandbox();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('displays an error if can not get connection to org', async () => {
```
lcampos, line 17: This test should also verify you are sending data to the channel service

packages/salesforcedx-vscode-soql/test/vscode-integration/sfdx.test.ts
xyc: Sounds good, I'm fine with this too. I was thinking that debounce was using named exports `import { debounce } from 'debounce'` as well as default exports which might be utilized. But let's not spend more time on this

packages/salesforcedx-vscode-soql/src/sfdx.ts
dehru: @lcampos - I was thinking that having an issue with the connection is just a normal part of sfdx development.  Scratch orgs expire as part of normal development.  Lemme know if you'd still like to see me adding telemetry.  Happy to.

lcampos: You are correct in that this error with expired orgs is expected. However, there is value in knowing the amount of times users run into this. It gives us visibility so we can make product decisions to minimize them running into this or also potentially catching an unrelated error causing spikes on this error scenario.

dehru: Gotcha.  Sometimes we use telemetry to catch edge cases instead of metrics.  And that's a philosophy flaw.

packages/salesforcedx-vscode-soql/test/vscode-integration/sfdx.test.ts
dehru: and verified i'm sending telemetry as well.

lcampos (approved):

PR #3054, opened 2021-03-10
feat: changes to libraryCommandletExecutor interface to support cancellation

AnanyaJha:
### What does this PR do?
Changes to the libraryCommandletExecutor interface to support progress reporting and cancellation

@W-8965162@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -171,10 +171,6 @@ export abstract class LibraryCommandletExecutor<T>
           cancellable: this.cancellable
         },
         (progress, token) => {
-          token.onCancellationRequested(() => {
-            notificationService.showCanceledExecution(this.executionName);
-            return false;
-          });
           return this.run(response, progress, token);
         }
       );
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -118,6 +118,7 @@ export abstract class SfdxCommandletExecutor<T>
 export abstract class LibraryCommandletExecutor<T>
   implements CommandletExecutor<T> {
   protected cancellable: boolean = false;
+  protected cancelled: boolean = false;
   private readonly executionName: string;
   private readonly logName: string;
   private readonly outputChannel: vscode.OutputChannel;
@@ -171,6 +172,10 @@ export abstract class LibraryCommandletExecutor<T>
           cancellable: this.cancellable
         },
         (progress, token) => {
+          token.onCancellationRequested(() => {
+            this.cancelled = true;
+            notificationService.showCanceledExecution(this.executionName);
+          });
           return this.run(response, progress, token);
         }
       );
@@ -182,13 +187,16 @@ export abstract class LibraryCommandletExecutor<T>
         channelService.showChannelOutput();
       }
 
-      if (success) {
-        notificationService
-          .showSuccessfulExecution(this.executionName)
-          .catch(e => console.error(e));
-      } else {
-        notificationService.showFailedExecution(this.executionName);
+      if (!this.cancelled) {
+        if (success) {
+          notificationService
+            .showSuccessfulExecution(this.executionName)
+            .catch(e => console.error(e));
+        } else {
+          notificationService.showFailedExecution(this.executionName);
+        }
       }
+
       this.telemetry.addProperty('success', String(success));
       const { properties, measurements } = this.telemetry.build();
       telemetryService.sendCommandEvent(
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -170,7 +170,13 @@ export abstract class LibraryCommandletExecutor<T>
           location: vscode.ProgressLocation.Notification,
           cancellable: this.cancellable
         },
-        (progress, token) => this.run(response, progress, token)
+        (progress, token) => { 
+          token.onCancellationRequested(() => {
+            notificationService.showCanceledExecution(this.executionName);
+            return false;
+          })
+          return this.run(response, progress, token)
+        }
       );
       channelService.showCommandWithTimestamp(
         `${nls.localize('channel_end')} ${this.executionName}`
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -117,6 +117,7 @@ export abstract class SfdxCommandletExecutor<T>
 
 export abstract class LibraryCommandletExecutor<T>
   implements CommandletExecutor<T> {
+  protected cancellable: boolean = false;
   private readonly executionName: string;
   private readonly logName: string;
   private readonly outputChannel: vscode.OutputChannel;
@@ -144,9 +145,19 @@ export abstract class LibraryCommandletExecutor<T>
    * @param response Data from the parameter gathering step.
    * @returns Whether or not the execution was a success
    */
-  public abstract run(response: ContinueResponse<T>): Promise<boolean>;
+  public abstract run(
+    response: ContinueResponse<T>,
+    progress?: vscode.Progress<{
+      message?: string | undefined;
+      increment?: number | undefined;
+    }>,
+    token?: vscode.CancellationToken
+  ): Promise<boolean>;
 
   public async execute(response: ContinueResponse<T>): Promise<void> {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
+    const cancellationToken = cancellationTokenSource.token;
+
     const startTime = process.hrtime();
     const channelService = new ChannelService(this.outputChannel);
     const telemetryService = TelemetryService.getInstance();
@@ -159,9 +170,10 @@ export abstract class LibraryCommandletExecutor<T>
       const success = await vscode.window.withProgress(
         {
           title: nls.localize('progress_notification_text', this.executionName),
-          location: vscode.ProgressLocation.Notification
+          location: vscode.ProgressLocation.Notification,
+          cancellable: true
         },
-        () => this.run(response)
+        (progress, token) => this.run(response, progress, token)
       );
       channelService.showCommandWithTimestamp(
         `${nls.localize('channel_end')} ${this.executionName}`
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -158,9 +169,10 @@ export abstract class LibraryCommandletExecutor<T>
       const success = await vscode.window.withProgress(
         {
           title: nls.localize('progress_notification_text', this.executionName),
-          location: vscode.ProgressLocation.Notification
+          location: vscode.ProgressLocation.Notification,
+          cancellable: true
```
xyc, line 173: this should be called with `this.cancellable`

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -143,9 +144,19 @@ export abstract class LibraryCommandletExecutor<T>
...
+    response: ContinueResponse<T>,
+    progress?: vscode.Progress<{
+      message?: string | undefined;
+      increment?: number | undefined;
+    }>,
+    token?: vscode.CancellationToken
+  ): Promise<boolean>;
 
   public async execute(response: ContinueResponse<T>): Promise<void> {
+    const cancellationTokenSource = new vscode.CancellationTokenSource();
```
xyc, line 157: we can remove line 157 & 158 as tokens are provided in `vscode.window.withProgress`

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -159,9 +167,16 @@ export abstract class LibraryCommandletExecutor<T>
       const success = await vscode.window.withProgress(
         {
           title: nls.localize('progress_notification_text', this.executionName),
-          location: vscode.ProgressLocation.Notification
+          location: vscode.ProgressLocation.Notification,
+          cancellable: this.cancellable
         },
-        () => this.run(response)
+        (progress, token) => {
+          token.onCancellationRequested(() => {
```
xyc, line 174: can we have a test covering this?

brpowell, line 174: We need to handle the check below whether it's successful or not so that both the cancel message and failure message don't show up, just the former.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -159,9 +167,16 @@ export abstract class LibraryCommandletExecutor<T>
...
+            notificationService.showCanceledExecution(this.executionName);
+            return false;
```
brpowell, line 176: How does this `return false` work?

packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
```diff
@@ -97,7 +97,7 @@ export class NotificationService {
     );
   }
 
-  private showCanceledExecution(executionName: string) {
+  public showCanceledExecution(executionName: string) {
```
AnanyaJha, line 100:
```suggestion
  private showCanceledExecution(executionName: string) {
```

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
AnanyaJha: I decided to remove this change for now, keeping this limited to just changing the method definition

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
AnanyaJha: Removed this for this PR

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -159,9 +167,12 @@ export abstract class LibraryCommandletExecutor<T>
...
         {
           title: nls.localize('progress_notification_text', this.executionName),
-          location: vscode.ProgressLocation.Notification
+          location: vscode.ProgressLocation.Notification,
+          cancellable: this.cancellable
         },
-        () => this.run(response)
+        (progress, token) => {
+          return this.run(response, progress, token);
+        }
```
brpowell, line 175: In terms of showing the cancel message, should we do something like this?
```typescript
// kinda pseudocode, can't remember the apis off the top of my head
(progress, token) => {
  if (this.cancellable) {
    token.onCancellation(() => wasCancelled = true)
  }
  return this.run(response, progress, token);
}

// ...

if (success) {
  notificationService.showSuccessMessage()
} else if (wasCancelled) {
  notificationService.showCancellationMessage();
} else {
  notificationService.showFailureMessage();
}
```

xyc: Not sure the above code would make `this.run` return immediately. It depends on the underlying implementation of `this.run`.

we can move `notificationService.showSuccessMessage` and `notificationService.showFailureMessage` to the `(progress,token) => {...}` callback body, and utilize the `token.isCancellationRequested`.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -159,9 +168,16 @@ export abstract class LibraryCommandletExecutor<T>
...
+          location: vscode.ProgressLocation.Notification,
+          cancellable: this.cancellable
         },
-        () => this.run(response)
+        (progress, token) => {
+          token.onCancellationRequested(() => {
+            this.cancelled = true;
+            notificationService.showCanceledExecution(this.executionName);
+          });
+          return this.run(response, progress, token);
```
brpowell, line 179: Bringing up the conversation from before:

Correct me if I'm wrong, but this is still dependent on how `run` is implemented and it always will. There is no way to stop this promise unless the concrete class resolves run appropriately when cancellation is requested. If i don't resolve run after cancellation is requested in my concrete class, the progress notification will keep going and we'll also show the cancellation message, even though we'll still be waiting for `withProgress` to finish.

Nothing is wrong with the code here, just setting the expectation of this behavior.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -117,6 +117,8 @@ export abstract class SfdxCommandletExecutor<T>
 
 export abstract class LibraryCommandletExecutor<T>
   implements CommandletExecutor<T> {
+  protected cancellable: boolean = false;
+  protected cancelled: boolean = false;
```
brpowell, line 121:
```suggestion
  private cancelled: boolean = false;
```

Might not want concrete classes to be able to change this value, they should go through the token.

AnanyaJha: good call!

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
AnanyaJha: Yeah, that's basically it! This logic will only handle displaying the appropriate notification to the user, it does not affect the actual execution of the command in `run`. The only difference is that the progress notification actually does not keep going after the user hits cancel, it disappears and the `Cancel` notification is displayed.

packages/salesforcedx-utils-vscode/test/unit/commands/commandletExecutor.test.ts
```diff
@@ -102,6 +102,21 @@ describe('LibraryCommandletExecutor', () => {
...
+    const showInfoStub = sb
+      .stub(vscodeStub.window, 'showInformationMessage')
+      .resolves(nls.localize('notification_show_in_status_bar_button_text'));
+    const cancelledExecutor = new TestExecutor(new MockChannel());
+    cancelledExecutor.cancelled = true;
+
+    await cancelledExecutor.execute({ data: { success: true }, type: 'CONTINUE' });
+    expect(showErrStub.notCalled).to.be.true;
+    expect(showInfoStub.notCalled).to.be.true;
+  });
```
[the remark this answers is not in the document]
brpowell: Yeah it's a tricky test to write. Could probably do something like create our own cancellationTokenSource and make the vscodeStub.withProgress function call the task with the token from that source. But then we need to handle the event loop to make sure the onCancellationRequest listener is called before the test is over. I can try to do something like that in a subsequent PR, but this is a good start for some sanity.

brpowell (approved):

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
brpowell: Gotcha. Sounds like there's some internal behavior in withProgress that handles stopping the notification if the token is cancelled.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -171,13 +187,16 @@ export abstract class LibraryCommandletExecutor<T>
         channelService.showChannelOutput();
       }
 
-      if (success) {
-        notificationService
-          .showSuccessfulExecution(this.executionName)
-          .catch(e => console.error(e));
-      } else {
-        notificationService.showFailedExecution(this.executionName);
+      if (!this.cancelled) {
```
xyc, line 190: Do we want to move this to `(progress, token) => {...}` callback body, after this.run? ex.:
```
const result = this.run(response, progress, token);
if (!token || !token.isCancellationRequested) {
  // show success / fail notification
}
return result;
```
That way we don't have to define an extra class variable.

packages/salesforcedx-utils-vscode/test/unit/commands/commandletExecutor.test.ts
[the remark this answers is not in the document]
xyc: yeah, we can mock withProgress to be like:
```
withProgressStub.callsFake((options, task) => {
   // create our own mock token
   // See the (over)simplified version of cancellation token in the design doc
    task(mockProgress, mockToken);
});
// we can cancel because we own the token
```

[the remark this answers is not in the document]
AnanyaJha: I tried this approach and it didn't behave as expected. I don't want to block bryan and shelby's story so I've left it as is for now and I'll address this and remove the cancellable class property in a follow up story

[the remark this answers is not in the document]
xyc: Sounds good. We can address that later which is more about internal engineering. The class property is private and doesn't affect this API

PR #3059, opened 2021-03-12
Replay debugger test sidebar status & code coverage

lcampos:
### What does this PR do?
Support code coverage and status coloring in Apex test sidebar when using the one click replay debugger.

This PR should be merged after https://github.com/forcedotcom/salesforcedx-vscode/pull/3060

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
@@ -16,15 +16,21 @@ export function ensureDirectoryExists(filePath: string): void {
   fs.mkdirSync(filePath);
 }
 
-export function getTestResultsFolder(vscodePath: string, testType: string) {
+export function getTestResultsFolder(rootDirPath: string, testType: string) {
   const dirPath = path.join(
-    vscodePath,
+    rootDirPath,
     '.sfdx',
     'tools',
     'testresults',
     testType
   );
 
+  ensureDirectoryExists(dirPath);
+  return dirPath;
+}
+
+export function getLogDirPath(rootDirPath: string): string {
+  const dirPath = path.join(rootDirPath, '.sfdx', 'tools', 'debug', 'logs');
   ensureDirectoryExists(dirPath);
   return dirPath;
 }
--- a/packages/salesforcedx-vscode-apex-replay-debugger/src/commands/quickLaunch.ts
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/src/commands/quickLaunch.ts
@@ -7,6 +7,7 @@
 import {
   ApexTestResultData,
   LogService,
+  ResultFormat,
   TestLevel,
   TestResult,
   TestService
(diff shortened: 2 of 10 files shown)


packages/salesforcedx-vscode-apex-replay-debugger/src/utils/index.ts
```diff
@@ -4,9 +4,5 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { getRootWorkspacePath } from '@salesforce/salesforcedx-utils-vscode/out/src/workspaces';
-import * as path from 'path';
 
-export function getLogDirPath(): string {
```
lcampos, line 10: Moved this to the utils module so we can use it in apex and apex replay debugger extensions

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -16,9 +16,9 @@ export function ensureDirectoryExists(filePath: string): void {
   fs.mkdirSync(filePath);
 }
 
-export function getTestResultsFolder(vscodePath: string, testType: string) {
+export function getTestResultsFolder(rootDirPath: string, testType: string) {
   const dirPath = path.join(
```
AnanyaJha, line 20: we can add a return type to this method

packages/salesforcedx-vscode-apex-replay-debugger/src/commands/quickLaunch.ts
```diff
@@ -79,6 +88,20 @@ export class QuickLaunch {
...
+        const apexTestResultsPath = getTestResultsFolder(
+          getRootWorkspacePath(),
+          'apex'
+        );
+        await testService.writeResultFiles(
+          result,
+          { dirPath: apexTestResultsPath, resultFormats: [ResultFormat.json] },
+          retrieveTestCodeCoverage()
+        );
+      }
```
AnanyaJha, line 103: It looks like the test sidebar isn't getting updated after the first run due to how the result files are generated for synchronous test runs. Synchronous test runs don't have a test run id - so after the first test result file is generated, `test-result-.json`, each subsequent run just _updates_ the content of that file. We could change the fileWatcher to fire on the `onDidChange` event to get around this, otherwise does this command need to use a synchronous test run? We could also switch to asynchronous test runs to avoid this.

On a sidenote, I'll file a bug on the apex library because it looks like we're appending an extra `-` onto the end of the result file

lcampos: Yeah, I'll have to think about that. I'm leaning towards changing the watchers.

Specific to the additional `-` character, are you talking about the content or the name of the file ? The name of the file should be `test-result-<test result id>.json`, since sync test runs don't generate test result ids the current behavior is correct.

AnanyaJha: Sounds good, I think changing the watchers works!

I was referring to the name of the file, right now it gets created with the name `test-result-.json`. This makes sense, but it looks like the CLI currently creates it as just `test-result.json`. We can add this to the list of changes coming in the future, but I think we'll need to switch to the CLI version for now since we aren't touching the result files with this first release.

lcampos: Sounds good 👍

gbockus-sf:
Covered by PR #3381

PR #3067, opened 2021-03-16
chore: Render test for SOQL Builder webview

jgellin-sf:
### What does this PR do?
This PR implements a WebDriver-based test to verify that the SOQL Builder webview is actually rendered. Integration tests cover most SOQL extension function, but unnoticed dependency version mismatches recently caused a situation where all tests passed, yet the SOQL extension could not render the SOQL Builder.

WebDriver-based tests are inherently time-consuming and flaky, owing so much of their success to timing and built-in waits, so they should be used sparingly. This PR creates a single test verifying multiple criteria to save time:
- Open folder containing sfdx-project to enable SOQL extenion commands
- Open SOQL file in the text editor; verify it is open
- Toggle to SOQL Builder; verify it is open
- Verify rendering by checking existence of querybuilder-app element

This implementation is sufficient to avoid the particular recent problem, and is not currently designed to be an exhaustive survey of the SOQL extension's UI.

Considerations for follow-on work:
- This PR puts all verifications in a single test that takes about 20 seconds on average. Multiple tests would repeat many of the same setup steps and multiply this duration. More granular steps could be considered.
- This PR includes the UI tests as part of the `test:vscode-integration` script, so it can be run as part of CI; we may want to call this out as a separate kind of test in a special CI workflow that only runs on some commits.
- This PR verifies rendering minimally by only checking the existence of the querybuilder-app element. If, in the future, checks are needed to verify the existence of child elements within the SOQL Builder, note that this is complicated by the shadow DOM.
- This PR makes no attempt to verify rendering of query results. To do so, we would have to connect an org or fake running a query without an org, and there is currently no mechanism to do so.

@W-8977278@

### Functionality Before
N/A

### Functionality After
N/A


packages/salesforcedx-vscode-soql/test/ui-test/soql-ui-smoke.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      'test',
+      'ui-test',
+      'resources',
+      'sfdx-test-project'
+    ) + path.sep;
+  const filename = 'example.soql';
+
+  const openFolder = async (folder: string) => {
+    const workbench = new Workbench();
+    const cmdTitle = 'Extest: Open Folder';
```
jonnyhork, line 41: `extest` is a utility command provided by the testing framework. There are some recommended ways to open files and folders but they have been unreliable for us. 
Using the `extest` command has been simple and stable so far. 
https://github.com/redhat-developer/vscode-extension-tester/wiki#opening-files-and-folders

.circleci/config.yml
```diff
@@ -353,6 +361,17 @@ jobs:
             echo $SFDX_CI_DEVHUB_JWTKEY | base64 --decode > devhub.key
             sfdx force:auth:jwt:grant --clientid ${SFDX_CI_DEVHUB_CLIENTID} --username ${SFDX_CI_DEVHUB_USERNAME} --jwtkeyfile devhub.key --setdefaultdevhubusername --setalias devhub
             rm devhub.key
+      - run:
```
lcampos, line 364: Why do this instead of having a scheduled task that runs your tests every "x" amount of time e.g. run the tests once a day ?

jonnyhork: Thats a good question, I had not thought about a chron job for these tests.  Perhaps it would simplify the CI and be enough of a check.

One downside is a delay in knowing when something is broken and does that gap expose us to missing an issue before it released?

jonnyhork: One tricky part of the current script is that it will only run with workflows on the develop branch, we decided that is probably fine for now, keep it really simple. 
That wouldn't be an issue any more with a chron job, I would imagine we could just have it run on the main branch.

jonnyhork: Alrighty, after some discussion with the team, I'm going to switch this to a chron job.

jonnyhork: @lcampos @xyc Please see the `fix: run UI tests with chron job` commit to run Ui test once a day at midnight with the other jobs.

lcampos (approved):

PR #3091, opened 2021-03-25
fix: create new query command opens soql builder

dehru:
### What does this PR do?
With the last update of VSCode ( Version: 1.54.3 ), behavior around opening a new empty soql query command changed.  Digging into this, i discovered we were probably riding on top of a bug that got fixed or a loop-hole that got closed.  This PR fixes the issue.

#### Nitty Gritty Details
CustomEditors are associated with a filename pattern and not a languageId.  

```
"customEditors": [
      {
        ...
        "selector": [
          {
            "filenamePattern": "*.soql"
          }
        ]
      }
```

Previously, when we created a new untitled file for a soql query, we did not specify a file name that matched our filenamePattern, but rather just specified a languageId like this:

```
const doc = await vscode.workspace.openTextDocument({
      language: 'soql',
      content: ''
    });
```

Yet, user's were still previously presented with a Soql Builder ui.  Unfortunately, this stopped working in Version: 1.54.3.

This fix addresses that issue and honors the filenamePattern by naming the untitled file `untitled.soql`, as well as uses the special `untitled` scheme, which let's vscode know that the file has no title and is not in the file system ( until the user elects to  save it ).  There's a lot of discussion [on this issue here](https://github.com/microsoft/vscode/issues/93441).

Additionally, I discovered that the toggle feature requires a file on disk, and so the toggle feature does not present in the editor menu until the file is saved ( see the new when condition addition ).  This also matches the existing behavior, so there is no functional change to the UX, as the toggle editor menu would not appear until the file was saved.

@W-8990594@

### Functionality Before
https://user-images.githubusercontent.com/599418/112550580-37b97580-8d85-11eb-937c-4f6a28042474.mov

### Functionality After
https://user-images.githubusercontent.com/599418/112550616-43a53780-8d85-11eb-914f-12605389f8a3.mov


packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
```diff
@@ -20,9 +20,5 @@ export async function soqlBuilderToggle(doc: vscode.Uri): Promise<void> {
     ? BUILDER_VIEW_TYPE
     : EDITOR_VIEW_TYPE;
 
-  vscode.commands.executeCommand(
-    OPEN_WITH_COMMAND,
-    vscode.Uri.file(doc.fsPath),
```
dehru, line 25: this just produces a Uri, which is what `doc` already is.  So I just use that as is since the [openWith command takes a Uri](https://github.com/microsoft/vscode/blob/afca44535df55707d5f0b1dd65098060b448c8a8/src/vs/workbench/contrib/customEditor/browser/customEditors.ts#L137).

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -17,16 +17,16 @@ export async function soqlOpenNew(): Promise<void> {
...
 
   if (vscode.workspace) {
-    // create untitled file
-    const doc = await vscode.workspace.openTextDocument({
-      language: 'soql',
-      content: ''
+    const fileName = 'untitled.soql';
+    const newUri = vscode.Uri.file(fileName).with({
+      scheme: 'untitled',
+      path: fileName
```
dehru, line 23: ultimately, the Uri for an untitled file ends up looking like `untitled:untitled.soql` instead of when the file is on disc, which is a `file:` scheme.

packages/salesforcedx-vscode-soql/test/vscode-integration/commands/soqlFileCreate.test.ts
```diff
@@ -21,8 +21,6 @@ describe('soqlOpenNew should', () => {
   beforeEach(() => {
     sb = createSandbox();
     telemetryStub = sb.stub(telemetryService, 'sendCommandEvent') as SinonStub;
-    editorOpened = sb.stub();
```
dehru, line 24: this is no longer called, since we do not interact with the openTextDocument() api.  Therefore, removed from test and associated expectations.

packages/salesforcedx-vscode-soql/test/vscode-integration/commands/soqlFileCreate.test.ts
```diff
@@ -37,7 +35,7 @@ describe('soqlOpenNew should', () => {
     await soqlOpenNew();
 
     expect(telemetryStub.called).is.true;
-    expect(editorOpened.called).is.true;
     expect(executeCommandSpy.getCall(0).args[2]).contains(BUILDER_VIEW_TYPE);
+    expect(executeCommandSpy.getCall(0).args[1].scheme).contains('untitled');
```
dehru, line 39: verify the untitled scheme is used

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -69,7 +69,9 @@
     "papaparse": "^5.3.0",
     "sinon": "7.5.0",
     "typescript": "3.8.3",
-    "vscode-languageclient": "6.1.3"
+    "vscode-languageclient": "6.1.3",
+    "webpack": "4.30.0",
```
dehru, line 73: without this, we can't actually run `npm run compile` from within the `packages/salesforcedx-vscode-soql` directory.  which is a huge time saver.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -145,7 +147,7 @@
       ],
       "editor/title": [
         {
-          "when": "resourceLangId == soql",
+          "when": "resourceLangId == soql && resourceScheme == file",
```
dehru, line 150: Toggling from the text editor to the soql builder with an unsaved file causes this issue shown below.  By checking the resourceScheme, we just don't show the toggle icon until the file is saved.  Which prevents users from toggling back and forth and getting into this situation.  Once saved, toggle works as expected.

https://user-images.githubusercontent.com/599418/112553326-e8c20f00-8d89-11eb-8ca8-ccd6abda10c5.mov

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -17,16 +17,16 @@ export async function soqlOpenNew(): Promise<void> {
   telemetryService.sendCommandEvent('soql_builder_open_new', process.hrtime());
 
   if (vscode.workspace) {
-    // create untitled file
-    const doc = await vscode.workspace.openTextDocument({
-      language: 'soql',
-      content: ''
+    const fileName = 'untitled.soql';
```
dobladez, line 20: If I run the "Create Query in SOQL..." command a second time before saving, nothing happens. I guess it is because it is always the same filename here.

If we are OK with this behavior, fine. If not, you might want to extract that hard-coded filename into a function to generate a unique file for the current workspace.

dobladez:
Tricky stuff! Thanks for the explanations.

So, to make sure I understand:   Would this not be a problem if the "customEditors" selector supported "languageId" (instead of file extension) ?

dobladez (approved):
👍  Just double-check the observation about the uniqueness of the generated filename.

jgellin-sf (approved):
Nice

dehru:
> Tricky stuff! Thanks for the explanations.
> 
> So, to make sure I understand: Would this not be a problem if the "customEditors" selector supported "languageId" (instead of file extension) ?

Yes.  And it does kinda seem like a flaw in their design.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
dehru: I messed with this a bit, and the behavior of untitled files with Custom Editors even with a numeric file sequence is not any better.  Check out this movie.  I'm a little dumbfounded...and ignorant still about much of the vscode OS along with the behavior of Custom Editors.  If i use the openTextDocument() api, then the Save As doesn't behave as we expect.

https://user-images.githubusercontent.com/599418/112670483-9125b100-8e26-11eb-8d9e-c8c1aa3571d5.mov

dehru: At this point, I say we go out with this behavior and see what people think.  At least we are functioning again.

jonnyhork (approved):

sfsholden (approved):

PR #3092, opened 2021-03-26
Spike: Sobject refresh cancellation @W-8952920@

jag-sfdc:
### What does this PR do?
Spike to cancel SObject refresh at different stages

@W-8952920@


packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts
```diff
@@ -123,6 +122,14 @@ export class FauxClassGenerator {
         username: await ConfigUtil.getUsername(projectPath)
       })
     });
+
```
jag-sfdc, line 125: Cancellation check after getting connection

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -132,44 +133,56 @@ export class SObjectDescribe {
...
     }
-  }
 
-  public async fetchObjects(types: string[]): Promise<SObject[]> {
-    const batchSize = MAX_BATCH_REQUEST_SIZE;
-    const requests = [];
-    for (let i = 0; i < types.length; i += batchSize) {
-      const batchTypes = types.slice(i, i + batchSize);
-      requests.push(this.describeSObjectBatchRequest(batchTypes));
+    const fetchedObjects: SObject[] = [];
```
jag-sfdc, line 149: Check as part of each batch request and throw a custom error to handle cancellation

jag-sfdc:
Spike doc - https://salesforce.quip.com/3U2GAgP3wPsH

packages/salesforcedx-sobjects-faux-generator/src/describe/sObjectDescribe.ts
```diff
@@ -132,44 +133,56 @@ export class SObjectDescribe {
...
+    if (
+      cancellationToken &&
+      cancellationToken.isCancellationRequested
+    ) {
+      throw new Error('SObject Refresh Cancelled');
```
AnanyaJha, line 154: Do we need to throw an error if the request was cancelled or can we just exit early here?

jag-sfdc: Since we are doing this as a batch process, there would be multiple requests running in parallel. So, there is a chance we end up canceling only one request and not the others. But throwing an error will force the calling method to treat all the batch requests as cancelled.

xyc: I think it should be fine if we don't throw error and make this check after `Promise.all` call. By the time we reach here, the requests are already in progress.

That would also help to surface the actual error in the requests.

jag-sfdc: By doing a check after Promise.all, we are waiting for all the requests to complete. By doing it here, even if some of the requests are still in progress, we would exit. I agree about we missing actual errors. Let me think about this a bit more. Thank you.

xyc: I don't think in this scenario "we would exit" either. Consider the following code:

```
async function f() {
  try {
    const request1 = new Promise(async (resolve, reject) => {
      await new Promise((resolve) => setTimeout(resolve, 1000));
      throw new Error("interrupted");
    });
    const request2 = new Promise((resolve) => {
      setTimeout(() => {
        console.log("Im the long running request");
        resolve();
      }, 5000);
    });
    await Promise.all([request1, request2]);
  } catch (error) {}
}

f();
```
You should still be able to see "Im the long running request" output even after it has thrown an error. Once the request has been made, it generally wouldn't be cancellable.

jag-sfdc: Thank you @xyc!
That makes sense, so I could just add a check after Promise.all.

jag-sfdc: `async function f() {
  try {
    const request1 = new Promise(async (resolve, reject) => {
      await new Promise((resolve) => setTimeout(resolve, 1000));
      throw new Error("interrupted");
    });
    const request2 = new Promise((resolve) => {
      setTimeout(() => {
        console.log("Im the long running request");
        resolve();
      }, 5000);
    });
    await Promise.all([request1, request2]);
  } catch (error) {
      console.log("I have exited!");
  }
}

f();`

Just one more thing - What I was trying is not to cancel a request that has been made. If the "I have exited" is printed before "Im the long running request", it means we are cancelling without waiting for all requests to complete. But again this is not worth the complexity it brings. So, will just check after Promise.all like you suggested.

xyc: that's right, but we should note that once cancel button is clicked the dialog is dismissed immediately and we are handling this in background here.

Issue #3095, opened 2021-03-29
"Development Models" Documentation and the "Org Picker"


jeffblenmanindygov:
### Summary

The documentation for ["Create Project and Use Org Browser"](https://developer.salesforce.com/tools/vscode/en/user-guide/development-models/#create-project-and-use-org-browser) incorrectly tells the user to find the "Org Picker" in the Side Bar instead of the Status Bar.

### Steps To Reproduce (from the [documentation](https://developer.salesforce.com/tools/vscode/en/user-guide/development-models/#create-project-and-use-org-browser)):

1.  Open the VS Code editor and from the Command Palette, run SFDX: Create Project.
1.  In the Side Bar of the code editor, click Org Picker.

### Expected result

After creating an SFDX project, find an "Org Picker" icon in the Visual Studio Code "Side Bar."

### Actual result

After creating an SFDX project, the user is unable to find the "Org Picker" mentioned in the instructions.  First-time users get stuck since they have never authorized an org through VSC.

### Additional information

![VSC_Window](https://user-images.githubusercontent.com/49252516/112874401-e82ebe80-9090-11eb-93f9-3efd0f1e6538.png)

Eventually the user figures out (or is told) they can just go directly to the Command Palette to find what they need and skip step 2 of the instructions.  

This step seems unnecessary for basic project creation.  If the intent is to expose the user to the Org Picker it may be more useful to offer it as an alternative to just using the Command Palette the same way they used it in Step 1 to create their project.  You could provide a link to an Org Picker section with details about the 3 different authentication types and a screenshot of where the Org Picker is in VS Code.  For example, in or after Step 3 there could be something like, "You can also use the _Org Picker_ to see other authorization commands (for Dev Hub or scratch orgs) or select an existing authorized org from the list."

**VS Code Version**:  1.54.3

**SFDX CLI Version**:   sfdx-cli/7.91.0-6a6ed69ebe win32-x64 node-v14.15.4

**OS and version**:  Windows 10 Pro 20H2 (build 19042.868)

smaddox-sf:
HI @jeffblenmanindygov - Thanks for catching this and taking the time to report it!  We'll get the doc corrected.  CC @ekapner

PR #3099, opened 2021-03-29
[GitHub feedback] Update Org Picker location

ekapner:
https://github.com/forcedotcom/salesforcedx-vscode/issues/3095

### Functionality Before
We said Org Picker was located on Side Bar

### Functionality After
Update docs to say Org Picker is on status bar

### DIFF
--- a/docs/_articles/en/user-guide/development-models.md
+++ b/docs/_articles/en/user-guide/development-models.md
@@ -118,7 +118,7 @@ You can delete source from your project and from your non-source-tracked org.
 To develop in non-scratch orgs and use Org Browser to retrieve source:
 
 1. Open the VS Code editor and from the Command Palette, run **SFDX: Create Project**.
-1. In the code editor's status bar, click Org Picker. This opens the Command Palette and you can select a command to authorize an org, Dev Hub, or create a scratch org; Or you can select from the list of authorized orgs.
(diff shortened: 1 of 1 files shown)


docs/_articles/en/user-guide/development-models.md
```diff
@@ -118,7 +118,7 @@ You can delete source from your project and from your non-source-tracked org.
 To develop in non-scratch orgs and use Org Browser to retrieve source:
 
 1. Open the VS Code editor and from the Command Palette, run **SFDX: Create Project**.
-1. In the Side Bar of the code editor, click Org Picker. This opens the Command Palette and you can select a command to authorize an org, Dev Hub, or create a scratch org; Or you can select from the list of authorized orgs.
+1. In the code editor's status bar, click Org Picker. This opens the Command Palette and you can select a command to authorize an org, Dev Hub, or create a scratch org; Or you can select from the list of authorized orgs.
```
jshackell-sfdc, line 121: Two tiny nits:

- To get around the "This" not having a noun after it, you can reword the second sentence like this: "The Command Palette opens and you can select a command..."  (I'm personally not bothered by it, but Acrolinx complains). 
- I think the semi-colon after "..or create a scratch org;"  should be a period rather than semi-colon.

ekapner: Thanks, I noticed those things but didn't address them. Will do.

jshackell-sfdc (approved):

PR #3101, opened 2021-03-30
feat: turn experimental.deployRetrieve on by default

sfsholden:
### What does this PR do?
Flips source deploy retrieve on by default.

@W-9076651@


sfsholden:
Removing the SDR upgrade to next week - we have some code changes and test failures to address. Bryan and I decided it would be better to not rush those changes.

brpowell (approved):

PR #3103, opened 2021-03-30
Report Apex tests progress and allow cancellation

xyc:
### What does this PR do?
Consume changes in Apex lib for progress reporting and tests cancellation

@W-8911030@

### Functionality Before
no detailed progress reports and tests are not cancellable

### Functionality After
reporting progress and tests are cancellable
apex-progress


packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
```diff
@@ -222,10 +230,33 @@ export class ApexLibraryTestRunExecutor extends LibraryCommandletExecutor<
         payload = { testLevel: TestLevel.RunAllTestsInOrg };
     }
 
-    const result = await testService.runTestAsynchronous(payload, codeCoverage);
+    const progressReporter: Progress<ApexTestProgressValue> = {
```
AnanyaJha, line 233: this handles updating the vs code progress based on the events we receive from the `runTestsAsynchronous` operation?

AnanyaJha:
@ekapner could you take a look at the text in the progress notification in the gif above? Any comments on the formatting?

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
xyc: Yes

AnanyaJha (approved):
LGTM, let's just see if Emily has any thoughts about the formatting of the progress message

ekapner:
> LGTM, let's just see if Emily has any thoughts about the formatting of the progress message

My only comment was it took me a second to parse the command name. The message does the job - My first thought is to bold the command name so it stands out - but I don't want to be inconsistent with other similar messages - and I don't think we've done that anywhere else. So in short, it works - and I don't want to introduce any inconsistencies.

PR #3126, opened 2021-04-06
Xyc/all local tests

xyc:
### What does this PR do?
Matches VS Code behavior of running local tests with CLI.
- Run All Local Tests: Runs all tests in the current org except the ones that originate from installed managed packages
- Run All Tests: Runs all tests in the current org

@W-9074371@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
@@ -42,6 +42,7 @@ import * as settings from '../settings';
 
 export enum TestType {
   All,
+  AllLocal,
   Suite,
   Class
 }
@@ -66,6 +67,14 @@ export class TestsSelector
         description: testSuite.fsPath,
         type: TestType.Suite
       };
+    });
+
+    fileItems.push({
+      label: nls.localize('force_apex_test_run_all_local_test_label'),
+      description: nls.localize(
+        'force_apex_test_run_all_local_tests_description_text'
+      ),
+      type: TestType.AllLocal
     });
 
     fileItems.push({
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/src/messages/i18n.ja.ts
```diff
@@ -41,7 +41,10 @@ export const messages = {
   force_apex_log_list_text: 'Apex デバッグログを取得しています',
   force_apex_test_run_all_test_label: 'すべてのテスト',
   force_apex_test_run_all_tests_description_text:
-    '現在のプロジェクトのすべてのテストを実行',
+    'Runs all tests in the current org',
+  force_apex_test_run_all_local_test_label: 'All Local Tests',
+  force_apex_test_run_all_local_tests_description_text:
+    'Runs all tests in the current org except the ones that originate from installed managed packages',
```
AnanyaJha, line 47: @ekapner let us know if there are any edits, thanks!

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
```diff
@@ -68,6 +69,14 @@ export class TestsSelector
       };
     });
 
+    fileItems.push({
```
AnanyaJha, line 72: Nit - do we want to keep `Run All Tests` at the top of the picklist and move `Run All Local Tests` to the second spot? That way the picklist is arranged from general -> specific options to run tests

packages/salesforcedx-vscode-apex/src/messages/i18n.ja.ts
ekapner: LGTM

ekapner (approved):

lcampos (approved):
Aside from @AnanyaJha 's comment/question, everything else looks good.

jag-sfdc (approved):

packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts
xyc: That was actually my original change. I'm arranging it this way since running local tests is default for production deployments, and it's the behavior of "Run All Tests" with library toggle turning off (Before the changes introduced in this PR). I'd like the first/default option to be more likely/frequently the option that customers would be choosing.

AnanyaJha: Cool that makes sense to me as well

AnanyaJha (approved):

PR #3127, opened 2021-04-06
fix: better exception handling for type inference errors

brpowell:
### What does this PR do?

Properly handles file paths when throwing TypeInfereceError exceptions. Also fixes it for when we are creating component counts in the baseDeployExecutor. 

@W-9105085@

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
@@ -28,3 +28,31 @@ export function getTestResultsFolder(vscodePath: string, testType: string) {
   ensureDirectoryExists(dirPath);
   return dirPath;
 }
+
+/**
+ * Creates a project relative path version of an absolute path.
+ *
+ * @param fsPath Absolute file path
+ * @param packageDirs Package directory paths
+ * @returns Relative path for the project
+ */
+export function getRelativeProjectPath(
+  fsPath: string = '',
+  packageDirs: string[]
+) {
+  let packageDirIndex;
+  for (let packageDir of packageDirs) {
+    if (!packageDir.startsWith(path.sep)) {
+      packageDir = path.sep + packageDir;
+    }
+    if (!packageDir.endsWith(path.sep)) {
+      packageDir = packageDir + path.sep;
+    }
+    packageDirIndex = fsPath.indexOf(packageDir);
+    if (packageDirIndex !== -1) {
+      packageDirIndex += 1;
+      break;
+    }
+  }
+  return packageDirIndex !== -1 ? fsPath.slice(packageDirIndex) : fsPath;
+}
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -47,8 +47,7 @@ type DeployRetrieveOperation = MetadataApiDeploy | MetadataApiRetrieve;
 
 export abstract class DeployRetrieveExecutor<
   T
-  > extends LibraryCommandletExecutor<T> {
-
+> extends LibraryCommandletExecutor<T> {
   protected cancellable: boolean = true;
 
   constructor(executionName: string, logName: string) {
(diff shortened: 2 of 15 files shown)


packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -28,3 +28,31 @@ export function getTestResultsFolder(vscodePath: string, testType: string) {
...
 }
+
+/**
+ * Creates a project relative path version of an absolute path.
+ *
+ * @param fsPath Absolute file path
+ * @param packageDirs Package directory paths
+ * @returns Relative path for the project
+ */
+export function getRelativeProjectPath(
```
brpowell, line 39: Moved to helper utils

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
```diff
@@ -4,37 +4,23 @@
...
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
-import { ComponentSet, MetadataType } from '@salesforce/source-deploy-retrieve';
+import { getRelativeProjectPath } from '@salesforce/salesforcedx-utils-vscode/out/src';
 import { MetadataComponent } from '@salesforce/source-deploy-retrieve';
-import * as vscode from 'vscode';
-import { sfdxCoreSettings } from '../../settings';
-
-export function useBetaDeployRetrieve(
```
brpowell, line 13: This function isn't really needed anymore - it was from when we needed to detect the type before deploying or retrieving in addition to checking the setting. Now we can just check the setting. I got rid of it now because it was actually causing problems - If we couldn't infer the type here no deploy would trigger at all.

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
```diff
@@ -4,37 +4,23 @@
...
+  if (e.name === 'TypeInferenceError') {
+    const projectPath = getRelativeProjectPath(
+      e.message.slice(0, e.message.lastIndexOf(':')),
+      await SfdxPackageDirectories.getPackageDirectoryPaths()
+    );
+    formattedException.message = `${projectPath}: Could not infer metadata type`;
   }
 
-  return true;
+  return formattedException;
```
brpowell, line 23: Convert to relative project path if it's a TypeInferenceError. Otherwise only throw the exception name with "Unknown Exception" as message.

packages/salesforcedx-vscode-core/src/commands/util/index.ts
```diff
@@ -40,7 +40,4 @@ export {
   DevUsernameChecker,
   EmptyPreChecker
 } from './preconditionCheckers';
-export {
-  createComponentCount,
-  useBetaDeployRetrieve
-} from './betaDeployRetrieve';
+export { createComponentCount } from './betaDeployRetrieve';
```
lcampos, line 43: We should also export `formatException` here so we can import it elsewhere by just using the `./utils` path

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
```diff
@@ -4,37 +4,23 @@
...
-  for (const { fsPath } of uris) {
-    const componentsForPath = ws.resolveSourceComponents(fsPath);
-    if (supportedTypes && componentsForPath) {
-      for (const component of componentsForPath) {
-        if (!permittedTypeNames.has(component.type.name)) {
-          return false;
-        }
-      }
-    }
+  if (e.name === 'TypeInferenceError') {
```
lcampos, line 15: Is there any chance that the error message does not follow this format (e.g. <path> : <message>) ?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -40,15 +41,15 @@ import { nls } from '../messages';
 import { DeployQueue } from '../settings';
 import { SfdxPackageDirectories, SfdxProjectConfig } from '../sfdxProject';
 import { createComponentCount } from './util';
+import { formatException } from './util/betaDeployRetrieve';
```
lcampos, line 44: We should be able to import this by just using `./util` path

packages/salesforcedx-vscode-core/src/commands/util/betaDeployRetrieve.ts
brpowell: Only if we change the structure of the message :)

PR #3134, opened 2021-04-07
feat: improve support for relationship queries (nested SELECTs)

dobladez:
Related PR: https://github.com/forcedotcom/soql-language-server/pull/25

### What does this PR do?

* complete nested FROM with parent-child relationships:
    `SELECT (SELECT Id FROM |) FROM Account`
* complete nested SELECT with fields of the relationship object:
    `SELECT (SELECT | FROM Users) FROM Account`
* complete nested ORDER BY with fields of the relationship object:
    `SELECT (SELECT Id FROM Users ORDER BY |) FROM Account`
* fix: don't propose more than one level of SELECT nesting
* fix: don't propose aggregate functions or GROUP BY on nested SELECT
* only propose Id fields on semi-join SELECTs

@W-8558994@

### Functionality Before
Code-completion on FROM clauses (even on nested queries) was suggesting all sobjects as candidates.

### Functionality After
On nested queries, only parent-child relationship names are proposed as candidates:

https://user-images.githubusercontent.com/152839/113082260-82d1ef00-91b0-11eb-8ef8-e277a0869b2e.mov


dobladez:
Reminder: Codecov does NOT capture the coverage of code-completion integration tests

rcoringrato-sfdc (approved):

Issue #2243, opened 2020-06-08
Setting to always force push


lukecotter:
Alot of the time we are never going to make changes directly in the org but sometimes there are still conflicts that occur. 

An option in settings to always force push via sfdx force:source:push -f (--forceoverwrite) would help avoid the prompt to retry the push with the override, saving time.

smaddox-sf:
@lukecotter - Thanks for the taking the time to provide feedback.  Can you tell me more about the scenarios when you always want to overwrite?  

For example, do you always want to override for certain Orgs?  (such as a Scratch Org or a personal sandbox that you don't share with others)

lukecotter:
The case I am thinking about only affect scratch orgs. I am a developer and each developer works in their own scratch org. There are a few cases where this can be helpful. 

### Always force push because local is true
Sometimes I do not care what changes I may have made in the org and for development of an entire feature, I consider only what is local to be true. If I wanted to make a change in the org and have it persist locally I would pull before pushing.

In that case I do no want to run the risk of a conflict and waste time having to push and then force push after, I would rather know it will always succeed and be the same as local.

### Issues with incorrect conflicts being reported
In the passed I seem to have had some cases where conflicts were reported when one was not expected. I can not remember specifics but I seem to remember this occurring on the second push after pushing a change to the admin profile.

Again in those cases I do not want to run the risk of hitting the conflict and I am happy the org is always overridden with what is in local.

### Sometimes on large repos the push fails
I have seen the push fail due to memory usage. In those cases force pushing has succeeded where the normal push failed. Something else that has solved this issues was deleting the org folder in the `.sfdx` folder but this takes longer on the second push than force pushing.

lukecotter:
Generally normal push is fine there are just those few cases where I find my self going to the command line to use --forceoverwrite adding a setting to add --forceoverwrite when a file is saved would prevent me from having to go to the commandline.

lukecotter:
closed because this has been merged and is available to use 👍🏻

PR #3144, opened 2021-04-12
feat: setting to always force push on save

lukecotter:
### What does this PR do?
Adds a new vscode setting to always use `--forceoverwrite` when doing `force:source:push` on save.
Useful if you do not care about conflicts or never make changes directly in an org
e.g prettier adds new line at end of a class which would result in a conflict on push 
#2243

### Functionality Before
Push on save would just do a push meaning if conflicts existed the command with overwrite would need to be run.

### Functionality After
With the setting enabled it can be assumed that you local copy is correct and always overwrite on push on save.


packages/salesforcedx-vscode-core/src/commands/forceSourcePush.ts
```diff
@@ -33,8 +34,11 @@ export class ForceSourcePushExecutor extends BaseDeployExecutor {
       .withArg('force:source:push')
       .withJson()
       .withLogName('force_source_push_default_scratch_org');
-    if (this.flag === '--forceoverwrite') {
-      builder.withArg(this.flag);
+    if (
+      sfdxCoreSettings.getForcePushAndPullEnabled() ||
+      this.flag === '--forceoverwrite'
+    ) {
+      builder.withArg('--forceoverwrite');
```
brpowell, line 41: Might want to enable the force push here instead:

https://github.com/forcedotcom/salesforcedx-vscode/blob/06e7e5fa1b8ef4c19b2057d56d679003a3ccbcba/packages/salesforcedx-vscode-core/src/settings/pushOrDeployOnSave.ts#L84

and have a check like

```typescript
let command = 'sfdx.force.source.push';
if (sfdxCoreSettings.getForcePushOnSave()) {
  command += '.force'
}
vscode.commands.executeCommand(command);
```

brpowell:
Thanks for contributing @lukecotter 💯 . For `pull` I'm not sure we want to have a setting to force overriding conflicts, since we already a command for both `pull` and `force pull`. With `push` though this makes sense with the deploy on save scenario - we just call the force push version of the command instead since like pull, push has two versions. I left a comment on where to go about making that change.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -845,6 +845,11 @@
       "type": "object",
       "title": "%feature_previews_title%",
       "properties": {
+        "salesforcedx-vscode-core.force-push-and-pull.enabled": {
```
brpowell, line 848: @smaddox-sf Looking for your feedback on the setting name. Following my comment on how this is a deploy on save feature, should we make the setting in that scope?

e.g.

`salesforcedx-vscode-core.push-or-deploy-on-save.enabled`
`salesforcedx-vscode-core.push-or-deploy-on-save.overrideConflictsOnPush`

lukecotter:
@brpowell I had originally done it that way. Lol.
I'll make the change to be push only, I agree that it makes sense when push on save is on.

packages/salesforcedx-vscode-core/package.json
smaddox-sf: @brpowell - I like your suggestion and agree we should keep it within the 'upon save' scope.  For the description of this setting, we should also note it's upon save and remove the mention of pull...so something like "Specifies whether to always use --forceoverwrite when doing force:source:push on save".  When ready, Sonal should review the final wording for the setting & description

lukecotter:
@brpowell This should be done now.
Let me know if there is anything else, needed.

packages/salesforcedx-vscode-core/src/commands/forceSourcePush.ts
lukecotter: done

packages/salesforcedx-vscode-core/package.json
lukecotter: I have used the suggested text and the setting is called `salesforcedx-vscode-core.push-or-deploy-on-save`
Let me know if this needs to change

packages/salesforcedx-vscode-core/package.json
```diff
@@ -837,6 +837,11 @@
       "type": "object",
       "title": "%feature_previews_title%",
       "properties": {
+        "salesforcedx-vscode-core.force-push-on-save": {
```
brpowell, line 840:
```suggestion
        "salesforcedx-vscode-core.push-or-deploy-on-save.overrideConflictsOnPush": {
```

Can we move this setting right below the `push-or-deploy-on-save.enabled` entry to group related setting?

@smaddox-sf @sbudhirajadoc what do you think of my suggested setting name (Override conflicts on push)? Also moved the setting into the `push-or-deploy-on-save` namespace.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -37,6 +37,7 @@
   "force_project_create_text": "SFDX: Create Project",
   "force_project_with_manifest_create_text": "SFDX: Create Project with Manifest",
   "force_apex_trigger_create_text": "SFDX: Create Apex Trigger",
+  "force_push_on_save_description": "Specifies whether to always use --forceoverwrite when doing force:source:push on save.",
```
brpowell, line 40: @sbudhirajadoc

brpowell, line 40:
```suggestion
  "override_conflicts_on_push_description": "Specifies whether to always use --forceoverwrite when doing force:source:push on save.",
```

lukecotter: done

packages/salesforcedx-vscode-core/package.json
lukecotter: done

smaddox-sf: Looks good on both.  @sbudhirajadoc will do a final review on the setting name & description

packages/salesforcedx-vscode-core/package.json
sbudhirajadoc: Slight modification -- "Specifies whether to always use --forceoverwrite when you run force:source:push on save"

packages/salesforcedx-vscode-core/package.nls.json
sbudhirajadoc: Slight modification -- "Specifies whether to always use --forceoverwrite when you run force:source:push on save"

packages/salesforcedx-vscode-core/package.json
sbudhirajadoc: added a small update on description. setting name is good.

packages/salesforcedx-vscode-core/package.nls.json
sbudhirajadoc: Slight modification -- "Specifies whether to always use --forceoverwrite when you run force:source:push on save"

packages/salesforcedx-vscode-core/package.json
```diff
@@ -837,6 +837,11 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "%force_push_on_save_description%"
```
brpowell, line 843:
```suggestion
          "description": "%override_conflicts_on_push_description%"
```

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -44,6 +44,7 @@
   "retrieve_test_code_coverage_text": "Specifies whether code coverage results are calculated and retrieved when you run Apex tests.",
   "telemetry_enabled_description": "Specifies whether to enable Salesforce telemetry. Even when enabled, still abides by the overall `telemetry.enableTelemetry` vscode setting.",
   "push_or_deploy_on_save_enabled_description": "Specifies whether or not to automatically run force:source:push (for source-tracked orgs) or force:source:deploy (for non-source-tracked orgs) when a local source file is saved.",
+  "override_conflicts_on_push_description": "Specifies whether to always use --forceoverwrite when doing force:source:push on save.",
```
brpowell, line 47:
```suggestion
  "override_conflicts_on_push_description": "Specifies whether to always use --forceoverwrite when you run force:source:push on save",
```

lukecotter:
@brpowell is this ok to merge now?

brpowell (approved):

PR #3158, opened 2021-04-21
fix: retrieving all components of org browser type

brpowell:
### What does this PR do?

Fixes an issue retrieving all components of a type using the org browser. If there was a mixed of locally present components and remote only components, only the locally present ones were retrieved.

@W-9102368@

### Functionality Before

Only locally present components were retrieved when retrieving all components of a type.

### Functionality After

Both remote components and locally present ones are retrieved.


sfsholden (approved):
Thank you for catching this!

lcampos (approved):

PR #3165, opened 2021-04-21
fix: disable lwc type generation

rcoringrato-sfdc:
### What does this PR do?

Disable LWC type generation in the core extension during sobject refresh.

@W-9168796@


lcampos (approved):

lcampos:
btw, this would need porting to `release/v51.10.0`

PR #3167, opened 2021-04-22
feat(apex): soql-in-apex support

dobladez:
### What does this PR do?
Adds support for special `_SOQL_` completion item which the Apex LSP server may return
when the cursor is within a SOQL block.  When Apex LSP server returns such item,
"forward" the code-completion request to be handled by the SOQL LSP.

For context: This is the Apex LSP client-side implementation of Approach 2.b described
in https://salesforce.quip.com/NZcqAcKLMCs4.

The server-side of this PR is at https://git.soma.salesforce.com/apex/apex-jorje/pull/800.

Pending:
 * Replace out/apex-jorje-lsp.jar with a final signed one (Apex LSP server PR [here](https://git.soma.salesforce.com/apex/apex-jorje/pull/800))
 * Snippets always show up, even on SOQL blocks. One possible solution: move them to Apex LSP server
 * Possible optimization: create the virtual document with only the SOQL block (and then translate
   Positions) instead of replacing everything around the block with blanks
 * Telemetry?

@W-9037497@

### Functionality Before
Editor completes with Apex items when cursor is inside a SOQL block

### Functionality After
Editor completes forwards completion request to SOQL LSP when cursor is within SOQL block of an Apex file


packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
```diff
@@ -0,0 +1,145 @@
...
+  const rangeAtCursor = document.getWordRangeAtPosition(
+    position,
+    /[:(_\.\w)]+/
+  );
+  const wordAtCursor = rangeAtCursor
+    ? document.getText(rangeAtCursor)
+    : undefined;
+
+  return !!wordAtCursor && wordAtCursor.startsWith(':');
+}
```
brpowell, line 60: What makes the VS Code context more likely to catch this scenario of whether or not the cursor is in a binding expression? Is it not possible for the extra check to be in the language server?

packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
```diff
@@ -0,0 +1,145 @@
...
+  );
+  return soqlCompletions || [];
+}
+
+function eolForDocument(doc: TextDocument) {
+  switch (doc.eol) {
+    case EndOfLine.LF:
+      return '\n';
+    case EndOfLine.CRLF:
+      return '\r\n';
```
brpowell, line 142: Unless the codecov report is wrong, might want to verify this case for windows

packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
dobladez: I agree that this belongs to the server side.... however, this is a "hack", which is much simpler to implement here. The real solution is to fix the start/end indexes bug on the language server side.

packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
dobladez: Both: codecov doesn't capture any of the code-completion tests. *And* I have not manually tested any of this on Windows, I'll try to find a volunteer :-)

packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
brpowell: Gotcha. Proper solution aside, I guess what makes the "hack" easier to do here than have the "hack" in the language server?

brpowell:
Thought I would ask - from stepping through I feel like this might be something to address in the soql language server rather than this middleware or the apex language server. Just wanted to confirm with you @dobladez in case we could catch it earlier:

https://user-images.githubusercontent.com/10779873/116909401-c7b2d080-ac09-11eb-9aa6-afecf2a3c148.mp4

packages/salesforcedx-vscode-apex/src/embeddedSoql/soqlCompletion.ts
brpowell: Cool cool. You can also probably have a test case that correctly builds a virtual document from a document with CRLF.

dobladez:
> Thought I would ask - from stepping through I feel like this might be something to address in the soql language server rather than this middleware or the apex language server. Just wanted to confirm with you @dobladez in case we could catch it earlier:
> 
>  soql_completion.mp4

Hmm... this is working for me. Just in case,  I've  just created a unit test on the LSP server side (Java) to produce your example on the video and it seems to work fine. Are you sure it is loading the right Apex LSP .jar file?

BTW: Among the improvements I thought of for the future was to provide SOQL completion also when inside empty brackets `[ ]`. Today it only assumes it's a SOQL block if it starts with `[SELECT `.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -148,20 +168,16 @@ export async function createLanguageServer(
...
-
-  client.onTelemetry(data =>
-    telemetryService.sendEventData('apexLSPLog', data.properties, data.measures)
-  );
-
-  return client;
+function isSOQLExtensionInstalled() {
+  const soqlExtensionName = 'salesforce.salesforcedx-vscode-soql';
+  const soqlExtension = vscode.extensions.getExtension(soqlExtensionName);
+  return soqlExtension !== undefined;
```
brpowell, line 182: Let's also check if the extension is enabled, since you can have it installed but disabled. I believe it's just `soqlExtension.isEnabled`.

dobladez: When the extension is disabled, `getExtension` returns `undefined`. There's no `isEnabled`, although there might be in the future: https://github.com/microsoft/vscode/issues/84997

I did try with `.isActive` like you suggested on the quip doc: unfortunately, the SOQL extension is not yet activated when this check runs.

brpowell: ah i see, thanks for clarifying.

brpowell (dismissed):

brpowell (approved):

PR #3173, opened 2021-04-23
chore: add version logic to create-release-branch flow for major version bumps

sfsholden:
### What does this PR do?
Adds the ability to change the type of version bump when creating a release branch.

@W-9177561@

### Functionality Before
Creating a release branch always resulted in a minor bump. There was not a way to use the script to bump the major or patch version. 

### Functionality After
Creating a release branch allows for flexibility when bumping the version.


.circleci/config.yml
```diff
@@ -596,18 +599,12 @@ workflows:
           message: Pending Approval for Creation of Release Branch
           channel: 'pdt_releases'
           color: '#0E1111'
-      - build-all
```
sfsholden, line 599: I decided to remove jobs that we weren't running as part of the scheduled build. I don't see these as being needed.

.vscode/tasks.json
```diff
@@ -87,7 +87,7 @@
         "${input:verboseOutput}",
         "-i",
         "${input:packagesToIgnore}",
-        "-r",
+        "-o",
```
sfsholden, line 90: renamed release override value to "o" so that "r" could be used for the version bump type. This way it matches what we have in other repos.

scripts/create-release-branch.js
```diff
@@ -7,14 +7,44 @@ const logger = require('./logger-util');
...
+      patch = 0;
+      break;
+    case 'patch':
+      patch = parseInt(patch) + 1;
+      break;
+  }
+  return `${major}.${minor}.${patch}`;
+}
 
+shell.env['SALESFORCEDX_VSCODE_VERSION'] = getReleaseVersion();
```
brpowell, line 47: Will this persist across jobs?

sfsholden: Yes, it should.

brpowell (approved):

PR #3180, opened 2021-04-27
feat(faux-generator): refactor generators. Add one for SOQL LSP

dobladez:
### What does this PR do?

Refactor faux-generator to accommodate new generator for SOQL LSP.

Summary of changes:
* Change `SObject to be a subset of jsforce's raw data (Pick<DescribeSObjectResult ...>).
* Remove most types from `types/describe.ts` as they are the same as those in `jsforce` lib.
* Change `SObjectDescribe` functions to convert the API responses to the (smaller) `SObject`. This reduces memory consumption, since we'll keep the complete sobject metadata in memory for as little as possible.
* Change `SObjectRefreshOutput` to use `SObject` instead of `SObjectDefinition`
* Adapt `OrgObjectDetailRetriever`: Remove call to `declGenerator.generateSObjectDefinition()`. Pass `SObject` straight as output.
* Adapt `FauxClassGenerator` and `TypingGenerator` to make the call to `declGenerator.generateSObjectDefinition()` internally, to convert the `SObject`s into `SObjectDefinition`.
* Re-create `minSObjects.json` to match the new `SObject` type format
* Adapt `MinObjectRetriever`
* Add new generator for SOQL LSP (`SOQLMetadataGenerator`)
* Add `scripts/regenerateMinsSObjects.ts` to generate `minSObjects.json` when needed

__IMPORTANT:__ file `minSObjects.json` must be re-generated from the right Org before merging.

@W-9037488@

### Functionality Before
N/A

### Functionality After
On generation (either from mins json or "Refresh SObjects" command) a new subdirectory gets created at `.sfdx/tools/soqlMetadata`  with one json file per SObject for SOQL LSP consumption.


packages/salesforcedx-sobjects-faux-generator/scripts/regenerateMinsSObjects.ts
```diff
@@ -0,0 +1,128 @@
...
+  const token = args[0];
+  const instanceUrl = args[1];
+
+  const connection = await createConnection(token, instanceUrl);
+
+  await generateLocalSobjectJSON(connection);
+})();
+
+async function createConnection(token: string, instanceUrl: string) {
+  return await Connection.create({
```
rcoringrato-sfdc, line 45: I prefer to use a pattern like:
    **const connection = (await Org.create({ aliasOrUsername })).getConnection();**

It's simpler, and you don't need to pass around a token you can just use a connection alias that you establish in your environment.

packages/salesforcedx-sobjects-faux-generator/src/generator/soqlMetadataGenerator.ts
```diff
@@ -0,0 +1,58 @@
...
+import { SObject, SObjectGenerator, SObjectRefreshOutput } from '../types';
+
+const REL_BASE_FOLDER = [TOOLS_DIR, SOQLMETADATA_DIR];
+
+export class SOQLMetadataGenerator implements SObjectGenerator {
+  public constructor() {}
+
+  public generate(output: SObjectRefreshOutput): void {
+    const outputFolderPath = path.join(output.sfdxPath, ...REL_BASE_FOLDER);
+    if (!this.resetOutputFolder(outputFolderPath)) {
```
rcoringrato-sfdc, line 22: This has the effect of clearing all previously generated json data, so if the user refreshes just custom objects it will remove data for all of the standard objects.

packages/salesforcedx-sobjects-faux-generator/scripts/regenerateMinsSObjects.ts
dobladez: 👍  much better now. Thanks!

packages/salesforcedx-sobjects-faux-generator/src/generator/soqlMetadataGenerator.ts
dobladez: Great point. I'll about how to best handle it

rcoringrato-sfdc (approved):

Issue #2089, opened 2020-03-20
add a parameter for `sfdx force:org:open` to indicate whether it is lightning or classic to open


garychenming:
As a SFDC developer, my development scenario was limited to classic, with the default lightning experience every time I opened org, and I had to switch to classic, which was time wasting when the network was bad.

So I hope there is a parameter to specify the experience mode of opening org. like this:

```
$: sfdx force:org:open -u orgName -l // open an org with lightning experience
$: sfdx force:org:open -u orgName -c // open an org with classic mode
```

jackTylerDurden:
What should be the default value of this parameter(i.e. if not specified)?

garychenming:
> What should be the default value of this parameter(i.e. if not specified)?

Lightning is the current trend, so I think the default value should be lightning.

jackTylerDurden:
Hi @garychenming,
Thanks for the quick reponse. 

I went through the documentation of Salesforce CLI Command Reference and found that there already exists a command which opens the org in lightning when a parameter is given.

The command I used is - 
sfdx force:org:open -u <username>  -p lightning

Please find the documentation [here](https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_org.htm#cli_reference_force_org). Please look for '  $ sfdx force:org:open -r -p lightning' once you go to the documentation.

Please let me know if I'm missing anything.

Thanks, and Regards,
Tanmay

gbockus-sf:
nope...long never'd. Thanks for the callout @AllanOricil

PR #3201, opened 2021-05-01
feat: add landing page setting

osieckiAdam:
### What does this PR do?

Enable the user to set default landing page which will be opened when the command "SFDX: Open Default Org" runs

#2089

### Functionality Before
When SFDX: Open Default Org command was used, it was automatically redirecting to lightning home page.

### Functionality After
User is able to define in settings which page will be opened after using SFDX: Open Default Org command. User may choose one of the predefined values Default, Home (Classic), Home (Lightning), Setup (Classic), Setup (Lightning) or define its own URL.
image


smaddox-sf:
Hi @osieckiAdam - I'm sorry it has taken so long to reply on this PR.  This one fell through the cracks.

Thank you for submitting this!  I'd like to make some tweaks to simplify a bit and match some of the terminology to the CLI for consistency.  Here's the changes needed for us to accept the PR:

Let's simplify this to two settings:

**Setting 1:**
- Name: Path for Org Open
- Description: Defines which page will open when using SFDX: Open Default Org. specify the portion of the URL after 'yourInstance.salesforce.com/'. 
- Default is no value, which will open the setup page
This setting is similar to the 'other' option in the 'Salesforce Landing Page' setting that you created.


**Setting 2:**
- Name: Open Org in Classic
- Description: Check this box to open the current org in Classic instead of Lightning
- Default is unchecked, which will open the org in Classic

For the first setting, the user would provide the same path that the CLI expects ([CLI reference documentation](https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_org.htm#cli_reference_force_org_open))

osieckiAdam:
@smaddox-sf no problem :)
This is a good idea to simplify this. I was thinking about exactly the same approach as You described in Your comment.  I came to a conclusion that I would be not sure what should be the behaviour when the checkbox for setting 2 is checked. There is no easy way to 'translate' all links from Lightning to Classic. For example: if I want to open the User page in setup, for LE I need to use the parameter:

```lightning/setup/ManageUsers/home```

But for Classic I need to go with a small workaround. I need to use 005 with at least one parameter to not get redirected to Lightning Page:

```005?isUserEntityOverride=1```

If I had a checkbox which You described, I would probably expect that if I put the lightning/setup/ManageUsers/home in the first setting then after checking this checkbox the expected result will be that the User page is opened in classic. This is why I think the checkbox could be misleading. 

The best option that I came up with is to give few standard options which in my opinion could be used in most cases and let users experiment with custom URLs. This is exactly the same approach as in Chrome extension ORGanizer for Salesforce created by @enreeco so I thought some users could be already familiar with that behaviour.  

The final decision is obviously yours, once you will specify what is the expected behaviour when checkbox is checked I would be very happy to change the implementation. I would also add text 'This option requires restart to take effect.' to descriptions for both options.

smaddox-sf:
Hi @osieckiAdam - Yeah, you're right that trying to translate the classic vs. lightning paths adds complexity and could result in some unexpected results.  We tested some other permutations and believe we can simplify to one setting (essentially Setting #1 that I described above).  The path provided would dictate whether lightning or classic is opened.  Examples:
 - lighting - Opens the default home page with lightning
 - lighting/customURL - Opens the page dictated by the customURL in lightning
 - customURL - Opens the page dictated by the customURL in Classic

What the user supplies for this setting would be identical as what is supplied for the `-p` (path) argument of the org:open CLI command.  If nothing is specified in the setting, current behavior will continue.

osieckiAdam:
Ok, all clear, let me implement the changes and update the PR.
It's a pity we are getting rid of predefined values but I see that it's a great idea to make this as simple as possible.  
I guess that I will be able to make changes on Saturday/Sunday

gbockus-sf:
Closing due to inactivity.

PR #3206, opened 2021-05-04
tests: relaunch crashed VSCode Integration Tests

sfsholden:
### What does this PR do?
Relaunches VSCode Integration Tests that have crashed. If the resulting junit result file is not present, this is what would constitute a rerun. Note that this does not rerun tests that have failed and have results in the junit result file.

This PR should hopefully cut down on the number of integration test failures due to:
```
Missing junit results for the following packages:

* vscode-integration:
	- <insert vscode package name>
```

Note that to see an example of what this looks like, refer to [Assert all tests ran and aggregate results](https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/4634/workflows/15eeecd3-3d05-437c-bf84-d0ebd5d81f5e/jobs/20430).

@W-9138899@


.circleci/config.yml
```diff
@@ -357,10 +357,6 @@ jobs:
       - SFDX_AUTOUPDATE_DISABLE: true
       - SFDX_DOMAIN_RETRY: << pipeline.parameters.domain-retry >>
       - NODE_VERSION: << pipeline.parameters.node-version >>
-    parameters:
```
sfsholden, line 360: Dead code that is no longer used.

scripts/aggregate-junit-xml.js
```diff
@@ -23,96 +23,166 @@
...
   integration: [],
   unit: [],
   system: []
 };
 
-for (const entry of fs.readdirSync(packagesDir)) {
-  const packagePath = path.join(packagesDir, entry);
+function getTestCategories() {
+  let flags;
+  if (process.argv.indexOf(TEST_TYPE_ARG) > -1) {
```
sfsholden, line 57: Explaining why I'm keeping this:

Originally I needed to make this change because I had another flag that I was using as well. This other flag was going to be used to determine when we wanted the script to return the failed package list, but not throw an exit code. Originally I was going to relaunch the tests outside of this script.

I opted to go with the simpler approach of just relaunching the tests from within this script itself.

scripts/aggregate-junit-xml.js
```diff
@@ -23,96 +23,160 @@
  * By default, all categories will be tested. To only process specific
  * categories, call the script with desired categories separated by a space.
  *
- * e.g. node aggregate-junit-xml.js integration vscode-integration
+ * e.g. 
+ * node aggregate-junit-xml.js -t integration vscode-integration
+ * npm run aggregateJUnit -- -t integration vscode-integration
```
brpowell, line 28: Can we add in this string that it will attempt to rerun tests for packages where no results were found one more time?

brpowell (approved):

sfsholden:
> Based on your research, we can also probably remove the code_version flag in the CI variables. That was set to see if there was a particular version of VS Code causing problems but doesn't seem to be the case.

I think we should keep it. It's a good thing to be able to adjust on the fly. We might have a time in the future where this will be nice to have.

We could adjust how it is being set though. Right now since it's an environment variable it's hard to know for sure which version it is using. But we could change it to a pipeline configuration in circleci. This would let us see what that value is since it won't be masked. Going with that approach would also mean you could easily change the version of VSCode for your own workflow without affecting others'.

Issue #3151, opened 2021-04-14
API version in manifest or sfdx-project.json has no impact on retrieve and org explorer


fishygeek91:
### Summary

When retrieving object metadata via the Org Explorer or right clicking on a file, the returned results differ from using force:source:retrieve -x “path/to/package.xml”. This is causing issues for developers who are diffing meta files from production with source control, and is dependent on the way the source is retrieved. 

### Steps To Reproduce:

1. Used force:source:retrieve -x “path/to/package.xml” to pull test object
2. Next I used the Org Explorer to pull down same object
3. Diff'ed meta. Metadata for object differs
4. Changed API version is package.xml
5. Used force:source:retrieve -x “path/to/package.xml” and source differs from step 1
6. Used Org explorer and received same metadata as step 2.

### Expected result

force:source:retrieve -x “path/to/package.xml” returns same metadata as Org Explorer or right clicking on file to retrieve source

### Actual result

force:source:retrieve -x “path/to/package.xml” returns different metadata as Org Explorer or right clicking on file to retrieve source

### Additional information

[https://github.com/forcedotcom/cli/issues/528](url)

_Feel free to attach a screenshot_.

**VS Code Version**:

**SFDX CLI Version**:

**OS and version**:

fishygeek91:
![Screen Shot 2021-04-22 at 9 00 33 AM](https://user-images.githubusercontent.com/62761381/115738610-6f96e700-a34a-11eb-8c53-a4b4831c797a.png)

Unchecking this box seems to have fixed the issue for me. At least for a test object I have been working with to come up with a solution for our source control.

smaddox-sf:
Hi @fishygeek91 - Thanks for reaching out.  We are working on a fix for this.  You are correct on the workaround to uncheck that setting for now.  With the setting unchecked, these commands use the CLI directly for these commands vs. the new library for deploy/retrieve.  You will see better performance with the library, so I suggest turning this setting back on once we have this fixed.  We'll update the issue once we merge a fix.

ChuckJonas:
@smaddox-sf could you give us a sense of the timeline for this fix?

We are trying to decide if we should invest in upgrading all of our projects to use the Experimental flag or just waiting for the fix to come out.

Thanks

brpowell:
[...]

PR #3208, opened 2021-05-04
fix: respect sfdx api version override during new deploy/retrieve

brpowell:
### What does this PR do?

Respects the overridden API version in an SFDX project when using the new deploy/retrieve logic. Note that deploying/retrieving with a manifest will always honor the version specified in the manifest - this is the CLI behavior.

#3151, @W-9179305@

### Functionality Before

New deploy/retrieve logic always used the latest API version when using Org Browser or source paths.

### Functionality After

Now it only uses the latest if a version isn't overridden through SFDX configuration or with a manifest file's version.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -20,6 +20,7 @@ import {
   DeployResult,
   MetadataApiDeploy,
   MetadataApiRetrieve,
+  registry,
   RetrieveResult
 } from '@salesforce/source-deploy-retrieve';
 import {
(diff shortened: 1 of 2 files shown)


AnanyaJha (approved):

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -62,6 +64,16 @@ export abstract class DeployRetrieveExecutor<
     try {
       const components = await this.getComponents(response);
 
+      // concrete classes may have purposefully changed the api version.
```
sfsholden, line 67: I find this documentation a bit confusing. When you say concrete classes - what exactly are you referring to? Is this in reference to the manifest file?

brpowell: By concrete classes I'm referring to implementations of the abstract class. So if getComponents changes the api version on the component set in the child class, this code respects that. Particularly this is for the manifest-based classes, but if anything extends this base class and manually sets the api version this will work.

sfsholden (approved):

fishygeek91:
Hi @brpowell. I finally got around to testing these changes, and it appears that this only works if the experimental checkbox "Use library for deploy and retrieve commands instead of the CLI". Without this checked, the same version of the SObject metadata is retrieved (most recent?) no matter where I change the apiVersion. I have changed the apiVersion in package.xml, sfdx-project.json, and using sfdx config:set apiVersion. None of those make a difference in the metadata retrieved, even when I set the apiVersion to 45 (down from 51). I also went as far as testing this in both the stable version of VSCode and Insiders and got the same results. 

I am trying to sync up our CI with the meta we are pulling from VSCode as we are developing. So far it seems that if we want this, then we have to regularly update (or automate) our apiVersion in the package.xml (manifest) used in the project.

PR #3213, opened 2021-05-05
chore: add change case management

sfsholden:
### What does this PR do?
Adds change case management to VSCode.

@W-9230973@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.2.0
-  release-management: salesforce/npm-release-management@4.12.1
+  release-management: salesforce/npm-release-management@4
   slack: circleci/slack@3.4.2
 
 parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,9 +49,6 @@ parameters:
   pre-publish:
     type: boolean
     default: false
-  scheduled-build:
-    type: string
-    default: ''
 
 commands:
   build-artifacts:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -49,6 +49,9 @@ parameters:
   pre-publish:
     type: boolean
     default: false
+  scheduled-build:
```
sfsholden, line 52: TODO - set this value automatically to match whatever is in the package.json. I don't think we will ever have a time where this value will be different.

sfsholden: Redoing this piece - just realized this isn't needed.

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.2.0
-  # sfchangecase: salesforce/change-case-management@3
+  release-management: salesforce/npm-release-management@4.12.1
```
sfsholden, line 14: Just like with source-deploy-retrieve, pin this version of npm-release-management in case there are breaking changes in one of the patch updates.

brpowell: Is it reasonable to expect that minor/patch won't have breaking changes?

sfsholden: No, in the past when we started to use the CLI's orb we had some incidents where jobs/commands were renamed (as part of a minor or patch version bump) that broke us. This blocked us from being able to release.

I'm happy to change this though - it could be a good learning experience for everyone on just how important correct semver is.

brpowell: Gotcha. Let's leave it at `@4` since the documentation suggests that. If breaking changes happen again we can bring it up since that seems unintentional.

brpowell (approved):

PR #3240, opened 2021-05-14
feat: single file diff

xyc:
### What does this PR do?
Single file diff using cache service lib

@W-9166550@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -187,15 +187,7 @@
...
-          "when": "sfdx:project_opened && resourceFilename =~ /.*(app|cmp|auradoc|evt|intf|css|design|html|js|svg|permissionset-meta.xml|layout-meta.xml|labels-meta.xml)$/"
-        },
-        {
-          "command": "sfdx.force.diff",
-          "when": "sfdx:project_opened && resourceLangId == apex"
-        },
-        {
-          "command": "sfdx.force.diff",
-          "when": "sfdx:project_opened && resourceLangId == visualforce"
+          "when": "sfdx:project_opened && resourceLangId != 'forcesourcemanifest'"
```
jag-sfdc, line 190: Why are we making this change? Just a cleanup or is there any benefit to this?

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
```diff
@@ -5,130 +5,88 @@
...
-      await handleDiffResponse(exitCode, stdOut);
-    });
+/**
+ * Perform file diff and execute VS Code diff comand to show in UI.
+ * It matches the correspondent file in compoennt.
+ * @param localFile local file
+ * @param remoteComponent remote source component
+ * @returns {Promise<void>}
+ */
+async function diffFile(
```
jag-sfdc, line 31: I refactored this to conflictDetectionService to keep things cleaner here. Let me know if that sounds good once we merge ours.

packages/salesforcedx-vscode-core/package.json
xyc: MetadataCacheService allows us to diff more file types, example: `js-meta.xml`

packages/salesforcedx-vscode-core/src/commands/forceSourceDiff.ts
xyc: yeah, that looks good

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
```diff
@@ -495,6 +495,8 @@ export const messages = {
   conflict_detect_diff_command_title: 'ファイルを比較',
 
   force_source_diff_text: 'SFDX: 組織のファイルとの差分を表示',
+  force_source_diff_remote_not_found:
+    'Selected file or directory was not found in org',
```
jag-sfdc, line 499: Can you include Sonal and get her approval for this. Rest looks good.

AnanyaJha:
If a user doesn't have a file in their org, we're currently displaying two error notifications: 
Screen Shot 2021-05-17 at 5 01 10 PM

Do we need to display two? Is there a way we could consolidate the info to one pop up for the user?

packages/salesforcedx-vscode-core/package.json
AnanyaJha: This is going to cause the `Diff` command to be exposed for across most contexts ie. a user's `.forceignore`, `.gitignore`, the `manifest` folder etc. Is there anyway to keep this more selective, ie. to just files/directories in the project's source folders?

xyc:
@AnanyaJha this is the current pattern of error messages. The text notification_unsuccessful_execution_text doesn't allow us to customize with additional messaging.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -561,6 +561,8 @@ export const messages = {
   conflict_detect_diff_command_title: 'Compare Files',
 
   force_source_diff_text: 'SFDX: Diff File Against Org',
+  force_source_diff_remote_not_found:
+    'Selected file or directory was not found in org',
```
xyc, line 565: @sbudhirajadoc Could I have a review on the text? This is shown when the remote file diffed is not found.

AnanyaJha:
> @AnanyaJha this is the current pattern of error messages. The text notification_unsuccessful_execution_text doesn't allow us to customize with additional messaging.

Ah I see, in that case maybe we should just log the message in the output window so we follow the pattern we're using for other commands? Included an example of an error while deploying 

Screen Shot 2021-05-17 at 5 25 04 PM

packages/salesforcedx-vscode-core/package.json
xyc: not really, the source folders can be arbitrary. Also we are keeping this consistent with sdr commands. (Deploy/Retreive commands are also available for the aforementioned files)

xyc:
I'd argue that is inferior UX since it shows the error far away & it takes much more effort for users to locate the root cause. Consolidate messages is not a bad idea, but that would take a much larger refactoring.

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
sbudhirajadoc: looks good to me.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sbudhirajadoc: change to 'Selected file or directory was not found in the org'

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
sbudhirajadoc: change to 'Selected file or directory was not found in the org'

sbudhirajadoc (commented):
minor changes in text.

rcoringrato-sfdc (approved):

PR #3246, opened 2021-05-18
feat: use SDR for conflict detect perf improvements

rcoringrato-sfdc:
### What does this PR do?

- Use SDR library to improve performance of conflict detection.
- Add manifest support for caching project metadata

@W-9164033@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
+++ b/packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
@@ -36,17 +36,17 @@ export class ConflictDetector {
   private diffs: DirectoryDiffResults;
   private error?: Error;
   private static instance: ConflictDetector;
-  private static EMPTY_DIFFS = Object.freeze({
+  private static EMPTY_DIFFS = {
     localRoot: '',
     remoteRoot: '',
     different: new Set<string>(),
     scannedLocal: 0,
     scannedRemote: 0
-  });
+  };
 
   constructor(differ?: DirectoryDiffer) {
     this.differ = differ || new CommonDirDirectoryDiffer();
-    this.diffs = ConflictDetector.EMPTY_DIFFS;
+    this.diffs = Object.assign({}, ConflictDetector.EMPTY_DIFFS);
   }
 
   public static getInstance(): ConflictDetector {
@@ -62,7 +62,7 @@ export class ConflictDetector {
     data: ConflictDetectionConfig
   ): Promise<DirectoryDiffResults> {
     const startTime = process.hrtime();
-    this.diffs = ConflictDetector.EMPTY_DIFFS;
+    this.diffs = Object.assign({}, ConflictDetector.EMPTY_DIFFS);
     this.error = undefined;
 
     try {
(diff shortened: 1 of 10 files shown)


packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -57,33 +65,48 @@ export class MetadataCacheService {
...
-  public getSourceComponents(): ComponentSet {
-    return this.componentPath
-      ? (this.sourceComponents = ComponentSet.fromSource(this.componentPath))
-      : new ComponentSet();
+  public async getSourceComponents(): Promise<ComponentSet> {
+    if (this.componentPath && this.projectPath) {
+      this.sourceComponents = this.isManifest
+        ? await ComponentSet.fromManifest({
+            manifestPath: this.componentPath,
+            resolveSourcePaths: [this.projectPath]
```
brpowell, line 104: I think we need to make this a list of the project's package directories. If `projectPath` is the root of the project, it'll still work but will end up over-scanning irrelevant files like node_modules which could impact the performance

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -247,21 +279,28 @@ export type MetadataCacheCallback = (
...
 
   protected async getComponents(response: any): Promise<ComponentSet> {
-    this.cacheService.initialize(response.data, getRootWorkspacePath());
-    return this.cacheService.getSourceComponents();
+    this.cacheService.initialize(
+      response.data,
+      getRootWorkspacePath(),
+      this.isManifest
+    );
+    return await this.cacheService.getSourceComponents();
```
brpowell, line 303: Nit, remove `await` since we're already returning a promise

packages/salesforcedx-vscode-core/src/conflict/conflictDetectionService.ts
```diff
@@ -57,189 +61,79 @@ export class ConflictDetector {
...
 
-  private clearDirectory(dirToRemove: string, throwErrorOnFailure: boolean) {
-    try {
-      shell.rm('-rf', dirToRemove);
-    } catch (error) {
-      this.reportError('error_cleanup_temp_files', error);
-      if (throwErrorOnFailure) {
-        throw error;
-      }
+  private async handleCacheResults(
```
jag-sfdc, line 114: Where is this handleCacheResults used and how is this different from the one in forceSourceDiff.ts ?

jag-sfdc: Never mind, so this is specifically used for the conflict detection part here. I think we need a separate diffService.ts and move some of the diff specific code there. Can be part of a separate devchoice/refactoring story.

rcoringrato-sfdc: Agreed. There is some overlap, but they will diverge as conflict detection progresses.

AnanyaJha (approved):

PR #3247, opened 2021-05-18
feat: use new cli for function creation + java support @W-9282205@

jag-sfdc:
### What does this PR do?
Uses new cli for function creation - either JS or Java.
Remove support for typescript

@W-9282205@
@W-9282425@

### Functionality Before
N/A

### Functionality After
Screen Shot 2021-05-18 at 11 51 10 AM
Screen Shot 2021-05-18 at 11 51 31 AM

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
@@ -33,6 +33,15 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
 > {
   private metadataType?: MetadataInfo;
 
+  constructor(type?: string) {
+    super();
+    const info = MetadataDictionary.getInfo(type!);
+    if (!info) {
+      throw new Error(`Unrecognized metadata type ${type}`);
+    }
+    this.metadataType = info;
+  }
+
   public execute(response: ContinueResponse<DirFileNameSelection>): void {
     const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
@@ -91,26 +100,22 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
   }
 
   public getSourcePathStrategy(): SourcePathStrategy {
-    // @ts-ignore
-    return this.metadataType.pathStrategy;
+    return this.metadataType!.pathStrategy;
   }
 
   public getFileExtension(): string {
-    // @ts-ignore
-    return `.${this.metadataType.suffix}`;
+    return `.${this.metadataType!.suffix}`;
   }
 
   public setFileExtension(extension: string): void {
-    // @ts-ignore
-    this.metadataType.suffix = extension;
+    this.metadataType!.suffix = extension;
   }
 
   public getDefaultDirectory(): string {
-    // @ts-ignore
-    return this.metadataType.directory;
+    return this.metadataType!.directory;
   }
 
-  public setMetadataType(type: string): void {
+  public set metadata(type: string) {
     const info = MetadataDictionary.getInfo(type);
     if (!info) {
       throw new Error(`Unrecognized metadata type ${type}`);
--- a/packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
@@ -33,15 +33,6 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
 > {
   private metadataType?: MetadataInfo;
 
-  constructor(type?: string) {
-    super();
-    const info = MetadataDictionary.getInfo(type!);
-    if (!info) {
-      throw new Error(`Unrecognized metadata type ${type}`);
-    }
-    this.metadataType = info;
-  }
-
   public execute(response: ContinueResponse<DirFileNameSelection>): void {
     const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -90,7 +81,7 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
       : 'customDir';
   }
 
-  private getPathToSource(outputDir: string, fileName: string): string {
+  public getPathToSource(outputDir: string, fileName: string): string {
```
xyc, line 84: is this necessary?

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -100,18 +91,30 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
   }
 
   public getSourcePathStrategy(): SourcePathStrategy {
+    // @ts-ignore
```
xyc, line 94: why the `ts-ignore` here?

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
jag-sfdc: I did this to test something. I will revert it.

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
jag-sfdc: since metadataType is not assigned in the constructor anymore.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -26,23 +26,25 @@ import {
...
 } from '../util';
 import { BaseTemplateCommand } from './baseTemplateCommand';
-import { FUNCTION_TYPE } from './metadataTypeConstants';
+import { FUNCTION_TYPE_JAVA, FUNCTION_TYPE_JS } from './metadataTypeConstants';
 
 export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
-  constructor() {
-    super(FUNCTION_TYPE);
-  }
+  // constructor() {
```
jag-sfdc, line 32: I'll remove this

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -100,18 +91,30 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
...
     this.metadataType.suffix = extension;
   }
 
   public getDefaultDirectory(): string {
+    // @ts-ignore
     return this.metadataType.directory;
   }
+
+  public setMetadataType(type: string): void {
```
xyc, line 113: we can use javascript's native getter here. `public set metadataType`

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
xyc: `this.metadataType!.pathStrategy` could work around this. 
But the real solution is, we should use a getter for metadata type (so that we don't have to repeat the try/catch blocks) & catch exception (programmer error most likely) if metadataType is undefined.

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -30,17 +30,8 @@ import {
 
 export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
   DirFileNameSelection
-  > {
-  private metadataType: MetadataInfo;
-
-  constructor(type: string) {
```
xyc, line 36: we can keep the constructor (make type optional here) but your call. this is only for CLI templates and should retire gradually.

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
jag-sfdc: I'll use the "!". Not sure if adding a getter is going to provide much help since metadataType will be undefined only due to a programmer error and tests should right away catch it as it did for forceLightningLwcTestCreate.ts when I made the change.

xyc: yeah, the only thing is if it's missed by the test it will get an error like cannot get suffix of undefined. If we don't expect this class to be used a lot since we are moving to libraries, it's fine using `!`.

xyc (approved):

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -26,23 +26,25 @@ import {
   SfdxWorkspaceChecker
 } from '../util';
 import { BaseTemplateCommand } from './baseTemplateCommand';
-import { FUNCTION_TYPE } from './metadataTypeConstants';
+import { FUNCTION_TYPE_JAVA, FUNCTION_TYPE_JS } from './metadataTypeConstants';
 
 export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
   constructor() {
```
xyc, line 32: nit: can remove this constructor

xyc:
looks good w/ a minor comment

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Took care of this

PR #3253, opened 2021-05-20
chore: remove function project template @W-9282325@

jag-sfdc:
### What does this PR do?
remove function project template as an option when user creates a new project

@W-9282325@

### Functionality Before
Screen Shot 2021-05-20 at 1 24 06 PM

### Functionality After
Screen Shot 2021-05-20 at 1 25 03 PM


packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts
```diff
@@ -127,14 +127,6 @@ export class SelectProjectTemplate
         'force_project_create_analytics_template'
       )
     ];
-    if (sfdxCoreSettings.getFunctionsEnabled()) {
```
xyc, line 130: don't we need to remove `projectTemplateEnum.functions`?

AnanyaJha: +1, I think we'll also need to remove [this reference](https://github.com/forcedotcom/salesforcedx-vscode/blob/0289155546ba1d3923a99e07189bef8797d5f2d0/packages/salesforcedx-vscode-core/src/commands/forceProjectCreate.ts#L143)

jag-sfdc: Good call! Will take care of these.

xyc (approved):

PR #3262, opened 2021-05-24
fix: pull func dependencies by default + java/maven support @W-9282387@

jag-sfdc:
### What does this PR do?
Remove the user setting to opt-out of pulling function dependencies. There is no option for users to turn off this feature.
Do a maven run at the end of java function creation

@W-9282387@

### Functionality Before
N/A

### Functionality After
Screen Shot 2021-05-24 at 3 01 01 PM


packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -585,6 +585,8 @@ export const messages = {
     'Installing NPM dependencies',
   force_function_install_npm_dependencies_error:
     "%s. Make sure you have NodeJS installed (https://nodejs.org/) and then run 'npm install' to install dependencies from package.json",
+  force_function_install_mvn_dependencies_error:
+    "%s. Make sure you have Maven installed (https://maven.apache.org/) and then run 'mvn install' to install dependencies from pom.xml",
```
jag-sfdc, line 589: FYI @sbudhirajadoc

sbudhirajadoc: Looks good!

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -47,33 +50,42 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
...
-              resolve();
-            });
-          });
-        }
-      );
+  public runPostCommandTasks(data: FunctionInfo) {
+    const language = data.language;
+    if (language === LANGUAGE_JAVA) {
+      const pathToSource = this.getPathToSource(data.outputdir, data.fileName);
+      const targetDir = path.join(path.dirname(pathToSource), '..', '..', '..', '..', '..');
```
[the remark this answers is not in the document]
jag-sfdc: This is where I spent most time for this story. Yes, for java it is consistently 5 directories up. The alternative approach is setting java path strategy to be at this top level and we navigate 5 levels deep to open the java file at the time of creation.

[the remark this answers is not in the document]
jag-sfdc: Btw, 'Should pull dependencies for java' test will be the pinning test for this behavior. If in future Functions team change this directory structure and we change the pathstrategy, this test would fail.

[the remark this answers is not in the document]
xyc: can we add a comment on the test so when it breaks we know what could go wrong?

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -81,7 +81,7 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
       : 'customDir';
   }
 
-  private getPathToSource(outputDir: string, fileName: string): string {
+  public getPathToSource(outputDir: string, fileName: string): string {
```
xyc, line 84: does this have to be public?

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -47,33 +50,42 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
...
+            notificationService.showWarningMessage(
+              nls.localize(
+                'force_function_install_mvn_dependencies_error',
+                err.message
+              )
+            );
+            reject(err);
+          }
+          resolve();
+        });
```
brpowell, line 70: Is it possible to stream the output to the output channel? I couldn't tell if anything was happening until it finally failed. I did have `mvn` installed and it wasn't until I switched to the function directory and ran it until i knew what the error was. I'm also not sure when the dependencies actually finish installing

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
jag-sfdc: Changed to protected

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: One of the acceptance criteria Stephanie wanted in this story was to "Remove all the user notifications we show at the time of dependency installation." So we removed all the notifications and make the dependencies installation as silent as possible.

brpowell (approved):

PR #3271, opened 2021-05-26
feat: cache for FileProperties using Memento

maxcamp:
### What does this PR do?
Draft of a wrapper for Memento that can serve as a cache for conflict detection.

@W-9258968@

### Functionality Before
N/A

### Functionality After
This class exposes 4 public methods.  The first is initialize() which is called upon extension activation (the extension context must be passed in in order to use Memento). The second is getInstance() because I thought that a singleton pattern made sense. The third and fourth take in an array of FileProperties and get/set using the file names as the keys.


packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private getPropertiesForFile(fileName: string): ConflictFileProperties | undefined {
+    return this.storage.get<ConflictFileProperties>(fileName);
+  }
+
+  private setPropertiesForFile(fileName: string, value: ConflictFileProperties) {
+    this.storage.update(fileName, value);
+  }
+
+  public getPropertiesForFiles(fileProperties: FileProperties[]): LabelledConflictFileProperties[] {
```
jag-sfdc, line 49: This looks good. I'd suggest we leave out these multiple getters and setters. We may or may not need them based on how we use this at the time of retrieval or deploy. Let's revisit at that point.

jag-sfdc (approved):

PR #3273, opened 2021-05-27
feat: add Java support for Functions debugging

rcoringrato-sfdc:
### What does this PR do?

- Allow use of Debug Invoke feature for Java Functions

@W-9330349@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
@@ -76,6 +76,9 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
   }
 
   public execute(response: ContinueResponse<string>) {
+    const runtimeDetection: RegExp = new RegExp(
+      '.*Installing (.*) function runtime.*'
+    );
     const startTime = process.hrtime();
     const cancellationTokenSource = new vscode.CancellationTokenSource();
     const cancellationToken = cancellationTokenSource.token;
@@ -121,6 +124,7 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
         rootDir: functionDirPath,
         port: FUNCTION_DEFAULT_PORT,
         debugPort: FUNCTION_DEFAULT_DEBUG_PORT,
+        debugType: 'node',
         terminate: () => {
           return execution.killExecution('SIGTERM');
         }
@@ -143,7 +147,10 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
     );
 
     execution.stdoutSubject.subscribe(data => {
-      if (data.toString().includes('Debugger running on port')) {
+      const matches = String(data).match(runtimeDetection);
+      if (matches && matches.length > 1) {
+        FunctionService.instance.updateFunction(functionDirPath, matches[1]);
+      } else if (data.toString().includes('Debugger running on port')) {
         progress.complete();
         taskViewService.removeTask(task);
         notificationService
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -76,6 +76,9 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
   }
 
   public execute(response: ContinueResponse<string>) {
+    const runtimeDetection: RegExp = new RegExp(
```
xyc, line 79: nit: upper case and move out of the class / make static since this is a constant

xyc: also in the variable name describe we are detecting debug type here

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -143,7 +147,10 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
     );
 
     execution.stdoutSubject.subscribe(data => {
-      if (data.toString().includes('Debugger running on port')) {
+      const matches = String(data).match(runtimeDetection);
```
jag-sfdc, line 150: Curious if this is the only option to find the function type? Any information about the language or the directory structure at the time of registering the function we could use instead?
I'm just trying to avoid the need to maintain a regex for any future changes in the output.

rcoringrato-sfdc: That was my first approach, but I decide to change it. Since the build-pack is creating the environment seems like it is best suited to determine the runtime/debug settings.

jag-sfdc: In that case, can we add an else condition for the line below where we check for matches and give a user warning that reads something like "Language could not be identified. 'Debug Invoke' may not work".

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -143,7 +147,10 @@ export class ForceFunctionStartExecutor extends SfdxCommandletExecutor<string> {
...
+      if (matches && matches.length > 1) {
+        FunctionService.instance.updateFunction(functionDirPath, matches[1]);
+      } else if (data.toString().includes('Debugger running on port')) {
```
jag-sfdc, line 153: why is here an else condition? Don't we want to close the notification irrespective of a language match or not.

rcoringrato-sfdc: The language match occurs fairly early, whereas the "running" message is one of the last messages.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
rcoringrato-sfdc: I'll see if I can move that type of logic to the 'Debug Invoke' - it should not be visible/enabled if the language hasn't been determined, or at least produce the warning you mentioned.

jag-sfdc: Also looks like there is enough business logic now to add a separate section of tests in forceFunctionStart.test.ts

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
jag-sfdc: Synced offline and now understand data is one line of stdout at a time. So, this is fine.

jag-sfdc (approved):

PR #3277, opened 2021-06-02
fix: invoke works when there are spaces in the path @W-9282250@

jag-sfdc:
### What does this PR do?
Invoke works when directory path to payload file has spaces

@W-9282250@

### Functionality Before
Invoke would fail when payload file is in a directory path with spaces

### Functionality After
Invoke passes when payload file is in a directory path with spaces


rcoringrato-sfdc (approved):

brpowell:
I'm actually observing the opposite behavior, where this change causes an error and not having quotes works okay. My payload path is `/Users/b.powell/Desktop/My Project/functions/test/payload.json`

jag-sfdc:
I was able to repro what Bryan saw. It works in CLI, but not through VSCode. I'll close this PR.

PR #3282, opened 2021-06-03
Use @heroku/functions-core library implementation of 'Start Function' and 'Run Function'

YousefHaggyHeroku:
Things I need help with:
- Rebase commits to match repository guidelines
- Testing function cancellation - can we / do we need to do this with the Library executor?

### What does this PR do?
This PR switches the Function Invoke and Function Start commands to using the `functions-core` helper library directly instead of interfacing with the cli. Ditching the CLI decreases the execution time of those commands significantly

The CLI also now uses the `functions-core` shared helper library; that change has been deployed to production. https://github.com/salesforcecli/plugin-functions/pull/65

**FAQ**
**Q**: Where did all the metric logging go?
**A**: I believe `LibraryCommandletExecutor` does a good job at handling the various metric types in a uniform manner. However, if there's a metric that I've overlooked please let me know!

**Q**:


packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart.test.ts
```diff
@@ -430,46 +327,6 @@ describe('Force Function Start', () => {
         nls.localize('force_function_start_no_org_auth')
       );
     });
-
-    it('Should handle cancellation requested', async () => {
```
YousefHaggyHeroku, line 434: I couldn't figure out how to mock and trigger cancellation for the progress bar provided by `LibraryCommandletExecutor`

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -177,6 +177,13 @@ export abstract class LibraryCommandletExecutor<T>
           token.onCancellationRequested(() => {
             this.cancelled = true;
             notificationService.showCanceledExecution(this.executionName);
+
+            telemetryService.sendCommandEvent(
```
jag-sfdc, line 181: @YousefHaggyHeroku - Can you please explain this change? FYI @randi274

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
```diff
@@ -8,31 +8,60 @@
...
+import { LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode/out/src';
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+export class ForceFunctionInvoke extends LibraryCommandletExecutor<string> {
+  constructor(debug: boolean = false) {
+    super(
+      nls.localize('force_function_invoke_text'),
+      debug ? 'force_function_debug_invoke' : 'force_function_invoke',
+      OUTPUT_CHANNEL
+    );
+    if (debug) {
```
jag-sfdc, line 34: @YousefHaggyHeroku - Not sure if I understand this. We need language based telemetry for both debug and regular invokes. What does this do?

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -135,11 +134,13 @@ export abstract class LibraryCommandletExecutor<T>
   constructor(
     executionName: string,
     logName: string,
-    outputChannel: vscode.OutputChannel
+    outputChannel: vscode.OutputChannel,
+    cancellable: boolean = false
```
xyc, line 138: You can add `protected cancellable = true`; in the subclass instead of introducing this change on the base class .

xyc: TypeScript helps but still `super(..., true)` isn't as readable.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
YousefHaggyHeroku: Good question, previously we had a `force_function_start_cancelled` metric logging [in the actual `forceFunctionStart code`](https://github.com/forcedotcom/salesforcedx-vscode/blob/158d4794e1a9498c13508b21809848c953fe7588/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts#L107-L111). Since the cancellation request is handled by the library executor now, I figured it might be worth moving that metric here in case other long running Library based-commands exist in the future.

If there's a downside to this, I can move it to `token?.onCancellationRequested` in `forceFunctionStart` instead

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
YousefHaggyHeroku: Great suggestion, thanks! I agree the `super(..., true)` could be a bit hard to read

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
YousefHaggyHeroku: I'm not entirely sure, but right now in forceFunctionInvoke we only use language based telemetry on debug invokes: https://github.com/forcedotcom/salesforcedx-vscode/blob/158d4794e1a9498c13508b21809848c953fe7588/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts#L72-L75

I tried to adapt closely to what already existed. I *think* we only log language based telemetry in the debug invoke because we don't assume there are any `FunctionService.startedExecutions` with the regular invoke. The regular invoke can technically be called on a function running in CLI directly, not VS Code. @randi274, you  probably have more context on this than I do, curious to hear your thoughts!

YousefHaggyHeroku: Roger that, making that change now

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
jag-sfdc: This is good 👍

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
```diff
@@ -8,31 +8,58 @@
...
-      .withLogName('force_function_invoke')
-      .build();
+import { runFunction } from '@heroku/functions-core';
+import { LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode/out/src';
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+export class ForceFunctionInvoke extends LibraryCommandletExecutor<string> {
+  constructor(debug: boolean = false) {
+    super(
+      nls.localize('force_function_invoke_text'),
+      debug ? 'force_function_debug_invoke' : 'force_function_invoke',
```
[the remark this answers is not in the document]
YousefHaggyHeroku: Done!

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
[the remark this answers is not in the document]
randi274: Solid change here, thanks for the diligence!

YousefHaggyHeroku:
Merged in https://github.com/forcedotcom/salesforcedx-vscode/pull/3399

PR #3283, opened 2021-06-03
feat: cache for FileProperties using Memento

maxcamp:
### What does this PR do?
 A wrapper for Memento that can serve as a cache for conflict detection.

@W-9258968@, @W-9275211@, @W-9275205@, @W-9275169@

### Functionality Before
Mostly new things. The one place where I made edits was in the MockMemento class for unit tests. It now has functionality (implemented with arrays) rather than just method stubs.

### Functionality After
PersistentStorageService class which stores LastModifiedDates using FileNames as the keys. It's methods are initialize() and getInstance() which implement a singleton pattern, a get() method, a set() method, and a setMultiple() method which proves to be useful.


packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,54 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { FileProperties } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
+import {
+  ExtensionContext,
+  Memento
+} from 'vscode';
+
+interface ConflictFileProperties {
+  lastModifiedDate: string;
```
maxcamp, line 14: Currently only storing lastModifiedDate because it's the only thing needed to detect a conflict. Could easily add more fields but I don't think they're necessary

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export class PersistentStorageService {
```
rcoringrato-sfdc, line 17: @maxcamp I like the interface of this service. It hides the details of the underlying storage and gives flexibility to change it if necessary.

packages/salesforcedx-vscode-core/src/conflict/index.ts
```diff
@@ -77,3 +78,9 @@ function openResource(node: ConflictNode) {
     window.showTextDocument(local).then(() => { });
   }
 }
+
+export function setupFilePropertyStorage(
+  extensionContext: ExtensionContext
+) {
+  PersistentStorageService.initialize(extensionContext);
+}
```
AnanyaJha, line 86: do we need to have a wrapper around the initialize() here? could we call initialize() directly when needed?

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { FileProperties } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
```
AnanyaJha, line 7: Can we import `FileProperties` from the top level? ie.

`import { FileProperties } from '@salesforce/source-deploy-retrieve';`

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.storage = context.workspaceState;
+  }
+
+  public static initialize(context: ExtensionContext) {
+    PersistentStorageService.instance = new PersistentStorageService(context);
+  }
+
+  public static getInstance(): PersistentStorageService {
+    if (!PersistentStorageService.instance) {
+      throw new Error('Storage should have been initialized upon extension activation');
```
AnanyaJha, line 31: Let's localize this message, we can do that by adding the message to [this file](https://github.com/forcedotcom/salesforcedx-vscode/blob/844a6497b46400b5556a0c4eb97d52b863b978bb/packages/salesforcedx-vscode-core/src/messages/i18n.ts)

packages/salesforcedx-vscode-core/test/vscode-integration/telemetry/MockContext.ts
```diff
@@ -16,19 +16,38 @@ import {
...
   private telemetryGS: boolean;
+  private keys: string[] = [];
+  private values: any[] = [];
 
   constructor(setGlobalState: boolean) {
     this.telemetryGS = setGlobalState;
   }
 
-  public get(key: string): any {
+  private getIndex(key: string): number {
```
xyc, line 26: let's just use `Array.prototype.findIndex`

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -0,0 +1,54 @@
...
+interface ConflictFileProperties {
+  lastModifiedDate: string;
+}
+
+export class PersistentStorageService {
+  private storage: Memento;
+  private static instance?: PersistentStorageService;
+
+  private constructor(context: ExtensionContext) {
+    this.storage = context.workspaceState;
```
AnanyaJha, line 22: it looks like we switched over to using the `workspaceState` storage but from my local testing, I'm not seeing this storage persist data between sessions. are there any drawbacks to using the `globalState` storage?

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
jag-sfdc: Yes, with yesterday's library release you should be able to do this.

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
maxcamp: My thought was just that using the smaller scope of workspaceState would be preferable if it did in fact work. I saw the data persist for simple cases but did not thoroughly test it, so I have switched back to global in the latest commit.

packages/salesforcedx-vscode-core/test/vscode-integration/telemetry/MockContext.ts
maxcamp: Updated!

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
maxcamp: Thanks for pointing me to that file. I made an addition, let me know if that works

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
maxcamp: Would need the extensions to import SDR 2.1.5 which is currently causing me problems, perhaps related to @salesforce/core. Will update this as soon as SDR 2.1.5 is being used

packages/salesforcedx-vscode-core/src/conflict/index.ts
maxcamp: Updated it to do that in the latest commit

jag-sfdc (approved):

PR #3289, opened 2021-06-04
remove functions feature-flag and update doc @W-9311863@

jag-sfdc:
### What does this PR do?
Function enabled by default and doc has been updated to point to the latest functions plugin

@W-9311863@

### Functionality Before
'Experimental: Support Functions' had to be turned on to see functions commands

### Functionality After
Functions commands appear in command palette and right-click by default
Doc has been updated to point to the latest functions plugin.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -547,16 +547,8 @@ export async function activate(context: vscode.ExtensionContext) {
     return internalApi;
   }
 
-  // Set functions enabled context
-  const functionsEnabled = sfdxCoreSettings.getFunctionsEnabled();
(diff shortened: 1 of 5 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -547,16 +547,7 @@ export async function activate(context: vscode.ExtensionContext) {
     return internalApi;
   }
 
-  // Set functions enabled context
-  const functionsEnabled = sfdxCoreSettings.getFunctionsEnabled();
```
xyc, line 551: do we want to remove `getFunctionsEnabled` as well?

xyc (approved):

PR #3298, opened 2021-06-07
feat: cache FileProperties after a retrieve

maxcamp:
### What does this PR do?
FileProperties are added to the metadata cache after a retrieve

@W-9312957@, @W-9275242@

### Functionality Before
The metadata cache was unused

### Functionality After
After a retrieve, the FileProperties are passed to the cache within PersistentStorageService. Important note: this PR addresses both retrieve from source path and retrieve with manifest, because they overlap within baseDeployRetrieve.


packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -231,6 +232,7 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
       const relativePackageDirs = await SfdxPackageDirectories.getPackageDirectoryPaths();
       const output = this.createOutput(result, relativePackageDirs);
       channelService.appendLine(output);
+      PersistentStorageService.getInstance().setPropertiesForFiles(result.response.fileProperties);
```
maxcamp, line 235: I believe that fileProperties will contain the information for files that were successfully retrieved and only for files that were successfully retrieved. If that is not true, can perhaps add an if statement here.

maxcamp: Is a check necessary? It's not as long as there are no extra fileProperties in the result right

jag-sfdc: We discussed offline and decided not to have this check.

jag-sfdc (approved):

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
AnanyaJha: Doesn't need to be part of this PR, but in a future one we could consider exporting the instance of the PersistentStorageService similar to what we do with telemetry [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/7a8858b60cca1534197f04e764cbbcca2edfed53/packages/salesforcedx-vscode-core/src/telemetry/index.ts#L11)

packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -26,8 +27,9 @@ import {
 } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
```
AnanyaJha, line 27: Can these be referenced from the top level of `@salesforce/source-deploy-retrieve`?

violetyao: Great point! I have updated `ComponentStatus` and `MetadataApiRetrieveStatus` so that they are referenced from the top level. I'll update `MetadataApiDeployStatus` once the SDR version is bumped.

AnanyaJha (approved):

PR #3301, opened 2021-06-08
refactor: move getLogDirPath from replay debugger to utils

violetyao:
### What does this PR do?
This PR moves the `getLogDirPath()` method to `salesforcedx-utils-vscode` from `salesforcedx-vscode-apex-replay-debugger`, so we can use it in both `salesforcedx-vscode-apex` and `salesforcedx-vscode-apex-replay-debugger`.

@W-9236553

### Functionality Before
`getLogDirPath` was defined in `packages/salesforcedx-vscode-apex-replay-debugger/src/utils/index.ts`. Files using `getLogDirPath` (`quickLaunch.ts` and `quickLaunch.test.ts`) were importing the method from replay debugger. 

### Functionality After
`getLogDirPath` is defined in `packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts`. Files using `getLogDirPath` (`quickLaunch.ts` and `quickLaunch.test.ts`) are importing the method from utils.


packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -115,3 +115,7 @@ export class WorkspaceContextUtil {
     return this._alias;
   }
 }
+
+export function getLogDirPath(): string {
+  return join(getRootWorkspacePath(), '.sfdx', 'tools', 'debug', 'logs');
+}
```
brpowell, line 121: Lets put this in `salesforcedx-utils-vscode/src/helpers/paths.ts`. There are other similar path utilities in there.

violetyao: Great point! I actually started with putting `getLogDirPath` in `paths.ts` and was getting an error message of `mocha cannot find module ‘vscode’`. Interestingly, the error was coming from ‘CommandBuilderTest.ts’ and at that point I didn’t add any unit test in `paths.test.ts` yet. 

@xyc pointed out that `vscode` does not exist in unit tests’ context. In order to resolve the error from ‘CommandBuilderTest.ts’, there would be a lot of refactoring required in order to mock vscode by using proxyquire and adding workspace mock in ‘mock.ts’. 

We then found that ’vscode' was already imported in `workspaceContextUtil.ts` and ‘proxyquire’ was already used in `workspaceContextUtil.ts`. Putting `getLogDirPath` in `workspaceContextUtil.ts` would thus result in much less amount of refactoring. 

Thus, we think putting `getLogDirPath` in `context` is a better approach. Let me know what you think!

xyc: agreed, it needs a larger refactoring if this is in the `helpers` folder. The helpers are strictly without `vscode` context, and tests of any function that use the helpers would need to mock out vscode if we go that route.

Also, just wants to point out if you put this function in any file under `context` folder, not only `workspaceContextUtil`, it should work too.

brpowell: Gotcha, thank you both for clarifying. I think that approach makes sense.

Also wanted to point out though If you look at `getTestResultsFolder()` in `paths.ts`, it's doing something very similar - reading cached information from the project's sfdx folder. That function gets around this issue you're describing by asking for the workspace path as a parameter. You could also take that approach for the sake of consistency.

violetyao: Thanks for your reply Bryan! 

We discussed taking `rootDirPath` as an argument for `getLogDirPath` in this [slack thread](https://platformcloud.slack.com/archives/GCDT2EVAQ/p1623093024220100?thread_ts=1623087112.218200&cid=GCDT2EVAQ). 

"""
Gotcha. It would be easier for utils unit test if we use the older version of getLogDirPath
```
export function getLogDirPath(rootDirPath: string): string {
  const dirPath = path.join(rootDirPath, '.sfdx', 'tools', 'debug', 'logs');
  ensureDirectoryExists(dirPath);
  return dirPath;
} 
```
which takes `rootDirPath` as an argument and thus does not involve the vscode module. However, i would then need to modify `quickLaunch.ts` and `quickLaunch.test.ts` under the replay debugger module. Which approach would you prefer? sticking to the current version of getLogDirPath that use the vscode module, or use the old version of getLogDirPath ? :slightly_smiling_face: 
"""

@jag-sfdc mentioned that we should first try without adding `rootDirPath`. I think it makes sense because modifying `quickLaunch.ts` and `quickLauch.test.ts` would introduce more refactoring.

jag-sfdc: The changes look good to me. If @brpowell is fine with approach, let's merge this.

PR #3304, opened 2021-06-09
fix: use correct message for file diffs

rcoringrato-sfdc:
### What does this PR do?

Display the appropriate message in the multi-file diff view when a diff operation detects no differences.

@W-9421428@


packages/salesforcedx-vscode-core/src/conflict/conflictView.ts
```diff
@@ -48,12 +49,17 @@ export class ConflictView {
     title: string,
     remoteLabel: string,
     reveal: boolean,
-    diffResults?: DirectoryDiffResults
+    diffResults?: DirectoryDiffResults,
+    differences?: boolean
   ) {
+    this.diffsOnly = differences || false;
```
brpowell, line 55:
```suggestion
    diffsOnly = false
  ) {
    this.diffsOnly = diffsOnly;
```

nit - the name `differences` feels like it actually is the diffs

brpowell (approved):
Looks good, just one small comment

PR #3305, opened 2021-06-09
feat: spike component diff

rcoringrato-sfdc:
### What does this PR do?

Draft PR that demonstrates use of SourceComponent to check for diffs/conflicts.

PR #3314, opened 2021-06-10
feat: Set cache key to type and fullName

violetyao:
### What does this PR do?
This PR sets key of the metadata cache in `PersistentStorageService.ts` to `${fileType}#${fullName}`. The reason of making this change was discussed in this [thread](https://platformcloud.slack.com/archives/GCDT2EVAQ/p1623259455260400). 

@[W-9422594](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000O97EYAS/view)

### Functionality Before
Key of `PersistentStorageService` was `fileName`. 

### Functionality After
Key of `PersistentStorageService` is changed to `${fileType}#${fullName}`.


packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -35,22 +35,26 @@ export class PersistentStorageService {
...
   }
 
-  public setPropertiesForFile(fileName: string, value: ConflictFileProperties | undefined) {
-    this.storage.update(fileName, value);
+  public setPropertiesForFile(key: string, conflictFileProperties: ConflictFileProperties | undefined) {
+    this.storage.update(key, conflictFileProperties);
   }
 
-  public setPropertiesForFiles(fileProperties: FileProperties | FileProperties[]) {
+  public setPropertiesForFilesRetrieve(fileProperties: FileProperties | FileProperties[]) {
```
violetyao, line 46: Renamed the function to `setPropertiesForFilesRetrieve` because the deploy side would follow a different strategy which fetches `type` and `fullName` from `sourceComponent` and `lastModifiedDate` from `MetadataApiDeployStatus`. I’ll create a separate PR for deploy to avoid confusion.

violetyao (commented):
Borrowed the makeKey function from #3305. Thanks Roy!

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
maxcamp: I don't mind this idea! I think there is a challenge we will want to discuss which is that we would like to use `SourceComponent.type.id` as the first half of our keys. However, `FileProperties` do not come naturally correlated with `SourceComponent`s. Not sure what the best solution is.

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -35,22 +35,26 @@ export class PersistentStorageService {
...
     const fileArray = Array.isArray(fileProperties) ? fileProperties : [fileProperties];
     for (const fileProperty of fileArray) {
       this.setPropertiesForFile(
-        fileProperty.fileName,
+        this.makeKey(fileProperty.type, fileProperty.fullName),
```
brpowell, line 50: @violetyao @maxcamp So a component can be mapped to multiple files, which is where I'm having trouble understanding using this as a key for a single file. I might be missing some context though if you wouldn't mind breaking it down for me :)

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
maxcamp: Update: after a good amount of thought we have decided to do this just as Violet originally had it, using the type name and not the id as the key for the cache. I don't want to do an official "approving review" but I am now on board with the changes in this PR!

violetyao: Discussed offline and would also push `setPropertiesForFilesDeploy` in this PR.

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
violetyao: Great question! Max and I think the cache should save for each component because we don't have info on `lastModifiedDate` about individual files both on deploy and retrieve. Specifically, on the retrieve side (`setPropertiesForFilesRetrieve` ), we extract `lastModifiedDate` from `fileProperties` and `fileProperties` is for each single component; on the deploy side (`setPropertiesForFilesDeploy` ), we extract `lastModifiedDate` from `MetadataApiDeployStatus` and assign it to all components. Let us know what you think! :slightly_smiling_face:

brpowell: I see, that approach definitely makes sense. I just reviewed the API documentation [here](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_result_objects.htm) to add some more context to your response. Thank you!

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -5,6 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { FileProperties } from '@salesforce/source-deploy-retrieve';
+import { MetadataApiDeployStatus} from '@salesforce/source-deploy-retrieve/lib/src/client/types';
+import { LazyCollection } from '@salesforce/source-deploy-retrieve/lib/src/collections';
```
brpowell, line 9: Probably don't need to import `LazyCollection` directly, that's technically an internal export

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -5,6 +5,9 @@
...
+import { SourceComponent } from '@salesforce/source-deploy-retrieve/lib/src/resolve';
```
brpowell, line 10: Probably don't need to import `LazyCollection` directly, that's technically an internal export. As for the other two let's import them from the top level since that follows proper versioning:

```typescript
import {
  SourceComponent,
  MetadataApiDeployStatus,
  FileProperties
} from '@salesforce/source-deploy-retrieve'
```

I think `MetadataApiDeployStatus` should be exported at the top level...

violetyao: I see! I'll create a PR for SDR to export `MetadataApiDeployStatus` at the top level. On a similar note, I use `RequestStatus` in `persistentStorageService.test.ts` and it is not currently exported at the top level.  Shall we also export `RequestStatus` at the top level for SDR?

brpowell (approved):

brpowell:
:shipit:

PR #3325, opened 2021-06-15
feat: add org/project prefix to cache key

violetyao:
### What does this PR do?
The PersistentStorageService cache is global. This PR adds an org/project prefix to every cache entry's key so that they are unique identifiers. 

@[W-9456559](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000OPsiYAG/view)@

### Functionality Before
Cache key was `${type}#${fullName}`.

### Functionality After
Cache key is now `${org}#${project}#${type}#${fullName}`.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
+++ b/packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
@@ -47,29 +47,29 @@ export class PersistentStorageService {
     this.storage.update(key, conflictFileProperties);
   }
 
-  public setPropertiesForFilesRetrieve(fileProperties: FileProperties | FileProperties[]) {
+  public setPropertiesForFilesRetrieve(org: string, project: string, fileProperties: FileProperties | FileProperties[]) {
     const fileArray = Array.isArray(fileProperties) ? fileProperties : [fileProperties];
     for (const fileProperty of fileArray) {
       this.setPropertiesForFile(
-        this.makeKey(fileProperty.type, fileProperty.fullName),
+        this.makeKey(org, project, fileProperty.type, fileProperty.fullName),
         {
           lastModifiedDate: fileProperty.lastModifiedDate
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -70,6 +72,8 @@ export class PersistentStorageService {
   }
 
   public makeKey(type: string, fullName: string): string {
-    return `${type}#${fullName}`;
+    const orgUserName = workspaceContext.username;
```
violetyao, line 75: I didn't have error handling of 1) orgUserName undefined and 2) projectPath empty here because I believe they should trigger 1) `notificationService.showErrorMessage(nls.localize('missing_default_org'));` and 2) `workspaceChecker` in `forceSourceDiff.ts` before entering `PersistentStorageService.ts`. Let me know if I miss anything! :)

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
```diff
@@ -70,6 +72,8 @@ export class PersistentStorageService {
...
+    const projectPath = getRootWorkspacePath();
+    return `${orgUserName}#${projectPath}#${type}#${fullName}`;
```
jag-sfdc, line 77: This looks good. Just for a sanity check, can we test if memento has any limitation on the length of the key?

violetyao: Great question! I tested with key lengths of 0.1k chars, 1k chars, and 10k chars. 0.1k chars and 1k chars work as expected. 10k chars would return a `SIGPIPE` error, but I think the chance is really slim that keys exceed 1k chars.

jag-sfdc (approved):

PR #3332, opened 2021-06-16
feat: new conflict detection mechanism

maxcamp:
### What does this PR do?
This PR creates an alternative mechanism for conflict detection based firstly on the lastModifiedDate of the given components. This is a complete mechanism, with a PostconditionChecker that initiates the process, a conflict detector that compares the timestamps, and finally a ComponentDiffer that determines which file in the component actually had the diff. Note that this code only works for users 1) with a timestamp cache that has the needed information (PersistentStorageService) and 2) once the PostconditionChecker is added to the appropriate commands.

@W-9258989@

### Functionality Before
Only one conflict detection mechanism exists, based on diffing every file in the given directory.

### Functionality After
A second conflict detection mechanism is available to be used, which is smarter and applicable to source path commands, while still providing users with the exact same end interface. However, all commands are still tied to the old mechanism for now.


packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -0,0 +1,87 @@
...
+  private determineConflicts(
+    data: CorrelatedComponent[]
+  ) {
+    const cache = PersistentStorageService.getInstance();
+    const conflicts: Set<string> = new Set<string>();
+    data.forEach(component => {
+      let lastModifiedInOrg;
+      let lastModifiedInCache;
+
+      lastModifiedInOrg = component.fileProperties.lastModifiedDate;
```
jag-sfdc, line 57: Help me understand this - why are we using fileProperties here? Didn't we discuss about using SourceComponents for both retrieve and deploy. The same question for the "correlateResults()" method - why we are retrieving and storing fileProperties?

maxcamp: Good question! We need FileProperties for thing, and that is the lastModifiedDate. For the cache keys, we decided to start using type name and fullName because those are fields that exist on both FileProperties and SourceComponents and sometimes, such as after a deploy, we don't have FileProperties. You'll see that in this PR, FileProperties were added to the MetadataCacheResult just so we can get access to the lastModifiedDate here. 

A valid change we could make is only storing lastModifiedDate instead of the whole FileProperties in the CorrelatedComponent type.

jag-sfdc: I like the idea of storing lastModifiedDate instead of the whole FileProperties in the CorrelatedComponent type.
One more question - where do we get lastModifiedDate after deploy given that FileProperties is not available?

maxcamp: It is just a field in the MetadataApiDeployStatus (the supposed reason being that FileProperties are not necessary, they would be the exact same for every file getting deployed).

maxcamp: Made this change in a recent [commit](https://github.com/forcedotcom/salesforcedx-vscode/pull/3332/commits/c2fe3428dc4c1eb7654b5e42a13f70c0dd132ada)

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -284,3 +286,103 @@ export class ConflictDetectionChecker implements PostconditionChecker<string> {
...
+          false
+        );
+      } else {
+        channelService.appendLine(
+          nls.localize(
+            'conflict_detect_command_hint',
+            this.messages.commandHint(componentPath)
+          )
+        );
+        channelService.showChannelOutput();
```
AnanyaJha, line 372: It looks like we're also calling `showChannelOutput` on line 351, do we need to call it again here?

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { SourceComponent } from '@salesforce/source-deploy-retrieve/lib/src/resolve';
+import * as fs from 'fs';
+import * as path from 'path';
+
+export interface ComponentDiff {
+  projectPath: string;
+  cachePath: string;
+}
+
+export class ComponentDiffer {
```
AnanyaJha, line 17: Do these methods need to be part of a class? Could they be standalone functions too?

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { SourceComponent } from '@salesforce/source-deploy-retrieve/lib/src/resolve';
```
AnanyaJha, line 8:
```suggestion
import { SourceComponent } from '@salesforce/source-deploy-retrieve';
```

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface ComponentDiff {
+  projectPath: string;
+  cachePath: string;
+}
+
+export class ComponentDiffer {
+
+  constructor() {}
+
+  public diffComponents(
```
AnanyaJha, line 21: let's add some JSDocs on this method too!

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
maxcamp: My instinct is always to have a class, but there ended up being less methods than I first anticipated, so they could be standalone static methods. Would you put these methods in their own file still or include them in TimestampConflictDetector?

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
maxcamp: Sure happy to! Any there any general rules about when to use JSDocs?

Made one for this method in the latest [commit](https://github.com/forcedotcom/salesforcedx-vscode/pull/3332/commits/59311d19c0602fd852eca39f711ae0ccc3b8f0a3).

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
maxcamp: I assumed that this method flushed the stream, and therefore was responsible for showing the text that is appended in the line above. Having now tested it, I see that you're right and it's only needed once to switch the channel. Thanks!

Fixed in the latest [commit](https://github.com/forcedotcom/salesforcedx-vscode/pull/3332/commits/59311d19c0602fd852eca39f711ae0ccc3b8f0a3).

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
AnanyaJha: Awesome thanks! We generally try to include JSDocs for methods that have several parameters and cover broader functionality

packages/salesforcedx-vscode-core/src/conflict/componentDiffer.ts
AnanyaJha: Ah that makes sense! We could keep these methods in this file and just change the signatures to be `export diffComponents()` and keep the other method a private utility function

maxcamp: My recent changes are to implement this! I ended up reworking the unit test a lot, where now it writes and reads from temporary files

packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -0,0 +1,85 @@
...
+  MetadataCacheResult,
+  MetadataCacheService,
+  PersistentStorageService
+} from './';
+import { diffComponents } from './componentDiffer';
+import { CorrelatedComponent } from './metadataCacheService';
+
+export class TimestampConflictDetector {
+  private diffs: DirectoryDiffResults;
+  // Probably want a UI update to remove scannedLocal and scannedRemote
```
AnanyaJha, line 21: let's add a work item for this and remove the comment

AnanyaJha (approved):
Small comment and looks like it needs to be rebased again, but otherwise :shipit: !

PR #3337, opened 2021-06-17
chore(tests): disable other extensions on integration tests

dobladez:
### What does this PR do?
When running integration tests, disable any extensions already installed in the current users' local VS Code installation.

This is to avoid possible conflicts when running tests locally. For example: if you have extra extensions for code-completion (like, Tabnine) the completion items get intermixed with the ones from our extensions, sometimes breaking our test assertions.


brpowell (approved):

gbockus-sf:
This doesn't seem to be an issue currently so closing this out.

PR #3338, opened 2021-06-17
feat(functions): adds language data to telemetry for functions

randi274:
### What does this PR do?
* This PR sends the function language (java or node for now) to our telemetry service when the create, start,
stop and invoke functions commands complete, so that we can better understand what languages our 
developers are using when working with the new Functions feature.

@W-9282462@

### Functionality Before
Functions commands would run without logging language information.

### Functionality After
Language data is being recorded!


packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceFunctionCreate.test.ts
```diff
@@ -124,6 +125,17 @@ describe('Force Function Create', () => {
       );
     });
 
+    it('Log metric should add additional language property for telemetry', async () => {
+      const telemetryServiceStub = stub(telemetryService, 'sendCommandEvent');
```
maxcamp, line 129: Looks like the CircleCI check is failing because this stub needs to be restored! I would declare this variable at the top of the describe() block and then restore it in the afterEach(). Then the checks will likely pass

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -110,6 +110,19 @@ export class FunctionService {
     return this.startedExecutions.size > 0;
   }
 
+  /**
+   * Returns the debugType of the first of the startedExecutions as a way to determine the language
```
jag-sfdc, line 114: Nit pick - Currently we run all the commands with the assumption that the user has only running function.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -66,7 +72,7 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
             );
             reject(err);
           }
-          resolve();
+          resolve(void 0);
```
jag-sfdc, line 75: Let's not make this change for consistency.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -83,11 +89,24 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
...
         });
       });
     }
   }
+
+  public logMetric(
```
jag-sfdc, line 98: Where is this method called during creation? I'm thinking if there is any other way to readily get the language information.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceFunctionCreate.test.ts
randi274: Thanks @maxcamp!

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
randi274: Sounds good - I was seeing errors in VS Code for this, and it seems like there are some places in the code where we call it this way, but I'm happy to put it back.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
randi274: I was struggling with this one on Thursday, and this was the suggestion of @brpowell. It gets called further downstream in the extended class BaseTemplateCommand's execute, which here we aren't overriding in forceFunctionCreate.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
randi274: Bring on the nit picks - I'll learn more that way! 👍 I'm a little unclear on the recommended change, can you clarify? Is it just changing the comment, or something in the method?

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
brpowell: It's to attach the language property to the `commandExecution` event that already gets called in `BaseTemplateCommand` for this command and not send additional or new events

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: Just changing the comment.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Got it! One more question - if we are going with this.getFileExtension(), the language would be js, ts or java correct? Whereas it is node or java in other places that uses FunctionService's getFunctionLanguage.

jag-sfdc: @randi274 - Just read your comment on the story and you talk about language being java, javascript or typescript. Can we have consensus on what it is going to be? We should be fine with java, ts or js (Based on the file extension, if that's the easiest) as long it is consistent everywhere.

randi274: Ah right, that's a great point @jag-sfdc! I lost focus on that when we were talking through this on Thursday. Let me think on making that consistent, and then get back to you if I'm unclear on how to make that happen.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
```diff
@@ -71,7 +71,8 @@ export async function forceFunctionDebugInvoke(sourceUri: Uri) {
       await FunctionService.instance.stopDebuggingFunction(localRoot);
       telemetryService.sendCommandEvent(
         'force_function_debug_invoke',
-        startTime
+        startTime,
+        { language: FunctionService.instance.getFunctionLanguage() }
```
brpowell, line 75: This change is fine for the scope of the PR, but just wanted to FYI that by having `sendCommandEvent` here, I believe we are sending two events for the same execution since the `execute()` method of `SfdxCommandletExecutor` is already calling this through `logMetric`. This could report double the executions in insights

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
brpowell: It should be fine just to make it the file extension, unless we can think of a function that is two different languages with the same extension.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -88,6 +94,24 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
...
+    properties?: Properties,
+    measurements?: Measurements
+  ) {
+    if (properties) {
+      const fileExtension = this.getFileExtension().replace('.', '');
+      if (fileExtension === 'js') {
+        properties.language = 'node';
+      } else if (fileExtension === 'java') {
+        properties.language = 'java';
+      }
```
brpowell, line 110: Not sure if we should limit only reporting the language if it's one of these two things, maybe there should be a fallback

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceFunctionCreate.test.ts
```diff
@@ -116,19 +120,34 @@ describe('Force Function Create', () => {
...
-      sinon.assert.calledWith(
+      assert.calledOnce(execStub);
+      assert.calledWith(execStub, 'mvn install', { cwd: path.join('some', 'dir', 'functions', 'myFunc1') });
+      assert.calledWith(
         notificationServiceStub,
         nls.localize('force_function_install_mvn_dependencies_error', errorText)
       );
     });
 
+    it('Log metric should add additional language property for telemetry for java function', async () => {
```
brpowell, line 131: The convention for test subjects is to start it with "Should", so `it('Should.....') {}`

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceFunctionCreate.test.ts
```diff
@@ -59,18 +59,22 @@ describe('Force Function Create', () => {
...
     let withProgressStub: SinonStub;
     const functionInfoJS = { fileName: 'myFunc1', outputdir: 'some/dir', language: 'javascript' };
     const functionInfoJava = { fileName: 'myFunc1', outputdir: 'some/dir', language: 'java' };
     let rootWorkspacePathStub: SinonStub;
 
     beforeEach(() => {
-      execStub = stub(cp, 'exec');
-      notificationServiceStub = stub(notificationService, 'showWarningMessage');
-      rootWorkspacePathStub = stub(rootWorkspace, 'getRootWorkspacePath');
+      sandbox = createSandbox();
```
brpowell, line 71: Nice on using the sandbox. I know some tests create a sandbox in the beforeEach hook, but i think it really only needs to be made once at the top.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Can't think of a scenario where a fallback would be required. The metaDictionary definitions does not have a default MetaDataInfo, so good to follow the same pattern.

jag-sfdc: @randi274 - While looking into this, I see in metadataDictionary.ts the suffix is set to blank for functionjs and functionjava. I think we need to take care of this. Have we tested this end to end to see if it works? Testing telemetry in local is a bit of a challenge (you need to use the vsix from circleci). Let me know if you need help with that.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
randi274: @brpowell would you recommend another WI to make a similar change as we did in forceFunctionCreate? I think that would also apply to forceFunctionStop.

brpowell: That would be worth filing if that is the case, otherwise we're getting incorrect data on command usage.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -88,6 +94,24 @@ export class ForceFunctionCreateExecutor extends BaseTemplateCommand {
...
   }
+
+  public logMetric(
+    logName: string | undefined,
+    hrstart: [number, number],
+    properties?: Properties,
+    measurements?: Measurements
+  ) {
+    if (properties) {
+      const fileExtension = this.getFileExtension().replace('.', '');
```
xyc, line 105: can we compare with `.js` and `.java` instead of replacing the `.`?

randi274: Sure can!

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
randi274: @jag-sfdc some help would be great, thank you! I was able to get the vsix files installed, but I'm not quite sure how to run the telemetry locally, and could use some help triggering the functions as well since I had some difficulty getting that set up. I'll set up a meeting for tomorrow 👍

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionInvoke.ts
YousefHaggyHeroku: Hello 👋 , intern here. sorry to interject but I had a question about this. I'm not entirely sure what a WI is, but regarding the metric logging *I think* my [PR for switching function invocation to using a library instead of CLI](https://github.com/forcedotcom/salesforcedx-vscode/pull/3282) handles that metric. So maybe it's worth holding on that improvement for now? I'm not entirely sure, just figured I'd make note of it

randi274: Thanks for the heads up! If you get the sense that it's logging dupes and you have a chance to change it in your PR, feel free to make a change or note of it there, factoring in these changes too. Sounds like it's a good change if it gets us further off of the CLI!

PR #3342, opened 2021-06-20
Adds Java 11 path for Arch Linux installations

renatoliveira:
This PR adds the Java path for Arch Linux machines (which might prove to be useful for Arch-based installs that are more user-friendly, such as Manjaro).


randi274:
@sbudhirajadoc does this change work for you? If so, I can get it set up for merge.

PR #3348, opened 2021-06-22
feat: enable conflict detection on retrieve

maxcamp:
### What does this PR do?
This PR enables the new conflict detection mechanism for both retrieve with manifest and retrieve from source path. For the source path side, a new helper class CompositePostconditionChecker was created. 

@W-9164035@

### Functionality Before
Conflict detection was not used for retrieve from source path, the old conflict detection mechanism was used for retrieve with manifest.

### Functionality After
The new conflict detection mechanism is used for all retrieves.

### DIFF
--- a/docs/_articles/en/user-guide/detect-conflicts.md
+++ b/docs/_articles/en/user-guide/detect-conflicts.md
@@ -7,7 +7,7 @@ Whenever you retrieve or deploy source to keep the local project and your defaul
 
 If conflicts are detected, you can cancel the retrieve or deploy operation and view the differences between your local project and default org.
 
-To view the files that caused conflicts, click the Org Differences icon ({% octicon issue-opened %}) in the Activity Bar. The Org Differences: Conflicts view opens in the Side Bar and displays the list of files with conflicts. To resolve conflicts, click a file to open the diff editor and compare the remote file (to the left) with the local file (on the right). To edit a file, hover over the file name and click the open file icon.
+To view the files that caused conflicts, click the Org Differences icon ({% octicon issue-opened %}) in the Activity Bar. The Org Differences: Conflicts view opens in the Side Bar and displays the list of files with conflicts. To resolve conflicts, click a file to open the diff editor and compare the remote file (on the left) with the local file (on the right). To edit a file, hover over the file name and click the open file icon.
 
 You can also view the resource files with conflicts and the CLI command to overwrite the conflicts in the Output panel.
 
@@ -22,7 +22,7 @@ Because the conflict detection feature is in beta, you must enable the feature:
 
 You can also enter conflict detection in the search box to find the feature and then enable it.
 
-In this beta release, we have enabled conflict detection for the **SFDX: Retrieve Source in Manifest from Org** and **SFDX: Deploy Source in Manifest to Org** commands only. All manifest files that exists both in the org and in the local environment are checked for conflicts; files that don’t exist in both are not checked.
+Once enabled, conflict detection is available on all commands that deploy and retrieve source.
 
 ![Prompt for conflict detection](./images/DetectConflict_prompt.png)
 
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveManifest.ts
```diff
@@ -106,7 +106,7 @@ export async function forceSourceRetrieveManifest(explorerPath: vscode.Uri) {
     sfdxCoreSettings.getBetaDeployRetrieve()
       ? new LibrarySourceRetrieveManifestExecutor()
       : new ForceSourceRetrieveManifestExecutor(),
-    new ConflictDetectionChecker(messages)
+    new TimestampConflictChecker(true, messages)
```
AnanyaJha, line 109: Does this need to be gated behind a feature flag?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -32,6 +32,35 @@ import { PathStrategyFactory } from './sourcePathStrategies';
...
+  public constructor(...postcheckers: Array<PostconditionChecker<any>>) {
+    this.postcheckers = postcheckers;
+  }
+  public async check(inputs: CancelResponse | ContinueResponse<T>): Promise<CancelResponse | ContinueResponse<T>> {
+    if (inputs.type === 'CONTINUE') {
+      const aggregatedData: any = {};
+      for (const postchecker of this.postcheckers) {
+        const input = await postchecker.check(inputs);
+        if (input.type === 'CONTINUE') {
+          Object.keys(input.data).map(
```
xyc, line 46: let's use forEach since we intend to use side effects here

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveManifest.ts
maxcamp: The feature flag is the first thing hit in the check() method that gets called!

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
maxcamp: Are we able to use forEach given that there is an await being used inside the loop? If a simple switch to forEach is not an option is there anything else you would recommend?

docs/_articles/en/user-guide/detect-conflicts.md
```diff
@@ -22,7 +22,7 @@ Because the conflict detection feature is in beta, you must enable the feature:
 
 You can also enter conflict detection in the search box to find the feature and then enable it.
 
-In this beta release, we have enabled conflict detection for the **SFDX: Retrieve Source in Manifest from Org** and **SFDX: Deploy Source in Manifest to Org** commands only. All manifest files that exists both in the org and in the local environment are checked for conflicts; files that don’t exist in both are not checked.
+Once enabled, conflict detection is available on all commands that deploy and retrieve source.
```
maxcamp, line 25: @sbudhirajadoc With this PR, conflict detection will change to use timestamps and will be available for the 'Retrieve source from source path' and soon the 'Deploy source from source path' commands. So at the minimum we need a doc change like this, and we may want to describe the new mechanism as well.

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
maxcamp: Realized I misunderstood the comment! Changed in the latest commit

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postconditionCheckers.test.ts
```diff
@@ -57,6 +62,135 @@ describe('Postcondition Checkers', () => {
     });
   });
 
+  describe('CompositePostconditionChecker', () => {
```
jag-sfdc, line 65: Nice job covering all the scenarios 🥇

jag-sfdc (approved):

Issue #3294, opened 2021-06-04
Conflict detection error when retrieving source from manifest for the first time


ben-koch:
### Summary

When starting a new project and running "Retrieve Source in Manifest from Org" for the first time, an error is thrown during conflict detection and the retrieval action fails.

### Steps To Reproduce:

1.  Create Project with Manifest
2. Authorize a new SF Org
3. Run "Retrieve Source in Manifest from Org"

### Expected result

Conflict detection should be skipped or return no conflicts and the manifest retrieval should succeed, assuming no other issues.

### Actual result

An error is thrown and the retrieval fails. Error is as follows:

"No components in the package to retrieve
An error was encountered during conflict detection. Error: Conflict Detection failed to retrieve org file state"

### Additional information

Manually disabling conflict detection and running the retrieval is successful. Once there are components in the VS Code project, re-enabling conflict detection and running Retrieve Source in Manifest from Org no longer results in an error and conflict detection works normally.

**VS Code Version**: 1.56.2

**SFDX CLI Version**: 

**OS and version**: MacOS Big Sur 11.2.3

PR #3358, opened 2021-06-24
fix: conflict detection not working on new project

maxcamp:
### What does this PR do?
Solves a bug that caused conflict detection to throw an error in a new project.

The two fixes are
1. When doing a retrieve, forceAddWildcards should be true so that the ComponentSet can be created properly. See forceSourceRetrieveManifest.ts for an example.
2. If there are no components (specifically no local components) the common root should be '' not baseDir, because the common root is defined as the relative path from baseDir.

#3294, @W-9368479@

### Functionality Before
After creating a new project and running `Retrieve with Manifest` with conflict detection on, an error was triggered 'No components in the package to retrieve'

### Functionality After
Creating a new project and running `Retrieve with Manifest` with conflict detection on causes conflict detection to work as normal.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
+++ b/packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
@@ -110,7 +110,8 @@ export class MetadataCacheService {
       this.sourceComponents = this.isManifest
         ? await ComponentSet.fromManifest({
             manifestPath: this.componentPath,
-            resolveSourcePaths: packageDirs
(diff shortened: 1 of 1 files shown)


xyc (approved):

xyc:
Looks good on my side. I'd let @rcoringrato-sfdc weigh in this too

rcoringrato-sfdc:
@maxcamp @xyc Looks good, nice work.

rcoringrato-sfdc (approved):

PR #3371, opened 2021-06-29
fix: add missing Experience Cloud targets

muenzpraeger:
This PR adds missing Experience Cloud targets to the js-meta XML validation.

n/a

### Functionality Before
The js-meta validation is missing two Experience Cloud targets for LWR sites.

### Functionality After
The js-meta validation no longer declares the valid targets as invalid.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/resources/static/js-meta.xsd
+++ b/packages/salesforcedx-vscode-lwc/resources/static/js-meta.xsd
@@ -112,6 +112,16 @@
                 <xs:documentation>Enables a component to be used on a Lightning community page in Experience Builder.</xs:documentation>
               </xs:annotation>
             </xs:enumeration>
+            <xs:enumeration value="lightningCommunity__Page_Layout">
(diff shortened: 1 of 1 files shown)


xyc (approved):

smithgp (approved):
Sorry, I was OOTO last week. LGTM

gbockus-sf:
@smithgp @muenzpraeger is this still a thing.  Should we get this caught up and merged?

smithgp:
Well, one and half years ago, I think I was only added as a reviewer since I had just previously edited this file, so I'm not sure. I totally don't remember this.
But, according to https://developer.salesforce.com/docs/component-library/documentation/en/lwc/reference_configuration_tags, `lightningCommunity__Theme_Layout` and `lightningCommunity__Page_Layout` are valid targets still, so I imagine this should go in.

muenzpraeger:
Yup, this is still an issue. Would be great to have this resolved.

PR #3376, opened 2021-06-30
fix: retrieve nothing should not error

maxcamp:
### What does this PR do?
This PR changes conflict detection functionality so that it does not error if the retrieve result is undefined. Instead, it will simply show a warning. This makes more sense because a retrieve without conflict detect on would not throw an error in this situation. There are also many style changes in this PR.

@W-9537468@

### Functionality Before
Using an empty manifest file, or a manifest file that does not correspond to any components in the org, would result in the conflict error "failed to retrieve org file state".

### Functionality After
Now, the situation above results in the warning "no files retrieved during conflict detection" and the retrieve proceeds as normal.

UPDATE: Intentionally no longer showing even a warning. If there are no files, conflict detection simply proceeds to the deploy operation


packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -29,35 +30,45 @@ export class TimestampConflictDetector {
...
+  private determineConflicts(data: CorrelatedComponent[]) {
     const cache = PersistentStorageService.getInstance();
     const conflicts: Set<TimestampFileProperties> = new Set<TimestampFileProperties>();
     data.forEach(component => {
       let lastModifiedInOrg: string | undefined;
       let lastModifiedInCache: string | undefined;
 
       lastModifiedInOrg = component.lastModifiedDate;
-      const key = cache.makeKey(component.cacheComponent.type.name, component.cacheComponent.fullName);
+      const key = cache.makeKey(
```
jag-sfdc, line 51: What is making all these styling changes?

jag-sfdc (approved):

PR #3378, opened 2021-06-30
fix: correct response from CompositePostconditionChecker

maxcamp:
### What does this PR do?
During QA, a bug was found on deploying and retrieving with source path due to conflict detection changes. The problem went back to CompositePostconditionChecker not returning the correct data. This PR updates that class to chain the data rather than aggregate, which makes more sense for the common use case of data being file paths.

@W-9164035@, @W-9164037@

### Functionality Before
Running deploy or retrieve from source path would result in the error "fsPaths is not iterable", due to incorrect data being returned from CompositePostconditionChecker. 

### Functionality After
CompositePostconditionChecker returns the correct data and there is no longer an error.


randi274:
The change works great! That error isn't cropping up anymore. @maxcamp it looks like we've got a failing unit test, but once that gets cleared up this will be good to get in.

randi274 (approved):
Great quick fix!

maxcamp:
@randi274 The failing test was the same flapper Xiaoyi and I talked about the other day. It has been hitting me a lot. Merging in though!

PR #3379, opened 2021-06-30
fix: add back conflict detection progress/cancel popup

maxcamp:
### What does this PR do?
This PR adds back the bottom-right progress/cancel popup for conflict detection. This is done by using the MetadataCacheExecutor class instead of the loadCache() method.

@W-9545597@

### Functionality Before
The new conflict detection mechanism did not have the progress/cancel popup associated with running a command.

### Functionality After
Conflict detection now has that popup.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
@@ -14,10 +14,16 @@ import { existsSync } from 'fs';
 import { join, normalize } from 'path';
 import { channelService } from '../../channels';
 import {
+  SfdxCommandlet,
+  SfdxWorkspaceChecker,
+  SimpleGatherer
+} from '../../commands/util';
+import {
   ConflictDetectionConfig,
   conflictDetector,
   conflictView,
   DirectoryDiffResults,
+  MetadataCacheExecutor,
   MetadataCacheService
 } from '../../conflict';
 import { TimestampConflictDetector } from '../../conflict/timestampConflictDetector';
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -13,11 +13,17 @@ import {
...
+  SfdxWorkspaceChecker,
+  SimpleGatherer
+} from '../../commands/util';
 import {
   ConflictDetectionConfig,
   conflictDetector,
   conflictView,
   DirectoryDiffResults,
+  MetadataCacheExecutor,
   MetadataCacheService
```
AnanyaJha, line 27: unused import
```suggestion
```

packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -28,16 +28,20 @@ export class TimestampConflictDetector {
     this.diffs = Object.assign({}, TimestampConflictDetector.EMPTY_DIFFS);
   }
 
-  public createDiffs(
+  public getDiffs(): DirectoryDiffResults {
+    return this.diffs;
+  }
+
+  public async createDiffs(
+    username: string,
```
AnanyaJha, line 36: why did we need to add in the username parameter here?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
maxcamp: Thanks for catching this

packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
maxcamp: This is just so we can match a signature exported by MetadataCacheExecutor. It uses the type MetadataCacheCallback which requires a username parameter. So you're correct that the username parameter is unused, but I just added it in instead of changing the callback type (which I think would also be doable).

maxcamp:
This pull request is being closed because adding toast messages to conflict detection results in too much noise for the user on every command.

PR #3393, opened 2021-07-07
fix: conflict detection support deploy on save files under multiple directories

violetyao:
### What does this PR do?
This PR addresses the issue that the new conflict detection mechanism does not work for `deploy on save` multiple files under multiple directories.  
@W-9549835@

### Functionality Before
We used relative paths in `diffComponents` to match local and remote components. This became problematic when we `deploy on save` multiple files under multiple directories: keys of cacheIndex are of format `test1.cls` while keys of projectIndex are of format `folder1/test1.cls`. Thus, when we assign `cacheIndex.get(key)` to `cachePath`, the value would be undefined. 

### Functionality After
We now use `diffComponents` to match local and remote components. We also have `localRelPath` and `remoteRelPath` for visualizing file diffs.


AnanyaJha:
@violetyao quick question for you, how have you been testing these changes? I haven't been able to view multiple conflicts from deploy on save yet so I'd like to make sure I'm reproducing the issue correctly 😄 
My current strategy has involved deploying three classes to the org outside of VS Code, making changes to those three classes within VS Code, and saving the classes within VS Code. At this point, the conflict detection modal appears, but I'm only seeing a conflict in _one_ of the classes rather than all three

violetyao:
> @violetyao quick question for you, how have you been testing these changes? I haven't been able to view multiple conflicts from deploy on save yet so I'd like to make sure I'm reproducing the issue correctly 😄
> My current strategy has involved deploying three classes to the org outside of VS Code, making changes to those three classes within VS Code, and saving the classes within VS Code. At this point, the conflict detection modal appears, but I'm only seeing a conflict in _one_ of the classes rather than all three

Hi Ananya! Have you modified the three classes in their org web page after deploying them? The `remote last modified date` and `local last sync date` would be the same if we don't modify the classes after deployment. We can also talk about this in our sync today!

AnanyaJha (approved):

maxcamp:
While these changes are a great start to fixing the multiple directories problem, I found that using path.basename() leads to problems with metadata types that do not follow the "matchingContent" and "bundle" strategies. Examples include StaticResources ("mixedContent" strategy) and CustomObjects ("decomposed" strategy). I'm going to push some changes soon that hopefully fix those issues.

maxcamp:
Key resources about the different strategies:
SDR Handbook - [https://salesforce.quip.com/F3q6AmNSquvE#ZJKACAke3ls](https://salesforce.quip.com/F3q6AmNSquvE#ZJKACAke3ls)
SDR Code -[https://github.com/forcedotcom/source-deploy-retrieve/tree/develop/src/resolve/adapters](https://github.com/forcedotcom/source-deploy-retrieve/tree/develop/src/resolve/adapters)

AnanyaJha:
Looks good to me :shipit:

PR #3399, opened 2021-07-08
Pr/3282 Use @heroku/functions-core library implementation of 'Start Function' and 'Run Function'

jag-sfdc:
What does this PR do?
This PR switches the Function Invoke and Function Start commands to using the functions-core helper library directly instead of interfacing with the cli. Ditching the CLI decreases the execution time of those commands significantly

The CLI also now uses the functions-core shared helper library; that change has been deployed to production. salesforcecli/plugin-functions#65

@W-9536315@

Functionality Before
Uses CLI for functions interactions from VSCode

Functionality After
Use library for functions interactions from VSCode


randi274:
@jag-sfdc I can confirm the changes are the same and it looks like tests are passing again (huzzah!). I installed the latest build from this branch (22646) with ctv for a quick sanity check though, and it looks like something's broken. I can't use any commands in an SFDX repo except create project, and that's generating the attached error. I also checked out the current latest build (22633) and can confirm all expected commands are working, including create function.

Screen Shot 2021-07-08 at 12 59 36 PM

AnanyaJha (approved):

PR #3409, opened 2021-07-12
fix: remove setting to enable function @W-9311863

jag-sfdc:
### What does this PR do?
Do not show option to enable/disable function in the settings page

@W-9311863

### Functionality Before
image

### Functionality After
The above option is not shown


randi274 (approved):
Looks good! I'm not seeing the setting any longer and functions is still working as expected.

AnanyaJha (approved):

PR #3410, opened 2021-07-12
feat: support checkpoints for one-step replay debugger

violetyao:
### What does this PR do?
As part of the one-step replay debugger command, this PR checks for the presence of checkpoints. If they are present, this PR uploads them before we execute the Apex Test(s). Checkpoints have a limit of 5 per org/user, if the limit is exceeded, the command execution would be canceled and a message indicating this limit violation would be shown to the user.

@W-7757402@

### Functionality Before
The one-step replay debugger command did not support checkpoints. 

### Functionality After
The one-step replay debugger command uploads checkpoints if they are present.
Screen Shot 2021-07-12 at 11 12 08 AM
Screen Shot 2021-07-12 at 11 11 47 AM

If there are more than five checkpoints, the command executed is canceled and a message is reported to the user. 
Screen Shot 2021-07-12 at 11 11 07 AM


packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -901,6 +901,11 @@ export async function sfdxCreateCheckpoints() {
     telemetryService.sendCheckpointEvent(errorMsg);
     creatingCheckpoints = false;
   }
+  if (updateError) {
+    return false;
+  } else {
+    return true;
+  }
```
AnanyaJha, line 908:
```suggestion
  if (updateError) {
    return false;
  } 
  return true;
```

AnanyaJha (approved):
Small comment, otherwise it looks good!

AnanyaJha:
What's the behavior if you don't have any checkpoints? Do we still need to run this command? Is it possible to check whether checkpoints are present beforehand?

violetyao:
> What's the behavior if you don't have any checkpoints? Do we still need to run this command? Is it possible to check whether checkpoints are present beforehand?

Thank you for reviewing, Ananya! We discussed in one of our daily syncs and decided to close @W-7748717@ because the debug & run command should actually not give warnings (should just run through the test) if there is no checkpoint/breakpoint. 

We use `SFDX: Update Checkpoints in Org` to check if there is any checkpoint available and to upload checkpoints if they exist. If there is no checkpoint, the behavior of the debug & run button is the same with directly running tests using the green run button, except that those `SFDX: Update Checkpoints in Org` are printed. I've attached a screenshot below!

Screen Shot 2021-07-21 at 12 57 58 PM

AnanyaJha:
> > What's the behavior if you don't have any checkpoints? Do we still need to run this command? Is it possible to check whether checkpoints are present beforehand?
> 
> Thank you for reviewing, Ananya! We discussed in one of our daily syncs and decided to close @W-7748717@ because the debug & run command should actually not give warnings (should just run through the test) if there is no checkpoint/breakpoint.
> 
> We use `SFDX: Update Checkpoints in Org` to check if there is any checkpoint available and to upload checkpoints if they exist. If there is no checkpoint, the behavior of the debug & run button is the same with directly running tests using the green run button, except that those `SFDX: Update Checkpoints in Org` messages are printed. I've attached a screenshot below!
> 
> Screen Shot 2021-07-21 at 12 57 58 PM

Awesome thank you for reminding me! Is it possible for us to exit out of the Uploading Checkpoints process once we discover that there aren't any checkpoints so we can avoid printing the rest of the messages as well?

violetyao:
Great point! I agree that the messages are confusing if there is no checkpoint available. I've added a `hasOneOrMoreActiveCheckpoints` function and attached a screenshot of the outputs if there is no checkpoint available.
Screen Shot 2021-07-21 at 3 08 59 PM

Let me know what you think! If the change looks good to you, let's have Sonal review the `no_enabled_checkpoints` message wording.

AnanyaJha:
> Great point! I agree that the messages are confusing if there is no checkpoint available. I've added a `hasOneOrMoreActiveCheckpoints` function and attached a screenshot of the outputs if there is no checkpoint available.
> Screen Shot 2021-07-21 at 3 08 59 PM
> 
> Let me know what you think! If the change looks good to you, let's have Sonal review the `no_enabled_checkpoints` message wording.

Do you think the user needs to be notified that they don't have any checkpoints at all? @jag-sfdc thoughts? If anything, I think it should be a warning modal, not an error one.

violetyao:
> > Great point! I agree that the messages are confusing if there is no checkpoint available. I've added a `hasOneOrMoreActiveCheckpoints` function and attached a screenshot of the outputs if there is no checkpoint available.
> > Screen Shot 2021-07-21 at 3 08 59 PM
> > Let me know what you think! If the change looks good to you, let's have Sonal review the `no_enabled_checkpoints` message wording.
> 
> Do you think the user needs to be notified that they don't have any checkpoints at all? @jag-sfdc thoughts? If anything, I think it should be a warning modal, not an error one.

I'm leaning towards still giving users warnings if they don't have any checkpoints at all. It might be confusing to users when they try to`debug & run` and expect that tests stop at checkpoints. I agree that a warning modal is better than an error. Let us know what you think Jag!

packages/salesforcedx-vscode-apex-replay-debugger/src/messages/i18n.ts
```diff
@@ -21,6 +21,8 @@ export const messages = {
     'Apex language server could not provide information about valid breakpoints.',
   up_to_five_checkpoints:
     'You have %s of the maximum 5 enabled checkpoints. Remove or disable checkpoints until 5 or fewer are active.',
+  no_enabled_checkpoints:
```
violetyao, line 24: We decided in our demo that we would like to keep the warning. @sbudhirajadoc Could you help us review the `no_enabled_checkpoints` message? Thanks!

sbudhirajadoc: Can you change it to "You don't have any checkpoints enabled." 
Also, are they required to take a followup action that might require saying, "Enable a checkpoint and try again." ?

violetyao: Got it! Users are still able to run through tests without any checkpoints enabled, so **not** having an "Enable a checkpoint and try again" message makes more sense to me. @AnanyaJha @jag-sfdc Let us know what you think!

violetyao: We decided in daily sync that we should add "Enable a checkpoint and try again"! `i18n.ts` has been updated.

PR #3424, opened 2021-07-16
fix: support child components in correlateResults

maxcamp:
### What does this PR do?
This PR allows child components to work for conflict detection. Child components do not have their own FileProperties objects and thus have to use their parents'.

@W-9608415@

### Functionality Before
Deploying a single or multiple child components without their corresponding parent would never be able to detect conflicts.

### Functionality After
Conflicts can be detected in child components, but they use their parents' lastModifed timestamps.

Testing this PR made me realize a piece of strange behavior for conflict detection, which is that extra conflicts can result from timestamps existing at the component level and not the file level. For example: a conflict exists in CustomField X and CustomField Y from the same CustomObject is deployed with a local change. A conflict will be detected in Y because the conflict in X caused the CustomObject's timestamps to have a disparity.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
+++ b/packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
@@ -6,11 +6,12 @@
  */
 import {
   ComponentSet,
+  FileProperties,
   MetadataApiRetrieve,
   RetrieveResult,
   SourceComponent
 } from '@salesforce/source-deploy-retrieve';
-import { FileProperties } from '@salesforce/source-deploy-retrieve';
+import { RecompositionState } from '@salesforce/source-deploy-retrieve/lib/src/convert/convertContext';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
@@ -47,6 +48,11 @@ export interface CorrelatedComponent {
   cacheComponent: SourceComponent;
   projectComponent: SourceComponent;
   lastModifiedDate: string;
+}
+
+interface RecomposedComponent {
+  component?: SourceComponent;
+  children: Map<string, SourceComponent>;
 }
 
 export class MetadataCacheService {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -49,6 +50,11 @@ export interface CorrelatedComponent {
   lastModifiedDate: string;
 }
 
+interface RecomposedComponent {
+  component?: SourceComponent;
+  children: Map<string, SourceComponent>;
```
jag-sfdc, line 55: Why is component optional, but not children? Shouldn't this be the other way? We could simplify the code in line 283 just storing the sourcecomponent.

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -252,21 +258,64 @@ export class MetadataCacheService {
...
+        const parentEntry = projectIndex.get(parentKey);
+        if (parentEntry) {
+          parentEntry.children.set(key, comp);
+        } else {
+          projectIndex.set(parentKey, {
+            children: new Map<string, SourceComponent>().set(key, comp)
+          });
+        }
+      } else {
+        const entry = projectIndex.get(key);
```
jag-sfdc, line 283: Reference to the earlier comment. Can we simply do a projectIndex.set(key, sourceComponent) and have no child components?

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -252,21 +258,64 @@ export class MetadataCacheService {
...
   public static correlateResults(result: MetadataCacheResult): CorrelatedComponent[] {
     const components: CorrelatedComponent[] = [];
 
-    const projectIndex = new Map<string, SourceComponent>();
+    const projectIndex = new Map<string, RecomposedComponent>();
     for (const comp of result.project.components) {
-      projectIndex.set(MetadataCacheService.makeKey(comp.type.name, comp.fullName), comp);
+      const key = MetadataCacheService.makeKey(comp.type.name, comp.fullName);
+      if (comp.parent) {
+        const parentKey = MetadataCacheService.makeKey(comp.parent.type.name, comp.parent.fullName);
```
jag-sfdc, line 273: The next few lines are quite complex and will help to add why we are doing this. Even better if we can somehow refactor it into a method on its own (and reuse for cacheIndex).

maxcamp: I agree, will do that!! I added a bunch of comments in the latest commit, let me know what you think

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
maxcamp: The reason why component is optional is because we could have just the children and not the parent in the deploy. We could make the component non-optional but then I would want to add another field that is something like `parentIsInDeploy: boolean`.

Children could be optional. We would need to include a null check when adding to it, but that would be fine.

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
maxcamp: See above. But also sometimes we may have an entry which has children stored but does not have a parent yet. So in that case we still need to check whether the entry exists in order to not overwrite the children.

So in conclusion do you think both fields should be optional? Or leave as is? Or something else?

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -278,17 +281,67 @@ export class MetadataCacheService {
...
+          // Add the child component if we have an entry for the parent
+          parentEntry.children.set(key, comp);
+        } else {
+          // Create a new entry that does not have a parent yet
+          index.set(parentKey, {
+            children: new Map<string, SourceComponent>().set(key, comp)
+          });
+        }
+      } else {
+        const entry = index.get(key);
```
AnanyaJha, line 330: Would the component (since it's not a child component) not already be associated with the entry in the index?

AnanyaJha, line 330: how would the entry already be in the index at this point?

maxcamp: The input to this function, the SourceComponent[], could contain several parents and several children in an unknown order. An entry for a parent would exist if we had already seen one or more of its children (the entry would be created on line 325).

AnanyaJha (approved):

jag-sfdc (approved):

PR #3429, opened 2021-07-19
Use `functions-core` lib for `Create Function` command

YousefHaggyHeroku:
### What does this PR do?
Replace CLI usage with library for `Create Function`, reducing execution time greatly. Additionally, add typescript support
@W-9620937@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/types/index.ts
+++ b/packages/salesforcedx-utils-vscode/src/types/index.ts
@@ -88,6 +88,7 @@ export type LocalComponent = DirFileNameSelection & {
   suffix?: string;
 };
 
-export type FunctionInfo = DirFileNameSelection & {
+export type FunctionInfo = {
+  fileName: string;
   language: string;
 };
--- a/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
@@ -42,11 +42,7 @@ export class ForceFunctionCreateExecutor extends LibraryCommandletExecutor<
   any
 > {
   constructor() {
-    super(
-      nls.localize('force_function_create_text'),
-      'force_function_create',
-      OUTPUT_CHANNEL
-    );
+    super(nls.localize('force_function_create_text'), LOG_NAME, OUTPUT_CHANNEL);
   }
   public async run(response: ContinueResponse<FunctionInfo>): Promise<boolean> {
     const { fileName, language } = response.data;
--- a/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
@@ -82,37 +82,32 @@ export class ForceFunctionCreateExecutor extends LibraryCommandletExecutor<
     const document = await vscode.workspace.openTextDocument(outputFile);
     vscode.window.showTextDocument(document);
     channelService.appendLine('Installing dependencies...');
+
     if (language === LANGUAGE_JAVA) {
-      return new Promise((resolve, reject) => {
-        cp.exec('mvn install', { cwd: path.join(functionPath) }, err => {
-          if (err) {
-            notificationService.showWarningMessage(
-              nls.localize(
-                'force_function_install_mvn_dependencies_error',
-                err.message
-              )
-            );
-            reject(err);
-          }
-          resolve(true);
-        });
+      cp.exec('mvn install', { cwd: path.join(functionPath) }, err => {
+        if (err) {
+          notificationService.showWarningMessage(
+            nls.localize(
+              'force_function_install_mvn_dependencies_error',
+              err.message
+            )
+          );
+        }
       });
     } else {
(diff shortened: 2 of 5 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceFunctionCreate.test.ts
```diff
@@ -10,103 +10,141 @@ import * as cp from 'child_process';
...
+  let execStub: SinonStub;
+  const sandbox = createSandbox();
+  let notificationServiceStub: SinonStub;
+  let telemetryServiceStub: SinonStub;
+  let withProgressStub: SinonStub;
+  let openTextDocumentStub: SinonStub;
+  let showTextDocumentStub: SinonStub;
+  let gatherStub: SinonStub;
+  const functionInfoJS = {
+    type: 'CONTINUE' as 'CONTINUE',
```
YousefHaggyHeroku, line 41: This hurts my eyes a little bit, but I get a typescript error when I don't include `as CONTINUE`

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -22,47 +19,72 @@ import {
...
-
-  public runPostCommandTasks(data: FunctionInfo) {
-    const language = data.language;
+  public async run(response: ContinueResponse<FunctionInfo>): Promise<boolean> {
+    const { fileName, language } = response.data;
+    let metadata: MetadataInfo | undefined;
+    switch (language) {
+      case LANGUAGE_JAVASCRIPT:
+        metadata = MetadataDictionary.getInfo(FUNCTION_TYPE_JS);
+        metadata!.suffix = '.js';
```
YousefHaggyHeroku, line 53: I don't think we need to use metadata!suffix. A `const fileExtension` or something seems cleaner to me. Thoughts?

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -22,47 +19,72 @@ import {
...
+    if (welcomeText) channelService.appendLine(welcomeText);
+    channelService.showChannelOutput();
+    const outputFile = metadata!.pathStrategy.getPathToSource(
+      functionPath,
+      fileName,
+      metadata!.suffix
+    );
+    const document = await vscode.workspace.openTextDocument(outputFile);
+    vscode.window.showTextDocument(document);
+    channelService.appendLine('Installing dependencies...');
```
YousefHaggyHeroku, line 84: I wonder if it's also worth using `spawn` instead of `exec` below and streaming the install output so users can see specifically what is going on

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: Agree! But, all the definitions in metadataDictionary (apexclass, triggers etc) all use suffix and this will be a much bigger refactor. I'm fine with that if you still want to, but can we make it a separate PR?

packages/salesforcedx-utils-vscode/src/types/index.ts
```diff
@@ -88,6 +88,7 @@ export type LocalComponent = DirFileNameSelection & {
   suffix?: string;
 };
 
-export type FunctionInfo = DirFileNameSelection & {
+export type FunctionInfo = {
+  fileName: string;
```
jag-sfdc, line 92: Why don't we need the directory path anymore?

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
jag-sfdc: In this particular case, product wanted this to be a background process with limited user visibility and we don't show any notification popups either.

packages/salesforcedx-utils-vscode/src/types/index.ts
jag-sfdc: Ok, so generateFunction takes language and figures out the path itself it looks like.

YousefHaggyHeroku: I think previously we used `DirFileNameSelection` which provided `outputDir` in order to satisfy the requirements of `BaseTemplateCommand`, but it was never actually set.

packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
YousefHaggyHeroku: Ah I seem should I remove this line as well then?

jag-sfdc: As discussed on slack, this can remain as is.

jag-sfdc (approved):

YousefHaggyHeroku:
Merged by https://github.com/forcedotcom/salesforcedx-vscode/pull/3476

PR #3435, opened 2021-07-21
feat: introduce 3 Apex Test Suite commands

AnanyaJha:
### What does this PR do?

This PR introduces Test Suites to our command palette. It adds the following commands:

- `SFDX: Run Apex Test Suite`
   - Allows the user to choose an Apex Test Suite via quickpick. The test suite is run through the same code path as the test sidebar & the results are displayed after the test finishes running.
- `SFDX: Create Apex Test Suite`
   - User enters the name of the new Apex Test Suite & selects tests to add to the suite via a quickpick. After selection, the command is run.
- `SFDX: Add Tests to Apex Test Suite`
   - User chooses an existing Apex Test Suite and then selects tests to add to the select suite via the quickpick. After the selection, the command is run.

@W-9632080@
@W-9632072@
@W-9632048@

### Functionality Before
new functionality!

### Functionality After

https://user-images.githubusercontent.com/21326913/126533374-5df40b6b-4f24-4f2f-9fbe-104bb4a5c6ae.mov

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.nls.json
+++ b/packages/salesforcedx-vscode-apex/package.nls.json
@@ -17,5 +17,8 @@
   "force_apex_test_last_method_run_text": "SFDX: Re-Run Last Run Apex Test Method",
   "force_apex_test_class_run_text": "SFDX: Run Apex Test Class",
   "force_apex_test_method_run_text": "SFDX: Run Apex Test Method",
-  "force_apex_test_run_text": "SFDX: Run Apex Tests"
+  "force_apex_test_run_text": "SFDX: Run Apex Tests",
+  "force_apex_test_suite_run_text": "SFDX: Run Apex Test Suite",
+  "force_apex_test_suite_create_text": "SFDX: Create Apex Test Suite",
+  "force_apex_test_suite_build_text": "SFDX: Add Tests to Apex Test Suite"
 }
--- a/packages/salesforcedx-vscode-apex/src/commands/forceApexTestSuite.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceApexTestSuite.ts
@@ -1,24 +1,24 @@
 import { TestService } from '@salesforce/apex-node';
 import {
-  CancelResponse,
-  ContinueResponse,
-  ParametersGatherer
-} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
-import {
   LibraryCommandletExecutor,
   SfdxCommandlet,
   SfdxWorkspaceChecker
 } from '@salesforce/salesforcedx-utils-vscode/out/src';
+import {
+  CancelResponse,
+  ContinueResponse,
+  ParametersGatherer
+} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import { readFileSync } from 'fs';
+import { basename } from 'path';
+import * as vscode from 'vscode';
+import { OUTPUT_CHANNEL } from '../channels';
 import { workspaceContext } from '../context';
 import {
   ApexLibraryTestRunExecutor,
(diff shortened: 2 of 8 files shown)


randi274:
Wow, awesome to see this added to the product! 🎉🎉

packages/salesforcedx-vscode-apex/src/commands/forceApexTestSuite.ts
```diff
@@ -0,0 +1,204 @@
...
+async function listApexTestSuiteItems(): Promise<ApexTestQuickPickItem[]> {
+  const connection = await workspaceContext.getConnection();
+  const testService = new TestService(connection);
+  const testSuites = await testService.retrieveAllSuites();
+
+  const quickPickItems = testSuites.map(testSuite => {
+    return {
+      label: testSuite.TestSuiteName,
+      // @ts-ignore
+      description: testSuite.Id,
```
AnanyaJha, line 60: The field is named incorrectly in the apex-node package so we'll need to update that so we can remove the `ts-ignore` here

packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -17,5 +17,8 @@
   "force_apex_test_last_method_run_text": "SFDX: Re-Run Last Run Apex Test Method",
   "force_apex_test_class_run_text": "SFDX: Run Apex Test Class",
   "force_apex_test_method_run_text": "SFDX: Run Apex Test Method",
-  "force_apex_test_run_text": "SFDX: Run Apex Tests"
+  "force_apex_test_run_text": "SFDX: Run Apex Tests",
+  "force_apex_test_suite_run_text": "SFDX: Run Apex Test Suite",
```
xyc, line 21: please have @sbudhirajadoc review this

packages/salesforcedx-vscode-apex/src/commands/forceApexTestSuite.ts
```diff
@@ -0,0 +1,204 @@
...
+} from './forceApexTestRun';
+
+export type ApexTestSuiteOptions = { suitename: string; tests: string[] };
+
+async function listApexClassItems(): Promise<ApexTestQuickPickItem[]> {
+  const apexClasses = await vscode.workspace.findFiles('**/*.cls');
+  const apexClassItems: ApexTestQuickPickItem[] = [];
+
+  apexClasses.forEach(apexClass => {
+    const fileContent = readFileSync(apexClass.fsPath).toString();
```
xyc, line 38: We should keep in mind the perf impact of this. I tested in perf-project and it took ~1s for loading the menu.

xyc: A change at this moment might not be needed, but if there's performance issues we should consider if we can reuse information from language server.

packages/salesforcedx-vscode-apex/package.nls.json
AnanyaJha: the messages were approved beforehand in a separate doc! thanks sonal!

packages/salesforcedx-vscode-apex/src/commands/forceApexTestSuite.ts
AnanyaJha: I agree - I followed the pattern we're using [elsewhere](https://github.com/forcedotcom/salesforcedx-vscode/blob/5696483cc6e5dd71767312a9130ab8ddb1982fce/packages/salesforcedx-vscode-apex/src/commands/forceApexTestRun.ts#L82), but I think both should be examined to see if we can find a better approach.

xyc (approved):

PR #3476, opened 2021-08-05
Pr/3429 use functions lib to create functions

rcoringrato-sfdc:
### What does this PR do?

Replace CLI usage with library for Create Function, reducing execution time greatly. Additionally, add typescript support

@W-9709809@


packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -127,7 +124,7 @@ export class FunctionInfoGatherer implements ParametersGatherer<FunctionInfo> {
     }
 
     const language = await vscode.window.showQuickPick(
-      [LANGUAGE_JAVA, LANGUAGE_JAVASCRIPT],
+      [LANGUAGE_JAVA, LANGUAGE_JAVASCRIPT, LANGUAGE_TYPESCRIPT],
```
jag-sfdc, line 127: I missed to see this in Yousef's review. Don't think we want to support typescript yet. This can be removed.

rcoringrato-sfdc: @jag-sfdc Can you confirm that we'll keep it as is?

jag-sfdc: Yes, we will keep this as is.

jag-sfdc (approved):

PR #3481, opened 2021-08-06
Mco/test updates

maxcamp:
### What does this PR do?
This PR adds many testing constants from SDR to vscode-core. Along with moving the constants, this PR uses them in 6 different test files. More information about this work can be found in https://salesforce.quip.com/xJrPAdrFHheo including next steps. Here are some more advantages of having the test components available:
- Less code. Creating SourceComponents by hand takes effort and results in a lot of duplicated code within the tests.
- More abstraction. Tests in the past used specific examples of components (such as an ApexClass component) where now tests can use abstract categories of components (such as a MatchingContentFile component). These abstract categories exactly match with the different categories of functionality defined by SDR.
- Easier edge case testing. The components in mock/registry can be set up to account for edge cases and then easily used. For example, to test conflict detection with a child component we could now just import decomposed.DECOMPOSED_CHILD_COMPONENT_1.

@W-9662066@

### Functionality Before and After
The tests themselves should have the same functionality before and after.

### DIFF
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
@@ -44,6 +44,11 @@ import { nls } from '../../../src/messages';
 import { DeployQueue } from '../../../src/settings';
 import { SfdxPackageDirectories } from '../../../src/sfdxProject';
 import { getRootWorkspacePath } from '../../../src/util';
+import {
+  decomposed,
+  matchingContentFile,
+  mockRegistry
+} from '../mock/registry';
 import { MockContext } from '../telemetry/MockContext';
 
 const sb = createSandbox();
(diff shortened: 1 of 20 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -44,6 +44,11 @@ import { nls } from '../../../src/messages';
 import { DeployQueue } from '../../../src/settings';
 import { SfdxPackageDirectories } from '../../../src/sfdxProject';
 import { getRootWorkspacePath } from '../../../src/util';
+import {
```
jag-sfdc, line 47: Apart from changing these imports, do you see any other work once the library has been published?

violetyao: It's ready to go once we update those imports!

jag-sfdc (approved):

PR #3182, opened 2021-04-28
Paragraph title update: Update replay-debugger.md

eyewell:
### What does this PR do?
Changed line 54, since the enabling of automatic logging is not based on whether the org is the "default dev org", but instead, on whether the org is a DE or Scratch Org.

### What issues does this PR fix or reference?
#<Insert GitHub Issue>, @<Insert GUS WI>@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

PR #3551, opened 2021-09-20
Pr/3182 - Paragraph title update: Update replay-debugger.md

randi274:
### What does this PR do?
Improved documentation to reflect that since the enabling of automatic logging is not based on whether the org is the "default dev org", but instead, on whether the org is a DE or Scratch Org.

#3182 

### Functionality Before
Docs needed a small improvement.

### Functionality After
Docs are more accurate now 🙌


sbudhirajadoc (approved):

randi274:
@eyewell can you try one more time to sign the CLA from this item? I was hopeful it would record after the other ticket, but let's give it one more try here. If that doesn't work, I'll follow up with PDT to see what's happening.

PR #3581, opened 2021-10-12
fix: set first alias when config default org

### LINKED ISSUE #3579: "Set a Default Org" command fails with a username with more than one alias
### Summary

Sometimes multiple aliases are created for the same username of an org. But this doesn't work well when selecting these aliases from vscode.

### Steps To Reproduce:

1.  Create more than one alias for a username of an org.
1.  Use "SFDX: Set a Default Org" command from vscode for the previous username with more than one alias.

### Expected result

Should select the username or one of the aliases as default org.

### Actual result

An error occurs when trying to set the concatenation of the different aliases with commas as the default user of an org.
`ERROR running config:set:  No AuthInfo found for name DevHub test,DevHub test2`

### Additional information

image

**Salesforce Extension Version in VS Code**: 52.17.0

**SFDX CLI Version**: sfdx-cli/7.121.8 linux-x64 node-v16.7.0

**OS and version**: Ubuntu 20.04.2 LTS

DanielCalle:
Avoid error when a username has more than one alias by using the first one

fix #3579

### What does this PR do?
Use the first alias, separating it by a comma in case there is more than one for the same username. In other cases it works in the previous way. 

#3579

### Functionality Before
If there was more than one alias for a user, an error occurred when trying to set it by default 
image

### Functionality After
If there is more than one alias for the username, it will not produce an error and will select the first one as the username of a default org

### DIFF
--- a/docs/_articles/en/soql/soql-builder.md
+++ b/docs/_articles/en/soql/soql-builder.md
@@ -1,6 +1,4 @@
 ---
-
----
 title: SOQL Builder
 lang: en
 ---


DanielCalle:
Thanks @AnanyaJha ! Yes, no problem. I have already added a test for this scenario

PR #3608, opened 2021-10-22
chore: add npm ci command

AnanyaJha:
### What does this PR do?
This PR confirms the add of the `package-lock.json` file to our project and switches our circle ci config to start using the `npm ci` command [to install modules in a ci environment](https://docs.npmjs.com/cli/v7/commands/npm-ci) using the lockfile

@W-9953863@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ commands:
           shell: bash.exe
           command: |
             npm config set scripts-prepend-node-path true
-            npm install
+            npm ci
             npm install -g sfdx-cli@<< pipeline.parameters.cli-version >>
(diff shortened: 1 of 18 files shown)


jag-sfdc (approved):

AnanyaJha:
It looks like the package-lock file is not being generated for each individual package unfortunately because of limitations with lerna. Experimented with the following workaround: https://stackoverflow.com/questions/59768340/lerna-not-generating-package-lock-json-for-every-package, but there's some errors related to the installation of packages not published on the npm registry. We also need to investigate whether this method adds any additional time to installation in the builds. Closing this PR for now.

PR #3621, opened 2021-10-27
Added ability to deploy and retrieve multiple files

jeffb-sfdc:
### What does this PR do?
Adds the ability to deploy multiple files when multiple files are selected

@https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000E53RuYAJ/view@

### Functionality Before
One could select multiple files, but performing a deploy would only deploy a single file.

### Functionality After
Now, all selected files & folders are deployed.


packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -40,10 +45,11 @@ export class LibraryRetrieveSourcePathExecutor extends RetrieveExecutor<
   }
 
   public async getComponents(
-    response: ContinueResponse<string>
+    response: ContinueResponse<string | string[]>
```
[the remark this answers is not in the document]
jeffb-sfdc: yep, done.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceDeploySourcePath.test.ts
```diff
@@ -8,15 +8,17 @@
...
+} from '@salesforce/source-deploy-retrieve';
 import { expect } from 'chai';
 import * as path from 'path';
 import { createSandbox, SinonStub } from 'sinon';
-import {
-  LibraryDeploySourcePathExecutor
-} from '../../../src/commands';
+import * as vscode from 'vscode';
+import { LibraryDeploySourcePathExecutor } from '../../../src/commands';
+import * as forceSourceDeploySourcePath from '../../../src/commands/forceSourceDeploySourcePath';
```
jeffb-sfdc, line 20: Just a FYI... I ran the liter and it applied these formatting changes.  It wasn't something I explicitly did.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -51,21 +56,42 @@ export class LibraryDeploySourcePathExecutor extends DeployExecutor<
...
+    new SfdxWorkspaceChecker(),
+    new LibraryPathsGatherer(uris),
+    new LibraryDeploySourcePathExecutor(),
+    new TimestampConflictChecker(false, messages)
+  );
+  await commandlet.run();
+};
 
-export async function forceSourceDeploySourcePath(sourceUri: vscode.Uri) {
+export const forceSourceDeploySingleSourcePath = async (sourceUri: vscode.Uri) => {
```
jeffb-sfdc, line 94: Randi & I discovered that in order to get the stubs and spies to work, the syntax of the function declarations of the static functions needed to be changed.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -84,7 +90,42 @@ export class SourcePathChecker implements PostconditionChecker<string> {
...
+  } else {
+    await forceSourceRetrieveSingleSourcePath(explorerPath);
+  }
+};
+
+export const forceSourceRetrieveMultipleSourcePaths = async (uris: vscode.Uri[]) => {
+  const messages: ConflictDetectionMessages = {
+    warningMessageKey: 'conflict_detect_conflicts_during_retrieve',
+    commandHint: input => {
+      return new SfdxCommandBuilder()
```
jag-sfdc, line 111: I've seen this in other places as well. Since we are using LibraryRetrieveSourcePathExecutor, is this hint correct? Should add what SDR would expect or simply leave this empty?

jeffb-sfdc: Thanks for bringing this up.  I've been experimenting with this and here's what I found:
* If I leave empty (and use `commandHint: input => ""`, in the terminal, the user sees this:
 ```


Conflicts:
    Found 1 file(s) in conflict:

BookTests.cls

Run the following command to overwrite the conflicts:
```

The message reported to the user says, "Run the following" but yet nothing follows, so I think we should give them the commands to manually run if they wish.
* We're using LibraryRetrieveSourcePathExecutor, but the user doesn't have direct access to this, and would need to use sfdx in their terminal, so my preference would be to give them the commands.
* While looking into this, I found a build (which I introduced) pertaining to multiple files.  I pushed the update to the forceSourceDeployMultipleSourcePaths, but not forceSourceRetrieveMultipleSourcePaths just yet (so we don't loose this conversation - once we decide on want to do, I'll get the approach in both forceSourceDeployMultipleSourcePaths and forceSourceRetrieveMultipleSourcePaths to match.

Comment (jag-sfdc) [reply packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts:111]: Thanks for explaining this. This looks good as it looks now 👍

File: packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts [hunk 5]
```diff
@@ -55,66 +46,43 @@ export class LibraryRetrieveSourcePathExecutor extends RetrieveExecutor<
...
-    await forceSourceRetrieveSingleSourcePath(explorerPath);
+  // When a single file is selected and "Retrieve Source from Org" is executed,
+  // sourceUri is passed, and the uris array contains a single element, the same
+  // path as sourceUri.
+  //
+  // When multiple files are selected and "Retrieve Source from Org" is executed,
+  // sourceUri is passed, and is the path to the first selected file, and the uris
+  // array contains an array of all paths that were selected.
+  //
+  // When editing a file and "Retrieve This Source from Org" is executed,
```
jag-sfdc, line 61: Good catch! I did not test that path.

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -40,75 +36,60 @@ export class LibraryRetrieveSourcePathExecutor extends RetrieveExecutor<
...
     }
-  }
+  };
 
   const commandlet = new SfdxCommandlet(
     new SfdxWorkspaceChecker(),
-    new FilePathGatherer(explorerPath),
+    new LibraryPathsGatherer(uris),
     new LibraryRetrieveSourcePathExecutor(),
-    new SourcePathChecker()
```
jeffb-sfdc, line 111: @jag-sfdc I removed SourcePathChecker() and replaced with TimestampConflictChecker().  This seems to work but want to make you aware of this change.

jag-sfdc: @jeffb-sfdc - At one time we had both SourcePathChecker() and TimeStampConflictChecker() as in this PR- https://github.com/forcedotcom/salesforcedx-vscode/pull/3348/files. We then removed TimeStampConflictChecker likely to fix a bug. I'll dig through the PRs to understand why we did this. If this isn't a required fix for multi file deploy/retrieve to work, can we not change it now.

jeffb-sfdc: It's not a requirement.  I'll add it back in.  I removed it b/c TimestampConflictChecker() worked for both deploy & retrieved, and thought I'd consolidate.

jag-sfdc: @jeffb-sfdc - I was able to track down the reason. We decided to disable conflict detection for retrieves and pushed a PR for it - https://github.com/forcedotcom/salesforcedx-vscode/pull/3382. We should leave this with SourcePathChecker() for now.

jeffb-sfdc: Thanks, I reverted it back.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveSourcePath.test.ts
```diff
@@ -97,11 +109,87 @@ describe('Force Source Retrieve with Sourcepath Option', () => {
...
         type: 'CONTINUE',
-        data
-      } as ContinueResponse<string>;
+        data: [data]
+      } as ContinueResponse<string[]>;
       const componentSet = executor.getComponents(continueResponse);
       expect((await componentSet).sourceApiVersion).to.equal('11.0');
     });
+
+    it('validates the condition of when multiple files are retrieved', async () => {
```
jag-sfdc, line 118: Apologies I didn't notice this before. As a standard, we want to prefix tests with "should". Everything else looks good. Please merge the PR once you fix this.

PR #3650, opened 2021-11-16
feat: additional anonymous apex features

randi274:
### What does this PR do?
Enables Hover, Definition, and References in anonymous apex files.

@W-10178822@

### Functionality Before
Anonymous Apex files didn't interact much with the language server.

### Functionality After
Users can hover over variables and methods for more information, go to the definition of another class, and see References all from within Anonymous Apex.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docs",
-  "version": "53.3.0",
+  "version": "53.4.0",
   "description": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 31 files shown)


jag-sfdc:
Curious why is there a 1.11MB file size reduction?

jag-sfdc:
Ah! Looks like the jar is not signed.

randi274:
Good catch @jag-sfdc! I went through the process to sign it, but must have missed something. Will review with @rcoringrato-sfdc today.

rcoringrato-sfdc (approved):

PR #3660, opened 2021-11-19
Fb/org browser multi select

jeffb-sfdc:
### What does this PR do?
Adds the ability to deploy multiple files when multiple files are selected

@https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000E53RuYAJ/view@

### Functionality Before
One could select multiple files, but performing a deploy would only deploy a single file.

### Functionality After
Now, all selected files & folders are deployed.


packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -51,79 +49,85 @@ export class LibraryDeploySourcePathExecutor extends DeployExecutor<
...
-  }
-}
-
-export async function forceSourceDeploySourcePath(sourceUri: vscode.Uri) {
+export const forceSourceDeploySourcePaths = async (
+  sourceUri: vscode.Uri | vscode.Uri[] |undefined,
+  uris: vscode.Uri[] | undefined
+) => {
   if (!sourceUri) {
-    const editor = vscode.window.activeTextEditor;
```
jeffb-sfdc, line 70: @jag-sfdc I think what happened is that I originally had (pseudo code):
```
export async function forceSourceDeploySourcePath(sourceUri: vscode.Uri, uris: vscode.Uri[]) {
    if (uris) {
        forceSourceDeployMultipleSourcePaths()
        return;
    }

    // continue with the rest of forceSourceDeploySourcePath()...
```

When I consolidate and just used forceSourceDeployMultipleSourcePaths(), this check (lines 71-86) was removed & wasn't noticed, b/c 1) forceSourceDeployMultipleSourcePaths() didn't have this logic in it, and 2) the paths we exercised, never had an issue with that check, and we never ran into this issue.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -51,79 +49,85 @@ export class LibraryDeploySourcePathExecutor extends DeployExecutor<
...
+      uris = [];
+      uris.push(sourceUri);
     }
-  };
-
-  const commandlet = new SfdxCommandlet(
-    new SfdxWorkspaceChecker(),
-    new FilePathGatherer(sourceUri),
-    new LibraryDeploySourcePathExecutor(),
-    new CompositePostconditionChecker(
```
jeffb-sfdc, line 103: @jag-sfdc The implementations of the old "single" function vs. the old "multiple" function vary.  The single version calls
```
new CompositePostconditionChecker(
  new SourcePathChecker(),
    new TimestampConflictChecker(false, messages)
)
```

...yet the multiple version calls
```
new TimestampConflictChecker(false, messages)
```

We should discuss this on Monday.

jag-sfdc: I think this is a mistake that we skipped this. We should add new SourcePathChecker() to make sure the code is in packageDirectories directory defined in sfdx-project.json

jag-sfdc (approved):

PR #3661, opened 2021-11-20
Retrieve field data for a custom object type

RitamAgrawal:
### What does this PR do?
Migrating away from using sfdx:CLI to retrieve metadata components in Org Browser and using sfdx-core library for the same.
Added a new feature, on clicking any of the CustomObjects metadatatypes(like Account, Campaign etc.) it expands to list its  fields with field properties like name, type and length etc.

W-10003716
W-10065144

### Functionality Before
Fields and its properties were not displayed.

### Functionality After
Screen Shot 2021-11-19 at 7 16 30 PM


jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -84,79 +85,170 @@ export class ComponentUtils {
...
+      );
+      telemetryService.sendEventData('CustomObjects Fields quantity', { sObject }, { fields: fields.length });
+      return fields;
+    } catch (e) {
+      telemetryService.sendException('metadata_cmp_build_custom_objects_fields_list', e.message);
+      throw new Error(e);
+    }
+  }
+
+  public async listMetadataTypes(
```
jag-sfdc, line 125: Nit pick. Though this is calling metadata list, it'll be good to rename this method to something that explains what this does i.e. fetching and saving components of a specific metadataType.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -84,79 +85,170 @@ export class ComponentUtils {
...
+    const metadataQuery = {type: metadataType};
+    const metadataFileProperties = await connection.metadata.list(metadataQuery);
+    const result = {status: 0, result: metadataFileProperties};
+    const jsonResult = JSON.stringify(result);
+    fs.writeFileSync(componentsPath, jsonResult);
+    return jsonResult;
+  }
 
-    return fields;
+  public async listSObjectFields(
```
jag-sfdc, line 138: Same comment as above. Consider to rename this method name.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -84,79 +85,170 @@ export class ComponentUtils {
...
+            case 'email':
+              return `${field.name} (${field.type}(${field.length}))`;
+            case 'reference':
+              return `${field.relationshipName} (reference)`;
+            default:
+              return `${field.name} (${field.type})`;
+          }
+        }
+      );
+      telemetryService.sendEventData('CustomObjects Fields quantity', { sObject }, { fields: fields.length });
```
jag-sfdc, line 117: Great job adding telemetry. I think the field quantity won't be much help here. Can we instead add telemetry to see how many components we get in listMetadataTypes()? Also we want the event name to be on worded - something like orgBrowserMetadataFetched? This would show us this originated from orgBrowser?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
RitamAgrawal: Incorporating this and renamed listMetadataTypes to fetchAndSaveMetadataComponentProperties.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
RitamAgrawal: Renaming listSObjectFields to fetchAndSaveSObjectFieldsProperties.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
RitamAgrawal: Removing the telemetry record as it's not adding that much of value to it.

jag-sfdc (approved):

PR #3671, opened 2021-11-23
chore: adds java telemetry

randi274:
### What does this PR do?
Adds Java telemetry, for the extensions and the LSP.

@W-10165985@, @W-10215296@

### Functionality Before
No Java Telemetry sent

### Functionality After
Sending Java Telemetry


rcoringrato-sfdc (approved):

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -60,6 +60,11 @@ async function createServer(
     if (jvmMaxHeap) {
       args.push(`-Xmx${jvmMaxHeap}M`);
     }
+    telemetryService.sendEventData(
+      'JVM Max Heap',
+      undefined,
```
jag-sfdc, line 65: Curious why property is set to undefined?

rcoringrato-sfdc: Should it be a property instead of a measurement?

jag-sfdc: I'm not sure. I do see "apexLSPStartup" metric being passed like this. But no idea how to query it on AppInsights.

jag-sfdc: Played with this in AppInsights and looks "apexLSPStartup" should be queried as the following ```
union customEvents
| where name == 'salesforcedx-vscode/apexLSPStartup'
```
I think it would be good to change the name from 'JVM Max Heap' to "apexLSPStartup". @randi274 - Let me know if you want to sync up on this tomorrow. We can go over AppInsights together on how this would be queried.

randi274: Ah gotcha! That comes from the LSP, I've seen where that gets passed out of Jorje. I can make that change. Also Undefined is set here since it matched what we have [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/539fa5c214b9182bf888c10878ac547dc3f3d4d5/packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts#L75) for another metric. Let me know if you like the update, if not let's sync on it!

jag-sfdc: Sorry for the back n forth. I think a different name would be better here. How about "apexLSPSettings"? We can leave the property to be undefined.

randi274: I like it! We also use apexLSPLog elsewhere if you like that as a differentiator from Startup.

jag-sfdc (approved):

PR #3680, opened 2021-12-03
chore: add jorje build and sign task

randi274:
### What does this PR do?
Creates a task to automate building and signing the LSP jar.

@W-10244024@

### Functionality Before
Devs were using local scripts to build and sign the LSP jar.

### Functionality After
We're all using the same script!


scripts/build-jorje.js
```diff
@@ -0,0 +1,107 @@
...
+let SFDC_KEYSTORE='';
+let SFDC_KEYPASS='';
+const JORJE_DEST_DIR=`${CURRENT_DIR}/packages/salesforcedx-vscode-apex/out`;
+const JORJE_DEST_PATH=`${JORJE_DEST_DIR}/apex-jorje-lsp.jar`;
+
+function verifyPaths() {
+  console.log(LOG_HEADER);
+  try {
+    JORJE_DEV_DIR = checkJorjeDirectory();
+    fs.statSync(JORJE_DEV_DIR).isDirectory();
```
rcoringrato-sfdc, line 49: @randi274 The return value is not referenced, so this check has no effect, i.e. a non-directory value can be used and it will fail further on.

scripts/build-jorje.js
```diff
@@ -0,0 +1,107 @@
...
+  } catch (error) {
+    logger.error(util.format(NO_JORJE_MSG, JORJE_DEV_DIR));
+    process.exit();
+  }
+  if (needSigning) {
+    ({SFDC_KEYSTORE, SFDC_KEYPASS} = checkSigningAbility());
+    try {
+      fs.statSync(SFDC_KEYSTORE).isFile();
```
rcoringrato-sfdc, line 57: @randi274 The return value is not referenced here.

scripts/build-jorje.js
randi274: It actually does fail if it's not a valid directory (which is an interesting way write a function for a library), but it does lead to another question of: what if it's a directory, but the wrong one? I think I can catch for that later on, and add some better error handling.
Screen Shot 2021-12-06 at 4 14 27 PM

scripts/build-jorje.js
randi274: Same thing here; if it's not a file, an error will be thrown, so as long as it doesn't throw an error, we're a-ok to proceed.

scripts/build-jorje.js
rcoringrato-sfdc: Yes, an invalid file path will cause the check to fail, but if it is a valid file, but not a valid directory then it will continue and fail at **process.chdir(JORJE_DEV_DIR)** in **buildLSP()**.

randi274: I moved things around a bit and think that issue should be resolved now; let me know if you'd like me to be more explicit about handling it though!

rcoringrato-sfdc (approved):

PR #3688, opened 2021-12-07
feat: enable use of anon Apex execution codelens

rcoringrato-sfdc:
### What does this PR do?

Add a code lens for anonymous Apex files (*.apex) for execution of the entire file, or a selected range.

@W-9787744@


randi274 (commented):
Changes are looking great! Was able to test with the selected text, modified but not saved text, and the whole file. I'm not sure the isEmpty is working as expected though. When I highlight an empty line, I run into an issue here where it generates an '<EOF>'. If that's ok though, I'm good to approve.
Screen Shot 2021-12-08 at 2 19 45 PM

rcoringrato-sfdc:
@randi274 I think it is a little bit of a corner case, so its okay.

randi274 (approved):
Awesome addition to the product!

PR #3697, opened 2021-12-14
fix: display retrieve source button

RitamAgrawal:
### What does this PR do?
Provides the `Retrieve Source from Org` button for all custom objects in Org Browser
@W-10304052@

### Functionality Before
Screen Shot 2021-12-14 at 12 41 44 PM

### Functionality After

Screen Shot 2021-12-14 at 12 37 54 PM

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
+++ b/packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
@@ -134,6 +134,7 @@ export class BrowserNode extends vscode.TreeItem
     switch (this.type) {
       case NodeType.MetadataType:
         return RetrieveDescriberFactory.createTypeNodeDescriber(this);
+      case NodeType.Folder:
       case NodeType.MetadataComponent:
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -134,6 +134,7 @@ export class BrowserNode extends vscode.TreeItem
     switch (this.type) {
       case NodeType.MetadataType:
         return RetrieveDescriberFactory.createTypeNodeDescriber(this);
+      case NodeType.Folder:
```
jag-sfdc, line 137: What happens when we do a retrieve on a folder that is not sobject?

RitamAgrawal: It runs the command successfully.
Screen Shot 2021-12-15 at 10 08 21 AM
.

jag-sfdc (approved):

PR #3709, opened 2021-12-18
fix: remove log4j from apex language server

rcoringrato-sfdc:
### What does this PR do?

Build and package Apex Language Server without log4j library.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docs",
-  "version": "53.8.0",
+  "version": "53.8.1",
   "description": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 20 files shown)


trinasfdx:
I’m seeing references to log4j in the pom.xml file here: /apex-jorje-lsp/META-INF/maven/ch.qos.logback/logback-classic. Is this expected @rcoringrato-sfdc?

rcoringrato-sfdc:
@trinasfdx Yes, that is a reference to an slf4j bridge module (log4j-over-slf4j) that would allow logj4 loggers to work with slf4j (described [here](http://www.slf4j.org/legacy.html)). The Apex Language Server (ALS) does not contain any log4j loggers, however, so it is not used, and is not part of the log4j distribution.

randi274 (approved):

PR #2523, opened 2020-09-14
Delete clean org commands

lcampos:
### What does this PR do?
Adds a few label updates to PR #2505 

### What issues does this PR fix or reference?
#945, @W-8065926@

PR #3732, opened 2021-12-22
feat: add ability to remove expired orgs from the org list

randi274:
### What does this PR do?
We already had a command to remove dead scratch orgs (#2523), but that functionality isn't readily
available to users, and it can be frustrating for new users to dig it up. Meanwhile the list of
expired orgs only grows! This change surfaces that functionality in the org list directly, so users
can view the commands that interact with their orgs from one spot.

@W-10351429@

### Functionality Before
The command `SFDX: Remove Deleted and Expired Orgs` was only available from the command palette.

### Functionality After
It's now easier to clean up stale orgs. 🧹

### DIFF
--- a/packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
+++ b/packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
@@ -133,7 +133,8 @@ export class OrgList implements vscode.Disposable {
       '$(plus) ' + nls.localize('force_auth_web_login_authorize_org_text'),
       '$(plus) ' + nls.localize('force_auth_web_login_authorize_dev_hub_text'),
       '$(plus) ' + nls.localize('force_org_create_default_scratch_org_text'),
-      '$(plus) ' + nls.localize('force_auth_access_token_authorize_org_text')
+      '$(plus) ' + nls.localize('force_auth_access_token_authorize_org_text'),
+      '$(plus) ' + nls.localize('force_org_list_clean_text')
     ];
 
     const authInfoList = await this.updateOrgList();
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -172,6 +173,11 @@ export class OrgList implements vscode.Disposable {
         vscode.commands.executeCommand('sfdx.force.auth.accessToken');
         return { type: 'CONTINUE', data: {} };
       }
+      case '$(plus) ' +
```
jeffb-sfdc, line 176: minor point (and nothing to do w/your PR), do you mind changing lines 156-159 to
```
return { type: 'CONTINUE', data: {} };
```
(this will match the other cases)

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
```diff
@@ -330,6 +330,18 @@ describe('Set Default Org', () => {
     ).to.be.true;
   });
 
+  it('should return Continue and call force:org:list:clean command if SFDX: Remove Deleted and Expired Orgs is selected', async () => {
+    orgListStub.returns(orgsList);
```
jeffb-sfdc, line 334: I know this matches the other tests, but this doesn't appear to actually test the code.  When it runs, the ```'$(plus) ' +
        nls.localize('force_auth_web_login_authorize_dev_hub_text')``` case gets hit, not the  nls.localize('force_org_list_clean_text') case.

jeffb-sfdc: I spoke with Randi about this, and this isn't an issue.  This case is supposed to hit the default case, so all is good.

jeffb-sfdc (approved):

PR #3738, opened 2021-12-31
chore: webpack salesforcedx-vscode-core - WIP

jag-sfdc:
### What does this PR do?
Webpack salesforcedx-vscode-core
This is still work in progress


gbockus-sf:
replaced webpack with esbuild so this is no longer valid.

PR #3758, opened 2022-01-12
fix: added replay debugging to ApexLibraryExecuteExecutor

jeffb-sfdc:
### What does this PR do?
This PR adds Apex debug replay to when anonymous Apex is executed.

@W-10288326@

### Functionality Before
When the execute anonymous command was invoked, it would run but the flow was execution only.

### Functionality After
Now, when the execute anonymous command is invoked, the anonymous Apex runs, and then the replay debugger runs.


jeffb-sfdc:
@rcoringrato-sfdc , traceFlags.ts is in `packages/salesforcedx-utils-vscode/src/commands/traceFlags.ts`.  It isn't a command, and seems like the helpers directory (`packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts`) would be more appropriate.  Any objections to it being moved?

rcoringrato-sfdc:
@jeffb-sfdc I missed that initially, that would be a good move.

jag-sfdc (approved):

PR #3784, opened 2022-01-27
feat: manifest builder

randi274:
### What does this PR do?
Adds Manifest Builder functionality.

#@W-10288473@, @W-10288468@

### Functionality Before
Creating a manifest for a project had to be done manually, or at the start of a new project.

### Functionality After
Creating new manifests is easier than ever; multi-select all the files you'd like, then right click to 'Generate Manifest'. Enter a name to save the file, or cancel to view unsaved changes.


packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -64,6 +64,7 @@
   "refresh_components_text": "Refresh Components",
   "force_source_retrieve_display_text": "Retrieve Source from Org",
   "force_source_retrieve_and_open_display_text": "Retrieve and Open Source",
+  "force_create_manifest": "SFDX: Generate Manifest (package.xml)",
```
jag-sfdc, line 67: Has @sbudhirajadoc reviewed this and the other messages?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
```
jag-sfdc, line 2: Can the year be changed to 2021?

randi274: Yes!

packages/salesforcedx-vscode-core/package.nls.json
randi274: No, she was still waiting for the vsix files, so merge will have to wait until Monday when she's back.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+    uris.push(sourceUri);
+  }
+  const sourcePaths = uris.map(uri => uri.fsPath);
+  if (sourcePaths) {
+    const componentSet = ComponentSet.fromSource(sourcePaths);
+    const inputOptions = {
+        placeHolder: nls.localize('manifest_editor_save_placeholder'),
+        prompt: nls.localize('manifest_editor_save_prompt'),
+        validateInput: text => {
+          return text.includes('.xml') ? 'Remove file extension' : null;
```
jag-sfdc, line 31: We don't have "Remove file extension" in other "create" commands. Can we get some feedback from Ananya on this?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jag-sfdc: I'm stuck in 2021. I meant 2022 :)

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+  vscode.workspace.openTextDocument({
+    content: componentSet.getPackageXml(),
+    language: 'xml'
+  }).then(newManifest => {
+    vscode.window.showTextDocument(newManifest);
+  });
+}
+
+function saveDocument(response: string, componentSet: ComponentSet) {
+  const fileName = response ? response.concat('.xml') : 'package.xml';
```
jag-sfdc, line 54: One suggestion will be to make the extension optional when a user enters and add it here when missing. But it'll be good to get Ananya's thoughts on this.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    beforeEach(() => {
+      inputBoxSpy = env.stub(vscode.window, 'showInputBox');
+      openDocSpy = env.spy(vscode.workspace, 'openTextDocument');
+    });
+
+    afterEach(() => {
+      env.restore();
+    });
+
+    it('Should create first manifest from single sourceUri', async () => {
```
jag-sfdc, line 54: Are the uris any different for folders? I am wondering if there should be separate tests for folders and files (single and multiples).

jag-sfdc:
Looks great overall. File extension is the only piece that needs some input from Ananya and then we are good to go 👍

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jeffb-sfdc: HA!  I had 2021 in my latest PR as well.  (I just updated them)

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    let openDocSpy: sinon.SinonSpy;
+    let inputBoxSpy: sinon.SinonStub;
+
+    before(() => {
+      env = createSandbox();
+    });
+
+    beforeEach(() => {
+      inputBoxSpy = env.stub(vscode.window, 'showInputBox');
+      openDocSpy = env.spy(vscode.workspace, 'openTextDocument');
```
jeffb-sfdc, line 47: Just a reminder, to rename to showInputBoxStub, and openTextDocumentSpy

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      expect(writeFileSpy.called).to.equal(false);
+    });
+
+  });
+
+  describe('Exception Handling', () => {
+
+    let env: sinon.SinonSandbox;
+    let openDocSpy: sinon.SinonSpy;
+    let inputBoxSpy: sinon.SinonStub;
```
jeffb-sfdc, line 117: also, these three vars (env, openDocSpy, and ~~inputBoxSpy~~ inputBoxStub) can probably be hoisted up to the topmost describe()

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: I'm glad you clarified, because I'll also be in 2021 until at least February.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: Yes! Also making a note that this should definitely be localized if we use it.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
randi274: Yes, working on that currently!

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: Sure! I scheduled a message to Ananya on this in the team chat for Monday once she's back from PTO.

gbockus-sf: Could we make a common utility for checking for a set of file extensions? 
Something along the lines of 
```
verifyFileExtensions(text: string, invalidExtensions: string[], errorMsg: string): string | null 
```
This could also use `.endsWith` instead of `.includes` so it would allow a filename like `custom.xml.for.contacts`

gbockus-sf: also should we consider casing?  `.XML` also seems like a bummer for case insensitive file systems.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -285,6 +289,10 @@
         {
           "command": "sfdx.force.function.start",
           "when": "sfdx:project_opened && resource =~ /.*/functions/.*/[^/]+(/[^/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
+        },
+        {
+          "command": "sfdx.create.manifest",
+          "when": "sfdx:project_opened"
```
gbockus-sf, line 295: I'm not seeing the option to create manifest in the edited right clicked menu for files in an sfdx project.  Is that what this option is intended to configure? 
Screen Shot 2022-01-28 at 10 38 16 AM

packages/salesforcedx-vscode-core/package.json
```diff
@@ -404,6 +412,10 @@
           "command": "sfdx.force.project.with.manifest.create",
           "when": "!sfdx:internal_dev"
         },
+        {
+          "command": "sfdx.create.manifest",
+          "when": "false"
```
gbockus-sf, line 417: why do  we even include this is we're just going to set the when to false?

packages/salesforcedx-vscode-core/package.json
randi274: I believe that functionality is what's intended here - the right click option to create manifests should only work in the explorer

packages/salesforcedx-vscode-core/package.json
randi274: The syntax here is weird, but it's how we prevent the command from appearing in the command palette.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -811,6 +823,10 @@
           "dark": "resources/dark/code.svg"
         }
       },
+      {
+        "command": "sfdx.create.manifest",
+        "title": "%force_create_manifest%"
```
gbockus-sf, line 828: cool...where does the string replace happen here?  Is it custom or is that something baked into extensions?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
gbockus-sf: Are we using https://marketplace.visualstudio.com/items?itemName=doi.fileheadercomment for auto inserting the copyright? looks like there is a year template value if so.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ComponentSet } from '@salesforce/source-deploy-retrieve';
+import * as fs from 'fs';
+import { join } from 'path';
+import { format } from 'util';
+import * as vscode from 'vscode';
+import { nls } from '../messages';
+import { getRootWorkspacePath } from '../util';
```
gbockus-sf, line 14: do we have a preferred import order?

jeffb-sfdc: It's to be sorted by alphabet (of the name of the module being imported).  Our linter spits out an error if they are alphabetized.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -250,7 +250,10 @@ export const messages = {
   demo_mode_prompt:
     'Authorizing a business or production org is not recommended on a demo or shared machine. If you continue with the authentication, be sure to run "SFDX: Log Out from All Authorized Orgs" when you\'re done using this org.',
   force_auth_logout_all_text: 'SFDX: Log Out from All Authorized Orgs',
+  manifest_editor_dupe_error: 'Manifest with the name %s already exists. Delete this manifest or use another name.',
   force_auth_logout_default_text: 'SFDX: Log Out from Default Org',
+  manifest_editor_save_placeholder: 'Enter unique manifest name (no .xml)',
```
gbockus-sf, line 255: This seems like it should be `package` since that's what we default to if there's not text inputed. We tell them if the validation fails so I think it's fine to not call out that `.xml` isn't allowed until they try to enter it.

packages/salesforcedx-vscode-core/src/orgBrowser/browser.ts
```diff
@@ -44,7 +44,8 @@ export class OrgBrowser {
     const username = await OrgAuthInfo.getDefaultUsernameOrAlias(false);
     this._dataProvider = new MetadataOutlineProvider(username);
     this._treeView = window.createTreeView(OrgBrowser.VIEW_ID, {
-      treeDataProvider: this._dataProvider
+      treeDataProvider: this._dataProvider,
+      canSelectMany: true
```
gbockus-sf, line 48: still coming up to speed. Is there any we're doing with the org view where making this multiselect will be a problem?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -514,6 +515,11 @@ async function setupOrgBrowser(
       await forceSourceRetrieveCmp(trigger, true);
     }
   );
+
+  vscode.commands.registerCommand(
+    'sfdx.create.manifest',
```
gbockus-sf, line 520: Should this be an enum or constant? Having all the commands we support for a particular module live in one place seems like a win.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export async function forceCreateManifest(
+  sourceUri: vscode.Uri,
+  uris: vscode.Uri[] | undefined
+) {
+  if (!uris || uris.length < 1) {
+    uris = [];
+    uris.push(sourceUri);
```
gbockus-sf, line 22: dumb q: what use case is this supporting?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (sourcePaths) {
+    const componentSet = ComponentSet.fromSource(sourcePaths);
+    const inputOptions = {
+        placeHolder: nls.localize('manifest_editor_save_placeholder'),
+        prompt: nls.localize('manifest_editor_save_prompt'),
+        validateInput: text => {
+          return text.includes('.xml') ? 'Remove file extension' : null;
+        }
+    } as vscode.InputBoxOptions;
+    const response = await vscode.window.showInputBox(inputOptions);
```
gbockus-sf, line 34: nit (I call a thing a nit if it's not very important or just an opinion ) 
responseText would be a better name since we expect a string back.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (response === undefined) {
```
gbockus-sf, line 35: any reason you prefer the specific `===` to just `if (!response)`?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -0,0 +1,74 @@
...
+  const saveLocation = join(manifestPath, fileName);
+  checkForDuplicateManifest(saveLocation, fileName);
+
+  fs.writeFileSync(saveLocation, componentSet.getPackageXml());
+  vscode.workspace.openTextDocument(saveLocation).then(newManifest => {
+    vscode.window.showTextDocument(newManifest);
+  });
+}
+
+function checkForDuplicateManifest(saveLocation: string, fileName: string) {
```
gbockus-sf, line 69: This and the previous two functions seem specific to working with files.  Should we have a `FileService` that handles this kind of thing? It'd be less coupled to the manifest building that way and would make testing more straightforward.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+const uri2 = vscode.Uri.parse(classPath + class2);
+const manifest = `<?xml version="1.0" encoding="UTF-8"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+  <types>
+    %s
+    <name>ApexClass</name>
+  </types>
+</Package>`;
+
+describe('Force Create Manifest', () => {
```
gbockus-sf, line 34: We could call out unit tests here.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        }
+      });
+      env.stub(fs, 'existsSync').returns(false);
+      env.stub(fs, 'mkdirSync').returns(undefined);
+      env.stub(fs, 'writeFileSync').returns(undefined);
+      const packageName = 'package' + generateRandomSuffix() + '.xml';
+      inputBoxSpy.onCall(0).returns(packageName);
+      await forceCreateManifest(uri1, undefined);
+
+      expect(openDocSpy.calledOnce).to.equal(true);
```
gbockus-sf, line 68: should we verify that the other stubs were called here with the right parameters? If we open the wrong file in the wrong place that'd be sadness.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      env.restore();
+    });
+
+    it('Should handle exception while creating component set', async () => {
+      env.stub(ComponentSet, 'fromSource').throws(
+        new Error(nls.localize('error_creating_packagexml'))
+      );
+      let exceptionThrown = false;
+      try {
+        await forceCreateManifest(uri1, [uri2]);
```
gbockus-sf, line 138: can we use 
```
const manifestError = new Error(nls.localize('error_creating_packagexml'));
env.stub(ComponentSet, 'fromSource').throws(manifestError);
...
await expect(forceCreateManifest(uri1, [uri2]);).to.throw(manifestError);
```

it'd remove the need for the try/catch and extra checking of if an exception was thrown or not.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        .onFirstCall()
+        .returns(true)
+        .onSecondCall()
+        .returns(false);
+      inputBoxSpy.onCall(0).returns(undefined);
+      await forceCreateManifest(uri1, [uri1, uri2]);
+
+      expect(openDocSpy.calledOnce).to.equal(true);
+      expect(writeFileSpy.called).to.equal(false);
+    });
```
gbockus-sf, line 109: Two test cases to consider: 
1. When the user doesn't supply a name use the default. 
2. When there are no entries in package.xml (like when you do it for a file in the manifests directory it produces an empty file.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        expect(e.message).to.contain(fileName);
+      }
+      expect(exceptionThrown).to.equal(true);
+      expect(openDocSpy.called).to.equal(false);
+    });
+  });
+
+});
+
+function generateRandomSuffix() {
```
gbockus-sf, line 173: nit: I personally prefer to default to arrow function assigned to constants pretty much always.  It removes the need to consider the good ol `scope/this` in which the function is executed in. It's one of those things where it's usually not an issue, but when it is it totally sucks and is hard to debug.

gbockus-sf (commented):
Left probably more comments then necessary...some for my education. 
This code is so _cool_ :)

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: The event that it does become available as a command palette option - @jeffb-sfdc ran into this previously and recommended this just in case.

randi274:
@gbockus-sf thanks for the thorough feedback on this one! lots of great suggestions I’m looking through now 🎉

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jag-sfdc: @gbockus-sf - We don't use it. We just copy it from other files. This extension is a good idea for the whole team to use 👍

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
jag-sfdc: I always thought this was just a style preference. Looks like there is more to this. Can we take it up in one of the elevensies, please?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
gbockus-sf: cool I'll create a ticket and follow up with adding it to our suggested extensions.

packages/salesforcedx-vscode-core/package.json
gbockus-sf: sorry dumb follow up q, but if it's not included here at all it'll how up in the command pallet?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
gbockus-sf: @jag-sfdc totally.

packages/salesforcedx-vscode-core/package.json
randi274: Correct! I had to check my knowledge on this, but here's the doc that confirms: https://code.visualstudio.com/api/extension-guides/command#controlling-when-a-command-shows-up-in-the-command-palette

packages/salesforcedx-vscode-core/package.json
randi274: This is baked into the extensions - it's able to find strings automatically from `package.nls.json`, as shown in this example repo (which I also just confirmed lol): https://github.com/microsoft/vscode-extension-samples/tree/main/i18n-sample

packages/salesforcedx-vscode-core/package.nls.json
randi274: @sbudhirajadoc confirmed that the string here would be better as `SFDX: Generate Manifest File`, so will be changing in the next revision.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
randi274: Feedback from @sbudhirajadoc on the string if we want a heavier hand here: `Enter unique a manifest file name (without file extension)`

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: This pattern is present elsewhere, and undefined has a specific meaning to signify the escape key, so wanted to be clear on that. https://code.visualstudio.com/api/references/vscode-api#window

packages/salesforcedx-vscode-core/src/orgBrowser/browser.ts
randi274: @jag-sfdc do you have any concerns that you can think of here? I'm not seeing any errors with a spot check, but some interesting behavior with Diff and Delete (only takes the selected directory) and Retrieve and Deploy work well together.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
randi274: They are not - confirmed with objects. The important part is that sometimes there is only one URI, and others there are several to work with.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
gbockus-sf: word. That makes sense.  empty string would get through (!response) so the strict equality is def better

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
randi274: Which stubs are you thinking are missing here? The one that I can't seem to add with the stubs in place is showing the text document, since I don't actually have anything to open without the file being created.

packages/salesforcedx-vscode-core/package.json
gbockus-sf: Gotcha.  I'm still burning in the vscode terminology. Is it assumed that if we don't expose a command in the pallet then it also isn't available in the editor context menu?

packages/salesforcedx-vscode-core/package.json
gbockus-sf: 🆒

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
randi274: @gbockus-sf It would be nice, but I think this failing because of the async function. After digging for a while, I found this which was similar to my issue: https://stackoverflow.com/questions/21587122/mocha-chai-expect-to-throw-not-catching-thrown-errors

Which got me to here: https://github.com/chaijs/chai/issues/882

So I think I'd have to write something like .rejected(), but I'm not quite sure how that would work. I'll push what I've got now, and we can discuss more, but I'd like to get this out sooner than get stuck too far in the syntax 😅 (although I would very much like to figure out how to get it to work!)

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: I'm open to this, as I know the team does have some utility functions, but I'm not sure if it would be best to put it in the -core package, or in the -utils (which is something we don't ship, but reference internally).

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: I heard back from Ananya here, and she also voted doing the work to strip it out ourselves. So we can keep the messaging, but I'll remove the validation and handle it downstream.

randi274: @jag-sfdc this has been updated

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        .onSecondCall()
+        .returns(false);
+      showInputBoxStub.onCall(0).returns(undefined);
+      await forceCreateManifest(URI_1, [URI_1, URI_2]);
+
+      expect(openTextDocumentSpy.calledOnce).to.equal(true);
+      expect(writeFileSpy.called).to.equal(false);
+    });
+
+    it('Should use default manifest name if none is supplied', async () => {
```
gbockus-sf, line 111: I'm so glad we use the should descriptions.  Easy on the 👀

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: Created here: W-10522602

packages/salesforcedx-vscode-core/src/index.ts
randi274: Created here: W-10522606

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
randi274: @gbockus-sf I made this one with W-10522627, and feel free to add more details in the ticket with what you found.

jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
gbockus-sf: 👍

packages/salesforcedx-vscode-core/src/index.ts
gbockus-sf: 👍

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
gbockus-sf: we're still discussing as a team. Can follow up later after we decide on a coding convention.

packages/salesforcedx-vscode-core/src/orgBrowser/browser.ts
jag-sfdc: Diff and Deploy/Retrieve would be the areas I would be worried about. If you have tested those, we should be good I hope.

jag-sfdc (approved):

gbockus-sf (commented):
QE complete. Changes look good.  QE doc https://salesforce.quip.com/ADkIAoCajhbe

PR #3789, opened 2022-02-02
fix(soql): remove excess telemetry

jgellin-sf:
### What does this PR do?
This PR removes a telemetry call that clutters telemetry unnecessarily.

@[W-10523357](https://gus.lightning.force.com/a07EE00000kLzFqYAK)@

### Functionality Before
When SOQL tooling fails to connect to an org for any reason, telemetry logs the error.

### Functionality After
When SOQL tooling fails to connect to an org for any reason, there is no telemetry.


jonnyhork (approved):
Thanks for taking car of this Jon.

jag-sfdc (approved):

PR #3805, opened 2022-02-03
Update git2gus config default build value

jag-sfdc:
### What does this PR do?
Update git2gus config to 54.0 offcore build


mohanraj-r (approved):
Do we need a make a note to do this every release?

PR #3812, opened 2022-02-09
Jb/add trace flag auto delete

jeffb-sfdc:
### What does this PR do?
Implements the auto removal of temporary trace flags.

@W-10288383@

### Functionality Before
With the new updates to the "SFDX: Launch Apex Replay Debugger with Current File” command, a trace flag was created behind the scenes, but wasn't cleaned up after the debugging session finished.

### Functionality After
Now, after finished debugging, if the trace flag was created by the command, it is deleted from the org (existing trace flags are left untouched)


packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -207,6 +208,8 @@ export class FunctionService {
         if (functionExecution) {
           functionExecution.debugSession = undefined;
         }
+
+        TraceFlagsRemover.getInstance(undefined).removeNewTraceFlags();
```
jeffb-sfdc, line 212: If this seems hokey, let me know.  I don't want to create another "getInstance" function, and TypeScript doesn't support function overloading.  (yes, you can sometimes simulate function overloading in TS but it's hacky and I don't care for it).

I think this is safe b/c:
1) I know that by the time we get here, `getInstance(this.connection)1 (in `createTraceFlag()`) has already been called.
2) I added guards in getInstance() to guard against this.

packages/salesforcedx-utils-vscode/src/helpers/traceFlagsRemover.ts
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Connection } from '@salesforce/core';
+
+/**
+ * TraceFlagsRemover is a singleton which deletes trace flags not created by the user.
```
jag-sfdc, line 10: The comment is confusing. Did you mean "deletes trace flags created by the user"?

packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
```diff
@@ -141,7 +142,13 @@ export class TraceFlags {
       'TraceFlag',
       traceFlag
     )) as DataRecordResult;
-    return result.success && result.id ? result.id : undefined;
+
+    if (result.success && result.id) {
+      TraceFlagsRemover.getInstance(this.connection).addNewTraceFlagId(result.id);
```
randi274, line 147: If the user starts a new session, would this trace flag still be removed?

packages/salesforcedx-utils-vscode/test/unit/helpers/traceFlags.test.ts
```diff
@@ -155,6 +155,84 @@ describe('Trace Flags', () => {
...
+      .onFirstCall()
+      .resolves({ success: true, id: 'aBcDeF' })
+      .onSecondCall()
+      .resolves({ success: true, id: '01020304' });
+
+    await traceFlags.ensureTraceFlags();
+    expect(createTraceFlagSpy.callCount).to.equal(1);
+  });
+
+  it('should verify that createTraceFlag() is not called when a trace flag exist', async () => {
```
randi274, line 191: Can we also add a test that we don't delete existing flags, and that we are able to delete the flags that we create?

jeffb-sfdc: I'm working on tests for the other files I touched.  After I finished I'll circle back to this.  I think the ask of "add a test that we don't delete existing flags" might be covered in another file (either traceFlagsRemover, or this one, traceFlags)

packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
jeffb-sfdc: By new session, do you if they quit VS Code and re-run?  `connection` is coming from `workspaceContext.getConnection()`.  I looked at `connection.accessToken` and it was the same value between runs, even after I quit VSC, ran VSC again, and then ran the command.

packages/salesforcedx-utils-vscode/src/helpers/traceFlagsRemover.ts
jeffb-sfdc: done

packages/salesforcedx-utils-vscode/test/unit/helpers/traceFlags.test.ts
jeffb-sfdc: I think these two tests cover both of these scenarios.  Let's sync up & discuss.

packages/salesforcedx-utils-vscode/test/unit/helpers/traceFlagsRemover.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const instance = TraceFlagsRemover.getInstance(mockConnection);
+    instance.addNewTraceFlagId('123');
+
+    await instance.removeNewTraceFlags();
+
+    expect(toolingDeleteStub.called).to.equal(true);
+    expect(1).to.equal(1);
+  });
+
+  it('should delete multiple trace flags', async () => {
```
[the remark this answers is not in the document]
jeffb-sfdc: done (added the, "should validate that connection.tooling.delete is not called when a trace flag already exists" test")

randi274 (approved):

jag-sfdc (approved):

Issue #3801, opened 2022-02-03
Generating a manifest file in Org Browser results in an error


jeffb-sfdc:
### Summary

Generating a manifest file in Org Browser results in an error

### Steps To Reproduce:

1. Switch to the Org Browser
2. Navigate to Apex Classes
3. Selected a few classes
4. Right-click, and select “SFDX: Generate Manifest File”

### Expected result

If the new Generate Manifest File command supports the Org Browser, I'd expect it to run

If it's not support, I'd expect the command to not be available.

### Actual result

“Error running command sfdx.create.manifest: %s: File or folder not found. This is likely caused by the extension that contributes sfdx.create.manifest.”

jeffb-sfdc:
Git2Bot didn't automatically run.  The ticket I created to track this is https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000kuvcUYAQ/view

gbockus-sf:
Fixed

PR #3818, opened 2022-02-10
fix: Generating manifest file in Org Browser results in an error

RitamAgrawal:
### What does this PR do?
This PR removes the command `Generate Manifest File`, which was appearing on right-clicking the components listed under metadata types in the Org Browser e.g. `AccountController` component listed under `Apex Classes` metadata type.

#3801 , @W-10534623@

### Functionality Before
Command `Generate manifest file` was appearing on right clicking the components listed within metadata types and the command was not executable.
Screen Shot 2022-02-10 at 11 26 57 AM

### Functionality After
The command 'Generate Manifest file' doesn't appear on right clicking at components listed within metadata types


jag-sfdc (approved):

packages/salesforcedx-vscode-core/package.json
```diff
@@ -163,10 +163,6 @@
           "command": "sfdx.force.conflict.open",
           "when": "view == conflicts && viewItem == conflict-actions",
           "group": "inline"
-        },
-        {
-          "command": "sfdx.create.manifest",
-          "when": "view == metadata && viewItem == component"
```
gbockus-sf, line 169: A for my education question...why do we sometimes have the `"when": "false"` and other times don't?

gbockus-sf (approved):

packages/salesforcedx-vscode-core/package.json
gbockus-sf: @RitamAgrawal?

PR #3820, opened 2022-02-10
fix: issue with  types for jsforce

gbockus-sf:
### What does this PR do?
pins the version of @types/jsforce to one that works for the current version of the code.  
 
### Functionality Before
build was failing 

### Functionality After
build succeeds without code changes.

### DIFF
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -34,6 +34,7 @@
   "dependencies": {
     "@salesforce/apex-tmlanguage": "1.6.0",
     "@salesforce/core": "2.28.0",
+    "@types/jsforce": "1.9.40",
     "@salesforce/soql-builder-ui": "0.2.0",
     "@salesforce/soql-data-view": "0.1.0",
     "@salesforce/soql-language-server": "0.7.1",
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {
     "@salesforce/apex-tmlanguage": "1.6.0",
     "@salesforce/core": "2.28.0",
+    "@types/jsforce": "1.9.40",
```
jeffb-sfdc, line 37: @types/jsforce is already included in this file (line 53).  Let's include it only once.  Also, should this a dependency, or a dev dependdency?

gbockus-sf: good catch.  Yeah it _should_ be a devDep, but @salesforce/core has it as a dependency so in order for us to avoid getting the version they have (which breaks us) we have to have it be a dependency as well.

jeffb-sfdc (approved):

mohanraj-r:
Thanks @gbockus-sf I have tested out this in https://github.com/forcedotcom/salesforcedx-vscode/pull/3821
We can close this one

PR #3838, opened 2022-02-14
containerless function when running locally

gbockus-sf:
### What does this PR do?
Add support for running a function directly on the developers system.
Alters the Start Function command to run locally instead of in a container. 
Refactor the start function code to support container & containerless 
I'm not sure how to see unit test coverage (yet), but I attempted to get full code coverage for the classes I touched in this PR. 

@W-10535876@
Also deals with W-10535850

### Functionality Before
Previously we defaulted to running functions in the containers running locally on the developer system via the 'SFDX: Start Function' command. 

### Functionality After
There are now two commands available: 
'SFDX: Start Function' Starts the function locally on the developers system outside of the container.
'SFDX: Start Container Function' Starts the function in the container locally on the developers system.


packages/salesforcedx-vscode-core/package.json
```diff
@@ -61,6 +62,7 @@
     "mocha-multi-reporters": "^1.1.7",
     "mock-spawn": "0.2.6",
     "nyc": "^13",
+    "proxyquire": "2.1.3",
```
gbockus-sf, line 65: add proxyquire as a dev dep to the core project to enable unit testing outside the integration test flow.  We are using this in a few packages already, but didn't have any references in core.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -172,6 +174,10 @@
           "command": "sfdx.force.function.start",
           "when": "sfdx:project_opened && resource =~ /.*/functions/.*/[^/]+(/[^/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
         },
+        {
+          "command": "sfdx.force.function.container.start",
+          "when": "sfdx:project_opened && resource =~ /.*/functions/.*/[^/]+(/[^/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
+        },
```
gbockus-sf, line 180: Note the previous `sfdx.force.function.start` was left unaltered here, but is updated to run a new command function in index.ts.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -70,6 +70,7 @@
   "force_analytics_template_create_text": "SFDX: Create Sample Analytics Template",
   "force_function_create": "SFDX: Create Function",
   "force_function_start": "SFDX: Start Function",
+  "force_function_start_container": "SFDX: Start Container Function",
```
gbockus-sf, line 73: New command that will run the old container based start function code. @sbudhirajadoc

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionContainerStartCommand.ts
```diff
@@ -0,0 +1,41 @@
...
+} from './forceFunctionStart';
+
+export const CONTAINER_START_TEXT_KEY = 'force_function_container_start_text';
+export const FUNCTION_CONTAINER_LOG_NAME = 'force_function_container_start';
+
+/**
+ * Executes sfdx run:function:start --verbose
+ * @param sourceUri
+ */
+export async function forceFunctionContainerStartCommand(sourceUri?: Uri) {
```
gbockus-sf, line 26: Did some refactoring here to try and separate some concerns.  Previously we have one file that exported both the command and executor. I moved those to different files to enable easier unit testing of each piece of the puzzle.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,139 @@
+import * as vscode from 'vscode';
+import { channelService } from '../../../channels';
+import { notificationService } from '../../../notifications';
+import { FunctionService } from '../functionService';
+import { FUNCTION_RUNTIME_DETECTION_PATTERN } from '../types/constants';
+
```
gbockus-sf, line 6: I'll put all these import together.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionContainerlessStartCommand.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const CONTAINER_START_TEXT_KEY =
+  'force_function_containerless_start_text';
+export const FUNCTION_CONTAINER_LOG_NAME = 'force_function_containerless_start';
+
+/**
+ * Executes sfdx run:function:start:local --verbose
+ * @param sourceUri
+ */
+export const forceFunctionContainerlessStartCommand = async (
```
gbockus-sf, line 20: I'll make this consistent across both function start commands.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart.ts
```diff
@@ -1,242 +0,0 @@
...
-} = {
-  force_function_start_docker_plugin_not_installed_or_started: {
-    cliMessage: 'Cannot connect to the Docker daemon',
-    errorNotificationMessage: nls.localize(
-      'force_function_start_warning_docker_not_installed_or_not_started'
-    )
-  }
-};
-
-export class ForceFunctionStartExecutor extends LibraryCommandletExecutor<
```
gbockus-sf, line 59: Refactored this into a base abstract class that is extended by the container and containerless executors.  There was a large chunk of shared code which drove my decision here.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -0,0 +1,53 @@
...
+      debugPort: FUNCTION_DEFAULT_DEBUG_PORT
+    });
+
+    localRun
+      .exec()
+      .then(msg => {
+        console.log('Gordon resolved: ' + msg);
+      })
+      .catch(err => {
+        console.log('Gordon Error: ' + err.message);
```
gbockus-sf, line 50: Will address consoles. Wasn't sure what exactly should happen in these cases.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -0,0 +1,53 @@
...
+  ): Promise<void> {
+    const functionLanguage = FunctionService.instance.getFunctionType();
+
+    const localRun = new LocalRun(functionLanguage, {
+      path: functionDirPath,
+      port: FUNCTION_DEFAULT_PORT,
+      debugPort: FUNCTION_DEFAULT_DEBUG_PORT
+    });
+
+    localRun
```
gbockus-sf, line 44: LocalRun has a return type of `Promise<void>`.  So there is not handle to the started function process.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/index.ts
```diff
@@ -0,0 +1,4 @@
+
+export {ForceFunctionContainerlessStartExecutor} from './ForceFunctionContainerlessStartExecutor';
+export {ForceFunctionContainerStartExecutor} from './ForceFunctionContainerStartExecutor';
+export { validateStartFunctionsUri } from './validateStartFunctionsUri';
```
gbockus-sf, line 4: hum...I'm not sure why prettier didn't catch the space inconsistency in these exports.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -43,6 +52,8 @@ export interface FunctionExecution extends Terminable {
   debugSession?: vscode.DebugSession;
 }
 
+export const FUNCTION_TYPE_ERROR = 'Unable to determine type of exectucting function.';
```
gbockus-sf, line 55: @sbudhirajadoc New error message on invalid function type.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -352,7 +353,12 @@ function registerCommands(
 
   const forceFunctionStartCmd = vscode.commands.registerCommand(
     'sfdx.force.function.start',
-    forceFunctionStart
+    forceFunctionContainerlessStartCommand
+  );
```
gbockus-sf, line 357: this is where the containerless start function is replacing the previous behavior.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -143,7 +143,8 @@ export const messages = {
   force_lightning_event_create_text: 'SFDX: Create Aura Event',
   force_lightning_interface_create_text: 'SFDX: Create Aura Interface',
   force_function_create_text: 'SFDX: Create Function',
-  force_function_start_text: 'SFDX: Start Function',
+  force_function_container_start_text: 'SFDX: Start Container Function',
+  force_function_containerless_start_text: 'SFDX: Start Local Function',
```
gbockus-sf, line 147: DO I need to do anything with the package.nls.ja.json if new values are added here?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStartCommand.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    let validateStartFunctionsUriStub: SinonStub;
+    let commandletRunStub: SinonStub;
+    let forceFunctionContainerStartCommand: any;
+    let FUNCTION_CONTAINER_LOG_NAME: string;
+    let CONTAINER_START_TEXT_KEY: string;
+    let filePathGathererConstructorStub: SinonStub;
+    let sfdxWorkspaceCheckerConstructorStub: SinonStub;
+    let forceFunctionContainerStartExecutorConstructorStub: SinonStub;
+
+    beforeEach(() => {
```
gbockus-sf, line 29: So I probably spent entirely too much time trying to figure out how to deal with the `vscode` module when developing unit tests locally.  The issue is it doesn't exist locally so you can't mock it (or even run a test that imports code where it gets referenced) when running the tests outside of vscode (the ide). 

I found a work around that I don't love, but found it to be the lesser of the two evils of not writing unit tests.
Using proxyquire I can ensure we are stubbing all dependancies.  The downside is I'm losing the type checking when calling the function under test (note this impact the typescript compile for the test only).

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStartCommand.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('Verify container start command.', () => {
+    const fakeUri: any = 'what';
+    let validateStartFunctionsUriStub: SinonStub;
+    let commandletRunStub: SinonStub;
+    let forceFunctionContainerStartCommand: any;
```
gbockus-sf, line 22: this is our type checking sadness I mentioned below.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStartCommand.test.ts
```diff
@@ -0,0 +1,102 @@
...
+            FilePathGatherer,
+            SfdxCommandlet,
+            SfdxWorkspaceChecker
+          },
+          './forceFunctionStart': {
+            ForceFunctionContainerStartExecutor,
+            validateStartFunctionsUri: validateStartFunctionsUriStub
+          }
+        }
+      ));
```
gbockus-sf, line 79: So here is where I pull in the exported members from the forceFunctionContainerStartCommand module.  This is done here in order to provide the stubbed dependancies to the class under test.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart.test.ts
```diff
@@ -5,15 +5,15 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
```
gbockus-sf, line 5: I'll rename this to indicate it's an integration test for the container start flow.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/index.ts
jag-sfdc: Don't think we have a rule for this. But for consistency, I see we use space in both exports and imports.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jag-sfdc: Not required. We rely on an external contributor (and until then, it defaults to English).

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/index.ts
gbockus-sf: hum...yeah I double checked and prettier should take care of this.  I'm not sure how it slipped by the first time.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -35,12 +35,23 @@ const forceFunctionStartErrorInfo: {
...
+  CONTAINER: 'container',
+  ERROR: 'error',
+  LOG: 'log',
+  PACK: 'pack'
+};
+
 export class ForceFunctionContainerStartExecutor extends ForceFunctionStartExecutor {
+  // The function binary. It must be used as a single instance. Multiple class to the getFunctionBinary
+  // function will not provide the same instance of the binary.
+  private functionsBinary: Benny | undefined;
```
gbockus-sf, line 48: figured out in testing that this has to be the same instance of the binary across the various methods of the listeners do not work as expected.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -51,10 +57,18 @@ export class ForceFunctionContainerlessStartExecutor extends ForceFunctionStartE
...
       })
       .catch(err => {
-        console.log('Gordon Error: ' + err.message);
+        const errorNotificationMessage = nls.localize(
+          this.UNEXPECTED_ERROR_KEY
+        );
+        telemetryService.sendException(this.UNEXPECTED_ERROR_KEY, err.message);
+        notificationService.showErrorMessage(errorNotificationMessage);
+        channelService.appendLine(errorNotificationMessage);
+        channelService.showChannelOutput();
```
gbockus-sf, line 71: Added correct logging and telemetry to the local run executor.  Please let me know if I missed anything.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.test.ts
```diff
@@ -0,0 +1,548 @@
...
+    };
+
+    ({ ContinueResponse } = proxyquireStrict(
+      '@salesforce/salesforcedx-utils-vscode/out/src/types',
+      {
+        vscode: vscodeStub
+      }
+    ));
+
+    ({ ForceFunctionStartExecutor } = proxyquireStrict(
```
gbockus-sf, line 144: This should probably get pulled into a test util eventually.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.test.ts
```diff
@@ -0,0 +1,548 @@
...
+      await executor.setupFunctionListeners('/does/not/matter', {
+        dispose: stub()
+      });
+      await executor.startFunction(functionName, functionDirPath);
+      assert.calledOnce(appendLineStub);
+      assert.calledWith(fakeBinary.run, functionName, {});
+    });
+  });
+
+  describe('ForceFunctionStartExecutor run.', () => {
```
gbockus-sf, line 466: This exercises the method in the base class that is used by both the container and containerless executors.  I could move it to a standalone unit test and implement my own class extending it.  Lemme know your thoughts.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.test.ts
```diff
@@ -0,0 +1,285 @@
...
+ */
+
+import { vscodeStub } from '@salesforce/salesforcedx-utils-vscode/out/test/unit/commands/mocks';
+import { expect } from 'chai';
+import * as proxyquire from 'proxyquire';
+import { assert, createSandbox, SinonSandbox, SinonStub, stub } from 'sinon';
+
+const proxyquireStrict = proxyquire.noCallThru();
+
+describe('ForceFunctionContainerlessStartExecutor unit tests', () => {
```
gbockus-sf, line 15: similar to the container executor test, but only covering the containersless method implementations.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStart.test.ts
```diff
@@ -5,15 +5,15 @@
...
+import { getFunctionsBinary } from '@heroku/functions-core';
 import * as library from '@heroku/functions-core';
 import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode/out/src';
 import { EventEmitter } from 'events';
 import * as path from 'path';
 import { assert, createSandbox, SinonSandbox, SinonStub } from 'sinon';
 import { Uri, window } from 'vscode';
 import { channelService } from '../../../../src/channels';
-import { forceFunctionStart } from '../../../../src/commands/functions/forceFunctionStart';
+import { forceFunctionContainerStartCommand } from '../../../../src/commands/functions/forceFunctionContainerStartCommand';
```
jeffb-sfdc, line 16: This is testing forceFunctionContainerStartCommand, so this file should be named forceFunctionContainerStartCommand.test.ts (not forceFunctionContainerStart.ts)

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStartCommand.test.ts
```diff
@@ -0,0 +1,109 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { vscodeStub } from '@salesforce/salesforcedx-utils-vscode/out/test/unit/commands/mocks';
+import * as proxyquire from 'proxyquire';
+import { assert, createSandbox, SinonSandbox, SinonStub, stub } from 'sinon';
+
+const proxyquireStrict = proxyquire.noCallThru();
+
+describe.only('Force Function Start Container Command Unit Tests.', () => {
```
jeffb-sfdc, line 14: Maybe I'm not seeing something here, but both forceFunctionContainerStartCommand.test.ts and forceFunctionContainerStart.test.ts test forceFunctionContainerStartCommand.  Can these be consolidated into just forceFunctionContainerStartCommand.test.ts?  We like to have a 1:1 mapping between test and implementation files.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -170,6 +172,10 @@
           "command": "sfdx.force.function.start",
           "when": "sfdx:project_opened && resource =~ /.*/functions/.*/[^/]+(/[^/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
         },
+        {
```
jeffb-sfdc, line 175: Since sfdx.force.function.start is for forceFunctionContainerlessStartCommand, would it make sense to change the command to "sfdx.force.function.containerless.start"?

jeffb-sfdc: Not sure & just want to raise the pint.  On one have, the name change would be more intuitive to a reader, but on the other hand, "sfdx.force.function.containerless.start" is _**really**_ close to "sfdx.force.function.container.start and might lead to bugs.  Thoughts?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionContainerStartCommand.ts
```diff
@@ -0,0 +1,41 @@
...
+export const CONTAINER_START_TEXT_KEY = 'force_function_container_start_text';
+export const FUNCTION_CONTAINER_LOG_NAME = 'force_function_container_start';
+
+/**
+ * Executes sfdx run:function:start --verbose
+ * @param sourceUri
+ */
+export const forceFunctionContainerStartCommand = async (sourceUri?: Uri) => {
+  const validSourceUri = validateStartFunctionsUri(sourceUri);
+  if (validSourceUri === undefined) {
```
jeffb-sfdc, line 28: I recommend doing `if (!validSourceUri)` over `if (validSourceUri === undefined)`.  This way, if it's null or undefined it will return.  Yes, the TS says it returns undefined (and not undefined | null), but I've seen cases where undefined was expected, but one got  null instead.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionContainerlessStartCommand.ts
```diff
@@ -0,0 +1,44 @@
...
+  sourceUri?: Uri
+) => {
+  const validSourceUri = validateStartFunctionsUri(sourceUri);
+  if (validSourceUri === undefined) {
```
jeffb-sfdc, line 31: Same comment above, about using ! over ===.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
+/*
```
jeffb-sfdc, line 1: nit:  I think the convention within our projects is to name files using camelCase, and not PascalCase.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+      // evergreen:benny:message {"type":"log","timestamp":"2021-05-10T10:00:27.953248-05:00","level":"info","fields":{"debugPort":"9229","localImageName":"jvm-fn-init","network":"","port":"8080"}} +21ms
+      if (msg.fields && msg.fields.localImageName) {
+        channelService.appendLine(`'Running on port' :${msg.fields.port}`);
+        channelService.appendLine(
+          `'Debugger running on port' :${msg.fields.debugPort}`
+        );
+      }
+    });
+    // Allows for showing custom notifications
+    // and sending custom telemtry data for predefined errors
```
jeffb-sfdc, line 125: typo: telemetry

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,139 @@
...
+import { telemetryService } from '../../../telemetry';
+import { ForceFunctionStartExecutor as ForceFunctionStartExecutor } from './ForceFunctionStartExecutor';
+
+/**
+ * Error types when running SFDX: Start Function
+ * This is also used as the telemetry log name.
+ */
+type ForceFunctionStartErrorType = 'force_function_start_docker_plugin_not_installed_or_started';
+
+const forceFunctionStartErrorInfo: {
```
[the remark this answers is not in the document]
jeffb-sfdc: Reading through the code, I see what forceFunctionStartErrorInfo is doing, and looks like this code & its approach originally came from forceFunctionStart.ts.  But reading through the code, my first thought was... "what?"

This (the original code that was carried forward) seems (IMHO) to be a little over engineered.  This might be a good opportunity to clean it up and make it more simple. We could get rid of ForceFunctionStartErrorType and forceFunctionStartErrorInfo, and just change it to:
```
const forceFunctionStartErrorInfo = {
  force_function_start_docker_plugin_not_installed_or_started: {
    cliMessage: 'Cannot connect to the Docker daemon',
    errorNotificationMessage: nls.localize(
      'force_function_start_warning_docker_not_installed_or_not_started'
    )
  }
}
```

...and then further down, not have to do ` (Object.keys(forceFunctionStartErrorInfo) as ForceFunctionStartErrorType[]).forEach(...`

Thoughts?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+} = {
+  force_function_start_docker_plugin_not_installed_or_started: {
+    cliMessage: 'Cannot connect to the Docker daemon',
+    errorNotificationMessage: nls.localize(
+      'force_function_start_warning_docker_not_installed_or_not_started'
+    )
+  }
+};
+
+export const BINARY_EVENT_ENUM = {
```
jeffb-sfdc, line 38: TypeScript has support for enums.  Maybe change to:
```
enum BinaryEvent {
  Container,
  Log,
  Pack
}```

File: packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts [hunk 32]
```diff
@@ -0,0 +1,164 @@
...
+    const writeMsg = (msg: { text: string; timestamp: string }) => {
+      const outputMsg = msg.text;
+
+      if (outputMsg) {
+        channelService.appendLine(outputMsg);
+
+        const matches = String(outputMsg).match(
+          FUNCTION_RUNTIME_DETECTION_PATTERN
+        );
+        if (matches && matches.length > 1) {
```
jeffb-sfdc, line 65: Just curious, does this check need to also check for the presence of matches[1]?  So...
```
if (matches && matches.length > 1 && matches[1]) { ... }
```

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+      }
+    });
+    // Allows for showing custom notifications
+    // and sending custom telemtry data for predefined errors
+    this.functionsBinary.on(BINARY_EVENT_ENUM.ERROR, handleError);
+  }
+  public async cancelFunction(
+    registeredStartedFunctionDisposable: vscode.Disposable
+  ): Promise<void> {
+    if (this.functionsBinary !== undefined) {
```
jeffb-sfdc, line 131: Same comment about using `!` vs `!== undefined`

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    await this.functionsBinary.build(functionName, {
+      verbose: true,
+      path: functionDirPath
+    });
+  }
+
+  public async startFunction(
+    functionName: string,
+    functionDirPath: string
```
jeffb-sfdc, line 156: `functionDirPath` isn't used.  Can it be removed?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+  ): Promise<void> {
+    channelService.appendLine(`Starting ${functionName} in container`);
+    if (!this.functionsBinary) {
+      throw new Error('Unable to start function with no binary.');
+    }
+    this.functionsBinary.run(functionName, {}).catch(err => console.log(err));
```
jeffb-sfdc, line 162: I'm not a fan of putting multiple statements on a single line.  (at the very least) it makes placing break points difficult, and (me personally) I prefer:
```
    this.functionsBinary.run(functionName, {})
      .catch(err => console.log(err));
```

I'd like others on the team to weigh in.  Maybe it's just me, but I prefer the calls to be on seperate lines.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -0,0 +1,74 @@
...
+import {
+  FUNCTION_DEFAULT_DEBUG_PORT,
+  FUNCTION_DEFAULT_PORT
+} from '../types/constants';
+import { ForceFunctionStartExecutor } from './ForceFunctionStartExecutor';
+
+export class ForceFunctionContainerlessStartExecutor extends ForceFunctionStartExecutor {
+  public async setupFunctionListeners(
+    functionDirPath: string,
+    functionDisposable: Disposable
```
jeffb-sfdc, line 24: `functionDirPath` and `functionDisposable` aren't used.  Can they be removed?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -0,0 +1,74 @@
...
+  public async cancelFunction(
+    registeredStartedFunctionDisposable: Disposable
+  ): Promise<void> {
+    // TODO: how to stop the localRun
+    registeredStartedFunctionDisposable.dispose();
+  }
+
+  public async buildFunction(
+    functionName: string,
+    functionDirPath: string
```
jeffb-sfdc, line 38: `functionName` and `functionDirPath` aren't used.  Can they be removed?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
```diff
@@ -0,0 +1,133 @@
...
+  protected UNEXPECTED_ERROR_KEY = 'force_function_start_unexpected_error';
+
+  constructor(startMessageKey: string, logName: string) {
+    super(nls.localize(startMessageKey), logName, OUTPUT_CHANNEL);
+    this.cancellable = true;
+  }
+
+  public async run(
+    response: ContinueResponse<string>,
+    progress?: vscode.Progress<{
```
jeffb-sfdc, line 36: Is progress used, or can it be removed?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
```diff
@@ -0,0 +1,133 @@
...
+    channelService.appendLine('Parsing project.toml');
+    const descriptor = await getProjectDescriptor(
+      path.join(functionDirPath, 'project.toml')
+    );
+    const functionName = descriptor.com.salesforce.id;
+
+    this.buildFunction(functionName, functionDirPath);
+
+    channelService.appendLine(`Starting ${functionName}`);
+    await this.startFunction(functionName, functionDirPath);
```
jeffb-sfdc, line 111: VSC is reporting a warning here that `await` is't needed.  Can it be removed?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -12,6 +12,15 @@ import { Disposable } from 'vscode';
 import { workspaceContext } from '../../context';
 import { getRootWorkspace, getRootWorkspacePath } from '../../util';
 
+/**
+ * An enum for the different types of functions.
+ */
+export enum FUNCTION_TYPE {
```
jeffb-sfdc, line 18: I think the convention for enums in TS is PascalCase, so:
```
export enum FunctionType {
  JavaScript,
  TypeScript,
  Java
}
```

Here's what the TypeScript docs say: https://www.typescriptlang.org/docs/handbook/enums.html

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jeffb-sfdc: typo with "exectucting"

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -45,6 +54,8 @@ export interface FunctionExecution extends Terminable {
   debugSession?: vscode.DebugSession;
 }
 
+export const FUNCTION_TYPE_ERROR = 'Unable to determine type of exectucting function.';
+
```
jeffb-sfdc, line 58: FunctionService looks to be a singleton, but it's missing a private constructor.  Since you are making changes to this file, could you add a private constructor?

packages/salesforcedx-vscode-core/package.json
gbockus-sf: I'm fine with making that change.  It would mirror the rest of the code and makes sense 👍

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionContainerStartCommand.ts
gbockus-sf: def please call that out if you run into it again.  If we can't trust the typing of typescript then there's def a problem (probably :any fun). 
So the condition I was a little concerned about was '' (emptystring). 
The way the validate function works it should default to the activeTextEditor Uri, but I wasn't 100% clear at the time I touched this.  I'm fine with making the change though b/c I prefer that syntax as well.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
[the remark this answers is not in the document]
gbockus-sf: I felt that "what?" I've had to review this a couple of times to understand it and even then it def seem over engineered. I'll simplify. 👍

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
gbockus-sf: 👍 cleaned it up.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
gbockus-sf: turns out we have a tslint rule for including parans for a single param function. but {} added.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
gbockus-sf: good call. in the abstract method def the param is optional I can just exclude it (needed for containerless implementation)

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
```diff
@@ -0,0 +1,133 @@
...
+        defaultUsernameorAlias => {
+          if (!defaultUsernameorAlias) {
+            const message = nls.localize('force_function_start_no_org_auth');
+            channelService.appendLine(message);
+            channelService.showChannelOutput();
+            notificationService.showInformationMessage(message);
+          }
+        }
+      );
+    } catch (error) {
```
gbockus-sf, line 69: I don't really understand this line.  empty catches only lead to sadness 🤷

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionContainerStartCommand.test.ts
gbockus-sf, line 14: missed an only 🤦 . fixed.

gbockus-sf: So this file is the unit tests and the `forceFunctionContainerStart.test.ts` is the integration tests.  I updated the descriptions in the describes for each file, but didn't want to move things around just yet until we dig deeper on what we're going to do with the unit tests for this package.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
jeffb-sfdc: Truth be told, I'm a terrible speller.  I use a VS Code extension named "Code Spell Checker" and it does a good job of finding spelling errors.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
gbockus-sf: Discussed with Jeff offline.  We're going to discuss with the larger team the convention moving forward.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -39,12 +40,16 @@ export class ForceFunctionContainerlessStartExecutor extends ForceFunctionStartE
...
     );
+
     const localRun = new LocalRun(functionLanguage, {
       path: functionDirPath,
       port: FUNCTION_DEFAULT_PORT,
       debugPort: FUNCTION_DEFAULT_DEBUG_PORT
     });
 
+    const debugType = functionLanguage === functionType.JAVA ? 'java' : 'node';
+    FunctionService.instance.updateFunction(functionDirPath, debugType);
```
gbockus-sf, line 51: So in the container function flow this happens in an odd place so I missed it for containerless function.  This updates the debug type after the function is started (which is required to run the service method for getting the type.

jeffb-sfdc (approved):

PR #3839, opened 2022-02-15
feat: add legacy source pull, push, status cmds for scratch orgs

mohanraj-r:
### What does this PR do?
Create parallel commands for scratch org pull/push/status commands to use new source:legacy versions of the commands as per https://github.com/forcedotcom/cli/issues/1258

### What does this PR doesn't do?
* Use [source-tracking](https://github.com/forcedotcom/source-tracking) lib to determine the version of source tracking in use and issue the `beta` or `legacy` version of the command
* Create a generic way to handle different versions of commands (e.g. `beta`, `legacy`)

#### TODO

- [ ] ~~🛑  **BEFORE MERGE: WAIT for https://github.com/salesforcecli/toolbelt/pull/213 to be released**~~
   - [ ] ~~`legacy` subtopic available in latest stable release of CLI.  e.g. `sfdx force:source:legacy:status`~~
- [x] Add source legacy command
	- [x] 	Account for [Simplified flags and ignored on force:source:status](https://github.com/forcedotcom/cli/issues/1258)
- [x] Fix integration tests
- [x] Get docs updated/reviewed 
	- [x] Add notification to users about upcoming changes and requiring latest CLI (e.g. using pinned issue) 
	- [ ]  Highlight this in release notes

@W-10669583@

### Functionality Before
* Only one version of source pull/push/status commands are supported 
  * resulting in breakage with upcoming changes to CLI (https://github.com/forcedotcom/cli/issues/1258)  

### Functionality After
* Both the new (`beta`) and `legacy` versions of source pull/push/status commands are supported


mohanraj-r:
@jag-sfdc Can you please review this PR to see if there is anything major that is missing

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2017, salesforce.com, inc.
```
gbockus-sf, line 2: nit: 2022

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
```diff
@@ -18,6 +18,7 @@ import {
   SfdxWorkspaceChecker
 } from './util';
 export class ForceSourcePullExecutor extends SfdxCommandletExecutor<{}> {
+  public static readonly command = 'force:source:beta:pull';
```
gbockus-sf, line 21: interesting that we're calling a beta thing legacy

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.flag = flag;
+  }
+
+  public build(data: {}): Command {
+    const builder = new SfdxCommandBuilder()
+      .withDescription(
+        nls.localize('force_source_pull_legacy_default_scratch_org_text')
+      )
+      .withArg('force:source:pull')
+      .withLogName('force_source_pull_legacy_default_scratch_org');
```
gbockus-sf, line 34: it'd be cool to pull these string up into consts in the file.

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
```diff
@@ -0,0 +1,60 @@
...
+        );
+    }
+    return builder.build();
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+
+export async function forceSourcePullLegacy(this: FlagParameter<string>) {
```
gbockus-sf, line 50: Is this a pattern we use. Using a parameter named this seems like asking for trouble.  Why not call it flagParameter?

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
```diff
@@ -0,0 +1,60 @@
...
+  // tslint:disable-next-line:no-invalid-this
```
gbockus-sf, line 51: renaming the param would remove the need for this comment.

packages/salesforcedx-vscode-core/src/commands/forceSourcePushLegacy.ts
```diff
@@ -0,0 +1,66 @@
...
+    return DeployType.Push;
+  }
+}
+
+const workspaceChecker = new SfdxWorkspaceChecker();
+const parameterGatherer = new EmptyParametersGatherer();
+
+export interface FlagParameter {
+  flag: string;
+}
```
gbockus-sf, line 54: Should we put this in a file so it's easy to find?

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -20,22 +21,28 @@ import {
...
+  public command: string;
   private flag: SourceStatusFlags | undefined;
+  private description: string;
+  private logName: string;
 
-  public constructor(flag?: SourceStatusFlags) {
+  public constructor(flag?: SourceStatusFlags, cmdVersion?: CommandVersion) {
     super();
     this.flag = flag;
+    this.command = `force:source:${cmdVersion? cmdVersion : CommandVersion.Default}:status`;
```
gbockus-sf, line 36: `cmdVersion ?? CommandVersion.Default` is some TS syntax to make this a little shorter

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -20,22 +21,28 @@ import {
...
+    this.description = ['force_source_status', cmdVersion, 'text'].filter(Boolean).join('_');
```
gbockus-sf, line 37: if we're worried about it not being defined on line 36 should we not also have the same concern here?  maybe assign it to a variable and use the variable in both places?

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -20,22 +21,28 @@ import {
...
+    this.logName = ['force_source_status', cmdVersion].filter(Boolean).join('_');
```
gbockus-sf, line 38: whoa...what's this do?

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -28,8 +28,23 @@ import { taskViewService } from '../../statuses';
 import { telemetryService } from '../../telemetry';
 import { getRootWorkspacePath } from '../../util';
 
+export enum CommandVersion {
```
jag-sfdc, line 31: @mohanraj-r - What happens when we clean the legacy code in a couple of months? Do we still leave the command version when there is only one version of the commands?

mohanraj-r: - As long as users are using older versions of CLI (which we have no control over) wouldn't we require legacy support?
- Regardless we can use this if this we ever have to do this legacy/beta switchover with any of the other cmds as well

jag-sfdc: As long as users use older versions of CLI and newer version of extension, things would continue to work (no change in their workflow). It is when someone upgrades to new CLI and continue to use their old scratch org, the normal commands will not work and they'll have to use the legacy commands. This would continue only until the scratch orgs expire (1 to 30 days). Is the case you talking about when a user does not upgrade at all?

Got it. I do like the overall design and this is a nice legacy/beta switchover that can be extended to other commands as well in the future.

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
```diff
@@ -47,9 +68,9 @@ export class ForceSourcePullExecutor extends SfdxCommandletExecutor<{}> {
 const workspaceChecker = new SfdxWorkspaceChecker();
 const parameterGatherer = new EmptyParametersGatherer();
 
-export async function forceSourcePull(this: FlagParameter<string>) {
```
mohanraj-r, line 50: @gbockus-sf Yes I do see the `this` param used in other places for the cmd funcs. 
Guess it has to do with how they are passed to `registerCommand`: `export function registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: any): Disposable;` where `@param thisArg The `this` context used when invoking the handler function.`

But all our tests are passing even with this change. 
@jag-sfdc Is the `this` convention required? Should we keep it?

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
mohanraj-r: Yes, thanks for the clarification Jag.

mohanraj-r (commented):
Thanks for the @gbockus-sf @jag-sfdc 
The PR is ready for review - can you please review it.

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
mohanraj-r: It adds the `_` separator only if the vars in the array are all present - yeah its a bit wonky logic :) Its removed now - moved them to static values instead of computing them dynamically and incurring the complexity/overhead

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -12,6 +12,11 @@
   "force_source_push_default_scratch_org_text": "SFDX: Push Source to Default Scratch Org",
   "force_source_push_force_default_scratch_org_text": "SFDX: Push Source to Default Scratch Org and Override Conflicts",
   "force_source_status_text": "SFDX: View All Changes (Local and in Default Scratch Org)",
+  "force_source_legacy_pull_default_scratch_org_text": "SFDX: Pull Source from Default Scratch Org (Legacy)",
```
mohanraj-r, line 15: @sbudhirajadoc Can you please review the text for the new commands

mohanraj-r: Also @AnanyaJha are there any user facing docs to add/update about the legacy/beta switchover? Or would the pinned issue you will be creating suffice?

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
jag-sfdc: I don't see a reason to keep this convention (unless I'm missing something). Agree with an earlier comment that it is confusing.

mohanraj-r: Actually when testing manually found that without the `this` the flags are not passed when the command is invoked. Guess it is a vscode specific thing. But neither our unit or integ tests were able to catch this. Going to check if existing tests can be updated to be able to catch this.

packages/salesforcedx-vscode-core/src/commands/forceSourcePush.ts
```diff
@@ -12,30 +12,52 @@ import {
...
-      .withArg('force:source:push')
+      .withArg(this.params.command)
       .withJson()
-      .withLogName('force_source_push_default_scratch_org');
+      .withLogName(this.params.logName.default);
     if (this.flag === '--forceoverwrite') {
       builder.withArg(this.flag);
       builder.withDescription(
-        nls.localize('force_source_push_force_default_scratch_org_text')
+        nls.localize(this.params.description.forceoverwrite)
```
gbockus-sf, line 60: I really like this constant pattern for commands.  Good stuff.

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -28,8 +28,22 @@ import { taskViewService } from '../../statuses';
...
+  Beta = 'beta',
+  Legacy = 'legacy',
+  Default = Beta
+}
+
 export interface FlagParameter<T> {
```
gbockus-sf, line 37: nit: I know this isn't new, but since you're here the generic seems unnecessary.  Should we just make flag a `string`.

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
gbockus-sf: So I did some digging b/c who likes mysteries.  The best doc I found around it was this: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions.  So basically it's a way to tell Typescript, `I expected this function to be running in this 'this' context when executing`.  Super weird.  In the generated js there is no parameter. It's used only for typechecking. Also the vscode command for registers is any'd to the max so I'm wondering if it even buys us anything to type this parameter.

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
```diff
@@ -48,9 +69,9 @@ const workspaceChecker = new SfdxWorkspaceChecker();
 const parameterGatherer = new EmptyParametersGatherer();
 
 export async function forceSourcePull(this: FlagParameter<string>) {
-  // tslint:disable-next-line:no-invalid-this
-  const flag = this ? this.flag : undefined;
-  const executor = new ForceSourcePullExecutor(flag);
+  const {flag, commandVersion} = this || {};
```
gbockus-sf, line 72: nit: I don't think you need the `|| {}` here.

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
mohanraj-r: dup file no longer exists

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
mohanraj-r: done

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
mohanraj-r: Yes this is a pattern being used and vscode fails to pass the flags if we don't name it `this` 🤷

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
mohanraj-r: comment removed in recent refactoring

packages/salesforcedx-vscode-core/src/commands/forceSourcePullLegacy.ts
```diff
@@ -0,0 +1,60 @@
...
+  const flag = this ? this.flag : undefined;
```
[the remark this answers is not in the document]
mohanraj-r: we are using the same handler for multiple variations of the command - some with flags and some without. So its `undefined` when the handler is invoked without any additional params/flags.

packages/salesforcedx-vscode-core/src/commands/forceSourcePushLegacy.ts
mohanraj-r: duplicate interface removed.

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
mohanraj-r: Good to know - thanks. Current refactored code doesn't use this pattern.

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
mohanraj-r: Current refactored code doesn't use this.

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
mohanraj-r: Yes this is needed when the handler is invoked without a flag param in which case it is `undefined` and then de-structure errors

packages/salesforcedx-vscode-core/package.nls.json
sbudhirajadoc: looks good to me.

packages/salesforcedx-vscode-core/src/commands/forceSourcePull.ts
gbockus-sf: We discussed in slack.  The weird part of this is that _shouldn't_ be possible but due to how the registerCommand function works we lose type safety for the parameters when this function is executed.

mohanraj-r:
@gbockus-sf Can you please take a look - I have addressed the items we discussed yesterday.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -196,12 +223,17 @@ function registerCommands(
   const forceSourceStatusLocalCmd = vscode.commands.registerCommand(
     'sfdx.force.source.status.local',
     forceSourceStatus,
-    { flag: '--local' }
+    { flag: SourceStatusFlags.Local }
```
gbockus-sf, line 226: should we pull this and line 321 into constants to enforce type safety and stand consistent?

packages/salesforcedx-vscode-core/src/commands/forceSourceStatus.ts
```diff
@@ -55,9 +77,9 @@ const parameterGatherer = new EmptyParametersGatherer();
 export async function forceSourceStatus(
   this: FlagParameter<SourceStatusFlags>
 ) {
-  // tslint:disable-next-line:no-invalid-this
-  const flag = this ? this.flag : undefined;
-  const executor = new ForceSourceStatusExecutor(flag);
```
gbockus-sf, line 60: nit: knowing what we know now would it make sense to add a type guard here for this? 
```
if (!this instanceof FlagParameter) { throw error} 
```
I don't even know if this works with `this` so marking this a nit

gbockus-sf (changes requested):
@mohanraj-r one request to pull the other two FlagParams up into variables and one type guard to consider and I think we're good to go

mohanraj-r:
@jag-sfdc Since there has been some changes/refactoring since you last reviewed can you please give this another review/QA before merging.

jag-sfdc (approved):

PR #3844, opened 2022-02-15
Remove unessential apex classes from Apex LSP jar

jag-sfdc:
### What does this PR do?
Remove unessential apex classes from Apex LSP jar

@W-10433780@ 

### Functionality Before
N/A

### Functionality After
No change in functionality

### DIFF
--- a/packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
+++ b/packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// tslint:disable:no-unused-expression
+
+import { expect } from 'chai';
+import * as path from 'path';
+import * as shell from 'shelljs';
+
+describe('Apex LSP Jar Contents', () => {
+
+  it('should not contain perm-guarded apex classes', () => {
+    shell.config.execPath = process.execPath;
+    const apexJarPath = path.join(__dirname, '..', '..', 'apex-jorje-lsp.jar');
+    const permGuardedClasses = ['AuraException.cls', 'AuraHandledException.cls', 'Blacktab.cls', 'InstallContext.cls', 'InstallHandler.cls', 'UninstallContext.cls', 
+    'UninstallHandler.cls', 'BusOp.cls', 'AppExchange.cls', 'ApplicationReadWriteMode.cls', 'CollectSimilarCasesData.cls', 'FfxPortalData.cls', 'ProductSecurity.cls',
+    'LegalOrgOps.cls', 'Org62Ops.cls', 'OrgCSOps.cls', 'GusOps.cls', 'BlackTabFramework.cls', 'SfdcOps.cls', 'Org62LEXFeedbackController.cls', 'TrailblazerIdentityAdditionalInfo.cls',
+    'TrailblazerIdentityInfo.cls', 'TrailblazerIdentityInfoWrapper.cls', 'TrailblazerIdentityArrayLengthMismatchException.cls', 'ChatterGroupSummaryPage.cls',
+    'FeedItemInputAttachmentType.cls'];
+    const stdout = shell.exec(`jar tvf ${apexJarPath}`).stdout;
+    expect(
+      permGuardedClasses.some(permGuardedClass => stdout.includes(permGuardedClass))
+    ).to.be.true;
+  });
+});
--- a/packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
+++ b/packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
@@ -24,6 +24,6 @@ describe('Apex LSP Jar Contents', () => {
     const stdout = shell.exec(`jar tvf ${apexJarPath}`).stdout;
     expect(
       permGuardedClasses.some(permGuardedClass => stdout.includes(permGuardedClass))
-    ).to.be.true;
+    ).to.be.false;
   });
 });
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// tslint:disable:no-unused-expression
```
jeffb-sfdc, line 8: Is `no-unused-expression` needed?  Otherwise it looks good.

jag-sfdc: Not required. I copied the copyright text and part of the test and this came with it. Will remove.

jag-sfdc: Actually turns out it is required. If not, lint is not happy with line 25 (complains "unused expression, expected an assignment or function call")

gbockus-sf: you can remove this and use 
```expect(
      permGuardedClasses.some(permGuardedClass => stdout.includes(permGuardedClass))
    ).to.equal(false);```
either that or we should figure out how to disable this rule for the tests.

File: packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+    const apexJarPath = path.join(__dirname, '..', '..', 'apex-jorje-lsp.jar');
+    const permGuardedClasses = ['AuraException.cls', 'AuraHandledException.cls', 'Blacktab.cls', 'InstallContext.cls', 'InstallHandler.cls', 'UninstallContext.cls',
+    'UninstallHandler.cls', 'BusOp.cls', 'AppExchange.cls', 'ApplicationReadWriteMode.cls', 'CollectSimilarCasesData.cls', 'FfxPortalData.cls', 'ProductSecurity.cls',
+    'LegalOrgOps.cls', 'Org62Ops.cls', 'OrgCSOps.cls', 'GusOps.cls', 'BlackTabFramework.cls', 'SfdcOps.cls', 'Org62LEXFeedbackController.cls', 'TrailblazerIdentityAdditionalInfo.cls',
+    'TrailblazerIdentityInfo.cls', 'TrailblazerIdentityInfoWrapper.cls', 'TrailblazerIdentityArrayLengthMismatchException.cls', 'ChatterGroupSummaryPage.cls',
+    'FeedItemInputAttachmentType.cls'];
+    const stdout = shell.exec(`jar tvf ${apexJarPath}`).stdout;
+    expect(
+      permGuardedClasses.some(permGuardedClass => stdout.includes(permGuardedClass))
+    ).to.be.false;
```
gbockus-sf, line 27: nit: one bummer about this test is if it fails we'll have no idea why. Would be not be better to assert on each class and fail with the class name if it fails?

jag-sfdc: Like now :) 
It is failing and I have no idea why. I was lazy about writing separate tests. Looks like that's what I would do.

packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// tslint:disable:no-unused-expression
+
+import { expect } from 'chai';
+import * as path from 'path';
+import * as shell from 'shelljs';
+
+describe('Apex LSP Jar Contents', () => {
+
+  it('should not contain perm-guarded apex classes', () => {
```
mohanraj-r, line 16: nit: can we use `it.each` here instead ?

gbockus-sf: ehhhh.. I dunno that we should depend on a module that is infrequently used and hasn't been updated in a few years.

packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
gbockus-sf: holy moley we have this comment in our source ~110 times.  I'll just add a comment to the eslint ticket to ensure we just disable that rule and remove the comment for the eslint migration

packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
mohanraj-r: I wasn't referring to the 3rd party mod. I was hoping our test runner/framework would have something similar to https://jestjs.io/docs/api#testeachtablename-fn-timeout

gbockus-sf: ah...I'm not aware of anything in mocha that is equalivant.  Simplest thing might be something like
```
const clsNames = ['AuraException.cls', ...];
const checkClsName = (className :string, stdout: string) => {
  expect(stdout.includes(className).to.be.false;
}
clsName.forEach((name) => {
  checkClsName(name);
});
```

gbockus-sf (approved):

packages/salesforcedx-vscode-apex/test/vscode-integration/apexLspJarContents.test.ts
jag-sfdc: I generally try to avoid loops in tests. Ideally, I would simplify this test into multiple tests so each tests one exact thing (and one class in this case). But felt it was overkill in this scenario.

mohanraj-r (approved):

PR #3845, opened 2022-02-16
chore: bump node version

randi274:
### What does this PR do?
Updates Node version from 12 to 14.

@W-10250430@

### Functionality Before
We were developing the extensions with Node version 12, soon to be extinct. 🦕

### Functionality After
Now we're using... 14!


.circleci/config.yml
```diff
@@ -18,13 +18,13 @@ parameters:
   # Configurable Versions
   cli-version:
     type: string
-    default: '7.45.1'
+    default: '7.132.0'
```
mohanraj-r, line 21: Great job for finding and updating this. We need to discuss (with team) if we `latest` here so that the `npm install -g sfdx-cli@<< pipeline.parameters.cli-version >>` command below can use the latest released version of cli.

package.json
```diff
@@ -71,9 +71,12 @@
   "repository": "forcedotcom/salesforcedx-vscode",
   "license": "BSD-3-Clause",
   "dependencies": {
-    "node": "12.22.9",
+    "node": "^14.0.0",
```
mohanraj-r, line 74: We can still use `^14.16.0` as we are using `14.16.0` in other places and that would ensure that atleast 14.16.0 is the minimum version.

package.json
```diff
@@ -71,9 +71,12 @@
...
     "npm": "^6.9.0"
   },
+  "engines": {
+    "node": ">=14.0.0"
```
mohanraj-r, line 78: same as above comment about using 14.16.0

mohanraj-r (commented):
LGTM. Some minor comments.

package.json
```diff
@@ -71,9 +71,12 @@
   "repository": "forcedotcom/salesforcedx-vscode",
   "license": "BSD-3-Clause",
   "dependencies": {
-    "node": "12.22.9",
+    "node": "^14.0.0",
     "npm": "^6.9.0"
```
mohanraj-r, line 75: Can we also update npm version to [6.14.11](https://nodejs.org/en/download/releases/) to match the node version

.circleci/config.yml
mohanraj-r: @RitamAgrawal mentions that 7.45.1 is probably the version that uses toolbelt .. maybe that has something to do with this version. But why is the cli installed only in windows and how is it used?

.npmrc
```diff
@@ -1 +1,2 @@
 save-exact=true
+engine-strict=true
```
gbockus-sf, line 2: Can we doc in the README why these settings are included and what they do?

.github/workflows/create-release-branch.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: Setup Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: '12.x'
+        node-version: '14.x'
```
gbockus-sf, line 17: should we lock this down to 14.16.X as well?

.circleci/config.yml
jag-sfdc: If 7.45.1 uses toolbelt, this is a somewhat significant change and would deserve a separate PR. We run the integration tests in windows and probably that's why we set this only in windows.

.npmrc
randi274: @gbockus-sf in the main [README](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/README.md)? Or somewhere in developing/contributing?

gbockus-sf: Probably in the contributing docs would be best. good call.  Maybe in developing.md?

.circleci/config.yml
randi274: Noted during 11sies, I'll revert this change out of this PR, and make a separate WI/PR for it.

randi274: Also smart callout about the toolbelt crossover @RitamAgrawal 🎉

.github/workflows/create-release-branch.yml
```diff
@@ -7,27 +7,27 @@ on:
...
-    - name: Install Script Dependencies
-      run: npm install shelljs && npm install -g lerna
-    - name: Create and Push the Release Branch
-      run: |
-        git config --local user.email "releaseaction@github.com"
-        git config --local user.name "Release Action"
-        node scripts/create-release-branch.js
-      env:
-        SALESFORCEDX_VSCODE_VERSION: ${{ github.event.client_payload.version }}
+      - name: Setup Node.js
```
jag-sfdc, line 14: Nit! Can we remove these extra spaces (I didn't see them in the previous version)? Everything looks good and I'll approve now.

contributing/developing.md
```diff
@@ -180,11 +180,23 @@ rules](https://palantir.github.io/tslint/usage/type-checking/) which requires a
...
+
+## Node Configuration
+
+### .npmrc
+
+The npmrc allows for project-level [configuration](https://docs.npmjs.com/cli/v8/using-npm/config) of the npm environment. The settings that we care about are `engine-strict` (this enforces the minimum node version in our packages) and `save-exact` (saving exact versions of dependencies from the package.json).
+
+### .nvmrc
+
+Our nvmrc specifies the minimimum node version required to run the project.
```
gbockus-sf, line 202: cool thanks

mohanraj-r: Can the rc files have inline comments - if yes would it make sense to have these in the files themselves for better context and future proof / sync ?

.github/workflows/create-release-branch.yml
randi274: Yeah! It was forcing me to add them to save the file; let me clean that up.

jag-sfdc (approved):

PR #3846, opened 2022-02-16
fix: enable Diff File on command palette

floralan:
### What does this PR do?
Enable "SFDX: Diff File Against Org" on Command Palette when a SFDX project opened. 

[@W-10707374](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000orpm5YAA/view)

### Functionality Before

- When there is no file opened, and a folder is selected, click SFDX:>Diff Folder Against Org from command palette. result: "Diff for this metadata type is currently not supported"
- image
- when there is a file opened, click SFDX:>Diff Folder Against Org from command palette. result: "Source Diff successfully run"
- image

### Functionality After

- User opens a file
- click "SFDX:Diff File Against Org" from command palette
- result: Source Diff successfully run


packages/salesforcedx-vscode-core/package.json
```diff
@@ -536,6 +536,10 @@
         },
         {
           "command": "sfdx.force.diff",
+          "when": "sfdx:project_opened"
```
mohanraj-r, line 539: Are there any tests that we can add/update for this?

PR #3849, opened 2022-02-17
chore: use context for common env variables

RitamAgrawal:
### What does this PR do?
Uses context to access common environment variables across different projects

 @W-10714614@

### Functionality Before
All environment variables are defined at project settings.

### Functionality After
Few common env variables are now defined in a context and remaining at project settings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -540,6 +540,7 @@ workflows:
     when: << pipeline.parameters.pre-publish >>
     jobs:
       - slack/approval-notification:
+          context: IEAT_keys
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -540,14 +540,15 @@ workflows:
     when: << pipeline.parameters.pre-publish >>
     jobs:
       - slack/approval-notification:
+          context: IEAT_keys
```
mohanraj-r, line 543: Is slack auth now part of the `IEAT_keys` ?

mohanraj-r (approved):

PR #3866, opened 2022-02-28
chore: update SDR, @salesforce/core, and apex-node deps

gbockus-sf:
### What does this PR do?
Catch up the repo for SDR, apex-node, and @salesforce/core to have a consistent versions to enable updating SDR for fixing readFileSync issue. 
One thing to call out. I had to add the typescript configuration for skipping the lib checks in the tsconfig.json 
    "skipLibCheck": true,
This was fallout from SDR using a newer version of typescript then this repo and the two not being compatible.  I updated work item W-9865175 to include removing this configuration when we upgrade TS. 

First merge for @W-10120433@
Will also require a followup to pick up the latest from SDR. 

### Functionality Before
N/A

### Functionality After
N/A


packages/salesforcedx-apex-debugger/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "rootDir": ".",
     "outDir": "out",
     "preserveConstEnums": true,
+    "skipLibCheck": true,
```
gbockus-sf, line 13: This is required b/c the two version of TS don't play nice between this repo and SDR 
```
salesforcedx-vscode-core: node_modules/@salesforce/core/lib/sfdxError.d.ts(152,9): error TS2380: 'get' and 'set' accessor must have the same type.
salesforcedx-vscode-core: node_modules/@salesforce/core/lib/sfdxError.d.ts(153,9): error TS2380: 'get' and 'set' accessor must have the same type.```
are the kinds of errors found in several module.

File: packages/salesforcedx-sobjects-faux-generator/package.json [hunk 2]
```diff
@@ -9,7 +9,7 @@
     "vscode": "^1.49.3"
   },
   "dependencies": {
-    "@salesforce/core": "2.28.4",
+    "@salesforce/core": "^2.35.0",
```
gbockus-sf, line 12: This is now the version specific the same as it is in SRD and apex-node to avoid version mismatches.

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -25,11 +25,11 @@
...
   ],
   "dependencies": {
-    "@salesforce/apex-node": "0.2.9",
+    "@salesforce/apex-node": "0.10.0",
     "@salesforce/apex-tmlanguage": "1.4.0",
-    "@salesforce/core": "2.28.0",
+    "@salesforce/core": "^2.35.0",
     "@salesforce/salesforcedx-utils-vscode": "54.2.0",
-    "@salesforce/source-deploy-retrieve": "5.0.0",
+    "@salesforce/source-deploy-retrieve": "5.12.0",
```
gbockus-sf, line 32: this is the latest SDR.  Will need to bump 1 more time after readFileSync is updated.

packages/salesforcedx-apex-debugger/tsconfig.json
mohanraj-r: As discussed in standup, if there is no other way to specifically skip checks for selected libs, then we can go with this at the risk of skipping type checks and potentially introducing type errors until this is fixed. 
This adds more motivation to update typescript/tsconfig. 

Can you do a quick check on what happens if we update the ES target in tsconfig(s) in this repo/packages to match  the target libs?

gbockus-sf: This article was pretty good for skipLibCheck: https://dd.engineering/blog/typescript-the-skiplibcheck-option-explained
I verified there is no option to specific it per module. 

I can try altering our target to match SDR. However I have low expectation b/c then we'll be different the node-apex (and some others I'm sure). That def feel like it belongs in the ts upgrade work vs here.

gbockus-sf: attempted the updated and ran into the other way problem with `@salesforce/core` 
```
@salesforce/salesforcedx-utils-vscode: node_modules/@salesforce/core/lib/sfdxError.d.ts(152,9): error TS2380: 'get' and 'set' accessor must have the same type.
@salesforce/salesforcedx-utils-vscode: node_modules/@salesforce/core/lib/sfdxError.d.ts(153,9): error TS2380: 'get' and 'set' accessor must have the same type.
```

packages/salesforcedx-sobjects-faux-generator/package.json
mohanraj-r: Is there a way to ensure this is in sync with SDR and apex-node going forward?

mohanraj-r (approved):
LGTM

packages/salesforcedx-sobjects-faux-generator/package.json
gbockus-sf: not that I'm aware of out of the box, but this is a fairly unique use case.  I don't totally understand at this point how the different deps are all interconnected and how this results in the version issue encountered here.    I'll create a ticket for tracking and figuring out the issue

PR #3869, opened 2022-03-01
fix: add telemetry to manifest builder @W-10754639@

jag-sfdc:
### What does this PR do?
Move manifest builder to use LibraryExecutor so we get telemetry for free

@W-10754639@

### Functionality Before
No telemetry for manifest builder

### Functionality After
Telemetry for manifest builder. No change in functionality.


randi274 (approved):
Code and local changes look great! Tested save and escape paths. Did not test telemetry since I still don't have access there, but those events should be registered now.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -5,16 +5,47 @@
...
+  constructor(sourcePaths: string[], responseText: string | undefined) {
+    super(
+      nls.localize('force_create_manifest'),
+      'force_create_manifest',
+      OUTPUT_CHANNEL
+    );
+    this.sourcePaths = sourcePaths;
+    this.responseText = responseText;
+  }
+  public async run(response: ContinueResponse<string>, progress?: vscode.Progress<{ message?: string | undefined; increment?: number | undefined; }>, token?: vscode.CancellationToken): Promise<boolean> {
```
gbockus-sf, line 34: did prettier run? This line is super long

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -5,16 +5,47 @@
...
 
 const DEFAULT_MANIFEST = 'package.xml';
 
+export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
+  private sourcePaths: string[];
+  private responseText: string | undefined;
+  constructor(sourcePaths: string[], responseText: string | undefined) {
+    super(
+      nls.localize('force_create_manifest'),
+      'force_create_manifest',
```
gbockus-sf, line 28: can we pull this out into a constant?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -24,19 +55,18 @@ export async function forceCreateManifest(
     uris.push(sourceUri);
   }
   const sourcePaths = uris.map(uri => uri.fsPath);
+  const inputOptions = {
+    placeHolder: nls.localize('manifest_input_save_placeholder'),
+    prompt: nls.localize('manifest_input_save_prompt')
```
gbockus-sf, line 60: nit: constant

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jag-sfdc: Yes, it did run and didn't complain. I can add a few new lines.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jag-sfdc: This is a common theme across all executors. This is part of the whole story to consolidate telemetry constants to a separate file. Until then, I can definitely extract into a constant in the local file.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -24,19 +55,18 @@ export async function forceCreateManifest(
...
-      openUntitledDocument(componentSet);
-    } else {
-      saveDocument(responseText, componentSet);
-    }
+    const commandlet = new SfdxCommandlet(
+      new SfdxWorkspaceChecker(),
+      new FilePathGatherer(sourceUri),
+      new ManifestCreateExecutor(sourcePaths, responseText)
+    );
+    await commandlet.run();
```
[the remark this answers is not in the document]
jag-sfdc: Sure! Btw, some of the integration tests are failing in circleci and I can repro it now locally. I'll fix them and add this test as well. I haven't tested telemetry in AppInsights. I need vsix files for telemetry to be published. Will do that shortly. FYI @randi274

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
randi274: I'm not opposed to this getting changed, but I think we very rarely use constants when calling nls.localize. Is this something we should go back and clean up everywhere?

gbockus-sf: 100%.  There should be a single source of truth for what this string value.  I'm always in favor of letting the computer do the work instead of the dev when it comes to remembering what a string value should be.  Benefits us when referencing the value in code as well as verifying it in test.

gbockus-sf: FWIW I'm also a big fan of doing it different now even if we can't go update all the things.  So even if we haven't done it that way in the past if it's a better way then we should start doing it now and adjust as we go. Also I can get out voted :P

jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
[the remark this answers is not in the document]
jag-sfdc: @gbockus-sf - I timeboxed my effort to use proxyquire and couldn't get it in. Looks like there is quite a bit wiring that has to be done precisely for this to work and don't think I got everything right. I'll go ahead and merge the way it is now. We should definitely revisit if proxyquire is the right away to remove vscode dependencies and if yes, any way to make it simpler.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
jag-sfdc: @gbockus-sf @randi274 - I made this change for now. Let's try and enforce this for future changes and we should come back and make one sweeping change at a later point. It is really hard to find the telemetry strings every time to query something on appinsights.

randi274: Cool! Sounds like a worthy effort to me, just wanted to make sure we're ready to enforce it going forward 😎

PR #3875, opened 2022-03-02
Simplify Release by creating changelog as part of release branch cut - Part 1

jag-sfdc:
### What does this PR do?
Simplifies the weekly release process by creating changelog for the release when the automated process cuts the release branch. The functions are mostly copied from change-log-generator.js. Part2 of this will have the existing change log generator call these functions rather than their local ones.

@W-10785024@

### Functionality Before
N/A

### Functionality After
No functionality change to the end user


scripts/change-log-generator-utils.js
```diff
@@ -0,0 +1,274 @@
+const constants = require('./change-log-constants');
```
gbockus-sf, line 1: dumb q: why do we write our script in straight JS instead of TS?

scripts/create-release-branch.js
```diff
@@ -89,10 +100,7 @@ shell.exec('git add lerna.json');
 shell.exec(`git commit -m "chore: update to version ${nextVersion}"`);
 
 // Push version update commits to develop
-shell.exec(`git push origin develop`);
-
-// Create the new release branch and switch to it
-shell.exec(`git checkout -b ${releaseBranchName}`);
+// shell.exec(`git push origin develop`);
```
gbockus-sf, line 103: nit: delete.  no dead code.

scripts/change-log-generator-utils.js
```diff
@@ -0,0 +1,274 @@
...
+  return filteredHeaders;
+}
+
+/**
+ * Generate the key to be used in the grouped messages map. This will help us
+ * determine whether this is an addition or fix, along with the package header
+ * that the commit should be inserted under.
+ *
+ * If we have a type that should be ignored, return an empty key.
+ */
```
gbockus-sf, line 243: 😂  jsdoc FTW. Should you go ahead and include the params? more vscode friendly

scripts/change-log-generator-utils.js
```diff
@@ -0,0 +1,274 @@
...
+      message = message.trim();
+      map[MESSAGE] = message.charAt(0).toUpperCase() + message.slice(1);
+      map[FILES_CHANGED] = getFilesChanged(map[COMMIT]);
+      map[PACKAGES] = getPackageHeaders(map[FILES_CHANGED]);
+    }
+  }
+
+  console.log('\nCommit: ' + commit);
+  console.log('Commit Map:');
+  console.log(map);
```
gbockus-sf, line 114: did you purposely leave these consoles here?

scripts/change-log-generator-utils.js
```diff
@@ -0,0 +1,274 @@
...
+  'chore',
+  'style',
+  'refactor',
+  'test',
+  'build',
+  'ci',
+  'revert'
+];
+
+module.exports = {
```
gbockus-sf, line 32: should this go at the bottom?  In regular js land it's usually not good to use a thing before you declare it.

scripts/change-log-generator-utils.js
jag-sfdc: To save an extra step to compile? I assume this was started when our release process was simple, but over time it grew and now I too felt having them in TS would help.

scripts/create-release-branch.js
jag-sfdc: I'm on the fence about this one. This would eliminate the cherry-pick process to add one of changes, but the downside is that we are no longer tied to the week. I think the last time we discussed the majority seemed to agree to this, so let me just delete it.

scripts/change-log-generator-utils.js
jag-sfdc: :) I'll add them to other functions as well.

scripts/change-log-generator-utils.js
jag-sfdc: Didn't know about that about exports. I can definitely move it to the bottom if that's a better JS practice.

scripts/change-log-generator-utils.js
jag-sfdc: I moved most of these functions from changelog generator and this is something I want to leave it as is.

mohanraj-r (approved):

scripts/change-log-generator-utils.js
gbockus-sf: cool.  I'll create a ticket.  We can use ts-node to run them as well so compile happens on the fly

gbockus-sf (approved):
cool

Issue #3806, opened 2022-02-04
Unable to retrieve Metadata of sub folder with extension


ronika-das:
### Summary

If the metadata is available inside another folder(i.e. it is under a sub folder), the Org Browser, doesn;t have the option to retrieve it

### Steps To Reproduce:

1.  Open Org Browser-> Open Report. Lets say we a Report folder called ABC and inside that we have few reports. It doesnt retrieve the reports within the sub folder
2. In the attached screenshot, There is a report Folder named work_order_reports. Inside that we have 10-15 reports. We are not able to retrieve that with the org browser

### Expected result

It should allow us to retrieve the metadata files within sub folders

### Actual result

It is not allowing to retrieve

### Additional information

![image](https://user-images.githubusercontent.com/32714099/152559453-957df2c2-cf43-4727-9d18-85e4e00d3d23.png)

**Salesforce Extension Version in VS Code**: 53.16.0

**SFDX CLI Version**:sfdx-cli/7.122.1 win32-x64 node-v14.18.0

**OS and version**: Windows OS , 10.0.19042 Build  19042

jorgemontalvao:
the same for me

AnanyaJha:
Hi @ronika-das thanks for reporting this issue! The team is aware of this and we'll be addressing this bug in the current iteration. We'll post here with updates

ronika-das:
Thank you @AnanyaJha  :)

trinasfdx:
@W-10541413@

randi274:
This has now been delivered with the latest version of the extensions (v54.6.1), with the exception of custom fields. We're tracking that separately with #2711, so we'll close this one for now.

jorgemontalvao:
> This has now been delivered with the latest version of the extensions (v54.6.1), with the exception of custom fields. We're tracking that separately with #2711, so we'll close this one for now.

@randi274 @AnanyaJha @ronika-das 
It works perfectly now! Thank you very much for all your effort here!

AnanyaJha:
No problem, hope you enjoy using the Extensions! Feel free to leave a review or rating here - https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode&ssr=false#review-details

Issue #3702, opened 2021-12-16
Can not retrieve the email templates from the org browser. 


anshumishra07sstk:
### Summary

Clicking on an email template from the org browser doesn't give an option to retrieve the code.

### Steps To Reproduce:

1.  Connect to Default org (My SF sandbox)
1.  Go to Org Browser and look for Email Templates Metadata.
1. Choose the correct folder. It shows the email template name.

### Expected result

_After right clicking the email template name, there should be an option to retrieve the code.
_And selected email should be added to the branch.

### Actual result

_Email Template name doesn't show an option to retrieve the code.
_Nothing gets added to the branch. There is no option to pull the email template into the branch for deployment.

### Additional information

_Feel free to attach a [screenshot_.](url)

**Salesforce Extension Version in VS Code**:

**SFDX CLI Version**:

**OS and version**:

anshumishra07sstk:
https://www.loom.com/share/61ebc6b3ca9b46a08edb725bd2349e1f 

The org browser doesn't retrieve the email templates and the explorer just shows email template folders but not the actual templates under "email" metadata.

Please take a look and if possible schedule a call to fix this issue. Thank you!

AnanyaJha:
Hi @anshumishra07sstk thanks for reporting this! Are you seeing this behavior with any other metadata types in the org browser? Could you also share what version of the Salesforce Extensions for VS Code you're using? This can be found in the Extensions Marketplace, thanks!

anshumishra07sstk:
Hi Ananya,

Sorry I couldn't reply earlier.

I see this problem with other data types too.
Today I added a new field in an object and when I opened VS Code Org
Browser, I saw that field but after doing the right click, there was no
option to retrieve that field so I retrieved the object itself. That adds
all the changes to my new branch.

However, to add all those changes in the branch, choose mine and discard
others; makes the whole process too time consuming.
But at least that option works for other metadata but there is nothing for
email templates deployment.


[image: image.png]

I will appreciate it if you can schedule a call and help me fix this issue.
I will be available from 3:00-5:00 PM EST today. If that doesn't work for
you, please let me know.

Thank you!

Regards,
Anshu Mishra
CRM Administrator, Global Contact Center Operations


Empire State Building
350 Fifth Avenue, 21st Floor
New York, NY 10118
www.shutterstock.com

RitamAgrawal:
[...]

PR #3892, opened 2022-03-08
Ra/retrieve components within folders

RitamAgrawal:
### What does this PR do?
In Org Browser, there are metadata types, such as CustomObject, Dashboard, Document, EmailTemplate and Report which has folders. This one makes the components within these folders retrievable
#3702, #3806, @W-10660751@, @W-10541413@

### Functionality Before
1. Components within folders didn't have 'Retrieve Source from Org' and 'Open Source and Retrieve' inline options.
2. Components within folders were not retrievable. 

### Functionality After
1. Components within folders have 'Retrieve Source from Org' and 'Open Source and Retrieve' inline options.
2. Components within folders are retrievable.

Untitled 4


packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
```diff
@@ -176,6 +176,46 @@ describe('load org browser tree outline', () => {
...
+
+    const metadataObject = {
+      xmlName: 'foldertypeNode1',
+      directoryName: 'testDirectory',
+      suffix: 'cls',
+      inFolder: false,
+      metaFile: false,
+      label: 'Folder Type Node 1'
+    };
+    for (const folderMetadataType of folderTypeMetadata){
```
jag-sfdc, line 206: It's better to remove for loops and "if" conditions in tests. Can we expand this test into multiple tests to test different folder metadata types?

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -84,10 +84,10 @@ export class MetadataOutlineProvider
         break;
       case NodeType.Folder:
       case NodeType.MetadataType:
-        let nodeType = NodeType.MetadataComponent;
+        let nodeType: NodeType = NodeType.MetadataComponent;
         if (TypeUtils.FOLDER_TYPES.has(element.fullName)) {
           nodeType = NodeType.Folder;
-        } else if (element.type === NodeType.Folder && element.fullName) {
+        } else if (element.parent && element.parent.fullName === 'CustomObject') {
```
mohanraj-r, line 90: I am guessing 'CustomObject' has element.type as NodeType.Folder ?

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
```diff
@@ -205,24 +391,65 @@ describe('load org browser tree outline', () => {
...
+    };
+    const parentNode = new BrowserNode(
+    'Custom Objects',
+    NodeType.MetadataType,
+    'CustomObject',
+    customObjectmetadataObject
+    );
+    parentNode.setComponents(['TestAccount', 'TestCleanInfo'], NodeType.Folder);
+    const childNodes = parentNode.children;
+    // @ts-ignore
```
mohanraj-r, line 427: when suppressing/bypassing checks it would be good to specify what is being suppressed (the error details that is being ignored) and maybe also why

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
gbockus-sf, line 90: nit: Can you make 'CustomObject' a constant? 
Also this logic change hurts my brain a little.  So previous
if it's a folder and has a fullName then it's a MetadataField
now
if it has a parent and it's parent has a full name that's equal to Custom Object then it's a MetadataField. 

Does this still work for not things in inside folders?

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
```diff
@@ -351,6 +578,154 @@ describe('load org browser tree outline', () => {
...
+      label: 'Report Folder'
+    };
+    const reportFolderNode = new BrowserNode(
+      'Report folder',
+      NodeType.Folder,
+      'ReportFolder',
+      metadataObject
+    );
+    const cmpsNodes = await metadataProvider.getChildren(reportFolderNode);
+    compareNodes(cmpsNodes, expected);
```
mohanraj-r, line 725: minor refactoring suggestion: since the `expected` has a simple and similar structure across the diff tests, wonder if would make sense to refactor it out e.g. have a util func e.g. getExpectedNodes(name, type) that returns the expected nodes and then call it from `compareNodes(..)` in place of expected

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
```diff
@@ -351,6 +578,154 @@ describe('load org browser tree outline', () => {
...
+    getCmpsStub.restore();
```
mohanraj-r, line 726: Ideally cleanup should be done as part of the test(suite) teardown or in try..finally so that if a test fails it still gets called. But if thats how all existing tests are done, it's fine for now.

mohanraj-r (commented):
LGTM. Some minor comments, questions.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
```diff
@@ -176,6 +176,192 @@ describe('load org browser tree outline', () => {
...
+        fullName: 'Account',
+        type: NodeType.Folder
+      },
+      {
+        label: 'AccountCleanInfo',
+        fullName: 'AccountCleanInfo',
+        type: NodeType.Folder
+      }
+    ];
+    const getCmpsStub = stub(
```
gbockus-sf, line 192: nit: can we pull this out into a beforeEach for all these related tests? 
It'd look something like
```
let getComponentsStub: Sinon.Stub;

beforeEach(() => {
  getComponentsStub = MetadataOutlineProvider.prototype, 'getComponents');
});

afterEach(() => {
  getComponentsStub.restore(); // If you were using mocha sandbox functionality you could make this easier 
});

it('imma test', () => {
  getComponentsStub.returns(expected.map(n => n.fullName));
  ....
});

File: packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts (hunk 2 above)
Comment (RitamAgrawal) [reply packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts:90]: Yes, rest everything listed under metadata type is set to the default i.e. MetadataComponent, be it within folder or without folders.

Comment (RitamAgrawal) [reply packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts:90]: `Custom Objects` metadata type has element.type as NodeType.MetadataType. `Account` , an object listed under `Custom Objects` has element.type as NodeType.Folder.

Comment (mohanraj-r) [reply packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts]: So this is the source of the bug and not just a refactor ?

File: packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts [hunk 7]
```diff
@@ -416,6 +529,32 @@ function compareNodes(actual: BrowserNode[], expected: any[]) {
...
+    },
+    {
+      label: label2,
+      fullName: label2,
+      type
+    }
+  ];
+}
+
+function getMetadataObject(label: string) {
```
mohanraj-r, line 547: Refactor looks good - thanks 👍 . As the metadata object returned from here seems to be used only creating a new BrowserNode, wonder if it would make sense to extract creation of BrowserNode as the util method here instead.
But I will leave it up to you since I failed to notice this in the last review iteration.

mohanraj-r (commented):
@RitamAgrawal Changes look good. Have couple of suggestions - please take a look.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
gbockus-sf: agreed.  Looks like this has been removed but for future reference also big +1 to doc'ing why we need to tsignore if we do.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -21,6 +21,8 @@ const validManageableStates = new Set([
   undefined // not part of a package
 ]);
 
+const CUSTOMOBJECTS_FULLNAME = 'CustomObject';
```
gbockus-sf, line 24: nit: why not export this here and use it in `metadataOutlineProvider` as well?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
RitamAgrawal: Yes! it's the source of the bug.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
RitamAgrawal: Thanks @mohanraj-r setting a default value here, takes care of the ts error.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataOutlineProvider.test.ts
RitamAgrawal: Thank you! Apart from metadataObject, other parameters like label, nodeType or fullName are all different each time we are creating a BrowserNode. `BrowserNode()` itself is taking care of the node creation. I think it's better this way.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -21,6 +21,8 @@ const validManageableStates = new Set([
   undefined // not part of a package
 ]);
 
+export const CUSTOMOBJECTS_FULLNAME = 'CustomObject';
```
mohanraj-r, line 24: Can we use `registry.types.customobject.name` from SDR (`import { registry } from '@salesforce/source-deploy-retrieve') instead ?

mohanraj-r (commented):
Just one minor comment from recent changes.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
mohanraj-r: Can you also please try to use `registry.types..` similarly for other types for the changes in the PR as well.

RitamAgrawal: No. By using `` registry.types.customobject.name from SDR (`import { registry } from '@salesforce/source-deploy-retrieve')``  I can get the type of the property of CustomObject like `string` but not the value of customobject.fullname .

mohanraj-r: From the docs https://github.com/forcedotcom/source-deploy-retrieve/blob/main/HANDBOOK.md#the-registry-object and the example given there looks like `name` would be `fullname`. But we can double check with CLI team.

mohanraj-r: Also please refer https://github.com/forcedotcom/salesforcedx-vscode/blob/e3ce5027f59a6ba735bcd78141b9efac66afa879/packages/salesforcedx-vscode-core/src/commands/util/parameterGatherers.ts#L195-L199

mohanraj-r: From the conversation with CLI Looks like the full name and type name are different.

mohanraj-r (approved):

RitamAgrawal:
Querying types and searching the registry is oftentimes easier and cleaner using registryAccess and contains built-in checking for whether or not a metadata type exists.
```
+import { RegistryAccess, registry } from '@salesforce/source-deploy-retrieve';
-export const CUSTOMOBJECTS_FULLNAME = 'CustomObject';
+const registryAccess = new RegistryAccess();
+export const CUSTOMOBJECTS_FULLNAME = registryAccess.getTypeByName('customobject').name
```

mohanraj-r:
@randi274 Can you please check if the linked issues can be closed.

PR #3916, opened 2022-03-11
build: remove unused yeoman generator dep from core pkg

mohanraj-r:
### What does this PR do?
- remove unused yeoman generator dep from core pkg

@W-10832373@

### Functionality Before
- `salesforcedx-vscode-core/package.json` listed `yeoman-generator` as a dependency

### Functionality After
- `salesforcedx-vscode-core/package.json` does not list `yeoman-generator` as a dependency


gbockus-sf:
@mohanraj-r can you create an associated WI and be sure it's on the current kanban board?  Just so we have it for tracking. Thanks.

gbockus-sf (approved):
cool

PR #3923, opened 2022-03-16
feat: rename lightning component bundles

floralan:
### What does this PR do?
Provide a new command: "SFDX: Rename Component" to rename LWC or Aura component. All files(js, html, css, js-meta.xml) for LWC component or all files(auradoc, cmp, cmp-meta.xml, css, design, svg, contoller, helper, renderer, js) for Aura component and the folder are automatically renamed with a new name that user provides. 

@W-10288367@

### Functionality Before
Right click "rename" to rename the files(at least 4 essential files for LWC and 9 essential files for Aura) and the folder one by one manually. 

### Functionality After
Right click with the new command: "SFDX: Rename Component", input the new component name, then all the files (that matches with the component name) and folder are renamed automatically.


packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
jeffb-sfdc, line 2: since this is a new file, the copyright should be 2022

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+      return false;
+  }
+}
+
+export async function forceRenameLightningComponent(sourceUri: vscode.Uri) {
+  const sourceFsPath = sourceUri.fsPath;
+  const inputOptions = {
+    placeHolder: nls.localize(RENAME_INPUT_PLACEHOLDER),
+    promopt: nls.localize(REAME_INPUT_PROMPT)
```
jeffb-sfdc, line 57: I think "REAME_INPUT_PROMPT" is meant to be "RENAME_INPUT_PROMPT"

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+          path.join(componentPath, item),
+          path.join(componentPath, newItem));
+      }
+    }
+  }
+  const newComponentPath = path.join(path.dirname(componentPath), newName);
+  fs.renameSync(
+    componentPath,
+    newComponentPath);
+  notificationService.showWarningMessage(nls.localize(RENAME_COMP_WARNING));
```
jeffb-sfdc, line 91: I'm not a fan of abbreviating code, and prefer code to be verbose.  Could you rename RENAME_COMP_WARNING to RENAME_COMPONENT_WARNING?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function getComponentPath(sourceFsPath: string): string {
+  const stats = fs.statSync(sourceFsPath);
+  return stats.isFile() ? path.dirname(sourceFsPath) : sourceFsPath;
+}
+
+function checkForDuplicateName(componentPath: string, newName: string) {
+  if (isDuplicate(componentPath, newName)) {
+    notificationService.showErrorMessage(nls.localize(REANME_INPUT_DUP_ERROR));
+    throw new Error(format(nls.localize(REANME_INPUT_DUP_ERROR)));
```
jeffb-sfdc, line 102: `nls.localize(REANME_INPUT_DUP_ERROR)` is called twice here.  How about something like:
```
errorMessage = nls.localize(REANME_INPUT_DUP_ERROR)
```
...and then use `errorMessage`?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+function isDuplicate(componentPath: string, newName: string): boolean {
+  const isLwc = isLwcComp(componentPath);
```
jeffb-sfdc, line 107: I'd prefer it if `isLwcComp()` was named, `isLwcComponent()`

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+  const lwcPath = isLwc ? path.dirname(componentPath) : path.join(path.dirname(path.dirname(componentPath)), 'lwc');
+  const auraPath = isLwc ? path.join(path.dirname(path.dirname(componentPath)), 'aura') : path.dirname(componentPath);
```
jeffb-sfdc, line 109: `path.dirname(componentPath)` occurs here in 4 places.  How about this instead:
```
const componentPathDirName = path.dirname(componentPath);
const lwcPath = isLwc ? componentPathDirName : path.join(path.dirname(componentPathDirName), 'lwc');
const auraPath = isLwc ? path.join(path.dirname(componentPathDirName), 'aura') : componentPathDirName;
```

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (fs.existsSync(path.join(lwcPath, newName)) || fs.existsSync(path.join(auraPath, newName))) {
+    return true;
+  }
+  return false;
```
jeffb-sfdc, line 113: There are are lot "if LWC else Aura" taking place here.  I think the LWC code should be moved into its own function, and the Aura code moved into its own function.  So something like this:
```
function isDuplicate(componentPath: string, newName: string): boolean {
  if (isLwcComp(componentPath)) {
    return isDuplicateLwcComponent((componentPath, newName);
  } else {
    return isDuplicateAuraComponent((componentPath, newName);
  }
```

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (isLwc) {
+    regularExp = new RegExp(`${componentName}\.(html|js|js-meta.xml|css|svg)`);
+  } else {
+    regularExp = new RegExp(`${componentName}(((Controller|Renderer|Helper)?\.js)|(\.(cmp|app|css|design|auradoc|svg)))`);
+  }
+  return item.match(regularExp) ? true : false;
+}
+
+function isLwcComp(componentPath: string): boolean {
+  return path.basename(path.dirname(componentPath)) === 'lwc' ? true : false;
```
jeffb-sfdc, line 128: since a string comparison is taking place, a call to `.toLowerCase()` should probably be added.
```
return path.basename(path.dirname(componentPath)).toLowerCase() === 'lwc' ? true : false;
```

packages/salesforcedx-vscode-core/package.json
```diff
@@ -305,6 +305,10 @@
         {
           "command": "sfdx.create.manifest",
           "when": "sfdx:project_opened"
+        },
+        {
+          "command": "sfdx.lightning.rename",
+          "when": "sfdx:project_opened && resource =~ /.*/(lwc|aura)/[^/]+(/[^/]+\\.(html|css|js|xml|svg|cmp|app|design|auradoc))?$/"
```
jeffb-sfdc, line 311: If I"m reading this correctly, the command is available when the parent is named lwc or aura, and the file type is html, css, js, xml, svg, cmp, app, design, auradoc.  Wouldn't we want to include all files?  What if the file type was .TS, or .PNG?  Are there any file types we'd want to exclude?  If we want to include all, can the RegEx be changed to include all?  Maybe this:
```
"when": "sfdx:project_opened && resource =~ /.*/(lwc|aura)/[^/]+(/[^/]+\\.*)?$/"
```

RegEx isn't my strongest suit, but I thin you will get the idea.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
RitamAgrawal: Is the function `isLwcComp` case sensitive?

floralan: @RitamAgrawal @jeffb-sfdc  'lwc' folder in a SFDX project is always lower-case, which is defined when user create a new SFDX project.

packages/salesforcedx-vscode-core/package.json
floralan: From my understanding, if users need JPG files, they have to store them in somewhere else like the assets folder that is located outside of lwc and aura component folder.

jeffb-sfdc: OIC, Yea, images are normally in the `staticresources` directory, and I hadn't heard about images being within the components directory.  I looked into it, and the image is used for an icon in AppBuilder.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
jeffb-sfdc: Got it.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @jeffb-sfdc We need to check duplication in bot LWC and Aura path instead of either of them. But, I revised my code to use only one "if lwc else" logic here to add more readability.

jeffb-sfdc: With the updated code, there looks to be a logic flip -the "isLwcComponent is true" condition has:
`auraPath = path.join(path.dirname(componentPathDirName), 'aura');`

...shouldn't it be lwc instead of Aura?

jeffb-sfdc: nm, I think I understand now... when renaming, we need to check for the name in `/lwc` as well as `/aura`.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as vscode from 'vscode';
+import { OUTPUT_CHANNEL } from '../channels';
+import { nls } from '../messages';
+import { FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
+
+const RENAME_LIGHTNING_COMPONENT_EXECUTOR = 'force_rename_lightning_component';
+const RENAME_INPUT_PLACEHOLDER = 'rename_component_input_placeholder';
+const RENAME_INPUT_PROMPT = 'rename_component_input_prompt';
+const RENAME_INPUT_DUP_ERROR = 'rename_component_input_dup_error';
+const RENAME_COMPONENT_WARNING = 'rename_component_warning';
```
jeffb-sfdc, line 23: nit pick:  Go make the names consistent, `RENAME_COMPONENT_WARNING` should be `RENAME_WARNING` (or add `_COMPONENT` to RENAME_INPUT_PLACEHOLDER, RENAME_INPUT_PROMPT, and RENAME_INPUT_DUP_ERROR)

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,137 @@
...
+    await commandlet.run();
+  }
+}
+
+function renameComponent(sourceFsPath: string, newName: string) {
+  const componentPath = getComponentPath(sourceFsPath);
+  const componentName = path.basename(componentPath);
+  checkForDuplicateName(componentPath, newName);
+  const items = fs.readdirSync(componentPath);
+  if (items) {
```
jeffb-sfdc, line 76: The `if(items)` here isn't needed and can be removed.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,137 @@
...
+        fs.renameSync(
+          path.join(componentPath, item),
+          path.join(componentPath, newItem));
+      }
+    }
+  }
+  const newComponentPath = path.join(path.dirname(componentPath), newName);
+  fs.renameSync(
+    componentPath,
+    newComponentPath);
```
jeffb-sfdc, line 90: Nit pick: when placing args on separate lines (which I like), I think our style guild is to place the closing parenthesis on a new line, like so:
```
fs.renameSync(
    componentPath,
    newComponentPath
);
 ```

File: packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts [hunk 13]
```diff
@@ -0,0 +1,138 @@
...
+      return false;
+  }
+}
+
+export async function forceRenameLightningComponent(sourceUri: vscode.Uri) {
+  const sourceFsPath = sourceUri.fsPath;
+  const inputOptions = {
+    placeHolder: nls.localize(RENAME_INPUT_PLACEHOLDER),
+    promopt: nls.localize(RENAME_INPUT_PROMPT)
+  } as vscode.InputBoxOptions;
```
jeffb-sfdc, line 58: I'm starting to QA this.  As I use this new command, one thing I wish was here, was when I go to rename, I wish the input box was pre-populated with the existing name.  I took a look, and it's the `value` field of InputBoxOptions which needs to be set.  What are your thoughts on pre-populating the field?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const RENAME_LIGHTNING_COMPONENT_EXECUTOR = 'force_rename_lightning_component';
+const RENAME_INPUT_PLACEHOLDER = 'rename_component_input_placeholder';
+const RENAME_INPUT_PROMPT = 'rename_component_input_prompt';
+const RENAME_INPUT_DUP_ERROR = 'rename_component_input_dup_error';
+const RENAME_WARNING = 'rename_component_warning';
+
+export class RenameLwcComponentExecutor extends LibraryCommandletExecutor<string> {
+  private sourceFsPath: string;
+  private responseText: string | undefined;
```
gbockus-sf, line 27: Nit: you could make this just `private responseText: string;` and make this class a little simpler. Note you'd also update the constructor parameter.   It doesn't look like we handle when responseText is undefined below so it probably makes sense to to the undefined check before constructing this class instance.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+import * as vscode from 'vscode';
+import { OUTPUT_CHANNEL } from '../channels';
+import { nls } from '../messages';
+import { FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
+
+const RENAME_LIGHTNING_COMPONENT_EXECUTOR = 'force_rename_lightning_component';
+const RENAME_INPUT_PLACEHOLDER = 'rename_component_input_placeholder';
+const RENAME_INPUT_PROMPT = 'rename_component_input_prompt';
+const RENAME_INPUT_DUP_ERROR = 'rename_component_input_dup_error';
+const RENAME_WARNING = 'rename_component_warning';
```
gbockus-sf, line 23: ❤️

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.responseText = responseText;
+  }
+
+  public async run(
+    response: ContinueResponse<string>,
+    progress?: vscode.Progress<{ message?: string | undefined; increment?: number | undefined; }>,
+    token?: vscode.CancellationToken
+    ): Promise<boolean> {
+      if (this.sourceFsPath) {
+        if (this.responseText) {
```
gbockus-sf, line 44: nit:  you can combine these into a single if statement

gbockus-sf: Should we add a log or telemetry for when we encounter the case where either of these values is undefined?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
gbockus-sf: good idea

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export async function forceRenameLightningComponent(sourceUri: vscode.Uri) {
+  const sourceFsPath = sourceUri.fsPath;
+  const inputOptions = {
+    placeHolder: nls.localize(RENAME_INPUT_PLACEHOLDER),
+    promopt: nls.localize(RENAME_INPUT_PROMPT)
+  } as vscode.InputBoxOptions;
+
+  const responseText = await vscode.window.showInputBox(inputOptions);
+  if (sourceFsPath) {
```
gbockus-sf, line 61: should this go before we ask for responseText?  sourceFsPath may never be falsy, but there's no point in asking for the user input is we can't proceed here.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
+/*
```
gbockus-sf, line 1: nit: b/c naming is always a pain.  Should this be `forceRenameUiComponent`?  I could see looking for where we rename aura components and discounting this file b/c it's explicitly calls out LightningComponent. 

If we want to keep it I could also see adding a JSDoc to the executor to document that this is for both LWC and Aura components.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+    const commandlet = new SfdxCommandlet(
+      new SfdxWorkspaceChecker(),
+      new FilePathGatherer(sourceUri),
+      new RenameLwcComponentExecutor(sourceFsPath, responseText)
+    );
+    await commandlet.run();
+  }
+}
+
+function renameComponent(sourceFsPath: string, newName: string) {
```
gbockus-sf, line 71: This smells like a service method vs something that should live in the command.  The command code should only really concern itself with command execution.  The check for existing and actual renaming should live elsewhere.  Maybe under a services directory in core or utils.  It looks like we have some one off services here and there but nothing organizing the logic. @jag-sfdc any thoughts or collections of services that I'm unaware of in the codebase?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+function isDuplicate(componentPath: string, newName: string): boolean {
+  // A LWC component can't share the same name as a Aura component
+  const componentPathDirName = path.dirname(componentPath);
+  let lwcPath: string;
+  let auraPath: string;
+  if (isLwcComponent(componentPath)) {
+    lwcPath = componentPathDirName;
+    auraPath = path.join(path.dirname(componentPathDirName), 'aura');
+  } else {
+    lwcPath = path.join(path.dirname(componentPathDirName), 'lwc');
```
gbockus-sf, line 116: nit: can you make 'lwc' and 'aura' constants?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+    auraPath = componentPathDirName;
+  }
+  const allLwcComponents = fs.readdirSync(lwcPath);
+  const allAuraComponents = fs.readdirSync(auraPath);
```
gbockus-sf, line 120: more syncs that could be async

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (allLwcComponents.includes(newName) || allAuraComponents.includes(newName)) {
```
gbockus-sf, line 121: nit: you could just return this vs having the if/else.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (isLwc) {
+    regularExp = new RegExp(`${componentName}\.(html|js|js-meta.xml|css|svg)`);
+  } else {
+    regularExp = new RegExp(`${componentName}(((Controller|Renderer|Helper)?\.js)|(\.(cmp|app|css|design|auradoc|svg)))`);
+  }
+  return item.match(regularExp) ? true : false;
+}
+
+function isLwcComponent(componentPath: string): boolean {
+  return path.basename(path.dirname(componentPath)) === 'lwc' ? true : false;
```
gbockus-sf, line 139: another place to use that 'lwc' constant

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+  const allLwcComponents = fs.readdirSync(lwcPath);
+  const allAuraComponents = fs.readdirSync(auraPath);
+  if (allLwcComponents.includes(newName) || allAuraComponents.includes(newName)) {
+    return true;
+  }
+  return false;
+}
+
+function isNameMatch(item: string, componentName: string, componentPath: string) {
```
gbockus-sf, line 127: we should have a test for this method.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+  const isLwc = isLwcComponent(componentPath);
+  let regularExp: RegExp;
+  if (isLwc) {
+    regularExp = new RegExp(`${componentName}\.(html|js|js-meta.xml|css|svg)`);
+  } else {
+    regularExp = new RegExp(`${componentName}(((Controller|Renderer|Helper)?\.js)|(\.(cmp|app|css|design|auradoc|svg)))`);
+  }
+  return item.match(regularExp) ? true : false;
```
gbockus-sf, line 135: does item.match return true/false?  If so just return it.

gbockus-sf (changes requested):
Had some thoughts.  Give em a read and let me know what you think.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -27,6 +27,7 @@
   "force_lightning_interface_create_text": "SFDX: Create Aura Interface",
   "force_lightning_lwc_create_text": "SFDX: Create Lightning Web Component",
   "force_lightning_lwc_test_create_text": "SFDX: Create Lightning Web Component Test",
+  "force_lightning_rename_component_text": "SFDX: Rename Component (Files Only)",
```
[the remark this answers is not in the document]
floralan: @gbockus-sf a good call! I don't think this command text is a good one either and may bring some confusion to users. I was planning to ask Sonal to revise this.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf  I think "Lightning Component" includes both Aura and LWC according to this [Salesforce Lightning Component Library page](https://developer.salesforce.com/docs/component-library/overview/components).

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf Revised those part of code to be included in a parameter gatherer.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: Because of the if/else logic, it doesn't allow me to make them constants. Is there a better way for this?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+        path.join(componentPath, newItem)
+      );
+    }
+  }
+  const newComponentPath = path.join(path.dirname(componentPath), newName);
+  fs.renameSync(
+    componentPath,
+    newComponentPath
+  );
+  notificationService.showWarningMessage(nls.localize(RENAME_WARNING));
```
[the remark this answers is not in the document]
floralan: Yes, so I think "(Files Only)" in command is redundant.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: If I make them const, can't it be accessed outside of the if/else block?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: I don't quite understand why here needs 'lwc' constant.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf item.match does not return true/false but returns an array that includes matched string,

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf Updated the test file to include this function.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf Does these conditions are already includes into telemetry since this class extends LibraryCommandletExecutor?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
gbockus-sf: Not I mean made string constants for 'aura' and 'lwc'

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
gbockus-sf: It's def a person preference but having the string 'lwc' spread though out our code base can just open us up to typo errors.   I prefer to have a constants file somewhere where we have 
`export const LWC = 'lwc';` that we reference everywhere instead of have the string 'lwc'.  If we only use it in this file then having it just defined in the file is fine.  The most obviously benefit outside of just us have to avoid the mental tracking of when 'lwc' should be use is in tests.  It's much better to use a constant LWC in the source, then verify in the test that LWC was used.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+function renameComponent(sourceFsPath: string, newName: string) {
+  const componentPath = getComponentPath(sourceFsPath);
+  const componentName = path.basename(componentPath);
+  checkForDuplicateName(componentPath, newName);
+  const items = fs.readdirSync(componentPath);
+  for (const item of items) {
+    // only rename the file that has same name with component
+    if (isNameMatch(item, componentName, componentPath)) {
+      const newItem = item.replace(componentName, newName);
+      fs.renameSync(
```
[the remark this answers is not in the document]
gbockus-sf: Let's unresolve this one to discuss in as a team when you get back.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
gbockus-sf: gotcha...returns null if no match.  I still think `return Boolean(item.match(regularExp));` is more readable.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  public async run(
+    response: ContinueResponse<ComponentName>
+    ): Promise<boolean> {
+      const newComponentName = response.data.name;
+      if (newComponentName && this.sourceFsPath) {
+        await renameComponent(this.sourceFsPath, newComponentName);
+        return true;
+      }
+      return false;
```
gbockus-sf, line 46: swinging back around to the telemetry.  We do get a ping about this command being called b/c of the LibraryCommandletExecutor extends...however we don't know if we actually do it as determined by this logic.  It's fine to push that off for now as we need a plan, but I was just calling it out previously.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+
```
gbockus-sf, line 49: I still think everything below here should probably be in a service/util class...but get have have existing patterns we should break away from.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (isLwcComponent(componentPath)) {
+    lwcPath = componentPathDirName;
+    auraPath = path.join(path.dirname(componentPathDirName), AURA);
+  } else {
+    lwcPath = path.join(path.dirname(componentPathDirName), LWC);
+    auraPath = componentPathDirName;
+  }
+  const allLwcComponents = await fs.promises.readdir(lwcPath);
+  const allAuraComponents = await fs.promises.readdir(auraPath);
+  return allLwcComponents.includes(newName) || allAuraComponents.includes(newName) ? true : false;
```
gbockus-sf, line 140: nit: here you can just do 
`return (allLwcComponents.includes(newName) || allAuraComponents.includes(newName));

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,156 @@
...
+  const isNameDuplicate = await isDuplicate(componentPath, newName);
+  if (isNameDuplicate) {
+    const errorMessage = nls.localize(RENAME_INPUT_DUP_ERROR);
+    notificationService.showErrorMessage(errorMessage);
+    throw new Error(format(errorMessage));
+  }
+}
+
+async function isDuplicate(componentPath: string, newName: string): Promise<boolean> {
+  // A LWC component can't share the same name as a Aura component
```
gbockus-sf, line 127: ah good comment that's sneaky

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -0,0 +1,194 @@
...
+          return false;
+        }
+      });
+    });
+
+    afterEach(() => {
+      env.restore();
+    });
+
+    it('should not rename when input text is empty', async () => {
```
gbockus-sf, line 143: total nit: this describe is called Exception Handling...but this test doesn't handle any exceptions.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      } catch (e) {
+        exceptionThrown = true;
+      }
+      expect(exceptionThrown).to.equal(true);
+      expect(renameStub.callCount).to.equal(0);
+    });
+  });
+});
+
+describe ('#isNameMatch', () => {
```
gbockus-sf, line 193: nit: There should only be one top level describe in each test file.  So these are rename tests...this describe block should go in the top level `Force Rename Lightning Component` describe.

gbockus-sf (approved):
some nits to consider. Thanks for they async'ing

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -0,0 +1,156 @@
...
+  const newComponentPath = path.join(path.dirname(componentPath), newName);
+  await fs.promises.rename(
+    componentPath,
+    newComponentPath
+  );
+  notificationService.showWarningMessage(nls.localize(RENAME_WARNING));
+}
+
+function getComponentPath(sourceFsPath: string): string {
+  const stats = fs.statSync(sourceFsPath);
```
gbockus-sf, line 109: can we make this async?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: @gbockus-sf This can be Dev Choice problem that we can discuss within the team later, yes, I followed the existing pattern.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      env.stub(fs.promises, 'readdir')
+      .onFirstCall().resolves([])
+      .onSecondCall().resolves([])
+      .onThirdCall().resolves(itemsInHero);
+      const executor = new RenameLwcComponentExecutor(sourceUri.fsPath);
+      await executor.run({
+        type: 'CONTINUE',
+        data: {name: 'hero1'}
+      });
+      expect(renameStub.callCount).to.equal(5);
```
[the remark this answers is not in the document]
floralan: @gbockus-sf The previous test focus on the call arguments, this test is intended to make sure it only rename the files and folder that have same name with LWC component. Here in this lwc component directory, it has 6 files and only 4 files match the component name, so rename should be called 5 times (plus the folder rename).

jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
floralan: Renamed description to "Exception and corner cases handling"

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
floralan: has updated.

PR #3950, opened 2022-03-24
fix: the SFDX: Create Project command now removes leading and trailing whitespaces from the project name

### LINKED ISSUE #2605: The  SFDX: Create Project  command allows trailing spaces for Project Name
### Summary
The  SFDX: Create Project  command allows trailing spaces for Project Name, which makes files that are not able to be renamed or deleted through Windows File Explorer.

### Steps To Reproduce:

1.  Open VSCode with the Salesforce extensions.
1.  Run the command SFDX: Create Project from the command palate
1.  Select Standard (probably does not matter)
1.  Enter a Project Name with a trailing space
1.  Provide a directory to save
1. Try to rename or delete it in file explorer and you will get  "could not find this item" errors
### Expected result

It should trim the trailing white spaces

### Actual result

It does not trim the trailing white spaces

### Additional information

image

image

**SFDX CLI Version**:
sfdx-cli/7.74.1-32db2396ed win32-x64 node-v12.18.3

**OS and version**:
Windows 10

klewis-sfdc:
fix #2605

### What does this PR do?
Leading and trailing white spaces will now get stripped when using the SFDX: Create Project command

#2605, @W-8190305@

### Functionality Before
The SFDX: Create Project command allows leading and trailing whitespaces in the project name.

### Functionality After
The SFDX: Create Project command strips leading and trailing whitespaces from the project name.


packages/salesforcedx-vscode-core/src/util/inputUtils.ts
```diff
@@ -0,0 +1,8 @@
+import * as vscode from 'vscode';
+
+export async function getTrimmedString(projectNameInputOptions: vscode.InputBoxOptions) {
+  const input = await vscode.window.showInputBox(
+    projectNameInputOptions
+  );
+  return input !== undefined ? input.trim() : input;
```
jeffb-sfdc, line 7: This can be reduced to just:
```
  return input ? input.trim() : undefined;
  ```

File: packages/salesforcedx-vscode-core/src/util/inputUtils.ts [hunk 2]
```diff
@@ -0,0 +1,8 @@
+import * as vscode from 'vscode';
```
jeffb-sfdc, line 1: Please add the copyright here.  Just copy-paste from one of the other files, and make sure the year is 2022 since this is a new file.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -123,6 +125,16 @@ describe('Force Project Create', () => {
         expect.fail('Response should be of type ContinueResponse');
       }
     });
+
+    it('Should return Continue with trimmed project name if project name input has leading and or trailing spaces', async () => {
+      const gatherer = new SelectProjectName();
+      const response = await gatherer.gather();
+      if (response.type === 'CONTINUE') {
```
gbockus-sf, line 132: nit: I'd let expect do it's job and go with 
```
expect(response.type).to.equal('CONTINUE');
expect(response.data.projectName).to.equal(PROJECT_NAME);
```
 and do away with the condition.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -93,6 +94,7 @@ describe('Force Project Create', () => {
       inputBoxSpy.onCall(0).returns(undefined);
       inputBoxSpy.onCall(1).returns('');
       inputBoxSpy.onCall(2).returns(PROJECT_NAME);
+      inputBoxSpy.onCall(3).returns(PROJECT_NAME_WITH_LEADING_TRAILING_SPACES);
```
gbockus-sf, line 97: my apologies as I know you're following the pattern..but having a single before/after here and depending on state in each of the tests leads towards us having a bad time.  Would you mind changing this to a beforeEach/after Each and just assign the returns for each test in each it?   So it'd be something like 
```
beforeEach(() => { 
});

afterEach(() => {
  inputBoxSpy.restore();
});

....

    it('Should return Continue with trimmed project name if project name input has leading and or trailing spaces', async () => {
      inputBoxSpy.returns(PROJECT_NAME_WITH_LEADING_TRAILING_SPACES);
      const gatherer = new SelectProjectName();
...
```

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -33,6 +33,7 @@ import { getRootWorkspacePath } from '../../../src/util';
 // tslint:disable:no-unused-expression
 describe('Force Project Create', () => {
   const PROJECT_NAME = 'sfdx-simple';
+  const PROJECT_NAME_WITH_LEADING_TRAILING_SPACES = '  sfdx-simple  ';
```
gbockus-sf, line 36: nit: you could use a template literal here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

 =  \` ${PROJECT_NAME}  \` 

here

packages/salesforcedx-vscode-core/src/util/inputUtils.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export async function getTrimmedString(projectNameInputOptions: vscode.InputBoxOptions) {
```
gbockus-sf, line 3: nice:  want to write some sweet unit tests for this?

gbockus-sf (commented):
good stuff 👍   a few suggestions in there

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: Just took a crack at this.  Let me know how things look!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -86,41 +87,51 @@ describe('Force Project Create', () => {
   });
 
   describe('SelectProjectName Gatherer', () => {
-    let inputBoxSpy: sinon.SinonStub;
+    let gatherer: SelectProjectName;
+    let inputBox: sinon.SinonStub;
```
gbockus-sf, line 91: nit: I'd name this inputBoxStub

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -86,41 +87,51 @@ describe('Force Project Create', () => {
...
       expect(response.type).to.equal('CANCEL');
     });
 
     it('Should return Continue with inputted project name if project name is not undefined or empty', async () => {
-      const gatherer = new SelectProjectName();
+      inputBox.returns(PROJECT_NAME);
       const response = await gatherer.gather();
-      expect(inputBoxSpy.calledThrice).to.be.true;
+      expect(response.type).to.equal('CONTINUE');
+      if (response.type === 'CONTINUE') {
```
gbockus-sf, line 124: this if isn't necessary with the above expect.  The test will fail and bail if type doesn't equal Continue.

gbockus-sf: same on line 120

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -86,41 +87,51 @@ describe('Force Project Create', () => {
...
+        expect(response.data.projectName).to.equal(PROJECT_NAME);
+      }
+    });
+
+    it('Should return Continue with trimmed project name if project name input has leading and or trailing spaces', async () => {
+      inputBox.returns(PROJECT_NAME_WITH_LEADING_TRAILING_SPACES);
```
gbockus-sf, line 130: sweeeeeeeeeet much clearer how this test works 👍

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(() => {
+    sandbox = createSandbox();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should call showInputBox once', async () => {
+    const inputBox = sandbox.stub(vscode.window, 'showInputBox').resolves(INPUT_VAL);
```
gbockus-sf, line 26: you should follow the same pattern here that you do in the other test.  Create the stub in the before each.

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { expect } from 'chai';
+import { createSandbox, SinonSandbox } from 'sinon';
+import * as vscode from 'vscode';
+import { getTrimmedString } from '../../../src/util/inputUtils';
+
+describe('getTrimmedString', () => {
+  let sandbox: SinonSandbox;
+  const INPUT_VAL = 'Test Input';
+
+  beforeEach(() => {
```
gbockus-sf, line 17: this can be before.  Note a beforeEach should be separately to create the stub

gbockus-sf: To be clear you only need to create the sandbox once per suite.

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should call showInputBox once', async () => {
+    const inputBox = sandbox.stub(vscode.window, 'showInputBox').resolves(INPUT_VAL);
+    const trimmedString = await getTrimmedString({});
+    sandbox.assert.calledOnce(inputBox);
```
gbockus-sf, line 28: interesting.  I wonder if you have to use `sandbox.assert`. 🤷

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
gbockus-sf: oh also the same naming nit.  I usually use the thing I'm stubbing with a postfix of Stub: `showInputBoxStub`.

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
+import { createSandbox, SinonSandbox } from 'sinon';
+import * as vscode from 'vscode';
+import { getTrimmedString } from '../../../src/util/inputUtils';
+
+describe('getTrimmedString', () => {
```
gbockus-sf, line 13: nit: Can we have this be called `inputUtils Unit tests`?

gbockus-sf: could probably just add a top level describe with that name and keep it as is.  Will help when we add a new function to inputUtils and want to write tests for it.

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    sandbox.stub(vscode.window, 'showInputBox').resolves(`${INPUT_VAL}  `);
+    const trimmedString = await getTrimmedString({});
+    expect(trimmedString).to.be.eq(INPUT_VAL);
+  });
+
+  it('should remove leading and trailing whitespace', async () => {
+    sandbox.stub(vscode.window, 'showInputBox').resolves(`  ${INPUT_VAL}  `);
+    const trimmedString = await getTrimmedString({});
+    expect(trimmedString).to.be.eq(INPUT_VAL);
+  });
```
gbockus-sf, line 47: good good stuff.  Real deal unit tests..awesome.  

Should we consider the the following cases? 
- empty string input
- undefined input 
- null input
- just whitespace input

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: Removing the check for response type gives a compile-time error.  Do you know how I can get around it?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
jeffb-sfdc: +1

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: Resolved with Jeff's help (thanks @jeffb-sfdc!).  Was able to remove the conditional.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -86,42 +88,47 @@ describe('Force Project Create', () => {
...
-      inputBoxSpy.onCall(0).returns(undefined);
-      inputBoxSpy.onCall(1).returns('');
-      inputBoxSpy.onCall(2).returns(PROJECT_NAME);
+    beforeEach(() => {
+      gatherer = new SelectProjectName();
     });
 
     after(() => {
-      inputBoxSpy.restore();
+      inputBoxStub.restore();
```
jeffb-sfdc, line 99: I suggest:
1. move inputBoxStub into each `it()` which requires it
2. remove `inputBoxStub.restore()` from `after()` (and also remove `after()`)

At the moment, here's the lifecycle of inputBoxStub:
* The test starts 
* The instance of the stub is created (line 92)
* test#1 starts (line 102)
* inputBoxStub is set to return the value the test is setting it to (line 103)
* test#1 finishes
(inputBoxStub is not cleaned up)
* test#2 starts
* inputBoxStub is reused
* inputBoxStub is set to return the value the test is setting it to
* test#2 finishes
.
.
.
* test N finishes
* `after()` is called, and `inputBoxStub.restore()` executes


In one of your tests, if you add something like this:

```
    it('Should return cancel if user input is empty string', async () => {
      inputBoxStub.returns('');
      expect(response.type).to.equal('CANCEL');
      expect(inputBoxStub.calledOnce).to.be.true; // <<<<<<<<<<<<<<
    });
````

...this is the 3rd test, and I suspect that the "is called once" is going to fail.  For stubs like this, I think it is cleaner to do:

```
describe('SelectProjectName Gatherer', () => {
    let gatherer: SelectProjectName;

    beforeEach(() => {
      gatherer = new SelectProjectName();
    });


    it('Should make one call to showInputBox', async () => {
      const inputBoxStub = sinon.stub(vscode.window, 'showInputBox');
      inputBoxStub.returns(undefined);
      expect(inputBoxStub.calledOnce).to.be.true;
    });
.
.
.
```

Comment (klewis-sfdc) [reply packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts:99]: OK - when I try to do this I am getting:



Here is the attempted change:

Comment (klewis-sfdc) [reply packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts:99]: Do you know how I can do this and avoid the "already wrapped" error?

File: packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts [hunk 16]
```diff
@@ -0,0 +1,74 @@
...
+      expect(trimmedString).to.be.eq(EMPTY_STRING);
+    });
+
+    it('should return undefined when given undefined', async () => {
+      showInputBoxStub.resolves(undefined);
+      const trimmedString = await getTrimmedString({});
+      assert.isUndefined(trimmedString);
+    });
+
+    it('should return null when given null', async () => {
```
klewis-sfdc, line 76: Since getTrimmedString uses showInputBox internally, which only returns a string or undefined, this might not be a valid test.  Thoughts?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: Thanks for the help, Jeff! I think its right now.

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
jag-sfdc: @klewis-sfdc - Sorry for looking at this so late. Instead of getTrimmedString taking in InputBoxOptions, if it just takes a string (whatever vscode.window.showInputBox() returns) and trims it, would the tests and the test changes be simpler?

klewis-sfdc: Yes, it would definitely be simpler, but there was some discussion in the first few rounds of review on this PR about having a reusable service method that could be used anytime we wanted to get a trimmed string from the user to avoid having to do it _after_ collection of the value each time.  I'm ok either way and could update the PR to just trim before returning from SelectProjectName (and removing getTrimmedString) if you think that is better.  

Alternatively:

klewis-sfdc: It does seem a bit weird passing input options to the getTrimmedString function, but if we switch to passing it a String, then probably better to just call string.trim() locally here.

jag-sfdc: Agree, just calling string.trim() would be my preference as well. I see the point about a reusable function - but this is such a simple cleanup on the string vs a validation for function name or apex class name.

gbockus-sf (approved):
🍠  Looks good.  Don't forget to wait for your QE to approve before hitting the button 👍

packages/salesforcedx-vscode-core/test/vscode-integration/util/inputUtils.test.ts
gbockus-sf: We chatted in slack to coordinate. Ended up keeping the utility method as it was already mostly there.

floralan:
Since this is a bug fix, PR title should start with **fix:** so changelog generator will categorize this into fix column in changelog. I added Sonal as the reviewer so we need wait her approval on the title to get this PR merged.

floralan:
Used vsixes and local build to verify this PR:

✅ Verified that SFDX:Create Project command allows leading and trailing space in extensions v54.7.0
✅ Verified that, with the changes in this PR, SFDX:Create Project command automatically delete the leading and trailing space for project name.

PR #3951, opened 2022-03-24
[LWC Intellisense] Enhance integration tests in salesforce-vscode-lwc package

RitamAgrawal:
### What does this PR do?
Add integration tests in intellisense.test.ts in lwc-package.
@W-10759603@

### Functionality Before
Editor was not getting cleared after each tests.

### Functionality After
1. Added `LWC JS @Salesforce Import Intellisense Test` and `LWC JS Lightning Import Test`
2. Fixed existing tests to have a clean editor after each tests.


packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
+        'force-app',
+        'main',
+        'default',
+        'lwc'
       );
-    } catch (e) {
-      throw e;
-    }
-  });
+      lwcExtension = extensions.getExtension(
```
gbockus-sf, line 50: Should this and the next line go in a before function for the top level describe?   It should only have to verify it's activated once.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-    await editor.edit(editBuilder => {
-      editBuilder.replace(rangeReplace, text);
+    afterEach(() => {
+      try {
+        selection = new Selection(startPosition, endPosition);
+        // We need to clear the editor's line or the input text will change and test will fail.
+        editor.edit(editBuilder => {
+          editBuilder.delete(selection)
+        });
+      } catch (e) {
```
gbockus-sf, line 69: Do we have to try/catch here for some reason?  Seems it'd be the same result (a failed afterEach) without it.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-        items: [
-          {
-            label: 'lightning-accordion',
-            kind: vscode.CompletionItemKind.Property
-          },
-          { label: 'c-hello-binding', kind: vscode.CompletionItemKind.Property }
-        ]
+    it('LWC Extension Activation', async () => {
+      expect(lwcExtension.isActive);
+    });
```
gbockus-sf, line 76: This test can go at the top level describe. Esp if we move the activate wait to the top level before.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-      startPosition.character + text.length
-    );
-    const rangeReplace = new vscode.Range(startPosition, endPosition);
-    await editor.edit(editBuilder => {
-      editBuilder.replace(rangeReplace, text);
+      try {
+        await testCompletion(docUri, endPosition, items);
+      } catch (error) {
+        throw error;
+      }
```
gbockus-sf, line 103: another spot where I don't think we need the try/catch

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
+          label: '@commitlint/cli',
+          kind: CompletionItemKind.Module
+        }
+      ]
+
+      try {
+        await testCompletion(docUri, endPosition, items);
+      } catch (error) {
+        throw error;
+      }
```
gbockus-sf, line 136: try/catch begone

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-  docUri: vscode.Uri,
-  position: vscode.Position,
-  expectedCompletionList: vscode.CompletionList
+  docUri: Uri,
+  position: Position,
+  expectedCompletionList: CompletionItem[]
 ) {
   // Simulate triggering a completion
-  const actualCompletionList = (await vscode.commands.executeCommand(
+  let actualCompletionList = (((await commands.executeCommand(
```
gbockus-sf, line 258: this should be const right?

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
     position
-  )) as vscode.CompletionList;
+  )) as CompletionList).items);
+
+  actualCompletionList.sort();
+  expectedCompletionList.sort();
 
-  expectedCompletionList.items.forEach(expectedItem => {
-    const actualItem = actualCompletionList.items.find(obj => {
+  expectedCompletionList.map(expectedItem => {
```
gbockus-sf, line 267: Probably shouldn't use map here b/c we don't need to create a new altered array. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#when_not_to_use_map

I'd go with for...of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

gbockus-sf (commented):
Had a few comments in there....good stuff 👍

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-            kind: vscode.CompletionItemKind.Module
-          },
-          {
-            label: '@salesforce/user/Id',
-            kind: vscode.CompletionItemKind.Module
-          }
-        ]
+
+    it('LWC JS Lightning Import Intellisense', async () => {
+      // We have to have some text or we'll just get generic completions
```
floralan, line 140: Can we remove the comment here since it's duplicate with the previous test.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
RitamAgrawal: oh yes! that's a miss. I will change to const.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
-      await testCompletion(docUri, endPosition, {
-        items: [
-          {
-            label: 'c/helloBinding',
-            kind: vscode.CompletionItemKind.Folder
-          }
-        ]
+    it('LWC JS @Salesforce Import Intellisense', async () => {
+      // We have to have some text or we'll just get generic completions
+      text = "import {} from '@sales";
```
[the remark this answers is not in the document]
RitamAgrawal: We are using double quotes because we have a single quote(') character before _@sales_ within the string.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -8,217 +8,264 @@
...
+    let text: string;
+    let startPosition: Position;
+    let endPosition: Position;
+    let selection: Selection;
 
-  afterEach(async () => {
-    try {
-      await vscode.commands.executeCommand(
-        'workbench.action.closeActiveEditor'
+    beforeEach(async() => {
```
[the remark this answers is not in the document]
RitamAgrawal: I ran `npm run lint`. Do we have any other command as well?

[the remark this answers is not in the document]
gbockus-sf: `npm run reformat` I think it's also suppose to happen automagically before a push but I've seen a few instances where folks have gotten around it somehow

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
RitamAgrawal: Yes! It would be the same result: failed afterEach. So have removed it from both instances.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -7,218 +7,230 @@
...
-      startPosition.character + text.length
-    );
-    const rangeReplace = new vscode.Range(startPosition, endPosition);
-    await editor.edit(editBuilder => {
-      editBuilder.replace(rangeReplace, text);
+  describe('LWC JS Intellisense Test Suite', function() {
+
+    // Time taken to execute the command to fetch actualcompletion list, varies with different environment and system.
+    // tslint:disable-next-line:no-invalid-this
+    this.timeout(10000);
```
gbockus-sf, line 45: just curious why you bumped this from 20s -> 10s

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -7,218 +7,230 @@
...
+  extensions,
+  Extension,
+  Position,
+  Selection,
+  TextDocument,
+  TextEditor,
+  Uri,
+  window,
+  workspace
+} from 'vscode';
```
gbockus-sf, line 24: interesting...why'd you decide to go away from just the * import?

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
```diff
@@ -7,218 +7,230 @@
...
+
+    beforeEach(async () => {
+      lwcDir = path.join(
+        workspace.workspaceFolders![0].uri.fsPath,
+        'force-app',
+        'main',
+        'default',
+        'lwc'
+      );
+      docUri = Uri.file(path.join(lwcDir, 'hello', 'hello.js'));
```
gbockus-sf, line 63: just observation that the above two lines don't ever change so they could have been assigned and been const's outside the `beforeEach`

gbockus-sf (approved):
Good stuff.  Def way better then when you started 👍

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
RitamAgrawal: Looked at the execution time of each tests in circleCI and that's all less than 10seconds. I think 5000 can also suffice but just having the buffer.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
RitamAgrawal: In order to avoid writing `vscode.` everywhere where these keywords(like Uri, window etc..) are used.

packages/salesforcedx-vscode-lwc/test/vscode-integration/intellisense.test.ts
RitamAgrawal: Yes! Thank you. Now, since we are dealing with JS and Html separately, `doc` and `docUri` can be treated as consts.

Issue #1579, opened 2019-08-22
Retrieve Standard Value Sets in Org Browser 


FabienTaillon:
### Summary

No Standard Value Set visible in Org Browser.

### Steps To Reproduce:

1.  Open Org Browser.
1.  Click on Standard Value Sets.

### Expected result

The list of Standard Value Sets should be displayed.

### Actual result

Just displaying "No components available".

### Additional information

**VS Code Version**: 1.37.1 + Extension pack 46.10

**SFDX CLI Version**: sfdx-cli 7.20.1 (core)

**OS and version**: MacOS Mojave

FabienTaillon:
Actually there are several metadata where it does not work, so I'm sure you already know about it.

aldoforce:
any update around this? I'm facing the same problem. Can't retrieve Case status fields.

aldoforce:
@FabienTaillon even if you don't use OrgBrowser, the issue is the same with CLI:

```
sfdx force:source:retrieve -m StandardValueSet

=== Retrieved Source
No results found

```

FabienTaillon:
If you know which StandardValueSet, you can specify then name directly:
`sfdx force:source:retrieve -m StandardValueSet:CaseStatus`

You can't retrieve all at once with the command because this Metadata Type doesn’t support the wildcard character * (asterisk).

However, regarding the VS Code extension, I would expect to see the list anyway because from a user perspective we don't care that the API is not good. I would even prefer an hardcoded list that would be updated at every release than seeing nothing.

Obviously the best fix would be to fix the API itself.

AltiusRupert:
You can stick this file into your manifest folder, then click-right and "SFDX: Retireve Source in Manifest from Org"


[package-StandardValueSet.xml.zip](https://github.com/forcedotcom/salesforcedx-vscode/files/4635848/package-StandardValueSet.xml.zip)

MooseAteJack:
Also having this issue.

There are ways to work around the issue and force:source:pull is fine with seeing change in StandardValueSets but  the browser can't find them.

kennethkim1221:
Hi all. I am experiencing the same thing now. I tried Fabien's suggestion in using sfdx force:source:retrieve -m StandardValueSet:LeadSource but the result was:

=== Retrieved Source
No results found

Anyone figured out how to to this?

FabienTaillon:
[...]

PR #3982, opened 2022-03-29
feat(orgbrowser): add support for StandardValueSet

sbresin:
### What does this PR do?
Adds support for retrieving StandardValueSet Metadata with the Org Browser.

There is no API call that lists all StandardValueSets, so i hardcoded the list of their names. The list comes from here: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/standardvalueset_names.htm

#1579 

### Functionality Before
None: Currently it is not possible at all to retrieve StandardValueSets from the OrgBrowser.

### Functionality After
Screen Shot 2022-03-29 at 16 38 46


packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -21,6 +21,89 @@ const validManageableStates = new Set([
   undefined // not part of a package
 ]);
 
+const STANDARDVALUESET_FULLNAME = 'StandardValueSet';
```
mohanraj-r, line 24: Can we reuse `standardValueSet.fullnames` from `source-deploy-retrieve` (`import { standardValueSet } from '@salesforce/source-deploy-retrieve/lib/src/registry'`) here instead

mohanraj-r (changes requested):
Thanks for the contribution @sbresin 👍 

@RitamAgrawal Can you please take a look at this PR since you made changes to org browser recently

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
sbresin: addressed.

thanks for checking it out this quick :)

mohanraj-r:
Superseded by https://github.com/forcedotcom/salesforcedx-vscode/pull/3992

PR #3992, opened 2022-03-31
feat(orgbrowser): add support for StandardValueSet

mohanraj-r:
### What does this PR do?
Builds the community contribution #3982 from @sbresin 🎉 

Adds support for retrieving StandardValueSet Metadata with the Org Browser.

There is no API call that lists all StandardValueSets, so i hardcoded the list of their names. The list comes from here: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/standardvalueset_names.htm

#1579 

### Functionality Before
None: Currently it is not possible at all to retrieve StandardValueSets from the OrgBrowser.

### Functionality After
Screen Shot 2022-03-29 at 16 38 46


klewis-sfdc (approved):
Looks great to me!  Thanks for the contribution @sbresin!

:white_check_mark: Org Browser Retrieve Source works for StandardValueSets
:white_check_mark: Org Browser Retrieve Source and Open works for StandardValueSets
:white_check_mark: Org Browser Retrieve Source for all Standard Value Sets via directory-level retrieve button works
:white_check_mark: Adding a new picklist option to a standard value set and then re-running Retrieve Source via Org Browser to retrieve the new value works (also works via right click on file in Explorer)
:white_check_mark: Adding a new picklist option to a standard value set and then running SFDX: Diff file against org works and shows correct diff

Issue #3693, opened 2021-12-10
not stopping on breakpoint for javascript function


jjsfdc:
### Summary

"Debug Invoke" of my sample JSON payload invokes locally running JS Function, but doesn't stop on my breakpoint.

### Steps To Reproduce:

1.  sf generate project -n SomeProject
2. cd SomeProject
3. sf generate function -l javascript -n somefn
4. code .
5. edit index.js so function body is just one line: return "hello world";
6. click in left edge of editor to set a breakpoint on this return line
7. right click and choose "SFDX: Start Function"
8. upon launch, create a json file within the "functions/somefn" directory and save
9. click "debug invoke" link at top of editor page

### Expected result

debugger would pause on the breakpoint I set

### Actual result

function is invoked, but execution does not stop on the breakpoint.

### Additional information

**Salesforce Extension Version in VS Code**: 53.7.0

**SFDX CLI Version**: node-v16.13.1

**OS and version**: Mac OS Big Sur version 11.6.1

randi274:
I don't believe this was in the scope of the original add-on for functions, but is something we should consider adding going forward.

gbockus-sf:
Associated Work Item: [W-10738230](https://gus.lightning.force.com/a07EE00000rEzczYAC)

PR #4001, opened 2022-04-05
Add support for debugging JavaScript with containerless functions

jeffb-sfdc:
### What does this PR do?
This PR adds support for debugging JavaScript with containerless functions, by adding a new property to `FunctionExecution`, `isContainerless`.

#@W-10738230@

### Functionality Before
When one attempted to debug a JavaScript file using a containerless function, the code would be invoked, but one was not able to debug.

### Functionality After
One can now debug a JavaScript file using a containerless function.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
@@ -75,7 +75,7 @@ export abstract class ForceFunctionStartExecutor extends LibraryCommandletExecut
         rootDir: functionDirPath,
         port: FUNCTION_DEFAULT_PORT,
         debugPort: FUNCTION_DEFAULT_DEBUG_PORT,
-        // Note this defaults to node but will be updated by the updateFunction method after the function is started if necessarly.
+        // Note this defaults to node but will be updated by the updateFunction method after the function is started if necessary.
         debugType: 'node',
         terminate: () => {
           return new Promise(resolve =>
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
@@ -81,7 +81,8 @@ export abstract class ForceFunctionStartExecutor extends LibraryCommandletExecut
           return new Promise(resolve =>
             resolve(this.cancelFunction(registeredStartedFunctionDisposable))
           );
-        }
+        },
+        isContainerless: false
       }
     );
 
--- a/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
@@ -82,7 +82,7 @@ export abstract class ForceFunctionStartExecutor extends LibraryCommandletExecut
             resolve(this.cancelFunction(registeredStartedFunctionDisposable))
           );
         },
-        isContainerless: false
+        isContainerLess: false
       }
     );
 
--- a/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
@@ -10,6 +10,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { Disposable } from 'vscode';
 import { workspaceContext } from '../../context';
+import { nls } from '../../messages';
 import { getRootWorkspace, getRootWorkspacePath } from '../../util';
 
 /**
@@ -57,9 +58,6 @@ export interface FunctionExecution extends Terminable {
    */
   isContainerLess: boolean;
 }
-
-export const FUNCTION_TYPE_ERROR =
-  'Unable to determine type of executing function.';
 
 export class FunctionService {
   private static _instance: FunctionService;
@@ -161,7 +159,8 @@ export class FunctionService {
 
       return functionType.JAVA;
     }
-    throw new Error(FUNCTION_TYPE_ERROR);
+
+    throw new Error(nls.localize('error_function_type'));
   }
 
   /**
@@ -188,8 +187,7 @@ export class FunctionService {
   public async debugFunction(rootDir: string) {
     const functionExecution = this.getStartedFunction(rootDir);
     if (!functionExecution) {
-      // TODO: report error?
-      return;
+      throw new Error(nls.localize('error_unable_to_get_started_function').replace('{0}', rootDir));
     }
 
     if (!functionExecution.debugSession) {
@@ -221,7 +219,7 @@ export class FunctionService {
     };
 
     if (functionExecution.isContainerLess) {
-      debugConfiguration.remoteRoot = undefined;
+      delete debugConfiguration.remoteRoot;
     }
 
     return debugConfiguration;
--- a/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
@@ -55,7 +55,7 @@ export interface FunctionExecution extends Terminable {
   /**
    * Flag to determine whether running in a container
    */
-  isContainerless: boolean;
+  isContainerLess: boolean;
 }
 
 export const FUNCTION_TYPE_ERROR =
@@ -114,7 +114,7 @@ export class FunctionService {
     };
   }
 
-  public updateFunction(rootDir: string, debugType: string, isContainerless: boolean): void {
+  public updateFunction(rootDir: string, debugType: string, isContainerLess: boolean): void {
     const functionExecution = this.getStartedFunction(rootDir);
     if (functionExecution) {
       const type = debugType.toLowerCase();
@@ -124,7 +124,7 @@ export class FunctionService {
         functionExecution.debugType = 'java';
       }
 
-      functionExecution.isContainerless = isContainerless;
+      functionExecution.isContainerLess = isContainerLess;
     }
   }
 
@@ -220,7 +220,11 @@ export class FunctionService {
       port: debugPort
     };
 
-    if (this.getFunctionType() === functionType.JAVASCRIPT && functionExecution.isContainerless) {
+    // The following check is specifically for when debugging a JavaScript function and not using
+    // a container (see W-10738230).  If there are issues with debugging TypeScript in the future,
+    // this is something to look into for TypeScript as well (though, at this time, debugging
+    // TypeScript w/o a container works).
+    if (functionExecution.isContainerLess && this.getFunctionType() === functionType.JAVASCRIPT) {
       debugConfiguration.remoteRoot = undefined;
     }
 
--- a/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
@@ -60,10 +60,6 @@ export interface FunctionExecution extends Terminable {
 
 export const FUNCTION_TYPE_ERROR =
   'Unable to determine type of executing function.';
-// localize?  This appears in the output tab  (but not a notification)
-
-
-
 
 export class FunctionService {
   private static _instance: FunctionService;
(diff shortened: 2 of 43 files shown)


packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -180,27 +187,44 @@ export class FunctionService {
...
-        port: debugPort
-      };
-      if (!functionExecution.debugSession) {
-        await vscode.debug.startDebugging(
-          getRootWorkspace(),
-          debugConfiguration
-        );
-      }
+    if (!functionExecution) {
+      // TODO: report error?
```
jeffb-sfdc, line 191: Should we report an error to the user?  Either in the output or using a notification pill?

jeffb-sfdc, line 191: Should we report an error here?  And the sting on line 62, should it be localized?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -180,27 +187,44 @@ export class FunctionService {
...
+      resolveSourceMapLocations: ['**', '!**/node_modules/**'],
+      console: 'integratedTerminal',
+      internalConsoleOptions: 'openOnSessionStart',
+      localRoot: rootDir,
+      remoteRoot: '/workspace',
+      hostName: '127.0.0.1',
+      port: debugPort
+    };
+
+    if (this.getFunctionType() === functionType.JAVASCRIPT && functionExecution.isContainerless) {
```
jag-sfdc, line 223: Interesting! How did you find out this was the change that was required? How is that typescript in containerless mode works even without this fix?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -52,6 +52,10 @@ export interface FunctionExecution extends Terminable {
    * Active debug session attached
    */
   debugSession?: vscode.DebugSession;
+  /**
+   * Flag to determine whether running in a container
```
jag-sfdc, line 56: Nit pick - Can we add a "why" here. i.e. we need this flag in invoke/debug which is a separate operation and we won't have context as to which function was started.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: Is this the case where the function was not started and the user is trying to debug it?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: Never mind. Let's keep it this way. I think the "why" part is quite obvious.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: If so, I see the behavior is the same - the function invoke fails and an error in the output. Let's keep it this way.

jag-sfdc: Yes. Let's localize that message. I think this is something we missed.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: Jeff explained this in the elevensies today.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jag-sfdc: Except for this, everything looks good. I'll approve the PR now and you can merge after localizing this message and QA.

jag-sfdc (approved):

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -180,27 +187,48 @@ export class FunctionService {
...
+      remoteRoot: '/workspace',
+      hostName: '127.0.0.1',
+      port: debugPort
+    };
+
+    // The following check is specifically for when debugging a JavaScript function and not using
+    // a container (see W-10738230).  If there are issues with debugging TypeScript in the future,
+    // this is something to look into for TypeScript as well (though, at this time, debugging
+    // TypeScript w/o a container works).
+    if (functionExecution.isContainerLess && this.getFunctionType() === functionType.JAVASCRIPT) {
```
gbockus-sf, line 227: So this is a nit consideration, but should we get rid of the function type check.  If it works for TS/java/js without remoteRoot that seems more clear then having it for 2 out of 3 cases even if it works today as is.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -180,27 +187,48 @@ export class FunctionService {
...
+      debugConfiguration.remoteRoot = undefined;
```
gbockus-sf, line 228: Nit (as you already verified it works) but here would have been cleaner to delete the property instead of setting it to undefined. `delete debugConfiguration.remoteRoot`.  The diff being with the current flow we get 
```
{
      ...
      localRoot: rootDir,
      remoteRoot: undefined,
      ...
} 
```
where as if we use delete the property doesn't exist on the object at all.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
```diff
@@ -327,4 +338,130 @@ describe('Function Service', () => {
...
+      fsSyncStub.returns(true);
+      const service = FunctionService.instance;
+      service.registerStartedFunction(functionDef);
+      const rootDir = 'FirstFunction';
+      const functionExecution = service.getStartedFunction(rootDir);
+      const debugConfiguration = service.getDebugConfiguration(functionExecution!, rootDir);
+      expect(debugConfiguration.remoteRoot).to.equal('/workspace');
+    });
+
+    it('Should set remoteRoot to /workspace when TypeScript and containerless.', () => {
```
gbockus-sf, line 408: This might be a good reason to remove the remoteRoot on local debugging.  This test  and the next one tells us we expect this setting for locally running TS/java functions, but we don't really need it so it could lead to future confusion.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
```diff
@@ -327,4 +338,130 @@ describe('Function Service', () => {
...
+        isContainerLess: false
+      };
+
+      fsSyncStub.onCall(0).returns(false);
+      fsSyncStub.onCall(1).returns(true);
+      const service = FunctionService.instance;
+      service.registerStartedFunction(functionDef);
+      const rootDir = 'FirstFunction';
+      const functionExecution = service.getStartedFunction(rootDir);
+      const debugConfiguration = service.getDebugConfiguration(functionExecution!, rootDir);
```
gbockus-sf, line 365: oh I had to look it up...that trailing ! is handy

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
```diff
@@ -327,4 +338,130 @@ describe('Function Service', () => {
...
+        rootDir: 'FirstFunction',
+        debugPort: 7777,
+        port: 8080,
+        debugType: 'unknown',
+        terminate: () => Promise.resolve(),
+        isContainerLess: false
+      };
+
+      fsSyncStub.onCall(0).returns(false);
+      fsSyncStub.onCall(1).returns(true);
```
gbockus-sf, line 360: unit test approach question.  Why not stub `getFunctionType` in these tests to make it a bit more clear how we're getting JS vs TS vs Java function in the function under test?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
gbockus-sf: I think the user does currently get feedback in this situation.  Might be worth replacing the TODO with comment to info future us :)

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
```diff
@@ -81,7 +81,8 @@ export abstract class ForceFunctionStartExecutor extends LibraryCommandletExecut
           return new Promise(resolve =>
             resolve(this.cancelFunction(registeredStartedFunctionDisposable))
           );
-        }
+        },
+        isContainerLess: false
```
gbockus-sf, line 85: Can you update the unit tests for the three functionStartExecutor classes?

gbockus-sf (commented):
Some nits to consider.  Only bonafide is to update the startExecutor unit tests I think.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -180,27 +187,44 @@ export class FunctionService {
...
+      resolveSourceMapLocations: ['**', '!**/node_modules/**'],
+      console: 'integratedTerminal',
+      internalConsoleOptions: 'openOnSessionStart',
+      localRoot: rootDir,
+      remoteRoot: '/workspace',
+      hostName: '127.0.0.1',
+      port: debugPort
+    };
+
+    if (functionExecution.isContainerLess) {
```
jeffb-sfdc, line 223: @jag-sfdc I had added a comment, but then @gbockus-sf left a comment asking, "why even check the language type?"  I was checking the language type to leave the TS and Java paths as they were, but I tried out using `debugConfiguration.remoteRoot = undefined` for all three, and they all still worked, so I removed the language check and removed the code comments as well.

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jeffb-sfdc: done

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jeffb-sfdc: done

jeffb-sfdc: done

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
jeffb-sfdc: done

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
jeffb-sfdc: That was the existing pattern in the file.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
jeffb-sfdc: I looked but didn't see any places but didn't find any code (in out tests files) where they needed to be updated.  If there's something specific you have in mind, ping me & lets go over what needs to be updated.

floralan:
Used vsix files to verify this fix. 
✅ Verified, before this fix, debugger never hit the break point in javascript file when running "Debug Invoke" for a js function. 
✅ With this fix, run "Debug Invoke" in a javascript file, verified that break point works.
✅ Verified "Debug Invoke" does invoke function after the debugging finished 
✅ verified break point also works for Typescript Functions. 

A question:
after I edit the javascript file by adding more functions to it, I saved the file, and set break points inside the file. Then I run "Debug Invoke" again, it opens the old js file without my changes. Is there a further step required to update the index.js file or is this the behavior we don't expect?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
jeffb-sfdc: done

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
jeffb-sfdc: done

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts
jeffb-sfdc: I updated the tests (the containerless ones, for all 3 languages) to check for the prop to not be defined, and I also updated the titles of the tests.

gbockus-sf:
@floralan Yeah the behaviour you called out is the expected behaviour. To pick up the changes you'd need to stop/start the function again.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.test.ts
```diff
@@ -513,7 +514,7 @@ describe('ForceFunctionContainerStartExecutor unit tests', () => {
       assert.calledWith(startFunctionStub, fakeId, fakeResponse.data);
     };
 
-    it('Should be able to run a container executor.', async () => {
+    it.only('Should be able to run a container executor.', async () => {
```
gbockus-sf, line 517: rot roh: be sure to remove the only.

jeffb-sfdc: Thanks!  (and OOPS!) I just fixed this.

randi274:
Addressed #3693

PR #4003, opened 2022-04-05
fix: hover error fixed in LSP

randi274:
### What does this PR do?
Consumes the latest lightning-language-server packages, which fixes the textDocument/hover error in
the event that the lsp is incorrectly parsing the content of .js files.

#3929, @W-10857157@

### Functionality Before
Users were often experiencing the textDocument/hover error, which drew focus to the output panel and caused excessive noise.

### Functionality After
This error is now suppressed, and users can continue developing without being constantly pestered with unhelpful errors 😌

### DIFF
--- a/packages/salesforcedx-vscode-lightning/package.json
+++ b/packages/salesforcedx-vscode-lightning/package.json
@@ -25,9 +25,9 @@
     "Programming Languages"
   ],
   "dependencies": {
-    "@salesforce/aura-language-server": "3.7.0",
+    "@salesforce/aura-language-server": "3.8.0",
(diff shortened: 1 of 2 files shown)


floralan (approved):
I'm not seeing the hover error with this change! Great job for fixing this bug.

gbockus-sf (approved):
Using the ebike project I just edited some existing html to QE this issue. Used the currently released version (v54.6.1) to verify the error condition. Used the vsixes from the last build to verify this fixed.  

✅  Verified that I got the error when editing HTML in LWC prior to the update
✅  Verified that the error no longer occurred after update

PR #4005, opened 2022-04-05
chore: merge release branch into develop once it is cut

floralan:
### What does this PR do?
Bring back release branch into develop as soon as it is cut.

 @W-10943603@

### DIFF
--- a/scripts/create-release-branch.js
+++ b/scripts/create-release-branch.js
@@ -92,7 +92,7 @@ shell.exec('git add lerna.json');
 // Git commit
 shell.exec(`git commit -m "chore: update to version ${nextVersion}"`);
 
-// Merge release branch to develop
+// Merge release branch to develop as soon as it is cut to resolve conflict when merge main branch into develop after release 
 shell.exec(`git checkout develop`)
 shell.exec(`git merge ${releaseBranchName}`)
 shell.exec(`git push -u origin develop`)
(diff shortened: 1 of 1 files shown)


scripts/create-release-branch.js
```diff
@@ -102,5 +92,21 @@ shell.exec('git add lerna.json');
...
+
+// Generate changelog
+const previousBranchName = changeLogGeneratorUtils.getPreviousReleaseBranch(releaseBranchName);
+const parsedCommits = changeLogGeneratorUtils.parseCommits(changeLogGeneratorUtils.getCommits(releaseBranchName, previousBranchName));
+const groupedMessages = changeLogGeneratorUtils.getMessagesGroupedByPackage(parsedCommits, '');
+const changeLog = changeLogGeneratorUtils.getChangeLogText(releaseBranchName, groupedMessages);
+changeLogGeneratorUtils.writeChangeLog(changeLog);
+
+const commitCommand = `git commit -a -m "chore: generated CHANGELOG for ${releaseBranchName}"`;
+shell.exec(commitCommand);
```
gbockus-sf, line 109: Do we want to merge into develop before we commit the change log?  I fee like that kinda was the point, but hurts my head a little :)

gbockus-sf: I went back and looked at the ticket and it does indeed say to do exactly what you're doing. I do feel like we need a code comment here to call out why we do that b/c if I were to revisit this code in 6 months I'll def be tempted to change the order.

floralan: @gbockus-sf  Added some comment here.

floralan: @gbockus-sf  Added some comment here.

gbockus-sf (approved):
cool

PR #4020, opened 2022-04-11
fix: SFDX:Rename Component now also rename the test file that has the same name as component under __tests__ folder for LWC components

floralan:
### What does this PR do?
Enable SFDX: Rename Component command to rename test file under __tests__ folder for LWC components. Only the test file that has the same name as component will be renamed. 

@W-10976932@

### Functionality Before
SFDX: Rename Component does not rename test file under __tests__ folder in LWC components. 

### Functionality After
SFDX: Rename Component now rename the test file in LWC components as well.

### DIFF
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
@@ -10,8 +10,10 @@ const lwcPath = vscode.Uri.parse('/force-app/main/default/lwc');
 const auraPath = vscode.Uri.parse('/force-app/main/default/aura/');
 const lwcComponent = 'hero';
 const auraComponent = 'page';
-const itemsInHero = ['_test_', 'hero.css', 'hero.html', 'hero.js', 'hero.js-meta.xml', 'templateOne.html'];
-const itemsInPage = ['_test_', 'page.auradoc', 'page.cmp', 'page.cmp-meta.xml', 'page.css', 'page.design', 'page.svg', 'pageController.js', 'pageHelper.js', 'pageRenderer.js', 'templateOne.css'];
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -39,7 +41,7 @@ describe('Force Rename Lightning Component', () => {
       readdirStub
         .onFirstCall().resolves([])
         .onSecondCall().resolves([])
-        .onThirdCall().resolves([itemsInHero[1]]);
+        .onThirdCall().resolves([itemsInHero[0]]);
```
gbockus-sf, line 44: did this change for any particular reason?

gbockus-sf: oh I see... moving the __tests__ got it

gbockus-sf (approved):
good stuff thanks for the quick turn around 👍

gbockus-sf:
QE:

gbockus-sf (approved):

floralan:
> 



> 
"isDuplicate" function only check for the existing LWC and Aura Components name aka the folder names. If we check every file name under LWC or Aura for duplicate, when the number of components is large, "rename component" command can be slow. But we can try to implement it and test for the execution time to see if it's acceptable.

gbockus-sf:
> ause that "isDuplicate" function only check for the existing LWC and Aura Components name aka the folder names. If we check every file name under LWC or Aura for duplicate, when the number of components is large, "rename component" command can be slow. But we can try to implement it and test for the execution time to see if it's acceptable.

@floralan I don't think it's a blocker for this PR.  We can discuss the issue and see if it makes sense to worry about it. It would be good to add your above comment to the issue I created.

floralan:
> 

@gbockus-sf  Sounds great! I'll get this PR merged and discuss the issue in the charter of the ticket.

Issue #3873, opened 2022-03-02
Stopping a containerless function does not kill the local process


gbockus-sf:
### Summary

If a containerless function is started, then stopped via the command palette the process started for the function service is not killed. 

### Steps To Reproduce:

1.  For any function type start the function in containerless mode `SFDX: Start Function`
1.  After the function has started run the stop function command `SFDX: Stop Function`

### Expected result

Command should run successfully and the underlying process for the function should be gone. 

### Actual result

Command run successfully but the process for the function remains. 

### Additional information

Blocked by :https://github.com/heroku/sf-functions-core/issues/42

gbockus-sf:
Verified this is still an issue.  Associated work Item W-10737950.  Closing W-10934563 as a dupe that was created when I added the bug label.

PR #4025, opened 2022-04-12
fix: add ability to stop containerless function

mohanraj-r:
### What does this PR do?
Fixes [Stopping a containerless function does not kill the local process · Issue #3873 · forcedotcom/salesforcedx-vscode](https://github.com/forcedotcom/salesforcedx-vscode/issues/3873)

#3873, @W-10737950@

### Functionality Before
Stop Command runs successfully but the process for the function continues running.

### Functionality After
Stop Command runs successfully and the underlying process for the function is not running.


gbockus-sf (commented):
Changes look good so far.

sbudhirajadoc:
@mohanraj-r Is there a public facing string that you'd like me to review? Thanks!

mohanraj-r:
@sbudhirajadoc There are no new user facing strings in this PR. Just wanted to double-check if any changes need to be made to user docs e.g https://developer.salesforce.com/tools/vscode/en/functions/overview

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -51,10 +56,8 @@ export class ForceFunctionContainerlessStartExecutor extends ForceFunctionStartE
 
     localRun
```
gbockus-sf, line 57: nit: should we await this now that it resolves with the info we need to proceed?

gbockus-sf (approved):
QE: 
- Verify TS function is killed on stop ✅ 
- Verify JS function is killed on stop ✅ 
- Verify Java function is killed on stop ✅ 
- Kills worked after invoke and debug ✅ 

Verified process killed two ones.  1. check for processes using ps (existing after start, gone after kill) 2. verified I was able to start another function (no port conflict).

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
mohanraj-r: Good point.- but changing it to to use await requires making the `startFunction` as async which makes it different from `ForceFunctionContainerStartExecutor` and that could lead to inconsistencies when calling them from `ForceFunctionStartExecutor` ?

gbockus-sf: hum...true.  I think it's the lessor evil to make the startFunction return a `Promise<void>` verse having the unhandled promise.  Then we can have the async in the containerless version and the container version can just return a resolved promise.

gbockus-sf: I'm curious if this will fix the problem where VSCode says the containerless function has started, but then if you try to invoke too quickly it's not actually started and the invoke fails.

mohanraj-r: 👍 Refactored both container and containerless start to be async.

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
```diff
@@ -19,22 +19,27 @@ import {
 import { ForceFunctionStartExecutor } from './ForceFunctionStartExecutor';
 
 export class ForceFunctionContainerlessStartExecutor extends ForceFunctionStartExecutor {
+  private process: LocalRunProcess | undefined | void;
```
gbockus-sf, line 22: nit: void should encompass undefined here so I think it could just be `LocalRunProcess | void`

gbockus-sf (approved):
Thanks @mohanraj-r Looks good.  I ran though the QE again locally and everything worked as expected. 👍

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerlessStartExecutor.ts
mohanraj-r: Removing void results in error at `this.process = await localRun.exec()`
```
(property) ForceFunctionContainerlessStartExecutor.process: LocalRunProcess | undefined
Type 'void | LocalRunProcess' is not assignable to type 'LocalRunProcess | undefined'.
  Type 'void' is not assignable to type 'LocalRunProcess | undefined'.ts(2322)
```

Removing undefined results in error at `private process: LocalRunProcess | void;`
```
(property) ForceFunctionContainerlessStartExecutor.process: void | LocalRunProcess
Property 'process' has no initializer and is not definitely assigned in the constructor.ts(2564)
```

And looks like void can't be assigned to e.g. `this.process = undefined;` -> `this.process = void;` results in error

gbockus-sf: word.  Thanks for trying.

Issue #4024, opened 2022-04-12
Rename component can silently replace files


gbockus-sf:
### Summary

If you rename a lightning/aura component and there is an existing file withe same name it will be replace without warning. 

### Steps To Reproduce:

1.  Create a LWC component called accountMap.
2. Create a LWC test for the component (default name is accountMap.test.js in __tests__.
3. Create a new test file called `accountMapOther.test.js`. Map the content different then accountMap.test.js.
4. Execute the Rename component command via the context menu on the LWC component.
5. Enter `accountMapOther` as the new name for the component. 
6. See that there is now only a single test file and any edits to the original accountMapOther.test.js file are gone.

### Expected result

There should be a warning that the change is destructive giving the user the option to change their mind. 

### Actual result

File is silently deleted. 

### Additional information

I didn't verify but suspect this could also be an issue for files existing outside the `__tests__` directory. 

**Salesforce Extension Version in VS Code**:
v54.7.0

**SFDX CLI Version**:
sfdx-cli/7.144.2 darwin-x64 node-v16.14.2

**OS and version**:
MacOS

gbockus-sf:
@floralan This can be closed correct?

gbockus-sf:
Fixed in linked PR

PR #4039, opened 2022-04-14
fix: SFDX: Rename Component now shows an error message if one uses an existing file name as the new component name to avoid data loss.

floralan:
### What does this PR do?
Fix the bug that SFDX: Rename Component can silently replace a file if one uses an existing file name as new component name. 

#4024, @W-10983487@

### Functionality Before
if one renames a lightning/aura component and there is an existing file with the same name, it will be replaced without warning. 

### Functionality After
Added an error message saying that: "the component name is already in use for a file under current component directory" to prevent user from using an existing file name as new component name to avoid file loss.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
@@ -156,7 +156,9 @@ async function isDuplicate(componentPath: string, newName: string): Promise<bool
   return (allLwcComponents.includes(newName) || allAuraComponents.includes(newName));
 }
 
-// enforce unique new file name under current component directory to avoid file loss
+/**
+ * check duplicate name under current component directory and __tests__ directory to avoid file loss
+ */
 async function checkForDuplicateInComponent(componentPath: string, newName: string, items: string[]) {
   let allFiles = items;
   if (items.includes(TEST_FOLDER)) {
--- a/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
@@ -20,6 +20,7 @@ const RENAME_LIGHTNING_COMPONENT_EXECUTOR = 'force_rename_lightning_component';
 const RENAME_INPUT_PLACEHOLDER = 'rename_component_input_placeholder';
 const RENAME_INPUT_PROMPT = 'rename_component_input_prompt';
 const RENAME_INPUT_DUP_ERROR = 'rename_component_input_dup_error';
+const RENAME_INPUT_DUP_FILE_NAME_ERROR = 'rename_component_input_dup_file_name_error';
 const RENAME_WARNING = 'rename_component_warning';
 const LWC = 'lwc';
 const AURA = 'aura';
@@ -88,6 +89,7 @@ async function renameComponent(sourceFsPath: string, newName: string) {
   const componentName = getComponentName(componentPath);
   await checkForDuplicateName(componentPath, newName);
   const items = await fs.promises.readdir(componentPath);
+  await checkForDuplicateInComponent(componentPath, newName, items);
   for (const item of items) {
     // only rename the file that has same name with component
     if (isNameMatch(item, componentName, componentPath)) {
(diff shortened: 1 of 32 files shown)


sbudhirajadoc (changes requested):
Change -- Component name is already in use for a file in current component directory.

To. --  This component name is already in use in the current component directory. Choose a different name and try again.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -624,6 +624,7 @@ export const messages = {
   sobject_refresh_standard: 'Standard SObjects',
   force_rename_lightning_component: 'SFDX: Rename Component',
   rename_component_input_dup_error: 'Component name is already in use in LWC or Aura',
+  rename_component_input_dup_file_name_error: 'This component name is already in use in the current component directory. Choose a different name and try again.',
```
gbockus-sf, line 627: nit: should this be: This file name ?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -213,6 +214,48 @@ describe('Force Rename Lightning Component', () => {
...
+      try {
+        const executor = new RenameLwcComponentExecutor(sourceUri.fsPath);
+        await executor.run({
+          type: 'CONTINUE',
+          data: {name: 'templateOne'}
+        });
+      } catch (e) {
+        exceptionThrown = true;
+      }
+      expect(exceptionThrown).to.equal(true);
```
gbockus-sf, line 235: nit: you could verify what exception is getting thrown here. I've def had tests in the past where I thought it was throwing one kind of error, but it was actually failing for a different reason. same on line 256.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
gbockus-sf: nit^2: should we include the file name?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -154,6 +156,29 @@ async function isDuplicate(componentPath: string, newName: string): Promise<bool
   return (allLwcComponents.includes(newName) || allAuraComponents.includes(newName));
 }
 
+// enforce unique new file name under current component directory to avoid file loss
```
gbockus-sf, line 159: nit: for functions should we use the ```
/** 
 *
 **/```
comment syntax?

gbockus-sf (commented):
good stuff.  some nits to consider

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sbudhirajadoc: OK to switch to "This file name is already in use in the current component directory...."  if this is actually a file name in a component directory.

sbudhirajadoc (approved):

floralan:
> good stuff. some nits to consider

Great suggestions! Have updated the code based on your comments.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -154,6 +156,31 @@ async function isDuplicate(componentPath: string, newName: string): Promise<bool
...
+ * check duplicate name under current component directory and __tests__ directory to avoid file loss
+ */
+async function checkForDuplicateInComponent(componentPath: string, newName: string, items: string[]) {
+  let allFiles = items;
+  if (items.includes(TEST_FOLDER)) {
+    const testFiles = await fs.promises.readdir(path.join(componentPath, TEST_FOLDER));
+    allFiles = items.concat(testFiles);
+  }
+  // get all file names and filter out folder names
+  const allFileNames = allFiles.map(file => {
```
randi274, line 169: Is it possible to refactor this to create a named function so the comment can be removed?

floralan: @randi274 Sounds great! updated the code.

randi274 (approved):
**QA Complete**
* Can rename LWC component without an duplicate names → ✅
* Can rename LWC component w/ another component having a JS file with the intended name →   ✅
* Can’t rename LWC component with another LWC component bundle having the intended name → ✅
* Can’t rename LWC component w/ Aura having the intended bundle name → ✅
* Can’t rename Aura component w/ LWC having the intended bundle name → ✅
* Can’t rename LWC component that has a .js file (or any other file) matching the new intended name ✅
* Can’t rename LWC component that has a .js file (or any other file) matching the new intended name ✅
* Can’t run the command from the command palette at all ✅

I did a little extra QA since I didn't get a chance to on initial pass through. I also encountered a few additional questions about the feature, neither of which should prevent merge of this fix.

* ❓I noticed `SFDX: Rename Component` was not available on the __tests__ directory or any files therein. Should it be? 
* ❓Is it possible to reopen the new component after renaming? When I run the default Rename it does reopen the bundle, and currently functionality is that focus is lost and I have to find the component again.

floralan:
> **QA Complete**
> 
> * Can rename LWC component without an duplicate names → ✅
> * Can rename LWC component w/ another component having a JS file with the intended name →   ✅
> * Can’t rename LWC component with another LWC component bundle having the intended name → ✅
> * Can’t rename LWC component w/ Aura having the intended bundle name → ✅
> * Can’t rename Aura component w/ LWC having the intended bundle name → ✅
> * Can’t rename LWC component that has a .js file (or any other file) matching the new intended name ✅
> * Can’t rename LWC component that has a .js file (or any other file) matching the new intended name ✅
> * Can’t run the command from the command palette at all ✅
> 
> I did a little extra QA since I didn't get a chance to on initial pass through. I also encountered a few additional questions about the feature, neither of which should prevent merge of this fix.
> 
> * ❓I noticed `SFDX: Rename Component` was not available on the **tests** directory or any files therein. Should it be?
> * ❓Is it possible to reopen the new component after renaming? When I run the default Rename it does reopen the bundle, and currently functionality is that focus is lost and I have to find the component again.

@randi274 Thank you for the very detailed QA report! For your first questions, I changed the code to enable rename command for **__tests__** folder and any .js files under it. Could you take time to verify you can see the change on your end as well? For the second question, I think that's a very good feature users may expect. We can create a user story for it!

randi274:
Thanks for making that change so quickly @floralan! I think it might be a bit more of a complicated story, as it doesn't look like it's working as I'd expect quite yet; can you revert that commit, and then we'll add another user story to address that one?

floralan:
> 
Yes, making another story for the changes sounds a better plan! I just reset the branch, once all the tests done, we can get this PR merged 👍

PR #4061, opened 2022-04-25
chore: fix empty changelog issue

floralan:
### What does this PR do?
This PR will fix: 
- In the past two releases, changelog has been generated as an empty file: https://github.com/forcedotcom/salesforcedx-vscode/commit/f0a652a00af5ad7dab0568c3961f7548fb21eacf. 

@W-11026891@


scripts/change-log-generator-utils.js
```diff
@@ -46,7 +46,7 @@ function getPreviousReleaseBranch(releaseBranch) {
  function getReleaseBranches() {
   return shell
     .exec(
-      `git branch --remotes --list --sort='-creatordate' '${constants.RELEASE_BRANCH_PREFIX}*'`,
+      `git branch --remotes --list --sort='-creatordate' '${constants.REMOTE_RELEASE_BRANCH_PREFIX}*'`,
```
gbockus-sf, line 49: Is this what fixes our change log issue?

scripts/create-release-branch.js
```diff
@@ -68,9 +68,6 @@ if (isRemoteReleaseBranchExist) {
   process.exit(-1);
 }
 
-// Create the new release branch and switch to it
-shell.exec(`git checkout -b ${releaseBranchName}`);
-
```
gbockus-sf, line 73: So I'm lacking a bit in the how releases work department, but am I right that this script is assuming we're on the main branch to start?  So previously we were committing the changes to the release branch, and now we're committing directly to the main branch?  Do we ever push the changes back to the remote main branch?   I'm debating how I feel about committing directly to main.  Typically you don't want commits to be directly to the main branch.

gbockus-sf: actually never mind I finally processed the code above...so we're on develop.  The question is now how do we get from the release branch back to main? Usually you only want to catch up develop with main (otherwise why does main even exist?).  If I'm way off here please just ping me and straighten me out.

scripts/change-log-generator-utils.js
floralan: Yes, I think so. But I don't have the ability to test it since now v54.10.0 has been cut and there is no feat/fix PRs coming after it. So, I created this PR as a draft. Once we get PRs in this week, I can test if this is really the root cause of change log issue.

gbockus-sf (approved):
shipit

PR #4077, opened 2022-05-02
feat - Update Device Authentication UX in a Container Environment

jonnyhork:
### What does this PR do?
If a user wants to authenticate to an org with `SFDX: Authorize an Org` command in a container environment eg VS Code Remote Development or in Code Builder they have to use [Salesforce Device Flow](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_device_flow.htm&type=5) which requires them to copy a device code and go to a url to enter the code. Currently, we surface the information required to complete the auth in the output channel and can be easily missed or unclear that further action needs to be taken by the user.

This PR will update the UX for users so that when they run the `SFDX: Authorize an Org` in a container, we will open a browser window and pre-populate the device code in the input for them. The user can simply click through to complete the process.

 @[W-11038644](https://gus.lightning.force.com/a07EE00000vYEIwYAO)@

### Functionality Before
https://user-images.githubusercontent.com/25188632/166326234-a0f63d5d-4ad3-4a22-a409-19eea14395dc.mov

### Functionality After

https://user-images.githubusercontent.com/25188632/166326253-4225b741-2bc2-4ee0-ba23-204a0561f617.mov


packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +23,126 @@ import {
...
+
+  private parseAuthUrlFromStdOut(stdOut: string): string | undefined {
+    let authUrl;
+    try {
+      const response = JSON.parse(stdOut) as DeviceCodeResponse;
+      const verificationUrl = response.verification_uri;
+      const userCode = response.user_code;
+
+      if (verificationUrl && userCode) {
+        authUrl = `${verificationUrl}?user_code=${userCode}`;
```
jonnyhork, line 113: Adding the query param allows for the device code to be pre-populated in the form input as the first step for auth.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +23,126 @@ import {
...
+
+      if (!deviceCodeReceived) {
+        const authUrl = this.parseAuthUrlFromStdOut(stdOut);
+
+        if (authUrl) {
+          deviceCodeReceived = true;
+          // open the default browser
+          vscode.env.openExternal(vscode.Uri.parse(authUrl, true));
+        }
+      }
```
jonnyhork, line 89: This is the heart of the new functionality.

We will subscribe the to stdOut from the CLI which comes in two parts throughout this process:
1. DeviceCodeResponse when the CLI makes a request to authenticate
2. The OrgAuthInfo response once the user approves the "device" to access their org.

For this PR we only care about the `DeviceCodeResponse`
```js
{
  "user_code": "4VG9UGUD",
  "device_code": "UGxhdGZvcm1DTEk7OC4zNi4yMjYuMjI0OzE2NTExNjI5NzM5NzI7NFZHOVVHVUQ",
  "interval": 5,
  "verification_uri": "https://login.salesforce.com/setup/connect"
}
```
However, there is no guarantee that we will get the complete, parsable JSON in the stdOut stream all at once. So we will attempt to parse the data each time the subscriber is called until we get a complete authUrl from `parseAuthUrlFromStdOut `, then we will stop trying to parse the output from the CLI by flipping the `deviceCodeReceived ` flag.

If there is no authUrl we don't do anything and any errors will be handled by the CLI or SfdxCommandletExecutor as usual. 

For reference, here is the [CLI code](https://github.com/salesforcecli/plugin-auth/blob/main/src/commands/auth/device/login.ts) for this flow and [tests](https://github.com/salesforcecli/plugin-auth/blob/main/test/commands/auth/device/login.test.ts)

jonnyhork (commented):
Some of the changes in the i18n.ts file are formatting changes sorry for the noise.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -46,6 +46,8 @@ export const messages = {
     'SFDX: Authorize an Org using Session ID',
   force_auth_access_token_login_bad_oauth_token_message:
     'The session ID that you are trying to use is not valid. Check if it has expired, or use a valid session ID.',
+  force_auth_web_login_device_code_parse_error:
+    'There was an unexpected error authorizing to you org in a container environment.',
```
jonnyhork, line 50: This is the only new change in this file.

dehru: 'to your org'

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -129,9 +217,14 @@ const workspaceChecker = new SfdxWorkspaceChecker();
 const parameterGatherer = new AuthParamsGatherer();
 
 export function createAuthWebLoginExecutor(): SfdxCommandletExecutor<{}> {
-  return isDemoMode()
-    ? new ForceAuthWebLoginDemoModeExecutor()
-    : new ForceAuthWebLoginExecutor();
+  switch (true) {
```
dehru, line 220: fancy

dehru:
Let's add `Authorize a Dev Hub` to this flow too.  It also uses the old device flow.

cromwellryan:
For posterity and visibility...

Just want to jump in a say 👏 : for the quality of that PR’s description. 

* Thoughfully describes the situation and changes. 
* Videos :movie_camera: to describe the before and after. 
* Preemptive comments to highlight the areas of interest to the reviewer.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
jonnyhork: good catch, thanks.

jonnyhork:
@cromwellryan Thanks Ryan I really appreciate the feedback =)

jgellin-sf (commented):
LGTM

jonnyhork:
@dehru Thanks for pointing that out, I have updated the `SFDX: Authorize a Dev Hub` command with the same UX in [feat: impove ux of dev hub device auth flow](https://github.com/forcedotcom/salesforcedx-vscode/pull/4077/commits/f0e6a42c9bd23c8d689f6199cf9cceebc3e2d592)

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +24,136 @@ import {
...
+
+  private logToOutputChannel(code: string, url: string) {
+    channelService.appendLine(`${EOL}`);
+    channelService.appendLine(nls.localize('action_required'));
+    channelService.appendLine(
+      nls.localize('force_auth_device_login_enter_code', code, url)
+    );
+    channelService.appendLine(`${EOL}`);
+  }
+}
```
jonnyhork, line 138: I added this function to log the same user instructions to the output channel as the CLI. This way if we need to provide the user another way to get the information they need to complete authentication we can direct them to this. @AnanyaJha @gbockus-sf 
Note that the output channel will _not_ open on its own, they will have to get to it manually, as intended.
Screen Shot 2022-05-04 at 3 35 26 PM

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
```diff
@@ -34,22 +34,36 @@ import { isDemoMode } from '../../modes/demo-mode';
 import { isSFDXContainerMode } from '../../util';
```
gbockus-sf, line 34: Do you know why the above package-lock file is so much bigger then it was previously? It doesn't seem valid based on the changes in this PR.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
```diff
@@ -34,22 +34,36 @@ import { isDemoMode } from '../../modes/demo-mode';
...
+import { ForceAuthWebLoginContainerExecutor } from './forceAuthWebLogin';
+
+export class ForceAuthDevHubContainerExecutor extends ForceAuthWebLoginContainerExecutor {
+  public build(data: {}): Command {
+    const command = new SfdxCommandBuilder().withDescription(
+      nls.localize('force_auth_web_login_authorize_dev_hub_text')
+    );
+
+    command
+      .withArg('force:auth:device:login')
```
gbockus-sf, line 46: nit: We've (well me) has been pushing to starting using constants for these kind of string values.  Makes testing easier and reduces typo potential.  We don't quite have a good land place for the string yet but if you could just pull them up as constants in the file that'd be great. (note a nit from me means you can decide).  I realize we have this all over currently, but trying to make improvements as we go.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
```diff
@@ -119,9 +133,14 @@ const workspaceChecker = new SfdxWorkspaceChecker();
 const parameterGatherer = new EmptyParametersGatherer();
 
 export function createAuthDevHubExecutor(): SfdxCommandletExecutor<{}> {
-  return isDemoMode()
-    ? new ForceAuthDevHubDemoModeExecutor()
-    : new ForceAuthDevHubExecutor();
+  switch (true) {
+    case isSFDXContainerMode():
```
gbockus-sf, line 137: I have never seen a switch statement like this...interesting.  Def better then ? : ? : ? :

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +24,136 @@ import {
...
-      command
-        .withArg('force:auth:device:login')
-        .withLogName('force_auth_device_login');
-    } else {
-      command
-        .withArg('force:auth:web:login')
-        .withLogName('force_auth_web_login');
+
+    command
+      .withArg('force:auth:device:login')
```
gbockus-sf, line 56: nit: same nit on the string parameters.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +24,136 @@ import {
...
+
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+
+  private parseAuthUrlFromStdOut(stdOut: string): string | undefined {
+    let authUrl;
+    try {
+      const response = JSON.parse(stdOut) as DeviceCodeResponse;
+      const verificationUrl = response.verification_uri;
```
gbockus-sf, line 110: nit: kind of seems like a stretch, but is there any validation we should do on the URL here?  Could someone get a suspect URL to open in the browser if they do bad things to SFDX?

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -46,7 +46,11 @@ export const messages = {
     'SFDX: Authorize an Org using Session ID',
   force_auth_access_token_login_bad_oauth_token_message:
     'The session ID that you are trying to use is not valid. Check if it has expired, or use a valid session ID.',
-
```
gbockus-sf, line 49: @sbudhirajadoc Have you reviewed these stings?

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
jonnyhork: I am looking at the diff and some of the changes are formatting changes and what looks like just more metadata about the dependancies. I looks like the `lockfileversion` went from 1 ---> 2 which might explain a lot of the changes.
image

I didn't need to change anything with the dependancies for this work so we should be able to remove the commit with the lockfile if you want? The reason its in there is based on [this conversation](https://salesforce-internal.slack.com/archives/G0213NY8PMH/p1651253372344739) with Mohan

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
```diff
@@ -22,39 +24,136 @@ import {
...
+    notificationService.reportCommandExecutionStatus(
+      execution,
+      cancellationToken
+    );
+
+    ProgressNotification.show(execution, cancellationTokenSource);
+    taskViewService.addCommandExecution(execution, cancellationTokenSource);
+  }
+
+  private parseAuthUrlFromStdOut(stdOut: string): string | undefined {
```
gbockus-sf, line 106: This seems like a good Candidate for a service class/module.  Then it could be unit tested.  Maybe a static method on [AuthUtil.ts](https://github.com/forcedotcom/salesforcedx-vscode/blob/b032cf46357971e91d8e892239f806741e3cdf95/packages/salesforcedx-utils-vscode/src/auth/authUtil.ts#L13). thoughts?

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
gbockus-sf: Thanks for the explanation.  So in `parseAuthUrlFromStdOut` we show an error to the end user if there is a problem parsing the JSON.  Is that the right path is we have an expected flow where we'll get incomplete results?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/auth/forceAuthWebLogin.test.ts
```diff
@@ -243,42 +238,37 @@ describe('Force Auth Web Login is based on environment variables', () => {
...
+        loginUrl: TEST_URL
+      });
       expect(authWebLoginCommand.toCommand()).to.equal(
-        'sfdx force:auth:device:login --setalias  --instanceurl  --setdefaultusername'
+        `sfdx force:auth:device:login --setalias ${TEST_ALIAS} --instanceurl ${TEST_URL} --setdefaultusername --json --loglevel fatal`
+      );
+      expect(authWebLoginCommand.description).to.equal(
+        nls.localize('force_auth_web_login_authorize_org_text')
       );
     });
```
gbockus-sf, line 273: Do we test the execute method and related here? 
potential test cases: 
1. Was openExternal called and did it use the right url based on stdout from the command.
2. The case where the `execution.stdoutSubject.subscribe` is called first with invalid data then a second time with valid data. 
3. The case where the `execution.stdoutSubject.subscribe` is called with unexpected or error data.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -46,7 +46,11 @@ export const messages = {
...
+  force_auth_web_login_device_code_parse_error:
+    'There was an unexpected error authorizing to your org in a container environment.',
+  force_auth_device_login_enter_code:
+    'Enter %s user code in the verification URL %s',
+  action_required: '=== Action Required!',
```
gbockus-sf, line 53: @sbudhirajadoc  I think only the above messages are new. These are show to the user in the log output. See https://github.com/forcedotcom/salesforcedx-vscode/pull/4077/files#r865412273 for example

gbockus-sf (commented):
@jonnyhork mostly a few nits to consider and some questions around how it's tested.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
gbockus-sf: hum.. reverting seems safest here, but let's see what @mohanraj-r thinks.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
jonnyhork: I could try to string match parts of the URL we expect but I feel like we are really just a consumer of the CLI and SF OAuth, which I hope we could trust.

If an actor wanted to do something I think they would have to change the code here and get someone to install the new vsix, or man in the middle the cli response or SF auth and send back a new URL in the JSON.

Just let me know what you're comfortable with? =)

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
jonnyhork: Personally I'm not sure it deserves it own class yet. This function is very specific to the device auth and only used in this one execute method at this time. It also sounds like you're gonna rework the auth flows soon? Maybe then would be a better time to evaluate any abstractions that might help.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
gbockus-sf: Yeah def a stretch.  For the desktop vscode user we have them install sfdx as a pre-req so it wouldn't be too far of a stretch for them to install the wrong thing from a bad actor, but in code builder everything is installed for the end user so probably not as much of a concern.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
mohanraj-r: Yes please revert to version 1 - I have run into issues with getting vsce packaging to work correctly with npm v7/8 and the new lock file versions

packages/salesforcedx-vscode-core/test/vscode-integration/commands/auth/forceAuthWebLogin.test.ts
jonnyhork: I was thinking about these tests cases but ended up just going with what the current pattern was in the auth files. We would have to mock the stdOut etc, are there existing examples of that in place I can use?

Honestly, I think I was just lazy and hoping what we have would suffice, sorry 😬 
I'm fine adding these test cases.

gbockus-sf: ha no worries.  This is why code reviews exist :) If you search for `sandbox.stub` or just `stub(` in *.test.* files you'll find a lot of example (probably could get more specific and find exact stubs of the values you'll need to mock in the test file.  Only thing to watch out for is there are some tests that use proxyquire to attempt to be more pure unit tests and those can be more complex to setup.  Feel free to hit me up if you run into any trouble getting stuff mocked.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
jonnyhork: Sounds good, please see the [chore: use constants for cli commands](https://github.com/forcedotcom/salesforcedx-vscode/pull/4077/commits/99dab5ccbc87c730be6a76d17ed27bf03eb3913a) commit.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
jonnyhork: Done in [Revert "fix: update package lock"](https://github.com/forcedotcom/salesforcedx-vscode/pull/4077/commits/30a81d8db131129ed794ddf4ebcaa750977867ba)

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sbudhirajadoc: "**There was an unexpected error authorizing to your org in a container environment.**'" isn't very a helpful error message. I would like to add more context that would be helpful to a user. Let's work on this  together next week, @jonnyhork .Thanks, @gbockus-sf !

packages/salesforcedx-vscode-core/test/vscode-integration/commands/auth/forceAuthWebLogin.test.ts
jonnyhork: @gbockus-sf I added tests for the cases listed above, please see [chore: refactor for testing](https://github.com/forcedotcom/salesforcedx-vscode/pull/4077/commits/21975b6055a1d0e4586056cb466715fc037fcc6e)

I ended up going a different direction as I couldn't get any of the approaches, based on existing patterns, we discussed to work, so I refactored the code a bit and made it easier to tests just the parts we want.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthWebLogin.ts
jonnyhork: As far as the error message goes, I don't think we necessarily need to write to the channel service but I thought if there is a situation where the auth doesn't work, maybe from the users experience, nothing happens and maybe the VS Code says the command failed in a toast. It might be nice to have something logged to help indicate what went wrong if they report an issue. Unless they know to look in the output channel they wont see this error message, we will mostly just be logging a telemetry event.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/auth/forceAuthWebLogin.test.ts
```diff
@@ -243,43 +238,112 @@ describe('Force Auth Web Login is based on environment variables', () => {
...
+    const responseStr = JSON.stringify({
+      error: 500,
+      message: 'something went wrong'
+    });
+    deviceExecutor.injectResponse(responseStr);
+
+    expect(deviceExecutor.stdOut).to.be.equal(responseStr);
+    expect(openExternal.called).to.be.false;
+    expect(deviceExecutor.deviceCodeReceived).to.be.false;
+  });
```
gbockus-sf, line 348: ty! appreciate the effort.

gbockus-sf (approved):
Looks good!  Thanks @jonnyhork 
SFDX: Authorize an Org
✅  Word with out container env var
✅  Word with container env var (SFDX_CONTAINER_MODE)

PR #4091, opened 2022-05-09
fix: updated lerna from 3.22.1 to 4.0.0 [draft]

jeffb-sfdc:
<!--- PR title should follow the pattern: <type>(optional scope): <description>.
Updated lerna from 3.22.1 to 4.0.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
     default: '7.45.1'
   lerna-version:
     type: string
-    default: '3.13.1'
+    default: '4.0.0'
   node-version:
     type: string
     default: '14.16.0'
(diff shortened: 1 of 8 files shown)


jeffb-sfdc:
The unit tests in `salesforcedx-utils-vscode` is failing.  The error is `Error: Path contains invalid characters: C:\Users\circleci\project\packages\salesforcedx-utils-vscode\coverage\lcov-report\out\src\cli\webpack:\salesforcedx-utils-vscode` and it appears to be an issue with the `webpack:` (something that works on mac, but fails on Windows, which is what our build system uses.

From what we can tell, this has been failing for a long time, but lerna v3.22.1 ignored the error, but lerna v4.0.0 recognizes there's an error and the build fails.

The salesforcedx-utils-vscode project seems to be similar in structure to the salesforcedx-sobjects-faux-generator project (which is passing), so when this gets revisited, that would probably be the next place to revisit.

mohanraj-r:
Thanks for digging into this further @jeffb-sfdc 
I had seen and noted the error in the dependabot issue - but didn't debug it further and compare it to existing build logs. Can we create a WI and close this PR for now.

mohanraj-r:
@jeffb-sfdc Can you remove me from this PR please. I am getting daily notifications in slack about it being pending.

gbockus-sf:
covered by https://github.com/forcedotcom/salesforcedx-vscode/pull/4283

Issue #4053, opened 2022-04-21
Rename Component command should be available on the tests directory and for the JS files


randi274:
**Is your feature request related to a problem? Please describe.**
In order to create a more consistent user experience, the `SFDX: Rename Component` command should be available on directories underneath the main LWC bundle. 

**Describe the solution you'd like**
In LWCs, I'd like the command to be available for __tests__ and any JS files in that directory. The prompt should auto-populate the name of the component bundle (only adding the [regex](https://github.com/forcedotcom/salesforcedx-vscode/pull/4039/commits/5e0cfada8ff49e397f9b164a38f73143da9a1f02) for it to appear in the package.json causes __tests__ to auto-populate and the command to fail).

**Describe alternatives you've considered**
Click the main bundle directory or the main files to rename (js, html, css, etc all work).

PR #4139, opened 2022-05-17
fix: "SFDX: Rename Component" is available for test files in LWC now

floralan:
### What does this PR do?
Enable "SFDX: Rename Component" command for test files in LWC. 

#4053, @W-11152868@

### Functionality Before
"SFDX: Rename Component" command is not available for test files under `__tests__` directory under a lwc component. 

### Functionality After
Screen Shot 2022-05-17 at 10 17 46 AM

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -307,7 +307,7 @@
         },
         {
           "command": "sfdx.lightning.rename",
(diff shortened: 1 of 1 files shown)


randi274 (changes requested):
As discussed in hangout today, can we change this so the Rename Component is available for all children components/folders under a given component? This might require some rethinking of the logic to find the parent folder or name that we need to use (example: if I right click under __tests__ the JS file, it shouldn't auto-populate __test__, it should give the parent component name).

floralan:
> 
Since we need time to rethink about the logic and that may involve refactoring the class. Let me close this PR for now and create a new one when I have time to pick up this issue.

PR #4156, opened 2022-05-24
chore: update package-lock.json again

randi274:
### What does this PR do?
Commit that package-lock.json file from the release build again, similar to #4137.


klewis-sfdc (approved):
Thanks, Randi!

PR #4165, opened 2022-05-27
fix: renamed context to extensionContext

jeffb-sfdc:
### What does this PR do?
This PR renames instances of "context" to "extensionContext" to avoid confusion between ExtensionContext and WorkspaceContext variables.

#@W-11171762@

### Functionality Before / Functionality After
No functionality change

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/index.ts
+++ b/packages/salesforcedx-vscode-apex/src/index.ts
@@ -44,7 +44,7 @@ import { ApexTestRunner, TestRunType } from './views/testRunner';
 
 let languageClient: LanguageClient | undefined;
 
-export async function activate(context: vscode.ExtensionContext) {
+export async function activate(extensionContext: vscode.ExtensionContext) {
   const extensionHRStart = process.hrtime();
   if (vscode.workspace && vscode.workspace.workspaceFolders) {
     const apexDirPath = getTestResultsFolder(
@@ -63,18 +63,18 @@ export async function activate(context: vscode.ExtensionContext) {
       testOutlineProvider.onResultFileCreate(apexDirPath, uri.fsPath)
     );
 
-    context.subscriptions.push(testResultFileWatcher);
+    extensionContext.subscriptions.push(testResultFileWatcher);
   } else {
     throw new Error(nls.localize('cannot_determine_workspace'));
   }
 
   // Workspace Context
-  await workspaceContext.initialize(context);
+  await workspaceContext.initialize(extensionContext);
 
   // Telemetry
-  const extensionPackage = require(context.asAbsolutePath('./package.json'));
+  const extensionPackage = require(extensionContext.asAbsolutePath('./package.json'));
   await telemetryService.initializeService(
-    context,
+    extensionContext,
     APEX_EXTENSION_NAME,
     extensionPackage.aiKey,
     extensionPackage.version
@@ -83,11 +83,11 @@ export async function activate(context: vscode.ExtensionContext) {
   // Initialize Apex language server
   try {
     const langClientHRStart = process.hrtime();
-    languageClient = await languageServer.createLanguageServer(context);
+    languageClient = await languageServer.createLanguageServer(extensionContext);
     languageClientUtils.setClientInstance(languageClient);
     const handle = languageClient.start();
     languageClientUtils.setStatus(ClientStatus.Indexing, '');
-    context.subscriptions.push(handle);
+    extensionContext.subscriptions.push(handle);
 
     languageClient
       .onReady()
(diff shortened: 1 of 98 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -705,12 +705,12 @@ export async function activate(context: vscode.ExtensionContext) {
     SfdxCommandletExecutor,
     sfdxCoreSettings,
     SfdxWorkspaceChecker,
-    workspaceContext,
+    workspaceContext: workspaceContextInstance,
```
jeffb-sfdc, line 708: This is the one spot where it wasn't clear what the right thing to do is.  `api` is returned by `activate()` and I'm not familiar with how this is used, so I kept it as being returned as, "workspaceContext".  This compiles and all tests pass, and I performed a light smoke test and haven't found any issues.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -139,7 +139,7 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 function registerCommands(
-  extensionContext: vscode.ExtensionContext
+  // extensionContext: vscode.ExtensionContext
```
randi274, line 142: @jeffb-sfdc was this comment intentional?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveManifest.test.ts
```diff
@@ -11,8 +11,7 @@ import { expect } from 'chai';
 import * as path from 'path';
 import { createSandbox, SinonStub } from 'sinon';
 import { LibrarySourceRetrieveManifestExecutor } from '../../../src/commands/forceSourceRetrieveManifest';
-import { workspaceContext } from '../../../src/context';
-import { nls } from '../../../src/messages';
```
randi274, line 15: @jeffb-sfdc was NLS meant to be removed here too?

packages/salesforcedx-vscode-core/test/vscode-integration/telemetry/MockExtensionContext.ts
```diff
@@ -80,7 +80,7 @@ class MockEnvironmentVariableCollection
   }
 }
 
-export class MockContext implements ExtensionContext {
+export class MockExtensionContext implements ExtensionContext {
```
randi274, line 83: I actually do really like this change here - much clearer what/why we're mocking 👍

packages/salesforcedx-vscode-apex/src/index.ts
jeffb-sfdc: It's not used.  I'll go and remove it.

packages/salesforcedx-vscode-lwc/test/vscode-integration/activation/activationMode.test.ts
```diff
@@ -4,24 +4,26 @@ import * as vscode from 'vscode';
 import sinon, { stubInterface, stubObject } from 'ts-sinon';
 import * as sinonChai from 'sinon-chai';
 import { shared as lspCommon } from '@salesforce/lightning-lsp-common';
-import { MockContext } from './MockContext';
+import { MockExtensionContext } from './MockExtensionContext';
```
randi274, line 7: Just checking that the directory was renamed here as well? This was the case for all of the MockContext's?

randi274 (approved):
Those are all my questions - I think refactoring here makes it much easier to understand the different type of contexts used, especially as folks onboard to the product. Perhaps it was clearer to previous maintainers what the differences were, but I'm supportive of changes that make we think will make it easier to develop going forward! Given that builds are working and tests are passing, I'm ok approving once @klewis-sfdc has reviewed them with you as well. Maybe we can plan for a larger spot check during next week's release too?

klewis-sfdc (commented):
Thanks for the PR, Jeff!  Overall, I really like the changes from 
* 👍🏽 "context" --> "extensionContext"  and
* 👍🏽 "MockContext" --> "MockExtensionContext",  

but I'm not as in agreement with the changes to 
* ⚠️ add "Instance" into var names.

"Instance" is a technical implementation detail that IMO does not belong in the var name.  I actually prefer the expression on the left side here:
image
I would prefer not to establish or continue this naming convention.   Let me know what you think and how strongly you feel about wanting that change.  Happy to discuss more in standup today!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveManifest.test.ts
jeffb-sfdc: nls isn't used, so I removed it since workspaceContext (also not used) was removed.

packages/salesforcedx-vscode-lwc/test/vscode-integration/activation/activationMode.test.ts
jeffb-sfdc: No, just the file was renamed.  The path is `packages/salesforcedx-vscode-lwc/test/vscode-integration/activation/MockExtensionContext.ts`, and the parent directory, `activation` was not changed, just the filename.

jeffb-sfdc:
> but I'm not as in agreement with the changes to
> 
> * ⚠️ add "Instance" into var names.
> 

done

PR #4183, opened 2022-06-01
feat: added Org List panel [Draft]

jeffb-sfdc:
### What does this PR do?
This PR adds a new command, "SFDX: View Org List" which when invoked, opens a new WebView panel and displays information about the user's dev hubs and orgs.

This is a draft.  It can be cleaned up and be ready for production with only a few changes, but for now it's posted as-is to get an idea of how it works, and to solicit UI feedback.


gbockus-sf:
@jeffb-sfdc do we have an associated WI for this feature?  Def don't want to lose track of the value add here, but would like to close the PR until we can prioritize and take action.

jeffb-sfdc:
There hasn't been any movement on this so I'm going to close this PR.

PR #4192, opened 2022-06-06
fix(forcecreatemanifest): call and set sourceApiVersion correctly for…

angela-young:
### What does this PR do?
Fixed an issue where the manifest file generated would have a different version than the associated metadata version of the component.
https://github.com/forcedotcom/easywriter/issues/13, @W-10842770@

### Functionality Before
Create an Apex class, and its metadata file will say version 54. Generating a manifest file from that component will then have a version of 53.

### Functionality After
After generating a manifest file, it will now display the correct version (54).

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
@@ -14,6 +14,7 @@ import { format } from 'util';
 import * as vscode from 'vscode';
 import { OUTPUT_CHANNEL } from '../channels';
 import { nls } from '../messages';
+import { SfdxProjectConfig } from '../sfdxProject';
 import { getRootWorkspacePath } from '../util';
 import { FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
 
@@ -38,7 +39,9 @@ export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
                    progress?: vscode.Progress<{ message?: string | undefined; increment?: number | undefined; }>,
                    token?: vscode.CancellationToken): Promise<boolean> {
     if (this.sourcePaths) {
+      const sourceApiVersion = (await SfdxProjectConfig.getValue('sourceApiVersion')) as string;
       const componentSet = ComponentSet.fromSource(this.sourcePaths);
+      componentSet.sourceApiVersion = sourceApiVersion;
       if (this.responseText === undefined) {
         // Canceled and declined to name the document
         openUntitledDocument(componentSet);
--- a/packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
@@ -39,9 +39,13 @@ export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
                    progress?: vscode.Progress<{ message?: string | undefined; increment?: number | undefined; }>,
                    token?: vscode.CancellationToken): Promise<boolean> {
     if (this.sourcePaths) {
-      const sourceApiVersion = (await SfdxProjectConfig.getValue('sourceApiVersion')) as string;
       const componentSet = ComponentSet.fromSource(this.sourcePaths);
-      componentSet.sourceApiVersion = sourceApiVersion;
+      if (componentSet.apiVersion) {
+        // only get sourceApiVerion if there exists an apiVersion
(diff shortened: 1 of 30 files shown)


packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -38,7 +39,9 @@ export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
                    progress?: vscode.Progress<{ message?: string | undefined; increment?: number | undefined; }>,
                    token?: vscode.CancellationToken): Promise<boolean> {
     if (this.sourcePaths) {
+      const sourceApiVersion = (await SfdxProjectConfig.getValue('sourceApiVersion')) as string;
```
gbockus-sf, line 42: What type is returned from get value here? Just wondering why you had to cast it to a string.

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -38,7 +39,9 @@ export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
...
       const componentSet = ComponentSet.fromSource(this.sourcePaths);
+      componentSet.sourceApiVersion = sourceApiVersion;
```
gbockus-sf, line 44: 👍 Can you ensure that we have a test that covers this change here: `packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts`?

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
angela-young: When I ran it, it was telling me the type was string, but when I removed the cast, it gave me this error:
`Type 'string | number | boolean | JsonMap | JsonArray | null | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.`
So it seems like without the cast, the return type is undefined.

gbockus-sf: word. That makes sense 👍

packages/salesforcedx-vscode-core/src/commands/forceCreateManifest.ts
```diff
@@ -39,6 +40,12 @@ export class ManifestCreateExecutor extends LibraryCommandletExecutor<string> {
                    token?: vscode.CancellationToken): Promise<boolean> {
     if (this.sourcePaths) {
       const componentSet = ComponentSet.fromSource(this.sourcePaths);
+      if (componentSet.apiVersion) {
```
angela-young, line 43: Checks if there is an existing apiVersion before fetching the sourceApiVersion. Tests that use a dummy file will fail otherwise.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -184,6 +186,46 @@ describe('Force Create Manifest', () => {
       expect(openTextDocumentSpy.calledOnce).to.equal(true);
     });
 
+    describe('Verify manifest version.', () => {
```
RitamAgrawal, line 189: Nit-pick: Please check the indentation. It's off by one.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceCreateManifest.test.ts
```diff
@@ -184,6 +186,46 @@ describe('Force Create Manifest', () => {
...
+      const fakeVersion = '54.0';
+      let getValueStub: sinon.SinonStub;
+
+      beforeEach(() => {
+        getValueStub = env
+          .stub(SfdxProjectConfig, 'getValue')
+          .resolves(fakeVersion as unknown);
+      });
+
```
RitamAgrawal, line 198: I think you should add afterEach block here. This would ensure that environment is restored after each test. Something like this : 
```
afterEach(() => {
      env.restore();
    });
```

angela-young: Due to the describe block being nested, the afterEach of 'Happy Path Unit Tests' seems to apply, but I will add one just to be safe.

angela-young:
Currently looking into the desired behavior and use-cases of the Generate Manifest command.

angela-young:
Source of truth for apiVersion is correct currently. Should not be pulling from `sfdx-project.json`.

PR #4082, opened 2022-05-04
fix: add support for getRecords

shah-work:
### What does this PR do?
Add template for getRecords

### What issues does this PR fix or reference?
@W-10372964@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

PR #4193, opened 2022-06-07
chore : add template for getRecords

RitamAgrawal:
### What does this PR do?
Add template for getRecords
#4082 , @W-11263132@

### Functionality After
`lwc-wire-get-record-property` and `lwc-wire-get-record-function` ! Two new snippets are added.


gbockus-sf:
FYI @RitamAgrawal I think you'll need to make an issue to track this change and add the label community-contrib to that issue to have the related GUS WI created.

packages/salesforcedx-vscode-lwc/snippets/lwc.json
```diff
@@ -63,16 +63,16 @@
     "body": [
       "@api recordId;",
       "",
-      "@wire(getRecords, { records:[ { recordIds: [ ], fields: [ ] } ] } )",
-      "${2:records-property-name}"
+      "@wire(getRecords, { records:[ { recordIds: [ $2 ], fields: [ $0 ] } ] } )",
+      "${1:records-property-name}"
```
gbockus-sf, line 67: I updated the tab indexes for these two snippets to be in line with the getRecord snippits.

gbockus-sf (approved):
Changes looks good and consistent with what we see for the getRecord snippits. 

QE'd: 
- verified that I was able to utilize both snippits and they were at least as function as the existing getRecord snippts. ✅ 

On thing to call out is these snippits do not work stand alone and require the user to go add the correct imports.

PR #4203, opened 2022-06-10
Updated @salesforce/templates to 55.0.0.  The LWC template now adds a unit test

jeffb-sfdc:
### What does this PR do?
The PR adds a default/boilerplate unit test to the LWC template.

When a new LWC is created, only the .js and .html are created. We should encourage our customers to write unit tests, and adding a default unit test along with the component class will help encourage customers to write unit tests.

#@W-11171762@


gbockus-sf (approved):
Verified locally.  
When creating a new LWC component via command it includes unit test by default ✅ 

Thanks

PR #4208, opened 2022-06-13
fix: add additional context for scope of generate manifest command

angela-young:
### What does this PR do?
Adds an additional when clause to specify where the generate manifest file command should appear.

@W-10737780@

### Functionality Before
Selecting any file or folder in the project would allow you to generate a manifest file.

### Functionality After
The generate command now only appears in the package directory folder - ie. force-app directory.

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -303,7 +303,7 @@
         },
         {
           "command": "sfdx.create.manifest",
-          "when": "sfdx:project_opened"
+          "when": "resource =~ /force-app/ && sfdx:project_opened"
         },
         {
           "command": "sfdx.lightning.rename",


packages/salesforcedx-vscode-core/package.json
```diff
@@ -303,7 +303,7 @@
         },
         {
           "command": "sfdx.create.manifest",
-          "when": "sfdx:project_opened"
+          "when": "resource =~ /force-app/ && sfdx:project_opened"
```
marshallvaughn, line 306: This piece may break functionality for users who use alternate source root paths, e.g. `src` as is the standard [DX@Scale](https://docs.dxatscale.io)

elizabethiko: because of this piece we are not able to use the command as we are using multiple app folders - none of which is called force-app

randi274: Hey y'all, thanks for the feedback here! Can one of you log a bug on this in our issues tab so we can dig in and get a fix out soon? FYI on this one @angela-young https://github.com/forcedotcom/salesforcedx-vscode/issues

marshallvaughn: @randi274 done! Opened issue #4310.

Issue #4218, opened 2022-06-16
Bump SF schemas dependency in order to pick up the latest and greatest 


gbockus-sf:
**Is your feature request related to a problem? Please describe.**
Schemas was updated to address changes to project json files.  We need to update to pull in the json validation for those files. 

**Additional context**
See: https://github.com/forcedotcom/schemas/compare/v1.1.0...v1.1.3

PR #4219, opened 2022-06-16
build: update the salesforcedx-schemas version to pick up the latest release

gbockus-sf:
### What does this PR do?
Bumps salesforcedx-schemas to 1.1.3 -> https://github.com/forcedotcom/schemas/compare/v1.1.0...v1.1.3

#4218

Functionality updated to use latest json for validation of sfdx json files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6136,9 +6136,9 @@
       "integrity": "sha512-RbZBlljl+LjMZtzFsLpE9LneX7uiWS35M0ecJGkEI6ACIV0sKH48PRbSANVko+FNyzzSkLCc2m6HvXnwiaQoTQ=="
     },
     "node_modules/@salesforce/schemas": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/@salesforce/schemas/-/schemas-1.1.0.tgz",
-      "integrity": "sha512-6D7DvE6nFxpLyyTnrOIbbAeCJw2r/EpinFAcMh6gU0gA/CGfSbwV/8uR3uHLYL2zCyCZLH8jJ4dZ3BzCMqc+Eg=="
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/@salesforce/schemas/-/schemas-1.1.3.tgz",
+      "integrity": "sha512-XWohlOT2oQDqAJH00OXS3f2MGjkwZ6pr4emnnkHSQbg7UdGW0rvGpEnRKqBbDUfZ4K5YKSo9Gj216ZtaP3JLXg=="
     },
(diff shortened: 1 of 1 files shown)


jeffb-sfdc (approved):

package-lock.json
```diff
@@ -6136,9 +6136,9 @@
       "integrity": "sha512-RbZBlljl+LjMZtzFsLpE9LneX7uiWS35M0ecJGkEI6ACIV0sKH48PRbSANVko+FNyzzSkLCc2m6HvXnwiaQoTQ=="
     },
     "node_modules/@salesforce/schemas": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/@salesforce/schemas/-/schemas-1.1.0.tgz",
-      "integrity": "sha512-6D7DvE6nFxpLyyTnrOIbbAeCJw2r/EpinFAcMh6gU0gA/CGfSbwV/8uR3uHLYL2zCyCZLH8jJ4dZ3BzCMqc+Eg=="
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/@salesforce/schemas/-/schemas-1.1.3.tgz",
+      "integrity": "sha512-XWohlOT2oQDqAJH00OXS3f2MGjkwZ6pr4emnnkHSQbg7UdGW0rvGpEnRKqBbDUfZ4K5YKSo9Gj216ZtaP3JLXg=="
```
dehru, line 6141: I wonder what the benefit is of updating this vs updating the dependency in [package.json here](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.json#L32) to something like `^1.1.3`

dehru:
I build and tested this branch.  I found there was intellisense available on the 2 files that rely on this schema.
Screen Shot 2022-06-22 at 11 05 03 AM
Screen Shot 2022-06-22 at 10 59 32 AM
Screen Shot 2022-06-22 at 10 58 24 AM

dehru (approved):
QA passed

package-lock.json
gbockus-sf: So since the package.json has 
`    "@salesforce/schemas": "^1",`
we'll always update to the latest patch version when it is available (when using npm install). 

However the package-lock.json file is intended to ensure that we get exactly the same version of dependancies when we build for production.  I _think_ (I always have to review this) when doing a build on circleci with use `npm ci` which always adheres to what is specified in the lock file verse what is in the package.json files.  This is to ensure that we are able to have an specific set of dep versions installed when building for prod.  This is so we don't break if something bumps the version and we've never installed it locally.  It gives us a finer level of control on what gets included in the production build.  We can def chat more about this as I feel like this explanation is kind of confusing :P

Issue #3572, opened 2021-10-07
Sfdx create apex class button does not appear on nested folders


AllanOricil:
### Summary

If I click with the right mouse button on the classes directory, the command is available. If I click on a nested directory the command is missing. The command should appear on any nested folder within the classes directory.

### Steps To Reproduce:

1.  Create a folder inside the classes directory
2.  click with the right mouse button on this folder
3.  verify you cant see the create Apex class command

### Expected result

sfdx create Apex class command should appear when clicking on any nested directories that are inside the classes directory

### Actual results

Sfdx create Apex class command does not appear when clicking on a nested folder inside the classes directory

### Additional information

**Salesforce Extension Version in VS Code**:  

**SFDX CLI Version**:

**OS and version**:

PawelWozniak:
I see a menu item but when I click it and provide the name then it still creates apex class in main directory as there is no other choice. In such case I would expect even skipping this choice menu and direct file create under a subdirectory.
<img width="264" alt="2022-07-04_08h35_21" src="https://user-images.githubusercontent.com/5740564/177095872-51f92dcf-75fc-471d-871c-9c1caae12d08.png">

randi274:
@PawelWozniak are you finding that even if you chose another directory, the extensions still create the Apex class under the default `classes` directory? That sounds like it might be a different 🐛 worth logging separately.

randi274:
Also, closing this one as @dehru [delivered](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode/CHANGELOG.md#5530---june-29-2022) it for us last month 🎉

PR #4224, opened 2022-06-20
fix: use regex to match sub-directories in when clause

dehru:
### What does this PR do?

This PR fixes the issue where the "Create Apex Class" option is not available in sub-directories of the `classes` directory when a user right-clicks on windows ( or control clicks on mac ).

#https://github.com/forcedotcom/salesforcedx-vscode/issues/3572, @W-10288492@

### Functionality Before

https://user-images.githubusercontent.com/599418/174657434-e32a4636-8d60-4ff1-bc6a-153a3889be92.mov

### Functionality After

https://user-images.githubusercontent.com/599418/174658151-8ef486cd-9ee6-40c8-abd3-252b8ace425e.mov

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -215,7 +215,7 @@
         },
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && resourceFilename == classes && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourcePath =~ /classes/ && sfdx:project_opened"
         },
         {
           "command": "sfdx.force.folder.diff",


packages/salesforcedx-vscode-core/package.json
```diff
@@ -215,7 +215,7 @@
         },
         {
           "command": "sfdx.force.apex.class.create",
-          "when": "explorerResourceIsFolder && resourceFilename == classes && sfdx:project_opened"
+          "when": "explorerResourceIsFolder && resourcePath =~ /classes/ && sfdx:project_opened"
```
dehru, line 218: This could also be further protected by adding more restrictions to the regex.  We could look for the presence of a directory ( windows or linux ) before the word `classes`.  

```resource =~ /(\/|\\\\)classes/```

But I just wasn't sure if that was overly protective when this regex reads easier.

jeffb-sfdc: One thing I found was that if I created a `classes` directory in aura (so `/app/main/default/aura/classes`) the "SFDX: Create Apex Class" command is available.  Though not appropriate, I don't expect users will be creating folders called "classes".  (just something to be aware of).  I guess let's add it in and then see if we get any issues logged.

jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/package.json
dehru: It could also get fired if someone created a directory called 'myclasses' since there is no check for directory separators.  But I too was wondering what is over-protective compared to actual user behavior.

jeffb-sfdc:
I QA'd it and didn't find any issues.

PR #4225, opened 2022-06-21
Allow rename of lwc component inside of __tests__ directory

dehru:
### What does this PR do?

This PR adjusts the when clause for the Rename Component command so that it works in any file in a subdirectory of the default `lwc` or `aura` with the right file extensions.  It also adds a simple utility function that will resolve the directory above the `__tests__` directory so that the rest of the rename component functionality works as expected.

#https://github.com/forcedotcom/salesforcedx-vscode/issues/4053, @W-11152868@

### Functionality Before

https://user-images.githubusercontent.com/599418/174902502-77552e32-a291-4970-ad10-ba29a3735584.mov

### Functionality After

https://user-images.githubusercontent.com/599418/174902548-34d8394e-e768-4b8b-a960-fc66cbf1e5c8.mov


packages/salesforcedx-vscode-core/package.json
```diff
@@ -307,7 +307,7 @@
         },
         {
           "command": "sfdx.lightning.rename",
-          "when": "sfdx:project_opened && resource =~ /.*/(lwc|aura)/[^/]+(/[^/]+\\.(html|css|js|xml|svg|cmp|app|design|auradoc))?$/"
+          "when": "sfdx:project_opened && resource =~ /.*/(lwc|aura)/.*(/[^/]+\\.(html|css|js|xml|svg|cmp|app|design|auradoc))?$/"
```
dehru, line 310: Basically, show the command to the user on anything inside the /lwc/ and /aura/ directories that have these file extensions.  This includes the component subdirectory named `__tests__` and files inside there.

If the user comes up with their own folder structure, this function should still work to resolve the parent component directory.

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -132,9 +132,19 @@ async function renameComponent(sourceFsPath: string, newName: string) {
...
+  directories.splice(directories.length - 1);
+  return directories.join(pathSeparator);
+}
+
 async function getComponentPath(sourceFsPath: string): Promise<string> {
   const stats = await fs.promises.stat(sourceFsPath);
-  return stats.isFile() ? path.dirname(sourceFsPath) : sourceFsPath;
+  let dirname = stats.isFile() ? path.dirname(sourceFsPath) : sourceFsPath;
+  if (dirname.endsWith(TEST_FOLDER)) {
+    dirname = getParentDirectoryOfTestFolder(dirname, path.sep);
```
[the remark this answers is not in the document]
dehru: This is where i got tripped up with Windows vs OSX in unit testing.  How you suggested was my original implementation.  I'd appreciate your thoughts on it.

If I use path.sep inside the function itself instead of passing it in, it works great at runtime.  The path that is generated by the OS can be parsed using path.sep.

But during testing, i wanted to unit test that this function works on both platforms. if I want to pass in a string like this `force-app/main/default/lwc/helloworld/__tests__` and verify that it get parsed and the `__tests__` directory removed, the test only passes on my OSX operating system and fails on our Windows CI.  Because on Windows the function under unit test is now looking for `\` and the unit test fails.

I thought about generating the string in the unit test dynamically like `force-app${path.sep}main${path.sep}default${path.sep}lwc${path.sep}helloworld${path.sep}__tests__`.  And maybe that's better, but it just seemed weird that a unit test would run differently depending on where it was executed.

So anyway, I opted to pass in the path.sep into the function, and this allowed me to unit test the function for both operating systems, and it's easy to read the unit test.

So options are:
1. Leave it as is ( pass the separator into the function )
2. Generate the path dynamically in the unit test
3. Use an if/else in the unit test.  Like `if ( path.sep === '/' ) {} else {}` inside the it statement of the unit test based on environment.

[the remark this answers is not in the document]
jeffb-sfdc: I looked trough our codebase, but I don't see many places where we are creating platform-specific tests.  I just ran through the exercise of, "what tests would I write if I had written getParentDirectoryOfTestFolder()"?

I started off with just the implementation:
```
export function getParentDirectoryOfTestFolder(sourceFsPath: string): string {
  const directories = sourceFsPath.split(path.sep);
  directories.splice(directories.length - 1);
  return directories.join(path.sep);
}
```

...and then thought about writing a test for it.  This is what immediately came to mind:
```
it('validates getParentDirectoryOfTestFolder', () => {
    const folders = ['path', 'to', 'tests']; // mocked(fake) path
    const folderPath = folders.join(path.sep);

    const parentFolder = folders.slice(0, -1);
    const parentFolderPath = parentFolder.join(path.sep);

    const parentDirectory = getParentDirectoryOfTestFolder(folderPath);
    expect(parentDirectory).to.equal(parentFolderPath)
  });
```

With this test, it validates the logic & behavior inside the function, and leaves the implementation as a black box.

OK, so, thinking out loud here... I could then see one saying that my test is just mirroring the implementation of what's in `getParentDirectoryOfTestFolder()`, which I would say, that's OK, we just want to validate that it works. 
 But if that wan't acceptable, the the next thing I came up with was a test like this:
```
it('validates getParentDirectoryOfTestFolder', () => {
    const folderPath = 'path|to|tests'.replace(/\|/g, path.sep);
    const parentFolderPath = 'path|to'.replace(/\|/g, path.sep);

    const parentDirectory = getParentDirectoryOfTestFolder(folderPath);
    expect(parentDirectory).to.equal(parentFolderPath)
  });
```

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -399,5 +399,19 @@ describe('Force Rename Lightning Component', () => {
...
+  describe('getParentDirectoryOfTestFolder function', () => {
+    it('validates getParentDirectoryOfTestFolder', () => {
+      const folders = ['path', 'to', '__tests__'];
+      const folderPath = folders.join(path.sep);
+
+      const parentFolder = folders.slice(0, -1);
+      const parentFolderPath = parentFolder.join(path.sep);
+
+      const parentDirectory = getParentDirectoryOfTestFolder(folderPath);
+      expect(parentDirectory).to.equal(parentFolderPath);
```
dehru, line 414: @jeffb-sfdc - I ended up with your first suggestion.  I think this is a good balance.  Thanks for the input.

dehru:
@klewis-sfdc Thanks for pointing me to eBikes as a source of interesting folder structure inside the aura and lwc directories that I hadn't considered.  This led me to refactor how we look up the component directory.  Now we search the directory path until we find the last instance 'lwc' or 'aura', and then the component directory exists below that.

Here's a movie showing me exercise the bugs you found above.

https://user-images.githubusercontent.com/599418/176777356-f81b4141-86fe-45fc-adc3-06fdb727ae80.mov

dehru:
NOTE, I'm seeing that same error from CI locally.  sobjects-faux-generator is getting a 404 from npm trying to install @types for eslint.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -639,6 +639,8 @@ export const messages = {
     'Press Enter to confirm your input or Escape to cancel',
   rename_component_warning:
     'Warning: References to the old name will not be updated. Update manually and redeploy once all changes have been made.',
+  rename_component_error:
+    'Unable to rename component. Update manually and redeploy once all changes have been made.',
```
dehru, line 643: @sbudhirajadoc - I added a pop-tart message if there is ever an error running this command.  I noticed that we also throw a generic failure message.  So this might not be necessary.  Let me know what you think.  

The nasty details of the error message will display in the output channel.
Screen Shot 2022-06-30 at 2 31 33 PM

klewis-sfdc (approved):
Looks great @dehru!  Thanks for the improvements!  

- [x ] SFDX: Rename Component command shows up in right click menu when under main files within tests directory under LWC or Aura component structure
- [x ] SFDX: Rename Component command works to rename main files with same name as component under LWC or Aura component structure

packages/salesforcedx-vscode-core/src/messages/i18n.ts
sbudhirajadoc: @dehru I am not clear on what the action "Update manually and redeploy once all the changes have been made" means. What do they update? How  do they redeploy?

PR #4231, opened 2022-06-23
fix: add defaultUsername and defaultDevhubUsername to ui

dehru:
### What does this PR do?
This is my addition to jeff's work on hackday.  It adds a column for defaultDevhubUsername and defaultUsername.  I had to execute the command from the root workspace directory to get those fields into the results.

It also hacks together some styling changes in a very bad way....until we can get some time to investigate how better to style these vscode webview components.

### Functionality Before
<insert gif and/or summary>

### Functionality After
Screen Shot 2022-06-23 at 3 45 11 PM

### DIFF
--- a/packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
+++ b/packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
@@ -139,7 +139,7 @@ export class OrgListPanel {
 
                 <vscode-data-grid
                   class='all-non-scratch-orgs-data-grid'
-                  grid-template-columns='50px 1fr 2fr 1fr 1fr'
+                  grid-template-columns='50px 1fr 2fr 1fr 2fr'
                   aria-label='All Non Scratch Orgs'
                 >
                 </vscode-data-grid>
@@ -154,7 +154,7 @@ export class OrgListPanel {
 
                 <vscode-data-grid
                   class='all-scratch-orgs-data-grid'
-                  grid-template-columns='50px 1fr 2fr 1fr 1fr'
+                  grid-template-columns='50px 1fr 2fr 1fr 1fr 1fr'
                   aria-label='All Scratch Orgs'
                 >
                 </vscode-data-grid>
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
```diff
@@ -302,6 +303,19 @@ export class OrgListPanel {
                   // Now display the the content (the panel and the grids)
                   var orgListContentEl = document.getElementById('org-list-content');
                   orgListContentEl.style.display = 'contents';
+
+                  setTimeout(() => {
```
dehru, line 307: Forgive me.  This is nasty.  The dom is not updated immediately, so a slight timeout allowed me to query and find the elements i wanted to change the style of.  But this is nothing but a short-term hack.

jeffb-sfdc: Oh, that's ugly.  (but it made me laugh) Yea, we need to get this resolved b/f shipping.  I opened a ticket up on the issue at https://github.com/microsoft/vscode-webview-ui-toolkit/issues/378

packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
```diff
@@ -155,7 +154,7 @@ export class OrgListPanel {
 
                 <vscode-data-grid
                   class='all-scratch-orgs-data-grid'
-                  grid-template-columns='240px 320px 200px 200px 200px 200px'
+                  grid-template-columns='50px 1fr 2fr 1fr 1fr'
```
jeffb-sfdc, line 157: One of the columns is missing, and should probably be `'50px 1fr 2fr 1fr 1fr 1fr'`

However, even with the extra column, the alignment is off.  What I see in the "All" tab is:
image

packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
dehru: I'd love to hack on this some more if ananya wants to move forward.

packages/salesforcedx-vscode-core/src/panels/orgListPanel.ts
```diff
@@ -140,7 +139,7 @@ export class OrgListPanel {
 
                 <vscode-data-grid
                   class='all-non-scratch-orgs-data-grid'
-                  grid-template-columns='240px 320px 200px 200px'
+                  grid-template-columns='50px 1fr 2fr 1fr 2fr'
```
dehru, line 142: last one takes up 2 columns.

PR #4232, opened 2022-06-24
fix: fixed a bug pertaining to VS Code passing a stale file path after a file has been renamed

jeffb-sfdc:
### What does this PR do?

### Functionality Before

### Functionality After
1-3: the same
4. This deployment succeeds

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -35,9 +35,9 @@ export function flushFilePath(filePath: string): string {
 }
 
 export function flushFilePaths(filePaths: string[]): string[] {
-  for(let i=0; i<filePaths.length; i++) {
+  for (let i = 0; i < filePaths.length; i++) {
     filePaths[i] = flushFilePath(filePaths[i]);
-  };
+  }
 
   return filePaths;
 }
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -5,6 +5,8 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import * as fs from 'fs';
+
 export function isNullOrUndefined(object: any): object is null | undefined {
   if (object === null || object === undefined) {
     return true;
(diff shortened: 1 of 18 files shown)


packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +16,28 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
   return JSON.parse(jsonString);
 }
+
+// There's a bug in VS Code where, after a file has been renamed,
+// the URI that VS Code passes to the command is stale and is the
+// original URL.  See https://github.com/microsoft/vscode/issues/152993.
+//
+// To get around this, fs.realpathSync.native() is called to get the
+// URI with the actual file name.
+export function flushFilePath(filePath: string): string {
```
gbockus-sf, line 33: Nit: Naming is hard.  I'm ok with flushFilePath, but would getCaseSenstiveFilePath be more accurate as far as what we're trying to accomplish?

gbockus-sf, line 33: Should we have the input here be of type `string | undefined` and error if an undefined value is passed?  That would save us a lot of `if(inputFile)` checks in the calling code.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +16,28 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
+  return fs.realpathSync.native(filePath);
```
gbockus-sf, line 34: good find 👍

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
jeffb-sfdc: I don't think it wold be more accurate, because the path is already case sensitive, and the issue is that it is old & stale, and needs to be updated.  It took me a while to land on `flushFilePath`, just b/c I couldn't think of a name that's more fitting or more accurately described the issue or what it does.

But that being said, I'm OK with the current name but also not totally in love with it either.  I don't think it should be `getCaseSenstiveFilePath`, but if you have any other suggestions, please let me know.

jeffb-sfdc: Done (this will be in my other PR, as I'm going to close this one out)

jeffb-sfdc:
> @jeffb-sfdc this should be closed in favor or #4240 coerrect?

Yes, I wanted to make sire that I could get a new PR to build first.  With this PR, I found the issue, found other areas that need to have the same fix applied, and then had over 100 untegration tests fail.  Not b/c I broke anything, but b/c there were many test which end up calling this code, and use a mocked file path, and when `fs.realpathSync.native()` is called with a bogus file path, it throws an exception.

I found that I had dug myself into a hole and decided to star over, adding `fs.realpathSync.native()` and then changing on package at a time.

PR #4240, opened 2022-06-28
fix: added flushFilePath()

jeffb-sfdc:
### What does this PR do?

### Functionality Before

### Functionality After
1-3: the same
4. This deployment succeeds


packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.test.ts
```diff
@@ -314,10 +314,10 @@ describe('ForceFunctionContainerStartExecutor unit tests', () => {
...
       assert.calledOnce(fakeDisposible.dispose);
       assert.calledTwice(localizeStub);
-      console.log('what are the args', {
-        args0: sendExceptionStub.getCalls()[0].args[0],
-        args1: sendExceptionStub.getCalls()[0].args[1]
-      });
+      // console.log('what are the args', {
+      //   args0: sendExceptionStub.getCalls()[0].args[0],
+      //   args1: sendExceptionStub.getCalls()[0].args[1]
+      // });
```
klewis-sfdc, line 320: What do you think about either adding a comment here as to why this is no longer applicable or removing entirely?

klewis-sfdc (commented):
In QA so far, I can confirm that renaming by changing the casing only and then deploying now works correctly.  Nice!

Looking at the changes in the PR, are there delete or retrieve actions that should also be tested here?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.test.ts
jeffb-sfdc: I'll remove them.  This was there and was adding noise to the output.  salesforcedx-vscode-core has a lot of noise in it when the untegration tests run.  When the tests run (either locally, or in CI) there are lots of errors reported, and it's at the point where it's hard to tell when a failure is a failure or not.  This is just a "drop in the ocean" and I hope to focus on this at some point and clean up the build.

klewis-sfdc (approved):
Looks good to me!  Thanks @jeffb-sfdc !

- [x] SFDX: Deploy Source to Org works after renaming locally by only making changes to the casing in the file/class name
- [x] SFDX: Deploy Source to Org works when deploying multiple files after renaming one locally by only making changes to the casing in the file/class name
- [x] SFDX: Retrieve Source from Org works as before
    - Note: If you rename locally, you still need to deploy the class first so that it is renamed on the server before executing a “Retrieve” operation
- [x] SFDX: Delete from Project and Org works as before
    - Note: Similarly, if you rename the component/class locally, you need to deploy it so that the name is aligned on the server before running a delete operation

jeffb-sfdc:
> In QA so far, I can confirm that renaming by changing the casing only and then deploying now works correctly. Nice!
> 
Just a heads up - I updated the existing tests and they pass.  I'm now adding tests for `flushFilePath()`.

klewis-sfdc:
Also tested "SFDX: Launch Apex Replay Debugger with Current File" - working as before - looks good!  ✅

PR #4245, opened 2022-06-29
Upgrade "SetConfig" Command to use Core Library 

angela-young:
### What does this PR do?
Rewrites the force:config:set command to use the core library instead of the CLI library. All functionality should remain the same aside from minor changes in the output displayed.

@W-9466545@

### Functionality Before
CLI Output

### Functionality After
Core Library Output


packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,66 @@
...
+    await config.write();
+    this.responses.push({name: CONFIG_NAME, val: this.usernameOrAlias, success: String(true) });
+    const outputTable = this.formatOutput(this.responses);
+    channelService.appendLine(outputTable);
+    return true;
+  }
+
+  public getUsernameOrAlias() {
+    return this.usernameOrAlias;
+  }
```
angela-young, line 52: This is used in the testing files to make sure the username is set correctly.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -6,34 +6,35 @@
  */
 
 import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { createSandbox } from 'sinon';
+import * as vscode from 'vscode';
 import { ForceConfigSetExecutor } from '../../../src/commands';
-import { nls } from '../../../src/messages';
+
+const env = createSandbox();
```
gbockus-sf, line 14: nit: can you rename env -> sandbox. It would just be more consistent with existing code.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -6,34 +6,35 @@
...
     const usernameOrAlias = 'test-username1@gmail.com';
     const forceConfigSet = new ForceConfigSetExecutor(usernameOrAlias);
-    const forceConfigSetCommand = forceConfigSet.build({});
-    expect(forceConfigSetCommand.toCommand()).to.equal(
-      `sfdx force:config:set defaultusername=${usernameOrAlias}`
-    );
-    expect(forceConfigSetCommand.description).to.equal(
-      nls.localize('force_config_set_org_text')
-    );
+    env.stub(forceConfigSet, 'run').returns(true);
```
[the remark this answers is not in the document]
randi274: If not, it would be good to test the table code directly

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,66 @@
...
+
+  private formatOutput(input: Row[]): string {
+    const title = 'Set Config'; // todo: localize
+    const table = new Table();
+    const outputTable = table.createTable(
+      input,
+      [
+        { key: 'name', label: 'Name'},
```
randi274, line 60: nit: extra space before the first key, or would recommend adding spaces before each `key` and after the label value. not sure we have coding conventions for either, but I think the extra space on each side is nicer given the options

randi274 (changes requested):
Some localization and test comments mostly, but good to see this up and ready to go!

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,66 @@
...
+  public async run(response: ContinueResponse<string>): Promise<boolean> {
+    // In order to correctly setup Config, the process directory needs to be set to the current workspace directory
+    const path = getRootWorkspacePath(); // Get current workspace path
+    process.chdir(path); // Set process directory
+
+    const config = await Config.create(Config.getDefaultOptions());
+
+    config.set(CONFIG_NAME, this.usernameOrAlias);
+    await config.write();
+    this.responses.push({name: CONFIG_NAME, val: this.usernameOrAlias, success: String(true) });
```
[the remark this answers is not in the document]
angela-young: I think I was just copying what CLI seemed to be doing, so I can definitely see why there's some confusion here. Another thing to consider though is that the `createTable` method takes in a list, so it might make sense to keep it.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -6,34 +6,34 @@
...
-    );
-    expect(forceConfigSetCommand.description).to.equal(
-      nls.localize('force_config_set_org_text')
-    );
+    sandbox.stub(forceConfigSet, 'run').returns(true);
+    expect(forceConfigSet.getUsernameOrAlias()).to.equal(usernameOrAlias);
   });
-});
 
-describe('Force Config Set using multiple aliases for a single username', () => {
```
randi274, line 26: More notes while we're working through this - have you tested what happens when a user has multiple aliases for a single username?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -6,34 +6,34 @@
...
+import * as vscode from 'vscode';
 import { ForceConfigSetExecutor } from '../../../src/commands';
-import { nls } from '../../../src/messages';
+
+const sandbox = sinon.createSandbox();
+let openTextDocumentSpy: sinon.SinonSpy;
 
 describe('Force Config Set', () => {
+  beforeEach(() => {
+    openTextDocumentSpy = sandbox.spy(vscode.workspace, 'openTextDocument');
```
randi274, line 18: Ah, looks like this one can be removed!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
angela-young: Yes, I've had multiple aliases in the past, and the expected behavior is that the first one is the one that gets set in the config. I still plan to keep this test, just modify it so we're making sure config is set with the correct alias.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,66 @@
...
+    await forceConfigSet(usernameOrAlias);
+    expect(configSetSpy.calledOnce);
+    expect(configSetSpy.calledWith(CONFIG_KEY, usernameOrAlias)).to.equal(true);
+    expect(configWriteSpy.calledOnce);
   });
-});
 
-describe('Force Config Set using multiple aliases for a single username', () => {
-  it('should build the force config set command with first alias', async () => {
+  it('should set config with first alias', async () => {
```
randi274, line 45: Love to see this test covered here. Is there any scenario in which there would be no alias to set? It's more likely that we'd encounter this if the user was making modifications to that file directly, or through another program.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,66 @@
...
-    );
-    expect(forceConfigSetCommand.description).to.equal(
-      nls.localize('force_config_set_org_text')
-    );
+    await forceConfigSet(aliases.join(','));
+    expect(configSetSpy.callCount).to.equal(1);
+    expect(configSetSpy.calledWith(CONFIG_KEY, expectedAlias)).to.equal(true);
+  });
+
+  it('should display formatted output in output channel', async () => {
```
randi274, line 53: Looks great! Really covering all our bases here 😄

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,66 @@
...
+    configWriteSpy = sandbox.spy(Config.prototype, 'write');
+    tableSpy = sandbox.spy(Table.prototype, 'createTable');
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should set config with the given username or alias', async () => {
     const usernameOrAlias = 'test-username1@gmail.com';
```
randi274, line 38: Small nit: can we move usernameOrAlias to a const at the top of the test file since we're reusing it throughout?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
angela-young: I've been looking into that, and the only situation where the Set Config command gets called is when the user selects a valid alias from a dropdown. So invalid and blank inputs can not be entered.

randi274 (approved):
This is looking great! Testing on Windows was good too.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,66 @@
...
-      .withArg(`defaultusername=${this.usernameOrAlias}`)
-      .build();
+  public async run(response: ContinueResponse<string>): Promise<boolean> {
+    // In order to correctly setup Config, the process directory needs to be set to the current workspace directory
+    const path = getRootWorkspacePath(); // Get current workspace path
+    process.chdir(path); // Set process directory
+
+    const config = await Config.create(Config.getDefaultOptions());
+
+    config.set(nls.localize(CONFIG_NAME), this.usernameOrAlias);
```
angela-young, line 46: Expired orgs can not be set anymore, need to add a check for valid orgs now.

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,86 @@
...
+import { getRootWorkspacePath } from '../util/rootWorkspace';
 import {
   EmptyParametersGatherer,
   SfdxCommandlet,
-  SfdxCommandletExecutor,
   SfdxWorkspaceChecker
 } from './util';
 
-export class ForceConfigSetExecutor extends SfdxCommandletExecutor<{}> {
+const CONFIG_SET_EXECUTOR = 'force_config_set_org_text';
```
randi274, line 20: Can these be moved into that `const` file @angela-young, the one under src?

sbudhirajadoc (approved):

gbockus-sf:
closing this PR for now as it'll need some touches to get it ready.  We have two tickets tracking the work so I linked both to this PR for future reference.

PR #4241, opened 2022-06-28
fix: Correctly link errors in problems tab to problem location in file

klewis-sfdc:
### What does this PR do?
Fixes an issue where the URI for the file was built incorrectly, which prevented clicking on the problem in the problems tab from linking to the correct location in the file.

### What issues does this PR fix or reference?
#4194, #4228 @[W-11350093]@

### Functionality Before

https://user-images.githubusercontent.com/46458081/176308232-ac30c395-36e0-40f6-991d-99c040869dcc.mp4

### Functionality After

https://user-images.githubusercontent.com/46458081/176308217-3750e4a8-0447-4811-a143-f70f6cf2c308.mp4

PR #4246, opened 2022-06-29
fix: Build absolute filePath so problems tab links to error location correctly

klewis-sfdc:
### What does this PR do?

A follow up to PR #4241 to fix the fileUri when constructed after a deployment error.

##4228, @W-11350093@


packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -143,3 +144,16 @@ export function handleDeployDiagnostics(
 
   return errorCollection;
 }
+
+export function getAbsoluteFilePath(
```
gbockus-sf, line 148: Should we put this in the same place as where getRootWorkspacePath lives?

packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -143,3 +144,16 @@ export function handleDeployDiagnostics(
...
+  filePath: string | undefined,
+  workspacePath: string
```
gbockus-sf, line 150: should this default to the root?

packages/salesforcedx-vscode-core/test/vscode-integration/diagnostics/index.test.ts
```diff
@@ -246,4 +247,22 @@ describe('Diagnostics', () => {
...
+    const filePath = `${workspacePath}/src/classes/Testing.cls`;
+    const asoluteFilePath = getAbsoluteFilePath(filePath, workspacePath);
+    expect(asoluteFilePath).to.equal(filePath);
+  });
+
+  it('Should use the workspace path as fileUri when filePath is undefined', () => {
+    const filePath = undefined;
+    const asoluteFilePath = getAbsoluteFilePath(filePath, workspacePath);
+    expect(asoluteFilePath).to.equal(workspacePath);
+  });
```
gbockus-sf, line 267: I checked out the failing test and I'd just update it to the following: 
```
 const [fileDesc, reportedError] = setDiagnosticsStub.getCall(
            index
          ).args;
          const expectedFile = vscode.Uri.file(
            getAbsoluteFilePath(row.filePath, getRootWorkspacePath())
          );
          expect(fileDesc).to.deep.equal(expectedFile);
          expect(reportedError).to.deep.equal([
            {
              message: row.error,
              range: new vscode.Range(
                row.lineNumber - 1,
                row.columnNumber - 1,
                row.lineNumber - 1,
                row.columnNumber - 1
              ),
              source: row.type
            }
          ]);
``` lines 457-476

It separates the the assertions so we don't have to run it to know which on is failing.  A very uningration solution as it pulls the utility into the test, but it was the best path I could come up with.

gbockus-sf (approved):
Passed my QE: 
✅ Verified navigation from problem to file when file is open 
✅ Verified navigation from problem to file when file is closed
✅ Verified navigation from problem to file when file is open but not the active editor
✅ Verified when problem is generated during active editing 
✅ Verified navigation from problem when the problem is reported after deploy from navigation
✅ Verified navigation from problem to file deploy via Deploy this source to Org

:shipit:

packages/salesforcedx-vscode-core/test/vscode-integration/diagnostics/index.test.ts
klewis-sfdc: I updated the var names in this method to more closely reflect the params of the method where they are defined.

PR #4247, opened 2022-06-30
build: update the circleci config to use different context(s)

gbockus-sf:
### What does this PR do?
Update the circleci build to use the salesforce-ide &  NPM_Salesforce-Releases circleci contexts

@W-11372105@


.circleci/config.yml
```diff
@@ -601,15 +605,17 @@ workflows:
...
           message: Pending Approval for Creation of Release Branch
           channel: 'pdt_releases'
           color: '#0E1111'
       - hold: # Requires manual approval in Circle Ci
           type: approval
       - create-release-branch:
           e: default-executor
-          context: IEAT_keys
+          context:
+            - salesforce-ide
```
gbockus-sf, line 618: I moved IEAT_keys -> salesforce-ide & added the NPM context to the publish jobs.  I think that covers it

RitamAgrawal: Before this name change, were Npm tokens defined in IEAT_keys?

gbockus-sf: yup, now are found in NPM_Salesforce-Releases

randi274: Maybe a silly question @gbockus-sf, but this is a new token, which is expected, right? Do we need to disable the old token once we retire IEAT_keys?

gbockus-sf: The token is owned by the cli team and has already been rolled so the current value in IEAT_keys is already disabled/expired.  Good question though

PR #4281, opened 2022-07-14
chore: attempting to get org list working with core v3 updates

klewis-sfdc:
VSCE will initialize successfully, and the org list should populate correctly in the command palette.    Still working on getting the set command to be handled correctly when switching default org or auth'ing an org.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/constants.ts
+++ b/packages/salesforcedx-vscode-core/src/constants.ts
@@ -16,8 +16,6 @@ export const SFDX_CONFIG_DISABLE_TELEMETRY = 'disableTelemetry';
 export const ENV_SFDX_DISABLE_TELEMETRY = 'SFDX_DISABLE_TELEMETRY';
 export const SFDX_CLI_DOWNLOAD_LINK =
   'https://developer.salesforce.com/tools/sfdxcli';
-export const DEFAULT_USERNAME_KEY = 'target-org';
-export const DEFAULT_DEV_HUB_USERNAME_KEY = 'target-dev-hub';
 export const PKG_ID_PREFIX = '04t';
 
 export const TELEMETRY_GLOBAL_VALUE = 'sfdxTelemetryMessage';
--- a/packages/salesforcedx-vscode-core/src/constants.ts
+++ b/packages/salesforcedx-vscode-core/src/constants.ts
@@ -16,8 +16,8 @@ export const SFDX_CONFIG_DISABLE_TELEMETRY = 'disableTelemetry';
 export const ENV_SFDX_DISABLE_TELEMETRY = 'SFDX_DISABLE_TELEMETRY';
 export const SFDX_CLI_DOWNLOAD_LINK =
   'https://developer.salesforce.com/tools/sfdxcli';
-export const DEFAULT_USERNAME_KEY = 'defaultusername';
-export const DEFAULT_DEV_HUB_USERNAME_KEY = 'defaultdevhubusername';
+export const DEFAULT_USERNAME_KEY = 'target-org';
+export const DEFAULT_DEV_HUB_USERNAME_KEY = 'target-dev-hub';
 export const PKG_ID_PREFIX = '04t';
 
 export const TELEMETRY_GLOBAL_VALUE = 'sfdxTelemetryMessage';
@@ -30,7 +30,8 @@ export const BETA_DEPLOY_RETRIEVE = 'experimental.deployRetrieve';
 export const CONFLICT_DETECTION_ENABLED = 'detectConflictsAtSync';
 export const INTERNAL_DEVELOPMENT_FLAG = 'internal-development';
 export const PUSH_OR_DEPLOY_ON_SAVE_ENABLED = 'push-or-deploy-on-save.enabled';
-export const PUSH_OR_DEPLOY_ON_SAVE_OVERRIDE_CONFLICTS = 'push-or-deploy-on-save.overrideConflictsOnPush';
+export const PUSH_OR_DEPLOY_ON_SAVE_OVERRIDE_CONFLICTS =
+  'push-or-deploy-on-save.overrideConflictsOnPush';
 export const RETRIEVE_TEST_CODE_COVERAGE = 'retrieve-test-code-coverage';
 export const SFDX_CORE_CONFIGURATION_NAME = 'salesforcedx-vscode-core';
 export const SHOW_CLI_SUCCESS_INFO_MSG = 'show-cli-success-msg';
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -54,25 +54,7 @@ export class OrgList implements vscode.Disposable {
...
-          authFile.username
-        );
-        const fileData = readFileSync(filePath, 'utf8');
-        authInfoObjects.push(JSON.parse(fileData));
-      } catch (e) {
-        console.log(e);
-      }
-    }
-    return authInfoObjects;
+    return authFilesArray;
```
gbockus-sf, line 57: :|

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -16,8 +16,8 @@ export const SFDX_CONFIG_DISABLE_TELEMETRY = 'disableTelemetry';
 export const ENV_SFDX_DISABLE_TELEMETRY = 'SFDX_DISABLE_TELEMETRY';
 export const SFDX_CLI_DOWNLOAD_LINK =
   'https://developer.salesforce.com/tools/sfdxcli';
-export const DEFAULT_USERNAME_KEY = 'defaultusername';
-export const DEFAULT_DEV_HUB_USERNAME_KEY = 'defaultdevhubusername';
+export const DEFAULT_USERNAME_KEY = 'target-org';
+export const DEFAULT_DEV_HUB_USERNAME_KEY = 'target-dev-hub';
```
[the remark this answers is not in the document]
klewis-sfdc: yeah I think so - should just be able to import and use them.  I'll take a look

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -43,7 +43,7 @@ export class OrgList implements vscode.Disposable {
   }
 
   public displayDefaultUsername(defaultUsernameorAlias?: string) {
-    if (!defaultUsernameorAlias) {
+    if (defaultUsernameorAlias) {
```
klewis-sfdc, line 46: Not sure why this worked the way it did before, but flipping this now has the org list showing up in command palette, shows the default org in the bottom menu bar, and appears to successfully be running "Set a default org" command.

gbockus-sf: B/c somebody screwed up removing the deprecated function :mybad: 
`  if (!isNullOrUndefined(defaultUsernameorAlias)) {` was the original setting.

Your change totally makes sense. 👍

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
gbockus-sf: Our own @AnanyaJha made this change (well a version of it) [here](https://github.com/forcedotcom/salesforcedx-vscode/pull/890/files]

@AnanyaJha any memory about why we'd need to get the list of authorizations from global from core, then read the json files for each of those vs just using what they provide for us?

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -132,11 +136,18 @@ describe('OrgAuthInfo', () => {
       ).to.equal(true);
     });
 
-    it('should use default username/alias when invoked without argument', async () => {
+    it.only('should use default username/alias when invoked without argument', async () => {
```
gbockus-sf, line 139: don't forget to remove the `.only`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
gbockus-sf: Ok so I started feeling real paranoid about this so I looked at the types returned from each of these: 
```
export declare type OrgAuthorization = { // new structure
    orgId: string;
    username: string;
    oauthMethod: 'jwt' | 'web' | 'token' | 'unknown';
    aliases: Nullable<string[]>;
    configs: Nullable<string[]>;
    isScratchOrg?: boolean;
    isDevHub?: boolean;
    isSandbox?: boolean;
    instanceUrl?: string;
    accessToken?: string;
    error?: string;
    isExpired: boolean | 'unknown';
};
export interface FileInfo { // Old structure
    scratchAdminUsername?: string;  // Doesn't exist but isScratchOrg does
    isDevHub?: boolean;
    username: string;
    devHubUsername?: string; // doesn't exist 
    expirationDate?: string;  // Doesn't exist but isExpired does
  }

Comment (gbockus-sf) [reply packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts:57]: this kind of blew my mind...I was like...these types don't match...how is the even compiling?

Comment (gbockus-sf) [reply packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts:57]: I guess it's b/c the required properties for the two interfaces overlap

Comment (gbockus-sf) [reply packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts:57]: previously line 70 did a JSON.parse which returns an any so that made sense (not logically but for compilation)

**Review by gbockus-sf — APPROVED** [review]
I'm cool with merging this to get us in a better state.  I think we'll need to follow up with more changes to make the filtering work as expected.

File: packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts [hunk 5]
```diff
@@ -54,25 +54,7 @@ export class OrgList implements vscode.Disposable {
...
-        authInfoObjects.push(JSON.parse(fileData));
-      } catch (e) {
-        console.log(e);
-      }
-    }
-    return authInfoObjects;
+    return authFilesArray;
   }
 
   public async filterAuthInfo(authInfoObjects: FileInfo[]) {
```
gbockus-sf, line 60: Note: there is a filter option on `listAllAuthorizations` we cold utilize if we wanted.

gbockus-sf: just documenting that on line 80 below from the old PR `That field indicates that the file represents a new user, other than the default admin, within the org.` We should probably verify this is still valid

packages/salesforcedx-vscode-core/src/constants.ts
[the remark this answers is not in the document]
klewis-sfdc: imported them directly, removed the constants

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: FYI, I Took another look at this and I was able to make a much smaller change that got it working without changing the shape of the returned data.  Had to point this function to the correct directory where these auth files are stored, which might have shifted slightly with the upgrade to core v3.  That PR is much cleaner.  See the change here: https://github.com/forcedotcom/salesforcedx-vscode/pull/4290/files#diff-2b72c15ab3c2e3d66f269772e80aed9902d2d85d9acb2a462116d03541484d4fR66-R67

klewis-sfdc:
Closing in favor of #4290 which enables the same functionality with a smaller set of changes.

PR #4283, opened 2022-07-15
fix: Update build to improve build time for corev3

gbockus-sf:
### What does this PR do?
This PR was a result of much longer build times post corev3 update. 

- Alter how we cache resources on windows in circle to allow for faster npm installs. 
- Update to lerna5 (MEM buffer exceeded error be gone!)
- Remove coverage call in utils package for unit:test to get past issue on leran5
 
@W-11384712@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ commands:
           shell: bash.exe
           command: |
             npm config set scripts-prepend-node-path true
-            npm install
+            npm ci
             npm install -g sfdx-cli@<< pipeline.parameters.cli-version >>
             npm install -g junit-merge
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,12 +61,10 @@ commands:
     description: 'Building and archiving extension artifacts.'
     steps:
       - checkout
-      - restore_dependency_cache
       - install
       - compile
       - package-extensions
       - stage-and-store-artifacts
-      - save_dependency_cache
 
   build-win:
     description: 'Build for Windows.'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ commands:
           shell: bash.exe
           command: |
             npm config set scripts-prepend-node-path true
-            npm ci
+            npm list @heroku/functions-core || npm install
             npm install -g sfdx-cli@<< pipeline.parameters.cli-version >>
             npm install -g junit-merge
       - run:
@@ -138,7 +138,7 @@ commands:
   install:
     description: 'Installing dependencies'
     steps:
-      - run: npm ci
+      - run: npm install
       - run: npm --version
       - run: node --version
 
@@ -222,7 +222,26 @@ commands:
           key: << parameters.base-cache-path >>{{ checksum "package-lock.json" }}
           paths:
             - << parameters.path-to-save >>
-            - C:\Users\circleci\AppData\Local\npm-cache\
+            - node_modules
+            - packages/salesforcedx-apex-debugger/node_modules
+            - packages/salesforcedx-visualforce-language-server/node_modules
+            - packages/salesforcedx-vscode-apex-replay-debugger/node_modules
+            - packages/salesforcedx-vscode-soql/node_modules
+            - packages/salesforcedx-apex-replay-debugger/node_modules
+            - packages/salesforcedx-visualforce-markup-language-server/node_modules
+            - packages/salesforcedx-vscode-core/node_modules
+            - packages/salesforcedx-vscode-visualforce/node_modules
+            - packages/salesforcedx-sobjects-faux-generator/node_modules
+            - packages/salesforcedx-vscode/node_modules
+            - packages/salesforcedx-vscode-expanded/node_modules
+            - packages/salesforcedx-webview-ui/node_modules
+            - packages/salesforcedx-test-utils-vscode/node_modules
+            - packages/salesforcedx-vscode-apex/node_modules
+            - packages/salesforcedx-vscode-lightning/node_modules
+            - packages/system-tests/node_modules
+            - packages/salesforcedx-utils-vscode/node_modules
+            - packages/salesforcedx-vscode-apex-debugger/node_modules
+            - packages/salesforcedx-vscode-lwc/node_modules
 
   ssh-config:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ parameters:
   # Constants
   cache-version:
     type: string
-    default: 'dependency-cache-v4-'
+    default: 'dependency-cache-v5-'
   domain-retry:
     description: 'SFDX Domain Retry Wait in Seconds'
     type: integer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,6 +308,7 @@ jobs:
     executor:
       name: default-executor
       working-dir: ~/project
+      resource_class: large
     environment:
       - SFDX_URL_LINUX: https://developer.salesforce.com/media/salesforce-cli/sfdx-linux-amd64.tar.xz
       - SFDX_AUTOUPDATE_DISABLE: true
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -28,10 +28,13 @@ parameters:
...
     # Need to migrate from legacy CircleCI image (W-11027467)
     default: '16.13.0'
+  npm-version:
+    type: string
+    default: '8.12.1'
   # Constants
   cache-version:
     type: string
-    default: 'dependency-cache-v4-'
+    default: 'dependency-cache-v5-'
```
gbockus-sf, line 37: bump the cache default name since there have been changes to what is cached.

.circleci/config.yml
```diff
@@ -58,12 +61,10 @@ commands:
     description: 'Building and archiving extension artifacts.'
     steps:
       - checkout
-      - restore_dependency_cache
       - install
       - compile
       - package-extensions
       - stage-and-store-artifacts
-      - save_dependency_cache
```
gbockus-sf, line 66: for linux we are only caching the top level node_modules dir.  This is providing very little value for this particular job so I just removed it.

.circleci/config.yml
```diff
@@ -78,13 +79,18 @@ commands:
           command: |
             nvm install $NODE_VERSION
             nvm use $NODE_VERSION
-            nvm install-latest-npm
+      - run:
+          name: 'Install npm'
+          shell: bash.exe
+          command: (set -x; npm install -g npm@<< pipeline.parameters.npm-version >>)
```
gbockus-sf, line 85: set -x prints out the command name

.circleci/config.yml
```diff
@@ -78,13 +79,18 @@ commands:
...
       - run:
           name: 'Install dependent modules'
           shell: bash.exe
+          # Only run the npm install if the npm list command fails to find the module locally installed
+          # Note I picked a random module.
           command: |
             npm config set scripts-prepend-node-path true
-            npm install
+            npm list @heroku/functions-core || npm install
```
gbockus-sf, line 93: this change results in only running npm install on windows it the node_modules weren't previously cached.

.circleci/config.yml
```diff
@@ -138,7 +144,7 @@ commands:
   install:
     description: 'Installing dependencies'
     steps:
-      - run: npm ci
+      - run: npm install
```
gbockus-sf, line 147: install was faster then ci in this case.

.circleci/config.yml
```diff
@@ -200,30 +206,41 @@ commands:
...
-    # the package.json instead of the package-lock.json for our checksum.
     parameters:
       base-cache-path:
         type: string
         default: << pipeline.parameters.cache-version >>
     steps:
       - restore_cache:
           keys:
-            - << parameters.base-cache-path >>{{ checksum "package.json" }}
+            - << parameters.base-cache-path >>{{ checksum "package-lock.json" }}
```
gbockus-sf, line 216: so we were trigging caches off the package.json file vs package-lock.json.  I'm pretty sure this was a remnant of when we didn't have a package-lock.json file and didn't make sense.  I updated it to base the cache off the package-lock.json file.

.circleci/config.yml
```diff
@@ -200,30 +206,41 @@ commands:
...
+            - packages/salesforcedx-vscode-core/node_modules
+            - packages/salesforcedx-vscode-visualforce/node_modules
+            - packages/salesforcedx-sobjects-faux-generator/node_modules
+            - packages/salesforcedx-test-utils-vscode/node_modules
+            - packages/salesforcedx-vscode-apex/node_modules
+            - packages/salesforcedx-vscode-lightning/node_modules
+            - packages/system-tests/node_modules
+            - packages/salesforcedx-utils-vscode/node_modules
+            - packages/salesforcedx-vscode-apex-debugger/node_modules
+            - packages/salesforcedx-vscode-lwc/node_modules
```
gbockus-sf, line 243: cache all the node_modules directories.  I saw this result in a bump from 1 m to 5 min to restore the cache, but it reduced the npm install time from 12+ m to 4

.circleci/config.yml
```diff
@@ -284,6 +301,7 @@ executors:
     docker:
       - image: << parameters.image >>
     working_directory: << parameters.working-dir >>
+    resource_class: large
```
gbockus-sf, line 304: we were running at 100% of cpu on the linux build when generating artifacts so I bumped it up...we still hit 100% cpu, but for like 50% of the time we were previously and get more memory free.

.circleci/config.yml
```diff
@@ -521,8 +539,8 @@ jobs:
       - restore_dependency_cache
       - gh-config
       - run:
-          name: 'Install npm 8.12'
-          command: sudo npm install -g npm@8.12.1
+          name: 'Install npm'
+          command: sudo npm install -g npm@<< pipeline.parameters.npm-version >>
```
gbockus-sf, line 543: added a config var for the npm version.

package.json
```diff
@@ -16,7 +16,7 @@
     "@commitlint/cli": "^15.0.0",
     "@commitlint/config-conventional": "^15.0.0",
     "@monorepo-utils/workspaces-to-typescript-project-references": "2.7.6",
-    "@octokit/core": "3.6.0",
+    "@octokit/core": "4.0.4",
```
gbockus-sf, line 19: it started complaining about this dep which was our issue in publishing last week and this was the suggested fix. :shrug:

package.json
```diff
@@ -30,7 +30,7 @@
     "husky": "1.3.1",
     "istanbul": "1.1.0-alpha.1",
     "jsdom": "^15.1.1",
-    "lerna": "3.22.1",
+    "lerna": "5.1.8",
```
gbockus-sf, line 33: I haven't see a max buffer issue since I did this update.  Probably should also update the global version and the docs.

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -34,7 +34,7 @@
   "scripts": {
     "compile": "webpack",
     "webpack": "webpack",
-    "lint": "tslint --project . -e **/src/data/minSObjects.json",
+    "lint": "tslint --project . -e **/minSObjects.json",
```
gbockus-sf, line 37: weird tslint change with the lerna update

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -51,8 +51,9 @@
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "test": "npm run test:unit",
-    "test:unit": "node ./node_modules/nyc/bin/nyc.js ./node_modules/mocha/bin/_mocha --recursive out/test/unit/**/*.js --timeout 20000 --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaUnitTestsConfig.json",
-    "coverage": "./node_modules/.bin/nyc npm test"
+    "test:unit": "mocha --recursive out/test/unit/**/*.js --timeout 20000 --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaUnitTestsConfig.json",
+    "test:unit:coverage": "node ./node_modules/nyc/bin/nyc.js ./node_modules/mocha/bin/_mocha --recursive out/test/unit/**/*.js --timeout 20000 --reporter mocha-multi-reporters --reporter-options configFile=../../config/mochaUnitTestsConfig.json",
+    "coverage": "./node_modules/.bin/nyc npm test:unit:coverage"
```
gbockus-sf, line 56: So with the new version of lerna this was always encountering an error when using nyc to generate coverage number on utils due to some webpack issue.  I just removed the coverage generation for the default `test:unit` task and added a specific script for having coverage locally.  Note this was only occurring on the windows build.

.circleci/config.yml
gbockus-sf: the module isn't present if the cache isn't loaded which occurs when there is a change to package-lock.json

klewis-sfdc (approved):
👍🏽  Thank you!

gbockus-sf:
lerna updated separately.  With conversion to GH actions for CI circle improvements no longer needed.

PR #4306, opened 2022-07-21
test: rearrange before and after

randi274:
### What does this PR do?
We were receiving an unruly number of test failures from this one particular test, which has caused issues for a year +. Swapping the beforeEach / afterEach was a previous suggestion that we hadn't implemented yet, but based on [3 test runs ](https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode?branch=randi%2Fflapper-integration-fix)seems to have resolved the issue.

### Functionality Before
Receiving the following error `jsconfig Test Suite “after each” hook for “Should not update jsconfig.json when a component is saved”.` [regularly](https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/9544/workflows/eb559068-7851-4547-8d14-643a21f7af35/jobs/36228/tests): `Error: EBUSY: resource busy or locked, rmdir 'c:\Users\circleci\project\packages\system-tests\assets\lwc-recipes\force-app\main\default\lwc\testComponent'`.

### Functionality After
Error should not be received again 🤞


packages/salesforcedx-vscode-lwc/test/vscode-integration/jsconfig.test.ts
```diff
@@ -51,6 +45,12 @@ describe('jsconfig Test Suite', () => {
     }
   });
 
+  beforeEach(async () => {
```
jeffb-sfdc, line 48: Strange that the `beforeEach()` needs to come after the `afterEach()` (the order shouldn't matter).  Could you add a comment, letting the reader know why they are in that order, and warn them to not move them?  I could see someone seeing this one day and changing the order, thinking they are cleaning up the code.

jeffb-sfdc (approved):

Issue #941, opened 2017-08-28
Clear output log before executing new command (with preference)


muenzpraeger:
### Summary
With the frequent usage of the SFDX extension the output command can become quite overloaded. Especially when running multiple commands it's sometimes not that easy to see where the last command started.

### Expected result
The output log should be cleared automatically when a new command gets executed. This should be configurable (default: not to clear).

jefersonchaves:
:thumbsup:

And I agree that clearing is good and toggling it off by default is great.

OneMadhu:
I too agree with this feature this would be really helpful;  As suggested by @cwallsfdc also consider demarcation in case of setting turned off.

OneMadhu:
I am ready to contribute on this feature please let me know.

PR #4318, opened 2022-07-25
Add preference clear output channel

dehru:
### What does this PR do?

This PR adds a user/workspace preference and changes the preference category they display under in the preferences UI.  When set, the output channel is cleared between each common command.   There are a few commands that do not output to the channel ( Like Create Soql Query ) and those do not clear the channel.

#941, @W-10288504@

### Functionality Before

Old and New Preference Shown Below

Screen Shot 2022-07-21 at 3 31 36 PM

### Functionality After

https://user-images.githubusercontent.com/599418/180885264-b1a61c1b-42f8-4e73-86d2-e9d7f9238fbe.mov

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/channelService.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/channelService.ts
@@ -9,6 +9,7 @@ import stripAnsi from 'strip-ansi';
 import { OutputChannel, window } from 'vscode';
 import { CommandExecution } from '../cli';
 import { nls } from '../messages';
+import { SfdxSettingsService } from '../settings';
 
 export class ChannelService {
   private readonly channel: OutputChannel;
@@ -37,6 +38,9 @@ export class ChannelService {
   }
 
   public streamCommandStartStop(execution: CommandExecution) {
+    if (SfdxSettingsService.getEnableClearOutputBeforeEachCommand()) {
+        this.channel.clear();
+    }
     this.channel.append(nls.localize('channel_starting_message'));
     this.channel.appendLine(execution.command.toString());
     this.channel.appendLine('');
@@ -100,5 +104,9 @@ export class ChannelService {
 
   public appendLine(text: string) {
     this.channel.appendLine(text);
+  }
+
+  public clear() {
+    this.channel.clear();
   }
 }
--- a/packages/salesforcedx-utils-vscode/src/commands/channelService.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/channelService.ts
@@ -39,7 +39,7 @@ export class ChannelService {
 
   public streamCommandStartStop(execution: CommandExecution) {
     if (SfdxSettingsService.getEnableClearOutputBeforeEachCommand()) {
-        this.channel.clear();
+        this.clear();
     }
     this.channel.append(nls.localize('channel_starting_message'));
     this.channel.appendLine(execution.command.toString());
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -14,6 +14,7 @@ import {
   TelemetryService
 } from '../index';
 import { nls } from '../messages';
+import { SfdxSettingsService } from '../settings';
 import { CommandletExecutor, ContinueResponse } from '../types';
 import { getRootWorkspacePath } from '../workspaces';
 import { ChannelService } from './channelService';
@@ -161,6 +162,9 @@ export abstract class LibraryCommandletExecutor<T>
     const startTime = process.hrtime();
     const channelService = new ChannelService(this.outputChannel);
     const telemetryService = TelemetryService.getInstance();
+    if (SfdxSettingsService.getEnableClearOutputBeforeEachCommand()) {
+      channelService.clear();
+    }
 
     channelService.showCommandWithTimestamp(
       `${nls.localize('channel_starting_message')}${this.executionName}\n`
--- a/packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
@@ -7,6 +7,7 @@
 import { Observable } from 'rxjs/Observable';
 import * as vscode from 'vscode';
 import { CommandExecution } from '../cli';
+import { SFDX_CORE_CONFIGURATION_NAME } from '../constants';
 import { nls } from '../messages';
 import { ChannelService } from './index';
 
@@ -112,7 +113,7 @@ export class NotificationService {
       executionName
     );
     const showCLISuccessMsg = vscode.workspace
-      .getConfiguration('salesforcedx-vscode-core')
+      .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
       .get<boolean>('show-cli-success-msg', true);
     if (showCLISuccessMsg) {
       const showButtonText = nls.localize('notification_show_button_text');
@@ -129,7 +130,7 @@ export class NotificationService {
           channelService.showChannelOutput();
         } else if (selection === showOnlyStatusBarButtonText) {
           await vscode.workspace
-            .getConfiguration('salesforcedx-vscode-core')
+            .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
             .update('show-cli-success-msg', false);
         }
       }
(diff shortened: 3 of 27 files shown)


packages/salesforcedx-utils-vscode/src/commands/channelService.ts
```diff
@@ -37,6 +38,9 @@ export class ChannelService {
   }
 
   public streamCommandStartStop(execution: CommandExecution) {
+    if (SfdxSettingsService.getEnableClearOutputBeforeEachCommand()) {
```
dehru, line 41: The channel service is my first choice to attempt to handle this preference in one central location.  Unfortunately, the way the channel service is used throughout the app is very inconsistent.  This "streamCommandStartStop()" is used by some of the abstract executors ( SfdxCommandletExecutor, LibraryExectutor ) that are exported from salesforcedx-utils-vscode.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -161,6 +162,9 @@ export abstract class LibraryCommandletExecutor<T>
     const startTime = process.hrtime();
     const channelService = new ChannelService(this.outputChannel);
     const telemetryService = TelemetryService.getInstance();
+    if (SfdxSettingsService.getEnableClearOutputBeforeEachCommand()) {
```
dehru, line 165: The LibraryCommandletExecutor is exported from `salesforcedx-utils-vscode` and is another place that I found to check and clear the output log.

packages/salesforcedx-utils-vscode/src/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const SETTING_CLEAR_OUTPUT_TAB = 'clearOutputTab';
```
dehru, line 1: These are now all exported by salesforcedx-utils-vscode and consumed in other modules so that the string 'salesforcedx-vscode-core' is not in 20 places.

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -165,6 +166,9 @@ export class SfdxCommandlet<T> {
   }
 
   public async run(): Promise<void> {
+    if (sfdxCoreSettings.getEnableClearOutputBeforeEachCommand()) {
```
dehru, line 169: Adding this to the run() method here basically adds this feature to all the commands that extend `SfdxCommandlet` in the `salesforcedx-vscode-core` extension.  I wish there was an elegant single cut-point elsewhere in the `salesforcedx-utils-vscode` module like this, but there was not.

packages/salesforcedx-utils-vscode/src/commands/channelService.ts
```diff
@@ -37,6 +38,9 @@ export class ChannelService {
...
+        this.channel.clear();
```
randi274, line 42: Should this just be calling `clear()`?

packages/salesforcedx-utils-vscode/src/constants.ts
```diff
@@ -0,0 +1,3 @@
...
+export const SFDX_CORE_CONFIGURATION_NAME = 'salesforcedx-vscode-core';
```
randi274, line 2: What's your thought here having these two identical strings with different names?

randi274 (approved):
Looking great, and look at those lovely unit tests! 😍 Just a few quick qs. Good for me otherwise pending QA

packages/salesforcedx-utils-vscode/src/commands/channelService.ts
dehru: Yes, good catch

packages/salesforcedx-utils-vscode/src/constants.ts
dehru: I decided that since they represent 2 different things, that we should have 2 constants, even though they are the same string.  

I wanted anyone changing these in the future to be more aware of what this constant represents.

packages/salesforcedx-utils-vscode/src/commands/sfdxCommandlet.ts
```diff
@@ -5,6 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import * as vscode from 'vscode';
+import { SfdxSettingsService } from '../settings';
```
jeffb-sfdc: I don't think this needs to be imported

packages/salesforcedx-utils-vscode/src/commands/notificationService.ts
```diff
@@ -112,7 +113,7 @@ export class NotificationService {
       executionName
     );
     const showCLISuccessMsg = vscode.workspace
-      .getConfiguration('salesforcedx-vscode-core')
+      .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
```
jeffb-sfdc, line 116: Since
```
vscode.workspace
      .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
```

is called twice, how about changing to:

```
const coreConfigurationName = vscode.workspace.getConfiguration(SFDX_CORE_CONFIGURATION_NAME);
const showCLISuccessMsg = coreConfigurationName.get<boolean>('show-cli-success-msg', true);
```

..and then using `coreConfigurationName` on line 132 & 133.

packages/salesforcedx-utils-vscode/src/index.ts
```diff
@@ -41,3 +41,4 @@ export {
   getRootWorkspacePath,
   getRootWorkspaceSfdxPath
 } from './workspaces';
+export { SETTING_CLEAR_OUTPUT_TAB, SFDX_CORE_CONFIGURATION_NAME, SFDX_CORE_EXTENSION_NAME } from './constants';
```
jeffb-sfdc, line 44: the imports need to be alphabetized.  i'm surprised the linter didn't catch this.

Since this is being imported from `./constants`, it should be moved to line 25, before the import from `./context/...`

packages/salesforcedx-vscode-apex-replay-debugger/test/vscode-integration/commands/quickLaunch.test.ts
```diff
@@ -12,6 +12,7 @@ import { MockTestOrgData, testSetup } from '@salesforce/core/lib/testSetup';
 import { notificationService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
 import { TraceFlags } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
 import * as utils from '@salesforce/salesforcedx-utils-vscode/out/src/index';
+import { SFDX_CORE_CONFIGURATION_NAME } from '@salesforce/salesforcedx-utils-vscode/out/src/index';
```
jeffb-sfdc, line 15: since `@salesforce/salesforcedx-utils-vscode/out/src/index` is already being imported, can line 15 be removed and on line 54, use `utils.SFDX_CORE_CONFIGURATION_NAME`?

packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -75,6 +74,9 @@ export class SfdxCoreSettings {
     return this.getConfigValue(CONFLICT_DETECTION_ENABLED, false);
   }
 
+  public getEnableClearOutputBeforeEachCommand(): boolean {
+    return this.getConfigValue(SETTING_CLEAR_OUTPUT_TAB, false);
+  }
```
jeffb-sfdc, line 79: not pick: please add an extra carriage return after the closing curly brace.

jeffb-sfdc (approved):

randi274:
QA Notes:
✅  Salesforce Feature Previews renamed to Salesforce Core Configuration
✅  New setting "Clear Output Tab" added to top of Core Config
✅  When setting is enabled for the user, commands are cleared between each output
✅  When setting is disabled for the user, commands remain logged
✅  When setting is enabled for the workspace, commands are cleared between each output
✅  When setting is disabled for the workspace, commands remain logged
✅  Commands that don't output to the channel, like Create SOQL Query, do not clear the channel 

Notes:
- Like all settings, if the VS Code Setting for Workspace is modified but then set to false, the setting will be overridden as false even if it is set to true for the user. It's best to limit to one setting file or the other.
- It's also very sensitive, and will clear the output at the start of any command (like Create Apex Class)

PR #4341, opened 2022-08-02
fix: update for checkForExpiredOrgs() to use OrgAuthorization

jeffb-sfdc:
### What does this PR do?
This PR updates `checkForExpiredOrgs()` to use OrgAuthorization, which was a change with a recent merge.

No GitHub issue for this PR

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/index.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/index.ts
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-export { isNullOrUndefined, extractJsonObject } from './utils';
+export { asyncFilter, isNullOrUndefined, extractJsonObject } from './utils';
 export {
   isAlphaNumString,
   isInteger,
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -14,11 +14,16 @@ export function isNullOrUndefined(object: any): object is null | undefined {
 }
 
 export function extractJsonObject(str: string): any {
-
   const jsonString = str.substring(
     str.indexOf('{'),
     str.lastIndexOf('}') + 1
   );
 
   return JSON.parse(jsonString);
+}
+
+export async function asyncFilter<T>(arr: T[], predicate: any) {
+	const results = await Promise.all(arr.map(predicate));
+
+	return arr.filter((_v, index) => results[index]);
 }
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -19,8 +19,8 @@ export function extractJsonObject(str: string): any {
   return JSON.parse(jsonString);
 }
 
-export async function asyncFilter<T>(arr: T[], predicate: any) {
-  const results = await Promise.all(arr.map(predicate));
+export async function asyncFilter<T>(arr: T[], callback: (value: T, index: number, array: T[]) => unknown) {
+  const results = await Promise.all(arr.map(callback));
 
   return arr.filter((_v, index) => results[index]);
 }
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -14,16 +14,13 @@ export function isNullOrUndefined(object: any): object is null | undefined {
 }
 
 export function extractJsonObject(str: string): any {
-  const jsonString = str.substring(
-    str.indexOf('{'),
-    str.lastIndexOf('}') + 1
-  );
+  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
 
   return JSON.parse(jsonString);
 }
 
 export async function asyncFilter<T>(arr: T[], predicate: any) {
-	const results = await Promise.all(arr.map(predicate));
+  const results = await Promise.all(arr.map(predicate));
 
-	return arr.filter((_v, index) => results[index]);
+  return arr.filter((_v, index) => results[index]);
 }
--- a/packages/salesforcedx-vscode-core/src/util/metadataDictionary.ts
+++ b/packages/salesforcedx-vscode-core/src/util/metadataDictionary.ts
@@ -4,7 +4,6 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { LocalComponent } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
 import { PathStrategyFactory, SourcePathStrategy } from '../commands/util';
 
 const DEFINITIONS: { [key: string]: MetadataInfo } = {
--- a/packages/salesforcedx-vscode-core/src/util/orgUtil.ts
+++ b/packages/salesforcedx-vscode-core/src/util/orgUtil.ts
@@ -5,8 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { AuthFields, AuthInfo, OrgAuthorization } from '@salesforce/core';
-import { asyncFilter } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
+import { AuthFields, AuthInfo } from '@salesforce/core';
 import { channelService } from '../channels';
 import { nls } from '../messages';
 import { notificationService } from '../notifications';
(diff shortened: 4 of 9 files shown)


packages/salesforcedx-utils-vscode/src/helpers/index.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-export { isNullOrUndefined, extractJsonObject } from './utils';
+export { asyncFilter, isNullOrUndefined, extractJsonObject } from './utils';
```
jeffb-sfdc, line 8: Added `asyncFilter` (which BTW, calling `map()` or `filter()` with async functions in them is a real PITA)

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +14,16 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
-
   const jsonString = str.substring(
     str.indexOf('{'),
     str.lastIndexOf('}') + 1
   );
 
   return JSON.parse(jsonString);
 }
+
+export async function asyncFilter<T>(arr: T[], predicate: any) {
```
jeffb-sfdc, line 25: Added so `filter()` can be called with async function within it.

packages/salesforcedx-vscode-core/src/util/metadataDictionary.ts
```diff
@@ -4,7 +4,6 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { LocalComponent } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
```
jeffb-sfdc, line 7: not used so it was removed

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -27,6 +30,20 @@ export async function setUpOrgExpirationWatcher(orgList: OrgList) {
   */
 }
 
+export async function getAuthFieldsFor(username: string): Promise<AuthFields> {
```
jeffb-sfdc, line 33: @ken, I need to consolidate these two functions wit what's in OrgList.ts.  I'll probable move them out of OrgList and move into a utils file.

klewis-sfdc: Sounds good - utils/authUtil?

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
randi274: Love it - great addition!

randi274 (approved):
LGTM 👍

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: It isn't being used anywhere yet right? We should add a few units tests to both test and document usage via test. That can def be a follow on PR

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +14,13 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
-
-  const jsonString = str.substring(
-    str.indexOf('{'),
-    str.lastIndexOf('}') + 1
-  );
+  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
 
   return JSON.parse(jsonString);
 }
+
```
gbockus-sf, line 21: Should the type for predicate below be the same as the param for Array.map? `callbackfn: (value: T, index: number, array: T[]) => unknown`

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -40,24 +57,26 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
       daysUntilExpiration.getDate() + daysBeforeExpire
     );
 
-    const authInfoObjects = await orgList.getAuthInfoObjects();
-    if (!authInfoObjects) {
+    const orgAuthorizations = await AuthInfo.listAllAuthorizations()
```
gbockus-sf, line 60: missing semi;

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -40,24 +57,26 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
...
+    if (!orgAuthorizations) {
       return;
     }
 
-    const orgsAboutToExpire = authInfoObjects!.filter((authInfoObject: FileInfo) => {
+    const orgsAboutToExpire = await asyncFilter(orgAuthorizations, async (orgAuthorization: OrgAuthorization) => {
```
gbockus-sf, line 65: oh there's asyncFilter being used...ignore my previous q

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -70,28 +89,27 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
...
-    const formattedOrgsToDisplay = orgsAboutToExpire.map((orgAboutToExpire: any) => {
-      const alias = aliases.getKeysByValue(orgAboutToExpire.username);
-      const aliasName = alias.length > 0
-        ? alias.toString()
+    const formattedOrgsToDisplay = (await Promise.all(orgsAboutToExpire.map(async orgAboutToExpire => {
+      const aliasName = orgAboutToExpire.aliases && orgAboutToExpire.aliases.length > 0
+        ? orgAboutToExpire.aliases[0]
         : orgAboutToExpire.username;
 
+      const authFields = await getAuthFieldsFor(orgAboutToExpire.username);
```
gbockus-sf, line 97: seems like a bummer we make this call twice...once in the fiter then again in the formatter.  can we cache this data or add it to the org for processing?

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -70,28 +89,27 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
...
+
       return nls.localize(
         'pending_org_expiration_expires_on_message',
         aliasName,
-        orgAboutToExpire.expirationDate
+        authFields.expirationDate
       );
-    }).join('\n\n');
+    }))).join('\n\n');
```
gbockus-sf, line 104: this formatting code smells like it should be a separate function.  maybe a follow on sepearate ticket.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: If that causes issues I'd think we probably need a code comment calling out how to use predicate

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: found it in the orgUtil

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: yep

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: done (and thanks!)

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: I thought about this when developing this change.  The issue is that (IMHO) caching this would convolute the code.  Let's discuss at standup today.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
jeffb-sfdc: done

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: I changed this to a for loop

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: looking at the code, it's been reduced a bit, and I don't think moving the code in the `for()` loop into a separate function is needed now.  FWIW, I have unit tests in packages/salesforcedx-vscode-core/test/vscode-integration/util/orgUtil.test.ts which cover each of the code branches in there ("if isDevHub is true", "if expiration date < today", etc...)

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -40,58 +41,61 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
...
-    const authInfoObjects = await orgList.getAuthInfoObjects();
-    if (!authInfoObjects) {
+    const orgAuthorizations = await AuthInfo.listAllAuthorizations();
+    if (!orgAuthorizations) {
       return;
     }
 
-    const orgsAboutToExpire = authInfoObjects!.filter((authInfoObject: FileInfo) => {
+    let results: string[] = [];
+    for (let i=0; i<orgAuthorizations.length; i++) {
```
gbockus-sf, line 50: be sure to run the formatter...there's some space issues with this line.

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
```diff
@@ -40,58 +41,61 @@ export async function checkForExpiredOrgs(orgList: OrgList) {
...
+            authFields.expirationDate
+          )
+        );
+      }
+    }
 
-    if (!orgsAboutToExpire || orgsAboutToExpire.length < 1) {
+    if (results.length < 1) {
       return;
     }
```
gbockus-sf, line 89: this is much easier to grok.  Nice update.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: what did you end up doing for this?

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
jeffb-sfdc: done

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
jeffb-sfdc:
```
export async function asyncFilter<T>(arr: T[], callback: (value: T, index: number, array: T[]) => unknown) {
  const results = await Promise.all(arr.map(callback));

  return arr.filter((_v, index) => results[index]);
}
```

packages/salesforcedx-vscode-core/src/util/orgUtil.ts
klewis-sfdc: Agreed!

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +14,13 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
+export async function asyncFilter<T>(arr: T[], callback: (value: T, index: number, array: T[]) => unknown) {
```
klewis-sfdc, line 22: Should this be removed now that it is not used?

klewis-sfdc (approved):
Looks good to me with the one question of whether to remove asyncFilter.  Thanks for these updates, Jeff!

PR #4343, opened 2022-08-03
Make orgList use OrgAuthorizations

klewis-sfdc:
### What does this PR do?
Updates the orgList module to use core apis over reading from .sfdx files directly.  The orgList module is primarily used to populate the list of authorized orgs for display in the command palette. 

@W-11464626@

### Functionality Before
orgList is generated by reading auth files in .sfdx directory.

### Functionality After
orgList is generated by using core APIs.  No observable changes in behavior.


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
     }
   }
 
-  public async getAuthInfoObjects() {
-    const authFilesArray = await AuthInfo.listAllAuthorizations().catch(
+  public async getOrgAuthorizations(): Promise<OrgAuthorization[] | null> {
```
gbockus-sf, line 51: dumb q: was the return type inferred before and we just go explicit about it

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
...
+    for (const orgAuth of orgAuthorizations) {
+      const authFields: AuthFields = await this.getAuthFieldsFor(
+        orgAuth.username
+      );
+      if (authFields?.scratchAdminUsername) {
+        // non-Admin scratch org users
+        continue;
+      }
+      if (
+        authFields?.devHubUsername &&
```
gbockus-sf, line 92: is the `?` needed here?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
...
+        authFields.devHubUsername !== defaultDevHubUsername
+      ) {
+        // scratch orgs parented by other (non-default) devHub orgs
```
gbockus-sf, line 95: these comments are so _awesome_

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -121,11 +116,15 @@ export class OrgList implements vscode.Disposable {
   }
 
   public async updateOrgList() {
-    const authInfoObjects = await this.getAuthInfoObjects();
-    if (!authInfoObjects) {
+    const orgAuthorizations:
+      | OrgAuthorization[]
```
gbockus-sf, line 120: what does that leading `|` do again... is it like a dangling comma for easier diffs?

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
```diff
@@ -66,14 +66,14 @@ describe('orgList Tests', () => {
       readFileStub.callThrough();
       const orgList = new OrgList();
 
-      const authInfoObjects = await orgList.getAuthInfoObjects();
-      if (authInfoObjects) {
-        expect(authInfoObjects.length).to.equal(2);
+      const orgAuthorizations = await orgList.getOrgAuthorizations();
+      if (orgAuthorizations) {
```
gbockus-sf, line 70: I realize this is how the test was written before, but is it clean to just assert that orgAuthorizations is defined vs the if/else
`expect(!!orgAuthorizations).to.equal(true);` note there's probably a better assertion for this

gbockus-sf (approved):
great stuff.  A few nits to consider

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: Nope! authInfo.getFields() seems to always return an instance of AuthFields. Will remove.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: Actually this isn't needed.  I can remove the explicit typing since it is inferred.  Also updating the `if (!orgAuthorizations)` to use .length since an array is always returned here.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
jeffb-sfdc, line 51: Weird that `getOrgAuthorizations()` didn't have its return type defined.  Good catch and thanks for adding.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
...
+  public async getAuthFieldsFor(username: string): Promise<AuthFields> {
+    const authInfo: AuthInfo = await AuthInfo.create({
+      username
+    });
+    return authInfo.getFields();
+  }
 
-    const defaultDevHubUsernameorAlias = await getDefaultDevHubUsernameOrAlias();
+  public async filterAuthInfo(orgAuthorizations: OrgAuthorization[]) {
+    const defaultDevHubUsernameorAlias = await OrgAuthInfo.getDefaultDevHubUsernameOrAlias(
```
jeffb-sfdc, line 71: nit pick: `defaultDevHubUsernameorAlias` should probably be `defaultDevHubUsernameOrAlias`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
...
 
-  public async filterAuthInfo(authInfoObjects: FileInfo[]) {
-    authInfoObjects = authInfoObjects.filter(
-      fileData => !fileData.scratchAdminUsername
-    );
+  public async getAuthFieldsFor(username: string): Promise<AuthFields> {
+    const authInfo: AuthInfo = await AuthInfo.create({
+      username
+    });
+    return authInfo.getFields();
```
jeffb-sfdc, line 67: Please correct me if I'm wrong here, but `getFields()` returns an AuthFields object, not a promise for a AuthFields.

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
klewis-sfdc: Did some refactoring here - changes to be pushed shortly

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,61 @@ export class OrgList implements vscode.Disposable {
...
-        );
-        const fileData = readFileSync(filePath, 'utf8');
-        authInfoObjects.push(JSON.parse(fileData));
-      } catch (e) {
-        console.log(e);
-      }
-    }
-    return authInfoObjects;
+
+    return orgAuthorizations;
```
jeffb-sfdc, line 60: When I hover over `orgAuthorizations`, VSC's intellisense says it's of type "any"

maybe change line 52 to:
```
    const orgAuthorizations: OrgAuthorization[] = await AuthInfo.listAllAuthorizations().catch(
      err => null
    );
```

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
jeffb-sfdc: Nit pick (soap box): I know Prettier is changing the code when autosave and converts the code to that, but lines 119-121 are pretty unreadable to me.  I still don't know what it does.  I'd ever prefer writing this out long hand than what's on 119-121.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -121,11 +116,15 @@ export class OrgList implements vscode.Disposable {
...
-    if (!authInfoObjects) {
+    const orgAuthorizations:
+      | OrgAuthorization[]
+      | null = await this.getOrgAuthorizations();
+    if (!orgAuthorizations) {
       return null;
     }
-    const authUsernameList = await this.filterAuthInfo(authInfoObjects);
+    const authUsernameList = await this.filterAuthInfo(
+      orgAuthorizations as OrgAuthorization[]
```
jeffb-sfdc, line 126: The `as OrgAuthorization[]` isn't needed.

jeffb-sfdc: Also, `updateOrgList()` doesn't define a return type.  How about adding it and changing line 118 to:
`public async updateOrgList(): Promise<string[]>`

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: I removed this catch block - don't love this silent failure - let me know how the new version looks to you

jeffb-sfdc: looks good, thanks!

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,56 @@ export class OrgList implements vscode.Disposable {
...
+    const authInfo: AuthInfo = await AuthInfo.create({
+      username
+    });
+    return authInfo?.getFields();
   }
 
-  public async filterAuthInfo(authInfoObjects: FileInfo[]) {
-    authInfoObjects = authInfoObjects.filter(
-      fileData => !fileData.scratchAdminUsername
+  public async filterAuthInfo(orgAuthorizations: OrgAuthorization[]) {
```
jeffb-sfdc, line 63:
```
public async filterAuthInfo(orgAuthorizations: OrgAuthorization[]): Promise<string[]> {
```

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -121,11 +111,11 @@ export class OrgList implements vscode.Disposable {
   }
 
   public async updateOrgList() {
-    const authInfoObjects = await this.getAuthInfoObjects();
-    if (!authInfoObjects) {
+    const orgAuthorizations = await this.getOrgAuthorizations();
```
jeffb-sfdc, line 114: Should probably add the return type to line 113:
```
public async updateOrgList(): Promise<string[] | null> {
```

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: Yep.  dumb q: do we _always_ want to have explicit return types on functions?  Up until now I haven't been too concerned about it since TS seems to know about the return types even when they are not always explicit...

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: This block has changed - let me know if you have any comments on the latest!

jeffb-sfdc (approved):

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
gbockus-sf: Letting typescript do it's job and inferring the return type is usually fine.  The one exception is if we need to narrow the return type verse the inferred value. my 2c

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -50,64 +48,58 @@ export class OrgList implements vscode.Disposable {
...
-      } catch (e) {
-        console.log(e);
-      }
-    }
-    return authInfoObjects;
+  public async getAuthFieldsFor(username: string): Promise<AuthFields> {
+    const authInfo: AuthInfo = await AuthInfo.create({
+      username
+    });
+    return authInfo?.getFields();
```
gbockus-sf, line 60: I tested locally it doesn't look like the ? is needed here.  Super weird how the create method is doc'd so I see how got there, but I think it can be removed.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -120,12 +112,12 @@ export class OrgList implements vscode.Disposable {
     return authList;
   }
 
-  public async updateOrgList() {
-    const authInfoObjects = await this.getAuthInfoObjects();
-    if (!authInfoObjects) {
+  public async updateOrgList(): Promise<string[] | null> {
+    const orgAuthorizations = await this.getOrgAuthorizations();
+    if (orgAuthorizations && orgAuthorizations.length === 0) {
       return null;
```
gbockus-sf, line 118: nitty thing, but here you could return an empty array. Then the return type here can be just string[]; This simplifies lines 134-136 removing the need to check if authInfoList exists.

gbockus-sf (approved):
Works great.

Conditions verified (note by verified I mean checking the orgList modal)
✅ Add an existing connection
✅ Add a brand new connection 
✅ Add a default dev hub
✅ add a default scratch org

🐛 
- Found one issue when when I add the scratch org the alias is not included in the org list.  I'll hit you up in slack to discuss if we should address here or as a follow on.

gbockus-sf:
Created W-11610935 bug for tracking scratch org alias issue.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
klewis-sfdc: Updated!

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -75,4 +80,10 @@ export class ConfigUtil {
     const apiVersion = await ConfigUtil.getConfigValue('apiVersion');
     return apiVersion ? String(apiVersion) : undefined;
   }
+
+  public static async getAllAliasesFor(username: string): Promise<string[]> {
+    const stateAggregator = await StateAggregator.getInstance();
+    const aliases = stateAggregator.aliases.getAll(username);
+    return aliases;
+  }
```
klewis-sfdc, line 88: OK to use StateAggregator here in configUtil, or would we prefer to have a separate StateUtil module as a wrapper around StateAggregator?

gbockus-sf: I think it's fine here.  We probably want to have some kind of CoreService at some point, but this isn't the time.

gbockus-sf (approved):
🚢 🚢 🚢

PR #4348, opened 2022-08-04
refactor: use org api version if no user config

klewis-sfdc:
### What does this PR do?
Uses the Api Version from the order associated with the default username when deploying components, if there is no local or global value that has been configured by the user in settings.

@W-11555227@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -63,7 +63,7 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-      void this.setApiVersionOn(components);
+      await this.setApiVersionOn(components);
 
       this.telemetry.addProperty(
         TELEMETRY_METADATA_COUNT,
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -66,7 +66,7 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-      this.setApiVersionOn(components);
+      await this.setApiVersionOn(components);
 
       this.telemetry.addProperty(
         TELEMETRY_METADATA_COUNT,
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -92,10 +92,14 @@ export abstract class DeployRetrieveExecutor<
       | string
       | undefined = await ConfigUtil.getUserConfiguredApiVersion();
 
+    if (userConfiguredApiVersion) {
+      components.apiVersion = userConfiguredApiVersion;
+      return;
+    }
+
     // If no user-configured Api Version is present, then get the version from the Org.
     const orgApiVersion = await OrgAuthInfo.getOrgApiVersion();
-    components.apiVersion =
-      userConfiguredApiVersion ?? orgApiVersion ?? components.apiVersion;
+    components.apiVersion = orgApiVersion ?? components.apiVersion;
   }
 
   protected setupCancellation(
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -4,6 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { Org } from '@salesforce/core';
 import {
   ConfigUtil,
   getRelativeProjectPath,
@@ -38,6 +39,7 @@ import { handleDeployDiagnostics } from '../diagnostics';
 import { nls } from '../messages';
 import { DeployQueue } from '../settings';
 import { SfdxPackageDirectories } from '../sfdxProject';
+import { getDefaultDevHubUsernameOrAlias } from '../util';
 import { createComponentCount, formatException } from './util';
 
 type DeployRetrieveResult = DeployResult | RetrieveResult;
@@ -64,17 +66,7 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-
-      // check the SFDX configuration to see if there is an overridden api version
-      // Run sfdx config:list to enlist all config values
-      // Project level local sfdx-config takes precedence over global sfdx-config at system level.
-      // getComponents uses ComponentSet from SDR which assigns
-      // the default latest value to components.apiversion
-
-      const apiVersion = (await ConfigUtil.getConfigValue('apiVersion')) as
-        | string
-        | undefined;
-      components.apiVersion = apiVersion ?? components.apiVersion;
+      this.setApiVersionOn(components);
 
       this.telemetry.addProperty(
         TELEMETRY_METADATA_COUNT,
@@ -94,6 +86,27 @@ export abstract class DeployRetrieveExecutor<
     } finally {
       await this.postOperation(result);
     }
+  }
+
+  async setApiVersionOn(components: ComponentSet) {
+    // Check the SFDX configuration to see if there is an overridden api version.
+    // Project level local sfdx-config takes precedence over global sfdx-config at system level.
+    const userConfiguredApiVersion = await this.getUserConfiguredApiVersion();
+    
+    // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await this.getOrgApiVersion();
+    components.apiVersion = userConfiguredApiVersion ?? orgApiVersion;
+  }
+
+  async getUserConfiguredApiVersion() {
+    return await ConfigUtil.getConfigValue('apiVersion') as
+    | string
+    | undefined;
+  }
+
+  async getOrgApiVersion() {
+    const aliasOrUsername = await getDefaultDevHubUsernameOrAlias();
+    return await (await Org.create({ aliasOrUsername })).getConnection().getApiVersion();
   }
 
   protected setupCancellation(
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -93,8 +93,9 @@ export abstract class DeployRetrieveExecutor<
       | undefined = await ConfigUtil.getUserConfiguredApiVersion();
 
     // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await OrgAuthInfo.getOrgApiVersion();
     components.apiVersion =
-      userConfiguredApiVersion ?? (await OrgAuthInfo.getOrgApiVersion());
+      userConfiguredApiVersion ?? orgApiVersion ?? components.apiVersion;
   }
 
   protected setupCancellation(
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -96,6 +88,27 @@ export abstract class DeployRetrieveExecutor<
     }
   }
 
+  async setApiVersionOn(components: ComponentSet) {
```
randi274, line 91: Ohhh I like this naming convention

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -96,6 +88,27 @@ export abstract class DeployRetrieveExecutor<
...
+    // Project level local sfdx-config takes precedence over global sfdx-config at system level.
+    const userConfiguredApiVersion = await this.getUserConfiguredApiVersion();
+    
+    // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await this.getOrgApiVersion();
+    components.apiVersion = userConfiguredApiVersion ?? orgApiVersion;
+  }
+
+  async getUserConfiguredApiVersion() {
+    return await ConfigUtil.getConfigValue('apiVersion') as
```
randi274, line 102: Beggars can't be choosers, but could this move to one of the ConfigUtil's as a specifically named method?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -96,6 +88,27 @@ export abstract class DeployRetrieveExecutor<
     }
   }
 
+  async setApiVersionOn(components: ComponentSet) {
+    // Check the SFDX configuration to see if there is an overridden api version.
+    // Project level local sfdx-config takes precedence over global sfdx-config at system level.
+    const userConfiguredApiVersion = await this.getUserConfiguredApiVersion();
+    
+    // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await this.getOrgApiVersion();
```
randi274, line 97: I don't love that we have to get this when we might have the userConfiguredApiVersion already

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -96,6 +88,27 @@ export abstract class DeployRetrieveExecutor<
     }
   }
 
+  async setApiVersionOn(components: ComponentSet) {
+    // Check the SFDX configuration to see if there is an overridden api version.
+    // Project level local sfdx-config takes precedence over global sfdx-config at system level.
+    const userConfiguredApiVersion = await this.getUserConfiguredApiVersion();
```
randi274, line 94: Also the 'user' is maybe not the one who set this... maybe `projectConfiguredApiVersion`?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -64,17 +66,7 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-
-      // check the SFDX configuration to see if there is an overridden api version
```
randi274, line 68: Whooooo love a good comment trim ✂️

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -96,6 +88,27 @@ export abstract class DeployRetrieveExecutor<
...
+    
+    // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await this.getOrgApiVersion();
+    components.apiVersion = userConfiguredApiVersion ?? orgApiVersion;
+  }
+
+  async getUserConfiguredApiVersion() {
+    return await ConfigUtil.getConfigValue('apiVersion') as
+    | string
+    | undefined;
```
gbockus-sf, line 104: when you use `as` you are basically overriding what typescript knows about the type.  This might be better: 

    const apiVersion = await ConfigUtil.getConfigValue('apiVersion');

This way the inferred type is the same (string | undefined) but we still lean on not overriding what typescript is telling us.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
randi274: Oh, will this still get the global setting? Maybe not adding "project" in then...

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Yep, good call!  Moved it to ConfigUtil

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Cool!  Updated.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Also a good callout!  I updated it and added in an assertion to one of the tests to verify that this only gets called if the user-configured value is not set.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Yeah, it will be local if there is one, and global if there isn't a local one set.  Either way I think it is always a user that sets the value that is found in config.  Could go with just configApiVersion I suppose, but wanted it to be clearly distinct from the orgApiVersion.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -125,6 +125,10 @@ export class OrgAuthInfo {
       authInfo: await AuthInfo.create({ username })
     });
   }
+
+  public static async getOrgApiVersion() {
+    return await (await OrgAuthInfo.getConnection()).getApiVersion();
```
gbockus-sf, line 130: total int: For a situation like this I'd probably go with 
```
const connection = await OrgAuthInto.getConnection();
const apiVersion = await connection.getApiVersion();
return apiVersion;
```
Granted it could be more concise, but I love me some code where I can set a breakpoint and see the result of an async call.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -191,14 +194,28 @@ describe('Base Deploy Retrieve Commands', () => {
...
+      sb.stub(ConfigUtil, 'getUserConfiguredApiVersion').resolves(
+        configApiVersion
+      );
+      const getOrgApiVersionSpy = sb.spy(OrgAuthInfo, 'getOrgApiVersion');
 
       await executor.run({ data: {}, type: 'CONTINUE' });
       const components = executor.lifecycle.doOperationStub.firstCall.args[0];
 
       expect(components.apiVersion).to.equal(configApiVersion);
+      expect(getOrgApiVersionSpy.called).to.be.false;
```
gbockus-sf, line 206: nit: `expect(getOrgApiVersionSpy.called).to.equal(false);` does the same assertion without the linter being overriden or beefing.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -191,14 +194,28 @@ describe('Base Deploy Retrieve Commands', () => {
...
+
+    it('should use the api version from the Org when no User-configured api version is set', async () => {
+      const executor = new TestDeployRetrieve();
+      const orgApiVersion = '40.0';
+      sb.stub(ConfigUtil, 'getUserConfiguredApiVersion').resolves(undefined);
+      sb.stub(OrgAuthInfo, 'getOrgApiVersion').resolves(orgApiVersion);
+
+      await executor.run({ data: {}, type: 'CONTINUE' });
+      const components = executor.lifecycle.doOperationStub.firstCall.args[0];
+
```
gbockus-sf, line 217: nit city: in unit tests where I'm stubbing/spying I always like to ensure that the all stubs were called as expected.  So you could add a expects to verify each stub was called once here.

gbockus-sf (approved):
Great work!  A few nits to consider.

gbockus-sf (approved):
Great stuff

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -63,7 +63,7 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-      this.setApiVersionOn(components);
+      void this.setApiVersionOn(components);
```
gbockus-sf, line 66: just curious what lint error are you working around here?

gbockus-sf: also this void hurts my brain....what is it doing?

gbockus-sf: Ok so I pulled it local b/c short on time, but it looks like setApiVersionOn is async.  Are we purposely not waiting on the promise to finish before continuing on? Unless we're purposefully not waiting on the async method to finish before moving on (and if we are we should add a code comment about why) We should just use `await` here vs `void`

klewis-sfdc: When I don't use void here, I get this linter error: "Promises must be handled appropriately"

When I use await here, all the tests in the file fail with "Failure in test '[object Object]': AuthInfoCreationError: Must pass a username and/or OAuth options when creating an AuthInfo instance."

I'll switch to await and see if I can figure it why it makes the other tests fail.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -93,8 +93,9 @@ export abstract class DeployRetrieveExecutor<
       | undefined = await ConfigUtil.getUserConfiguredApiVersion();
 
     // If no user-configured Api Version is present, then get the version from the Org.
+    const orgApiVersion = await OrgAuthInfo.getOrgApiVersion();
     components.apiVersion =
-      userConfiguredApiVersion ?? (await OrgAuthInfo.getOrgApiVersion());
+      userConfiguredApiVersion ?? orgApiVersion ?? components.apiVersion;
```
gbockus-sf: whoa chained nullish coalescing

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -92,10 +92,14 @@ export abstract class DeployRetrieveExecutor<
...
+    if (userConfiguredApiVersion) {
+      components.apiVersion = userConfiguredApiVersion;
+      return;
+    }
+
     // If no user-configured Api Version is present, then get the version from the Org.
     const orgApiVersion = await OrgAuthInfo.getOrgApiVersion();
-    components.apiVersion =
-      userConfiguredApiVersion ?? orgApiVersion ?? components.apiVersion;
+    components.apiVersion = orgApiVersion ?? components.apiVersion;
```
gbockus-sf, line 102: this line has gone through a lot of weird progressions haha.

gbockus-sf (approved):
thanks for fixing that await.  shipit.

PR #4362, opened 2022-08-11
fix: flushFilePath() with fix for Windows

jeffb-sfdc:
### What does this PR do?

### Functionality Before

### Functionality After
1-3: the same
4. This deployment succeeds

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -5,6 +5,8 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import * as fs from 'fs';
+
 export function isNullOrUndefined(object: any): object is null | undefined {
   if (object === null || object === undefined) {
     return true;
@@ -14,11 +16,40 @@ export function isNullOrUndefined(object: any): object is null | undefined {
 }
 
 export function extractJsonObject(str: string): any {
-
-  const jsonString = str.substring(
-    str.indexOf('{'),
-    str.lastIndexOf('}') + 1
-  );
+  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
 
   return JSON.parse(jsonString);
+}
+
+// There's a bug in VS Code where, after a file has been renamed,
+// the URI that VS Code passes to the command is stale and is the
+// original URI.  See https://github.com/microsoft/vscode/issues/152993.
+//
+// To get around this, fs.realpathSync.native() is called to get the
+// URI with the actual file name.
+export function flushFilePath(filePath: string): string {
+  if (filePath === '') {
+    return filePath;
+  }
+
+  let nativePath = fs.realpathSync.native(filePath);
+  if (/^win32/.test(process.platform)) {
+    // The file path on Windows is in the form of "c:\Users\User Name\foo.cls".
+    // When called, fs.realpathSync.native() is returning the file path back as
+    // "C:\Users\User Name\foo.cls", and the capitalization of the drive letter
+    // causes problems further down stream.  To fix this, we'll use the path
+    // returned from fs.realpathSync.native() and then capitalize the first
+    // character.
+    nativePath = nativePath.charAt(0).toLowerCase() + nativePath.slice(1);
+  }
+
+  return nativePath;
+}
+
+export function flushFilePaths(filePaths: string[]): string[] {
+  for (let i = 0; i < filePaths.length; i++) {
+    filePaths[i] = flushFilePath(filePaths[i]);
+  }
+
+  return filePaths;
 }
--- a/packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
@@ -1,4 +1,3 @@
-
 /*
  * Copyright (c) 2019, salesforce.com, inc.
  * All rights reserved.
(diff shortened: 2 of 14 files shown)


jeffb-sfdc:
This PR was originally https://github.com/forcedotcom/salesforcedx-vscode/pull/4240, but needed to be backed out due to an issue running on Windows.  I looked into the issue, and it was b/c under Windows, `fs.realpathSync.native()` returns a sting with the first character (the drive letter) capitalized, and this causes code to fail further downstream.

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
```diff
@@ -7,20 +6,99 @@
...
+  flushFilePaths
+} from '../../../src/helpers';
 
 describe('getConfigSource', () => {
-
   it('should extract a JSON string from larger string and then return as an object', async () => {
-
-    const exampleJsonString = JSON.stringify({ name: 'exampleName', error: 'exampleError' });
+    const exampleJsonString = JSON.stringify({
+      name: 'exampleName',
```
jeffb-sfdc, line 20: Note:  there's a lot of format changes here in this PR.  This was not my doing.  When I save files, they are automatically being re-formatted.  I'm not sure if there was a recent configuration change, but it seems like we have a lot of PRs formatting changes lately.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -16,17 +16,12 @@ import {
...
+import { flushFilePath } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
-import {
-  testOutlineProvider
-} from '../views/testOutlineProvider';
-import {
-  forceAnonApexDebug
-} from './forceAnonApexExecute';
+import { testOutlineProvider } from '../views/testOutlineProvider';
```
jeffb-sfdc, line 23: more auto-formatting changes...

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
klewis-sfdc: I've seen this also.  Checked in w Gordon about it yesterday.  We're going to create a ticket to look into why this is happening.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,11 +16,40 @@ export function isNullOrUndefined(object: any): object is null | undefined {
...
+
+  let nativePath = fs.realpathSync.native(filePath);
+  if (/^win32/.test(process.platform)) {
+    // The file path on Windows is in the form of "c:\Users\User Name\foo.cls".
+    // When called, fs.realpathSync.native() is returning the file path back as
+    // "C:\Users\User Name\foo.cls", and the capitalization of the drive letter
+    // causes problems further down stream.  To fix this, we'll use the path
+    // returned from fs.realpathSync.native() and then capitalize the first
+    // character.
+    nativePath = nativePath.charAt(0).toLowerCase() + nativePath.slice(1);
```
klewis-sfdc, line 43: 👍🏽

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
```diff
@@ -7,20 +6,99 @@
...
+describe('flushFilePath', () => {
+  it('should call fs.realpathSync.native() to resolve a path', async () => {
+    const filePath = 'c:\\Users\\temp\\exampleFile.js';
+    const realpathSyncNativeStub = sinon
+      .stub(fs.realpathSync, 'native')
+      .returns(filePath);
+
+    const result = flushFilePath(filePath);
+
+    expect(realpathSyncNativeStub.called).to.equal(true);
```
klewis-sfdc, line 41: Since calledOnce is used below, is the expect for "called" needed here?

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
```diff
@@ -7,20 +6,99 @@
...
+    const filePaths = ['file.js', 'file.js', 'file.js'];
+    const realpathSyncNativeStub = sinon
+      .stub(fs.realpathSync, 'native')
+      .returns(filePaths[0]);
 
+    const result = flushFilePaths(filePaths);
+
+    expect(result).to.equal(filePaths);
+
+    realpathSyncNativeStub.restore();
```
klewis-sfdc, line 102: Would this restore() be better placed in an afterEach() block?  Or do you prefer to keep it here inside each test case block?

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -75,21 +70,26 @@ function isAnonymousApexFile(sourceUri: vscode.Uri): boolean {
...
   }
 
   await testOutlineProvider.refresh();
-  const testClassName = testOutlineProvider.getTestClassName(sourceUri);
+  let testClassName = testOutlineProvider.getTestClassName(sourceUri);
+  // This is a little bizarre.  Intellisense is reporting that getTestClassName() returns a string,
+  // but it actually it returns string | undefined.  Well, regardless, since flushFilePath() takes
+  // a string (and guards against empty strings) using the Non-null assertion operator
+  // (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)
+  // fixes the issue.
```
klewis-sfdc, line 91: 👍🏽

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceDelete.test.ts
```diff
@@ -51,43 +52,59 @@ describe('ManifestChecker', () => {
...
     expect(response).to.be.true;
+
+    flushFilePathStub.restore();
   });
 });
 
 describe('ConfirmationAndSourcePathGatherer', () => {
   const examplePath = path.join('example', 'path');
-  const explorerPath = { fsPath: examplePath } as vscode.Uri;
+  const explorerPathUri = { fsPath: examplePath } as vscode.Uri;
```
klewis-sfdc, line 81: 👍🏽

klewis-sfdc (approved):
Looks good to me @jeffb-sfdc !  Left one question about maybe using afterEach in the tests, up to you for consideration.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: stupid questions b/c windows.  The default is C:\ but can a user have a drive named a lower case letter?  So would d:\ be valid vs D:\?  There may be a corner case here were a drive is actually a lower case letter (a quick google told me this does happen in some instances, like when using a usb drive).  We probably need a follow on ticket here to address the corner case and having the ability to ask the filesystem the valid path.

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
gbockus-sf: +1 nice catch

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
jeffb-sfdc: For the moment, there's only one test, so I think keeping it central to the test is better.  If there were more tests, and that stub was used in eery test (and the stubbed value never changed), then placing it in `beforeEach` would be a better place, but for here & now, probably best to keep it where it is.

gbockus-sf: +1
Seems you could move the stub into a beforeEach and the restore into an afterEach FTW

gbockus-sf (commented):
One concerning point of discussion around windows machines that have a lower case letter drive.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
jeffb-sfdc: From what I can tell, what's displayed is always uppercase (`D:\`) but what the APIs/SDK uses is always lowercase (`d:\`).  I tested this on a VM running Windows, but it's VERY slow, so I'd like a more thorough test on a real Windows machine b/f we merge.

jeffb-sfdc: (also, if I remember correctly, Windows is case-insensitive, so `d:\` is the same as `D:\`.  When one performs a `FileExists("c:\\foo\\bar.txt")`, it's the same as `FileExists("C:\\FOO\\BAR.TXT")`

gbockus-sf: I'm full of dumb questions.  With the case insensitivity of windows why are we required to make the above change?  It's been too long I don't remember what the 'down stream' changes are that result in the issue.

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
gbockus-sf: hum... I can't tell from the formatting but the `flushFilePath` and `flushFilePaths` describe blocks both appear to have tests that create and restore the following stub
```const realpathSyncNativeStub = sinon
      .stub(fs.realpathSync, 'native')
      .returns(realpathSyncFilePath);
...
      realpathSyncNativeStub.restore();
```
should those not be consolidated? I typically lean towards following patterns in tests to make them easier to grok for both new and existing developers so I'd prefer we put the stub/restore in before/after each blocks. This could be at the top level describe if that makes sense or just in a describe wrapper for all `flushFilePath*` describe blocks.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
jeffb-sfdc: That's a really good question, and I wondered about that as well, but unfortunately, I don't know why the casing is failing.  When I was debugging this, I noticed that the path was "c:\bla\bla\bla.cls" but fs.realpathSync.native() was returning "C:\bla\bla\bla.cls", and when I changed the case for the first char (the drive letter), our actions (deploy, delete, etc...) worked again.

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
gbockus-sf: oh..above I mean I couldn't tell if both describe blocks were wrapped in a single top level describe.

jeffb-sfdc: @gbockus-sf maybe we should take this off line & discuss f2f.  realpathSyncNativeStub can't be hoisted or shared, b/c the mocked return value changes from function to function.

klewis-sfdc:
I ran the same tests on M1 Mac as on the last PR, all checks out:

✅  SFDX: Deploy Source to Org works after renaming locally by only making changes to the casing in the file+class name
✅  SFDX: Deploy Source to Org works when deploying multiple files after renaming one locally by only making changes to the casing in the file+class name
✅  SFDX: Retrieve Source from Org works as before
    * Note: If you rename locally, you still need to deploy the class first so that it is renamed on the server before executing a “Retrieve” operation
✅  SFDX: Delete from Project and Org works as before
    * Note: Similarly, if you rename the component/class locally, you need to deploy it so that the name is aligned on the server before running a delete operation

packages/salesforcedx-utils-vscode/test/unit/helpers/utils.test.ts
gbockus-sf: @jeffb-sfdc Sure I'll ping ya in slack

jeffb-sfdc: I wrapped everything in a `describe()`, added a `beforeEach()` and `afterEach()`, and moved realpathSyncNativeStub

gbockus-sf (approved):
changes look good 👍 
✅ verified on windows

Also attempted to create a volume to check the lower case volume letter and found it wasn't possible using the out of the box disk manager for windows so I think we're good to go 👍

PR #4370, opened 2022-08-16
Update configUtil and its usages to get configuration values from core APIs in salesforcedx-vscode-core

klewis-sfdc:
### What does this PR do?
Updates the salesforcedx-vscode-core package to reference configuration values using the salesforce/core APIs.  

@W-11464489@

### Functionality Before
Configuration values are referenced by accessing the .sfdx directory directly.

### Functionality After
Configuration values are referenced using salesforce/core APIs.


packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
```diff
@@ -92,6 +93,7 @@ describe('Metadata Cache', () => {
       pollStatusStub.callsFake(() => {});
       operationStub.resolves(mockOperation);
       processStub.resolves(undefined);
+      sinon.stub(ConfigUtil, 'getUserConfiguredApiVersion').resolves('55.0');
```
klewis-sfdc, line 96: Without this, this test failed - complaining about a usage of process.chdir()

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -5,37 +5,29 @@
...
   public static async getDefaultUsernameOrAlias(
     enableWarning: boolean
   ): Promise<string | undefined> {
     try {
-      const defaultUserName = await ConfigUtil.getConfigValue(
-        DEFAULT_USERNAME_KEY
-      );
-      if (defaultUserName === undefined) {
+      const defaultUsernameOrAlias = await ConfigUtil.getDefaultUsernameOrAlias();
+      if (defaultUsernameOrAlias === undefined) {
```
gbockus-sf, line 22: nit: unless null is a valid value I prefer `if (!defaultUsernameOrAlias)` here

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -127,21 +120,10 @@ export class OrgAuthInfo {
...
-      ? await AuthUtil.getInstance().getUsername(defaultUsernameOrAlias)
-      : undefined;
-    const connection = await Connection.create({
-      authInfo: await AuthInfo.create({ username })
-    });
+  public static async getOrgApiVersion(): Promise<string | undefined> {
+    const connection = await workspaceContext.getConnection();
     const apiVersion = connection.getApiVersion();
-    return apiVersion;
+    return (apiVersion as string) || undefined;
```
gbockus-sf, line 126: nit: this line is kinda tricky.  Like I had to think kind of hard if we were gonna end up with 'undefined'. 
What do you think of 
```
```
In an ideal world I'd like throw a specific error if there is no apiVersion set that we could catch and handle from the caller but get the scope creep that'd introduce.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -127,21 +120,10 @@ export class OrgAuthInfo {
...
   }
 }
```
gbockus-sf, line 128: The changes in this file make it remarkably better IMO

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -48,13 +48,7 @@ export function disableCLITelemetry() {
...
-    const disabledConfig =
-      (await ConfigUtil.getConfigValue(SFDX_CONFIG_DISABLE_TELEMETRY)) || '';
-    return disabledConfig !== 'true';
-  } catch (e) {
-    console.log('Error checking cli settings: ' + e);
-  }
-  return true;
+export async function isCLITelemetryAllowed() {
+  const isTelemetryDisabled = await ConfigUtil.isTelemetryDisabled();
+  return isTelemetryDisabled === 'true' ? false : true;
```
gbockus-sf, line 53: can this adjustment be made in the ConfigUtil wrapper?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
+  // before returning
+  process.chdir(origCurrentWorkingDirectory);
+  return configAggregator;
+}
+
+// The SfdxConfigAggregator is used only to get configuration
+// values that correspond with old/deprecated config keys.
+// Currently, the key used for the custom templates
+// directory is the only usage, since it is documented for use
+// here: https://developer.salesforce.com/tools/vscode/en/user-guide/byotemplate#set-default-template-location
```
gbockus-sf, line 46: 👍 nit: FWIW this would show up better in the IDE if it were a function block comment of the 
```
/**
 *
 **/ 
``` 
style

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
+  const origCurrentWorkingDirectory = process.cwd();
+  const rootWorkspacePath = getRootWorkspacePath();
+  // Change the current working directory to the project path,
+  // so that ConfigAggregator reads the local project values
+  process.chdir(rootWorkspacePath);
+  const configAggregator = await SfdxConfigAggregator.create();
+  // Change the current working directory back to what it was
+  // before returning
+  process.chdir(origCurrentWorkingDirectory);
+  return configAggregator;
```
gbockus-sf, line 57: total NIT: you could DRY up these two methods a bit if they both used a common method to do the chdir song and dance.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
+}
 
 export class ConfigUtil {
```
gbockus-sf, line 60: dumb q: we already have this call with static methods doing the lords work...why not just have the two get*ConfigAggregator methods here as well?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
-  public static async getConfigSource(key: string): Promise<ConfigSource> {
-    let value = await ConfigUtil.getConfigValue(key, ConfigSource.Local);
-    if (!isNullOrUndefined(value)) {
+  public static async getConfigSource(
+    key: string
+  ): Promise<ConfigSource.Local | ConfigSource.Global | ConfigSource.None> {
```
gbockus-sf, line 63: Did just having ConfigSource as the type here work?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
+    const configAggregator = await getConfigAggregator();
+    const configSource = configAggregator.getLocation(key);
+    if (configSource === ConfigAggregator.Location.LOCAL) {
```
gbockus-sf, line 66: nit: can a switch statement on configSource work here?  I always forget the deets of when switch does and doesn't work in javascript land.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -76,9 +108,64 @@ export class ConfigUtil {
     return undefined;
   }
 
-  public static async getUserConfiguredApiVersion() {
-    const apiVersion = await ConfigUtil.getConfigValue('apiVersion');
-    return apiVersion ? String(apiVersion) : undefined;
+  public static async getUserConfiguredApiVersion(): Promise<
```
gbockus-sf, line 111: nit: same as https://github.com/forcedotcom/salesforcedx-vscode/pull/4370/files#r949573633
Why do we have the user api version method here and the org api method over in authInfo. 

Should we maybe have an API specific service or is that too granular?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/customTemplates.test.ts
```diff
@@ -66,12 +66,12 @@ describe('Custom Templates Create', () => {
     openTextDocumentStub.restore();
     sendCommandEventStub.restore();
     sendExceptionStub.restore();
-    getConfigValue.restore();
+    getTemplatesDirectoryStub.restore();
```
gbockus-sf, line 69: nit: should we add a sandbox here so we don't have to restore all the things and have the ability to parallelize the tests?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceApexClassCreate.test.ts
```diff
@@ -58,7 +58,7 @@ describe('Force Apex Class Create', () => {
     openTextDocumentStub.restore();
     sendCommandEventStub.restore();
     sendExceptionStub.restore();
-    getConfigValue.restore();
+    getTemplatesDirectoryStub.restore();
```
gbockus-sf, line 61: nit: sandbox

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
gbockus-sf: good ol uningration tests

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -80,14 +91,17 @@ describe('OrgAuthInfo', () => {
       );
       expect(showMessageStub.calledOnce).to.equal(true);
 
-      configUtilStub.restore();
+      configAggregatorStub.restore();
```
gbockus-sf, line 94: you should never have to do this with a sandbox.  I'd just add an `afterEach` to the top level describe that does a `sandbox.restore()`.  If we have tests that depend on stubs across tests those shouldn't be happening in an individual test.

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -96,7 +110,7 @@ describe('OrgAuthInfo', () => {
       await OrgAuthInfo.getDefaultDevHubUsernameOrAlias(true);
 
       expect(infoMessageStub.calledOnce).to.equal(false);
-      configUtilStub.restore();
+      configAggregatorStub.restore();
```
gbockus-sf, line 113: same..can be handled in the `afterEach` restore.

gbockus-sf (commented):
Changes are looking awesome.  A few discussion points in there.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,19 +26,47 @@ export enum ConfigSource {
...
+  const rootWorkspacePath = getRootWorkspacePath();
+  // Change the current working directory to the project path,
+  // so that ConfigAggregator reads the local project values
+  process.chdir(rootWorkspacePath);
+  const configAggregator = await ConfigAggregator.create();
+  // Change the current working directory back to what it was
+  // before returning
+  process.chdir(origCurrentWorkingDirectory);
+  return configAggregator;
+}
```
jeffb-sfdc, line 40: When changing the current working directory, I suggest wrapping the code in a try/finally block, so that in case anything throws, the working directory is set back.  Something like this:
```
async function getConfigAggregator(): Promise<ConfigAggregator> {
  const origCurrentWorkingDirectory = process.cwd();
  const rootWorkspacePath = getRootWorkspacePath();
  try {
    // Change the current working directory to the project path,
    // so that ConfigAggregator reads the local project values
    process.chdir(rootWorkspacePath);
    const configAggregator = await ConfigAggregator.create();
  } finally {
    // Change the current working directory back to what it was
    // before returning
    process.chdir(origCurrentWorkingDirectory);
  }    
  return configAggregator;
}
```
(warning: untested code)



Also, does the CWD need to be changed every time `ConfigAggregator.create()` is called?  If so, maybe move the CWD code into `ConfigAggregator.create()`.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
klewis-sfdc: This one is a configuration value that is set by the user. It is intended to be used to override the other one in some cases, which is the API version of the currently authenticated Org. So I think they are in the correct places.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
klewis-sfdc: let me know what you think of the changes.  added an optional param and convenience method

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
klewis-sfdc: looks like there already is sandbox.restore in afterEach.  Ill remove these

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,63 +21,85 @@ export enum ConfigSource {
...
+    const configSource = configAggregator.getLocation(key);
+    switch (configSource) {
+      case ConfigAggregator.Location.LOCAL:
+        return ConfigSource.Local;
+        break;
+      case ConfigAggregator.Location.GLOBAL:
+        return ConfigSource.Global;
+        break;
+      default:
+        return ConfigSource.None;
```
gbockus-sf, line 36: nice

packages/salesforcedx-vscode-core/src/util/configUtil.ts
gbockus-sf: would this info be worth a method comment?

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,63 +21,85 @@ export enum ConfigSource {
...
-    }
-    return undefined;
+  public static async getUserConfiguredApiVersion(): Promise<
+    string | undefined
+  > {
+    const configAggregator = await ConfigUtil.getConfigAggregator();
+    const apiVersion = configAggregator.getPropertyValue(
+      OrgConfigProperties.ORG_API_VERSION
+    );
+    return (apiVersion as string) || undefined;
```
gbockus-sf, line 47: Nit: one of those `apiVersion ? String(apiVersion) : undefined` situations.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -22,63 +21,85 @@ export enum ConfigSource {
...
+    return isTelemetryDisabled === 'true';
+  }
+
+  public static async getDefaultDevHubUsername(): Promise<string | undefined> {
+    const configAggregator = await ConfigUtil.getConfigAggregator();
+
+    const defaultDevHubUserName = configAggregator.getPropertyValue(
+      OrgConfigProperties.TARGET_DEV_HUB
+    );
+    return (defaultDevHubUserName as string) || undefined;
```
gbockus-sf, line 92: Same nit: ternary is better

gbockus-sf: for all the `as sting` spots in this file.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -89,4 +110,40 @@ export class ConfigUtil {
...
+  /*
+   *  The SfdxConfigAggregator is used only to get configuration
+   *  values that correspond with old/deprecated config keys.
+   *  Currently, the key used for the custom templates
+   *  directory is the only usage, since it is documented for use
+   *  here: https://developer.salesforce.com/tools/vscode/en/user-guide/byotemplate#set-default-template-location
+   */
+  private static async getSfdxConfigAggregator(): Promise<ConfigAggregator> {
+    const sfdxConfigAggregator = ConfigUtil.getConfigAggregator(true);
+    return sfdxConfigAggregator;
```
gbockus-sf, line 147: nice!

gbockus-sf (commented):
Changes look good!

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+  globalValuesOnly?: boolean;
+};
+
+/*
+ * The ConfigAggregator class is used to abstract away
+ * some of the complexities around changing the process directory
+ * that are needed to accurately retrieve configuration values
+ * when using the ConfigAggregator in the VSCE context.
+ */
+export class ConfigAggregatorProvider {
```
gbockus-sf, line 85: nice concept 👍 .  I'd put this in a `src/providers` directory.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+    this.configAggregators = new Map<string, ConfigAggregator>();
+    this.sfdxConfigAggregators = new Map<string, ConfigAggregator>();
+  }
+
+  public async getConfigAggregator(): Promise<ConfigAggregator> {
+    const rootWorkspacePath = getRootWorkspacePath();
+    let configAggregator = this.configAggregators.get(rootWorkspacePath);
+    if (!configAggregator) {
+      configAggregator = await this.createConfigAggregator();
+      this.configAggregators.set(getRootWorkspacePath(), configAggregator);
```
gbockus-sf, line 109: nit: use `rootWorkspacePath` here as the key to avoid looking it up again.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+  }
+
+  public async getSfdxConfigAggregator(): Promise<ConfigAggregator> {
+    let sfdxConfigAggregator = this.sfdxConfigAggregators.get(
+      getRootWorkspacePath()
+    );
+    if (!sfdxConfigAggregator) {
+      sfdxConfigAggregator = await this.createConfigAggregator({ sfdx: true });
+      this.sfdxConfigAggregators.set(
+        getRootWorkspacePath(),
```
gbockus-sf, line 121: same nit: only call `getRootWorkspacePath` once.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+      configAggregator = options.sfdx
+        ? await SfdxConfigAggregator.create()
+        : await ConfigAggregator.create();
+    } finally {
+      // Change the current working directory back to what it was
+      // before returning.
+      // Wrapping this in a finally block ensures that the working
+      // directory is switched back to what it was before this method
+      // was called if SfdxConfigAggregator.create() throws an exception.
+      process.chdir(currentWorkingDirectory);
```
gbockus-sf, line 180: do we need any error handling in a catch maybe?  Seems like a bad state if we encountered an error attempting to create an aggregator and it fails.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+    }
+    return configAggregator;
+  }
+
+  private static ensureProcessIsRunningUnderUserHomeDir(path: string) {
+    const userHomePath = '/';
```
gbockus-sf, line 186: Just calling out this says user path, but '/' isn't a home path.  It's like a base file system path.  If this is an existing flow (or what sfdx does) then fine, but some users may not have permission to navigate to that directory

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -7,78 +7,104 @@
...
+  OrgConfigProperties,
+  SfConfigProperties,
+  SfdxPropertyKeys,
   StateAggregator
 } from '@salesforce/core';
-import * as path from 'path';
-import { isNullOrUndefined, isUndefined } from 'util';
-import { telemetryService } from '../telemetry';
-import { getRootWorkspacePath } from './index';
+import { ConfigAggregatorProvider } from '../context/workspaceContext';
```
gbockus-sf, line 15: well these imports are waaaay better

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
```diff
@@ -322,16 +324,18 @@ describe('Metadata Cache', () => {
...
-      fullName: 'HandlerCostCenter',
-      lastModifiedDate: 'Today',
-      type: 'ApexClass'
-    },
-    {
-      fullName: 'Account',
-      lastModifiedDate: 'Yesterday',
-      type: 'CustomObject'
-    }] as FileProperties[];
+    const fileProperties = [
```
gbockus-sf, line 327: nit: does `const: Array<FileProperties> = [{` work?

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -54,3 +63,136 @@ export class WorkspaceContext {
...
+    const rootWorkspacePath = getRootWorkspacePath();
+    let configAggregator = this.configAggregators.get(rootWorkspacePath);
+    if (!configAggregator) {
+      configAggregator = await this.createConfigAggregator();
+      this.configAggregators.set(getRootWorkspacePath(), configAggregator);
+    }
+    return configAggregator;
+  }
+
+  public async getSfdxConfigAggregator(): Promise<ConfigAggregator> {
```
gbockus-sf, line 114: a stand alone test suite for this class would be awwesome

gbockus-sf (commented):
some comments for you consideration.  Main beef would be moving `ConfigAggregatorProvider` to it's own file

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
klewis-sfdc: This is the default dir that VSCE runs processes under.  If the caller wants a global values only configAgg, then the process needs to be running under the Users root dir (same dir where the global .sf and .sfdx dirs are) when the ConfigAgg is created.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
klewis-sfdc: There is at least one [place in plugin-source](https://github.com/salesforcecli/plugin-source/blob/b7829b99935b472833f8faf8652566fe446aa6b8/src/deployCommand.ts#L139-L140) where it is used and not wrapped in a t/c.  I think an exception would be rare and if there is one it should be thrown here.

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
klewis-sfdc: Looks like some props are missing:

gbockus-sf: Weird: I swear the linter just yelled at me to use Array<T> vs T[] but whatever.  
Hum....if you wanted to leave this cleaner than you found it you could add the properties and remove the `as`.  By saying `as FileProperties[]` we are telling typescript we know better then it does what type the object is which can lead to badness where we _think_ that a property has to exist on an object of the type but then it reality it doesn't.  I get this is a test file and it may be less painful to allow `as` at will in tests to keep the LOC down.  I'm open to discussion about it.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
gbockus-sf: in the finally we won't know if there's an exception right?  So we'd need to catch it and save it then re-throw post chdir right?

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
gbockus-sf: Screen Shot 2022-08-25 at 6 04 24 PM

So does ConfigAgg do the global lookup in the user home dir if no config directories are found?  Or does it need to be in the users home dir in order to find and load the globals?

In the screenshot I navigated to '/' first and there are no sfdx config dirs.  Then the second dir is my home dir where my global config dirs exits.  

I suspect if no config directories are found then configAgg will check the user home space for global config dirs so we're good to go. 

One naming nit here to avoid confusing might be to rename the method to `ensureCurrentDirectoryOutsideProject` or something...stupid naming.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -6,79 +6,110 @@
...
+    const defaultDevHubUserName = configAggregator.getPropertyValue(
+      OrgConfigProperties.TARGET_DEV_HUB
+    );
+    return defaultDevHubUserName ? String(defaultDevHubUserName) : undefined;
+  }
+
+  public static async getGlobalDefaultDevHubUsernameOrAlias(): Promise<
+    string | undefined
+  > {
+    const globalConfig = await Config.create({ isGlobal: true });
```
klewis-sfdc, line 105: Was able to figure out how to use Config here to get the global value.  This simplifies things in the ConfigAggregatorProvider and its tests and feels like a more solid way to get the global value than trying to ensure the process is not in a project dir before instantiating the ConfigAggregator..

packages/salesforcedx-vscode-core/src/providers/configAggregatorProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+ * ConfigAggregator and SfdxConfigAggregator in the VSCE context.
+ * In order to accurately retrieve both project and global variable values,
+ * It needs to be instantiated while process.cwd() returns the root project
+ * workspace path. ConfigAggregatorProvider handles this and switches the
+ * cwd back so that it is the same when the function finishes.
+ */
+export class ConfigAggregatorProvider {
+  protected configAggregators: Map<string, ConfigAggregator>;
+  protected sfdxConfigAggregators: Map<string, ConfigAggregator>;
+  protected globalConfigAggregator: ConfigAggregator | undefined = undefined;
```
RitamAgrawal, line 33: Any specific reason for defining undefined like this? Why can't we just go with `ConfigAggregator | undefined`

klewis-sfdc: updated!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/customTemplates.test.ts
```diff
@@ -38,40 +39,39 @@ describe('Custom Templates Create', () => {
...
   let sendExceptionStub: SinonStub;
-  let getConfigValue: SinonStub;
+  let getTemplatesDirectoryStub: SinonStub;
 
   beforeEach(() => {
-    showInputBoxStub = stub(vscode.window, 'showInputBox');
-    quickPickStub = stub(vscode.window, 'showQuickPick');
-    appendLineStub = stub(channelService, 'appendLine');
-    showSuccessfulExecutionStub = stub(
+    sandbox = createSandbox();
```
gbockus-sf, line 45: late to the game comment: sandbox only needs to be created once so this should be done either outside of the describe or just inside to the top level describe.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/customTemplates.test.ts
```diff
@@ -38,40 +39,39 @@ describe('Custom Templates Create', () => {
...
+      'showFailedExecution'
+    );
+    openTextDocumentStub = sandbox.stub(vscode.workspace, 'openTextDocument');
+    sendCommandEventStub = sandbox.stub(telemetryService, 'sendCommandEvent');
+    sendExceptionStub = sandbox.stub(telemetryService, 'sendException');
+    getTemplatesDirectoryStub = sandbox.stub(
+      ConfigUtil,
+      'getTemplatesDirectory'
+    );
+    getTemplatesDirectoryStub.returns(undefined);
```
gbockus-sf, line 65: nit: you can combine these thow lines like ```getTemplatesDirectoryStub = sandbox.stub(
      ConfigUtil,
      'getTemplatesDirectory'
    ).returns(undefined);
    ```

gbockus-sf: looking at the tests all of them set a returns value on the stub so I don't think you need to include it here.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceApexClassCreate.test.ts
```diff
@@ -30,35 +31,30 @@ describe('Force Apex Class Create', () => {
...
   let sendExceptionStub: SinonStub;
-  let getConfigValue: SinonStub;
+  let getTemplatesDirectoryStub: SinonStub;
 
   beforeEach(() => {
-    showInputBoxStub = stub(vscode.window, 'showInputBox');
-    quickPickStub = stub(vscode.window, 'showQuickPick');
-    appendLineStub = stub(channelService, 'appendLine');
-    showSuccessfulExecutionStub = stub(
+    sandbox = createSandbox();
```
gbockus-sf, line 37: same sandbox concern..should be done higher up and only once.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/forceApexClassCreate.test.ts
```diff
@@ -30,35 +31,30 @@ describe('Force Apex Class Create', () => {
...
+    showInputBoxStub = sandbox.stub(vscode.window, 'showInputBox');
+    quickPickStub = sandbox.stub(vscode.window, 'showQuickPick');
+    appendLineStub = sandbox.stub(channelService, 'appendLine');
+    showSuccessfulExecutionStub = sandbox
+      .stub(notificationService, 'showSuccessfulExecution')
+      .returns(Promise.resolve());
```
gbockus-sf, line 43: nit: using `.resolves()` is the same as `.returns(Promise.resolve())`

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -122,14 +122,18 @@ describe('Force Project Create', () => {
       inputBoxStub.returns(PROJECT_NAME);
       const response = await gatherer.gather();
       expect(response.type).to.equal('CONTINUE');
-      expect((response as ContinueResponse<ProjectName>).data.projectName).to.equal(PROJECT_NAME);
```
gbockus-sf, line 125: nit: we could add a typecheck to remove the need for the as casting here. 

```const isContinueResponse = (
  response: CancelResponse | ContinueResponse<ProjectName>
): response is ContinueResponse<ProjectName> => {
  return (response as ContinueResponse<ProjectName>).data !== undefined;
};```

Then the test would look like 
```
      expect(isContinueResponse(response)).to.equal(true);
      if (isContinueResponse(response)) {
        expect(response.type).to.equal('CONTINUE');
        expect(response.data.projectName).to.equal(PROJECT_NAME);
      }
      ```
      Check out type predicates here: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -46,8 +51,11 @@ describe('OrgAuthInfo', () => {
 
   describe('getDefaultDevHubUsernameOrAlias', () => {
     it('should return notification if there is no dev hub set', async () => {
-      const configUtilStub = sandbox.stub(ConfigUtil, 'getConfigValue');
-      configUtilStub.returns(undefined);
+      const configAggregatorStub = sandbox.stub(
+        ConfigAggregator.prototype,
+        'getPropertyValue'
+      );
+      configAggregatorStub.returns(undefined);
```
gbockus-sf, line 58: nit: can be chained.

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -79,15 +88,14 @@ describe('OrgAuthInfo', () => {
...
     });
 
     it('should not show a message if there is a dev hub set', async () => {
-      const configUtilStub = sandbox.stub(ConfigUtil, 'getConfigValue');
-      configUtilStub.returns('username');
+      const configAggregatorStub = sandbox.stub(
+        ConfigAggregator.prototype,
+        'getPropertyValue'
+      );
+      configAggregatorStub.returns('username');
```
gbockus-sf, line 98: nit: totally a best practices thing, but if I'm creating a stub in a test I always have an assertion to verify it was called and potentially an assertion to verify the args passed.

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ConfigAggregator, SfdxConfigAggregator } from '@salesforce/core';
+import { expect } from 'chai';
+import { createSandbox, SinonSpy, SinonStub } from 'sinon';
+import Sinon = require('sinon');
+import { ConfigAggregatorProvider } from '../../../src/providers/configAggregatorProvider';
```
gbockus-sf, line 11: You should add `stub` to the list of things imported from sinon.

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      // Assert
+      // createConfigAggregator should store the cwd initially,
+      // and check it again after creating the ConfigAggregator
+      // to ensure that the cwd is set back to its original value.
+      expect(getCurrentDirectoryStub.callCount).to.equal(2);
+      // Since the stubbed current directory is the same as the root
+      // workspace path, the directory should not have been changed.
+      expect(changeCurrentDirectoryToStub.callCount).to.equal(0);
+      expect(configAggregatorCreateSpy.callCount).to.equal(1);
+      expect(configAggregator).to.not.equal(undefined);
```
gbockus-sf, line 92: total nit: I typically do the assertion the returned value first.  If you didn't get back what you expected does it make sense to do the other assertions?

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+        'create'
+      );
+
+      // Act
+      const sfdxConfigAggregator = await (configAggregatorProvider as any).createConfigAggregator(
+        { sfdx: true }
+      );
+
+      // Assert
+      expect(sfdxConfigAggregatorCreateSpy.callCount).to.equal(1);
```
gbockus-sf, line 112: should we assert here that the regular create method was not called?

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      expect(sfdxConfigAggregator).to.not.equal(undefined);
+    });
+
+    it('should change back to the original current directory if ConfigAggregator creation fails', async () => {
+      // Arrange
+      configAggregatorCreateSpy.restore();
```
gbockus-sf, line 118: interesting...I've never seen this pattern in code before.  Might be simpler to just make it a stub to start or wrap the tests in describe blocks to do the different flavors of setup on the `create` method.

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      const configAggregatorCreateStub = Sinon.stub(ConfigAggregator, 'create');
+      configAggregatorCreateStub.throws(
```
gbockus-sf, line 120: nit: can chain.

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      // createConfigAggregator should store the cwd initially,
+      // and check it again after creating the ConfigAggregator
+      // to ensure that the cwd is set back to its original value.
+      expect(getCurrentDirectoryStub.callCount).to.equal(2);
+      expect(changeCurrentDirectoryToStub.callCount).to.equal(2);
+      expect(changeCurrentDirectoryToStub.getCall(1).args[0]).to.equal(
+        ConfigAggregatorProvider.defaultBaseProcessDirectoryInVSCE
+      );
+      expect(configAggregatorCreateStub.callCount).to.equal(1);
+      expect(caughtError).to.not.equal(undefined);
```
gbockus-sf, line 146: nit: typically it's good to verify your error here to ensure the thrown error was the one you expected. Typically I'd make 'There was a problem creating the Config Aggregator.' a const and pass that to the new Error and use it in the expect ala `expect(error.message).to.equal(errMessage);`

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      );
+
+      // Act
+      await configAggregatorProvider.reloadConfigAggregators();
+
+      // Assert
+      expect(configAggregatorReloadSpy.callCount).to.equal(2);
+    });
+  });
+});
```
gbockus-sf, line 205: These tests are *awesome* :celebrate:

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -6,79 +6,110 @@
  */
 
 import {
+  Config,
   ConfigAggregator,
   ConfigFile,
```
gbockus-sf, line 11: nit: unused

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -6,79 +6,110 @@
...
+
+  /*
+   * Currently, the docs tell Users to manually create this entry in the .sfdx
+   * configuration file.  For that reason, getTemplatesDirectory uses the
+   * SfdxConfigAggregator specifically.
+   */
+  public static async getTemplatesDirectory(): Promise<string | undefined> {
+    const sfdxConfigAggregator = await ConfigAggregatorProvider.getInstance().getSfdxConfigAggregator();
+    const templatesDirectory = sfdxConfigAggregator.getPropertyValue(
+      SfdxPropertyKeys.CUSTOM_ORG_METADATA_TEMPLATES
```
gbockus-sf, line 79: Debugging through this the currect setup results in always getting undefined back.  If I update it to use `OrgConfigProperties.ORG_CUSTOM_METADATA_TEMPLATES` we are in business.

gbockus-sf (changes requested):
First pass at QE
* ✅ Verify getRootWorkspacePath()
    * Verified via breakpoint
* ❌ Verify getting custom template setting
    * Added the following to the .sfdx-config.json   "customOrgMetadataTemplates": "https://github.com/gbockus-sf/salesforcedx-templates/tree/main/packages/templates/src/templates"
    * After addition still did not use custom templates
    * Tracking in code I see getTemplatesDirectory() returning undefined. 
    * I attempted to restart vscode to see if the value was cached and it didn’t make any difference. 
    * Updating getTemplatesDirectory() to use OrgConfigProperties.ORG_CUSTOM_METADATA_TEMPLATES fixes the issue. 
* ✅ Change a config with the cli and verify it is loaded in the workspace context
    * Verified current default dev or in .sfdx-config.json
    * ran sfdx force:config:set defaultusername=devOrg1 on the command
    * Verified that configAggs were reloaded. 
    * See the defaultOrg update 
* ✅ Verify local read of ApiVersion
    * Run a deploy and see current api version being used: 55.0
    * run sfdx force:config:set apiVersion=54.0
    * Run deploy/retrieve and verify version 54.0 is used on the componentSet.
* ❌ Verify reading cli telemetry setting
    * set breakpoint in telemetry.ts to verify that it is currently enabled
    * disable telemetry by running: sfdx config:set disableTelemetry=true
    * restart vscode → verify telemetry setting is off. 
    * Via breakpoint I can still see it enabling the reporter even with telemetry off. 
    * The this.isTelemetryEnabled() needs to be awaited on line 106 of telemetry.ts

Couple of issues to be addressed and potentially have tests to cover the error cases.

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
klewis-sfdc: Removed since no need for sfdx flavor of Config Aggregator anymore - the new ConfigAgg works with the old key

packages/salesforcedx-vscode-core/test/vscode-integration/providers/configAggregatorProvider.test.ts
klewis-sfdc: moved to a const before the describe

packages/salesforcedx-vscode-core/src/util/configUtil.ts
klewis-sfdc: Regular Config Agg with the new key works.  Removed the sfdx config agg part of the provider!  Yay for simpler

gbockus-sf (approved):
Follow up after changes: 
    *  ✅ Template setting working as expected for remote
        *   "customOrgMetadataTemplates": "https://github.com/gbockus-sf/salesforcedx-templates/tree/main/packages/templates/src/templates"
              

* ✅ Verify reading cli telemetry setting
    * Followed steps from first QE and see that telemetry is being read correctly.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
gbockus-sf: @klewis-sfdc did you decide against the typecheck here?

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
gbockus-sf: @klewis-sfdc thoughts on this one?

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -101,11 +101,8 @@ export class TelemetryService {
     const isDevMode = machineId === 'someValue.machineId';
 
     // TelemetryReporter is not initialized if user has disabled telemetry setting.
-    if (
-      this.reporter === undefined &&
-      this.isTelemetryEnabled() &&
-      !isDevMode
-    ) {
+    const isTelemetryEnabled = await this.isTelemetryEnabled();
+    if (this.reporter === undefined && isTelemetryEnabled && !isDevMode) {
```
gbockus-sf, line 105: oh...interesting. So what change here that might have been the root cause of the test timeouts is by pulling the assignment to the variable outside of the conditional we now await it every time this is called...where as previously if reported is defined it would never be called after the first attempted unless it was disabled. Might be worth moving back to 
`this.reporter === undefined && await this.isTelemetryEnabled() && !isDevMode` to avoid making the async call everytime

gbockus-sf (approved):
Reverify 
✅ getRootWorkspace
✅ Templates config 
✅ Reload on config change
✅ API no changes good to go
✅ verify cli telemetry setting is used

* I did add a code comment here as when I was testing I realized that we hit that flow several times on app load, and with the changes it now awaits that calls to get the config every time, where as when it was in the conditional it was only called when necessary.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
klewis-sfdc: Yep, def should switch this back with that in mind.  Updated!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: Nope, meant to ask you about this just to clarify further - Ill take a look at this this afternoon

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
klewis-sfdc: Updated! let me know how it looks

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
klewis-sfdc: updated!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -120,20 +128,26 @@ describe('Force Project Create', () => {
...
-      expect(response.type).to.equal('CONTINUE');
-      expect(
-        (response as ContinueResponse<ProjectName>).data.projectName
-      ).to.equal(PROJECT_NAME);
+
+      expect(isContinueResponse(response)).to.equal(true);
+      if (isContinueResponse(response)) {
+        expect(response.type).to.equal('CONTINUE');
+        expect(response.data.projectName).to.equal(PROJECT_NAME);
+      }
```
gbockus-sf, line 150: sweeeeet

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/metadataCacheService.test.ts
```diff
@@ -324,18 +324,32 @@ describe('Metadata Cache', () => {
         name: 'CustomField'
       }
     };
-    const fileProperties = [
+    const fileProperties: FileProperties[] = [
```
gbockus-sf, line 327: Thanks!

gbockus-sf (approved):
🚢 :shipit: 🚢 :shipit: 🚢 :shipit: 🚢 :shipit:

PR #4401, opened 2022-09-01
chore: workflow fix for node version

randi274:
### What does this PR do?
- Reverts latest, since that syntax [did not work](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/2974301105).

### DIFF
--- a/.github/workflows/publish-beta-release.yml
+++ b/.github/workflows/publish-beta-release.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v1
         with:
-          node-version: latest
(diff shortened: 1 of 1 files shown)


klewis-sfdc (approved):

.github/workflows/publish-beta-release.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v1
         with:
-          node-version: latest
+          node-version: '16.13.x'
```
gbockus-sf, line 14: dumb q: what if you just don't include a node version?

PR #4408, opened 2022-09-07
feat: call core lib to get the sfdx state folder name in Org Browser

RitamAgrawal:
### What does this PR do?
This PR replaces string`.sfdx` with core lib call . Earlier we were passing the directory name as string  but now calling `Global.SFDX_STATE_FOLDER` to obtain the directory name.
@W-11614723@

### Functionality Before
To form the metadata component json file path, string`.sfdx` was passed to path.join() method.

### Functionality After
Calling sfdx-core lib to obtain the state folder through Global class `Global.SFDX_STATE_FOLDER`.


packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -56,4 +59,21 @@ export class WorkspaceContext {
   get alias(): string | undefined {
     return WorkspaceContextUtil.getInstance().alias;
   }
+
+  public getSfdxDirectoryPath(): string {
```
RitamAgrawal, line 63: Need to remove this.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -62,3 +64,24 @@ export function fileExtensionsMatch(sourceUri: vscode.Uri, targetExtension: stri
...
+function getMetadataDirectoryPath(username: string): string {
+  return path.join(
+    getSfdxDirectoryPath(),
+    'orgs',
+    username,
+    'metadata'
+  );
+}
+
+export const projectPath = {
```
klewis-sfdc, line 84: Q: does this have to be a named export?  Could it be done like:

```
export {
  sfdxDirectoryPath,
  metadataDirectoryPath
}
```

so that when using these from the outside it would be:

```
import * from paths;

const sfdxDirPath = paths.sfdxDirectoryPath;
```

gbockus-sf: hum...This is one of the situations where I don't know why I prefer namedExports.  I think it's b/c the only two states I'm aware of is named or default exports [doc](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) I aways avoid default exports b/c it's easier to stay consistent with named exports. 

TBH I'm not even sure how it works if you export an object directly as suggested.  I'd have to test but I bet it just attaches each property as a stand along export we we'd end up with the reference vs value problem when stubbing in tests. 

bottom line I'd go with what we know and stick with the named export.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
```diff
@@ -86,7 +86,7 @@ describe('get metadata components path', () => {
   function expectedPath(fileName: string) {
     return path.join(
       rootWorkspacePath,
-      '.sfdx',
```
gbockus-sf, line 89: here instead of adding the core import I'd just stub projectPath.getMetadataDirectoryPath and use the value that stub returns here.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
klewis-sfdc: Ok sounds good to me!  How about making the exported named object plural so it is `paths.projectPaths.getSfdxDirectoryPath()`

gbockus-sf (commented):
There's the one test update otherwise looks good to go. I can QE quickly

gbockus-sf (approved):
QE'd locally.  Verified the path through code inspection after opening and refresh components in the org browser. ✅ 

Good to merge after the test update

PR #4421, opened 2022-09-13
[WIP] feat: Add support for Python functions

edmorley:
See also:
https://github.com/heroku/sf-functions-core/pull/59

GUS-W-11716024.


edmorley:
Closing for now, since this was only a partial WIP, and it's likely that when the time comes (when Python Functions is ready), it will be the owners of this repository that add support, rather than the Languages team.

PR #4422, opened 2022-09-13
chore: move configAggProvider to from core to utils

gbockus-sf:
### What does this PR do?
This is a change to prove out having a dependency from core to the utils-vscode package.  The intention is to only have classes for accessing configs in the untils package that all other extensions use to utilize. 

@W-11528403@

### Functionality Before
configAggragatorProvider lives in core package.

### Functionality After
configAggragatorProvider lives in the utils package.


packages/salesforcedx-utils-vscode/src/providers/configAggregatorProvider.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { ConfigAggregator } from '@salesforce/core';
-import { workspaceUtils } from '../util';
+import { workspaceUtils } from '../workspaces';
```
gbockus-sf, line 9: So the way we have our directory structure across packages is kind of wacky right now.  The workspaceUtils seems like it would live in the workspace dir so I moved it there.

packages/salesforcedx-utils-vscode/src/workspaces/rootWorkspace.ts
```diff
@@ -9,6 +9,7 @@ import { join } from 'path';
 import { workspace, WorkspaceFolder } from 'vscode';
 import { SFDX_FOLDER } from '../types';
 
+// TODO: consolidate all of these into workspaceUtils
```
gbockus-sf, line 12: Note there was already a file that has some of the same functionality as what's in workspaceUtils, but I didn't want to blow up this PR with the refactor.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -6,12 +6,12 @@
  */
 import { Connection } from '@salesforce/core';
 import {
+  ConfigAggregatorProvider,
```
gbockus-sf, line 9: All this work for this one line :)

packages/salesforcedx-vscode-core/src/providers/index.ts
```diff
@@ -0,0 +1 @@
+// export { configAggregatorProvider } from './configAggregatorProvider';
```
gbockus-sf, line 1: whoops..will delete.

packages/salesforcedx-vscode-core/src/util/configUtil.ts
```diff
@@ -12,7 +12,7 @@ import {
   SfConfigProperties,
   StateAggregator
 } from '@salesforce/core';
-import { ConfigAggregatorProvider } from '../providers/configAggregatorProvider';
+import { ConfigAggregatorProvider } from '@salesforce/salesforcedx-utils-vscode';
```
gbockus-sf, line 15: configUtils is next on my list to move.

packages/salesforcedx-vscode-core/test/vscode-integration/util/authInfo.test.ts
```diff
@@ -5,26 +5,26 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import {
-  AuthInfo,
-  ConfigAggregator,
-  Connection,
-  StateAggregator
-} from '@salesforce/core';
-import { OrgConfigProperties } from '@salesforce/core/lib/exported';
```
gbockus-sf, line 14: since moving the configAggProvider to utils we can't directly stub that method in core.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -92,10 +93,8 @@ export class ConfigUtil {
    * @param projectPath The project path for the currently SF project.
    * @returns The username for the configured Org if it exists.
    */
-  public static async getUsername(
-    projectPath: string
-  ): Promise<string | undefined> {
-    const configAggregator = await ConfigUtil.getConfigAggregator(projectPath);
+  public static async getUsername(): Promise<string | undefined> {
+    const configAggregator = await ConfigAggregatorProvider.getInstance().getConfigAggregator();
     const defaultUsernameOrAlias = configAggregator.getPropertyValue(
```
gbockus-sf, line 98: We (@RitamAgrawal and I) found a race condition around loading the configAggregator directly where it would only contain local values.  Since the whole point of moving it to vscode-utils and sharing it is to avoid these issues updated this function to use the provider 🎉

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -92,10 +93,8 @@ export class ConfigUtil {
    * @param projectPath The project path for the currently SF project.
    * @returns The username for the configured Org if it exists.
    */
-  public static async getUsername(
-    projectPath: string
-  ): Promise<string | undefined> {
-    const configAggregator = await ConfigUtil.getConfigAggregator(projectPath);
+  public static async getUsername(): Promise<string | undefined> {
```
gbockus-sf, line 96: this doesn't need the project path when using the provider

PR #4432, opened 2022-09-15
chore: Updates in the core package Part 6

gbockus-sf:
### What does this PR do?
Files touched in the core package

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/auth/authParamsGatherer.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/auth/authParamsGatherer.ts
@@ -11,7 +11,7 @@ import {
   CancelResponse,
   ContinueResponse,
   ParametersGatherer
-} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+} from '@salesforce/salesforcedx-utils-vscode';
 import { nls } from '../../messages';
 import { SfdxProjectConfig } from '../../sfdxProject';
 
--- a/packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
@@ -9,7 +9,7 @@ import {
   CliCommandExecutor,
   Command,
   SfdxCommandBuilder
-} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
+} from '@salesforce/salesforcedx-utils-vscode';
 
 import {
   EmptyParametersGatherer,
@@ -18,11 +18,11 @@ import {
   SfdxWorkspaceChecker
 } from '../util';
 
-import { getRootWorkspacePath } from '../../util';
+import { workspaceUtils } from '../../util';
 
 import { ConfigFile } from '@salesforce/core';
-import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
-import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode';
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode';
 import { homedir } from 'os';
 import * as vscode from 'vscode';
 import { CLI } from '../../constants';
@@ -73,7 +73,7 @@ export class ForceAuthDevHubExecutor extends SfdxCommandletExecutor<{}> {
     const cancellationToken = cancellationTokenSource.token;
 
     const execution = new CliCommandExecutor(this.build(response.data), {
-      cwd: getRootWorkspacePath()
+      cwd: workspaceUtils.getRootWorkspacePath()
     }).execute(cancellationToken);
 
     execution.processExitSubject.subscribe(() =>
--- a/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
@@ -5,10 +5,14 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode/out/src';
-import { notificationService } from '@salesforce/salesforcedx-utils-vscode/out/src/commands';
-import { CancelResponse, ContinueResponse, ParametersGatherer } from '@salesforce/salesforcedx-utils-vscode/src/types';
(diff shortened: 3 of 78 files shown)


packages/salesforcedx-vscode-core/src/commands/auth/authParamsGatherer.ts
```diff
@@ -11,7 +11,7 @@ import {
   CancelResponse,
```
gbockus-sf, line 11: This PR is big, but mostly just prettier and module name changes.

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
```diff
@@ -18,11 +18,11 @@ import {
   SfdxWorkspaceChecker
 } from '../util';
 
-import { getRootWorkspacePath } from '../../util';
+import { workspaceUtils } from '../../util';
 
 import { ConfigFile } from '@salesforce/core';
-import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
-import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode';
```
jeffb-sfdc, line 24: `isNullOrUndefined` seems generic and not vs-specific.  Can it be moved from salesforcedx-utils-vscode to salesforcedx-utils?

jeffb-sfdc: maybe address in a follow up ticket?

packages/salesforcedx-vscode-core/src/commands/forceRenameLightningComponent.ts
```diff
@@ -93,7 +99,10 @@ export class GetComponentName
   }
 }
 
-export async function inputGuard(sourceFsPath: string, newName: string): Promise<string> {
+export async function inputGuard(
+  sourceFsPath: string,
+  newName: string
+): Promise<string> {
```
jeffb-sfdc, line 105: I noticed something here.  Is `newName` a string, or is it a `Promise<string>`?  Seems like it's just a string, and `inputGuard` should be changed to:
```
export async function inputGuard(
  sourceFsPath: string,
  newName: string
): string {
```

jeffb-sfdc: nm...

packages/salesforcedx-vscode-core/src/commands/templates/baseTemplateCommand.ts
```diff
@@ -39,18 +34,23 @@ export abstract class BaseTemplateCommand extends SfdxCommandletExecutor<
...
     execution.processExitSubject.subscribe(async data => {
       this.logMetric(execution.command.logName, startTime, {
         dirType: this.identifyDirType(response.data.outputdir)
       });
-      if (data !== undefined && String(data) === '0' && hasRootWorkspace()) {
-        const outputFile = this.getPathToSource(response.data.outputdir, response.data.fileName);
-        const document = await vscode.workspace.openTextDocument(
-          outputFile
+      if (
+        data !== undefined &&
```
jeffb-sfdc, line 45: how about changing this from `data !== undefined &&` to `data &&`

packages/salesforcedx-vscode-core/src/commands/util/preconditionCheckers.ts
```diff
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
+import { isNullOrUndefined } from '@salesforce/salesforcedx-utils-vscode';
```
jeffb-sfdc, line 7: same comment - isNullOrUndefined seems generic and should be in salesforcedx-utils

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -10,10 +10,14 @@ import * as os from 'os';
...
 import { nls } from '../messages';
-import { getRootWorkspacePath, hasRootWorkspace } from '../util';
+import { workspaceUtils } from '../util';
 
-const CONFIG_FILE = hasRootWorkspace()
-  ? path.join(getRootWorkspacePath(), '.sfdx', 'sfdx-config.json')
+const CONFIG_FILE = workspaceUtils.hasRootWorkspace()
+  ? path.join(
+      workspaceUtils.getRootWorkspacePath(),
+      '.sfdx',
```
jeffb-sfdc, line 18: Is it OK to use a path with `.sfdx` here, or is this another case where we need to call an API?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -43,7 +43,7 @@ export class ComponentUtils {
       folderName ? `${metadataType}_${folderName}` : metadataType
     }.json`;
     const componentsPath = path.join(
-      getRootWorkspacePath(),
+      workspaceUtils.getRootWorkspacePath(),
```
jeffb-sfdc, line 46: same comment as before... is it OK to use a path with `.sfdx` here, or is there an API we should use?

jeffb-sfdc: If we stick with sfdx, then to match the other code this should be added:
`import { SFDX_PROJECT_FILE } from '../constants';`

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
jeffb-sfdc: If we stick with sfdx, then to match the other code this should be added:
`import { SFDX_PROJECT_FILE } from '../constants';`

packages/salesforcedx-vscode-core/src/commands/auth/forceAuthDevHub.ts
gbockus-sf: 100% agree. Will add it to my list of follow ups

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
gbockus-sf: yup, ok for now.  We have a bunch of work items to address all these in the core v3 epic.

jeffb-sfdc (approved):

PR #4429, opened 2022-09-15
chore: jest configuration and first set of test file updates Part 3

gbockus-sf:
### What does this PR do?
The jest configuration for each module. Includes unit and integration configs where applicable. 
The first set up test updates that is larges just updating the utils path.

PR #4433, opened 2022-09-15
chore: changes to non-core packags part 7

gbockus-sf:
### What does this PR do?
Code updates to all packages but the core package.

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -6,15 +6,13 @@
  */
 
 import {
+  extractJsonObject,
   ForceConfigGet,
-  ForceOrgDisplay
-} from '@salesforce/salesforcedx-utils-vscode/out/src/cli';
-import { extractJsonObject } from '@salesforce/salesforcedx-utils-vscode/out/src/helpers';
-import { RequestService } from '@salesforce/salesforcedx-utils-vscode/out/src/requestService';
-import {
+  ForceOrgDisplay,
+  RequestService,
   SFDX_CONFIG_ISV_DEBUGGER_SID,
   SFDX_CONFIG_ISV_DEBUGGER_URL
-} from '@salesforce/salesforcedx-utils-vscode/out/src/types';
+} from '@salesforce/salesforcedx-utils';
 import * as AsyncLock from 'async-lock';
 import { basename } from 'path';
 import {
@@ -548,7 +546,7 @@ export class ApexDebug extends LoggingDebugSession {
 
   private lock = new AsyncLock({ timeout: DEFAULT_LOCK_TIMEOUT_MS });
 
-  protected idleTimers: NodeJS.Timer[] = [];
+  protected idleTimers: Array<ReturnType<typeof setTimeout>> = [];
 
   constructor() {
     super('apex-debug-adapter.log');
@@ -590,8 +588,8 @@ export class ApexDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
-  private getSessionIdleTimer(): NodeJS.Timer[] {
-    const timers: NodeJS.Timer[] = [];
+  private getSessionIdleTimer(): Array<ReturnType<typeof setTimeout>> {
+    const timers: Array<ReturnType<typeof setTimeout>> = [];
     timers.push(
       setTimeout(() => {
         this.warnToDebugConsole(
@@ -637,7 +635,7 @@ export class ApexDebug extends LoggingDebugSession {
     }
   }
 
-  public resetIdleTimer(): NodeJS.Timer[] {
+  public resetIdleTimer(): Array<ReturnType<typeof setTimeout>> {
     this.clearIdleTimers();
     this.idleTimers = this.getSessionIdleTimer();
     return this.idleTimers;
@@ -883,7 +881,7 @@ export class ApexDebug extends LoggingDebugSession {
         await this.myRequestService.execute(new RunCommand(requestId));
         response.success = true;
       } catch (error) {
-        response.message = error;
+        response.message = error.message;
       }
     }
     this.resetIdleTimer();
@@ -901,7 +899,7 @@ export class ApexDebug extends LoggingDebugSession {
         await this.myRequestService.execute(new StepOverCommand(requestId));
         response.success = true;
       } catch (error) {
-        response.message = error;
+        response.message = error.message;
       }
     }
     this.sendResponse(response);
@@ -918,7 +916,7 @@ export class ApexDebug extends LoggingDebugSession {
         await this.myRequestService.execute(new StepIntoCommand(requestId));
         response.success = true;
       } catch (error) {
-        response.message = error;
+        response.message = error.message;
       }
     }
     this.sendResponse(response);
@@ -935,7 +933,7 @@ export class ApexDebug extends LoggingDebugSession {
         await this.myRequestService.execute(new StepOutCommand(requestId));
         response.success = true;
       } catch (error) {
-        response.message = error;
+        response.message = error.message;
       }
     }
     this.sendResponse(response);
@@ -1053,7 +1051,7 @@ export class ApexDebug extends LoggingDebugSession {
       response.body = { stackFrames: clientFrames };
       response.success = true;
     } catch (error) {
-      response.message = error;
+      response.message = error.message;
     }
     this.sendResponse(response);
   }
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -219,9 +219,11 @@ export abstract class LibraryCommandletExecutor<T>
         measurements
       );
     } catch (e) {
-      telemetryService.sendException(e.name, e.message);
-      notificationService.showFailedExecution(this.executionName);
-      channelService.appendLine(e.message);
+      if (e instanceof Error) {
+        telemetryService.sendException(e.name, e.message);
+        notificationService.showFailedExecution(this.executionName);
+        channelService.appendLine(e.message);
+      }
       channelService.showChannelOutput();
     }
   }
--- a/packages/salesforcedx-utils-vscode/src/config/configUtil.ts
+++ b/packages/salesforcedx-utils-vscode/src/config/configUtil.ts
@@ -5,7 +5,12 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { ConfigAggregator, ConfigFile, ConfigValue } from '@salesforce/core';
+import {
+  ConfigAggregator,
+  ConfigFile,
+  ConfigValue,
+  OrgConfigProperties
+} from '@salesforce/core';
 import * as path from 'path';
 import { TelemetryService } from '../telemetry/telemetry';
 import { getRootWorkspacePath } from '../workspaces';
@@ -78,5 +83,30 @@ export class ConfigUtil {
       }
     }
     return undefined;
+  }
+
+  public static async getUsername(
+    projectPath: string
+  ): Promise<string | undefined> {
+    const configAggregator = await ConfigUtil.getConfigAggregator(projectPath);
+    const defaultUsernameOrAlias = configAggregator.getPropertyValue(
+      OrgConfigProperties.TARGET_ORG
+    );
+    return (defaultUsernameOrAlias as string) || undefined;
+  }
+
+  private static async getConfigAggregator(
+    projectPath: string
+  ): Promise<ConfigAggregator> {
+    const origCurrentWorkingDirectory = process.cwd();
+    // Change the current working directory to the project path,
+    // so that ConfigAggregator reads the local project values
+    process.chdir(projectPath);
+    const configAggregator = await ConfigAggregator.create();
+    await configAggregator.reload();
+    // Change the current working directory back to what it was
+    // before returning
+    process.chdir(origCurrentWorkingDirectory);
+    return configAggregator;
   }
 }
--- a/packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
+++ b/packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
@@ -13,7 +13,7 @@ import { nls } from '../messages';
 import { SFDX_CONFIG_FILE, SFDX_FOLDER } from '../types';
 import { getRootWorkspacePath } from '../workspaces';
 
-export interface OrgInfo {
+export interface OrgUserInfo {
   username?: string;
   alias?: string;
 }
@@ -26,15 +26,15 @@ export class WorkspaceContextUtil {
 
   protected cliConfigWatcher: vscode.FileSystemWatcher;
   protected sessionConnections: Map<string, Connection>;
-  protected onOrgChangeEmitter: vscode.EventEmitter<OrgInfo>;
+  protected onOrgChangeEmitter: vscode.EventEmitter<OrgUserInfo>;
   protected _username?: string;
   protected _alias?: string;
 
-  public readonly onOrgChange: vscode.Event<OrgInfo>;
+  public readonly onOrgChange: vscode.Event<OrgUserInfo>;
 
   protected constructor() {
     this.sessionConnections = new Map<string, Connection>();
-    this.onOrgChangeEmitter = new vscode.EventEmitter<OrgInfo>();
+    this.onOrgChangeEmitter = new vscode.EventEmitter<OrgUserInfo>();
     this.onOrgChange = this.onOrgChangeEmitter.event;
 
     const bindedHandler = () => this.handleCliConfigChange();
@@ -49,6 +49,10 @@ export class WorkspaceContextUtil {
     this.cliConfigWatcher.onDidChange(bindedHandler);
     this.cliConfigWatcher.onDidCreate(bindedHandler);
     this.cliConfigWatcher.onDidDelete(bindedHandler);
+  }
+
+  public static getLogDirPath(): string {
+    return join(getRootWorkspacePath(), '.sfdx', 'tools', 'debug', 'logs');
   }
 
   public getAuthUtil(): AuthUtil {
@@ -114,8 +118,4 @@ export class WorkspaceContextUtil {
   get alias(): string | undefined {
     return this._alias;
   }
-}
-
-export function getLogDirPath(): string {
-  return join(getRootWorkspacePath(), '.sfdx', 'tools', 'debug', 'logs');
 }
--- a/packages/salesforcedx-utils-vscode/src/date/format.ts
+++ b/packages/salesforcedx-utils-vscode/src/date/format.ts
@@ -19,7 +19,7 @@ export function makeDoubleDigit(currentDigit: number): string {
   return ('0' + currentDigit).slice(-2);
 }
 
-export const optionYYYYMMddHHmmss = {
+export const optionYYYYMMddHHmmss: Intl.DateTimeFormatOptions = {
   year: 'numeric',
   month: '2-digit',
   day: '2-digit',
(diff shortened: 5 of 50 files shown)


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -548,7 +546,7 @@ export class ApexDebug extends LoggingDebugSession {
 
   private lock = new AsyncLock({ timeout: DEFAULT_LOCK_TIMEOUT_MS });
 
-  protected idleTimers: NodeJS.Timer[] = [];
+  protected idleTimers: Array<ReturnType<typeof setTimeout>> = [];
```
gbockus-sf, line 549: another spot where jest tests make a different.  This was the recommending type for NodeJS.Timer.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -79,4 +84,29 @@ export class ConfigUtil {
...
+    // Change the current working directory to the project path,
+    // so that ConfigAggregator reads the local project values
+    process.chdir(projectPath);
+    const configAggregator = await ConfigAggregator.create();
+    await configAggregator.reload();
+    // Change the current working directory back to what it was
+    // before returning
+    process.chdir(origCurrentWorkingDirectory);
+    return configAggregator;
+  }
```
gbockus-sf, line 111: This is the commit in faux the started all this noise.  Fixed in #4422

gbockus-sf: Will merge immediately following the large PR.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -13,7 +13,7 @@ import { nls } from '../messages';
 import { SFDX_CONFIG_FILE, SFDX_FOLDER } from '../types';
 import { getRootWorkspacePath } from '../workspaces';
 
-export interface OrgInfo {
+export interface OrgUserInfo {
```
gbockus-sf, line 16: We had two interfaces called OrgInfo. renamed this one to avoid the conflict.

packages/salesforcedx-utils-vscode/src/date/format.ts
```diff
@@ -28,12 +28,12 @@ export const optionYYYYMMddHHmmss = {
...
 };
 
-export const optionHHmm = {
+export const optionHHmm: Intl.DateTimeFormatOptions = {
   hour: '2-digit',
   minute: '2-digit'
 };
 
-export const optionMMddYYYY = {
+export const optionMMddYYYY: Intl.DateTimeFormatOptions = {
```
gbockus-sf, line 36: another typescript update beef.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
       return false;
     }
   }
```
gbockus-sf, line 424: just a prettier change.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -270,6 +270,8 @@ export abstract class TestNode extends vscode.TreeItem {
     dark: DARK_BLUE_BUTTON
   };
 
+  // TODO: create a ticket to address this particular issue.
+  // @ts-ignore
```
gbockus-sf, line 274: this was a weird case with extending a class from another module.  I didn't find a good work around so I went the heavy handed route.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -590,8 +588,8 @@ export class ApexDebug extends LoggingDebugSession {
     this.sendResponse(response);
   }
 
-  private getSessionIdleTimer(): NodeJS.Timer[] {
-    const timers: NodeJS.Timer[] = [];
+  private getSessionIdleTimer(): Array<ReturnType<typeof setTimeout>> {
+    const timers: Array<ReturnType<typeof setTimeout>> = [];
```
jeffb-sfdc, line 592: @gbockus-sf the syntax of `Array<ReturnType<typeof setTimeout>>` is more verbose, and (IMO) isn't as easy to grok as `NodeJS.Timer[]`.  I changed the type here back to `NodeJS.Timer[]` and that worked (for getSessionIdleTimer(), at least).  If not needed, my preference would be to  go back to NodeJS.Timer[].  I suspect thought there might have been a reason.  If so, what was it?

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -219,9 +219,11 @@ export abstract class LibraryCommandletExecutor<T>
         measurements
       );
     } catch (e) {
-      telemetryService.sendException(e.name, e.message);
-      notificationService.showFailedExecution(this.executionName);
-      channelService.appendLine(e.message);
+      if (e instanceof Error) {
```
jeffb-sfdc, line 222: This is something to watch out for.  I did a search, and it looks like Error object are being thrown everywhere, except 3 places:
packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts, line 40
packages/salesforcedx-sobjects-faux-generator/src/generator/fauxClassGenerator.ts, line 68
packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts, line 65

If these 3 places were updated to `throw new Error(string)` instead of `throw string`, the check for `e instanceof Error` wouldn't bee needed.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -51,6 +51,10 @@ export class WorkspaceContextUtil {
     this.cliConfigWatcher.onDidDelete(bindedHandler);
   }
 
+  public static getLogDirPath(): string {
+    return join(getRootWorkspacePath(), '.sfdx', 'tools', 'debug', 'logs');
```
jeffb-sfdc, line 55: Should the const for SFDX be used here?

packages/salesforcedx-utils-vscode/src/telemetry/configUtil.ts
```diff
@@ -52,10 +52,12 @@ export class ConfigUtil {
           return localValue;
         }
       } catch (err) {
-        TelemetryService.getInstance().sendException(
-          'get_config_value_local',
-          err.message
-        );
+        if (err instanceof Error) {
```
jeffb-sfdc, line 55: same comment as before, after converting those three instances, we probably don't need to check if err is an instance of Error

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+  public static async sfdxCreateCheckpoints(): Promise<boolean> {
+    // In-spite of waiting for the lock, we still want subsequent calls to immediately return
+    // from this if checkpoints are already being created instead of stacking them up.
+    if (!creatingCheckpoints) {
+      creatingCheckpoints = true;
+    } else {
+      return false;
+    }
+    let updateError = false;
+    // The status message isn't changing, call to localize it once and use the localized string in the
```
jeffb-sfdc, line 443: A few comments about `sfdxCreateCheckpoints()`
* Most of the comments don't seem to be needed, as they comment what is obvious on the next line, and should be removed
* This function is pretty large, and it would be better if it was broken up into smaller functions
* Unit tests?

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+            writeToDebuggerOutputWindow(
+              `${localizedProgressMessage}, ${nls.localize(
+                'checkpoint_creation_status_org_info'
+              )}`
+            );
+            progress.report({
+              increment: 0,
+              message: localizedProgressMessage
+            });
+            const orgInfoRetrieved: boolean = await checkpointService.retrieveOrgInfo();
```
jeffb-sfdc, line 472: The `: boolean` doesn't need to be added

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+              `${localizedProgressMessage}, ${nls.localize(
+                'checkpoint_creation_status_uploading_checkpoints'
+              )}`
+            );
+            progress.report({
+              increment: 70,
+              message: localizedProgressMessage
+            });
+            // This should probably be batched but it makes dealing with errors kind of a pain
+            for (const cpNode of checkpointService.getChildren() as CheckpointNode[]) {
```
jeffb-sfdc, line 541: I'm not a fan of abbreviating variables, even when the code is only a few lines line (as a reader it always makes me have to got back and look at what it is).  How about instead of `cpNode` it be named `checkpoiuntNode`?

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+            writeToDebuggerOutputWindow(
+              `${localizedProgressMessage}, ${nls.localize(
+                'checkpoint_creation_status_processing_complete_success'
+              )}`
+            );
+          }
+        );
+      });
+    } finally {
+      writeToDebuggerOutputWindow(
```
jeffb-sfdc, line 566: Even here, what's in the `finally` could be refactored into its own function.
Also, I think the syntax for finally is:
```
.finally(() => {
    // code
});
```

I haven't verified this, but that's the syntax I'm used to seeing.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+            writeToDebuggerOutputWindow(
+              `${localizedProgressMessage}, ${nls.localize(
+                'checkpoint_creation_status_source_line_info'
+              )}`
+            );
+            progress.report({
+              increment: 20,
+              message: localizedProgressMessage
+            });
+            const sourceLineInfoRetrieved: boolean = await retrieveLineBreakpointInfo();
```
jeffb-sfdc, line 487: The `:boolean` isn't needed

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+        errorMsg = nls.localize(
+          'checkpoint_upload_error_wrap_up_message',
+          nls.localize('sfdx_update_checkpoints_in_org')
+        );
+        writeToDebuggerOutputWindow(errorMsg, true, VSCodeWindowTypeEnum.Error);
+      }
+      telemetryService.sendCheckpointEvent(errorMsg);
+      creatingCheckpoints = false;
+    }
+    if (updateError) {
```
jeffb-sfdc, line 580: `return !updateError` ?

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+      let errorMsg = '';
+      if (updateError) {
+        errorMsg = nls.localize(
+          'checkpoint_upload_error_wrap_up_message',
+          nls.localize('sfdx_update_checkpoints_in_org')
+        );
+        writeToDebuggerOutputWindow(errorMsg, true, VSCodeWindowTypeEnum.Error);
+      }
+      telemetryService.sendCheckpointEvent(errorMsg);
+      creatingCheckpoints = false;
```
jeffb-sfdc, line 578: yikes!  `creatingCheckpoints` (defined on line 949) is essentially a global. (meaning, it is used to control logic & flow in several functions, yet aren't local to the functions.  Not of your doing, I know.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -426,6 +422,166 @@ export class CheckpointService implements TreeDataProvider<BaseNode> {
...
+              updateError = true;
+              return false;
+            }
+
+            writeToDebuggerOutputWindow(
+              `${localizedProgressMessage}, ${nls.localize(
+                'checkpoint_creation_status_clearing_existing_checkpoints'
+              )}`
+            );
+            progress.report({
```
jeffb-sfdc, line 520: I'm not familiar with progress.report().  Is `increment` "increment by" or "increment to"?
I looks like it's probably "increment to".  If that's the case, the progress is being set to 50% on lines 506 and 521.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -791,142 +947,6 @@ function setTypeRefsForEnabledCheckpoints(): boolean {
 // 3. Remove any existing checkpoints
 // 4. Create the new checkpoints
 let creatingCheckpoints = false;
-export async function sfdxCreateCheckpoints(): Promise<boolean> {
```
jeffb-sfdc, line 794: this threw me.  it seems sfdxCreateCheckpoints() was moved.  (why?)  if it hadn't been moved it wouldn't have appeared as a new function above.  I read your comment about a prettier format change, but didn't see code that matched

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
gbockus-sf: I was originally running into a compile issue (or maybe it was when having ts-jest do the compilation although using the same tsconfig so I dunno why it was different) I just tired changing it back and it compiled and the tests passed without issue locally.  I'll see if it was specific to the CI build or something 🤷 I grabbed the change from https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript but honestly don't remember the specifics of how/why it was causing an issue.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
gbockus-sf: Good call. Yeah I think at least one of those places I originally checked for an Error and ran into the issue of it being a string.  I'll add an item to the list for having the linter only allow us to throw errors. I started adding this check b/c with the typescript update they tightened up on assuming Errors in catch blocks (see [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#defaulting-to-the-unknown-type-in-catch-variables---useunknownincatchvariables).  I ended up deciding there were too many touches for this and ran into the issue in faux generate so I turned off the check in the tsconfig and added a note to follow up.  Will also ensure we have a ticket to address this.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
gbockus-sf: Ordinarily I'd agree but these will all be updated with the core v3 work so I was cool with leaving it.

packages/salesforcedx-utils-vscode/src/telemetry/configUtil.ts
gbockus-sf: agreed. see https://github.com/forcedotcom/salesforcedx-vscode/pull/4433#discussion_r976036460

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
gbockus-sf: Agreed on all points.  (and those made below).  I'll create a ticket and reference this PR.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
gbockus-sf: See this [test](https://github.com/forcedotcom/salesforcedx-vscode/pull/4429/files#r976041850).  It's the same issue where a stub wasn't working for an exported function so in order to get around it I moved the function to a static method on the class.  Sorry for the misdirection on the prettier comment I didn't look closely enough on my initial review of the change.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
gbockus-sf: Good q, I had no idea but looking at the docs it appears to sum the increment
```
To report discrete progress, use `increment` to indicate how much work has been completed. Each call with
         * a `increment` value will be summed up and reflected as overall progress until 100% is reached (a value of
         * e.g. `10` accounts for `10%` of work done).
         * 
         * 
```
From `packages/salesforcedx-utils-vscode/node_modules/@types/vscode/index.d.ts`

Honestly I don't ever recall progress ever being reported for any of our operations so it makes me wonder if it's even working as expected in the product today.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
gbockus-sf: yeah this whole file needs a hard look for sure.  14 imports, 18 exports.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
gbockus-sf: Ah the dreaded only fails on the CI build unit test 
```
FAIL test/unit/adapter/apexDebugVariablesHandling.test.ts
@salesforce/salesforcedx-apex-debugger:   ● Test suite failed to run
@salesforce/salesforcedx-apex-debugger:     src/adapter/apexDebug.ts:594:7 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'Timer'.
@salesforce/salesforcedx-apex-debugger:     594       setTimeout(() => {
@salesforce/salesforcedx-apex-debugger:               ~~~~~~~~~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     595         this.warnToDebugConsole(
@salesforce/salesforcedx-apex-debugger:         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     ... 
@salesforce/salesforcedx-apex-debugger:     601         );
@salesforce/salesforcedx-apex-debugger:         ~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     602       }, DEFAULT_IDLE_WARN1_MS),
@salesforce/salesforcedx-apex-debugger:         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger: FAIL test/unit/adapter/apexDebug.test.ts
@salesforce/salesforcedx-apex-debugger:   ● Test suite failed to run
@salesforce/salesforcedx-apex-debugger:     src/adapter/apexDebug.ts:594:7 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'Timer'.
@salesforce/salesforcedx-apex-debugger:     594       setTimeout(() => {
@salesforce/salesforcedx-apex-debugger:               ~~~~~~~~~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     595         this.warnToDebugConsole(
@salesforce/salesforcedx-apex-debugger:         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     ... 
@salesforce/salesforcedx-apex-debugger:     601         );
@salesforce/salesforcedx-apex-debugger:         ~~~~~~~~~~
@salesforce/salesforcedx-apex-debugger:     602       }, DEFAULT_IDLE_WARN1_MS),
```
The updated type was the fix for these tests I found.  Would a code comment about why that type is used be useful?

jeffb-sfdc (approved):

PR #4441, opened 2022-09-21
fix: support for renaming aura event bundles

CristiCanizales:
### What does this PR do?
- Adds the .evt extensions to the ones supported for aura components

[@W-11786256@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000016sDxdYAE/view)

### Functionality Before
- sfdx Rename Component command doesn't work for Aura Event: only the bundle is renamed, not the files inside

### Functionality After
- sfdx Rename Component works for aura event: the bundle and the files inside end up with the new name given

### DIFF
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
@@ -14,7 +14,7 @@ const auraPath = vscode.Uri.parse('/force-app/main/default/aura/');
 const lwcComponent = 'hero';
 const auraComponent = 'page';
 const itemsInHero = ['hero.css', 'hero.html', 'hero.js', 'hero.js-meta.xml', 'templateOne.html'];
-const itemsInPage = ['page.auradoc', 'page.cmp', 'page.cmp-meta.xml', 'page.css', 'page.design', 'page.svg', 'pageController.js', 'pageHelper.js', 'pageRenderer.js', 'templateOne.css'];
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceRenameLightningComponent.test.ts
```diff
@@ -286,6 +286,8 @@ describe('Force Rename Lightning Component', () => {
       expect(isNameMatch(itemsInPage[6], componentName, componentPath)).to.equal(true);
       expect(isNameMatch(itemsInPage[7], componentName, componentPath)).to.equal(true);
       expect(isNameMatch(itemsInPage[8], componentName, componentPath)).to.equal(true);
+      expect(isNameMatch(itemsInPage[9], componentName, componentPath)).to.equal(true);
+      expect(isNameMatch(itemsInPage[10], componentName, componentPath)).to.equal(true);
     });
 
     it('should return false if file type is not in LWC or Aura or file name and component name do not match', () => {
```
gbockus-sf, line 293: Can you update this test as well to add a negative test case like `handleEvt.js`. I could see it being common for people to name things with Evt.

gbockus-sf (commented):
good stuff.  I can QE this change.

gbockus-sf (approved):
Thanks!
Good work

Verify I can rename an Aura event ✅

* right click existing aura event and select SFDX: Rename Component
* provide new name 
* default location 
* Success. Both the .evt and .evt-meta.xml file are renamed.

Verify Error if I attempt a name that already exists ✅

* Attempt to rename to existing name 
* Fails with error message Unable to rename the component. Try renaming the component manually and then redeploying your changes.
* No changes are made

Verify rename with same name ✅

* Fail with same error as using name already in use

PR #4455, opened 2022-09-28
Replace '.sdfx' directory references with core enum in apex package

klewis-sfdc:
### What does this PR do?
To prevent potential disruption as the internals of the .sf and .sfdx evolve, use the enums from core instead of using the string reference to the state folder.

@W-11633781@, @W-11633776@

### Functionality Before
The state folder was referenced via string value ".sfdx"

### Functionality After
The state folder is referenced via imported STATE_FOLDER enum from core: https://github.com/forcedotcom/sfdx-core/blob/a6e10b250ed8b4c23ec13f2c7f15ef218bd52135/src/global.ts#L33

Note: PR order: apex <-- lwc <-- replay-debugger


packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -64,11 +65,9 @@ async function createServer(
...
       args.push(`-Xmx${jvmMaxHeap}M`);
     }
-    telemetryService.sendEventData(
-      'apexLSPSettings',
-      undefined,
-      { maxHeapSize: jvmMaxHeap != null ? jvmMaxHeap : 0 }
-    );
+    telemetryService.sendEventData('apexLSPSettings', undefined, {
+      maxHeapSize: jvmMaxHeap != null ? jvmMaxHeap : 0
+    });
```
klewis-sfdc, line 70: auto format

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -64,11 +65,9 @@ async function createServer(
     if (jvmMaxHeap) {
       args.push(`-Xmx${jvmMaxHeap}M`);
     }
-    telemetryService.sendEventData(
-      'apexLSPSettings',
-      undefined,
-      { maxHeapSize: jvmMaxHeap != null ? jvmMaxHeap : 0 }
-    );
+    telemetryService.sendEventData('apexLSPSettings', undefined, {
```
randi274, line 68: @klewis-sfdc was this formatting change intentional? I thought it was a bit easier to read before

klewis-sfdc: No, it just happened when I saved the file.  Our formatter thinks this is the correct way I guess

packages/salesforcedx-vscode-apex/test/vscode-integration/commands/forceApexExecute.test.ts
```diff
@@ -5,15 +5,20 @@
...
+} from '@salesforce/core';
 import { MockTestOrgData, testSetup } from '@salesforce/core/lib/testSetup';
-import { ChannelService } from '@salesforce/salesforcedx-utils-vscode';
 import {
+  ChannelService,
+  ContinueResponse,
   getRootWorkspacePath,
-  SFDX_CORE_CONFIGURATION_NAME
+  SFDX_CORE_CONFIGURATION_NAME,
+  TraceFlags
```
klewis-sfdc, line 20: consolidated the import statements

packages/salesforcedx-vscode-apex/src/codecoverage/colorizer.ts
```diff
@@ -6,6 +6,7 @@
  */
 
 import { CodeCoverageResult } from '@salesforce/apex-node';
+import { Global } from '@salesforce/core';
```
gbockus-sf, line 9: Dumb q: should we create a place in utils-vscode for getting all the things from core like this? Maybe a `PathService` that will give us path locations similar to how we're using ConfigUtils today?   Has utility functions that looks something like `PathService.getApexPath();`
That way we can limit core to just the utils packages

randi274: 💯 yes please

klewis-sfdc: Yep!  Will wait for PathService to be introduced in @RitamAgrawal's PR here: https://github.com/forcedotcom/salesforcedx-vscode/pull/4408/files and once that is merged, will add functions for them there.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -102,19 +101,14 @@ export function setupDB(): void {
...
   ) {
-    const dbPath = path.join(
-      vscode.workspace.workspaceFolders[0].uri.fsPath,
-      '.sfdx',
-      'tools',
-      'apex.db'
-    );
-    if (fs.existsSync(dbPath)) {
+    const dbPath = projectPaths.getApexLanguageServerDatabasePath();
+    if (dbPath && fs.existsSync(dbPath)) {
```
klewis-sfdc, line 105: TS beefed about this being potentially undefined, hence the additional conditionals.

packages/salesforcedx-vscode-apex/test/vscode-integration/commands/forceApexExecute.test.ts
```diff
@@ -86,7 +91,7 @@ describe('Force Apex Execute', () => {
     it('should return the selected file to execute anonymous apex', async () => {
       const fileName = path.join(
         getRootWorkspacePath(),
-        '.sfdx',
+        Global.STATE_FOLDER,
```
klewis-sfdc, line 94: todo: update tests to not import from core

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -60,28 +64,55 @@ export function getRelativeProjectPath(
...
-    getSfdxDirectoryPath(),
-    'orgs',
-    username,
-    'metadata'
+function apexTestResults(): string {
+  const apexTestResultsFolder = path.join(
+    stateFolder(),
+    'tools',
+    'testresults',
+    'apex'
```
gbockus-sf, line 94: I'd like to see these strings moved up to consts at the top of this file.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -60,28 +64,55 @@ export function getRelativeProjectPath(
...
   );
+  return apexTestResultsFolder;
+}
+
+function apexLanguageServerDatabase(): string | undefined {
+  if (!hasRootWorkspace()) {
+    return undefined;
+  }
+  const apexLangServerDbPath = path.join(stateFolder(), 'tools', 'apex.db');
```
gbockus-sf, line 103: interesting...is this a path or a file? 
more opportunity to create sting consts (one resuable!).

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -60,28 +64,55 @@ export function getRelativeProjectPath(
...
+  return apexLangServerDbPath;
+}
+
+function debugLogs(): string | undefined {
+  const logsDirectory = path.join(stateFolder(), 'tools', 'debug', 'logs');
```
gbockus-sf, line 108: consts.

packages/salesforcedx-vscode-apex/src/languageServer.ts
gbockus-sf: Should our projectPaths functions never return undefined?  Throwing an error if conditions aren't met to build the path seems like it'd be fine.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
```diff
@@ -101,21 +102,21 @@ describe('get metadata components path', () => {
...
     const metadataType = 'Report';
     const folder = 'TestFolder';
-    expect(
-      await cmpUtil.getComponentsPath(metadataType, alias, folder)
-    ).to.equal(expectedPath(metadataType + '_' + folder));
-    expect(getMetadataDirectoryPathStub.called).to.equal(true);
-    expect(getMetadataDirectoryPathStub.calledWith(username)).to.equal(true);
+    expect(await cmpUtil.getComponentsPath(metadataType, folder)).to.equal(
+      expectedPath(metadataType + '_' + folder)
+    );
```
gbockus-sf, line 117: nit: I'd typically assign the values under test to variables here instead of inlining two calls (one async) in an expect.
```
const compPath = await cmpUtil.getComponentsPath(metadataType, folder);
const expectedPath = expectedPath(metadataType + '_' + folder);
expect(compPath).to.equal(expectedPath);
```
makes it easier to grok a failure and debug IMO

gbockus-sf (commented):
great stuff.  A few small requests

packages/salesforcedx-vscode-apex/src/languageServer.ts
klewis-sfdc: This block is taking action only if there is already a path to the Apex language server database. So it checks and then if there is a path, it tries to delete the existing resource. If we changed it to throw when it tried to get the path to the database, then we would have to wrap this in a try catch and specifically handle the thrown exception, and then continue with execution. Seems to be like it would be more complicated, but open to further discussion.

gbockus-sf: hum...so apexLanguageServerDatabase only has the undefined type if getRootWorkspace returns undefined. I don't think we should ever get to this point if we don't have a root workspace so shouldn't we be ok to throw in that case and remove the undefined return type.

klewis-sfdc: Ok yeah that makes sense.  I updated it.

gbockus-sf (commented):
Good stuff.  Code looks good.  I'll QE tonight

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -110,9 +111,9 @@ function apexTestResultsFolder(): string {
   return pathToApexTestResultsFolder;
 }
 
-function apexLanguageServerDatabase(): string | undefined {
+function apexLanguageServerDatabase(): string {
   if (!hasRootWorkspace()) {
-    return undefined;
+    throw new Error(nls.localize('error_no_root_workspace_found'));
```
gbockus-sf: sweeeeeeet

gbockus-sf (approved):
QE: 
- get tools folder ✅ 
- Verify when in a SF workspace folders are found ✅ 
- Verified correct Log directory is supplied for get Apex debug logs flow. ✅ 
- Verified when not in a SF project extensions don't run so not logs. ✅

gbockus-sf:
merging b/c @klewis-sfdc is OOO today

PR #4457, opened 2022-09-28
Replace '.sdfx' directory references with core enum in lwc package

klewis-sfdc:
### What does this PR do?
To prevent potential disruption as the internals of the .sf and .sfdx evolve, use the enums from core instead of using the string reference to the state folder.

@W-11633619@

### Functionality Before
The state folder was referenced via string value ".sfdx"

### Functionality After
The state folder is referenced via imported STATE_FOLDER enum from core: https://github.com/forcedotcom/sfdx-core/blob/a6e10b250ed8b4c23ec13f2c7f15ef218bd52135/src/global.ts#L33

Note: PR order: apex <-- lwc <-- replay-debugger


packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -104,6 +104,37 @@ function apexLanguageServerDatabase(): string | undefined {
   return apexLangServerDbPath;
 }
 
+function lwcTestResults(expectedCwd: string): string {
+  // todo: should this use getRootWorkspacePath instead?
```
gbockus-sf, line 108: as discussed..yup for sure.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -104,6 +104,37 @@ function apexLanguageServerDatabase(): string | undefined {
...
+    'testresults'
+  );
+  return testResultsDirPath;
+}
+
+function getApexLanguageServerDatabasePath(): string | undefined {
+  if (!vscode.workspace.workspaceFolders) {
+    return undefined;
+  }
+  const apexLangServerDbPath = path.join(stateFolder(), 'tools', 'apex.db');
```
gbockus-sf, line 134: use consts

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -104,6 +104,37 @@ function apexLanguageServerDatabase(): string | undefined {
...
+  const testResultsDirPath = path.join(
+    vscodePath,
+    Global.STATE_FOLDER,
+    'tools',
+    'testresults'
+  );
+  return testResultsDirPath;
+}
+
+function getApexLanguageServerDatabasePath(): string | undefined {
```
gbockus-sf, line 130: kinda think we should throw an error here instead of returning undefined. This method doesn't seem like the right place to verify there is a workspaceFolder (that should be done by the caller I'd think).  We should never hit this code if there is no project open b/c the apex extension should never load I think.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
klewis-sfdc: I'm getting a little hung up on trying to remove this expectedCwd param.  It looks to me like the current root ws path is the only thing that is passed in, but some tests fail when I remove it, and the tests are fairly complex so I'm trying to figure out what else needs to be updated to make the tests work.  I'll keep looking at this

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -112,13 +113,30 @@ function apexTestResultsFolder(): string {
 }
 
 function apexLanguageServerDatabase(): string {
-  if (!hasRootWorkspace()) {
-    throw new Error(nls.localize('cannot_determine_workspace'));
-  }
```
klewis-sfdc, line 117: I removed this because:
- the function is only returning a path.  It should be on the caller to verify whether there is a file/resource at that path before doing anything with it
- the single caller for this function currently does just that:



Let me know what you think!

gbockus-sf: winner winner

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
gbockus-sf: cool. lemme know if you want to pair on funky tests.

gbockus-sf (approved):
During QE realized this is only used for the testing in lwc so we're good to go 👍 
Note merged with a failing integration test in circle b/c something changed in vscode and we're seeing the failure across the board (even in main that hasn't changed in weeks).  Fix is in the bundling PR.

PR #4469, opened 2022-09-30
Use ConfigAggProvider Everywhere

klewis-sfdc:
On the develop branch, after moving the creation of the configAggregator to the utils package, we were seeing some race condition issues where a global config aggregator would get cached for the project when the initial connection was created. This PR creates all instances of ConfigAggregator using the provider, resolving the race condition and ensuring that it always picks up local project config variable values.

### What does this PR do?
Reloads the configAggregator in the utils package inside the cliConfigChangeHandler function.  This ensures that when workspaceContextUtils fires the orgChange event, it has already reloaded the configAggregator and retrieved the latest configuration values from the config file.

This PR also modifies the way VSCE uses sf-core's testSetup() function.  This function mocks many core types globally when called outside of the describe block.  In order to prevent mocking the core types globally for all types, and to selectively mock core for specific describe blocks only, this PR changes many call sites in the salesforcedx-vscode-core package from calling testSetup() to using instantiateContext() with stubContext() and restoreContext() in the before() and after() blocks. 

@W-11848639@

### Functionality Before
When a local default username is configured, ConfigUtil.getDefaultUsernameOrAlias() returns the globally configured value

### Functionality After
When a local default username is configured, it is returned by ConfigUtil.getDefaultUsernameOrAlias()


packages/salesforcedx-vscode-core/test/vscode-integration/util/cliConfiguration.test.ts
```diff
@@ -125,4 +131,45 @@ describe('SFDX CLI Configuration utility', () => {
...
+        dummyLocalDefaultUsername
+      );
+      localSfdxConfigFile.writeSync(); // doesn't fire the file watcher logic to run
+      process.chdir('/'); // Change back to the default process.cwd
+
+      // Act
+      const localProjectDefaultUsernameOrAlias = await ConfigUtil.getDefaultUsernameOrAlias();
+
+      // Assert
+      // when the test initializes it creates a ConfigAggregator instance for the SFDX simple project, and put it in the config aggregator provider internal map. I have to get that provider and call reload in order to reset it before this test.
```
klewis-sfdc, line 169: Not sure if we can get this test to work.  I can set up the config file using the core lib types, but since a config aggregator instance is created for the project during the test initialization before the it block even runs, that global one is being returned for this project. In order to get the ConfigAggregator to be accurate for the project I would have to call reload, but I think that kind of defeats the point of this test... Open to any suggestions or ideas here.

packages/salesforcedx-vscode-core/test/vscode-integration/util/cliConfiguration.test.ts
```diff
@@ -125,4 +131,45 @@ describe('SFDX CLI Configuration utility', () => {
...
+      }
+
+      localSfdxConfigFile = await MyConfig.create({
+        isGlobal: false
+      });
+      localSfdxConfigFile.set(
+        OrgConfigProperties.TARGET_ORG,
+        dummyLocalDefaultUsername
+      );
+      localSfdxConfigFile.writeSync(); // doesn't fire the file watcher logic to run
```
gbockus-sf, line 162: interesting...I wonder why not

gbockus-sf: if we used sfdx here to set the default username, would it trigger the filewatcher that would retrigger the reload giving us the correct value below?

packages/salesforcedx-vscode-core/test/vscode-integration/util/cliConfiguration.test.ts
```diff
@@ -125,4 +131,45 @@ describe('SFDX CLI Configuration utility', () => {
...
+      process.chdir('/'); // Change back to the default process.cwd
```
gbockus-sf, line 163: note: I don't think this will work on windows?  maybe it will..I dunno

packages/salesforcedx-vscode-core/test/vscode-integration/util/cliConfiguration.test.ts
gbockus-sf: So is this identifying a current error condition we have with the flow of auth'ing an org for the first time? 
The steps seems like: 
1. VSCode opens and extensions load (no local config so global configAgg gets populated).
2. The localConfig gets set some how ( not sure if above flow is same as what we have when a user auths or not). 
3. We should ensure that all configAggs get reloaded to the local value.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -130,7 +130,7 @@ export class ConfigUtil {
 
     const info = await StateAggregator.getInstance();
     const username = defaultUsernameOrAlias
-      ? info.aliases.getUsername(String(defaultUsernameOrAlias))
+      ? info.aliases.resolveValue(String(defaultUsernameOrAlias))
```
klewis-sfdc, line 133: getUsername returns null when this method is passed the username.  resolveValue returns the passed in string (the username) instead of returning null

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -92,12 +88,11 @@ export class WorkspaceContextUtil {
...
   protected async handleCliConfigChange() {
-    const usernameOrAlias = await this.getAuthUtil().getDefaultUsernameOrAlias(
-      false
-    );
+    await ConfigAggregatorProvider.getInstance().reloadConfigAggregators();
+    const usernameOrAlias = await ConfigUtil.getDefaultUsernameOrAlias();
 
     if (usernameOrAlias) {
-      this._username = await this.getAuthUtil().getUsername(usernameOrAlias);
+      this._username = await ConfigUtil.getUsername();
```
klewis-sfdc, line 95: This is why the change to resolveValue is needed.  when there is no alias, username is passed in here, and username should come back

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -46,7 +46,6 @@ export class WorkspaceContext {
       // error reported by setupWorkspaceOrgType
       console.error(e)
     );
-    await ConfigAggregatorProvider.getInstance().reloadConfigAggregators();
```
klewis-sfdc, line 49: this is now being called in the utils package where the listener first reacts to the config file change, instead of here in the workspaceContext module where the event is handled

gbockus-sf (commented):
Looks good! I'll QE this shortly

gbockus-sf (approved):
QE for W-11848639 Use ConfigAggProvider Everywhere

* config changes via sfdx are detected and update the configAgg ✅
    * execute `sfdx force:config:set defaultusername=jsforce` from the terminal
    * breakpoint in workspaceContextUtils → handleCliConfigChange
    * Verify that the username returned from getDefaultUsernameOrAlias matches the value passed in the sfdx command.
* Deploy ✅
* Retrieve via org browser ✅

PR #4486, opened 2022-10-07
chore: update no response gha

randi274:
### What does this PR do?
Replaces our old No Response action with a No Response workflow.

@W-11871928@

### Functionality Before
No response bot was flaky, and not really working anymore.

### Functionality After
New action should perform as we previously expected ↺

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -1,0 +1,26 @@
+name: No Response
+
+## Closes issues marked with 'more information required' label after 7 days
+##    if original poster does not respond within that time
+
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
+        with:
+          token: ${{ secrets.SF_IDEE_BOT_GITHUB_TOKEN }}
+          daysUntilClose: 7
+          responseRequiredLabel: 'more information required'
+          closeComment: >
+            This issue has been automatically closed because there has been no response
+            to our request for more information from the original author. Currently, there
+            is not enough information provided for us to take action. Please reply and 
+            reopen this issue if you need additional assistance.
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,9 +1,0 @@
-# Configuration for probot-no-response - https://github.com/probot/no-response
-
-daysUntilClose: 7
-responseRequiredLabel: 'more information required'
-closeComment: >
-  This issue has been automatically closed because there has been no response
-  to our request for more information from the original author. Currently, there
-  is not enough information provided for us to take action. Please reply and 
-  reopen this issue if you need additional assistance.


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,26 @@
...
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
```
gbockus-sf, line 17: I assume this is already being used somewhere in SF land?

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          token: ${{ secrets.SF_IDEE_BOT_GITHUB_TOKEN }}
+          daysUntilClose: 7
+          responseRequiredLabel: 'more information required'
+          closeComment: >
+            This issue has been automatically closed because there has been no response
+            to our request for more information from the original author. Currently, there
+            is not enough information provided for us to take action. Please reply and 
+            reopen this issue if you need additional assistance.
```
gbockus-sf, line 26: Does it remove the label if it finds activity on the issue or no?

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,26 @@
+name: No Response
+
+## Closes issues marked with 'more information required' label after 7 days
+##    if original poster does not respond within that time
+
+on:
+  issue_comment:
+    types: [created]
```
gbockus-sf, line 8: sorry dumb q: do you know what this means?

randi274: Yeah, so on create of an issue comment (not edited/deleted). So if someone has a new comment

.github/workflows/no-response.yml
randi274: Yep, this one does remove the label! It does have to be the original poster though.

Also one interesting [quirk](https://github.com/lee-dohm/no-response#issue_comment-event) that I don't think the app did: "Reopen the issue if it was closed by someone other than the original author of the issue" if the original author comments again. I think that's good though, because if we close it, but then the OG gets back to us, we probably want it to be reopened, and that'll save us a click.

.github/workflows/no-response.yml
randi274: Nope :) CLI is still using the Probot App that's having issues. We are pinned to a specific version here though. I could go the extra step and pin to the commit hash.

.github/workflows/no-response.yml
gbockus-sf: sweet. def sounds better

.github/workflows/no-response.yml
```diff
@@ -14,9 +14,9 @@ jobs:
   noResponse:
     runs-on: ubuntu-latest
     steps:
-      - uses: lee-dohm/no-response@v0.5.0
+      - uses: lee-dohm/no-response@66174372c9c5576aa8a4287e33653f2fdd12d517
         with:
-          token: ${{ secrets.SF_IDEE_BOT_GITHUB_TOKEN }}
+          token: ${{ github.token }}
```
gbockus-sf, line 19: good thinking

randi274:
@gbockus-sf - take a look next week when you get a chance; I've updated this to use the stale issues action instead.

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,26 @@
...
+##    if original poster does not respond within that time
+
+
+permissions:
+  issues: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '30 1 * * *'
```
gbockus-sf, line 13: looks like we're back to only once a day.   Was that intentional?

gbockus-sf (approved):
🚢

gbockus-sf (approved):
I think we're good to merge 👍

gbockus-sf:
Verified that only build failure for this PR is the single int test that started failing across the board earlier this week.  Fixed in the bundling PR.  Merge at will. @randi274

PR #4510, opened 2022-10-24
chore: update to allow soql ui test job to complete with new version of vscode

gbockus-sf:
### What does this PR do?
- Update the engine value for the package.json files to match the new version specified for `@types/vscode`
- Add creating the vsix files to the test:ui job
- Update the test to account for a change in the vscode dom for finding the button to toggle the SOQL builder UI.

@W-11950108@

### Functionality Before
Test Fails: https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-vscode/10690/workflows/ea83f680-72e8-4d51-a4e0-34fcba589033/jobs/39917

### Functionality After
Test Passes in Circle.


.circleci/config.yml
```diff
@@ -342,6 +342,7 @@ jobs:
       - NODE_VERSION: << pipeline.parameters.node-version >>
     steps:
       - build-win
+      - package-extensions
```
gbockus-sf, line 345: I had to update how we run the 'test:ui' script as it was building a vsix file on the fly previously and now we need to following the bundling path.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -83,7 +83,10 @@
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "test": "npm run test:vscode-integration",
-    "test:ui": "extest setup-and-run -s ~/.vscode-test-resources -o test/ui-test/resources/settings.json -e test/ui-test/resources/extensions out/test/ui-test/*.test.js",
+    "test:ui:prepare": "extest get-vscode -s ~/.vscode-test-resources && extest get-chromedriver -s ~/.vscode-test-resources ",
+    "test:ui:install": "extest install-vsix -f ./salesforcedx-vscode-soql-56.2.0.vsix -s ~/.vscode-test-resources -e test/ui-test/resources/extensions ",
+    "test:ui:run": "extest run-tests -s ~/.vscode-test-resources -o test/ui-test/resources/settings.json -e test/ui-test/resources/extensions out/test/ui-test/*.test.js",
+    "test:ui": "npm run test:ui:prepare && npm run test:ui:install && npm run test:ui:run",
```
gbockus-sf, line 89: SO previously the extest command would call vsce and create a vsix file.  This is no longer possible without our bundle/package wrapper so I just updated it here to do the different pieces individually instead of using the setup-and-run task.  This works locally, but will need to verify in Circle after merge as the test:ui job only runs as part of the scheduled build.

packages/salesforcedx-vscode-soql/test/ui-test/soql-ui-smoke.test.ts
```diff
@@ -82,6 +82,7 @@ describe('In project folder, SOQL files should', function() {
     expect(toggle).is.not.undefined;
     await toggle?.click();
 
+    console.log('**** GORDON toggle clicked.');
```
gbockus-sf, line 85: I swear I removed this....will remove in follow on commit.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -71,7 +71,7 @@
     "papaparse": "^5.3.0",
     "sinon": "13.0.1",
     "typescript": "^4.7.4",
-    "vscode-extension-tester": "^4.0.2",
+    "vscode-extension-tester": "^5.0.0",
```
gbockus-sf, line 74: I thought this might help...it didn't, but doesn't hurt so I let it to pick up the latest.

.circleci/config.yml
gbockus-sf, line 345: See https://github.com/forcedotcom/salesforcedx-vscode/pull/4510/files#r1003744311 about why this is necessary.

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -29,7 +29,7 @@
     "@types/proxyquire": "1.3.28",
     "@types/shelljs": "0.7.9",
     "@types/sinon": "^2.3.7",
-    "@types/vscode": "1.72.0",
+    "@types/vscode": "^1.61.2",
```
gbockus-sf, line 32: So 1.61 is Sept 2021. 1.72 is 09/2022  I intended to have the year old version for types in the bundling PR but must have lost track of making that update.  The types only update shouldn't be a problem b/c if it builds it builds.  Note we looked in app insights and verified that over 99% of our users are on a recent release. 

TL;DR.  1.61.2 is support back one year and covers the use case necessary for the bundling.

packages/salesforcedx-vscode-soql/test/ui-test/soql-ui-smoke.test.ts
```diff
@@ -72,7 +72,7 @@ describe('In project folder, SOQL files should', function() {
 
     for (const action of actions) {
       if (
-        (await action.getAttribute('title')) ===
+        (await action.getAttribute('aria-label')) ===
```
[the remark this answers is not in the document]
randi274: Sneaky! How did this get changed?

randi274 (approved):

packages/salesforcedx-vscode-soql/test/ui-test/soql-ui-smoke.test.ts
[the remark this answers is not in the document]
gbockus-sf: There was an update to the vscode toolbar DOM (this is finding the button in the toolbar when looking at a soql file for toggling the file vs solqbuilder view).  This may have been broken for a while as these tests fail all the time and I just bothered to actually follow up this Monday as part of trust and the failed build.

PR #4514, opened 2022-10-25
refactor: update usage of .sfdx directory

CristiCanizales:
### What does this PR do?
- Updates scratchOrgDecorator’s usage of ‘.sfdx’ directory

[@W-11633519@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000014hjOKYAY/view)


packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +71,39 @@ describe('test project paths', () => {
       expect(relativeStateFolder).toEqual(FAKE_STATE_FOLDER);
     });
   });
+
+  describe('test SFDXGlobalConfigPath', () => {
+    it('should return the global config path', () => {
+      const result = projectPaths.sfdxGlobalConfig();
+      expect(result).toEqual(path.join(os.homedir(), '.sfdx', 'sfdx-config.json')
```
klewis-sfdc, line 78: Can os.homedir() be stubbed here?, so it would be `path.join(FAKE_HOME_DIR, FAKE_STATE_FOLDER, FAKE_CONFIG_FILE)`

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +71,39 @@ describe('test project paths', () => {
...
+      );
+    });
+  });
+
+  describe('test SFDXConfigPath', () => {
+    const FAKE_CONFIG = path.join(FAKE_WORKSPACE, 'sfdx-config.json');
```
klewis-sfdc, line 84: I think the state folder is missing here, should be: `(FAKE_WORKSPACE, FAKE_STATE_FOLDER, FAKE_CONFIG_FILE)`

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +71,39 @@ describe('test project paths', () => {
...
+
+      const sfdxConfig = projectPaths.sfdxConfig();
+      expect(sfdxConfig).toEqual(FAKE_CONFIG);
+    });
+
+    it('should return a path to the global config file', () => {
+      hasRootWorkspaceStub.mockReturnValue(false);
+
+      const sfdxConfig = projectPaths.sfdxConfig();
+      expect(sfdxConfig).toEqual(path.join(os.homedir(), '.sfdx', 'sfdx-config.json'));
```
klewis-sfdc, line 106: Same comment about stubbing os.homedir here and making this something like` path.join(FAKE_HOME_DIR, FAKE_STATE_FOLDER, FAKE_CONFIG_FILE)`

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,44 @@ describe('test project paths', () => {
...
+
+  describe('test SFDXConfigPath', () => {
+
+    it('should be defined', () => {
+      expect(projectPaths.sfdxConfig).toBeDefined();
+    });
+
+    it('should return a path to the project config file', () => {
+      const FAKE_CONFIG = path.join(FAKE_WORKSPACE, FAKE_STATE_FOLDER, FAKE_CONFIG_FILE);
+      const FAKE_PATH = path.join(FAKE_WORKSPACE, FAKE_STATE_FOLDER);
```
klewis-sfdc, line 96: What do you think about moving these consts to the top of the file so they can be reused in the expect statements?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,44 @@ describe('test project paths', () => {
       expect(relativeStateFolder).toEqual(FAKE_STATE_FOLDER);
     });
   });
+
+  describe('test SFDXGlobalConfigPath', () => {
+    it('should return the global config path', () => {
+      const mockOS = jest.requireActual('os');
+      mockOS.homedir = jest.fn().mockReturnValue(FAKE_HOME_DIR);
+      jest.mock('os', () => mockOS);
```
klewis-sfdc, line 81: Nice job figuring this out.  I wonder, can the` import `statement at the top be removed with the use of `requireActual` here?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
CristiCanizales: I think the first one is something specific to this it so I'd leave it there, but the second one I'm going to move it to the top because there's another it using the same path:)

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
CristiCanizales: Yes! Thank you :)

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -137,6 +138,17 @@ function sfdxProjectConfig(): string {
...
+  const pathToSFDXGlobalConfig = path.join(os.homedir(), relativeStateFolder(), SFDX_CONFIG_FILE);
+  return pathToSFDXGlobalConfig;
+}
+
+function sfdxConfig(): string {
+  const pathToSFDXConfig = workspaceUtils.hasRootWorkspace()
+  ? sfdxProjectConfig() : sfdxGlobalConfig();
+  return pathToSFDXConfig;
+}
+
```
klewis-sfdc, line 151: When I save this file in my IDE, these two blocks are formatted slightly differently.  What do you think about the following formatting:

```suggestion
function sfdxGlobalConfig(): string {
  const pathToSFDXGlobalConfig = path.join(
    os.homedir(),
    relativeStateFolder(),
    SFDX_CONFIG_FILE
  );
  return pathToSFDXGlobalConfig;
}

function sfdxConfig(): string {
  const pathToSFDXConfig = workspaceUtils.hasRootWorkspace()
    ? sfdxProjectConfig()
    : sfdxGlobalConfig();
  return pathToSFDXConfig;
}

```

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,43 @@ describe('test project paths', () => {
...
+    });
+  });
+
+  describe('test SFDXConfigPath', () => {
+
+    it('should be defined', () => {
+      expect(projectPaths.sfdxConfig).toBeDefined();
+    });
+
+    it('should return a path to the project config file', () => {
```
klewis-sfdc, line 94:
```suggestion
    it('should return a path to the project config file when in a project workspace', () => {
```

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,43 @@ describe('test project paths', () => {
...
+      let stateFolderStub: jest.SpyInstance;
+      stateFolderStub = jest
+        .spyOn(projectPaths, 'stateFolder')
+        .mockReturnValue(FAKE_PATH);
+
+      const sfdxConfig = projectPaths.sfdxConfig();
+      expect(sfdxConfig).toEqual(FAKE_CONFIG);
+    });
+
+    it('should return a path to the global config file', () => {
```
klewis-sfdc, line 106:
```suggestion
    it('should return a path to the global config file when not in a project workspace', () => {
```

klewis-sfdc (commented):
Looking great, Cristi!  A few more suggestions and then I think this is ready to go.  Thank you!

klewis-sfdc (approved):
Looks great, Cristi! Thanks for these changes!

QE looks good:
✅  Confirmed via breakpoint that the scratchOrgDecorator CONFIG_FILE var value is the same as before and is returning valid path to config file.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,43 @@ describe('test project paths', () => {
       expect(relativeStateFolder).toEqual(FAKE_STATE_FOLDER);
     });
   });
+
+  describe('test SFDXGlobalConfigPath', () => {
+    it('should return the global config path', () => {
+      const mockOS = jest.requireActual('os');
+      mockOS.homedir = jest.fn().mockReturnValue(FAKE_HOME_DIR);
+      jest.mock('os', () => mockOS);
```
gbockus-sf, line 81: jest.mock calls are hoisted to the top of the file by jest so that the execution will happen prior to anything else. See [here](jest.mock). Top avoid running into weird execution state issue I think we should always put the jest.mock calls at the top of the file directly after the imports (so on line 12 for this particular case).  Would you mind moving this mock call to the top of the file?  Does it make sense why I'm requesting it?

CristiCanizales: yess. Thank you!!

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -137,6 +138,22 @@ function sfdxProjectConfig(): string {
...
+function sfdxGlobalConfig(): string {
+  const pathToSFDXGlobalConfig = path.join(
+    os.homedir(),
+    relativeStateFolder(),
+    SFDX_CONFIG_FILE
+  );
+  return pathToSFDXGlobalConfig;
+}
+
+function sfdxConfig(): string {
```
gbockus-sf, line 150: So quick question When would we ever want to get a config file and not know if it's global or project level? I'll review the use case in this PR but that seems like a dangerous flow.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +73,43 @@ describe('test project paths', () => {
       expect(relativeStateFolder).toEqual(FAKE_STATE_FOLDER);
     });
   });
+
+  describe('test SFDXGlobalConfigPath', () => {
+    it('should return the global config path', () => {
+      const mockOS = jest.requireActual('os');
+      mockOS.homedir = jest.fn().mockReturnValue(FAKE_HOME_DIR);
```
gbockus-sf: in a huddle we figured out that this was altering the os.homedir function for all tests.  An FYI if we have to use this method in the future

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
klewis-sfdc: The way this was originally written, I thought maybe the path to the global config file was desired when a VS Code folder was open that wasn't an sfdx project.  But, it doesn't make a lot of sense to look for the global sfdx config file if you are not in an sfdx project at all.  When I fire up the extensions when NOT in an sfdx project, this code isn't hit at all since this code block is not run:

So, maybe the project level fn (`sfdxProjectConfig()`) is the only one that is actually needed

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -5,20 +5,12 @@
...
 import { nls } from '../messages';
-import { workspaceUtils } from '../util';
 
-const CONFIG_FILE = workspaceUtils.hasRootWorkspace()
-  ? path.join(
-      workspaceUtils.getRootWorkspacePath(),
-      '.sfdx',
-      'sfdx-config.json'
-    )
-  : path.join(os.homedir(), '.sfdx', 'sfdx-config.json');
```
gbockus-sf, line 21: hummmmmmmmmmm. I gotta look at the source more closely here.  I don't really understand how you could want the global config file in this flow.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
gbockus-sf: yeah that's my suspicion.  Looks like you already did the deeper digging work I was gonna do.  I'd say we get rid of that and just only use the workspace config.  

So the only commands that can be run when not in a sf project are listed in the activation events ```    "workspaceContains:sfdx-project.json",
    "onCommand:sfdx.force.project.create",
    "onCommand:sfdx.force.project.with.manifest.create",
    "onCommand:sfdx.debug.isv.bootstrap"```

Project create and create with manifest do not depend on have a default username...check on isv.bootstrap now

gbockus-sf: Verified...there is a check already in that sfdx.debug.isv.bootstrap command that we are in a sfdx project so we can get rid of the global config file watch 👍

gbockus-sf: hum....so I can never leave well enough alone.  I looked at the commit where the global was added and it was specifically around project create.  So if we are gonna switch that to only respect local we should def verify the empty state -> create project command still works as expected.  I would think so, but best to be sure.

klewis-sfdc: What would you think about merging this, which maintains previous behavior, and follow up WIs to 
A: investigate whether the global path can be removed from this flow, and 
B: to refactor what this flow actually does (scratch org decorator reads the config file and can use ConfigUtil instead)

gbockus-sf: I'm fine with that. Good call on configUtils.  I would prefer that we get rid of the `sfdxConfig` function and just use workspaceUtils in the calling code ala 
```
// In an ideal world we'd log if we using local or global config file.
const configFile = workspaceUtils.hasRootWorkspace() ? sfdxProjectConfig() : sfdxGlobalConfig();
```
Getting a configFile and not knowing which one just seems like it'd be a bad thing.

klewis-sfdc: Yep.  If scratch org decorator is updated to use `ConfigUtil.getDefaultUsernameOrAlias()` inside `displayDefaultUserName`, then both `paths.sfdxConfig` and `paths.sfdxGlobalConfig` can be removed.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -7,6 +7,7 @@
 
 import { Global } from '@salesforce/core';
 import * as fs from 'fs';
+import * as os from 'os';
```
klewis-sfdc, line 10: Can this be removed now that the global function was removed?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -11,12 +12,19 @@ jest.mock('@salesforce/core', () => {
...
+  const FAKE_HOME_DIR = '/fake/home/dir/';
   const FAKE_WORKSPACE = '/here/is/a/fake/path/to/';
   const FAKE_STATE_FOLDER = '.sfdx';
+  const FAKE_CONFIG_FILE = 'sfdx-config.json';
+  const FAKE_PATH = path.join(FAKE_WORKSPACE, FAKE_STATE_FOLDER);
+  let homeDirStub: jest.SpyInstance;
 
+  beforeEach(() => {
+    homeDirStub = jest.spyOn(os, 'homedir').mockReturnValue(FAKE_HOME_DIR);
+  });
```
klewis-sfdc, line 25: I think this beforeEach can be removed now also since the global function was removed.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -11,12 +11,14 @@ jest.mock('@salesforce/core', () => {
 });
 
 describe('test project paths', () => {
+  const hasRootWorkspaceStub = jest.spyOn(workspaceUtils, 'hasRootWorkspace');
   const FAKE_WORKSPACE = '/here/is/a/fake/path/to/';
   const FAKE_STATE_FOLDER = '.sfdx';
+  const FAKE_CONFIG_FILE = 'sfdx-config.json';
+  const FAKE_PATH = path.join(FAKE_WORKSPACE, FAKE_STATE_FOLDER);
```
klewis-sfdc, line 18: nit: Could this be renamed FAKE_PATH_TO_STATE_FOLDER?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -31,7 +33,7 @@ describe('test project paths', () => {
     it('should return a path to the state folder if the project has a root workspace', () => {
```
klewis-sfdc, line 33: I'm not sure that these "should be defined" tests are necessary (the one above this test).  I think there are compilation problems if the function is removed since it is referenced in the the rest of the functions below.  @gbockus-sf do these provide value since we get the same signal at compile time if the function is no longer defined?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -31,7 +33,7 @@ describe('test project paths', () => {
...
       hasRootWorkspaceStub.mockReturnValue(true);
       expect(projectPaths.stateFolder()).toEqual(
-        path.join(FAKE_WORKSPACE, FAKE_STATE_FOLDER)
+        path.join(FAKE_PATH)
```
klewis-sfdc, line 36: The path.join can be removed here.

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -5,50 +5,38 @@
...
   }
-  displayDefaultUserName(CONFIG_FILE);
+  await displayDefaultUserName();
 }
 
 export function monitorOrgConfigChanges() {
   const watcher = workspace.createFileSystemWatcher(CONFIG_FILE);
   watcher.onDidChange(uri => {
-    displayDefaultUserName(uri.fsPath);
+    displayDefaultUserName().catch(err => console.error(err));
```
klewis-sfdc, line 30:
```suggestion
    displayDefaultUserName();
```

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -5,50 +5,38 @@
...
   });
   watcher.onDidCreate(uri => {
-    displayDefaultUserName(uri.fsPath);
+    displayDefaultUserName().catch(err => console.error(err));
```
klewis-sfdc, line 33:
```suggestion
    displayDefaultUserName();
```

klewis-sfdc: I think the catch blocks can be removed here.  We can do some things in the test to get around the error that motivated this change.

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -5,50 +5,38 @@
...
-function displayDefaultUserName(configPath: string) {
-  fs.readFile(configPath, (err, data) => {
-    if (!err) {
-      const config = JSON.parse(data.toString());
-      if (config['defaultusername']) {
-        statusBarItem.text = `$(browser)`;
-      }
-    }
-  });
+async function displayDefaultUserName() {
```
klewis-sfdc, line 37: What do you think about renaming this function?  displayBrowserIcon?

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
```diff
@@ -5,50 +5,38 @@
...
+  const defaultUsernameOrAlias = await ConfigUtil.getDefaultUsernameOrAlias();
+  if (defaultUsernameOrAlias) {
```
klewis-sfdc, line 39: It will make the tests easier to combine the logic from 'showOrg' and the logic from this function displayDefaultUserName/displayBrowserIcon so that the statusBarItem only gets created when needed:

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  let mockGetDefaultUsername: jest.SpyInstance;
+  let mockCreateStatusBar: jest.SpyInstance;
+
+  beforeEach(() => {
+    mockCreateStatusBar = (vscode.window.createStatusBarItem as any).mockReturnValue(
+      mockStatusBarItem
+    );
+    mockFileSystemWatcher = (vscode.workspace
+      .createFileSystemWatcher as any).mockReturnValue(mockWatcher);
+    mockGetDefaultUsername = jest
```
klewis-sfdc, line 39: nit: Could these be renamed to be consistent with the pattern we've followed in other recent tests:
_"methodName-Mock"_:

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  describe('monitor org changes', () => {
+    it('should show the browser icon when the config file changes', () => {
+      onDidChangeEventEmitter.fire(fakePathURI);
+      // Act
+      monitorOrgConfigChanges();
+      expect(mockFileSystemWatcher).toHaveBeenCalled();
+      expect(mockWatcher.onDidChange).toHaveBeenCalledWith(expect.any(Function));
+      // expect(mockGetDefaultUsername).toHaveBeenCalled();
+    });
```
klewis-sfdc, line 71: I think the event emitter can be removed from this test file.  By getting a handle on the callback function passed to `watcher.onDidChange()`, we can then just call that callback function, faking like a file change happened.  Let me know if you have questions or would like to huddle to discuss.

klewis-sfdc (changes requested):
Looking good, Cristi!  I left a handful of comments.  Happy to huddle today if you'd like to discuss!

packages/salesforcedx-vscode-core/src/decorators/scratch-org-decorator.ts
CristiCanizales: That it makes a lot more sense :)

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    show: jest.fn(),
+    text: ''
+  };
+  let onDidCreateEventEmitter: vscode.EventEmitter<vscode.Uri>;
+  let onDidChangeEventEmitter: vscode.EventEmitter<vscode.Uri>;
+  onDidCreateEventEmitter = new vscode.EventEmitter<vscode.Uri>();
+  onDidChangeEventEmitter = new vscode.EventEmitter<vscode.Uri>();
+  const mockWatcher = {
+    // onDidChange: onDidChangeEventEmitter.event,
+    // onDidCreate: onDidCreateEventEmitter.event
```
gbockus-sf, line 25: if you left these comments as info we should add another comment to say why...other wise remove the dead code.

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    onDidChange: jest.fn(),
+    onDidCreate: jest.fn()
+  };
```
gbockus-sf, line 28: You should do the mockWatcher assignment in the beforeEach so that the spys are reset between tests. Also mockStatusBarItem.show.  Basically you should always set spy values in the before* function.

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    mockGetDefaultUsername = jest
+      .spyOn(ConfigUtil, 'getDefaultUsernameOrAlias');
+  });
+
+  describe('show Org', () => {
+    it('should not show the browser icon in the status bar when a default username is not set', async () => {
+      await showOrg();
+      expect(mockCreateStatusBar).toHaveBeenCalled();
+      expect(mockStatusBarItem.text).toEqual('');
+      expect(mockStatusBarItem.tooltip).toEqual(nls.localize('status_bar_open_org_tooltip'));
```
gbockus-sf, line 48: nice...I was hard coding in my PR.  Def better to use localize

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
gbockus-sf: agreed.  def better from a unit test perspective 👍

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      expect(mockFileSystemWatcher).toHaveBeenCalled();
+      expect(mockWatcher.onDidChange).toHaveBeenCalledWith(expect.any(Function));
+      // expect(mockGetDefaultUsername).toHaveBeenCalled();
+    });
+
+    it('should show the browser icon when the config file is created', () => {
+      monitorOrgConfigChanges();
+      expect(mockFileSystemWatcher).toHaveBeenCalled();
+      expect(mockWatcher.onDidCreate).toHaveBeenCalledWith(expect.any(Function));
+      // expect(mockGetDefaultUsername).toHaveBeenCalled();
```
gbockus-sf, line 77: nit: no dead code.

scripts/setup-jest.ts
```diff
@@ -73,6 +75,10 @@ const getMockVSCode = () => {
         onDidDelete: jest.fn()
       }),
       workspaceFolders: []
+    },
+    StatusBarAlignment: {
```
gbockus-sf, line 79: can we put this in alpha order by property name? So move it to after ProgressLocation

gbockus-sf (commented):
great stuff.  A few minor comments.

klewis-sfdc (commented):
Code changes look great, @CristiCanizales!  Thanks for making the change and upgrading all the tests!  Will do QA on this branch now and report back.

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceContext.test.ts
```diff
@@ -191,6 +189,7 @@ describe('WorkspaceContext', () => {
     ]);
 
     expect(someLogic.callCount).to.equal(3);
+    expect(showOrgStub.called).to.equal(true);
```
klewis-sfdc, line 192: 👍🏽

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      expect(mockStatusBarItem.show).not.toHaveBeenCalled();
+      expect(mockStatusBarItem.text).toEqual('');
+    });
+    it('should dispose and set to undefined the status bar when a default username is not set and the status bar exists', async () => {
+      mockStatusBarItem.text = 'text';
+      getDefaultUsernameOrAliasMock.mockResolvedValue(undefined);
+      await showOrg();
+      expect(getDefaultUsernameOrAliasMock).toHaveBeenCalled();
+      expect(createStatusBarItemMock).not.toHaveBeenCalled();
+      expect(mockStatusBarItem.dispose).toHaveBeenCalled();
```
klewis-sfdc, line 62: The status bar item hasn't been created yet in this test, so dispose won't be called here.  In order to test that dispose was called on the internal instance of statusBarItem, we'll have to call showOrg legitimately first to create the statusBarItem and have a handle on the mock:

packages/salesforcedx-vscode-core/test/jest/decorators/scratch-org-decorator.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      expect(mockStatusBarItem).toEqual(undefined);
```
klewis-sfdc, line 63: I don't think this can be tested since we don't have a reference to the internal instance of statusBarItem after setting it to undefined.  

What if instead of setting it back to undefined, we use show and hide here instead:



This might be using show and hide as intended, and is easier to test since we can test that `hide()` was called.

klewis-sfdc: This test could then look like:

CristiCanizales: We considered using `.hide()` but thought `.dispose()` was better and is already calling `.hide()`. I think then maybe just testing that dispose was called is ok ?

CristiCanizales: 👀

klewis-sfdc: OK, I think this is fine as is.

klewis-sfdc:
Looks good from a QA perspective!  Just need to get that last test issue figured out, then good to go!  Thank you @CristiCanizales!

✅ Log Out from All Authorized Orgs makes browser icon disappear
✅ Auth an org makes the browser icon appear
✅ Browser icon appears at initialization when default org is set
✅ Browser icon does not appear at initialization when default org is not set

klewis-sfdc (approved):

Issue #4465, opened 2022-09-29
Error when creating a new project when VS Code is empty and doesn't have any folders


jeffb-sfdc:
In the develop branch, if you:
1. start with an empty window in VS Code (no files or folders opened)
2. run SFDX: Create Project
…this is now failing (develop branch).

This is b/c getRootWorkspacePath(), getRootWorkspace(), and hasRootWorkspace() is getting called.  When called and no folders exists, ws is empty and getRootWorkspacePath() returns an empty string, which has ramifications further downstream.

Some other thoughts:
1. if getRootWorkspacePath(), if getRootWorkspace() doesn’t return anything, maybe it should throw instead of returning an empty string.

2. in getRootWorkspacePath(), getRootWorkspace() is being called twice:
```
export function getRootWorkspacePath(): string {
  return getRootWorkspace().uri ? getRootWorkspace().uri.fsPath : '';
}
```
…calling getRootWorkspace() just once would be better:

```
export function getRootWorkspacePath(): string {
  const rootWorkspace = getRootWorkspace();
  return rootWorkspace.uri ? rootWorkspace.uri.fsPath : '';
}
```

3. There are two copies of rootWorkspace.ts:
packages/salesforcedx-utils-vscode/src/workspaces/rootWorkspace.ts
packages/salesforcedx-vscode-core/src/util/rootWorkspace.ts

gbockus-sf:
Fixed and release!

PR #4517, opened 2022-10-26
fix: update to verify path exists before chdir in configAggProvider

gbockus-sf:
### What does this PR do?
Update the config agg provider to not attempt to chdir when the current working directory is undefined or empty string.

#4465

### Functionality Before
Screen Shot 2022-10-26 at 11 36 10 AM

### Functionality After
Successful creation of a new project from the empty project state.


klewis-sfdc (approved):
Looks good.  Confirmed via QE that 

✅ SFDX: Create Project works when not already in a project directory
✅ Once the project is created, ConfigAggregator for the root project workspace has local config values present

PR #4521, opened 2022-10-28
update usage of '.sfdx' dir in isvdebugging 

RitamAgrawal:
### What does this PR do?
This PR adds `relativeToolsFolder` function in projectPaths which updates the usage of string '.sfdx' and 'tools'

 @W-11633488@

### Functionality Before
`.sfdx/tools` dir was referenced as string
### Functionality After
`.sfdx/tools` dir is referenced using dynamic functions from `salesforcedx-utils-vscode`


packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -55,17 +56,15 @@ export interface InstalledPackageInfo {
...
     'isvdebuggermdapitmp'
   );
   public readonly relativeApexPackageXmlPath = path.join(
     this.relativeMetdataTempPath,
     'package.xml'
   );
   public readonly relativeInstalledPackagesPath = path.join(
-    '.sfdx',
-    'tools',
+    projectPaths.relativeToolsFolder(),
```
gbockus-sf, line 67: I realize adding a full blown jest unit test suite for this file is probably more lift then is worthwhile for the work being addressed, but you would add a test(s) to verify the `relativeInstalledPackagesPath` and `relativeMetdataTempPath` readonly values are set appropriately.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -70,4 +71,15 @@ describe('test project paths', () => {
...
   });
+
+  describe('test relativeToolsFolder', () => {
+    it('should be defined', () => {
+      expect(projectPaths.relativeToolsFolder).toBeDefined();
+    });
+
+    it('should return a path to the relative tools folder', () => {
+      const relativeToolsFolder = projectPaths.relativeToolsFolder();
+      expect(relativeToolsFolder).toEqual(FAKE_TOOLS_FOLDER);
```
gbockus-sf, line 82: So for this test you should be verifying that the value returned is the combo of the relative state folder and the tools constant.  I'd probably be ok with something like 
```      
expect(relativeToolsFolder).toEqual(
        path.join(FAKE_STATE_FOLDER, TOOLS) // TOOLS exported from paths.ts
      );
```

But even better would be to mock the projectPaths.relativeStateFolder() response for this test and have something like 
```
     relativeStateFolderMock.resolves(FAKE_STATE_FOLDER);
     const relativeToolsFolder = projectPaths.relativeToolsFolder();
      expect(relativeToolsFolder).toEqual(
        path.join(FAKE_STATE_FOLDER, TOOLS) // TOOLS exported from paths.ts
      );
```

gbockus-sf (commented):
couple of testing suggestions 👍

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
gbockus-sf: Looks like your test is failing with 
```
    Expected: ".sfdx/tools"     Received: ".sfdx\\tools"```
```
another good reason to update to the above suggested method for the assertion.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -13,6 +13,7 @@ jest.mock('@salesforce/core', () => {
 describe('test project paths', () => {
   const FAKE_WORKSPACE = '/here/is/a/fake/path/to/';
   const FAKE_STATE_FOLDER = '.sfdx';
+  const FAKE_TOOLS_FOLDER = 'tools';
```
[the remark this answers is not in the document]
gbockus-sf: Importing the tools constant from the single location would def be better

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -26,7 +26,6 @@ import {
 } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
 import { join } from 'path';
 import * as vscode from 'vscode';
-import { BaseDeployExecutor } from '.';
```
gbockus-sf, line 29: So doing an import like this results in scanning all things exported from the index.ts file.  Which can lead to complications in unit testing.  Updated to pull directly from the file where exported instead.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -23,7 +18,7 @@ import { SfdxPackageDirectories } from '../sfdxProject';
 import { telemetryService } from '../telemetry';
 import { workspaceUtils } from '../util';
 import { DeployExecutor } from './baseDeployRetrieve';
-import { FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
+import { ConflictDetectionMessages, FilePathGatherer, SfdxCommandlet, SfdxWorkspaceChecker } from './util';
```
gbockus-sf, line 21: Ran an organize imports.  quite a few unused here.

packages/salesforcedx-vscode-core/src/commands/util/conflictDetectionMessages.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
```
gbockus-sf, line 7: So for these two exports they were originally in `postconditionCheckers.ts` with a ton of other stuff.  This lead to always importing this metadata caching checker that had a ton of dependancies and eventually had issues.  Simplified by moving these out to their own files and re-exporting from there for index.ts.

packages/salesforcedx-vscode-core/src/commands/util/sfdxCommandlet.ts
```diff
@@ -7,27 +7,19 @@
...
 } from '@salesforce/salesforcedx-utils-vscode';
 import * as vscode from 'vscode';
-import { EmptyPostChecker } from '.';
 import { channelService } from '../../channels';
 import { notificationService, ProgressNotification } from '../../notifications';
 import { sfdxCoreSettings } from '../../settings';
 import { taskViewService } from '../../statuses';
 import { telemetryService } from '../../telemetry';
 import { workspaceUtils } from '../../util';
+import { EmptyPostChecker } from './emptyPostChecker';
```
gbockus-sf, line 22: organize imports.

packages/salesforcedx-vscode-core/test/jest/commands/isvdebugging/bootstrapCmd.test.ts
```diff
@@ -11,16 +11,15 @@ describe('isvdebugging unit test', () => {
 
   beforeEach(() => {
     relativeToolsFolderStub = jest.spyOn(projectPaths, 'relativeToolsFolder').mockReturnValue(TOOLS_FOLDER);
-    isvDebugBootstrapExecutorInst = new IsvDebugBootstrapExecutor();
-
   });
 
   it('should be defined', () => {
+    isvDebugBootstrapExecutorInst = new IsvDebugBootstrapExecutor();
```
gbockus-sf, line 17: So this wasn't required, but when testing an _instance_ of a class it's always good to create the instance for each test.

scripts/setup-jest.ts
```diff
@@ -64,7 +67,11 @@ const getMockVSCode = () => {
         };
       },
       onDidChangeConfiguration: jest.fn(),
-      createFileSystemWatcher: jest.fn(),
+      createFileSystemWatcher: jest.fn().mockReturnValue({
+        onDidChange: jest.fn(),
+        onDidCreate: jest.fn(),
+        onDidDelete: jest.fn()
+      }),
```
gbockus-sf, line 74: new vscode mock setup to get things working for core.  Note that b/c the src/context/index.ts creates an instance of WorkspaceContext on initialization
 `export const workspaceContext = WorkspaceContext.getInstance();`
We need to mock the file watcher results prior to loading any test dependancies.  
This is an anti-pattern we should avoid for singletons.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -26,7 +26,6 @@ import {
 } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
```
gbockus-sf, line 26: Some touches to get jest unit tests working in core.

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -53,19 +54,21 @@ export interface InstalledPackageInfo {
   versionNumber: string;
 }
 
+export const ISVDEBUGGER = 'isvdebuggermdapitmp';
+export const INSTALLED_PACKAGES = 'installed-packages';
+export const PACKAGE_XML = 'package.xml';
```
gbockus-sf, line 59: we should update the below paths to use these constants.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -156,5 +161,7 @@ export const projectPaths = {
   sfdxProjectConfig,
   toolsFolder,
   lwcTestResultsFolder,
-  relativeStateFolder
+  relativeStateFolder,
+  relativeToolsFolder,
+  TOOLS
```
gbockus-sf, line 166: nit: I'd just export the TOOLS constant on line 17 vs including constants and functions here.

gbockus-sf (approved):
Couple of simple polish suggestions and you're good to go 👍 
- Verified unit test execution and pass in circle ✅ 
- Verified unit test execution local ✅ 
- Not sure how to execute the ISV debugger but verified as best I could ✅

PR #4532, opened 2022-11-02
fix: update bundling for core extension to exclude functions-core

### LINKED ISSUE #4531: Unable to run SFDX: Create function command
### Summary

After the update to version v56.3.1 the `SFDX: Create function` command fails for all function types. Reported in slack. 

### Steps To Reproduce:

1.  Command Pallet -> SFDX: Create Function
1.  Enter name and select and type 

### Expected result

Should create the function. 

### Actual result

An error similar to 
```
14:47:33.806 Starting SFDX: Create Function

ENOENT: no such file or directory, lstat '/Users/gbockus/.vscode-insiders/extensions/tmpl/javascript'
```

### Additional information

Also occurring on code builder, but was not present when code builder was using extensions version v55.8.0. 

**Salesforce Extension Version in VS Code**:
SF Extensions v56.3.1

gbockus-sf:
### What does this PR do?
Update the core extension bundling to exclude the functions-core module from bundling. 

### Testing
To test this change do the following: 
1. npm install 
2. npm run compile 
3. cd package/salesforcedx-vscode-core
4. npm run bundle:extension
5. npm run vscode:package

This will generate the `salesforcedx-vscode-core-56.4.0.vsix` extension.

To install: 
Go to whatever version of vscode you'd like to use for verifying the extension.  Uninstall add salesforce extension & reload.
Click the '...' button at the top of the extensions view and then 'install from vsix' 
select the generated vsix file 
The bundled extension should now be installed in your vscode and you can test with 
'SFDX: Create Function'

fix #4531 
@W-11998949@

### Functionality Before
Create Functions command fails for all types when used from a bundled core extension.

### Functionality After
Create Function works using bundled core extension.

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -25,7 +25,7 @@
     "Other"
   ],
   "dependencies": {
-    "@heroku/functions-core": "^0.4.0",
+    "@heroku/functions-core": "0.4.2",
     "@octokit/plugin-paginate-rest": "2.21.3",
     "@salesforce/core": "^3.23.2",
     "@salesforce/salesforcedx-sobjects-faux-generator": "56.4.0",
@@ -86,6 +86,7 @@
       "main": "dist/index.js",
       "dependencies": {
         "applicationinsights": "1.0.7",
+        "@heroku/functions-core": "0.4.2",
         "@salesforce/core": "3.30.9",
         "@salesforce/templates": "55.1.0",
         "@salesforce/source-deploy-retrieve": "6.0.4",
@@ -95,7 +96,7 @@
     }
   },
   "scripts": {
-    "bundle:extension": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode  --external:@salesforce/core --external:applicationinsights --external:shelljs --external:@salesforce/templates --external:@salesforce/source-deploy-retrieve --minify",
+    "bundle:extension": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode  --external:@salesforce/core --external:applicationinsights --external:shelljs --external:@salesforce/templates --external:@salesforce/source-deploy-retrieve --external:@heroku/functions-core --minify",
     "vscode:prepublish": "npm prune --production",
     "vscode:package": "ts-node -P ./tsconfig.json ../../scripts/vsce-bundled-extension.ts",
     "vscode:sha256": "node ../../scripts/generate-sha256.js >> ../../SHA256",
--- a/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
@@ -6,15 +6,9 @@
  */
 
 import {
-  Command,
-  SfdxCommandBuilder
-} from '@salesforce/salesforcedx-utils-vscode';
-import { LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode';
-import {
   CancelResponse,
   ContinueResponse,
-  FunctionInfo,
-  ParametersGatherer
+  FunctionInfo, LibraryCommandletExecutor, ParametersGatherer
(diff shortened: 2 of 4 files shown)


packages/salesforcedx-vscode-core/src/commands/templates/forceFunctionCreate.ts
```diff
@@ -5,16 +5,10 @@
...
-  SfdxCommandBuilder
-} from '@salesforce/salesforcedx-utils-vscode';
-import { LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode';
 import {
   CancelResponse,
   ContinueResponse,
-  FunctionInfo,
-  ParametersGatherer
+  FunctionInfo, LibraryCommandletExecutor, ParametersGatherer
 } from '@salesforce/salesforcedx-utils-vscode';
```
gbockus-sf, line 12: Just ran Organize imports b/c I was here and it needed it.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -25,7 +25,7 @@
     "Other"
   ],
   "dependencies": {
-    "@heroku/functions-core": "^0.4.0",
+    "@heroku/functions-core": "0.4.2",
     "@octokit/plugin-paginate-rest": "2.21.3",
```
gbockus-sf, line 29: So on review of this flow I realized it's probably not a good thing to have ^ deps in the dependancies list and not when we do the packaging so I just pinned both to 4.0.2 which was the current version.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -86,6 +86,7 @@
       "main": "dist/index.js",
       "dependencies": {
         "applicationinsights": "1.0.7",
+        "@heroku/functions-core": "0.4.2",
         "@salesforce/core": "3.30.9",
```
gbockus-sf, line 90: Add functions-core as a module to be installed for the packaging.  The size increase was pretty minimal I think, but will verify.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -95,7 +96,7 @@
     }
   },
   "scripts": {
-    "bundle:extension": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode  --external:@salesforce/core --external:applicationinsights --external:shelljs --external:@salesforce/templates --external:@salesforce/source-deploy-retrieve --minify",
+    "bundle:extension": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode  --external:@salesforce/core --external:applicationinsights --external:shelljs --external:@salesforce/templates --external:@salesforce/source-deploy-retrieve --external:@heroku/functions-core --minify",
     "vscode:prepublish": "npm prune --production",
```
gbockus-sf, line 100: Update the bundle command to not bundle functions-core

packages/salesforcedx-vscode-core/package.json
randi274: Curious what the difference was here?

gbockus-sf: previously the code was functions core was included in our 'bundle' 
This was the root of the bug as they depend on being able to read a templates directory in their package node module and it didn't exist after bundling.  I can show you if this makes no sense.

gbockus-sf: also I opened a bug in that repo about them not being 'bundle-able'.  We'll see if it can pick up some traction

gbockus-sf: Before: 
`Uploading /home/circleci/project/extensions/salesforcedx-vscode-core-56.4.0.vsix (24 MB): DONE`
After: 
Uploading /home/circleci/project/extensions/salesforcedx-vscode-core-56.4.0.vsix (29 MB): DONE
5 MB diff or a 20% size increase 😢 Not as trivial as I'd hoped

randi274 (approved):
The bump in size isn't great, but the good news is that functions are indeed working again! Was able to create a Java, JS, and TS function without issue. I did see this new error with the replay debugger, but am willing to chalk it up to weird issues I was having with CTV, so this is good to me.

Screen Shot 2022-11-03 at 11 58 33 AM

Issue #4536, opened 2022-11-04
Apex interactive debugger extension fails to load for v56.4.1


gbockus-sf:
### Summary

For version 56.4.1 we are seeing an issue with loading the Apex interactive debugger. 

### Expected result

Should activate on startup.

### Actual result

```
[2022-11-04 12:32:54.266] [exthost] [error] Error: ENOENT: no such file or directory, scandir '/Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/messages'
	at Object.readdirSync (node:fs:1392:3)
	at e.readdirSync (node:electron/js2c/asar_bundle:5:10830)
	at Ki.importMessagesDirectory (/Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:30:1530)
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:185:30717
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:1:326
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:201:4321
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:1:326
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:209:1399
	at /Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:1:326
	at Object.<anonymous> (/Users/gbockus/.vscode-insiders/extensions/salesforce.salesforcedx-vscode-apex-debugger-56.4.1/dist/index.js:209:8581)
	at u._compile (/Applications/Visual Studio Code - Insiders.app/Contents/Resources/app/out/vs/loader.js:4:1313)
	at Module._extensions..js (node:internal/modules/cjs/loader:1220:10)
	at Module.load (node:internal/modules/cjs/loader:1035:32)
	at Module._load (node:internal/modules/cjs/loader:876:12)
[...]

PR #4538, opened 2022-11-07
fix: update apex interactive debugger package step to include @salesforce/core

gbockus-sf:
### What does this PR do?
Update the apex interactive debugger extension to exclude @salesforce/core and install it directly in the packaged extension.

This was due to updating the apex debugger to use the common code in utils for getting the config directory.  B/c it pulled in utils and utils depends on core this required a move to having core external for the extension. 

#4536 @W-120117332@

### Functionality Before
apex interactive debugger failed to activate

### Functionality After
apex interactive debugger activates without issue.

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-debugger/package.json
@@ -47,7 +47,7 @@
   "scripts": {
     "bundle:extension": "npm run bundle:extension:build && npm run bundle:extension:copy",
     "bundle:extension:copy": "cp ../salesforcedx-apex-debugger/dist/apexdebug.js ./dist/",
-    "bundle:extension:build": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode --minify",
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -83,7 +83,8 @@
     "packageUpdates": {
       "main": "dist/index.js",
       "dependencies": {
-        "applicationinsights": "1.0.7"
+        "applicationinsights": "1.0.7",
+        "@salesforce/core": "3.30.9"
```
klewis-sfdc, line 87: Just to confirm, we want this to be pinned and not carroted?

gbockus-sf: correct. We want all the extensions to stay consistent 👍 good q though

klewis-sfdc (commented):
Looks good to me.  Just the one Q.  Seems like core is carroted in most of the other packages' pkg.json files.  

Re: QA, I built via the vsixes and saw Apex Interactive Debugger load and get included in the "running extensions" list ✅.  I don't have an org with the interactive debugger license, so can't QA by using the feature.

klewis-sfdc (approved):
👍🏽 Thanks, Gordon!

PR #4551, opened 2022-11-17
chore: update .sfdx ref in Org Browser folder

RitamAgrawal:
### What does this PR do?
Update all '.sfdx' ref in _salesforcedx-vscode-core_ package with a function call from _salesforcedx-utils-vscode_

 @W-12068891@

### Functionality Before
Creating folder path was hardcoded with string values.

### Functionality After
Creating folder path is dynamic with _projectPaths_. _projectPaths.metadataFolder()_ is used to create metadata directory path.


packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -45,15 +46,7 @@ export class TypeUtils {
...
-    const workspaceRootPath = workspaceUtils.getRootWorkspacePath();
-    const username = await OrgAuthInfo.getUsername(usernameOrAlias);
-    const metadataTypesPath = path.join(
-      workspaceRootPath,
-      '.sfdx',
-      'orgs',
-      username,
-      'metadata'
-    );
+    const metadataTypesPath = projectPaths.metadataFolder(usernameOrAlias);
```
gbockus-sf, line 49: So dumb question.  Can this username or alias ever be different what what we'd get out of the workspaceContextUtil? I did a quick trace of the code flow and it looks like outside of tests this value being passed if read the same way we read it in the WorkspaceContextUtil.  So does it make sense just to refactor here and remove the parameter and always assume the one provided by default?

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
```diff
@@ -14,9 +14,14 @@ jest.mock('@salesforce/core', () => {
 describe('test project paths', () => {
   const hasRootWorkspaceStub = jest.spyOn(workspaceUtils, 'hasRootWorkspace');
   const FAKE_WORKSPACE = '/here/is/a/fake/path/to/';
+  const FAKE_WORKSPACE_INSTANCE: any = {};
   const FAKE_STATE_FOLDER = '.sfdx';
   const FAKE_CONFIG_FILE = 'sfdx-config.json';
+  const FAKE_USERNAME = 'testuser';
+  const FAKE_METADATA = 'metadata';
+  const FAKE_ORGS = 'orgs';
```
gbockus-sf, line 22: We should export these values in paths.ts and re-use them here.  It doesn't make sense to create the constant in both places.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -45,15 +46,7 @@ export class TypeUtils {
       telemetryService.sendException('metadata_type_workspace', err);
```
gbockus-sf, line 46: Can we write a jest unit test for this method?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
gbockus-sf: I chatted with @klewis-sfdc and got some context of why this path was followed.  I still think it's the lesser of two evils here to remove the parameter for this method and just use the default value provided in paths vs passing it along.  We'll clean up the value tracked in this class with that follow up ticket to remove the custom listener to changes here in favor of the common one in workspace utils.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataType.test.ts
```diff
@@ -18,30 +18,24 @@ import { OrgAuthInfo, workspaceUtils } from '../../../src/util';
...
     getDefaultUsernameStub = stub(OrgAuthInfo, 'getDefaultUsernameOrAlias');
     getUsernameStub = stub(OrgAuthInfo, 'getUsername');
+    metadataFolderStub = stub(projectPaths, 'metadataFolder').returns(
+      metadataDirectoryPath
+    );
   });
   afterEach(() => {
     getDefaultUsernameStub.restore();
     getUsernameStub.restore();
   });
```
gbockus-sf, line 36: I think if you write some jest unit tests for this method we can just remove this int test as it's not really an int test anyway.

packages/salesforcedx-utils-vscode/test/jest/helpers/paths.test.ts
RitamAgrawal: Have exported METADATA, ORGS and other constants in paths.ts, which are reused in the test file.

RitamAgrawal:
The changes are reverted to the initial commit [1f789c5](https://github.com/forcedotcom/salesforcedx-vscode/pull/4551/commits/1f789c5727af391daca9f3fbc4296faf134bd976) and a new PR #4600 has been raised.

PR #4554, opened 2022-11-18
chore: update how we import command const and alter test 

gbockus-sf:
Just a place holder PR for showing updates to get things working again for int tests for PR #4514 

Fix two things here.
1. So the issue with running the integration tests in core was a circular dependency.  index.ts -> decorators.ts -> index.ts.  It def wasn't obvious and I only eventually found the solution by looking at what changed paired with the stack trace. Fix was to move the command constant from index to the constants file and 🏁 
2. After getting the int tests running again I was seeing a failure in the workspaceContest test.  There were two issues there.  The first was how the  stub for showOrg was getting created: 
`showOrgStub = env.stub(showOrg).resolves();` ❌ 
`showOrgStub = env.stub(decorators, 'showOrg').resolves();` ✅ 
The second was our old pass by value problem friend.  I fixed it by wrapping show org in an object.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/constants.ts
+++ b/packages/salesforcedx-vscode-core/src/constants.ts
@@ -49,3 +49,6 @@ export const AURA_PATH = '/force-app/main/default/aura/';
 export const APEX_CLASSES_PATH = '/force-app/main/default/classes/';
 export const LWC_PATH = '/force-app/main/default/lwc/';
 export const FUNCTIONS_PATH = '/functions/';
+
+// Commands
+export const ORG_OPEN_COMMAND = 'sfdx.force.org.open';
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceContext.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceContext.test.ts
@@ -9,7 +9,7 @@ import * as vscode from 'vscode';
 import { SFDX_CONFIG_FILE, SFDX_FOLDER } from '../../../src/constants';
 import { workspaceContextUtils } from '../../../src/context';
 import { WorkspaceContext } from '../../../src/context/workspaceContext';
-import { showOrg } from '../../../src/decorators';
+import { decorators } from '../../../src/decorators';
 import { workspaceUtils } from '../../../src/util';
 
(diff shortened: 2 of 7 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceContext.test.ts
```diff
@@ -9,7 +9,7 @@ import * as vscode from 'vscode';
 import { SFDX_CONFIG_FILE, SFDX_FOLDER } from '../../../src/constants';
 import { workspaceContextUtils } from '../../../src/context';
 import { WorkspaceContext } from '../../../src/context/workspaceContext';
-import { showOrg } from '../../../src/decorators';
+import { decorators } from '../../../src/decorators';
```
klewis-sfdc, line 12: 👍🏽

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceContext.test.ts
```diff
@@ -118,7 +118,7 @@ describe('WorkspaceContext', () => {
       .stub(workspaceContextUtil, 'username')
       .get(() => testUser);
     aliasStub = env.stub(workspaceContextUtil, 'alias').get(() => testAlias);
-    showOrgStub = env.stub(showOrg).resolves();
+    showOrgStub = env.stub(decorators, 'showOrg').resolves();
```
klewis-sfdc, line 121: 👍🏽 !

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -49,3 +49,6 @@ export const AURA_PATH = '/force-app/main/default/aura/';
 export const APEX_CLASSES_PATH = '/force-app/main/default/classes/';
 export const LWC_PATH = '/force-app/main/default/lwc/';
 export const FUNCTIONS_PATH = '/functions/';
+
+// Commands
+export const ORG_OPEN_COMMAND = 'sfdx.force.org.open';
```
klewis-sfdc, line 54: 👍🏽

CristiCanizales (approved):
Nicee! Thank you 🙌🏼🥹

PR #4555, opened 2022-11-22
Create dumb-action.yml

CristiCanizales:

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/.github/workflows/dumb-action.yml
+++ b/.github/workflows/dumb-action.yml
@@ -1,0 +1,35 @@
+name: Dumb action
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/dumb-action.yml
```diff
@@ -0,0 +1,35 @@
+name: Dumb action
```
gbockus-sf, line 1: I like yours better( dumb action heh) but I didn't see it before creating the same thing.

PR #4568, opened 2022-12-01
chore: force config set remove cli dependency

RitamAgrawal:
### What does this PR do?
- Removes CLI dependency in ForceConfigSet
- Incorporates sfdx-core V3
### Functionality Before
Implement `force:config:set` command using _SfdxCommandLetExecutor_, making it dependent on CLI

### Functionality After
Implement `force:config:set` using _LibraryCommandLetExecutor_, removing the dependency from CLI. It is using sfdx-core lib.


packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -134,4 +136,20 @@ export class ConfigUtil {
       : undefined;
     return username ? String(username) : undefined;
   }
+
+  public static async setDefaultUsernameOrAlias(usernameOrAlias: string): Promise<void> {
+    const originalDirectory = process.cwd();
```
RitamAgrawal, line 141: Need to add test for this one.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,85 @@
...
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should set config with the given username or alias', async () => {
+    orgStub.resolves();
+    await forceConfigSet(usernameOrAlias);
+    expect(setDefaultUsernameOrAliasStub.callCount).to.equal(1);
+    expect(setDefaultUsernameOrAliasStub.calledWith(usernameOrAlias)).to.equal(true);
+    // expect(configWriteSpy.callCount).to.equal(1);
```
RitamAgrawal, line 46: config.write is called within ConfigUtil.setDefaultUsernameOrAlias function.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,76 @@
...
+  it('should display correct output to user', async () => {
+    const outputTableRow = { name: DEFAULT_USERNAME_KEY, val: usernameOrAlias, success: String(true) };
+    const forceConfigSetInstance = new ForceConfigSetExecutor(usernameOrAlias);
+    const formatOutput = (forceConfigSetInstance as any).formatOutput(outputTableRow);
+    expect(tableSpy.callCount).to.equal(1);
+    expect(formatOutput).to.contain(nls.localize(CONFIG_SET_NAME), DEFAULT_USERNAME_KEY);
+    expect(formatOutput).to.contain(usernameOrAlias, String(true));
+  });
+
+  it('should not set config with an invalid username or alias', async () => {
```
klewis-sfdc, line 67: This test might not be needed.  It looks like its logic is also covered in the next test, so I think it'd be ok to remove this one.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,76 @@
...
+    expect(channelSpy.callCount).to.equal(2);
+  });
+
+  it('should display error message in output channel', async () => {
+    setDefaultUsernameOrAliasStub.throws(new Error(errorMessage));
+    await forceConfigSet(usernameOrAlias);
+    expect(channelSpy.callCount).to.equal(2);
+    expect(channelSpy.lastCall.args.length).to.equal(1);
+    expect(channelSpy.lastCall.args[0]).to.contain(errorMessage);
+  }).timeout(10000);
```
klewis-sfdc, line 79: Is this timeout still needed here?  Does the test fail without it?

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -8,10 +8,12 @@
 import {
   Config,
   ConfigAggregator,
+  Org,
   OrgConfigProperties,
   SfConfigProperties,
   StateAggregator
 } from '@salesforce/core';
+import { DEFAULT_USERNAME_KEY, getRootWorkspacePath } from '..';
```
klewis-sfdc, line 16: Looks like DEFAULT_USERNAME_KEY is unused and can be removed here.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -134,4 +136,22 @@ export class ConfigUtil {
       : undefined;
     return username ? String(username) : undefined;
   }
+
+  public static async setDefaultUsernameOrAlias(usernameOrAlias: string): Promise<void> {
```
klewis-sfdc, line 140: I think configUtil.test.ts needs some tests to cover this new function.  Those tests should cover:

* That the cwd is switched to the project root if necessary (and switched back to the orig cwd after)
* That the core lib is being called correctly.  i.e., when called with <username>, Org.create is called
  with <username>, and then Config.set() is called with "<OrgConfigProperties.TARGET_ORG>, <username>", etc.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,76 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import { Config, Org } from '@salesforce/core';
```
klewis-sfdc, line 8: Looks like Config and Org are unused here and can be removed.

klewis-sfdc (commented):
Looking really good Ritam.  The tests for forceConfigSet look great.  I think there should be some tests added for the new function in configUtil.  You can look at the tests for config agg provider for help, and let me know if you'd like to discuss!

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
RitamAgrawal: Tests are passing without the timeout. Removed it.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -144,6 +146,28 @@ export class ConfigUtil {
...
+    const originalDirectory = process.cwd();
+    // In order to correctly setup Config, the process directory needs to be set to the current workspace directory
+    const workspacePath = workspaceUtils.getRootWorkspacePath();
+    try {
+      if (usernameOrAlias) { // check if username is an empty string
+        await Org.create({ aliasOrUsername: usernameOrAlias }); // check if username is valid
+      }
```
gbockus-sf, line 157: nit: I could see pulling the above logic out into a `isUsernameValid` function.  It's a little concerning now that it's not very clear what happens if it is invalid.

packages/salesforcedx-utils-vscode/test/jest/config/configUtil.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it('should change the current working directory to the original working directory', async () => {
+    const username = 'vscodeO';
+    await ConfigUtil.setDefaultUsernameOrAlias(username);
+    expect(workspacePathStub).toHaveBeenCalledTimes(1);
+    expect(chdirStub).toHaveBeenCalledTimes(2);
+    expect(chdirStub).toHaveBeenNthCalledWith(1, fakeWorkspace);
+    expect(chdirStub).toHaveBeenNthCalledWith(2, fakeOriginalDirectory);
+  });
+
+  it('should set username or alias as default if it is an empty string', async () => {
```
gbockus-sf, line 44: is this test name accurate?  It's not default but the passed empty string correct? I might go with, `Should be able to set username to empty string.`

packages/salesforcedx-vscode-core/src/commands/forceConfigSet.ts
```diff
@@ -4,33 +4,72 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { ConfigUtil, LibraryCommandletExecutor } from '@salesforce/salesforcedx-utils-vscode/out/src';
+import { Row, Table } from '@salesforce/salesforcedx-utils-vscode/out/src/output';
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode/src/types';
```
gbockus-sf, line 9: The three above imports should be updated to just import from `@salesforce/salesforcedx-utils-vscode`

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -25,6 +25,12 @@ export const TELEMETRY_OPT_OUT_LINK =
   'https://developer.salesforce.com/tools/vscode/en/faq/telemetry';
 export const TELEMETRY_METADATA_COUNT = 'metadataCount';
 
+export const CONFIG_SET_EXECUTOR = 'force_config_set_org_text';
+export const CONFIG_SET_NAME = 'force_config_set_title';
+export const TABLE_NAME_COL = 'table_header_name';
+export const TABLE_VAL_COL = 'table_header_value';
+export const TABLE_SUCCESS_COL = 'table_header_success';
```
gbockus-sf, line 32: So close to alpha order...wanna swap the last two?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceConfigSet.test.ts
```diff
@@ -5,35 +5,69 @@
...
+    expect(formatOutput).to.contain(nls.localize(CONFIG_SET_NAME), DEFAULT_USERNAME_KEY);
+    expect(formatOutput).to.contain(usernameOrAlias, String(true));
+  });
+
+  it('should display error message in output channel', async () => {
+    setDefaultUsernameOrAliasStub.throws(new Error(errorMessage));
+    await forceConfigSet(usernameOrAlias);
+    expect(channelSpy.callCount).to.equal(2);
+    expect(channelSpy.lastCall.args.length).to.equal(1);
+    expect(channelSpy.lastCall.args[0]).to.contain(errorMessage);
```
gbockus-sf, line 71: nice tests 👍

gbockus-sf (commented):
Great stuff...Thanks for picking this up.  A few suggestions.

packages/salesforcedx-utils-vscode/test/jest/config/configUtil.test.ts
RitamAgrawal: No it's not! I need to update the name.

packages/salesforcedx-vscode-core/src/constants.ts
RitamAgrawal: Thanks for pointing this one out.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
gbockus-sf: Ritam and I dug into this together and found that this isn't validating the username, but actually verifying there is a valid connection.  She's going to update the comment to more clearly state what is being checked and leave the function as is.

RitamAgrawal: `Org.create()`creates a connection with the active usernameOrAlias and throws an error if the usernameOrAlias is expired. Will update the comment for it to make more sense.

packages/salesforcedx-vscode-core/src/constants.ts
RitamAgrawal: Checked it by swapping the order! It throws a linting error.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/auth/forceAuthLogout.test.ts
```diff
@@ -191,10 +191,10 @@ describe('Force Auth Logout Default', () => {
     expect(
       notificationStub.callCount,
       'should not have posted an error message'
-    ).to.equal(1);
+    ).to.equal(2);
     const notificationArgs = notificationStub.getCall(0).args;
     expect(notificationArgs).to.deep.equal([
-      'SFDX: Log Out from Default Org successfully ran',
+      'SFDX: Set a Default Org successfully ran',
       'Show',
```
RitamAgrawal, line 198: I have updated the tests for now and logged a new WI [W-12431853](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001K0MdvYAF/view) to address the two notification pop-ups while running `SFDX: Log Out from Default Org` to log out from a scratch org.

randi274 (approved):
QE Approved:
✅ Change default org w/ one Alias -> shouldn't be using CLI, should display as formatted table in the output channel
✅ Change default org w/ two Aliases -> should use the first one as the Value
✅ Change default org w/ no Alias -> shows the username as the Value
✅ Confirmed the same on Windows

PR #4570, opened 2022-12-06
ci: Add commit workflow in github Actions 

gbockus-sf:
### What does this PR do?
Add the commit validation and testing to github actions.  
Also enabled NX to speed up the process. 

@W-11949791@

### Functionality Before
Validations of PRs occurred via circle ci.

### Functionality After
PR validation and build happens in Github actions.


.github/workflows/buildAll.yml
```diff
@@ -0,0 +1,31 @@
...
+      - run: npm ci
+      - run: npm --version
+      - run: node --version
+      - run: npm run compile
+      - run: npm run vscode:package
+      - name: Stage Artifacts
+        run: |
+          mkdir extensions
+          find packages -name "*.vsix" -type f -exec cp {} ./extensions \;
+      - name: Upload Extensions
```
gbockus-sf, line 26: I'd debated if we actually need to create and upload the vsix files for every commit on a PR.  I thought about just verifying we could build the vsix file and not saving the artifacts.  We could then have a manual step for saving the vsixes as artifacts if we actually want to install them.  It would save us much time but it would save us a ton of space as I think there's a limit in GHA.  Figured I could swing back to this update since we already do this for all builds in circle.

.github/workflows/integrationTests.yml
```diff
@@ -0,0 +1,6 @@
+on:
+  workflow_call:
+
+jobs:
+  windows-integration-tests:
+    uses: ./.github/workflows/integrationTestsWindows.yml
```
gbockus-sf, line 6: You'll not I'm not reusing any of the shared GHA workflows.  This is primarily due to this repo using npm instead of yarn.

.github/workflows/nightlyBuildDevelop.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 10 20 * * *
+
+jobs:
+  nightly-build-develop:
+    uses: ./.github/workflows/buildAll.yml
+    with:
+      branch: develop
```
gbockus-sf, line 12: scheduled jobs to run the build on develop and main

.github/workflows/validatePR.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node_version }}
+          cache: npm
+      - run: npm ci
+      - name: Lint
+        run: npm run lint
```
gbockus-sf, line 23: validate using the shared workflow + a lint step

lerna.json
```diff
@@ -11,5 +11,6 @@
         "release/*"
       ]
     }
-  }
+  },
+  "useNx": true
```
gbockus-sf, line 15: NX greatly improves the output from tests and allows us to better paralyze the unit tests. Note this is already available as it's baked into lerna now I just had to turn it on.

package.json
```diff
@@ -66,12 +66,13 @@
     "lint:fix": "lerna run lint:fix",
     "publish": "node scripts/publish-circleci.js",
     "test": "lerna exec --concurrency 1 --stream --bail=false -- npm run test --if-present",
-    "test:unit": "lerna exec --concurrency 1 --stream --bail=false -- npm run test:unit --if-present",
+    "test:unit": "lerna run test:unit",
```
gbockus-sf, line 69: nx didn't play great using this `lerna exec` command but does fine with the standard `lerna run`

package.json
```diff
@@ -66,12 +66,13 @@
...
     "test:ui": "lerna exec --concurrency 1 --stream --bail=false -- npm run test:ui --if-present",
     "test:unit-and-system-tests": "lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test:unit --if-present && npm run coverage:system-tests",
+    "test:download-vscode": "node scripts/download-vscode",
     "test:integration": "lerna run test:integration --concurrency 1 --stream -- --if-present",
-    "test:integration-tests-without-system-tests": "node scripts/download-vscode && lerna exec --ignore system-tests --concurrency 1 --stream --bail=false -- npm run test:integration --if-present && npm run test:vscode-integration --if-present",
-    "test:vscode-integration": "node scripts/download-vscode && lerna exec --concurrency 1 --stream --bail=false -- npm run test:vscode-integration --if-present",
+    "test:integration-tests-without-system-tests": "npm run test:download-vscode && lerna run test:integration && lerna run test:vscode-integration --concurrency 1",
```
gbockus-sf, line 74: the `test:vscode-integration` is the only tests that get grumpy if you try to paralyze due to launching vscode.  Hence it's the only one that keeps the concurrency flag.

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -52,8 +52,8 @@
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules && shx rm -rf out && shx rm -rf coverage && shx rm -rf .nyc_output",
     "test": "npm run test:unit && npm run test:integration",
-    "test:unit": "jest --coverage",
-    "test:integration": "jest --coverage -c ./jest.integration.config.js"
+    "test:unit": "jest --coverage --forceExit",
+    "test:integration": "jest --coverage --forceExit -c ./jest.integration.config.js"
```
gbockus-sf, line 56: These tests alway complain about open handles and in GHA hang instead of exiting like they do in circle so I added the forceExit flag

packages/salesforcedx-utils-vscode/src/cli/forceConfigGet.ts
```diff
@@ -8,6 +8,8 @@
 import { SfdxCommandBuilder } from './commandBuilder';
 import { CliCommandExecutor } from './commandExecutor';
 import { CommandOutput } from './commandOutput';
+
+export const FORCE_CONFIG_GET_COMMAND = 'force:config:get';
```
gbockus-sf, line 12: added for unit testing

packages/salesforcedx-utils-vscode/test/jest/cli/forceConfigGet.test.ts
```diff
@@ -0,0 +1,117 @@
+import { ForceConfigGet } from '../../../src';
```
gbockus-sf, line 1: The existing unigration tests were being dumb so I just converted to jest unit tests.

packages/salesforcedx-utils-vscode/test/unit/commands/channelService.test.ts
```diff
@@ -83,7 +83,8 @@ describe('Channel Service', () => {
     expect(mChannel.value).to.contain('ended with exit code 0');
   });
 
-  it('Should pipe stderr on unsuccessful command execution', async () => {
+  // Fails locally and on GHA
+  it.skip('Should pipe stderr on unsuccessful command execution', async () => {
```
gbockus-sf, line 87: These tests had been failing locally for awhile and there is an existing ticket to dig into why so I started skipping them since they also fail in GHA.

gbockus-sf:
You can find the GHA validate run [here](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/3633777775)

packages/salesforcedx-utils-vscode/test/unit/commands/channelService.test.ts
randi274: @gbockus-sf should we put the reference here/elsewhere so we know where to look?

.github/workflows/validatePR.yml
randi274: Love this addition - fwiw if you don't intend to use multiple node versions, I don't think you need to include that strategy section (especially since you've set `fail-fast:false`)

gbockus-sf: Yeah good call...I originally had it running on lts/* and lts/-1 but in the end decided it wasn't worth doing the lint on multiple node version but left the setup out of laziness.  Think I should remove that strategy setup?

.github/workflows/nightlyBuild.yml
```diff
@@ -0,0 +1,22 @@
+name: Nightly Build
+on:
+  workflow_call:
+    inputs:
+      branch:
+        type: string
+        required: true
+        default: 'ubuntu-latest'
```
randi274, line 8: Is this a helpful default?

gbockus-sf: lulz...good catch

.github/workflows/buildAll.yml
gbockus-sf: yeah I was mostly debating if we actually need it...or if it could be on demand

PR #4574, opened 2022-12-07
chore: add names to all GHA workflows.  put slack notification back

gbockus-sf:
### What does this PR do?
Add Name for any github actions workflows that were missing it.  Put the slack notification back on the cutReleaseBranch job since it is now merged into main. 

@W-11949791@ 

### Functionality Before
Screen Shot 2022-12-07 at 5 11 46 PM

### Functionality After
Names will be listed instead of workflow files.

### DIFF
--- a/.github/workflows/buildAll.yml
+++ b/.github/workflows/buildAll.yml
@@ -1,3 +1,4 @@
+name: Build All
 on:
   workflow_call:
 
--- a/.github/workflows/createReleaseBranch.yml
+++ b/.github/workflows/createReleaseBranch.yml
@@ -64,3 +64,15 @@ jobs:
     - run: echo “Release Type is ${{ github.event.inputs.releaseType || 'minor' }}”
     - run: echo “Release Branch is ${{ steps.branch.outputs.branch }}”
 
+  slack_notification:
+    if: ${{ always() }}
+    needs: create_branch
+    uses: ./.github/workflows/slackNotification.yml
+    secrets: inherit
+    with:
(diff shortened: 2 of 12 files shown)


.github/workflows/buildAll.yml
```diff
@@ -1,3 +1,4 @@
+name: Build All
```
gbockus-sf, line 1: these names are displayed in the checks and on the Actions screen.

.github/workflows/createReleaseBranch.yml
```diff
@@ -64,3 +64,15 @@ jobs:
     - run: echo “Release Type is ${{ github.event.inputs.releaseType || 'minor' }}”
     - run: echo “Release Branch is ${{ steps.branch.outputs.branch }}”
```
gbockus-sf, line 66: removed due to causing issues for the job this week.  Figured I'd put it back here since I was editing GHA files.

.github/workflows/createReleaseBranch.yml
```diff
@@ -64,3 +64,15 @@ jobs:
...
 
+  slack_notification:
+    if: ${{ always() }}
```
randi274, line 68: Do you need the 'if' line here?

randi274 (approved):

klewis-sfdc (approved):

.github/workflows/createReleaseBranch.yml
gbockus-sf: heh..I asked the same q in the original PR.  From what I understand/remember it's required b/c it executes on success and failure.

PR #4578, opened 2022-12-08
fix: add missing Experience Cloud targets

CristiCanizales:
### What does this PR do?
This PR adds missing Experience Cloud targets to the js-meta XML validation.

@W-12188466@

### Functionality Before
The js-meta validation is missing two Experience Cloud targets for LWR sites.

### Functionality After
The js-meta validation no longer declares the valid targets as invalid.


randi274 (approved):

CristiCanizales:
Thanks for contributing @muenzpraeger !😁

Issue #3490, opened 2021-08-11
SFDX: Create Apex Class adds blank line


pfrandsen:
### Summary

When you create an Apex class via salesforcedx-vscode a blank line is added to the end of the file. This blank line is removed from the class when it is deployed to Salesforce. This means that the org and the source repository automatically differs.

This has implications for automated comparison between your VCS and your org - you basically get false positives that the source and the org are not alligned.

### Steps To Reproduce:

1. Create Apex class via the VS Code extension
1. Deploy the class to the org.
1. Run SFDX: Diff File Against Org
1. See that there is a blank line in difference

### Expected result

The local class and the class in the org should be the same.

### Actual result

Local file has a blank line after the class definition.

### Additional information

<img width="548" alt="blankline" src="https://user-images.githubusercontent.com/1440545/128967104-f06821c0-85f7-4e66-9202-f9a7f4097d96.png">

**Salesforce Extension Version in VS Code**:
v52.8.0

**SFDX CLI Version**:
sfdx-cli/7.112.1 win32-x64 node-v14.17.4

**OS and version**:
Windows 10

xyc:
Already tracked at @W-9515647@.

kaustubhdapurkar:
Hi, does this issue actually belong to the cli code repo or the extensions code repo? Because I just ran this sfdx command in a completely new directory without creating any org and then opened the subsequently created apex class file and the file had the extra line so I am guessing this issue is with the cli.

PR #4581, opened 2022-12-09
fix: fix extra line from default apex template and all default versions

randi274:
### What does this PR do?
Bumps the templates library, which contains a fix removing the additional line which is added when creating Apex Classes from the default class template.

This PR also includes the version bump that sets the default API version for all metadata to the latest major Salesforce version 56.

#3490 

Compare behavior before (left) with after (right):
Screen Shot 2022-12-08 at 6 19 24 PM


gbockus-sf (approved):
LGTM

PR #4582, opened 2022-12-09
ci: fix nightly workflow for branch handling

gbockus-sf:
### What does this PR do?
Update the GHA nightly workflows to fix the issue of being able to pass a branch to run on the other workflows. Note that the branch inputs are optional on all but the nightly workflow and should default to an empty string which will continuing using the default behavior.  

Also updated the time when the nightly tasks are run to be more middle of the night.

@W-11949791@

### Functionality Before
nightly runs for main/develop failing due to branch inputs not being supported for the underlying workflows. 

### Functionality After
Update to allow passing the specific branch.

### DIFF
--- a/.github/workflows/buildAll.yml
+++ b/.github/workflows/buildAll.yml
@@ -1,6 +1,10 @@
 name: Build All
 on:
   workflow_call:
+    inputs:
+      branch:
+        type: string
+        required: false
 
 jobs:
   build-all:
@@ -11,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.branch }}
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node_version }}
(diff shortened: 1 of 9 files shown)


.github/workflows/buildAll.yml
```diff
@@ -1,6 +1,10 @@
 name: Build All
 on:
   workflow_call:
+    inputs:
+      branch:
+        type: string
+        required: false
```
gbockus-sf, line 7: note that required is set to false so that we did not have to update the commit workflow and it will still default to the associated branch

.github/workflows/nightlyBuildDevelop.yml
```diff
@@ -3,7 +3,7 @@ name: Nightly Build Develop
 on:
   workflow_dispatch:
   schedule:
-    - cron: 10 20 * * *
+    - cron: 10 6 * * *
```
gbockus-sf, line 6: 6:10 UCT => 12:10 AM CST

randi274 (approved):

PR #4596, opened 2022-12-15
test: update flakey lwc test

gbockus-sf:
### What does this PR do?
Int test failed in the nightly build so I took a look at it and made it more reliable. 

@W-12216803@

### Functionality Before
Test would fail depending on the order of the values returned in the code lens.
Find the failure in this GHA run https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/3701606628/jobs/6271072025

### Functionality After
Test will pass regardless of position of the expected file in the results.

### DIFF
--- a/packages/salesforcedx-vscode-lwc/test/vscode-integration/navigation.test.ts
+++ b/packages/salesforcedx-vscode-lwc/test/vscode-integration/navigation.test.ts
@@ -25,7 +25,7 @@ describe('LWC Definition Linking', () => {
 
   it('Should provide navigation to a selected LWC tag', async () => {
     // select the 'c-view-source' tag
-    testDefinitionNavigation(
+    return testDefinitionNavigation(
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-lwc/test/vscode-integration/navigation.test.ts
```diff
@@ -25,7 +25,7 @@ describe('LWC Definition Linking', () => {
 
   it('Should provide navigation to a selected LWC tag', async () => {
     // select the 'c-view-source' tag
-    testDefinitionNavigation(
+    return testDefinitionNavigation(
```
gbockus-sf, line 28: for tests that have async on the funciton we either need to return a promise or use the done function.  I just updated to return the promise here.

randi274 (approved):

Issue #3953, opened 2022-03-25
SFDX Extension is not working for Scratch Orgs


Schuchie:
### Summary

I only can use this extension for Sandboxes. When I'm switching to an existing Scratch Org. Im not able to push/pull/depoy/retrieve any source via the context menu.

It all works for Sandboxes. I also switched between multiple Scratch Orgs (also different from Different Dev Hubs). It looks like all Scratch Orgs are affected:

Connected to a sandbox:
<img width="441" alt="image" src="https://user-images.githubusercontent.com/18148112/160122408-3dc2f21b-83ef-4500-9556-bcda38d74a19.png">

Connected to a scratch org:
<img width="390" alt="image" src="https://user-images.githubusercontent.com/18148112/160122287-ce0b46db-4a16-4422-a4af-5fd3da7031cd.png">

As you can see in the scratch org-screen: the sfdx: commands are missing completely in the context menu.

### Steps To Reproduce:

1.  Auth a scratch org via the following: sfdx auth:sfdxurl:store -f ./test-randomscratchorg@example.com.auth -a   test-randomscratchorg@example.com
2.  Set default org in the vscode-project
3. You shouldnt be able to deploy anything via the context menu
4. Try to switch to a sandbox
5. You should be able.

### Expected result

I can see all options from the salesforce extension in vscode. 

### Actual result

I can not see all options from the salesforce extension in vscode. 

### Additional information

I can also see some errors in the vscode-dev-console. This error only occurs on activating your extension:
<img width="418" alt="image" src="https://user-images.githubusercontent.com/18148112/160123466-4954b122-b784-4441-b5ec-631fe01fa920.png">

**Salesforce Extension Version in VS Code**:

**SFDX CLI Version**: 7.143.0-7049747

**OS and version**: macOS Big Sur 11.6

Schuchie:
Found a closed issue without any workaround or bugfix from your site: https://github.com/forcedotcom/salesforcedx-vscode/issues/3558

Schuchie:
Also maybe one more hint:
- The Auth-File is in a whole different folder and not in the same folder as the sfdx project. After the authentication I just set the default org in my sfdx project to the before authenticated org.

jag-sfdc:
[...]

Issue #1510, opened 2019-07-25
Enabled deploy/retrieve/delete on Scratch Orgs


ntotten:
Now that the CLI supports deploy/retrieve/delete on scratch orgs we should allow this option for customers who want to. Similar to Illuminated Cloud:

![image](https://user-images.githubusercontent.com/282782/61881938-78346200-aec5-11e9-943d-3c949c6ff988.png)

lcampos:
In this scenario a user would not have conflict detection, which is one of the benefits of using push/pull. Will a user understand that's what they are giving up ?

holapancho:
what is the status about that?

brno32:
I want this

aimaj-anz:
+1 for this, or at least a toggle to remove the feature suppression for scratch orgs

with the failure rate of push/pull when constantly switching large branches for reviewing and merging, this would be invaluable instead of typing out directory, path, or metadata type based deploy commands

smaddox-sf:
We have work inflight that will make push/pull and deploy/retrieve interoperable.  As that foundational work is complete, we'll be loosening up the options exposed in VS Code.

PR #4597, opened 2022-12-15
fix: Deploy* Commands Should Update Source Tracking When Used Against Source-tracked Sandboxes

klewis-sfdc:
### What does this PR do?
* Fixes the SFDX: Deploy* commands so that they handle source-tracking when used against source-tracked Sandbox orgs.
* Updates @salesforce/source-deploy-retrieve to version 8.0.1 from 7.15.0

@W-11309172@
@W-13051164@
#1510
#3953

### Functionality Before
* The Deploy* commands did not correctly fix up source tracking when used against a source-tracked Sandbox.

### Functionality After
* When used against a source-tracked Sandbox org, the commands correctly handle source tracking 
  * This includes conflict handling.  
* Source Tracking now acts as the Conflict Detector when deploying to a **source-tracked Sandbox** and will always be actively detecting changes during deploy operations when connected to this org type.  

**Todos:**
- [x] Review
- [x] Make sure Override Conflicts and Deploy button works
- [ ] QA
- [ ] Docs Updates


packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -28,12 +28,12 @@ export async function getWorkspaceOrgType(
       err.message
     )
   );
-  return isScratchOrg ? OrgType.SourceTracked : OrgType.NonSourceTracked;
+  return isScratchOrg ? OrgType.ScratchOrg : OrgType.NonScratchOrg;
```
klewis-sfdc, line 31: Sandboxes can be source-tracked and are not accounted for here.  This change makes the Enum more accurate.

packages/salesforcedx-vscode-core/src/commands/DeployExecutor.ts
```diff
@@ -0,0 +1,108 @@
...
+import { WorkspaceContext } from '../context/workspaceContext';
+import { handleDeployDiagnostics } from '../diagnostics';
+import { nls } from '../messages';
+import { SourceTrackingService } from '../services';
+import { DeployQueue } from '../settings';
+import { SfdxPackageDirectories } from '../sfdxProject';
+import { BaseDeployExecutor } from './baseDeployCommand';
+import { DeployRetrieveExecutor } from './baseDeployRetrieve';
+
+export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T> {
```
klewis-sfdc, line 25: Extracted to make unit testing easier

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -4,15 +4,21 @@
...
 import { telemetryService } from '../telemetry';
-import { OrgAuthInfo, workspaceUtils } from '../util';
+import { OrgAuthInfo } from '../util';
 
 export enum OrgType {
   SourceTracked,
   NonSourceTracked
 }
 
+export const workspaceContextOrgTypeUtil = {
```
klewis-sfdc, line 17: this made mocking easier for these two functions

packages/salesforcedx-vscode-core/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,40 @@
...
+    // It is important to pass the connection from WorkspaceContext to
+    // Org.create() here.  Without this, core uses its cached version
+    // of State Aggregator and the "No auth info found" error can be
+    // thrown when deploying or retrieving immediately after creating a
+    // default scratch org.
+    const org: Org = await Org.create({ connection });
+
+    const options: SourceTrackingOptions = {
+      org,
+      project,
```
klewis-sfdc, line 32: I realized since STL is initialized with the `SfProject`, which includes the project path, we can probably rely on its internal cache and we don't have to use `process.chdir` like with `ConfigAggregator`.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -32,6 +32,7 @@
     "@salesforce/salesforcedx-utils-vscode": "56.15.0",
```
gbockus-sf, line 32: I think the package-lock.json in the salesforcedx-vscode-core folder above was added by mistake.

packages/salesforcedx-vscode-core/src/commands/DeployExecutor.ts
```diff
@@ -0,0 +1,108 @@
+/*
```
gbockus-sf, line 1: nit: should we put these in a folder called `executors`?

packages/salesforcedx-vscode-core/src/services/sourceTrackingService.ts
gbockus-sf: nice...worth a code comment?

packages/salesforcedx-vscode-core/src/commands/index.ts
```diff
@@ -48,7 +48,7 @@ export {
 export {
   forceSourceDelete,
   ConfirmationAndSourcePathGatherer,
-  ForceSourceDeleteExecutor,
+ForceSourceDeleteExecutor,
```
gbockus-sf, line 51: whitespace looks off..might want to run the format on this file.

packages/salesforcedx-vscode-core/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { Org, SfProject } from '@salesforce/core';
```
gbockus-sf, line 8: 😢

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,95 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode';
+import { ComponentSet } from '@salesforce/source-deploy-retrieve';
+import * as fs from 'fs';
+import { DeployExecutor } from '../../../src/commands/DeployExecutor';
+import { WorkspaceContext } from '../../../src/context/workspaceContext';
+import { SourceTrackingService } from '../../../src/services';
+
+const dummyProjectPath = '/a/project/path';
```
gbockus-sf, line 14: hum...I'm kinda surprised this doesn't complain that `dummyProjectPath` isn't defined during the mocking of `source-deploy-retrieve`.  jest.mock hoists to the verify beginning of execution so it doesn't have access to var declared outside of it.  You can move `dummyProjectPath` inside the handler function of the mock and just have another var in the test to expect against.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,95 @@
...
+jest.mock('@salesforce/source-deploy-retrieve', () => {
+  return {
+    ...jest.requireActual('@salesforce/source-deploy-retrieve'),
+    getRootWorkspacePath: () => dummyProjectPath,
+    ComponentSet: jest.fn().mockImplementation(() => {
```
gbockus-sf, line 19: just curious.  How do you decide when to use mockImplementation vs just `jest.fn().mockReturnValue({bla: bla})`. I _think_ mockImplementation is only needs when the thing is an object to be instantiated, but I always forget

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      return {};
+    }),
+    TelemetryService: { getInstance: jest.fn() },
+    TelemetryBuilder: jest.fn()
+  };
+});
+
+jest.mock('../../../src/messages', () => {
+  return { loadMessageBundle: jest.fn(), nls: { localize: jest.fn() } };
+});
```
gbockus-sf, line 47: Just calling it out, but I always try to use the auto mocking if possible just to keep life simple. 

So it'd be something like: 
```
import {messages} from '../../../src/messages';
jest.mock('../../../src/messages');

const mockedMessages = jest.mocked(messages);
```
Then you can just use the fully mocked module like 
```
mockedMessages.nls.localize.mockReturnsValue(jest.fn());
```

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    WorkspaceContext,
+    'getInstance'
+  );
+  const mockWorkspaceContext = { getConnection: jest.fn() } as any;
+  const createSourceTrackingSpy = jest.spyOn(
+    SourceTrackingService,
+    'createSourceTracking'
+  );
+  const dummyComponentSet = new ComponentSet();
+  const deploySpy = jest.spyOn(dummyComponentSet, 'deploy');
```
gbockus-sf, line 61: These mocks seems like they should go in the beforeEach. That way when another test is added they are reset on every run.

gbockus-sf: either that or the top of the it if only applicable to that test

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Act
+    await (executor as any).doOperation(dummyComponentSet, {});
+
+    // Assert
+    expect(createSourceTrackingSpy).toHaveBeenCalled();
+    expect(deploySpy).toHaveBeenCalled();
+    const createSourceTrackingCallOrder =
+      createSourceTrackingSpy.mock.invocationCallOrder[0];
+    const deployCallOrder = deploySpy.mock.invocationCallOrder[0];
+    expect(createSourceTrackingCallOrder).toBeLessThan(deployCallOrder);
```
gbockus-sf, line 93: It's all so simple after you get there :P

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    WorkspaceContext,
+    'getInstance'
+  );
+  const mockWorkspaceContext = { getConnection: jest.fn() } as any;
+  const createSourceTrackingSpy = jest.spyOn(
+    SourceTrackingService,
+    'createSourceTracking'
+  );
+  const dummyComponentSet = new ComponentSet();
+  const retrieveSpy = jest.spyOn(dummyComponentSet, 'retrieve');
```
gbockus-sf, line 61: same story for the mocks here. constants are fine wrapped in the describe function

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    await (executor as any).doOperation(dummyComponentSet, {});
+
+    // Assert
+    expect(createSourceTrackingSpy).toHaveBeenCalled();
+    expect(retrieveSpy).toHaveBeenCalled();
+    const createSourceTrackingCallOrder =
+      createSourceTrackingSpy.mock.invocationCallOrder[0];
+    const retrieveCallOrder = retrieveSpy.mock.invocationCallOrder[0];
+    expect(createSourceTrackingCallOrder).toBeLessThan(retrieveCallOrder);
+  });
```
gbockus-sf, line 97: for these two tests are there any error conditions worth checking?  None jump out at me but figured it was worth asking.

packages/salesforcedx-vscode-core/test/jest/services/sourceTrackingService.test.ts
```diff
@@ -0,0 +1,43 @@
...
+describe('Source Tracking Service', () => {
+  const mockConnection = {} as any;
+  const getConnectionStub = jest.fn().mockImplementation(() => {
+    return mockConnection;
+  });
+  const mockWorkspaceContext = { getConnection: getConnectionStub } as any;
+  const workspaceContextGetInstanceSpy = jest.spyOn(
+    WorkspaceContext,
+    'getInstance'
+  );
```
gbockus-sf, line 26: before each me for the future

packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -39,15 +39,14 @@ import { SinonSpy, SinonStub, spy } from 'sinon';
...
 import { channelService } from '../../../src/channels';
 import { BaseDeployExecutor } from '../../../src/commands';
-import {
-  DeployExecutor,
-  DeployRetrieveExecutor,
-  RetrieveExecutor
-} from '../../../src/commands/baseDeployRetrieve';
+import { DeployRetrieveExecutor } from '../../../src/commands/baseDeployRetrieve';
+import { DeployExecutor } from '../../../src/commands/DeployExecutor';
+import { RetrieveExecutor } from '../../../src/commands/RetrieveExecutor';
```
gbockus-sf, line 44: can these three imports be combined at `../../../src/commands`? or (`../../../source/commands/executors`?)

gbockus-sf (commented):
awesome work.  Some comments for you consideration

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -45,6 +32,7 @@ export abstract class DeployRetrieveExecutor<
   T
 > extends LibraryCommandletExecutor<T> {
   protected cancellable: boolean = true;
+  protected sourceTrackingService = new SourceTrackingService();
```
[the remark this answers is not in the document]
klewis-sfdc: Nope, don't think it is needed in this file anymore

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: I rolled this back at some point, because ideally we want this to be accurate as `SourceTracked` and `NonSourceTracked`.  

I updated the function in [this commit](https://github.com/forcedotcom/salesforcedx-vscode/pull/4597/commits/78f4af382eaad370d70226e5fb854dc956c9640c) to use core to determine if the org is source-tracked.  Core's [tracksSource() function](https://github.com/forcedotcom/sfdx-core/blob/main/src/org/org.ts#L574) takes into account Sandboxes and will return true if the org is a scratch or or a source-tracked sandbox.

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -123,7 +123,7 @@ export abstract class LibraryCommandletExecutor<T>
   protected cancellable: boolean = false;
   private cancelled: boolean = false;
   private readonly executionName: string;
-  private readonly logName: string;
+  protected readonly logName: string;
```
klewis-sfdc, line 126: I tucked the conflict msgs into baseDeployRetrieve, and this is the only piece of data I could use to figure out which command was being run while inside the catch block that handles the SourceConflict error.

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,31 @@
...
+  public static async createSourceTracking(
+    projectPath: string,
+    connection: Connection
+  ): Promise<SourceTracking> {
+    const project = await SfProject.resolve(projectPath);
+    const org = await Org.create({ connection });
+    const options: SourceTrackingOptions = {
+      org,
+      project,
+      ignoreLocalCache: false,
```
klewis-sfdc, line 24: I set this to false per the docs, but I occasionally saw some issues with cacheing during dev.  QA should include testing with creating new files etc to see if any cacheing issues can be found.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -104,15 +129,89 @@ export abstract class DeployRetrieveExecutor<
...
+      const username = await ConfigUtil.getUsername();
+      await conflictChecker.handleConflicts(
+        componentPaths,
+        String(username),
+        diffs
+      );
+    }
+  }
+
+  private getMessagesFor(
```
klewis-sfdc, line 157: The SourceConflict error is thrown asynchronously, so in the catch we've lost a handle on which command is being executed or the conflict messages that should be used for each command.  I put them here and am using the logName property to determine the command that is being executed and then which messages to use.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,102 @@
+/*
```
klewis-sfdc, line 1: I added these tests as separate files, but the deploy and retrieve executors are still inside baseDeployRetrieve.ts in this PR.  I actually think this works well because extracting the executors made the PR much larger, and with this arrangement the tests are well organized but the PR doesn't blow up with the refactor to move the executors out.  LMK what you think.

packages/salesforcedx-vscode-core/src/commands/DeployExecutor.ts
klewis-sfdc: In the current PR I left this in place in baseDepRetr.  I think the PR is cleaner this way, but let me know if you think I should pull these out again and put them under a executors dir.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
klewis-sfdc: Yeah, I *think* I would bring in mockImplementation when the mocked thing is being used with the new keyword to get an instance.  Like here I'm sure I tried  `ComponentSet: jest.fn(),` first, then when that didn't work I tried using mockImplementation which did the trick.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
klewis-sfdc: I was able to remove this entirely since the test wasn't using it anymore.  I think that workspaceContext getInstance WI helped make the unit tests much simpler to implement.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
klewis-sfdc: I always forget about this pattern to get a handle on the mock of the whole dep.  I was able to remove this this time but will keep this pattern in mind!

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -140,7 +159,8 @@ export class MetadataCacheService {
     const operation = await components.retrieve({
       usernameOrConnection: connection,
       output: this.cachePath,
-      merge: false
+      merge: false,
+      suppressEvents: true
```
klewis-sfdc, line 163: This allows the cache to be loaded (file retrieved) even when it has conflicts.  VSCE stores the file in a tmp dir to facilitate the line-by-line conflict view.

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,37 @@
...
+    const org = await Org.create({ connection });
+    const options: SourceTrackingOptions = {
+      org,
+      project,
+      ignoreLocalCache: true,
+      subscribeSDREvents: true,
+      ignoreConflicts: false
+    };
+    const sourceTracking = await SourceTracking.create(options);
+    await sourceTracking.ensureLocalTracking();
```
klewis-sfdc, line 34: With ignoreLocalCache: false, newly created files would not be seen by STL and it would allow a newly created Apex file to overwrite more recently made changes on the server.  

By setting ignoreLocalCache to true, and calling ensureLocalTracking(), we force the STL instance to re-load its cache, which causes it to re-read the filesystem.  When it does that it "sees" the newly created files and will throw when conflicts exist in this scenario.

What is needed:


How to get there:


The local repo is private to the STL instance, otherwise we could make a direct call to it to bust the cache.  As it is currently, ignoreLocalCache=true and ensureLocalTracking() is the only way to get there:

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -104,22 +123,103 @@ export abstract class DeployRetrieveExecutor<
...
     token: vscode.CancellationToken
   ): Promise<DeployResult | undefined> {
+    const projectPath = getRootWorkspacePath();
+    const connection = await WorkspaceContext.getInstance().getConnection();
+
+    const sourceTracking = await SourceTrackingService.createSourceTracking(
+      projectPath,
+      connection
+    );
+    await sourceTracking.ensureLocalTracking();
```
klewis-sfdc, line 141: In my observances during dev, if this call was made inside utils/SourceTrackingService#createSourceTracking it did not work - meaning newly created files would not be picked up.  When I put the call here in core/baseDeployRetrieve it does work to pick up newly created files.  The "new Apex Class" scenario should be thoroughly tested before merge of this PR.

packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -54,6 +63,7 @@ export class TimestampConflictDetector {
       lastModifiedInCache = cache.getPropertiesForFile(key)?.lastModifiedDate;
       if (
         !lastModifiedInCache ||
+        skipTimestampCheck ||
```
klewis-sfdc, line 66: This loop both checks whether the file has a conflict (by checking timestamps) and then creates the diff for the file if it does have a conflict.

In the case of a SourceConflictError, this same function is being re-used to create the diffs, and the skipTimestampCheck option is used to skip the step where it checks the timestamps.  This seemed to keep the refactoring here to a minimum, but let me know if separating out these two functions (timestamp checking and diff creating) would make more sense here.

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
klewis-sfdc: I added one test here and one in the deployExecutor.test.ts file to cover the handling of SourceConflict errors.  Let me know if you think of any other scenarios that we should have a unit test for here.

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,36 @@
...
+    // contains the project path, and PR #4643 made it so that VSCE is
+    // running with process.cwd set as the project root, there
+    // is no need to call process.chdir here as has been done in VSCE
+    // with other core types like Config and ConfigAggregator.
+    const project = await SfProject.resolve(projectPath);
+    const org = await Org.create({ connection });
+    const options: SourceTrackingOptions = {
+      org,
+      project,
+      ignoreLocalCache: true,
```
klewis-sfdc, line 29: The [docs in the stl library](https://github.com/forcedotcom/source-tracking/blob/5c29282008987a778a83bc27e6030d0ffad28990/src/sourceTracking.ts#L68) say that this should be set to false when used in long-running processes, however, when I set this to false there is the the issue where it doesn't recognize new files and allows conflicts to be overridden.  To get it to recognize newly created files, I had to set `ignoreLocalCache` to true here and make the call to `ensureLocalTracking()` later after creating the instance.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "@heroku/functions-core": "0.7.0",
     "@octokit/plugin-paginate-rest": "2.21.3",
-    "@salesforce/core": "^3.31.18",
+    "@salesforce/core": "^3.33.6",
```
randi274, line 30: Elsewhere this is set at 3.31.18 - does this need to be consistent?

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
```diff
@@ -119,25 +119,6 @@ export const forceSourceRetrieveSourcePaths = async (
     uris.push(sourceUri);
   }
 
-  const messages: ConflictDetectionMessages = {
-    warningMessageKey: 'conflict_detect_conflicts_during_retrieve',
```
randi274, line 123: I'm still seeing this message in tests + i18n; was it not needed, or moved around?

packages/salesforcedx-vscode-core/src/conflict/conflictView.ts
```diff
@@ -98,17 +99,20 @@ export class ConflictView {
...
 
     extensionContext.subscriptions.push(this._treeView);
   }
 
-  private updateEnablementMessage() {
+  private async updateEnablementMessage() {
+    const orgType = await getWorkspaceOrgType();
     this.treeView.message =
-      sfdxCoreSettings.getConflictDetectionEnabled() || this.diffsOnly
+      sfdxCoreSettings.getConflictDetectionEnabled() ||
```
randi274, line 113: I find this hard to grok at a glance - so we're saying if conflict detection is enabled OR diffsOnly is true, OR the org is source tracked, then we shouldn't show an error about conflict detection not being enabled?

What's the scenario where we would care about the orgType but be able to get here not setting diffsOnly (diff the folder command) or having conflict detection enabled?

randi274: Which I suppose is my way of asking if this is tested/where 😄

packages/salesforcedx-vscode-core/test/jest/conflict/metadataCacheService.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import * as sdrUtils from '../../../src/services/sdr/componentSetUtils';
+
+describe('MetadataCacheService', () => {
+  describe('createRetrieveOperation', () => {
+    const dummyComponentSet = new ComponentSet([
+      { fullName: 'Test', type: 'apexclass' },
+      { fullName: 'Test2', type: 'layout' }
+    ]);
+    let retrieveStub: jest.SpyInstance;
+    let getSourceComponentsStub: jest.SpyInstance;
```
randi274, line 13: Weirdly in my IDE I'm seeing 13, 15, and 16 aren't being called, but they clearly are referenced in the beforeEach... 🤔

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
```diff
@@ -0,0 +1,745 @@
+import { MetadataCacheService } from '../../../src/conflict';
+import * as diffUtils from '../../../src/conflict/componentDiffer';
+import { PersistentStorageService } from './../../../src/conflict/persistentStorageService';
+import { TimestampConflictDetector } from './../../../src/conflict/timestampConflictDetector';
+
+describe('TimestampConflictDetector', () => {
+  const dummyLastModifiedDateCache = '2023-03-17T17:52:51.000Z';
```
randi274, line 7: This is a fantastic example of stubbing! 👏  It does make the test a bit hard to read though. Would you be opposed to moving everything from line 7 - 683 (or at least 655) to an external file? That then gets this file down to about 70 lines and is much more readable.

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
```diff
@@ -0,0 +1,745 @@
...
+      );
+
+      expect(correlateResultsStub).toHaveBeenCalledWith(
+        dummyMetadataCacheResult
+      );
+      expect(persistentStorageServiceMock).toHaveBeenCalled();
+      expect(diffComponentsStub).toHaveBeenCalledTimes(2);
+      expect(diffs.different.size).toBe(4);
+    });
+  });
```
randi274, line 744: What would you think about adding 1 more test that returns no results when the timestamp conflict detector is not tripped?

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
```diff
@@ -0,0 +1,745 @@
+import { MetadataCacheService } from '../../../src/conflict';
+import * as diffUtils from '../../../src/conflict/componentDiffer';
+import { PersistentStorageService } from './../../../src/conflict/persistentStorageService';
+import { TimestampConflictDetector } from './../../../src/conflict/timestampConflictDetector';
+
+describe('TimestampConflictDetector', () => {
```
randi274, line 6: Also I noticed you didn't make any changes to the vscode-integration test for timestampConflictDetector.test.ts. That one seems to be more of a unit test than a uni-gration, since we're not importing vscode to test. Is that test file superseded by this one? ... can we 💣 it?!

packages/salesforcedx-vscode-core/test/jest/conflict/metadataCacheService.test.ts
```diff
@@ -0,0 +1,52 @@
+import { ComponentSet } from '@salesforce/source-deploy-retrieve';
+import { MetadataCacheService } from '../../../src/conflict';
+import { WorkspaceContext } from '../../../src/context';
+import * as sdrUtils from '../../../src/services/sdr/componentSetUtils';
+
+describe('MetadataCacheService', () => {
```
randi274, line 6: Alright similar question here as for for timestampConflictDetector.test.ts under vscode-integration. That one would be a lot more to move over, so might not be the right time, but is it worth making any changes over there?

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -211,7 +212,13 @@ export class TimestampConflictChecker implements PostconditionChecker<string> {
...
   ): Promise<ContinueResponse<string> | CancelResponse> {
-    if (!sfdxCoreSettings.getConflictDetectionEnabled()) {
+    // If the current org is source-tracked, then source tracking
+    // will handle conflict detection.
+    const orgType = await getWorkspaceOrgType();
+    if (
+      orgType === OrgType.SourceTracked ||
+      !sfdxCoreSettings.getConflictDetectionEnabled()
+    ) {
       return inputs;
```
randi274, line 222: So if the user has a source tracked org, but has conflict detection turned on, conflict detection will just quietly get skipped?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postconditionCheckers.test.ts
```diff
@@ -440,6 +444,10 @@ describe('Postcondition Checkers', () => {
       settingsStub = env.stub(sfdxCoreSettings, 'getConflictDetectionEnabled');
       conflictViewStub = env.stub(conflictView, 'visualizeDifferences');
       appendLineStub = env.stub(channelService, 'appendLine');
+      env.stub(WorkspaceContext, 'getInstance').returns(mockWorkspaceContext);
```
randi274, line 447: Not that I want to bloat this file any more but... is it worth adding a test where the OrgType is SourceTracked?

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
randi274: +1 for not blowing up the PR 👏  that big a refactor should probably go separately, and I'm not sure it's worth touching such core functionality if you were still able to accomplish enough testing here.

packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveSourcePath.ts
klewis-sfdc: Good eye.  It is not used at all.  Two routes: remove the tests and its entry in the i18 file, or leave this, since it will be needed again someday when conflict detection is enabled for retrieve.  Thoughts?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -16,30 +18,30 @@ export enum OrgType {
 export async function getWorkspaceOrgType(
   defaultUsernameOrAlias?: string
```
randi274, line 19: Looks like this isn't used - can we remove?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -16,30 +18,30 @@ export enum OrgType {
...
-  const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username).catch(err =>
-    telemetryService.sendException(
-      'get_workspace_org_type_scratch_org',
-      err.message
-    )
-  );
-  return isScratchOrg ? OrgType.SourceTracked : OrgType.NonSourceTracked;
+  const connection = await WorkspaceContext.getInstance().getConnection();
+  const org: Org = await Org.create({ connection });
+  // Org.supportsSourceTracking() checks the org at this point in time.  This
```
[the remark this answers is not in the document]
randi274: Or if it is worth keeping in code, could we move to a more traditional function-level /**/ comment?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -16,30 +18,30 @@ export enum OrgType {
...
+  // is important because a sandbox can be created without source tracking and
+  // then be refreshed to have source tracking.  So, in the case of sandboxes,
+  // we have to check the org's source tracking status in real time and can't
+  // rely on Org.tracksSource() because that property is populated at authorization-
```
randi274, line 27: Did you look into whether or not we should be updating tracksSource directly rather than using this workaround?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -16,30 +18,30 @@ export enum OrgType {
...
+  // Also opting to use this call only and specifically NOT use org.isScratch(),
+  // which was not returning consistent results during dev.
+  const isSourceTracked = await org.supportsSourceTracking();
+  return isSourceTracked ? OrgType.SourceTracked : OrgType.NonSourceTracked;
 }
 
 export function setWorkspaceOrgTypeWithOrgType(orgType: OrgType) {
   setDefaultUsernameHasChangeTracking(orgType === OrgType.SourceTracked);
-  setDefaultUsernameHasNoChangeTracking(orgType === OrgType.NonSourceTracked);
+  setDefaultUsernameHasNoChangeTracking(orgType !== OrgType.SourceTracked);
```
randi274, line 38: Was the thought for changing this just to track the state of SourceTracked only?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -16,30 +18,30 @@ export enum OrgType {
...
 }
 
 export async function setupWorkspaceOrgType(defaultUsernameOrAlias?: string) {
   try {
     setHasDefaultUsername(!!defaultUsernameOrAlias);
-    const orgType = await getWorkspaceOrgType(defaultUsernameOrAlias);
```
randi274, line 42: This might be out of scope but since you're touching this file... does anything actually use the getDefaultUsernameOrAlias? I'm wondering if we could remove that from this file and the vscode-integration test.ts to continue that cleanup too.

packages/salesforcedx-vscode-core/test/vscode-integration/context/workspaceOrgType.test.ts
```diff
@@ -88,114 +95,50 @@ describe('workspaceOrgType unit tests', () => {
...
         defaultUsername
       );
-
       expect(orgType).to.equal(OrgType.NonSourceTracked);
-      expect(createStub.getCall(0).args[0]).to.eql({
-        username: defaultUsername
-      });
-    });
-
-    it('throws an error when no defaultusername is set', async () => {
```
randi274, line 123: I LOVE that we're not throwing an unnecessary error anymore ❤️

packages/salesforcedx-vscode-core/test/jest/context/workspaceOrgType.test.ts
```diff
@@ -0,0 +1,42 @@
+import { Org } from '@salesforce/core';
```
randi274, line 1: So neat and tidy. Yay Jest!!

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -0,0 +1,36 @@
...
+  SourceTracking,
+  SourceTrackingOptions
+} from '@salesforce/source-tracking';
+
+export class SourceTrackingService {
+  public static async createSourceTracking(
+    projectPath: string,
+    connection: Connection
+  ): Promise<SourceTracking> {
+    // Since SourceTracking is initialized with an SfProject, which
```
randi274, line 19: Could we also move this comment up and out of the function?

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
randi274: That comment definitely seems like it was made for us 😄 do we need to file a bug there slash is that already done?

packages/salesforcedx-utils-vscode/test/jest/services/sourceTrackingService.test.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { SourceTracking } from '@salesforce/source-tracking';
+import { SourceTrackingService } from '../../../src/services';
+
+jest.mock('@salesforce/core', () => ({
+  ...jest.requireActual('@salesforce/core'),
```
randi274, line 11: oh this is interesting! What benefit did `requireActual` provide here?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -217,6 +325,16 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
     return operation.pollStatus();
   }
 
+  protected async handleSourceConflictError(e: any) {
+    // Retrieve operation - proceed and do not handle or throw.
```
randi274, line 329: can this also go up above?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -195,14 +295,22 @@ export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T> {
 }
 
 export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
+  public readonly sourceConflictErrorInfoMsg =
```
randi274, line 298: Can this be moved to ../messages/i18n.ts?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -104,22 +123,103 @@ export abstract class DeployRetrieveExecutor<
...
-      usernameOrConnection: await WorkspaceContext.getInstance().getConnection()
+      usernameOrConnection: connection
     });
 
     this.setupCancellation(operation, token);
 
     return operation.pollStatus();
   }
 
+  protected async handleSourceConflictError(e: any) {
```
randi274, line 152: Is this like an "error error" where something has gone wrong with our handling of source conflicts? Would it be worth logging with telemetry if so?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -195,14 +295,22 @@ export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T> {
...
+  public readonly sourceConflictErrorInfoMsg =
+    'SourceConflictError reported.  Use SFDX: Diff File Against Org and SFDX: Diff Folder Against Org to detect and view conflicts in advance of any retrieve operation.';
+
   protected async doOperation(
     components: ComponentSet,
     token: vscode.CancellationToken
   ): Promise<RetrieveResult | undefined> {
+    const projectPath = getRootWorkspacePath();
     const connection = await WorkspaceContext.getInstance().getConnection();
+    const sourceTracking = await SourceTrackingService.createSourceTracking(
```
randi274, line 307: I noticed above you're calling `ensureLocalTracking` on sourceTracking. Do we not care about that here?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
randi274: Can you give an example of the test flow you're concerned about for this?

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,185 @@
...
+    expect(createSourceTrackingSpy).toHaveBeenCalled();
+    expect(ensureLocalTrackingSpy).toHaveBeenCalled();
+    expect(deploySpy).toHaveBeenCalled();
+    const createSourceTrackingCallOrder =
+      createSourceTrackingSpy.mock.invocationCallOrder[0];
+    const ensureLocalTrackingSpyCallOrder =
+      ensureLocalTrackingSpy.mock.invocationCallOrder[0];
+    const deployCallOrder = deploySpy.mock.invocationCallOrder[0];
+    // In order to be sure that a Source Tracking instance is initialized
+    // and tracking files appropriately, create and ensureLocalTracking
```
randi274, line 136:
```suggestion
    // and tracking files appropriately, createSourceTracking and ensureLocalTracking
```

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
```diff
@@ -0,0 +1,185 @@
...
+    // need to be called before the deploy operation is started.
+    expect(createSourceTrackingCallOrder).toBeLessThan(deployCallOrder);
```
randi274, line 138: So this is just a way of looking at the order in which these are called, and making sure that createSourceTracking and ensureLocalTracking come before deployCallOrder? 

Also, is the 3rd assert a dupe of the first?

packages/salesforcedx-vscode-core/test/jest/commands/util/baseDeployRetrieve.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    setApiVersionOnStub = jest
+      .spyOn(sdrUtils, 'setApiVersionOn')
+      .mockImplementation(jest.fn());
+    formatExceptionStub = jest
+      .spyOn(betaDeployRetrieveUtils, 'formatException')
+      .mockImplementation(jest.fn());
+  });
+
+  describe('run', () => {
+    it('should handle a SourceConflictError', async () => {
```
randi274, line 96: Maybe a 3rd test for the absence of any error? Not wanting to repeat too much of what's covered in the vscode-integration test, but that's the 3rd option here right?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -77,7 +88,15 @@ export abstract class DeployRetrieveExecutor<
         status === RequestStatus.SucceededPartial
       );
     } catch (e) {
-      throw formatException(e);
+      if (e.name === 'SourceConflictError') {
+        this.handleSourceConflictError(e);
+        return true;
```
randi274, line 93: Why do we return `true` here instead of one of the `RequestStatus` enums?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -104,22 +123,103 @@ export abstract class DeployRetrieveExecutor<
...
+    const cacheResult = await metadataCacheService.loadCache(
+      componentPaths,
+      workspaceUtils.getRootWorkspacePath(),
+      false
+    );
+
+    const detector = new TimestampConflictDetector();
+    const diffs = detector.createDiffs(cacheResult, true);
+
+    const conflictMessages = this.getMessagesFor(this.logName);
```
randi274, line 167: Are `conflictMessages` and the following conditional being tested for? I found that block code a little tricky to read, so it'd be nice to have some additional coverage there.

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  it('should not handle a SourceConflict error, but should send info about conflicts to the console', async () => {
```
randi274, line 90: Is this test statement correct? It seems more like we're sending a generic error than anything specific about the conflicts.

packages/salesforcedx-utils-vscode/test/jest/services/sourceTrackingService.test.ts
klewis-sfdc: Working through these unit tests was an exercise in mocking lots of dependencies.  This one mocks SDR stuff that SDR depends on.  Without it the test fails with:

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Well, at this point the request did not succeed.  It threw an exception which was caught and handled.  So I think the true response is like, yep it ran.

If a status were returned, I'm not sure which one it would be:

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
randi274: Trying to jump around in the code to see where else we're aware of potential errors, and it looks like the postconditionChecker is wrapping loadCache, createDiffs, and handleConflicts all in a try block. There don't look to be any tests with more helpful information though... Not sure if they were just being overly cautious, but might be something to consider.

https://github.com/forcedotcom/salesforcedx-vscode/blob/ea0dc1b133a2e25dfdbceec2aebd611c08154bca/packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts#L237
https://github.com/forcedotcom/salesforcedx-vscode/blob/ea0dc1b133a2e25dfdbceec2aebd611c08154bca/packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postconditionCheckers.test.ts#L577

randi274 (commented):
Dang @klewis-sfdc another big feature for you! This one has done a LOT so I've left a bunch of questions and mostly testing feedback. Let me know when this is ready for re-review and when you'd like to talk testing strategy.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: No, the SourceConflictError is what STL throws when it detects conflicts.  Up above in DeployRetrieveExecutor run(), when a SourceConflictError is caught, the subclasses handleSourceConflictError() method is called so it can handle the error.  In a deploy operation, the conflict view is setup and shown, and in the RetrieveExecutor it is not handled currently.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Yep, repo steps are included in the PR description up above (buried in my checklist of todos):

 investigate fix potential issue with stl cacheing and possibly not recognizing new files
Repro:
Build this branch, launch Extension Host
Create a New sfdx project
Auth a DevHub
Auth a Scratch Org
Create a New Apex Class
Deploy the class to the scratch org
Modify the class locally
Modify the class on the server
Deploy the class to the scratch org
Expected:
Conflict
Actual:
class is deployed and overwrites the version on the server

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
randi274: Gotcha - so the experience for the user would be the same?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
randi274: This one might be good to discuss at 11sies if it could potentially have a meaningful impact in terms of code or functionality. For instance, when Apex tests run but return failures we show something like "the command failed to run", and in this case, it did fail to complete the deploy/retrieve operation because of those conflicts, right?

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: The Org.tracksSource method first checks the Auth file, and if the org has already been auth'd, it just grabs the property from there.  This method only makes a call to check the actual org if the Org has not yet been Auth'd.  So the core lib is doing some cacheing and trying to only make calls to the org when necessary.  It is useful as is for CLI so I don't think it should be changed for this PR.

The org.supportsSourceTracking() method always calls out to the org to get its source-tracking enabled state.  

In development I would see Org.tracksSource return false for scratch orgs sometimes in VSCE.   It seems built for transactional use in CLI, not long-running process use in VSCE.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: FYI this comment is outdated.  See discussions below about deciding to use org.supportsSourceTracking().

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Good call.  It is inconsistent because we are not actually using STL to determine conflicts on retrieve operations (currently).  It is possible that when conflict handling for retrieve operations is implemented that VSCE will need to call `ensureLocalTracking` or maybe `ensureRemoteTracking` here in the RetrieveExecutor, but for now it is not needed because VSCE doesn't do anything when STL throws the conflict error in a retrieve op anyway.

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
klewis-sfdc: Yep exactly.  And yes, third one was unneeded.  Good catch! Removed.

packages/salesforcedx-vscode-core/test/jest/commands/util/baseDeployRetrieve.test.ts
klewis-sfdc: Well, my intention here was to cover only the new functionality in unit tests, and maybe have a follow up WI to convert the rest of the integ tests.  Absence of any error would just be a successful deploy and testing those steps, which I think are covered pretty well in [the integ tests](https://github.com/forcedotcom/salesforcedx-vscode/blob/96eab0c65d44965e6286e32eee4063f7bba05d74/packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts#L256).

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Well the difference here is that the user has the whole conflict experience in the UI where they choose how to "finish" the operation.  So an extra notification that the deploy op didn't succeed might not be needed.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Yep.  I refactored these out into a separate file since they are now shared between multiple modules.  Open to any ideas for better way of doing this!

klewis-sfdc: I'll take a look re: test coverage.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: I revisited this and decided to remove it.  There is no equivalent console.info msg for the retrieve operation when retrieving against a non-src tracked org so it makes the operations slightly different depending on org.  Plus it seems kinda confusing to get an info msg about the conflicts but then still overwrite the file.  Happy to discuss further.

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
klewis-sfdc: I agree, I updated it at first, then realized this was a not so great test anyway and removed it.  The method has no implementation now, so I think this test can wait until conflict detection is implemented for retrieve operations.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: Yeah, part of the reason is that `SourceConflictError` in STL extends `SfError` from core:


which is [an interesting type](https://github.com/forcedotcom/sfdx-core/blob/4dbd4907f34e5c8495a21c5be31b2cf80f52a90d/src/sfError.ts#L27) in itself!:
(it can have actions that help the users get past errors)

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: 👍🏽 Updated this method!  LMK what you think of the updates.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: Can't remember why I changed this originally, but its not a necessary change.  Rolled it back.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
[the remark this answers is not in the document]
klewis-sfdc: Since we know that refreshing a Sb requires re-auth, I'm going to switch this back to use tracksSource().  If it isn't behaving as expected we will see that in QA.

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: See above comment about switching to use tracksSource().

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: Good call.  Removed!

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
klewis-sfdc: Yeah, it looks like that integ test file should be superseded by this one, but there is probably some additional coverage to port from that test to this unit test file.  Follow up WI?

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
klewis-sfdc: Currently, I think this is covered in one of the integration tests. I would love to fully port, does integration test over into this unit test file, but would also love to do it as a follow up work item. :-)

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: I added a test file to unit test this messages module.  Let me know if you have any feedback on it!

packages/salesforcedx-vscode-core/package.json
```diff
@@ -157,7 +157,7 @@
         {
```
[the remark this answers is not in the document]
klewis-sfdc: Thats the one line change that will enable Org Browser! #4744

packages/salesforcedx-vscode-core/package.json
klewis-sfdc: I think as long as the utils-vscode and vscode-core packages get the same version of core installed everything is ok.  But not sure if we should align the version in the other pkgs as part of this PR.

klewis-sfdc:
Closing temporarily to get Snyk check to re-run

klewis-sfdc:
Reopening

packages/salesforcedx-vscode-core/test/jest/conflict/timestampConflictDetector.test.ts
klewis-sfdc: @randi274 I created a follow up WI to finish the conversion of all conflict-related integration tests to jest unit tests: [W-13031042](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001PO56cYAD/view).

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -78,7 +84,15 @@ export abstract class DeployRetrieveExecutor<
...
+        return true;
+      } else {
+        // Error, but not a Source Conflict Error.  Prior to adding
+        // SourceTracking, this was the only statement in the catch
+        // block.
+        throw formatException(e);
+      }
     } finally {
       await this.postOperation(result);
     }
```
jeffb-sfdc, line 98: Since this method returns a boolean, at line 99 it should probably return false.

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -4,21 +4,28 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { ContinueResponse, SfdxCommandBuilder } from '@salesforce/salesforcedx-utils-vscode';
+import {
+  ContinueResponse,
+  SfdxCommandBuilder
```
jeffb-sfdc, line 9: SfdxCommandBuilder isn't needed

packages/salesforcedx-vscode-core/src/commands/forceSourceDeployManifest.ts
```diff
@@ -61,22 +68,19 @@ export async function forceSourceDeployManifest(manifestUri: vscode.Uri) {
...
-        .toString();
-    }
-  };
-
-  const commandlet = new SfdxCommandlet(
-    new SfdxWorkspaceChecker(),
-    new FilePathGatherer(manifestUri),
-    new LibrarySourceDeployManifestExecutor(),
-    new TimestampConflictChecker(true, messages)
+  const messages:
```
jeffb-sfdc, line 71: The typing here isn't needed, since the construction and assignment are inline and the type inferred by the function called.

```
  const messages = getConflictMessagesFor(
    'force_source_deploy_with_manifest_beta'
  );
```

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -82,36 +83,25 @@ export const forceSourceDeploySourcePaths = async (
...
-
-  const commandlet = new SfdxCommandlet<string[]>(
-    new SfdxWorkspaceChecker(),
-    new LibraryPathsGatherer(uris),
-    new LibraryDeploySourcePathExecutor(),
-    new CompositePostconditionChecker(
-      new SourcePathChecker(),
-      new TimestampConflictChecker(false, messages)
-    )
+  const messages:
```
jeffb-sfdc, line 86: Same comment about how the type isn't needed since it's inferred by `getConflictMessagesFor()`

packages/salesforcedx-vscode-core/src/commands/forceSourceDeploySourcePath.ts
```diff
@@ -9,6 +9,7 @@ import { ContinueResponse } from '@salesforce/salesforcedx-utils-vscode';
 import { ComponentSet } from '@salesforce/source-deploy-retrieve';
 import * as vscode from 'vscode';
 import { channelService } from '../channels';
+import { getConflictMessagesFor } from '../conflict/messages';
```
jeffb-sfdc, line 12: SfdxCommandBuilder on line 7 isn't needed

packages/salesforcedx-vscode-core/src/commands/util/postconditionCheckers.ts
```diff
@@ -294,7 +305,7 @@ export class TimestampConflictChecker implements PostconditionChecker<string> {
       );
 
       if (choice === nls.localize('conflict_detect_override')) {
-        conflictView.visualizeDifferences(
+        await conflictView.visualizeDifferences(
```
jeffb-sfdc, line 308: Nice catch with the missing `await` statements!

packages/salesforcedx-vscode-core/src/conflict/messages.ts
```diff
@@ -0,0 +1,45 @@
+import { SfdxCommandBuilder } from '@salesforce/salesforcedx-utils-vscode';
+import { ConflictDetectionMessages } from '../commands/util';
+
+const messagesByLogName: Map<string, ConflictDetectionMessages> = new Map();
```
jeffb-sfdc, line 4: Nit-pick here.  `messagesByLogName` and `warningMessageKey` get created when this file is imported.  That in of itself isn't too bad, but then `messagesByLogName.set()` is called twice.  So when this file is imported, that means this file executes code, and this makes it hard to write unit tests for.

I would rather see something like this:
```
import { SfdxCommandBuilder } from '@salesforce/salesforcedx-utils-vscode';
import { ConflictDetectionMessages } from '../commands/util';

export function getConflictMessagesFor(
  logName: string
): ConflictDetectionMessages | undefined {
  const messagesByLogName: Map<string, ConflictDetectionMessages> = new Map();
  const warningMessageKey = 'conflict_detect_conflicts_during_deploy';

  messagesByLogName.set('force_source_deploy_with_sourcepath_beta', {
    warningMessageKey,
    commandHint: inputs => {
      const commands: string[] = [];
      (inputs as string[]).forEach(input => {
        commands.push(
          new SfdxCommandBuilder()
            .withArg('force:source:deploy')
            .withFlag('--sourcepath', input)
            .build()
            .toString()
        );
      });
      const hints = commands.join('\n  ');

      return hints;
    }
  });

  messagesByLogName.set('force_source_deploy_with_manifest_beta', {
    warningMessageKey,
    commandHint: input => {
        .withArg('force:source:deploy')
        .withFlag('--manifest', input as string)
        .build()
        .toString();
    }
  });

  const conflictMessages = messagesByLogName.get(logName);
  if (!conflictMessages) {
    throw new Error(`No conflict messages found for ${logName}`);
  }
  return conflictMessages;
}
```
If I'm missing something here, let's discuss f2f.

packages/salesforcedx-vscode-core/src/conflict/messages.ts
```diff
@@ -0,0 +1,45 @@
+import { SfdxCommandBuilder } from '@salesforce/salesforcedx-utils-vscode';
```
jeffb-sfdc, line 1: Please add a copyright here

packages/salesforcedx-vscode-core/src/conflict/metadataCacheService.ts
```diff
@@ -4,7 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { projectPaths } from '@salesforce/salesforcedx-utils-vscode';
+import {
+  ConfigUtil,
```
jeffb-sfdc, line 8: ConfigUtil is not needed

packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -27,17 +30,23 @@ export class TimestampConflictDetector {
...
-    this.determineConflicts(components);
+    this.determineConflicts(components, skipTimestampCheck);
     return this.diffs;
   }
 
-  private determineConflicts(components: CorrelatedComponent[]) {
+  private determineConflicts(
+    components: CorrelatedComponent[],
+    skipTimestampCheck?: boolean
+  ) {
```
jeffb-sfdc, line 49: The return type of `: void` is missing (I forget this myself a lot)

packages/salesforcedx-vscode-core/test/jest/commands/deployExecutor.test.ts
jeffb-sfdc: +1 and I'll probably split it up after ken's PRs are in.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
klewis-sfdc: As per our convo in 11sies mtg today, no further changes needed here.  Resolving conversation.

packages/salesforcedx-vscode-core/test/jest/conflict/metadataCacheService.test.ts
klewis-sfdc: I refactored the test - let me know what you think!

packages/salesforcedx-utils-vscode/package.json
```diff
@@ -11,7 +11,7 @@
   ],
   "main": "out/src",
   "dependencies": {
-    "@salesforce/source-tracking": "2.2.24",
+    "@salesforce/source-tracking": "3.1.1",
```
klewis-sfdc, line 14: Bumped to the latest STL to get https://github.com/forcedotcom/source-tracking/pull/368

packages/salesforcedx-vscode-core/test/vscode-integration/commands/util/postconditionCheckers.test.ts
klewis-sfdc: I don't really want to add to the integration test.  I added just enough here to get it to pass.  When this integ test is converted to a unit test, that scenario could be added there.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -230,6 +283,16 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
...
 
+  /**
+   * @param error The DeployRetrieveExecutor base class catches
+   * SourceConflictErrors that are thrown by the source-tracking
+   * library and calls this method. Conflict handling is not
+   * currently implemented for retrieve operations.
+   * Per the docs, it is suggested to run SFDX: Diff* commands
+   * to check for conflicts before retrieving.
+   */
+  protected async handleSourceConflictError(error: any) {}
```
jeffb-sfdc, line 294: the return type `Promise<void>` is missing

jeffb-sfdc (approved):

jeffb-sfdc:
I tested:
* In Ken's source-tracked sandbox
- View all changes (local and in default org)
- View changes in default org
- View local changes
- Make a local change to a file
- Deploy source
- Make a remote (in server) change to a file
- Retrieve
- View all changes (local and in default org)
    ...might be a bug here...see video
- Make a local and remote change to a file
- View all changes (local and in default org)
- Deploy (via r-click in file)
    ...might be a bug here...see video
- Retrieve (via r-click in file)
    ...might be a bug here... same as above - no conflict detection...see video
- View local changes
- View remote changes
    ...might be a bug here...see video
- Made changes to two files
- Deployed a folder (the "classes" folder)
- Deployed force-app with changes (one Apex, one LWC)
- Deployed force-app again
    ...might be a bug here...see video

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -112,8 +112,18 @@ export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T> {
     components: ComponentSet,
     token: vscode.CancellationToken
   ): Promise<DeployResult | undefined> {
+    const projectPath = getRootWorkspacePath();
+    const connection = await WorkspaceContext.getInstance().getConnection();
+
+    components.projectDirectory = projectPath;
```
klewis-sfdc, line 118: This change was made to potentially fix the issue where a conflict wasn't being detected during Jeff's initial QA. 

It is also a potential fix for #4775.

And it is a property that should be set on the ComponentSet before deploying/retrieving anyway - see https://github.com/forcedotcom/source-deploy-retrieve/pull/929

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -33,7 +33,7 @@ export class SourceTrackingService {
       project,
       ignoreLocalCache: true,
       subscribeSDREvents: true,
-      ignoreConflicts: false
+      ignoreConflicts: true
```
klewis-sfdc, line 36: VSCE will not use STL for conflict detection.  Instead it will continue to use the logic behind the "Detect Conflicts at Sync" setting with uses the TimestampConflictDetector.

Until there is a way to deploy and override any conflicts in STL (when STL was already instantiated in the running process with `ignoreConflicts` set to `true`, we will need to continue to check for conflicts _before_ running the deploy operation (as opposed to catching the STL exception thrown mid-deploy operation)

packages/salesforcedx-sobjects-faux-generator/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "main": "out/src",
   "dependencies": {
-    "@salesforce/core": "3.34.7",
+    "@salesforce/core": "^3.34.8",
```
klewis-sfdc, line 13: In all packages we've had to update core to "^3.34.8" so that the [same version that is installed by @salesforce/source-deploy-retrieve](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/package.json#L28) is also installed for all of the VSCEs.

klewis-sfdc:
Superseded by #4824

PR #4600, opened 2022-12-16
chore: update .sfdx ref in Org Browser folder

RitamAgrawal:
### What does this PR do?
Update all '.sfdx' ref in salesforcedx-vscode-core package with a function call from salesforcedx-utils-vscode

@W-12068891@ #4551 

### Functionality Before
Creating folder path was hardcoded with string values.

### Functionality After
Creating folder path is dynamic with projectPaths. projectPaths.metadataFolder() is used to create metadata directory path.


packages/salesforcedx-vscode-core/src/orgBrowser/metadataType.ts
```diff
@@ -39,21 +40,13 @@ export class TypeUtils {
     'Scontrol'
   ]);
 
-  public async getTypesFolder(usernameOrAlias: string): Promise<string> {
+  public async getTypesFolder(): Promise<string> {
```
gbockus-sf, line 43: did you want to add a jest unit test for this method?

gbockus-sf (approved):
Changes look good 👍 
QE: 
- Verified metadata path in metadataType via inspection ✅ 
- If I log in as a different user the org browser needs a refresh (known issue still an issue) ✅ 
  - Note this is only for unexpanded tree items in the org browser. 
You could add one more unit test but it's fine to get it in before the break

Issue #4585, opened 2022-12-12
"Detect Conflict At Sync" doesn't work (still)


ralphcallaway:
### Summary

Conflict detection is firing on every save. I've made a video showing an example of the issue. https://www.youtube.com/watch?v=NMbWeDj2BLQ

Let me know if I can help at all in documenting test steps.

### Steps To Reproduce:

1.  Open SomeClass.cls
2. Run the "SFDX: Retrieve this Source from Org" command
3. Make a change to SomeClass.cls and save
5. Make a second change and save

### Expected result

Both saves should go through since the version of the class prior to the change is up to date.

### Actual result

The first change after doing a retrieve goes through without throwing a conflict error. The second change throws a conflict error.

### Additional information

**VS Code Version**: 1.74.0

**Salesforce Extension Version in VS Code**: Salesforce Extension Pack v56.5.1

**SFDX CLI Version**: sfdx-cli/7.178.0 darwin-x64 node-v18.12.0

**OS and version**: OS X 12.6.1

### Problem Outline

The cache is storing the last modified date for a component at the start of the save instead of using the final remote last modified date of the component from the server. This leads to the local timestamp to always be earlier than the remote and subsequent saves always appear to have a conflict.

For example, imagine the following sequence of events
1. VS Code: User Saves File
2. VS Code: stores last modified date of file (e.g. 1:00:23) (or saves the file if using the file system last modified date)
3. VS Code: makes a deploy (or other) call to sfdc server
4. Salesforce: processes and saves the file
5. Salesforce: updates last modified date of file (e.g. 1:00:24)

Because their are two timestamps stored marked at different points in the transaction they are almost always slightly out of sync preventing conflict detection from working effectively.

### Solution

After any push/deploy operation assuming either no conflict detection or user chose to proceed after conflict, update the cache of the remote last modified date upon **completion** of the transaction **after** the sfdx call has completed. This should happen regardless of whether the operation succeeded. 

After any successful retrieve/pull operation, update the cache/store of the remote last modified date of the component **after** the sfdx call has completed.

### History

This bug has been logged and closed repeatedly, let's get it fixed this time
- #3542
- #3522
- #3414
- #3438

Other Related Stuff
- #3566
- review discussion following [this comment on #2034](https://github.com/forcedotcom/salesforcedx-vscode/issues/2034#issuecomment-812648816)

ralphcallaway:
[...]

PR #4616, opened 2023-01-04
fix: Fix Conflict Detection at Sync

klewis-sfdc:
### What does this PR do?
Sets the last modified date for the component in the local cache after deployment and allows more recent local changes to be deployed.

#4585, @W-12302680@

### Functionality Before
Locally made changes would trip the conflict detector.

### Functionality After
Locally made changes will be deployed, unless more recent changes have been made on the server, which will trip the conflict detector.

**Todos:**

- [x] Unit tests
- [x] QA
- [x] Fix failing integration tests


packages/salesforcedx-vscode-core/src/conflict/timestampConflictDetector.ts
```diff
@@ -51,8 +53,11 @@ export class TimestampConflictDetector {
         component.cacheComponent.fullName
       );
       lastModifiedInCache = cache.getPropertiesForFile(key)?.lastModifiedDate;
-      if (!lastModifiedInCache || lastModifiedInOrg !== lastModifiedInCache) {
-        const differences = diffComponents(component.projectComponent, component.cacheComponent);
+      if (!lastModifiedInCache || lastModifiedInOrg > lastModifiedInCache) {
```
gbockus-sf, line 56: This date string comparaison still kind of blows my mind.  According to this [SO](https://stackoverflow.com/questions/14781153/how-to-compare-two-string-dates-in-javascript) this method will work for format YYYY-MM-DD, but it appears to be working for our format as well (at least for the dates we've tested).  I was kind of surprised (but not really) that we don't have a util for working with dates. To avoid blowing up this PR how would you feel about making a private method here to actually compare the dates ala ```
private dateIsGreater(dateStrOne: string, dateStrTwo: string): boolean {
  const dateNumOne = new Date(dateStrOne).getTime();
  const dateNumTwo = new Date(dateStringTwo).getTime();
  return dateNumOne > dateNumTwo;
}
```
working with dates always makes me nervous b/c it's always more complicated then it seems like it should be. 
In the future we should def pull in a module targeting date handling and have a service for handing it.

File: packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts [hunk 2]
```diff
@@ -63,6 +66,15 @@ export class PersistentStorageService {
     }
   }
 
+  public setPropertiesForFilesDeploy(result: DeployResult) {
```
gbockus-sf, line 69: mostly a nit, but we pass a DeployResponse here and a FileProperties in the set for the retrieve.  Seems like it's the same data, just one layer deeper on the retrieve method.  Did you consider if we could just make `setPropertiesForFilesRetrieve` -> `setPropertiesForFiles` and reuse that same logic?

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/timestampConflictDetector.test.ts
```diff
@@ -24,6 +24,8 @@ import { nls } from '../../../src/messages';
 import { stubRootWorkspace } from '../util/rootWorkspace.test-util';
 
 describe('Timestamp Conflict Detector Execution', () => {
+  const TODAY = '2023-01-28T00:15:28.000Z';
+  const YESTERDAY = '2023-01-27T00:15:28.000Z';
```
gbockus-sf, line 28: Would it be worth expanding this to test some other cases beside just yesterday and today?

gbockus-sf (commented):
A few comments for your consideration.  Thanks for digging into this problem!

packages/salesforcedx-vscode-core/test/vscode-integration/conflict/timestampConflictDetector.test.ts
gbockus-sf: actually on 2nd thought let's def not expand the int tests.  If you move to the date comparaison I'd feel less worried about this validation.

packages/salesforcedx-vscode-core/src/conflict/persistentStorageService.ts
klewis-sfdc: The data is coming from a different place though - the fileProperty for retreive (fileProperty.lastModifiedDate) and from the response itself for the deploy (result.response.lastModifiedDate).  I kinda think these should stay the way they are to keep it clear that the internal logic is actually slightly different for each operation.

gbockus-sf: word that makes sense.  I was assuming a DeployResponse and a RetrieveResponse would have the same shape but I guess not.

gbockus-sf (approved):
QE & Review complete. :shipit: 

* Attempt a single file deploy 
    * brand new apex class
    * without any changes
    * local files changed only
        * deployed no conflicts
        * Immediate deploy unchanged after
            * deployed no conflicts
    * remote file changed cancel flow
        * detected as conflict and showed changes in both cls and meta
        * Canceled and compared worked as expected (no change in org). 
        * [Image: Image.jpg]
    * remote file changed overwrite flow
        * overwrote remote file
    * both changed
        * overwrite
            * overwrote the remote file. 
        * view conflicts
            * Looks good 💯. Can see the diff in both the remote and local versions. 
* Attempt a deploy of a folder
    * without any changes
    * local files changed only
    * remote file changed 
        * view 
        * overwrite 
    * both changed
        * view 
        * overwrite 
* Do a deploy only altering the api version of the file
    * unchanged
    * local files changed only
    * remote file changed 
        * One odd use case here. If I change the API version in the remote, then view the diff I see the api version has changed.  However if I use the arrows in vscode to update the local file then attempt to deploy I get a conflict that is only an empty line in the remote file.
            * [Image: Image.jpg]
            * If I use the arrow to accept the change and save locally it still comes up as a conflict when I attempt to deploy. 
            * I can retrieve then deploy without issue. 
    * both changed

PR #4622, opened 2023-01-09
fix: add check for no open workspace folders

gbockus-sf:
### What does this PR do?
Add a check to ensure we have workspace folders before attempting to refresh sobjects during core extension activation.  This addresses the state where the core extension is activated in order to create a new project when VSCode currently doesn't have any open folders. 

@W-12347182@

### Functionality Before
Bundled extension would fail to create a project with a zero index error. 

### Functionality After
New project creation works as expected form the empty folder state.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -707,15 +707,15 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
   telemetryService.sendExtensionActivationEvent(extensionHRStart);
   console.log('SFDX CLI Extension Activated');
 
-  // Refresh SObject definitions if there aren't any faux classes
-  const sobjectRefreshStartup: boolean = vscode.workspace
-    .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
-    .get<boolean>(ENABLE_SOBJECT_REFRESH_ON_STARTUP, false);
-
   if (
     vscode.workspace.workspaceFolders &&
     vscode.workspace.workspaceFolders.length > 0
   ) {
+    // Refresh SObject definitions if there aren't any faux classes
+    const sobjectRefreshStartup: boolean = vscode.workspace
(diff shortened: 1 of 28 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -712,14 +712,19 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
...
-    initSObjectDefinitions(
-      vscode.workspace.workspaceFolders![0].uri.fsPath
-    ).catch(e => telemetryService.sendException(e.name, e.message));
-  } else {
-    checkSObjectsAndRefresh(
-      vscode.workspace.workspaceFolders![0].uri.fsPath
-    ).catch(e => telemetryService.sendException(e.name, e.message));
+  if (
+    vscode.workspace.workspaceFolders &&
+    vscode.workspace.workspaceFolders.length > 0
```
gbockus-sf, line 717: So if there are no folders open in vscode `workspaceFolders` is undefined and this blows up.  
Jest testing this is going to be interesting but I'll give it a shot.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -712,14 +712,19 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
...
+  ) {
+    if (sobjectRefreshStartup) {
```
klewis-sfdc, line 719: What do you think about moving the assignment of the `sobjectRefreshStartup` var inside of this new if(vscode.workspace...) block?

klewis-sfdc (approved):
One question about slightly moving the sobjectRefreshStartup const, and I know you mention trying to unit test the change...

Code change looks good and QA checks out:
* Before change, I get "Activating extension 'salesforce.salesforcedx-vscode-core' failed" when opening VS Code without a project loaded and running "SFDX: Create Project".
* After change, I do not see the error and I can create the project successfully. ✅

PR #4623, opened 2023-01-10
ci: update publish job to checkout code correctly

gbockus-sf:
### What does this PR do?
- Update the publishVSCode GHA to always checkout the code prior to interacting with GH
- Use npx with vsce so it can be found during publish. 
- 
@W-11596486@

### Functionality Before
Failed due to not checking out the code when generating a sha for publish

### Functionality After
checks out code prior to sha generation.

### DIFF
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -60,6 +60,7 @@ jobs:
           path: ./extensions
       - name: Display downloaded vsix files
         run: ls -R ./extensions
+      - run: npm ci
       - run: find ./extensions -type f -name "*.vsix" -exec npx vsce publish --pat ${{ env.VSCE_PERSONAL_ACCESS_TOKEN }} --packagePath {} \;
       - id: vsce
         run: | 
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -18,26 +18,22 @@ on:
 # The steps.publsh.outputs.result is an indicator that the pulish to the vscode marketplace succeeded.
 # The steps.mergeDevelop.outputs.status is an indicator that all the publish steps completed.
 jobs: 
-  checkout-code: 
-    runs-on: ubuntu-latest
-    steps:
-    - name: Setup Node.js
-      uses: actions/setup-node@v3
-      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publishVSCode.yml
```diff
@@ -18,52 +18,49 @@ on:
 # The steps.publsh.outputs.result is an indicator that the pulish to the vscode marketplace succeeded.
 # The steps.mergeDevelop.outputs.status is an indicator that all the publish steps completed.
 jobs: 
-  checkout-code:
```
gbockus-sf, line 21: So I dunno why I can't grok this, but each job run on separate VM.  So for each job you have to do a full checkout if you need access to the codebase.  I updated this flow to bring together the pieces required to do the sha update and push into main.

.github/workflows/publishVSCode.yml
```diff
@@ -72,27 +69,53 @@ jobs:
...
       - run: |
-          git tag v${{ inputs.version }}
-          git push origin v${{ inputs.version }}
-  pushMain: 
-    needs: [publish]
-    runs-on: ubuntu-latest
-    steps: 
-      - run: git push origin main
+            git tag v${{ inputs.version }}
+            git push origin v${{ inputs.version }}
```
gbockus-sf, line 82: to tag we'll need to have the code checked out.

.github/workflows/publishVSCode.yml
```diff
@@ -72,27 +69,53 @@ jobs:
...
+      - uses: ./.github/actions/gitConfig
+        with:
+          email: ${{ secrets.IDEE_GH_EMAIL }}
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.version }}
+          path: ./extensions
+      - name: Display downloaded vsix files
+        run: ls -R ./extensions
       - run: gh release create v${{ inputs.version }} ./extensions/*.vsix --title "Release v${{ inputs.version }}" --notes-file ./packages/salesforcedx-vscode/CHANGELOG.md
```
gbockus-sf, line 101: got create release we'll need to have vsixes and code for changelog

klewis-sfdc (approved):

PR #4625, opened 2023-01-11
ci: Fix how was generate the SHAs for validating published extensions

gbockus-sf:
### What does this PR do?
Update the script and related npm script for generating the SHAs during publishing.

Note this PR is targeting main b/c it needs to be in main in order to get executed for during the deploy.

@W-11596486@

### Functionality Before
Previous way SHAs were generated depended on the vsix being in the each package directory.  With GHA we have them all in a single top level directory. 

### Functionality After
sha report correctly built off the single directory of vsix files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "vscode:bundle:extension": "lerna run bundle:extension",
     "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
     "vsix:install": "find ./packages -name '*.vsix' -exec code --install-extension {} \\;",
-    "vscode:sha256": "lerna run vscode:sha256 --concurrency 1",
+    "vscode:sha256": "node ./scripts/generate-sha256.js >> ./SHA256",
     "vscode:publish": "lerna run vscode:publish --concurrency 1",
     "watch": "lerna run --parallel watch",
     "eslint-check": "eslint --print-config .eslintrc.json | eslint-config-prettier-check",
--- a/scripts/generate-sha256.js
+++ b/scripts/generate-sha256.js
@@ -1,19 +1,24 @@
 #!/usr/bin/env node
-
+const path = require('path');
 const shell = require('shelljs');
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "vscode:bundle:extension": "lerna run bundle:extension",
     "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
     "vsix:install": "find ./packages -name '*.vsix' -exec code --install-extension {} \\;",
-    "vscode:sha256": "lerna run vscode:sha256 --concurrency 1",
+    "vscode:sha256": "node ./scripts/generate-sha256.js >> ./SHA256",
```
gbockus-sf, line 88: So previously this would run the vscode:sha256 script in all packages.  Since for GHA we have all the vsix files gathered in a single directory this didn't make sense.

scripts/generate-sha256.js
```diff
@@ -1,19 +1,24 @@
 #!/usr/bin/env node
-
+const path = require('path');
 const shell = require('shelljs');
```
gbockus-sf, line 4: The output of this script is redirected in in the shell to a file.  I considered changing that behavior to make it less mysterious when things go wrong but decided I'll create a follow up ticket.

scripts/generate-sha256.js
```diff
@@ -1,19 +1,24 @@
...
+    'No VSIX files found matching the requested version in package.json'
+  );
   shell.exit(1);
 }
 
-if (/win32/.test(process.platform)) {
-  shell.exec(`CertUtil -hashfile ${vsix} SHA256`);
-} else {
-  shell.exec(`shasum -a 256 ${vsix}`);
+process.chdir('extensions');
```
gbockus-sf, line 16: The genreated SHA256 file should have the format
```
6da4adbf213c5deeadf5029cb37691d486649c192f96a8b3b878ddfc6485d37f  salesforcedx-vscode-apex-56.5.1.vsix
```
by changing to the extensions dir we are able to keep this format the same.

klewis-sfdc (approved):

PR #4643, opened 2023-01-24
feat: Switch to Project Path at Activation for Better Integration with Shared Libraries

klewis-sfdc:
### What does this PR do?
* Sets the current working directory to the projectPath at activation time for better integration with VSCE's's core shared libraries.

@W-12424555@

### Functionality Before
VSCE processes run in the default dir ('/') on a Mac

### Functionality After
VSCE processes run in the project directory (returning project path from process.cwd()) after activation.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -1,11 +1,15 @@
+import { getRootWorkspacePath } from '@salesforce/salesforcedx-utils-vscode';
 /*
  * Copyright (c) 2017, salesforce.com, inc.
  * All rights reserved.
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { getRootWorkspacePath } from '@salesforce/salesforcedx-utils-vscode';
-import { SFDX_CORE_CONFIGURATION_NAME } from '@salesforce/salesforcedx-utils-vscode';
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -595,32 +598,19 @@ async function setupOrgBrowser(
...
+  // node libraries work correctly.  @salesforce/core,
+  // @salesforce/source-tracking, etc. all use process.cwd()
+  // internally.  This causes issues when used from VSCE, as VSCE
+  // processes can run with a path that does not reflect the current
+  // project path (it often returns '/' from process.cwd()).
+  // Switching to the project path here at activation time ensures that
+  // commands are run with the project path returned from process.cwd(),
+  // thus avoiding the potential errors surfaced when the libs call
+  // process.cwd().
+  ensureCurrentWorkingDirIsProjectPath(rootWorkspacePath);
```
klewis-sfdc, line 613: I decided not to wrap this in a try/catch since downstream functions will all be depending on cwd() returning the projectPath from this point forward.  If something goes wrong with switching to the projectPath, then I think activate should throw here.

gbockus-sf (commented):
Looks awesome. 🚢 I'll QE now

gbockus-sf (approved):
Passed my QE.  Mostly just sanity checking the basic flows:
- Deploy works
- Create Project works
- Authorize works
- Retrieve works
- Org Browser works
- Verified that when I run a command it has the expected cwd of the project path.

PR #4651, opened 2023-01-27
ci: add manual publish option

gbockus-sf:
### What does this PR do?
- Add a button to the release job for publishing from main.
- Update the configuration for firing off the publish job to only happen for actual releases and not pre-releases instead of filtering in the workflow. 

@W-12370302@

### DIFF
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -1,15 +1,17 @@
 name: Publish in Microsoft Marketplace
 on:
   release:
-    types: [published, edited]
+    # This limits the workflow to releases that are not pre-releases 
+    # From the docs: A release was published, or a pre-release was changed to a release.
+    types: [ released ]
+  #  Button for publishing main branch in case there is a failure on the release.
+  workflow_call:
     
 jobs:
 
   publish-start-notification:
     name: "Publish Start Notification"
     uses: ./.github/workflows/slackNotification.yml
-    # Only run this job is not a prerelease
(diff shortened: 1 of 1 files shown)


.github/workflows/publishVSCode.yml
```diff
@@ -1,26 +1,27 @@
 name: Publish in Microsoft Marketplace
 on:
   release:
-    types: [published, edited]
+    # This limits the workflow to releases that are not pre-releases 
+    # From the docs: A release was published, or a pre-release was changed to a release.
+    types: [ released ]
```
gbockus-sf, line 6: doc says it but Randi found info to make this just run for real releases so I could remove the pre-release filtering.  Also removed the edited that didn't ever fire the workflow as expected anyway.

.github/workflows/publishVSCode.yml
```diff
@@ -1,26 +1,27 @@
...
-    if: ${{ !github.event.release.prerelease }}
     secrets: inherit
     with:
       title: "Publish waiting for approval in github."
       notification: "New Release triggered publish build and job is ready for approval."
       type: "notification"
       workflow: "publishVSCode.yml"
   
-  prepare_environment:
+  prepare-environment-from-main:
```
gbockus-sf, line 22: renamed to be clear this was happening on main.

CristiCanizales (approved):
LGTM 😁 🐑

randi274 (approved):
+1 to @CristiCanizales . 🛳️ it!

PR #4656, opened 2023-02-01
chore: Update tsconfig and cleanup

gbockus-sf:
### What does this PR do?
- Update the tsconfig to use the base tsconfig for node v16
- remove the out import paths where I could.  There are a few cases left in our repo around int tests and the apex-debugger that will need stand alone attention. 
- remove istanbul dependancy from int tests as it has been out of support for years and doesn't play nice with the new tsconfig version. 

@W-11272537@

### Functionality After
- Only noticeable change is we no longer have code coverage for int tests.  With those tests being on the way out it seemed like the lesser evil to remove the istanbul dep used for generating coverage data on those tests. 
- Not noticeable change is to update the generated js files to node14 which is the base for the versions of VSCode we support.  The lib which provides the typescript language feedback is set to node 16 which is the current version used for development. 

### Notes
Removing istanbul takes our number of deprecated modules in the repo down from 39 - 28. 🥳 
I will create a new WI to track updating the apex debugger and test specific imports that use the out path.  
The apex debugger is an odd module in how it is configured for top level exports so it was going to be a bigger lift to refactor it to remove the out imports.
There is also a single module in the core package unit test dir that is imported into some other packages via the out path.  This should be moved and exposed in the test specific package. 
The imports that are left in the project using the out path can be found in the attached image

Screenshot 2023-02-01 at 3 07 42 PM


package.json
```diff
@@ -32,14 +32,12 @@
     "eslint": "6.0.1",
     "eslint-config-prettier": "6.15.0",
     "husky": "1.3.1",
-    "istanbul": "1.1.0-alpha.1",
```
gbockus-sf, line 35: istanbul has been deprecated for a loooong time.  It was causing issues with compiling the source to node 14 `es2020`.  This is only used for the unigration tests.  I attempted to keep coverage for a bit but that test running is bananas so I ended up just punting and removing the coverage piece of the puzzle since those tests are headed for deletion anyway.

packages/salesforcedx-test-utils-vscode/src/testrunner.ts
```diff
@@ -1,19 +1,11 @@
-/*
- * rewritten test runner for integrating code coverage;
- * https://github.com/codecov/example-typescript-vscode-extension
```
gbockus-sf, line 3: This was a dead link so I just removed the comment.

packages/salesforcedx-test-utils-vscode/src/testrunner.ts
```diff
@@ -1,19 +1,11 @@
...
-import * as fs from 'fs';
 import * as glob from 'glob';
 import * as paths from 'path';
-import { isNullOrUndefined } from 'util';
 
 // tslint:disable:no-var-requires
-const istanbul = require('istanbul');
 // tslint:disable-next-line:variable-name
 const Mocha = require('mocha');
-const remapIstanbul = require('remap-istanbul');
```
gbockus-sf, line 16: So this is the custom test runner used for integration tests.   It is....complicated and more recent versions of mocha looks like it's not overly common to use this pattern anymore.  I just removed all the pieces related to code coverage.   

Note the replacement for istanbul (nyc) doesn't event have the same code based options and is only a command line client.  I attempted to pull it in outside of the int test run and it worked for me locally, but failed in GHA.

packages/salesforcedx-test-utils-vscode/src/testrunner.ts
```diff
@@ -156,9 +112,9 @@ function run(testsRoot: any, clb: any): any {
           .on('end', (): void => {
             console.log(`Tests ended with ${failureCount} failure(s)`);
             clb(undefined, failureCount);
-            if (!isNullOrUndefined(coverageRunner)) {
-              coverageRunner.reportCoverage();
-            }
+            // if (!isNullOrUndefined(coverageRunner)) {
+            //   coverageRunner.reportCoverage();
+            // }
```
gbockus-sf, line 117: will delete

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
```diff
@@ -4,36 +4,27 @@
...
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
 import {
+  SObjectCategory,
+  SObjectRefreshSource,
   SOBJECTS_DIR,
+  SObjectTransformerFactory,
   STANDARDOBJECTS_DIR
-} from '@salesforce/salesforcedx-sobjects-faux-generator/out/src';
```
gbockus-sf, line 11: so missed out paths removed here and organized the imports.

jeffb-sfdc (approved):

RitamAgrawal (approved):
QE Approved:
✅ Compiling successfully
✅ All Tests are passing
✅ Refresh SObjects -> All SObjects running successfully on DE as well as Scratch Org

PR #4661, opened 2023-02-03
fix: Update Core's Config Aggregator After Writing to Config File

klewis-sfdc:
### What does this PR do?

The extensions have seen some cacheing issues when using the types from the @salesforce/core library.  

This seems to be due to the differences in how the cli and VS Code use node processes.  

The cli is often running in one-off processes, while VS Code uses long-running processes.  

Because of this difference, we are seeing that some manual steps need to be taken to ensure that the core types are updated with the latest info after updating stored data (like configuration values) in VS Code.

* Forces the ConfigAggregator to read the Config file after it is updated with new information.

@W-9466545@

### Functionality Before
* The ConfigAggregator would not always be up-to-date in subsequent calls after calling `configUtil.setUsernameOrAlias`

### Functionality After
* The ConfigAggregator is always up-to-date in subsequent calls after calling `configUtil.setUsernameOrAlias`


RitamAgrawal (commented):
Tested with VSIX and the error is thrown when deploying the source to Org with Detect conflicts setting ON.

RitamAgrawal (approved):
QE (approved):
✅ SFDX: Create an Apex Class
✅ SFDX: Deploy Source to Org with Detect Conflict At Sync setting on.
✅ SFDX: Create a scratch Org
✅ SFDX: Push Source to Org

PR #4665, opened 2023-02-07
feat: Enable creating UTAM config files and running tests

maliroteh-sf:
### What does this PR do?

This PR adds a new feature to the SFDX extension for creating UI Test Automation Model (UTAM) config files and running tests using UTAM. When the user right-clicks on a test/spec `js` file under `lwc` or `modules` folders, the presented context menu now includes a new option to run the test using UTAM:

Screenshot 2023-02-07 at 6 54 16 AM

The user would need to open an SFDX project in VSCode and that project should already be set up for UTAM as documented [here](https://utam.dev/guide/get_started_utam) (with examples available [here](https://github.com/salesforce/utam-js-recipes)). It is also the responsibility of the user to have installed all the tools necessary for running tests using UTAM (e.g Appium, latest `@salesforce/lwc-dev-mobile` SFDX plugin, ...)

The PR also includes:
- Refactoring some of the code to promote code sharing/reuse across a few commands
- Adding more test coverage

@W-12263582@

### Functionality Before
No mechanism to automatically create a UTAM config file and run tests using UTAM.

### Functionality After
Ability to automatically create a UTAM config file and run tests using UTAM.


packages/salesforcedx-vscode-lwc/src/commands/forceLightningLwcPreview.ts
```diff
@@ -27,119 +23,33 @@ import { PreviewService } from '../service/previewService';
 import { telemetryService } from '../telemetry';
 import { openBrowser, showError } from './commandUtils';
 import { ForceLightningLwcStartExecutor } from './forceLightningLwcStart';
-
-enum PreviewPlatformType {
```
maliroteh-sf, line 31: Many of the deleted codes in this file are actually refactored/moved to the new `lwcUtils.ts`

packages/salesforcedx-vscode-lwc/test/vscode-integration/commands/forceLightningLwcPreview.test.ts
```diff
@@ -30,125 +24,23 @@ import {
...
+} from '../../../src/commands/lwcUtils';
 import { nls } from '../../../src/messages';
 import { DevServerService } from '../../../src/service/devServerService';
 import { WorkspaceUtils } from '../../../src/util/workspaceUtils';
-import {
-  ChannelService,
-  notificationService
-} from '@salesforce/salesforcedx-utils-vscode';
-
-const sfdxDeviceListCommand = 'force:lightning:local:device:list';
```
maliroteh-sf, line 48: Again, many of the deleted codes in this file are actually refactored/moved to the new `lwcUtils.test.ts`

sfdctaka (approved):
LGTM!

packages/salesforcedx-vscode-lwc/src/messages/i18n.ja.ts
```diff
@@ -37,11 +37,11 @@ export const messages = {
   force_lightning_lwc_stop_in_progress: 'ローカルの開発サーバを停止しています',
   force_lightning_lwc_preview_text:
     'SFDX: コンポーネントをローカルでプレビュー',
-  force_lightning_lwc_preview_file_undefined:
```
maliroteh-sf, line 40: Renaming these since they're now shared across commands and are not exclusive to the Preview command anymore.

packages/salesforcedx-vscode-lwc/test/jest/commands/forceLightningLwcPreview.test.ts
```diff
@@ -0,0 +1,336 @@
+/*
```
maliroteh-sf, line 1: converted from integration to jest test.

packages/salesforcedx-vscode-lwc/src/commands/lwcUtils.ts
```diff
@@ -0,0 +1,625 @@
...
+   */
+  public static isMacOS(): boolean {
+    return process.platform === 'darwin';
+  }
+
+  /**
+   * Takes a file path and converts it to a UNIX file path.
+   *
+   * @param filePath A file path.
+   * @return A string containg the file path in UNIX path style.
```
gbockus-sf, line 126: nit: containg -> containing

packages/salesforcedx-vscode-lwc/src/commands/lwcUtils.ts
```diff
@@ -0,0 +1,625 @@
...
+  CommandOutput,
+  SfdxCommandBuilder
+} from '@salesforce/salesforcedx-utils-vscode';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { channelService } from '../channel';
+import { nls } from '../messages';
+import { telemetryService } from '../telemetry';
+import { showError } from './commandUtils';
```
gbockus-sf, line 20: What was your thinking when choosing lwcUtils as the name for this module? It seems like kind of a catch all (which is def better then duplicated code btw :) ) although not specifically lwc related which is my primary concern. How would you feel about splitting it up further into areas of concern?  We're moving away from giant files with a bunch as exports as it leads to problems unit testing and requires more context knowledge to find available modules. 

I would see something like 
```
error/operationCancelledException.ts
enum/eachEnum
types/eachInterface
```
for the platform specific constants those can go in the constants.ts file or in a platormConstants.ts file

and for the rest we have a service directory so it could be renamed to LwcMobileCommandService and moved over there? 

I'm mostly looking to avoid an LWCUtils module that will turn into the catch all for any function needed for commands. thoughts?

maliroteh-sf: @gbockus-sf I agree... however, as per the team discussion we had, I thought there's going to be a move towards having contributing teams taking out their code from this repo and instead create their own plugins for the vscode extension. So I thought to keep everything in this 1 file since it is currently only used by `forceLightningLwcTestUIMobileRun` and `forceLightningLwcPreview` which MPE owns. I'm afraid if I break it up then others might use bits & pieces of this code and when time comes to pull these out of this repo then the clean up would be harder. But if you still want me to break up this file then I can do that.

gbockus-sf: oh yeah good point.  How would you feel about moving things MPE owns into a separate folder? It could have subfolders that follow the usual pattern of commands etc?  I totally understand we're sort of figuring out the best practices as we go and can adjust with more WI if that makes life easier.

maliroteh-sf: Sure yes I can do that as part of another WI. After this PR goes in, I'll create a follow up PR to split `lwcUtils` and move everything that MPE own into another folder.

maliroteh-sf:
In light the feedback MPE team received offline, we're going to close this PR and create another PR with a different approach to launching UTAM tests.

PR #4672, opened 2023-02-08
Fixed bug in getTestClassName(), added try/catch to forceLaunchApexReplayDebuggerWithCurrentFile()

jeffb-sfdc:
### What does this PR do?
This PR address a few issues that were recently discovered with the `SFDX: Launch Apex Replay Debugger with Current File` command.

#@W-12508503@

### Functionality Before
We discovered two issue:
First, when the user ran the `SFDX: Launch Apex Replay Debugger with Current File` command, it didn't work.  Meaning, to the user the Apex debugger never started.

Second, even though the operation failed, the user was never told anything, are weren't told it had failed.

### Functionality After
* A try/catch block was added to forceLaunchApexReplayDebuggerWithCurrentFile().  If an exception occurs, the exception is caught and we now display a notification to the user and log the exception into the Output panel.
* When `getApexTestClassName()` is called, we now no longer call `flushFilePath()`.  Instead, `getTestClassName()` converts the file path string and the paths stored in the Map (`testIndex`) to lower case and compares the two strings.

### Notes for reviewers and QE
* When running `SFDX: Launch Apex Replay Debugger with Current File`, this was only an issue with debugging Apex class files (not debugging anonymous Apex files, or Apex log files), though it wold be good to have all three paths/files types verified.
* To repro the handling of the exception and displaying the notification & output, this will be hard to naturally reproduce.  I suggest running code from this branch and modify `forceLaunchApexReplayDebuggerWithCurrentFile()` to throw an exception, and then verify the results in the UI.
* This should be QA'd on windows as well.
* Unit tests will be coming in a separate PR.  I'm going to perform some refactoring of files containing multiple classes in them into separate files, so there will be a few extra files touched in for the unit test work, and adding those changes here in this PR will (IMHO) cloud the changes I made to fix this issue.
* Other places in our code where `flushFilePath()` is called are:

- forceSourceDelete
- LibraryPathsGatherer, which is used in forceSourceDeploySourcePath and forceSourceDeploySourcePath

I verified that all three work, but the QE should also test/verify each of these three features/areas as well on both mac and Windows


packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -10,53 +10,65 @@ import {
...
+import { channelService } from '../channels';
 import { nls } from '../messages';
 import { testOutlineProvider } from '../views/testOutlineProvider';
 import { forceAnonApexDebug } from './forceAnonApexExecute';
 
 export async function forceLaunchApexReplayDebuggerWithCurrentFile() {
-  const editor = vscode.window.activeTextEditor;
-  if (!editor) {
+  try {
+    const editor = vscode.window.activeTextEditor;
```
jeffb-sfdc, line 26: Lines 26-60 in the new version are lines 25-59 in the old version.  No code changes were made here, other than wrapping in a `try` block and indenting.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -10,53 +10,65 @@ import {
...
-
-  if (isLogFile(sourceUri)) {
-    await launchReplayDebuggerLogFile(sourceUri);
-    return;
-  }
 
-  if (isAnonymousApexFile(sourceUri)) {
-    await launchAnonymousApexReplayDebugger();
-    return;
+    channelService.showChannelOutput();
```
jeffb-sfdc, line 66: In the `catch` block, a notification and logging to the Output view were added.  A new string was added and the verbiage was created by Sonal.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -81,13 +93,11 @@ async function getApexTestClassName(
...
-  // but it actually it returns string | undefined.  Well, regardless, since flushFilePath() takes
-  // a string (and guards against empty strings) using the Non-null assertion operator
-  // (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)
-  // fixes the issue.
-  testClassName = fileUtils.flushFilePath(testClassName || '');
+
+  const filePath = sourceUri.toString();
+  // filePath is in the format of "file:///Users/{user-name}/{path-to-apex-file.cls}"
+
+  const testClassName = testOutlineProvider.getTestClassName(filePath);
```
jeffb-sfdc, line 100: We no no long use `getTestClassName()`, and instead, within `getTestClassName() it converts the paths to lower case for the comparison.

I believe Windows isn't able to have multiple files with different casing in the same location (eg `fooBar.txt` an `foobar.txt`)

Linux (I think) does support this and one can have files with different casings t the same time (in the same location), but I thin the chances of someone having both `myapexclass.cls` and `MyApexClass.cls` in the same location are pretty slim.

And with Mac OSX, even though its file system is based in Linux, its UI (the Finder) prevents one from naming files with different casing in the same location.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -24,6 +24,7 @@ export function extractJsonObject(str: string): any {
 // To get around this, fs.realpathSync.native() is called to get the
 // URI with the actual file name.
 export function flushFilePath(filePath: string): string {
+  // filePath is in the format of "/Users/{user-name}/{path-to-apex-file.cls}"
```
gbockus-sf, line 27: I'm not sure this comment is valuable.  This is only true on mac, and only if the sf project is under the users home directory (it could be anywhere on the filesystem) It could be misleading to our future selves reading it.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
gbockus-sf: So what we're talking about here is a case-sensitive file system.  All OS's have the ability to have case sensitivity turned on for the file system (it's more complicated then that for windows in that it's a more recently supported thing and can be enabled on the folder level :cringe: [source](https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity).  It's been a while but last I recall salesforce doesn't enforce case sensitivity in the org, so I think we're clear to assume that there will never be two files with the same name but different cases. 

I did a quick google to see if I could find the official SF doc on this and found a few external hits but nothing official.  General agreement seems to be that apex class name are case insensitive so we're good to go.  I also verified that if I tried to create a new class with the same name but different casing it errored 👍

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -10,53 +10,65 @@ import {
...
-  }
 
-  if (isAnonymousApexFile(sourceUri)) {
-    await launchAnonymousApexReplayDebugger();
-    return;
+    channelService.showChannelOutput();
+    const message = (error && error.message)
+      ? error.message
+      : JSON.stringify(error);
+    channelService.appendLine(message);
```
gbockus-sf, line 70: I know there are existing patterns in place, but this feels dangerous in that JSON.stringify is a good way to get another error thrown that won't be caught (and it'll just be nonsense in the console out if displayed). 

The right thing to do here would be to add a method to our utils package to process errors and give us back a string to display to the user.  Some quick googling led me to [this](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript) which in it's final form is pretty bulletproof, but I get we don't want to blow this up so maybe the solution in this [SO](https://stackoverflow.com/questions/68240884/error-object-inside-catch-is-of-type-unknown) makes more sense with a WI to address error handling across the board.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
gbockus-sf, line 100: So the root of the issue here was flushFilePath throwing an error correct?  I was trying to wrap my head around this change and didn't really grok the root problem.  

One big downside to the change is we are switching from a single lookup in a map to iteration over all tests that exist to see if any match (with an extra operation to lowercase the path for each test).  Granted for small use cases this probably doesn't matter, but we support larger orgs that have hunderds of not thousands of tests so the cost here is potentially high. 

An alternate approach would be to move the flushFilePath call to before the getTestClassName call 
```
  const flushedPath =  fileUtils.flushFilePath(sourceUri);
  const testClassName = testOutlineProvider.getTestClassName(flushedPath);
```

To verify this I wanted to see what was happening with flushFilePath.  It appears there have been issues with the version of node used by VSCode (which explains why we didn't see this originally but it's showing up now) and the usage of `realpathSync.native` [ref](https://github.com/microsoft/vscode-eslint/issues/1229#issuecomment-814057935).

gbockus-sf: We may want to add a try/catch to `flushFilePath`.  I did a quick search and it appears that a pretty typical approach: [example](https://github.dev/facebook/jest/blob/80cf80055682919251b9567ef482755d5f060d47/packages/jest-util/src/tryRealpath.ts#L12%23L12)

Thoughts?

gbockus-sf (changes requested):
See comments.  Biggest concern is perf impact of iterating over all test files in this code flow.  Happy to chat here or in slack 👍

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
randi274: I had the opposite reaction of "oh how useful!" 😄 But I get your point here.

jeffb-sfdc: When I first discovered the bug and I was stepping through the code, it was not obvious what the issue was, which was, when `fs.realpathSync.native` is called, it's sometimes called with a path of `file://Users/...` and sometimes with a. path of `/Users/...`, and when passed a string with "file:" in it, it would thrown an exception.

Knowing what I do now, yea, it's obvious, but when I was initially looking into it, it wasn't obvious.

OK, so why use paths with "file:" in them at all?  We have code (like what's in testOutlineProvider.ts) which relies on it.  I'm not aware of the historical reasons why this is, but would prefer to not change additional components/areas with this PR.

gbockus-sf: That's cool.  In an ideal world we should create types and remove the mental type checking of strings [ref](https://stackoverflow.com/questions/51445767/how-to-define-a-regex-matched-string-type-in-typescript).  In a less ideal world we should validate the filePath parameter matches what we expect and throw an error if not correct. 

 The part I take issue with here is `/Users/{user-name}/`.  It's inaccurate and could lead devs down a bad assumption path about what to expect from the input parameter.

jeffb-sfdc: _"The part I take issue with here is /Users/{user-name}/..."_

I agree.  I still need to run this on Windows an verify the fix works.  (I only have a VM for this, and it's dog slow and I've been swamped this week and haven't gotten to this yet).  I'll post an update here with what I find.  I know the format will be different (eg `/Users/{user-name}/...` vs `C:/Users/{user-name}/...` so I will probably document both expected Mac and Windows paths in both locations.

jeffb-sfdc: Randi ran the branch on her Windows machine this morning it ran w/o any issues (the file paths worked).  I added comments on what the expected path for Mac OSX and Windows are expected.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -24,6 +24,8 @@ export function extractJsonObject(str: string): any {
 // To get around this, fs.realpathSync.native() is called to get the
 // URI with the actual file name.
 export function flushFilePath(filePath: string): string {
+  // filePath is in the format of "/Users/{user-name}/{path-to-apex-file.cls}" on Mac OSX
+  // filePath is in the format of "c:\\Users\\{user-name}\\{path-to-apex-file.cls}" on Windows
```
jeffb-sfdc, line 28: Randi ran this branch on her Windows machine this morning.  The code ran w/o any additional changes needed, and I added a comment here of what the format of `filePath` is here on Windows.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
```diff
@@ -81,13 +93,12 @@ async function getApexTestClassName(
...
-  // This is a little bizarre.  Intellisense is reporting that getTestClassName() returns a string,
-  // but it actually it returns string | undefined.  Well, regardless, since flushFilePath() takes
-  // a string (and guards against empty strings) using the Non-null assertion operator
-  // (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)
-  // fixes the issue.
-  testClassName = fileUtils.flushFilePath(testClassName || '');
+
+  const filePath = sourceUri.toString();
+  // filePath is in the format of "file:///Users/{user-name}/{path-to-apex-file.cls}" on Mac OSX
+  // filePath is in the format of "file:///c%3A/Users/{user-name}/{path-to-apex-file.cls}" on Windows
```
jeffb-sfdc, line 99: Randi ran this branch on her Windows machine this morning.  The code ran w/o any additional changes needed, and I added a comment here of what the format of `filePath` is here on Windows.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -146,8 +146,24 @@ export class ApexTestOutlineProvider
     }
   }
 
-  public getTestClassName(uri: vscode.Uri): string | undefined {
-    return this.testIndex.get(uri.toString());
+  public getTestClassName(filePath: string): string | undefined {
+    // filePath is in the format of "file:///Users/{user-name}/{path-to-apex-file.cls}" on Mac OSX
+    // filePath is in the format of "file:///c%3A/Users/{user-name}/{path-to-apex-file.cls}" on Windows
```
jeffb-sfdc, line 151: Randi ran this branch on her Windows machine this morning.  The code ran w/o any additional changes needed, and I added a comment here of what the format of `filePath` is here on Windows.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
klewis-sfdc: Does error.message need to be displayed in the channel at all?  Since it is already outputting the generic "Unknown error when launching..." message, maybe the specific error here could be output as a console.warn() or console.error().

Either way, I do agree that error.message needs to be checked before the call to JSON.stringify().  I like this structure on the bottom of the SO page that Gordon linked to: https://stackoverflow.com/a/68240891.

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
klewis-sfdc: Agree with what Gordon has said so far here. Re: iterating over all test files within getTestClassName(), what do you think about lowercasing the `testIndex `entries when it is built in `createTestIndex()` and then lowercase the uri filename when checking the index ` this.testIndex.get(uri.toString().toLowerCase());`?

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
jeffb-sfdc: @klewis-sfdc , yes, it should be displayed, otherwise we have knowledge of what happened or how to go about fixing the issue.

re. `if (err instanceof Error)`:  there's an edge case which this misses, which is if the Message is empty, the error reported is an empty string.

```
const message = (error && error.message)
      ? error.message
      : JSON.stringify(error);
```
...handles this case.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
gbockus-sf: I probably wasn't clear in my original comment.  The issue with this comment is the path isn't guaranteed to be in the users home space on either windows or mac.  `/Users/{user-name}/` and `\\Users\\{user-name}\\` aren't always present.  Is the point of this comment to communicate that the filePath string will not include the file:/ scheme?  If so I could see going with something like ```
// The filePath should be an fs filepath and not a VSCode URI string. No scheme (file:/) should be present.
// Examples: 
// "/Users/{user-name}/{path-to-apex-file.cls}" on Mac OSX
// "c:\\Users\\{user-name}\\{path-to-apex-file.cls}" on Windows```

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
gbockus-sf: @jeffb-sfdc do you agree we need to wrap `JSON.stringify` in a try/catch?

packages/salesforcedx-vscode-apex/src/commands/forceLaunchApexReplayDebuggerWithCurrentFile.ts
gbockus-sf: yeah I'm not a fan of unhandled exceptions myself...but I do  trust some of the larger projects have flushed out the details around the native method and I'd rather lean on their experience then learn all the quirks of that method ourselves.  I wish I could just share the link to the real work examples I'm seeing in vscode but alas it doesn't work that way. I see this by hovering over the native method and clicking the 'See Real World example from Github' link. 
Screenshot 2023-02-21 at 5 00 06 PM

[Typescript](https://github.dev/microsoft/TypeScript/blob/70585aa0c0e4fbd2921f340f276d87fef8801997/src/compiler/sys.ts#L1852%23L1852)
[Jest](https://github.dev/facebook/jest/blob/80cf80055682919251b9567ef482755d5f060d47/packages/jest-util/src/tryRealpath.ts#L12%23L12)
[eslint limits it to specific os](https://github.dev/microsoft/vscode-eslint/blob/1e44f688319912a3671e7070187f2fc7a5430ea1/server/src/eslintServer.ts#L825%23L825)
[opensumi (some framework to build vscode extenions 🤷 )](https://github.dev/opensumi/core/blob/574ebf8423f0fa8c6a7da2be90e3fa1a246e9e2c/packages/file-service/src/node/file-service-watcher.ts#L321%23L321)

There's a few others and not all swallow the exception but those that I looked at appears to be filtering by os and other method prior to calling the native method.  

I'd totally be down to have a catch that posts a custom event to our analytics with the path so we could try and track down when the native method throws. (we could also add a console.log so we could easily identify it's happening with the end user).

gbockus-sf:
closing for now

PR #4675, opened 2023-02-13
chore: update request-light dependency to avoid patching https.request

gbockus-sf:
### What does this PR do?
Update the version of request-light we import and ^ the dep so we pickup updates in the future. 

This is to fix the issue where when SDR attempts to go find the current API version via a request using the got module it is failing due to an older agent-base that patched https.request.

The error was only present in when running via vsix.  I believe with was b/c request-light is getting bundled with the source and it would load in such a way that the https.request was patched at startup vs when needed when running in debugging mode.

The problem code can be found in the before version bundle by running `npm run vscode:bundle` from the top level directory of the project.  Then view the `packages/salesforcedx-vscode-core/dist/index.js` file and search for `https.request =`.  
Note this is no longer present after the update. 

*Before*
npm ls agent-base 
```
├─┬ request-light@0.2.4 extraneous
│ ├── agent-base@4.3.0 extraneous
│ ├─┬ http-proxy-agent@2.1.0 extraneous
│ │ └── agent-base@4.3.0 deduped
│ └─┬ https-proxy-agent@2.2.4 extraneous
│   └── agent-base@4.3.0 deduped
```

*After*
 npm ls agent-base
request-light no longer imports agent-base

@W-9466545@

### Functionality Before
Retrieve operation would fail due with the following error 
`An error was encountered during conflict detection. RequestError: connect ECONNREFUSED 127.0.0.1:443`

### Functionality After
Retrieve works as expected (triggered via the deploy with conflict detection turned on)

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -15,7 +15,7 @@
     "@salesforce/salesforcedx-utils": "57.2.0",
     "async-lock": "1.0.0",
     "faye": "1.1.2",
-    "request-light": "0.2.4",
+    "request-light": "^0.7.0",
     "vscode-debugadapter": "1.28.0",
     "vscode-debugprotocol": "1.28.0"
   },
--- a/packages/salesforcedx-apex-replay-debugger/test/unit/commands/apexExecutionOverlayResultCommand.test.ts
+++ b/packages/salesforcedx-apex-replay-debugger/test/unit/commands/apexExecutionOverlayResultCommand.test.ts
@@ -684,8 +684,8 @@ export function createExpectedXHROptions(
       Accept: 'application/json',
       Authorization: `OAuth 123`,
       'Content-Length': requestBody
-        ? Buffer.byteLength(requestBody, 'utf-8')
-        : 0,
+        ? String(Buffer.byteLength(requestBody, 'utf-8'))
+        : '0',
       'Sforce-Call-Options': `client=${CLIENT_ID}`
     },
     data: requestBody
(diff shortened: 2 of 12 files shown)


packages/salesforcedx-apex-debugger/package.json
```diff
@@ -15,7 +15,7 @@
     "@salesforce/salesforcedx-utils": "57.2.0",
     "async-lock": "1.0.0",
     "faye": "1.1.2",
-    "request-light": "0.2.4",
+    "request-light": "^0.7.0",
```
gbockus-sf, line 18: We've wavered on if we should pin to versions or not.  For this update The changelog for request-light appears to follow correct semantic versioning and is really only updated to update dependancies so add the ^ seemed like the right thing to do.

packages/salesforcedx-apex-replay-debugger/test/unit/commands/apexExecutionOverlayResultCommand.test.ts
```diff
@@ -684,8 +684,8 @@ export function createExpectedXHROptions(
       Accept: 'application/json',
       Authorization: `OAuth 123`,
       'Content-Length': requestBody
-        ? Buffer.byteLength(requestBody, 'utf-8')
-        : 0,
+        ? String(Buffer.byteLength(requestBody, 'utf-8'))
+        : '0',
```
gbockus-sf, line 688: Headers in http land can only be strings.  They tightened this up on the types here. I verified that things still work as expected for toggling/clearing checkpoints where is one of the only places we actually use the requestService.

klewis-sfdc: Sweet.  Thx for converting these all to strings.

klewis-sfdc (approved):
Looks good. Thanks Gordon!

Issue #4673, opened 2023-02-09
Checkpoints don't update correctly in the Interactive Checkpoint Side Panel Window


AnanyaJha:
### Summary

The Checkpoint Window in the side panel only displays the most recent checkpoint that was enabled. If you toggle multiple checkpoints on and off, whether through the Command Palette or Right Click Context Menus, the window will only show the most recent checkpoint.

### Steps To Reproduce:

1. Enable a checkpoint in your file through the CP or right click context menu
2. Click on the "Debug" tab and look at the "Checkpoint" window. It should display your recently enabled checkpoint.
3. Enable another checkpoint in your file.

### Expected result

Both checkpoints are shown in the Checkpoint window

### Actual result

Only the last checkpoint is displayed in the window.

### Additional information

<img width="459" alt="Screen Shot 2023-02-08 at 5 32 29 PM" src="https://user-images.githubusercontent.com/21326913/217693068-f4ba5ad2-ef44-42eb-8b51-eda8dc662bd3.png">

**Salesforce Extension Version in VS Code**: 56.16.0

**SFDX CLI Version**: sfdx-cli/7.186.2

gbockus-sf:
Looking into this issue I found that there has been a change to how we can id breakpoints for tracking the existence of the checkpoint in the side panel. 

Repo steps:
- Add a checkpoint on a line (say 7) via the CP.
- Add another checkpoint via the context menu on the next line (add a breakpoint, then edit it setting the expression to checkpoint)

Issue #4534, opened 2022-11-03
Test tab somehow broken: missing indicators, shifted names


danikar208:
### Summary

_After some time the testing tab does not seem to be working properly. The indicators are missing at all times, after running tests the method names are shifted in a weird manner._.

### Steps To Reproduce:

1.  Run some tests
1.  Check the "Testing" tab for the ran tests

### Expected result

_We all know how should it look like_.

![2](https://user-images.githubusercontent.com/56158777/199737667-132fa24b-1bcc-493f-afe1-a6b5c60baa6f.PNG)

### Actual result

![1](https://user-images.githubusercontent.com/56158777/199737556-c904ff62-bd06-4179-8ca1-1f1c67060683.PNG)

**Salesforce Extension Version in VS Code**:
v56.3.1

**SFDX CLI Version**:
sfdx-cli/7.174.0 win32-x64 node-v16.17.1

**OS and version**:
Windows 10 21H2

mihailTr-19:
I have the same issue.

danikar208:
> Hi @danikar208 - I've seen this issue occasionally as well. Do you find that rebooting VS Code addresses the issue for you?

Hey, at first when the issues appeared, a reboot would had sufficed. Right now it never goes away, rebooting does not work of course, neither do reinstallation of extensions.

jbelloperezSFDC:
This issue happened from version 56.3.1 on to me. I've installed 55.8.0 and that works fine, but any version after that will hide the pass/fail icon

danikar208:
@jbelloperezSFDC How'd you install a later version?

danikar208:
Oh well, I loved the screaming red error diodes. I'll stick with an older version until something major comes out.

Won't close this issue, until we get a definite answer, if it's a feature now or not and the shifted faulty method names still feel really wonky.

randi274:
Thanks for your patience on this one - I'm seeing it consistently as well. We'll add to our backlog and get this one addressed soon as we can!

gbockus-sf:
Fix coming in the next release 👍

randi274:
Closing this one as it was fixed a few weeks ago, and should be available in the latest and great version of the extensions 🙌

PR #4686, opened 2023-02-16
fix: update test icons to account for bundling, fix apexDB location, and fix checkpoint issue

gbockus-sf:
### What does this PR do?
Primarily this PR addresses fixing accessing files in the resources directory for bundled vsixes. 
 - Icons shows in the lwc and Apex Test views.
 - Setting up the apexDb

Secondarily there was a simple fix to how checkpoints are handled so I just did the update here. 

@W-12545057@ #4534 
@W-12545057@ #4673 

### Functionality Before
No icons would show up in the Test views.
checkpoints would be removed out of order in the checkpoints view.

### Functionality After
We have icons.
Screenshot 2023-02-16 at 12 42 09 AM

checkpoints are correctly removed from the debug panel when toggle in a cls. 
Screenshot 2023-02-16 at 12 44 40 AM

Notes: 
To exercise the checkpoints functionality you can toggle a checkpoint via the `SFDX: Toggle Checkpoint` command in the CP or by creating a breakpoint and setting the expression to 'checkpoint'.  Checkpoints can also be removed by clicking an active checkpoint (breakpoint).

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "vscode:bundle:debugger": "lerna run bundle:debugger",
     "vscode:bundle:extension": "lerna run bundle:extension",
     "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
-    "vsix:install": "find ./packages -name '*.vsix' -exec code --install-extension {} \\;",
+    "vsix:install": "find ./packages -name '*.vsix' -exec code-insiders --install-extension {} \\;",
     "vscode:sha256": "node ./scripts/generate-sha256.js >> ./SHA256",
     "vscode:publish": "lerna run vscode:publish --concurrency 1",
     "watch": "lerna run --parallel watch",
--- a/packages/salesforcedx-utils-vscode/src/helpers/extensionUris.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/extensionUris.ts
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as vscode from 'vscode';
+
+function extensionUri(extensionName: string) {
+  const extensionRef = vscode.extensions.getExtension(extensionName);
+
+  if (extensionRef) {
+    return extensionRef.extensionUri;
+  }
+  throw new Error(`Unable to find extension ${extensionName}`);
+}
+
+function join(baseUri: vscode.Uri, relativePath: string) {
+  return vscode.Uri.joinPath(baseUri, relativePath);
+}
+
+export const extensionUris = {
+  extensionUri,
+  join
+};
--- a/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/paths.ts
@@ -147,7 +147,10 @@ function relativeStateFolder(): string {
 }
 
 function relativeToolsFolder(): string {
-  const relativePathToToolsFolder = path.join(projectPaths.relativeStateFolder(), TOOLS);
+  const relativePathToToolsFolder = path.join(
+    projectPaths.relativeStateFolder(),
+    TOOLS
+  );
   return relativePathToToolsFolder;
 }
 
--- a/packages/salesforcedx-utils-vscode/test/jest/helpers/extensionUris.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/jest/helpers/extensionUris.test.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as vscode from 'vscode';
+import { extensionUris } from '../../../src';
+
+describe('extensionUris Unit Tests.', () => {
+  describe('extensionUri()', () => {
+    const fakeExtensionRef = {
+      extensionUri: { url: 'fake' }
+    };
+    const fakeExtensionName = 'salesforce.someExtension';
+
+    beforeEach(() => {
+      vscode.extensions.getExtension = jest.fn();
+    });
+
+    it('Should return extensionUrl if found.', () => {
+      (vscode.extensions.getExtension as any).mockReturnValue(fakeExtensionRef);
+      const result = extensionUris.extensionUri(fakeExtensionName);
+      expect(vscode.extensions.getExtension).toHaveBeenCalledWith(
+        fakeExtensionName
+      );
+      expect(result).toEqual(fakeExtensionRef.extensionUri);
+    });
+
+    it('Should throw an error if extension is not found.', () => {
+      (vscode.extensions.getExtension as any).mockReturnValue(undefined);
+      expect(() => {
+        extensionUris.extensionUri(fakeExtensionName);
+      }).toThrowError(/^Unable to find extension/);
+    });
+  });
+
+  describe('join()', () => {
+    const fakeResultUri = { url: 'file://totallyfake' };
+    it('Should return joined Uri and relative path.', () => {
+      const fakeUri: any = { scheme: true };
+      const path = '/a/b/c/e';
+      (vscode.Uri.joinPath as any).mockReturnValue(fakeResultUri);
+      const result = extensionUris.join(fakeUri, path);
+      expect(vscode.Uri.joinPath).toHaveBeenCalledWith(fakeUri, path);
+      expect(result).toEqual(fakeResultUri);
+    });
+  });
+});
--- a/packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
@@ -808,14 +808,14 @@ export async function processBreakpointChangedForCheckpoints(
   for (const bp of breakpointsChangedEvent.removed) {
     if (bp.condition && bp.condition!.toLowerCase().indexOf(CHECKPOINT) >= 0) {
       await lock.acquire(CHECKPOINTS_LOCK_STRING, async () => {
-        const breakpointId = (bp as any)._id;
+        const breakpointId = bp.id;
         checkpointService.deleteCheckpointNodeIfExists(breakpointId);
       });
     }
   }
 
   for (const bp of breakpointsChangedEvent.changed) {
-    const breakpointId = (bp as any)._id;
+    const breakpointId = bp.id;
     if (
       bp.condition &&
       bp.condition!.toLowerCase().indexOf(CHECKPOINT) >= 0 &&
@@ -862,7 +862,7 @@ export async function processBreakpointChangedForCheckpoints(
       bp instanceof vscode.SourceBreakpoint
     ) {
       await lock.acquire(CHECKPOINTS_LOCK_STRING, async () => {
-        const breakpointId = (bp as any)._id;
+        const breakpointId = bp.id;
         const checkpointOverlayAction = parseCheckpointInfoFromBreakpoint(bp);
         const uri = code2ProtocolConverter(bp.location.uri);
         const filename = uri.substring(uri.lastIndexOf('/') + 1);
--- a/packages/salesforcedx-vscode-apex-replay-debugger/test/vscode-integration/checkpoints/checkpointService.test.ts
+++ b/packages/salesforcedx-vscode-apex-replay-debugger/test/vscode-integration/checkpoints/checkpointService.test.ts
@@ -199,7 +199,6 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
       CHECKPOINT,
       undefined
     );
-    (breakpoint as any)._id = breakpointId;
     bpAdd.push(breakpoint);
     await processBreakpointChangedForCheckpoints({
       added: bpAdd,
@@ -209,7 +208,7 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
 
     // Verify that a single checkpoint has been added to the checkpoint service
     const theNode = checkpointService.returnCheckpointNodeIfAlreadyExists(
-      breakpointId
+      breakpoint.id
     );
     if (!theNode) {
       assert.fail(
@@ -232,8 +231,7 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
       CHECKPOINT,
       undefined
     );
-    const breakpointId1 = breakpointId + '1';
-    (breakpoint1 as any)._id = breakpointId1;
+
     bpAdd.push(breakpoint1);
     const breakpoint2 = new vscode.SourceBreakpoint(
       location,
@@ -241,8 +239,7 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
       CHECKPOINT,
       undefined
     );
-    const breakpointId2 = breakpointId + '2';
-    (breakpoint2 as any)._id = breakpointId2;
+
     bpAdd.push(breakpoint2);
 
     await processBreakpointChangedForCheckpoints({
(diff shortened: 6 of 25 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "vscode:bundle:debugger": "lerna run bundle:debugger",
     "vscode:bundle:extension": "lerna run bundle:extension",
     "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
-    "vsix:install": "find ./packages -name '*.vsix' -exec code --install-extension {} \\;",
+    "vsix:install": "find ./packages -name '*.vsix' -exec code-insiders --install-extension {} \\;",
```
gbockus-sf, line 85: if I'm the only one using this for vsix testing figured I'd update instead of always editing it.

packages/salesforcedx-utils-vscode/src/helpers/extensionUris.ts
```diff
@@ -0,0 +1,26 @@
...
+    return extensionRef.extensionUri;
+  }
+  throw new Error(`Unable to find extension ${extensionName}`);
+}
+
+function join(baseUri: vscode.Uri, relativePath: string) {
+  return vscode.Uri.joinPath(baseUri, relativePath);
+}
+
+export const extensionUris = {
```
gbockus-sf, line 23: In the spirit of always handing Uri's the same way created a common helper in utils to handle repeatable tasks.  Think pathUtils but for Uris.

packages/salesforcedx-utils-vscode/src/helpers/paths.ts
```diff
@@ -147,7 +147,10 @@ function relativeStateFolder(): string {
 }
 
 function relativeToolsFolder(): string {
-  const relativePathToToolsFolder = path.join(projectPaths.relativeStateFolder(), TOOLS);
+  const relativePathToToolsFolder = path.join(
+    projectPaths.relativeStateFolder(),
+    TOOLS
+  );
```
gbockus-sf, line 153: format 🤷

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -808,14 +808,14 @@ export async function processBreakpointChangedForCheckpoints(
   for (const bp of breakpointsChangedEvent.removed) {
     if (bp.condition && bp.condition!.toLowerCase().indexOf(CHECKPOINT) >= 0) {
       await lock.acquire(CHECKPOINTS_LOCK_STRING, async () => {
-        const breakpointId = (bp as any)._id;
+        const breakpointId = bp.id;
```
gbockus-sf, line 811: this was the fix for the breakpoints panel. I'm not sure when this changed in the vscode spec, but def calls to light not referencing private properties in 3rd party libs as a best practice.

packages/salesforcedx-vscode-apex-replay-debugger/test/vscode-integration/checkpoints/checkpointService.test.ts
```diff
@@ -209,7 +208,7 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
 
     // Verify that a single checkpoint has been added to the checkpoint service
     const theNode = checkpointService.returnCheckpointNodeIfAlreadyExists(
-      breakpointId
+      breakpoint.id
```
gbockus-sf, line 211: the id value is readonly so we were unable to manipulate the breakpoint Id as was done previously.  I just updated to use the existing id and it worked

packages/salesforcedx-vscode-apex/jest.config.js
```diff
@@ -0,0 +1,4 @@
+// eslint:disable-next-line:no-var-requires
+const baseConfig = require('../../config/jest.base.config');
+
+module.exports = Object.assign({}, baseConfig);
```
gbockus-sf, line 4: first jest test in vscode-apex package.

packages/salesforcedx-vscode-apex/src/constants.ts
```diff
@@ -1,97 +1,21 @@
...
  */
 
-import * as path from 'path';
 import * as vscode from 'vscode';
-import { nls } from './messages';
 
 export const DEBUGGER_LINE_BREAKPOINTS = 'debugger/lineBreakpoints';
 export const DEBUGGER_EXCEPTION_BREAKPOINTS = 'debugger/exceptionBreakpoints';
 
-export const LIGHT_BLUE_BUTTON = path.join(
```
gbockus-sf, line 15: So the issue with the icons was the code assumed a location of the source and found the resources directory from the current file location.  That doesn't work with bundling b/c the source moves from this dir to the minified js file. I created a new class to handle it in a sane way instead of trying to hack it together as a constant value.

packages/salesforcedx-vscode-apex/src/helpers/languageServerUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+import {
+  extensionUris,
+  projectPaths
+} from '@salesforce/salesforcedx-utils-vscode';
+import { copyFileSync, existsSync, unlinkSync } from 'fs';
+import { join } from 'path';
+import * as vscode from 'vscode';
+import { VSCODE_APEX_EXTENSION_NAME } from '../constants';
+
+const setupDB = (): void => {
```
gbockus-sf, line 16: Just refactored this out of the file of the class that uses it to make it easier to unit test.

packages/salesforcedx-vscode-apex/src/helpers/languageServerUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (
+    vscode.workspace.workspaceFolders &&
+    vscode.workspace.workspaceFolders[0]
+  ) {
+    const dbPath = projectPaths.apexLanguageServerDatabase();
+    if (existsSync(dbPath)) {
+      unlinkSync(dbPath);
+    }
+
+    try {
```
gbockus-sf, line 26: Same story as the icons logic wise.  This was assuming previously it's location in the file system to find the resources directory ala `const systemDb = path.join(__dirname, '..', '..', 'resources', 'apex.db');`
I updated it to use the extensionUri method to find the extension directory which works for both dev and published versions.

packages/salesforcedx-vscode-apex/src/views/icons/IconsEnum.ts
```diff
@@ -0,0 +1,31 @@
...
+export enum IconsEnum {
+  LIGHT_BLUE_BUTTON = 'LIGHT_BLUE_BUTTON',
+  LIGHT_RED_BUTTON = 'LIGHT_RED_BUTTON',
+  LIGHT_ORANGE_BUTTON = 'LIGHT_ORANGE_BUTTON',
+  LIGHT_GREEN_BUTTON = 'LIGHT_GREEN_BUTTON',
+  DARK_BLUE_BUTTON = 'DARK_BLUE_BUTTON',
+  DARK_RED_BUTTON = 'DARK_RED_BUTTON',
+  DARK_GREEN_BUTTON = 'DARK_GREEN_BUTTON',
+  DARK_ORANGE_BUTTON = 'DARK_ORANGE_BUTTON'
+}
```
gbockus-sf, line 31: So my goal here was to make using this icons in code a simple and type safe thing.  Will break it down in the helper file.

packages/salesforcedx-vscode-apex/src/views/icons/iconHelpers.ts
```diff
@@ -0,0 +1,27 @@
...
+import { extensionUris } from '@salesforce/salesforcedx-utils-vscode';
+import { VSCODE_APEX_EXTENSION_NAME } from '../../constants';
+import { iconKey, ICONS } from './IconsEnum';
+
+/**
+ * Get the Uri for an icon located in the resources directory.
+ * @param key A key from the {@link IconsEnum}
+ * @returns The Uri to the icon image.
+ */
+const getIconPath = (key: iconKey) => {
```
gbockus-sf, line 17: So we needed a way to get iconPaths.  My first approach here was to initialize with the extensionContext so I could get the extensionPath.  That lead to some ugliness due to the file using this being initialized early in the activation flow.  To avoid race condition games and have to store the extensionContext here I altered this to use the extensionUri to get the base extension location.

packages/salesforcedx-vscode-apex/src/views/icons/iconHelpers.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { extensionUris } from '@salesforce/salesforcedx-utils-vscode';
+import { VSCODE_APEX_EXTENSION_NAME } from '../../constants';
+import { iconKey, ICONS } from './IconsEnum';
+
+/**
+ * Get the Uri for an icon located in the resources directory.
+ * @param key A key from the {@link IconsEnum}
+ * @returns The Uri to the icon image.
+ */
```
gbockus-sf, line 16: Screenshot 2023-02-16 at 1 02 26 AM

Note the param here is limited to the key values from the Icons const (which are also the keys for the IconsEnum). 
This allows us to use this method safely always limiting the param to the available icons (or getting a compile error)

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -368,4 +359,11 @@ export class ApexTestNode extends TestNode {
   public contextValue = 'apexTest';
 }
 
-export const testOutlineProvider = new ApexTestOutlineProvider(null);
```
gbockus-sf, line 371: So this was leading to all kinds of sadness trying to both write the tests and ensure when this ApexTestOutputProvider was being constructed.  I don't love the poor substitute for a singleton, but I would have had to majorly refactor the int tests if I altered how this property is referenced. 

This seemed like the lesser of two evils path but I'm open to suggestions.

packages/salesforcedx-vscode-lwc/src/testSupport/testExplorer/iconPaths.ts
```diff
@@ -4,62 +4,51 @@
...
+import { extensionUris } from '@salesforce/salesforcedx-utils-vscode';
 import * as path from 'path';
+import * as vscode from 'vscode';
+import { VSCODE_LWC_EXTENSION_NAME } from '../../constants';
 import { TestResult, TestResultStatus } from '../types';
 
-const extensionPath = path.join(__filename, '..', '..', '..', '..', '..');
-const LIGHT_BLUE_BUTTON = path.join(
+const extensionPath = extensionUris.extensionUri(VSCODE_LWC_EXTENSION_NAME);
+const LIGHT_BLUE_BUTTON = extensionUris.join(
```
gbockus-sf, line 15: The LWC constants were setup to be more easily updated so I went the easy path here instead of having two identical icon helpers.

packages/salesforcedx-vscode-lwc/test/vscode-integration/testSupport/testExplorer/testNode.test.ts
```diff
@@ -88,34 +88,48 @@ describe('Sfdx Test Node', () => {
       testExecutionInfo.testResult = undefined;
       const mockLabel = 'mockTest';
       const groupNode = new SfdxTestNode(mockLabel, testExecutionInfo);
-      expect(groupNode.iconPath.dark.endsWith('testNotRun.svg')).to.equal(true);
-      expect(groupNode.iconPath.light.endsWith('testNotRun.svg')).to.equal(
-        true
-      );
+      expect(
+        groupNode.iconPath.dark.toString().endsWith('testNotRun.svg')
```
gbockus-sf, line 92: These updates are b/c the icon is now a Uri instead of a string path.

package.json
klewis-sfdc: Hmm - maybe we could have two?  vsix:install and vsix:install:insiders?

packages/salesforcedx-vscode-apex/src/views/icons/iconHelpers.ts
klewis-sfdc: Sweeet

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
klewis-sfdc: This looks like a solid solution to avoid the big refactor to me.

packages/salesforcedx-vscode-apex/src/helpers/languageServerUtils.ts
klewis-sfdc: Solid

packages/salesforcedx-utils-vscode/test/jest/helpers/extensionUris.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const fakeExtensionRef = {
+      extensionUri: { url: 'fake' }
+    };
+    const fakeExtensionName = 'salesforce.someExtension';
+
+    beforeEach(() => {
+      vscode.extensions.getExtension = jest.fn();
+    });
+
+    it('Should return extensionUrl if found.', () => {
```
klewis-sfdc, line 22:
```suggestion
    it('Should return extensionUri if found.', () => {
```

klewis-sfdc:
These code changes look great.  Thanks for the update to make icons much more easy to work with!   

I left a question re the script and a super small suggested fix.  And it looks like the integration tests need to be tweaked but this looks really good.

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -280,20 +271,20 @@ export abstract class TestNode extends vscode.TreeItem {
     if (outcome === 'Pass') {
```
randi274, line 271: Asking because you're in here anyway; is it worth extrapolating all the outcome strings into variables?

gbockus-sf: good idea..on it

packages/salesforcedx-vscode-apex/src/constants.ts
```diff
@@ -1,97 +1,27 @@
...
 const endPos = new vscode.Position(0, 1);
 export const APEX_GROUP_RANGE = new vscode.Range(startPos, endPos);
 export const SET_JAVA_DOC_LINK =
   'https://developer.salesforce.com/tools/vscode/en/vscode-desktop/java-setup';
 export const SFDX_APEX_CONFIGURATION_NAME = 'salesforcedx-vscode-apex';
 export const APEX_EXTENSION_NAME = 'salesforcedx-vscode-apex';
+export const VSCODE_APEX_EXTENSION_NAME = `salesforce.${APEX_EXTENSION_NAME}`;
 export const LSP_ERR = 'apexLSPError';
+
+export const PASS_RESULT = 'Pass';
```
randi274, line 23: 🎉

packages/salesforcedx-vscode-apex/src/views/testOutlineProvider.ts
```diff
@@ -34,6 +31,9 @@ const NO_TESTS_DESCRIPTION = nls.localize(
   'force_test_view_no_tests_description'
 );
 
+const TEST_RUN_ID_FILE = 'test-run-id.txt';
+const TEST_RESULT_JSON_FILE = `test-result.json`;
```
randi274, line 35: Nit: ` vs ' here?

randi274:
QA:
✅ Colorful test indicators appear in both light and dark mode for Apex & LWC tests (confirmed blue, green, red, and orange)
✅ Checkpoints are now added/removed in a logical order, including when a breakpoint is converted
✅ Uploaded checkpoints and used replay debugger to make sure they were properly set
✅ Tested on both Mac and Windows

randi274 (approved):

packages/salesforcedx-vscode-apex-replay-debugger/test/vscode-integration/checkpoints/checkpointService.test.ts
```diff
@@ -322,188 +322,6 @@ describe('Verify checkpoint callback for vscode.debug.onDidChangeBreakpoints', (
     expect(lockSpy.getCall(1).args[0]).to.equal(CHECKPOINTS_LOCK_STRING);
   });
```
gbockus-sf, line 324: These three tests depended on all checkpoints sharing the same _id property (undefined).  I tried to see if I could use the real id property to get these tests to still work, but not luck (id is readonly).  I also attempted to _fake it_ but b/c there is an instanceof check in the code flow there was no getting around it.  Elected to just remove these int tests vs continuing the fight or changing logic to accommodate the int tests.

PR #4687, opened 2023-02-16
feat: remove container start command

jabrown85:
### What does this PR do?
We are removing this command. Users are to use the local version instead.

[@W-12553533@](https://gus.lightning.force.com/a07EE00001LMfHyYAL) / @W-12345678@

https://github.com/salesforcecli/plugin-functions/pull/612
https://github.com/heroku/sf-functions-core/pull/231

### Functionality Before
Users could start a function using Docker.

### Functionality After
Users will no longer be able to start a function using Docker.


packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionContainerStartExecutor.ts
```diff
@@ -1,148 +0,0 @@
...
- */
-
-import { Benny, getFunctionsBinary } from '@heroku/functions-core';
-import * as vscode from 'vscode';
-import { channelService } from '../../../channels';
-import { nls } from '../../../messages';
-import { notificationService } from '../../../notifications';
-import { telemetryService } from '../../../telemetry';
-import { FunctionService } from '../functionService';
-import { FUNCTION_RUNTIME_DETECTION_PATTERN } from '../types/constants';
```
edmorley, line 15: `FUNCTION_RUNTIME_DETECTION_PATTERN` is now unused:
https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/src/commands/functions/types/constants.ts#L30-L35

edmorley (commented):
Thank you for tackling this! :-)

I've found a few other things that can be removed:
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/package.nls.json#L80
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/src/messages/i18n.ts#L161
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/package.json#L203-L206
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/package.json#L329-L332
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/package.json#L627-L630
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/package.json#L962-L965
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/docs/_articles/en/functions/overview.md?plain=1#L19
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts#L356-L376 (plus a few others like it in that file)
- Things that refer to `isContainerless: false`: https://github.com/search?q=repo%3Aforcedotcom%2Fsalesforcedx-vscode%20isContainerLess&type=code
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStop.ts#L17
- https://github.com/forcedotcom/salesforcedx-vscode/blob/7a5f09583630a1518787abf8b28ab95d34460cff/packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts#L99

jabrown85:
I guess this is ready to build if someone wants to kick off the actions for us? I didn't run tests locally so I'm sure there are things to fix 😄

docs/_articles/en/functions/overview.md
```diff
@@ -30,4 +29,4 @@ A sample `payload.json` file that can invoke a function:
 Stop a function by running `SFDX: Stop Function`, which simply kills the local process.
 
 ### Resources
-For more information see [Get Started with Salesforce Functions](https://developer.salesforce.com/docs/platform/functions/guide/index.html). 
+For more information see [Get Started with Salesforce Functions](https://developer.salesforce.com/docs/platform/functions/guide/index.html).
```
gbockus-sf, line 32: @sbudhirajadoc Want to verify a doc change?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
```diff
@@ -232,9 +225,7 @@ export class FunctionService {
       port: debugPort
     };
 
-    if (functionExecution.isContainerLess) {
-      delete debugConfiguration.remoteRoot;
-    }
+    delete debugConfiguration.remoteRoot;
```
gbockus-sf, line 228: can we just remove the removeRoot property from the `getDebugConfiguration` function?

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStop.ts
```diff
@@ -1,38 +0,0 @@
-/*
```
gbockus-sf, line 1: Unless I'm missing something this is still necessary.  From what I recall there are three interactions with function 
`start` -> start the function process 
`invoke` code lens -> run/debug the function
`stop` -> stop the running function

If there's no stop how does the user stop a function that has been started?

packages/salesforcedx-vscode-core/src/commands/functions/functionService.ts
gbockus-sf: we'll need to remove it from [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/b5bacd0dd44e6d8b4e2e1f6a593eb8af64f40181/packages/salesforcedx-vscode-core/test/vscode-integration/commands/functions/functionService.test.ts#L106) as well

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -459,8 +447,6 @@ function registerCommands(
     forceFunctionInvokeCmd,
     forceFunctionDebugInvokeCmd,
     forceFunctionStartCmd,
-    forceFunctionContainerStartCmd,
-    forceFunctionStopCmd,
```
gbockus-sf, line 463: I think these stops will need to be put back as well.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -431,11 +424,6 @@ function registerCommands(
     forceFunctionDebugInvoke
   );
 
-  const forceFunctionStopCmd = vscode.commands.registerCommand(
-    'sfdx.force.function.stop',
-    forceFunctionStop
-  );
```
gbockus-sf, line 437: This is what wires in the command palette to have the `SFDX: Stop Function` command which is still necessary I believe

packages/salesforcedx-vscode-core/src/commands/functions/forceFunctionStart/ForceFunctionStartExecutor.ts
```diff
@@ -70,64 +70,26 @@ export abstract class ForceFunctionStartExecutor extends LibraryCommandletExecut
...
 
-    await this.setupFunctionListeners(
-      functionDirPath,
-      registeredStartedFunctionDisposable
-    );
-
-    token?.onCancellationRequested(() => {
-      this.cancelFunction(registeredStartedFunctionDisposable);
-      registeredStartedFunctionDisposable.dispose();
-    });
```
gbockus-sf, line 102: THis is how we ensure we cleanup if the user clicks the cancel button while the command is executing. Why'd it go away?

gbockus-sf (changes requested):
Build can be seen [here](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/4235573550) We currently don't have a good way to start the GHA flow for non-contributors so if you're going to be here a lot we can make you a contributor or we can just keep on keeping on with pushing a branch of your forked changed. 

Only one int test failure that should be an easy update. 

Couple of things to consider from review
1. Why'd we remove the stop function command and related code.
2. Why'd we remove the Cancel handling?

jabrown85:
> Why'd we remove the stop function command and related code.
It looked to me like it was only ever using the container code that was being removed, not the container-less. But perhaps I missed something? Does container stop work with containerless today?
>Why'd we remove the Cancel handling?
Same as above. At least that is how I read it. I see `isContainerLess: false` in that code and nothing leads me to believe it works for containerless.

gbockus-sf:
@jabrown85 Yeah the way this thing was built def leads to it not being overly clear.
We def still need the Stop b/c that's what stops the running process that was started by the start command for locally running functions (and we're only using one port today so if you have a running function you can't start another function) 

`forceFunctionStop -> FunctionService.instance.stopFunction(); -> functionExecution.terminate();`
That terminate is the property on `registeredStartedFunctionDisposable` that calls cancel function. 

`isContainerLess` get updated in the `updateFunction` call on the FunctionService from the ContainerlessStartExecutor.

clear as mud.  :) Lemme know if you want to pair up and run through it.

jabrown85:
I tried to recover the files/bits I removed that I think drive Function Stop. Can you let actions run again?

gbockus-sf:
looks like a bit more polish: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/4245425078/jobs/7380912334 
You can verify that locally by running `npm run compile` from the project root

jabrown85:
Locally `npm run compile` works for me now. Hoping any failed tests this go round will be easy enough to fix without learning how to run the world locally 😄

gbockus-sf:
The tests were all green except that one int test before the polish work.  I'll be qe'ing after we get the build right. 👍

jabrown85:
Hmm I don't know why that test is failing. Do you have a guess @gbockus-sf?

gbockus-sf:
@jabrown85 Taking a look at the failing stop int tests now.

gbockus-sf:
@jabrown85 After much head scratching I realized the stop int test were previously using container functions (or mocking them anyway). I see what I can do to update to local functions

gbockus-sf:
Merged here: https://github.com/forcedotcom/salesforcedx-vscode/pull/4707. Thanks @jabrown85

PR #4691, opened 2023-02-18
chore: Upgrade source-deploy-retrieve

klewis-sfdc:
### What does this PR do?
* Upgrades source-deploy-retrieve library to pull in recent changes that will enable #4597 

@W-12566946@

Note: I figured this should be its own PR since it is also doing the upgrade for salesforcedx-vscode-apex and there might be some specific QA we'd like to do for that pkg before merging.

### Before
`"@salesforce/source-deploy-retrieve": "7.5.5"`

### After
`"@salesforce/source-deploy-retrieve": "7.10.0"` (or higher)

**Todos:**
- [ ] QA 
  - deploy/retrieve to cover upgrade in core, 
  - run apex tests to cover upgrade in apex pkg


packages/salesforcedx-vscode-apex/package.json
```diff
@@ -29,7 +29,7 @@
     "@salesforce/apex-tmlanguage": "1.4.0",
     "@salesforce/core": "^3.31.18",
     "@salesforce/salesforcedx-utils-vscode": "57.2.1",
-    "@salesforce/source-deploy-retrieve": "7.5.5",
+    "@salesforce/source-deploy-retrieve": "7.10.0",
```
gbockus-sf, line 32: Needs to be bumped in the packageUpdates -> dependancies as well

packages/salesforcedx-vscode-core/package.json
```diff
@@ -31,7 +31,7 @@
     "@salesforce/salesforcedx-sobjects-faux-generator": "57.2.1",
     "@salesforce/salesforcedx-utils-vscode": "57.2.1",
     "@salesforce/schemas": "^1",
-    "@salesforce/source-deploy-retrieve": "7.5.5",
+    "@salesforce/source-deploy-retrieve": "7.10.0",
```
gbockus-sf, line 34: Needs to be bumped in the packageUpdates -> dependancies as well

package-lock.json
```diff
@@ -8089,20 +8089,20 @@
       }
     },
     "node_modules/@salesforce/source-deploy-retrieve": {
```
gbockus-sf, line 8091: I'm kind of amazed this compiles without syncing the apex/templates dependancies as well

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
```diff
@@ -153,8 +153,8 @@ describe('Force Source Retrieve Component(s)', () => {
           new SourceComponent({
             name: 'MyClassB',
             type,
-            content: path.join(type.directoryName, 'MyClassB.cls'),
-            xml: path.join(type.directoryName, 'MyClassB.cls-meta.xml')
+            content: path.join(String(type.directoryName), 'MyClassB.cls'),
```
jeffb-sfdc, line 156: `String()` here doesn't seem to be needed.  Also, in general it's probably best to avoid using `String()` since there's always the confusion over whether it's a TypeScript `String` or a JavaScript `String`, and when it's a TypeScript `String`, it's a class object and String !== string.

package-lock.json
klewis-sfdc: Should I submit a PR against templates to upgrade SDR to the same version in that repo also?

packages/salesforcedx-vscode-apex/package.json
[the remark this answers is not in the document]
klewis-sfdc: Yep, good eye - should also be updated now in the latest commit.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
klewis-sfdc: TypeScript is beefing about it if I don't make the change.  What is the best way to resolve this TS error?

package-lock.json
klewis-sfdc: Oh wait nm templates doesn't depend on SDR.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceSourceRetrieveMetadata/forceSourceRetrieveCmp.test.ts
jeffb-sfdc: Weird, I must have had a cached version of something.  I wasn't getting any TS errors, and previously `MetadataType.directoryName` was defined as a string.

I just did a git clean, install, and then compile, and now it's defined as `directoryName?: string`
To get past this, use the `!` operator...
```
content: path.join(type.directoryName!, 'MyClassB.cls'),
```
...using `!` is preferred over using `String()`. because `String` has different meaning in TS vs JS, and b/c calling `String()` is calling the `String` class constructor.

klewis-sfdc: 👍🏽 updated to use the non-null assertion operator in latest commit.

klewis-sfdc: I was re-visiting this change and decided to revert back to using String() for a few reasons:

* In the upgraded SDR library, `directoryName` is now optional as you mentioned above.  The non-null assertion operator tells TypeScript that VSCE _knows_ that this value is not null, when it could be.  
* String() is a node function that [coerces the value to a string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_primitives_and_string_objects) (`new String()` is the pattern that creates a String object and could cause comparison issues and should generally be avoided)
* Using String() in this scenario is an [established pattern in the VSCE code base](https://github.com/forcedotcom/salesforcedx-vscode/pull/4348#discussion_r938251211)

@jeffb-sfdc let me know if you'd like to discuss.  For these reasons I'd prefer `String()` over using the `!`.

gbockus-sf (approved):
QE result: 
Test Cases 

* Deploy
    *  an apex class ✅
    * locally changed class ✅
    * class with conflict ✅
        * override ✅
        * conflicts ✅
    * LWC component 
        * new ✅
        * existing ✅
    * manifest ✅
    * folder ✅
* Retrieve a new apex class
    * class✅
    * class with changes on server ✅
    * LWC component
        * org browser ✅
        * in explorer ✅
    * manifest ✅
    * folder ✅
    * From Org Browser ✅
* Run Apex Tests
    * single test ✅
    * Test Explorer ✅

I just deleted the cache for the windows unit tests and reran.  Very odd it was kind of hung up on whatever was in that cache.

package.json
```diff
@@ -33,6 +35,7 @@
     "eslint-config-prettier": "6.15.0",
     "husky": "1.3.1",
     "jest": "28.1.3",
+    "jest-junit": "14.0.1",
```
gbockus-sf, line 38: It's odd these came up in this PR. They are dev deps at the top level so it's fine, but I am curious if it was due to being on node v18 that triggered the necessity.

gbockus-sf (changes requested):
Hold for a bit.  I'm looking at the changes in the package-lock file and those seem out of sorts

package.json
klewis-sfdc: I'd be up for trying to remove them now and see if everything still compiles/looks ok

gbockus-sf (approved):

PR #4697, opened 2023-02-21
ci: update build/publish steps to use nvm for node version

gbockus-sf:
### What does this PR do?
Update the setup-node steps in our GHA configs to use the `.nvmrc` file to install the same version of node we use locally and define in package.json. 

@W-12573832@

### Functionality Before
GHA scripts failed due to node version mismatch

### Functionality After
GHA script succeed. 

### Notes
Verified the buildall flow and createReleaseBranch flow in my fork [here](https://github.com/gbockus-sf/salesforcedx-vscode/actions)

### DIFF
--- a/.github/workflows/buildAll.yml
+++ b/.github/workflows/buildAll.yml
@@ -12,10 +12,6 @@ on:
 
 jobs:
   build-all:
-    strategy:
-      matrix:
-        node_version: [lts/-1]
-      fail-fast: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


.github/workflows/buildAll.yml
```diff
@@ -12,18 +12,14 @@ on:
...
-      fail-fast: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.branch }}
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node_version }}
+          node-version-file: '.nvmrc'
```
gbockus-sf, line 22: matrix was not necessary since we only use the single version of node.

randi274 (approved):
Looks great! Noted that we are testing different versions of the extension with Linux and Windows builds specifically, but this solidifies more of our release and publishing process.

PR #4715, opened 2023-02-28
chore: fix unit tests

gbockus-sf:
### What does this PR do?
There was a single unit test that was regularly failing locally due to a timeout.  Fixed it locally by adding a longer timeout.
There were some unit tests that were failing locally due to the recent request-light update.  Fixed. 

Related to @W-9466545@

### Functionality Before
Unit tests failed local

### Functionality After
Unit tests pass. 

### Notes
Created 
https://github.com/forcedotcom/easywriter/issues/140 to track why a couple of the tests suites are exiting early in GHA which is why we didn't see the unit test failure in the visualforce package on GHA.

### DIFF
--- a/packages/salesforcedx-apex-debugger/test/unit/commands/baseDebuggerCommand.test.ts
+++ b/packages/salesforcedx-apex-debugger/test/unit/commands/baseDebuggerCommand.test.ts
@@ -32,7 +32,7 @@ export function getDefaultHeaders(contentLength: number): any {
     'Content-Type': 'application/json;charset=utf-8',
     Accept: 'application/json',
     Authorization: `OAuth 123`,
-    'Content-Length': contentLength,
+    'Content-Length': String(contentLength),
     'Sforce-Call-Options': `client=${CLIENT_ID}`
   };
 }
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-apex-debugger/test/unit/commands/baseDebuggerCommand.test.ts
```diff
@@ -32,7 +32,7 @@ export function getDefaultHeaders(contentLength: number): any {
     'Content-Type': 'application/json;charset=utf-8',
     Accept: 'application/json',
     Authorization: `OAuth 123`,
-    'Content-Length': contentLength,
+    'Content-Length': String(contentLength),
```
gbockus-sf, line 35: request-light bump altered the type here to be a string so this fixes all the tests that use this header.

packages/salesforcedx-visualforce-language-server/test/unit/javascriptMode.test.ts
```diff
@@ -46,7 +46,8 @@ describe('HTML Javascript Support', () => {
     }
   }
 
-  it('Should display completions', () => {
+  it('Should display completions', function() {
+    this.timeout(30000);
```
gbockus-sf, line 50: This was regularly timing out. I didn't dig at all on figuring out what it's waiting on locally but it now passes without a significant change in the run time.

randi274 (approved):
:shipit:

Issue #2952, opened 2021-02-16
Setting to change to pull/push operations in context menu for source-tracked Sandboxes


dschibster:
I am currently finding my way around source-tracking on Sandboxes, which I think is an amazing feature, however I am finding it hard currently to change the workflow that I (and many others) have gotten used to: Deploying / Pushing on Save and using the Command Palette for everything else.

If I am not mistaken, Auth with a Sandbox will always give you Deploy and Retrieve commands, both in the Command Palette and the context menu currently. Does not matter if the Sandbox is source-tracked or not. That means in order to get to the push/pull commands, I always have to go to the Terminal and type in what I want to do. That is definitely possible, will however hamper acceptance among colleagues. 

Is it feasible to detect a source-tracked Sandbox and enable push/pull accordingly in the command palette? If not that, maybe a setting that we could use on the workspace-level would help.

In the fear of repeating something everyone already knows: I have updated the CLI and checked everywhere I thought logical, didn't find anything regarding this issue however.

Thanks in advance.

smaddox-sf:
Hi @dschibster - Thanks for providing this feedback.  You are correct that push/pull is not currently exposed in the VS Code Extensions for source-tracked sandboxes.  The PM of source-tracked Sandboxes is considering exposing this and I will pass this feedback along to her for consideration.

jdgirald:
Hi @smaddox-sf, any updates on this. Thanks

JSaterdalen:
@smaddox-sf Would love to see this addressed, as it is actively causing issues in my source-tracked sandboxes. When the VS Code setting "Salesforcedx-vscode-core › Push-or-deploy-on-save" is enabled, VS Code incorrectly runs a source:deploy on save—instead of source:push—when saving a file against a source-tracked sandbox.

This messes with source tracking, due to source tracking [not supporting source:deploy and source:retrieve](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_source_tracking_best_practices.htm).

FabienTaillon:
Any update on this one ? What's the point of having to manually type the source:push/pull manually on a source tracked sandbox ?

jag-sfdc:
@FabienTaillon @JSaterdalen - We have created a bug to take care of this. Thank you so much for reporting this.

jeffb-sfdc:
@dschibster we are working at this at the moment.

ulistaerk:
[...]

PR #4727, opened 2023-03-08
feat: add pull/push for source-tracked Sandboxes

FabienTaillon:
### What does this PR do?
Add pull/push operations in command palette for Sandboxes with Source Tracking enabled.

#2952

### Functionality Before
In a source-tracked sandbox, push/pull commands were not available in the command palette:
- SFDX: Push Source to Default Scratch Org
- SFDX: Pull Source to Default Scratch Org
- SFDX: Push Source to Default Scratch Org and Override Conflicts
- ...

### Functionality After
All these commands that are source-tracking related become available for all orgs having source tracking enabled, not only Scratch Orgs.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
+++ b/packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
@@ -22,13 +22,15 @@ export async function getWorkspaceOrgType(
     throw e;
   }
   const username = await OrgAuthInfo.getUsername(defaultUsernameOrAlias);
-  const isScratchOrg = await OrgAuthInfo.isAScratchOrg(username).catch(err =>
+  const hasChangeTracking = await OrgAuthInfo.hasChangeTracking(
+    username
+  ).catch(err =>
     telemetryService.sendException(
-      'get_workspace_org_type_scratch_org',
+      'get_workspace_org_type_has_change_tracking',
       err.message
     )
   );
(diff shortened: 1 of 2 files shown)


FabienTaillon:
I submitted this PR as draft for now because there is one thing that remains to be done, but I would like to be sure you agree with it before doing it.

Today messages explicitly reference Scratch Orgs, like `force_source_pull_default_scratch_org_text` [here]( 
https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/messages/i18n.ts#L106).

I would suggest to rename them to `force_source_pull_default_org_text` (`SFDX: Pull Source to Default Org`), but as this needs to be changed in several places (including translations), I would like to be sure you're ok with that.

Other option would be to have 2 sets of messages (`SFDX: Pull Source to Default Scratch Org` and `SFDX: Pull Source to Default Sandbox`), but I don't see a need for that.

klewis-sfdc:
Hi @FabienTaillon - Thank you for this contribution!  I had a similar change within a PR that is underway already.  I pulled this specific change out and put it up in a new PR, with UI changes confirmed with our doc writer and unit tests here: #4755.  We will try to get this PR in soon so we can get this change in an upcoming release.  Thanks again for submitting this draft PR!

randi274:
Thanks again @FabienTaillon!

PR #4742, opened 2023-03-16
fix: set home as save path for unsaved query results

jgellin-sf:
### What does this PR do?
By default, SOQL query results are saved to the same directory as the .soql file from which the query is generated. However, when the SOQL query is not yet saved, the default directory was set to the root. This PR changes the behavior so that the default directory for unsaved query results is set to the home directory instead of the root.

https://github.com/forcedotcom/try-code-builder-feedback/issues/201
@W-12081773@

### Functionality Before
View query results of unsaved SOQL query. Save as CSV or JSON. The save dialog shows the root directory as the default save directory.

### Functionality After
View query results of unsaved SOQL query. Save as CSV or JSON. The save dialog shows the home directory as the default save directory.


floralan (approved):
This looks great to me! Good job! 👍

PR #4744, opened 2023-03-20
feat: Enable Org Browser for Scratch Orgs

klewis-sfdc:
### What does this PR do?
* PR #4597 enabled the SFDX: Retrieve* commands to work correctly for source-tracked orgs, which was a pre-requisite for exposing the Org Browser feature, since it relies on the retrieve operation to pull metadata components down into a local project.  
* This PR follows and exposes Org Browser for source-tracked orgs.  Enabling it for use when connected to either a Scratch org or a source-tracked Sandbox org. 

@W-10931259@

### Functionality Before
Org Browser would not be accessible when connected to a source-tracked org.

### Functionality After
Org Browser is accessible and metadata components can be retrieved using its download icons/buttons.


packages/salesforcedx-vscode-core/package.json
```diff
@@ -150,7 +150,7 @@
         {
           "id": "metadata",
           "name": "Metadata",
-          "when": "sfdx:project_opened && sfdx:default_username_has_no_change_tracking"
```
klewis-sfdc, line 153: Org Browser is the only consumer of this setting.  I would propose a follow up WI to remove this setting entirely, and update the consumer of the other like setting, 'sfdx:default_username_has_change_tracking' to use the newer `getWorkspaceOrgType()` function instead.

PR #4749, opened 2023-03-22
fix: update package.json files to comply with new parser

CristiCanizales:
### What does this PR do?
- Updates package.json files across the Salesforce Extensions to comply with new parser

@W-12645770@

### Functionality Before
- squigglies in package.json's

### Functionality After
<insert gif and/or summary>
- no squigglies in package.json's

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -140,11 +140,11 @@
       "view/item/context": [
         {
           "command": "sfdx.force.test.view.showError",
-          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)_Fail/"
+          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest||apexTestGroup)_Fail/"
         },
         {
           "command": "sfdx.force.test.view.goToDefinition",
-          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest|apexTestGroup)(_Pass|_Skip|\\b)/"
+          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest||apexTestGroup)(_Pass||_Skip||\\b)/"
         },
         {
           "command": "sfdx.force.test.view.runClassTests",
@@ -153,7 +153,7 @@
         },
         {
           "command": "sfdx.force.test.view.runSingleTest",
-          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest)(_.*|\\b)/",
+          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest)(_.*||\\b)/",
           "group": "inline"
         }
       ],
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -176,7 +176,7 @@
         },
         {
           "command": "sfdx.force.metadata.view.component.refresh",
-          "when": "view == metadata && viewItem =~ /(folder|type)/",
+          "when": "view == metadata && viewItem =~ /(folder||type)/",
           "group": "inline"
         },
         {
@@ -186,7 +186,7 @@
         },
         {
           "command": "sfdx.force.source.retrieve.component",
-          "when": "view == metadata && viewItem =~ /(type|component|folder)/",
+          "when": "view == metadata && viewItem =~ /(type||component||folder)/",
           "group": "inline"
         },
         {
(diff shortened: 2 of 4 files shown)


klewis-sfdc (commented):
Thanks for this work, Cristi! Looks great.  And thanks for sending me the [link to the PR](https://github.com/microsoft/vscode/issues/175540) that explains the new parser changes!  I've got one question about clearing out one more set of squigglies but otherwise this is looking great.

packages/salesforcedx-vscode-core/package.json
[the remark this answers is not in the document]
gbockus-sf: Probably got to hold on that request as we're currently targeting
```
  "engines": {
    "vscode": "^1.61.2"
  },```

File: packages/salesforcedx-vscode-apex/package.json [hunk 1]
```diff
@@ -153,7 +153,7 @@
         },
         {
           "command": "sfdx.force.test.view.runSingleTest",
-          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest)(_.*|\\b)/",
+          "when": "view == sfdx.force.test.view && viewItem =~ /(apexTest)(_.*||\\b)/",
```
gbockus-sf, line 156: dumb reg ex q.  Seems like they moved to js style regular expressions.  So did it just change for a|b  -> a or b  to a||b -> a or b?

CristiCanizales: it seems so! According to [this issue](https://github.com/microsoft/vscode/issues/175540) and [this page](https://code.visualstudio.com/api/references/when-clause-contexts)

packages/salesforcedx-vscode-core/package.json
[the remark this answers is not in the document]
CristiCanizales: I had seen those squigglies, and one more in another file, but didn't remove them because they were not red squigglies (? but Gordon's reason seems better🥲

packages/salesforcedx-vscode-apex/package.json
gbockus-sf: hum...interestingly the matches example on that page uses the single `|` 
Screenshot 2023-03-24 at 12 01 35 PM

CristiCanizales: Oh right... mm then the || would be for other cases?
image

ulugbekna: > dumb reg ex q. Seems like they moved to js style regular expressions. So did it just change for a|b -> a or b to a||b -> a or b?
> it seems so! According to https://github.com/microsoft/vscode/issues/175540 and [this page](https://code.visualstudio.com/api/references/when-clause-contexts)

Hmm, would you mind pointing which wording made you think so? Because we didn't change how we interpret regex patterns, it's just I probably had worded myself poorly.

Also, I think you don't need to change a when clause if there are no errors underlined.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -198,7 +198,7 @@
       "editor/context": [
         {
           "command": "sfdx.force.function.containerless.start",
-          "when": "sfdx:project_opened && resource =~ /.*/functions/.*/[^/]+(/[^/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
+          "when": "sfdx:project_opened && resource =~ /.*\\/functions\\/.*\\/[^\\/]+(\\/[^\\/]+\\.(ts||js||java||json||toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json"
```
ulugbekna, line 201: So here, you're right that `/` needs to be escaped within the regular expression and become `\\/`, but `|` remains same. I guess this error message was misleading: 

image

but it says what it says because of unescaped slashes. 

I think this should be the correct when clause in this case (just slashes escaped): 

```
sfdx:project_opened && resource =~ /.*\\/functions\\/.*\\/[^\\/]+(\\/[^\\/]+\\.(ts|js|java|json|toml))?$/ && resourceFilename != package.json && resourceFilename != package-lock.json && resourceFilename != tsconfig.json
```

CristiCanizales: Right! Thank you for taking the time to review this PR🙌🏼

randi274:
Remaining QE, given this is a dev-only impact:
- Build + Spot-check with VSIXes in 1.77+
- Build + Spot-check with VSIXes in 1.77-

randi274 (approved):
QA:
✅ - Builds locally, and also fixes the issue with the extra commands no longer appearing!
✅ - Extensions look good with VSIXes in 1.77.3
✅ - start function, lightning rename and lwc preview commands appear in proper context
✅ - Extensions look good with VSIXes in 1.75.1 (January build)
✅ - start function, lightning rename and lwc preview commands appear in proper context

PR #4761, opened 2023-03-30
fix: add ExperiencePropertyType to metadata dictionary

Kaaviyan:
### What does this PR do?
This PR adds ExperiencePropertyType to the metadata dictionary. Similar to the wave template, the ExperiencePropertyType metadata type doesn't have a -meta.xml instead it has a static schema.json file for each metadata. Adding an entry to the metadata dictionary to fix this issue

@W-12766917@

### Functionality Before
Retrieving ExperiencePropertyType throws an error and doesn't show the overwrite popup when metadata locally exists

### Functionality After
Retrieving ExperiencePropertyType shows overwrite popup warning if it exists locally and no error is thrown when retrieving this type


CristiCanizales (commented):
Hello! Looks good. Could you please add a jest test that covers your changes? 🙂

Kaaviyan:
Thank you for taking a look @CristiCanizales 
I've added some tests. Please let me know if it works :)

PR #4763, opened 2023-03-30
fix: added notes to the publishing doc on what to do when a release d…

jeffb-sfdc:
Add notes to the publishing doc on what to do when a release does not contain any changes

@W-12777981@


contributing/publishing.md
```diff
@@ -29,6 +29,12 @@ Creating a release branch automatically generates the change log based off of th
 
 One of the members of [Doc Maintainers](https://github.com/orgs/forcedotcom/teams/doc-maintainers/members) will review the changelog and make any changes to the release branch. A pull request will be opened against the release branch with updates to be included.
 
+## Compare Changes in the Release
+
+When verifying the release, verify that it contains changes. One can see the changes in GitHub using an URL to diff the changes between releases, with an URl in the format of https://github.com/forcedotcom/salesforcedx-vscode/compare/release/v57.7.0...release/v57.8.0.
+
+If no substantial changes were made the previous week, the the release will be skipped.
```
randi274, line 36:
```suggestion
If no changes were made the previous week, then the release can be skipped.
```

randi274: I was thinking substantial might be a bit subjective, and it would be good for us to release even small changes. Maybe omitting substantial altogether? Also, is there any additional action we should take for empty release, like deleting the branch, or just don't begin the pre-publish process?

randi274: (Actually probs not deleting the release, since I think our changelog may rely on finding that previous release?)

jeffb-sfdc: I changed it, but why "substantial" was there was for the case of when only dependabot updates (or GHA updates) were made during that week.  I changed it, but if there's anything better I'm open to changing it.

randi274 (approved):

contributing/publishing.md
randi274: Got it - I don't mind calling those out explicitly! Either that or scratching it works for me 👍

sbudhirajadoc (approved):

contributing/publishing.md
```diff
@@ -57,7 +63,7 @@ Before approving the release to the marketplace, download the vsix files from th
 
 Alternatively, you can download the files using the [gh cli](https://cli.github.com/) and then upload them all at once. Replace `v57.3.0` with the tag name for the release that you are testing, and to whatever download directory you would like. Additionally, `code` can be replaced by `code-insiders`.
 
-`> gh release download v57.3.0 --dir ~/Downloads/v57.3.0 --pattern '*.vscode'`  
+`> gh release download v57.3.0 --dir ~/Downloads/v57.3.0 --pattern '*.vscode'`
```
randi274, line 66:
```suggestion
`> gh release download v57.3.0 --dir ~/Downloads/v57.3.0 --pattern '*.vsix'`
```

randi274: @jeffb-sfdc I noticed that I had the wrong pattern here during my release testing!

PR #4793, opened 2023-04-11
fix: remove octokit plugin

CristiCanizales:
### What does this PR do?
- Remove `@octokit/plugin-paginate-rest` dep from salesforcedx-vscode-core package.json

@W-11410543@


gbockus-sf (approved):
When this was put in place it was due to vsix generation failing so as long as we get the build-all passing we're good to go 👍 TY

PR #4807, opened 2023-04-17
fix: update lls to fix core devs

randi274:
### What does this PR do?
Fixes an issue where internal developers weren't able to use the Lightning Language Server with the BLT setup.

@W-13007946@. Includes forcedotcom/lightning-language-server#564.

### Functionality Before
The LWC Language Server would show errors if BLT was not included in Core developer setup. 

### Functionality After
The Language Server checks for a configuration setup at a different location that is more commonly used, and doesn't require the use of the core config if it can't be located.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6543,11 +6543,11 @@
       }
     },
     "node_modules/@salesforce/aura-language-server": {
-      "version": "4.4.0",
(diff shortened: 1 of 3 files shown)


randi274:
QA Complete - the language servers are up and running, with standard functionality available.

gbockus-sf (approved):

PR #4773, opened 2023-04-04
fix: Retrieve* Commands Should Update Source Tracking When Used Against Source-tracked Sandboxes

klewis-sfdc:
### What does this PR do?
* Fixes the SFDX: Retrieve* commands so they update source-tracking when used against a source-tracked Sandbox.

### What issues does this PR fix or reference?
@W-12971145@

### Functionality Before
When connected to a source-tracked Sandbox, SFDX: Retrieve* commands would retrieve metadata components, but would not fix up source tracking correctly.

### Functionality After
When connected to a source-tracked Sandbox, SFDX: Retrieve* commands retrieve metadata components and fix up source tracking correctly.

Todos:
- [ ] Confirm SFDX: Retrieve Source in Manifest from Org works as expected

PR #4808, opened 2023-04-17
chore: bump core lib

klewis-sfdc:
### What does this PR do?
* Bumps core across the packages so that it is at latest and compiles across the vsce monorepo
* This PR is a requirement for #4773 

@W-12971145@

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6636,9 +6636,9 @@
       }
     },
     "node_modules/@salesforce/core": {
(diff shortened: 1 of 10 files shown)


randi274 (approved):
Installs, compiles, and runs all the extensions - also spot checked some functionality 👍

randi274:
Approved overriding Snyk test for this run.

PR #4809, opened 2023-04-18
feat: Expose Deploy* and Retrieve* Commands for Scratch Orgs

klewis-sfdc:
### What does this PR do?
* Exposes the SFDX: Deploy* and SFDX: Retrieve* commands when connected to a scratch org.

### What issues does this PR fix or reference?
@W-11309172@

### Functionality Before
* The SFDX: Deploy* and SFDX: Retrieve* commands were unavailable when connected to a scratch org.

### Functionality After
* The SFDX: Deploy* and SFDX: Retrieve* commands are available when connected to a scratch org.
* The commands handle source tracking correctly and fix it up appropriately after deploy/retrieve operations

PR #4810, opened 2023-04-18
feat: Expose Org Browser for Scratch Orgs

klewis-sfdc:
### What does this PR do?
* Enables the Org Browser feature to be used with scratch orgs.

@W-10931259@

### Functionality Before
* Org Browser was not available when connected to a scratch org.

### Functionality After
* Org Browser is available when connected to a scratch org.
* Retrieving from within Org Browser fixes up source-tracking correctly when connected to a scratch org.


randi274 (commented):
I noticed while reviewing #4809 that this is the last hanger-on we have around is_scratch_org and default_username_has_no_change_tracking. Can we remove them here? 🤞 Or make a follow-up WI to remove if we have any timing concerns?

randi274 (commented):
This has become an exciting cleanup PR!

QA is started, and the basics are looking good:
- Org Browser appears for scratch orgs
- Can retrieve singular file
- Refresh works 
- Can retrieve whole directories

Retesting today, and source tracking seems to be having issues for changes in remote. We'll track that issue separately.

Also regarding changes to `workspaceOrgType.ts`:
- ✅  Org Browser shows “No Default Org Set” if I log out
- ✅ Same if I refresh

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
```diff
@@ -25,47 +24,28 @@ export enum OrgType {
...
   const workspaceContext = WorkspaceContext.getInstance();
-  const connection = await workspaceContext.getConnection();
+  let connection;
+  try {
+    connection = await workspaceContext.getConnection();
+  } catch (error) {
+    console.warn(
+      `An error was encountered while getting the org connection:\n ${error}`
+    );
+    return OrgType.NonSourceTracked;
```
randi274, line 34: If we can't return the org connection, why do we assume we have a NonSourceTracked org?

randi274 (approved):

packages/salesforcedx-vscode-core/src/context/workspaceOrgType.ts
klewis-sfdc: Its basically a default value for this function to return.  When set to NonSrcTracked, commands like view changes and push/pull are hidden.

PR #4811, opened 2023-04-18
ci: add e2e testing workflow

randi274:
### What does this PR do?
Introduces a manual workflow run for our end to end testing framework that checks out any branch in our main VS Code repository and can also be run against any branch in our automation testing repository.

Here is a default test run: https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4734122783/jobs/8402595922
Here is a test run where both the VS Code branch and the Automation Tests branch have been specified: https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4734367144/jobs/8403184051

@W-12392382@

### Functionality Before
Our end to end tests could only be run locally.

### Functionality After
They can now be run in Github 🎉  (Caveat: some additional work will be required to get them happy and passing!)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+            cache: npm
+            cache-dependency-path: |
+              salesforcedx-vscode/package-lock.json
+              salesforcedx-vscode-automation-tests/package-lock.json
+        - name: Clone automation tests
+          uses: actions/checkout@v2
+          with:
+            repository: forcedotcom/salesforcedx-vscode-automation-tests
+            path: salesforcedx-vscode-automation-tests
+            ref: ${{ inputs.automationBranch }}
```
randi274, line 37: Fun fact - this used to be more verbose, as seen in our other workflows. ${{ github.event.inputs.automationBranch }} but according to documentation (and confirmed with running here) we can now omit the first two keys: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_callinputs

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+        os: [macos-latest]
+        vscodeVersion:
+          - stable
+    runs-on: ${{ matrix.os }}
+    steps:
+        - name: Checkout
+          uses: actions/checkout@v2
+          with:
+            path: ./salesforcedx-vscode
+            ref: ${{ github.event.ref }}
```
randi274, line 24: This might be shorter too now, but since it's not all that verbose compared to the inputs I don't hate it as is.

gbockus-sf: might not even need it as this might be the default value.  also don't hate just having it.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+        - uses: actions/setup-node@v3
+          with:
+            node-version-file: 'salesforcedx-vscode/.nvmrc' 
+            cache: npm
+            cache-dependency-path: |
+              salesforcedx-vscode/package-lock.json
+              salesforcedx-vscode-automation-tests/package-lock.json
+        - name: Clone automation tests
+          uses: actions/checkout@v2
```
gbockus-sf, line 33: did you purposely pick v2 over v3 here?

gbockus-sf: same on line 21

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+            SFDX_CLI_VERSION=$(sfdx version)
+            if [[ ((`echo $SFDX_CLI_VERSION | grep -c "sfdx-cli/"` > 0))]]
+            then
+              echo "sfdx-cli installed -" $SFDX_CLI_VERSION
+            else
+              echo "The sfdx-cli installation could not be verified"
+              exit 1
+            fi
+        - name: Run headless test
+          uses: GabrielBB/xvfb-action@v1
```
gbockus-sf, line 62: did we ever get this 3pp approved?

gbockus-sf (commented):
Great stuff!

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+          working-directory: salesforcedx-vscode-automation-tests
+        - name: Install the SFDX CLI
+          run: npm install -g sfdx-cli
+        - name: Verify CLI
+          shell: bash
+          run: |
+            set -e
+            sfdx version
+            SFDX_CLI_VERSION=$(sfdx version)
+            if [[ ((`echo $SFDX_CLI_VERSION | grep -c "sfdx-cli/"` > 0))]]
```
jeffb-sfdc, line 54: I'm not sure line 54 will work on Windows

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+              echo "The sfdx-cli installation could not be verified"
+              exit 1
+            fi
+        - name: Run headless test
+          uses: GabrielBB/xvfb-action@v1
+          with:
+            run: npm run automation-tests
+            working-directory: salesforcedx-vscode-automation-tests
+          env:
+            VSCODE_VERSION: ${{ matrix.vscodeVersion }}
```
jeffb-sfdc, line 67: What uses `VSCODE_VERSION`?

.github/workflows/e2e.yml
randi274: not yet - we can bind to a specific commit hash, or try and run it without?

.github/workflows/e2e.yml
randi274: yes, I think I ran into errors at some point... I'm not sure if they were resolved eventually, but I could play around with that

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+        default: 'develop'
+        type: string
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [macos-latest]
+        vscodeVersion:
+          - stable
```
jeffb-sfdc, line 17: should the node version in the matrix?

.github/workflows/e2e.yml
randi274: Since we're not using this now, I'll remove it. This is the way that we can access `process.env.VAR` within the automation-tests.

.github/workflows/e2e.yml
randi274: So once we get the workflow merged in this is what will let us run against specific branches, so this one is useful for us. Right now that's all manual, but eventually we could be have it be kicked off from a pull request or another event.

.github/workflows/e2e.yml
randi274: Good catch - this one I grabbed from the CLI, but they were using for Docker builds specifically, which was running on ubuntu. For now it's fine if we're only running against mac, but we could take the time to change it now, or kick the can to fix this once we start working on Windows. I also have a vague memory of this working when we did the initial windows tests, but re-verification/testing would definitely be necessary to confirm.

.github/workflows/e2e.yml
randi274: and by "without" I mean I've seen some ugly code that lets us run tests in a headless chrome environment, without using an action specifically :)

.github/workflows/e2e.yml
randi274: Test run to check: https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4756581716/workflow

.github/workflows/e2e.yml
randi274: ohhh better yet, that repo actually recommends using the Open Source Community Action now umbrella'd under Github: https://github.com/coactions + https://github.com/coactions/setup-xvfb . I'll still pin at a commit hash, but this seems a better bet?

.github/workflows/e2e.yml
randi274: Looks like v3 was good, so I'll update for latest 👍

.github/workflows/e2e.yml
jeffb-sfdc: Actually, I'm going to go add it to EnvironmentSettigs.ts

randi274: Update: @jeffb-sfdc is making some changes to factor in the VSCODE_VERSION in the event we would like to use it going forward, so it will stay.

.github/workflows/e2e.yml
randi274: Looks like the new action works too: https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4756852368/workflow

.github/workflows/e2e.yml
randi274: Also it looks like this does work on Windows as well 😎  https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4788500534/jobs/8515192099

.github/workflows/e2e.yml
randi274: Our testing here is pretty limited at the moment, so I've manually set the node version here to use in the matrix so we can test against different node versions. When we want multiple versions, we can make an update to use the .npmrc value here as well.

.github/workflows/e2e.yml
randi274: Confirmed this syntax looks ok for us: https://github.com/randi274/salesforcedx-vscode-fork/actions/runs/4800025121/jobs/8540419374

jeffb-sfdc (approved):

Issue #4775, opened 2023-04-05
Getting "This directory does not contain a valid Salesforce DX project." when "Deploy This Source to Org"


wrone:
### Summary
I'm not able to deploy classes from VSC using the "Deploy This Source to Org" button. I'm getting the below error:
![image](https://user-images.githubusercontent.com/9321029/230008741-ee8ea6fc-f316-402a-a135-e725c867ce20.png)

Yesterday this error started to occur, I created a new VSC project and in the new project it was working. But today it is not working in the new project anymore.

My project is created with **manifest**, not scratch org.

In the same time it is working when I'm using a terminal command to deploy:
![image](https://user-images.githubusercontent.com/9321029/230009568-5f5cf5dc-6254-4c38-84ee-be104aed2cc3.png)

### Steps To Reproduce:

**Repository to reproduce:** 
1.  Right click on the class and the click the "SFDX: Deploy This Source to Org" button

### Expected result
The class is deployed successully

### Actual result
Got the "Component conversion failed: This directory does not contain a valid Salesforce DX project." error

### System Information
Terminal:
bash

Version:
{
  "cliVersion": "sfdx-cli/7.194.1",
  "architecture": "win32-x64",
  "nodeVersion": "node-v18.15.0",
  "osVersion": "Windows_NT 10.0.19045",
  "shell": "cmd.exe",
  "rootPath": "...\\AppData\\Local\\sfdx\\client\\7.194.1-faa045e",
  "pluginVersions": [
    "@copado/copado-cli 1.1.19 (user)",
    "@oclif/plugin-autocomplete 2.1.6 (core)",
    "@oclif/plugin-commands 2.2.11 (core)",
    "@oclif/plugin-help 5.2.8 (core)",
    "@oclif/plugin-not-found 2.3.22 (core)",
    "@oclif/plugin-plugins 2.4.3 (core)",
    "@oclif/plugin-search 0.0.14 (core)",
    "@oclif/plugin-update 3.1.8 (core)",
    "@oclif/plugin-version 1.3.1 (core)",
    "@oclif/plugin-warn-if-update-available 2.0.32 (core)",
    "@oclif/plugin-which 2.2.17 (core)",
    "apex 2.2.7 (core)",
    "auth 2.7.10 (core)",
    "community 2.2.8 (core)",
    "custom-metadata 2.1.9 (core)",
    "data 2.3.8 (core)",
    "info 2.6.2 (core)",
[...]

PR #4816, opened 2023-04-20
fix: set project directory for component set

CristiCanizales:
#4775>


randi274 (changes requested):
In chatting with Ken, it sounds like this didn't solve our component set errors, but this is probably still a good thing to consider merging in.

One ask though - unit tests? :)

CristiCanizales:
Ugh I had the test local but as we talked this wasn't the fix I deleted it and now can't ctrl+z🥲 I'll rethink it and push later today

Issue #4571, opened 2022-11-28
User mode syntax not supported by language server


jonathanwiesel:
### Summary

Since the [recent introduction of User Mode beta](https://developer.salesforce.com/docs/atlas.en-us.240.0.apexcode.meta/apexcode/apex_classes_enforce_usermode.htm) in Apex it seems that the `as user` syntax is not supported since it shows a `Unexpected token 'as'.` error. 

Deploying the class does compile in the server.

### Steps To Reproduce:

1.  Create an Apex class with v56.0.
2. Add the new DML statement like the one in the docs:

```apex
Account acc = new Account(Name='test');
insert as user acc;
```

### Expected result

It should show no error and probably highligh `as user` in a different color similarly of how `WITH SECURITY_ENFORCED` or `WITH USER_MODE` is shown

### Actual result

The following local error is shown:

```
`Unexpected token 'as'.`
```

Which also generates other subsequent errors because of not parsing correctly the statement.

### Additional information

![Screenshot 2022-11-28 at 15 00 25](https://user-images.githubusercontent.com/1326781/204296177-ac816664-ab7c-4c7b-872b-cfcd48e7176d.png)

**Salesforce Extension Version in VS Code**: v56.5.1

**SFDX CLI Version**: 7.178.0

**OS and version**: macOS 12.6.1

PR #4817, opened 2023-04-20
feat: update lsp jar

### LINKED ISSUE #4586: New Assert class unresolved
### Summary
The new Assert class (https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_class_System_Assert.htm) always comes up as Unresolved.

### Steps To Reproduce:

1.  Create a unit test class in VSCode, API version 56.
2.  Create a test method, and in it, try typing `Assert.` and note that code completion is not functional
3. Try typing `Assert.areEqual(1,1);` and then mouse-over `Assert`. It comes up `!Unresolved!`
 
### Expected result

Code completion should function for the new `Assert` class.

### Actual result

Code completion does not function for the new `Assert` class.

**Salesforce Extension Version in VS Code**: v56.5.1

**SFDX CLI Version**: 7.177.1

**OS and version**: Windows 11

RitamAgrawal:
### What does this PR do?
Updates the Apex Language Server with updates for our current Apex compiler (updating features for built-in classes like auto-complete and hover), and adds the `Debug` Codelens to Anonymous Apex files so that users can both Execute and Debug Anonymous Apex.

@W-12593119@, @W-12432185@, @W-11727631@. Fixes #4586, #4571, dangmai/prettier-plugin-apex#738. Should unblock trailheadapps/visualforce-to-lwc#290.

### Functionality Before
The Apex LSP was showing errors for newly valid syntax like `insert as user` and the `Assert` class. Anonymous Apex only had a codelens available to Execute code.

### Functionality After
That syntax is shown correctly. We have updated our `Execute Anonymous Apex` to just read as `Execute`, and added `Debug` which will kick of the Apex Replay Debugger. (Breakpoints must still be set within `.cls` files, not `.apex` files).

### DIFF
--- a/SHA256.md
+++ b/SHA256.md
@@ -8,7 +8,7 @@ make sure that their SHA values match the values in the list below.
    following the instructions at
    https://code.visualstudio.com/docs/editor/extension-gallery#_common-questions.
    For example, download,
-   https://salesforce.gallery.vsassets.io/_apis/public/gallery/publisher/salesforce/extension/salesforcedx-vscode-core/57.7.0/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage.
+   https://salesforce.gallery.vsassets.io/_apis/public/gallery/publisher/salesforce/extension/salesforcedx-vscode-core/57.10.2/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage.
 
 2. From a terminal, run:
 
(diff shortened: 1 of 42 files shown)


packages/salesforcedx-vscode-apex/out/apex-jorje-lsp.jar
randi274, line 1: Surprising that this ended up being smaller than before, but it has also been a long time since we've been able to update.

gbockus-sf: an 89% decrease 👀

RitamAgrawal: This increased after with the latest commit. The size was decreased in the first commit because I didn't pull the master in jorje.

randi274: 89% isn't that much right 😅 I'm going to resolve to see if we can see the updated file size.

randi274: ahhh the UI here isn't great. It's 89% of the size it previously was. So - 

Main Size: 15.3 MB
Updated Size: 13.6 MB

RitamAgrawal:
Empty commit didn't fix the missing cla error. I don't know why it's showing missing cla suddenly.

RitamAgrawal:
Raised a new PR https://github.com/forcedotcom/salesforcedx-vscode/pull/4819 to eliminate missing cla error.

Issue #4813, opened 2023-04-19
Unable to deploy on save (only push, but I don't want that)


JimR-FAN:
### Summary

My sandbox, refreshed from Prod, somehow has source-tracking turned on.   Now when I use "Deploy (or Push) On Save",  only the Push command runs -- I should be able to have Deploy run instead

### Steps To Reproduce:

1.  My developer sandbox was refreshed from Prod.   Someone turned on source-tracking on Prod.  Now my sandbox has source-tracking turned on. 
1.  Now when I save a file in VSCode, the  force:source:push command is run instead of  the force:source:deploy command.  I want the saved file to go through  force:source:deploy --  *not* push.     Even though the org has source-tracking turned on, I do not develop this way.   I like it how it used to be - that I can deploy a single file without relying on the source-tracking feature, which I am not using for this org-development project. 

**THANK YOU :-)**

### Expected result

I should be able to control whether or not  force:source:push  is used instead of  force:source:deploy.   If anything,  force:source:push should be used only on DX projects,  where my project is and always has been an org-development project.

### Actual result

When I save,  the IDE automatically runs   force:source:push   instead of force:source:deploy like it did before my last org refresh.

### Additional information

**Salesforce Extension Version in VS Code**:
v57.10.2

**SFDX CLI Version**:
v7.196.9

**OS and version**:
Windows + Chrome (lastest)

gian-ribo:
Same Issue!

IttesITNick:
Same issue since update of the Salesforce Extensions in VScode since yesterday.

gian-ribo:
No problem with v57.7.0
I forced downgrade and now i can work as always. I hope for bugfix soon

CristiCanizales:
Hello @JimR-FAN, @gian-ribo, @IttesITNick! We're sorry for that, and wanted to let you know that there is a fix in progress for this issue and we'll let you know once it's available!

IttesITNick:
@JimR-FAN, @gian-ribo We found another workaround without downgrading the CLI.
- In setup > Dev Hub
- Disable the setting "Enable Source Tracking in Developer and Developer Pro Sandboxes"
- Then refresh your sandbox and authorize VScode again with the new sandbox.
Now on save we get deploy instead of push again.

JimR-FAN:
Thanks @IttesITNick  - in my case I do not control Prod and am unable to refresh my sandbox.    Thank you!

AnanyaJha:
[...]

PR #4820, opened 2023-04-21
feat: Add "Prefer deploy on save" setting

klewis-sfdc:
### What does this PR do?
* Adds a setting that will run a **deploy** instead of a **push** when "Push or deploy on save" is enabled.

#4813, @W-13083377@

### Functionality Before
* When "Push or deploy on save" is enabled, **push** is run on save for source-tracked orgs.

### Functionality After
* When 
  * "Push or deploy on save" is **enabled** and 
  * "Prefer deploy on save" is **disabled**, 
    * **push** is run on save for source-tracked orgs.
    
* When 
  * "Push or deploy on save" is **enabled** and 
  * "Prefer deploy on save" is **enabled**, 
    * **deploy** is run.


CristiCanizales (approved):
Looks good to me!
I verified it:
- pushes if the preference is not enabled ✅ 
- deploys if the preference is enabled ✅ 
Thanks for this Ken! 😄

PR #4827, opened 2023-04-28
Break up files in forceSourceRetrieveMetadata

jeffb-sfdc:
### What functionality has changed?
Absolutely nothing.
No code was modified, it was only moved.

@W-10273533@


klewis-sfdc (commented):
@jeffb-sfdc this looks awesome.  

It looks like the stuff that used to be in the "describers" module is also used by Org Browser.  What do you think about moving that stuff to the util directory since it is shared?

jeffb-sfdc:
@klewis-sfdc this changes in this PR are only for file in packages/salesforcedx-vscode-core/src/commands/forceSourceRetrieveMetadata

The changes here are only for moving functions and classes into their own files, and other than updating paths (for module imports) no other changes were made.  Additional changes (like code clean up, removing duplicate code) would increase the scope of this PR and the number of files that were changes, and it make it harder to review.

After this PR gets in, I plan to do the same refactoring in `/packages/salesforcedx-vscode-core/src/commands/util`, and then (finally) in `/packages/salesforcedx-vscode-core/src/commands`

Any code modification (like removing duplicate code) would be in a future PR.

klewis-sfdc (approved):
Thanks, Jeff!  Looking forward to the next refactors also!

klewis-sfdc:
I also built the branch and spot checked some retrieve commands -
* retrieve on classes directory ✅ 
* retrieve a manifest ✅ 
* retrieve this source from org ✅

PR #4839, opened 2023-05-04
fix: Only Update Source Tracking After Retrieve When Org Type Is Source-Tracked

klewis-sfdc:
### What does this PR do?
* only makes the call to `sourceTracking.updateTrackingFromRetrieve()` when connected to a source-tracked org

@W-13153391@

### Functionality Before
**NOTE: this behavior is currently only observed on the develop branch, and is not in released production VSCE.**
* If `sourceTracking.updateTrackingFromRetrieve()` is called when connected to a non source tracked org, it throws an error.
  * VSCE was not handling the error so it gets surfaced to users:

  

### Functionality After
* VSCE does not call `sourceTracking.updateTrackingFromRetrieve()` when connected to a non source tracked org, avoiding the error.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -30,6 +30,7 @@ import { channelService, OUTPUT_CHANNEL } from '../channels';
 import { PersistentStorageService } from '../conflict/persistentStorageService';
 import { TELEMETRY_METADATA_COUNT } from '../constants';
 import { WorkspaceContext } from '../context';
+import { getWorkspaceOrgType, OrgType } from '../context/workspaceOrgType';
 import { handleDeployDiagnostics } from '../diagnostics';
 import { nls } from '../messages';
 import { setApiVersionOn } from '../services/sdr/componentSetUtils';
@@ -232,10 +233,13 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
     this.setupCancellation(operation, token);
 
     const result: RetrieveResult = await operation.pollStatus();
-    await SourceTrackingService.updateSourceTrackingAfterRetrieve(
-      sourceTracking,
-      result
-    );
+    const orgType = await getWorkspaceOrgType();
+    if (orgType === OrgType.SourceTracked) {
+      await SourceTrackingService.updateSourceTrackingAfterRetrieve(
+        sourceTracking,
+        result
+      );
+    }
 
     return result;
   }


packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -232,10 +233,13 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
...
-      sourceTracking,
-      result
-    );
+    const orgType = await getWorkspaceOrgType();
+    if (orgType === OrgType.SourceTracked) {
+      await SourceTrackingService.updateSourceTrackingAfterRetrieve(
+        sourceTracking,
+        result
+      );
+    }
```
gbockus-sf, line 242: Does this cover the conflict detection retrieve?

klewis-sfdc: Hmm.  VSCE doesn't have conflict detection on retrieve.  Our setting doesn't apply and STL is configured with ignoreConflicts: true.  But, I should probably double check here that the retrieve was successful.

randi274 (approved):
Spot checking for scratch orgs & source-tracked sandboxes appear that source tracking is still being cleaned up correctly, and I no longer see the error message when retrieving from my dev org (conflict detection seems to be working as expected there too). Let's chat if there's thorough QA you'd like to see done on this one!

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
gbockus-sf: cool..yeah I was having to knock some rust off about how that worked exactly.  Another use case to verify here is delete as I saw the retrieve error in that flow on release testing 👍

klewis-sfdc: I updated it to only make the call on a successful retrieve operation.  I did some spot testing with deleting files and trying to retrieve them and such from a non-src-tracked org and didn't see any errors.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -232,10 +239,19 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
...
-    await SourceTrackingService.updateSourceTrackingAfterRetrieve(
-      sourceTracking,
-      result
-    );
+
+    if (
+      result &&
+      result.response &&
+      (result.response.status === 'Succeeded' ||
+        result.response.status === 'SucceededPartial') &&
```
gbockus-sf, line 247: nit: you could use [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) here 
```
if  ((result?.response?.status === 'Succeeded' || result?.response?.status === 'SucceededPartial') && this.sourceTracking) {
```
or slightly cleaner 
```
const status = result?.response?.status;
if  ((status === 'Succeeded' || status === 'SucceededPartial') && this.sourceTracking) {
```

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -24,12 +24,14 @@ import {
   ComponentStatus,
   RequestStatus
 } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
+import { SourceTracking } from '@salesforce/source-tracking';
```
gbockus-sf, line 27: Do we already import @salesforce/source-tracking to the core module? I took a gander and don't think we do. 
Two thing should probably happen here. 
1. we should just export SourceTracking from the SourceTrackingService to isolate the dependency into utils. 
2. We should add @salesforce/source-tracking to the packageUpdates.dependancies in package.json for core so if something does fail there we don't get the message blow up error

packages/salesforcedx-vscode-core/test/jest/commands/retrieveExecutor.test.ts
```diff
@@ -97,4 +105,49 @@ describe('Retrieve Executor', () => {
...
+
+    // Act
+    await (executor as any).doOperation(dummyComponentSet, {});
+
+    // Assert
+    expect(workspaceContextGetInstanceSpy).toHaveBeenCalled();
+    expect(createSourceTrackingSpy).not.toHaveBeenCalled();
+    expect(retrieveSpy).toHaveBeenCalled();
+    expect(pollStatusMock).toHaveBeenCalled();
+    expect(updateTrackingAfterRetrieveMock).not.toHaveBeenCalled();
```
gbockus-sf, line 151: nice unit tests 👍

gbockus-sf (approved):
🚢

PR #4855, opened 2023-05-12
chore: Github Action for publishing extensions in Open VSX

floralan:
### What does this PR do?
This PR brings a new GitHub Action to automate publishing extensions in OpenVSX. 

This PR also updates the `publishing.md` file to add instructions for publishing in OpenVSX. 

@W-12222167@

### Functionality Before
Manual work is needed for publishing extensions in OpenVSX

### Functionality After
The publishing in OpenVSX Github Actions runs after the release job is finished. 
It will also send a Slack notification to ask for approval. 
The publishing process in OpenVSX is almost the same as publishing on MarketPlace.


.github/workflows/publishOpenVSX.yml
```diff
@@ -0,0 +1,94 @@
+name: Publish in Open VSX Registry
```
floralan, line 1: The reason why I prefer creating a seperate GHA file over adding a step in the current `publishVSCode.yml` is that:

If the publishing in the OpenVSX step failed, the whole `publishing` section will fail. However, the `ctcCloseSuccess` section needs `publishing` to be successful to move on. 

Alternatively, if we have a separate section, `publishing in Open VSX` is in the `publishVSCode.yml` file. When it fails, we can config it only to be re-run manually not to interrupt the publishing marketplace job. However, this still can't solve the `ctcCloseSuccess` dependency. When only the `publishing in Open VSX` section is run, the `ctcCloseSuccess` can't be finished because of the absence of the `publishing` section. 

Please advise if you think it can be added in the `publishVSCode.yml` file and if it's a better way to do this.

package.json
```diff
@@ -49,7 +49,8 @@
     "ts-node": "10.9.1",
     "tslint": "5.5.0",
     "typescript": "^4.7.4",
-    "vsce": "2.15.0"
+    "vsce": "2.15.0",
+    "ovsx": "latest"
```
floralan, line 53: Not sure if this is required to add to this file. ovsx is used in the GHA where we publish extensions in oepnvsx: "npx ovsx publish". npx will automatically get the latest version of ovsx for each execution.

.github/workflows/publishOpenVSX.yml
```diff
@@ -0,0 +1,94 @@
...
+  ctc-open:
+    needs: [ prepare-environment-from-main ]
+    uses: salesforcecli/github-workflows/.github/workflows/ctcOpen.yml@main
+    secrets: inherit
+
+  publish:
+    needs: [ 'ctc-open', 'prepare-environment-from-main' ]    
+    runs-on: ubuntu-latest
+    env: 
+      OVSX_PAT: ${{ secrets.OVSX_PAT }}
```
floralan, line 57: I suppose this secret `OVSX_PAT` already exists, but I don't have access to verify that. I could use the reviewer's help to confirm this.

floralan:
Checks failed because of the missing these packages in the local file. 
```
npm ERR! Missing: ovsx@0.8.1 from lock file
npm ERR! Missing: commander@6.2.1 from lock file
npm ERR! Missing: semver@5.7.1 from lock file
npm ERR! Missing: tmp@0.2.1 from lock file
```
Not sure if we want to add them all in this PR or if there is a way to work around it.

floralan:
Another question/concern is that we will need to let the on-call person know to approve this GHA on Slack. I would expect there is an internal document to guide the on-call person for every week's release. I would like to use one team member's help to add one step in the doc for this publishing GHA.

CristiCanizales (approved):
This looks good to me! I also think it's better to have a different file for the workflow. Thanks Flora! 😄

PR #4879, opened 2023-05-23
feat: expose TelemetryService class in core API

lahernandezb:
Service will be used to report back telemetry data from within the GPT extension.

### What does this PR do?
Exposes the `TelemetryService` in the core API.

@W-13218122@

### Functionality Before
Only the telemetry service which had been instantiated in the core extension was exposed via the API

### Functionality After
We have access to the TelemetryService class to instantiate and report from within the GPT extension.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -9,6 +9,7 @@ import {
   getRootWorkspacePath,
   SFDX_CORE_CONFIGURATION_NAME
 } from '@salesforce/salesforcedx-utils-vscode';
+import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode';
 import * as fs from 'fs';
 import * as vscode from 'vscode';
 import { channelService } from './channels';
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -672,7 +673,8 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
     SfdxWorkspaceChecker,
     WorkspaceContext,
     taskViewService,
-    telemetryService
+    telemetryService,
+    TelemetryService
```
lahernandezb, line 677: Only concern is whether or not the capitalization distinction is enough to differentiate it from the core instance.

gbockus-sf: Yeah I think it's fine as is.  We could also put it under a property like 
```services: {
  TelemetryService
  // Service2,
  // Service3, 
}, ```

I don't hate the idea of having a separated group of things we had to add to support an external extension. 

either way...good to go.

gbockus-sf (approved):
Cool.  Normally I'd harass about unit test coverage, but this activation method the worse for pulling in all the things so we're good to go.

PR #4888, opened 2023-05-26
chore: add ChannelService to the core extension API

gbockus-sf:
### What does this PR do?
Add the ChannelService to the core extension API for use in the EGPT extension. 

@W-13216290@

### Functionality Before
Only instance of channelService in API

### Functionality After
Base ChannelService class is available in core API


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -6,11 +6,11 @@
  */
 import { ensureCurrentWorkingDirIsProjectPath } from '@salesforce/salesforcedx-utils';
 import {
+  ChannelService,
   getRootWorkspacePath,
-  SFDX_CORE_CONFIGURATION_NAME
+  SFDX_CORE_CONFIGURATION_NAME,
+  TelemetryService
```
gbockus-sf, line 12: Just cleaned up two imports from utils-vscode.

lahernandezb (approved):

PR #4895, opened 2023-05-30
chore: need this change on default branch

peternhale:
@W-13190794@

### What does this PR do?
sets up a e2e test that uses yourkit profiling

POC -


.github/workflows/e2ePerf.yml
```diff
@@ -0,0 +1,90 @@
...
+            salesforcedx-vscode-automation-tests/package-lock.json
+      - name: Clone automation tests
+        uses: actions/checkout@v3
+        with:
+          repository: forcedotcom/salesforcedx-vscode-automation-tests
+          path: salesforcedx-vscode-automation-tests
+          ref: ${{ inputs.automationBranch }}
+      - name: Install Extension Dependencies
+        run: |
+          npm install
```
CristiCanizales, line 53: I think we also need to run npm run compile here ?

randi274: Yes - that was a missing step for us previously!

randi274 (changes requested):
@peternhale looks pretty straightforward! We do like to test the workflows first - can you create a fork for VS Code to run it, and then add a sample action run to the description? The PR can be opened from the fork or we can keep this one open, as long as the files are identical.

Example PR: https://github.com/forcedotcom/salesforcedx-vscode/pull/4811

.github/workflows/e2ePerf.yml
```diff
@@ -0,0 +1,90 @@
+name: End to End Tests (Performance)
+on:
+  workflow_dispatch:
+    inputs:
+      automationBranch:
+        description: 'Set the branch to use for automation tests'
+        required: false
+        default: 'phale/perf-poc'
```
randi274, line 8: And then this should probably be 'develop'

PR #4903, opened 2023-06-01
fix: re-run pr validation

RitamAgrawal:
### What does this PR do?
Runs the validate PR workflow when just the PR description is edited.
https://github.com/forcedotcom/easywriter/issues/173, @W-13522251@

### Functionality Before
Validate PR workflow runs when commits are pushed to the branch.

### Functionality After
In addition to the existing functionality, Now Validate PR workflow also runs when just the PR description is edited with no commits pushed.

### DIFF
--- a/.github/workflows/validatePR.yml
+++ b/.github/workflows/validatePR.yml
@@ -1,7 +1,7 @@
 name: Validate PR
 on:
   pull_request:
-    types: [opened, reopened, edited]
+    types: [opened, reopened, edited, synchronize]
     branches: [develop]
 
 jobs:
--- a/.github/workflows/validatePR.yml
+++ b/.github/workflows/validatePR.yml
@@ -1,12 +1,13 @@
 name: Validate PR
 on:
   pull_request:
+    types: [opened, reopened, edited]
     branches: [develop]
 
 jobs:
   pr-validation:
     uses: salesforcecli/github-workflows/.github/workflows/validatePR.yml@main
-  code-quality: 
+  code-quality:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/validatePR.yml
```diff
@@ -1,12 +1,13 @@
 name: Validate PR
 on:
   pull_request:
+    types: [opened, reopened, edited]
```
gbockus-sf, line 4: Just an FYI when I did this in the EGPT repo it put open PRs into a weird state.  I got past it by just edit each description and it fired off the GHA, but initially they were stuck waiting on validation.  I think it was just a one time issue.

gbockus-sf (approved):
LGTM!

gbockus-sf (commented):
The type check failures are super weird...shouldn't have anything to do with this change

.github/workflows/validatePR.yml
RitamAgrawal: Okay! I will keep that in mind and Watch out for PRs opened after this one is merged.

RitamAgrawal:
> The type check failures are super weird...shouldn't have anything to do with this change

Re-run is also not helpful here.

gbockus-sf:
> > The type check failures are super weird...shouldn't have anything to do with this change
> 
> Re-run is also not helpful here.

Is it happening on develop locally?

RitamAgrawal:
> The type check failures are super weird...shouldn't have anything to do with this change

I think it's because locally it was using node 14.18.3 instead of v16.15.1

PR #4906, opened 2023-06-04
fix: pin apex-node to 1.6.1 due to conflicts

peternhale:
Pinned apex-node to v1.6.1 as v1.6.2 introduced a major release of @salesforce/core. The new version is causing compile errors with conflicts on class Connection from core. 

@W-13532434@


gbockus-sf (approved):
Monorepo fun!  Thanks for the quick fix.

PR #4917, opened 2023-06-12
feat: add orgId to telemetry

klewis-sfdc:
### What does this PR do?
* adds orgId to telemetry data

@W-13554780@

### Functionality Before
* orgId was not included in telemetry

### Functionality After
* orgId is included in telemetry


gbockus-sf (commented):
This looks legit 👍 Can we add the orgId to exception reporting as well or no?

packages/salesforcedx-utils-vscode/src/telemetry/telemetryReporter.ts
```diff
@@ -177,9 +184,12 @@ export class TelemetryReporter extends Disposable {
         measurements
       });
 
+      const orgId = WorkspaceContextUtil.getInstance().orgId;
+      const properties = { orgId };
       if (this.logStream) {
         this.logStream.write(
           `telemetry/${exceptionName} ${JSON.stringify({
+            properties,
```
klewis-sfdc, line 192: I'm just following the pattern above of sending a `properties` object to the `logStream.write` method.  Hoping that this just works, but not sure if we need to adjust anything on the Azure side to accept or map the new property?

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -103,6 +104,14 @@ export class WorkspaceContextUtil {
         defaultUsernameOrAlias !== this._username
           ? defaultUsernameOrAlias
           : undefined;
+      try {
+        const connection = await this.getConnection();
+        this._orgId = connection?.getAuthInfoFields().orgId;
+      } catch (error) {
+        console.log(
+          `There was an problem getting the orgId of the default org: ${error}`
```
klewis-sfdc, line 112: this.getConnection() throws so...

gbockus-sf: nit: should we set _orgId to '' or undefined in this case of a catch?  Looking to avoid having an orgId for the previously auth'd org if this throws when getting the orgId for whatever reason.

packages/salesforcedx-utils-vscode/src/telemetry/telemetryReporter.ts
gbockus-sf: I think it should work :+1:

gbockus-sf (approved):
Verified OrgId is present in telemetry calls ✅ 
🚢

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
klewis-sfdc: Yep, good call

klewis-sfdc: I put this in and added a test for it

gbockus-sf (approved):

lexcraw4d:
Error as posted above

PR #4928, opened 2023-06-20
refactor: consolidate configUtil and authUtil

smarvez:
### What does this PR do?
It centralizes and de-duplicates the two remaining overlapping methods that existed in `authUtil.ts` and `configUtil.ts`, moving both methods to `configUtil.ts` and deleting `authUtil.ts`.

#[113](https://github.com/forcedotcom/easywriter/issues/113), @W-12227695@

### Functionality Before
Both `username` and `defaultUsernameOrAlias` are able to be fetched from the org config.

### Functionality After
Both `username` and `defaultUsernameOrAlias` are able to be fetched from the org config.


packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -54,11 +55,26 @@ export class ConfigUtil {
...
+    try {
+      const configAggregator = await ConfigAggregatorProvider.getInstance().getConfigAggregator();
+      const defaultUsernameOrAlias = configAggregator.getPropertyValue(
+        OrgConfigProperties.TARGET_ORG
+      );
+      if (defaultUsernameOrAlias === undefined) {
+        return undefined;
+      }
+
+      return JSON.stringify(defaultUsernameOrAlias).replace(/\"/g, '');
```
smarvez, line 67: I'm curious why we're doing this (and also, would `String(defaultUsernameOrAlias)` be nicer? What types can be returned from `configAggregator.getPropertyValue()`?

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -54,11 +55,26 @@ export class ConfigUtil {
...
+    } catch (err) {
+      console.error(err);
+      if (err instanceof Error) {
+        TelemetryService.getInstance().sendException(
+          'get_default_username_alias',
+          err.message
+        );
+      }
+      return undefined;
```
jeffb-sfdc, line 76: Instead of catching the exception and returning undefined, it's probably better to re-throw the error.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
jeffb-sfdc: I think the issue with `String(defaultUsernameOrAlias)` is that if `defaultUsernameOrAlias` is undefined, it returns back `"undefined"` and not `undefined`.

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -54,11 +55,26 @@ export class ConfigUtil {
...
-    const defaultUsernameOrAlias = configAggregator.getPropertyValue(
-      OrgConfigProperties.TARGET_ORG
-    );
-    return defaultUsernameOrAlias ? String(defaultUsernameOrAlias) : undefined;
+    try {
+      const configAggregator = await ConfigAggregatorProvider.getInstance().getConfigAggregator();
+      const defaultUsernameOrAlias = configAggregator.getPropertyValue(
+        OrgConfigProperties.TARGET_ORG
+      );
+      if (defaultUsernameOrAlias === undefined) {
```
jeffb-sfdc, line 63: Instead of explicitly looking for undefined, perhaps change it to:
```
if (!defaultUsernameOrAlias) {
  return undefined;
}
```

This will work if defaultUsernameOrAlias is undefined or if it is null.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -52,7 +52,7 @@ export class OrgAuthInfo {
         }
       }
 
-      return JSON.stringify(defaultUsernameOrAlias).replace(/\"/g, '');
+      return defaultUsernameOrAlias;
```
jeffb-sfdc, line 55: There are two copies of `getDefaultUsernameOrAlias()` (one in packages/salesforcedx-utils-vscode/src/config/configUtil.ts, and the other one here, in packages/salesforcedx-vscode-core/src/util/authInfo.ts)

Can they be consolidated?
Or if not consolidated, can they match each other?

smarvez: I feel like the two methods are different enough to remain separate since `OrgAuthInfo.getDefaultUsernameOrAlias()` is doing an extra check for the global default username and it's also displaying warning messages to the user. Would we want that to happen by default from `configUtil.ts`? If not, maybe the method in `authInfo.ts` could be renamed to better reflect the differences.

Issue #4938, opened 2023-06-21
Error: exports.getOrgApiVersion is not a function


Deepak-K-Anand:
### Summary

Unable to deploy code after the recent extension update (58.3.0 - June 21, 2023). Error: `exports.getOrgApiVersion is not a function`

### Steps To Reproduce:

1.  Open any Metadata (Apex, VF Page, LWC etc.)
2.  Right-click and run **SFDX: Deploy This Source to Org**

### Expected result

The metadata should be deployed to the Salesforce org.

### Actual result

Presented with an error in the console that says: `exports.getOrgApiVersion is not a function`

### Additional information

**Observation**: This most likely started appearing after the recent release of 58.3.0 - https://github.com/forcedotcom/salesforcedx-vscode/pull/4917

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/3683725/c3caac8c-c02e-48b5-9964-905bcf3fe427)

**Salesforce Extension Version in VS Code**:
`v58.3.0`

**SFDX CLI Version**:
```
{
  "cliVersion": "@salesforce/cli/1.77.1",
  "architecture": "darwin-x64",
  "nodeVersion": "node-v18.16.0",
  "osVersion": "Darwin 22.5.0",
  "shell": "zsh",
  "rootPath": "/Users/deepakkanand/.local/share/sfdx/client/7.199.7-cc56357/sf",
  "pluginVersions": [
    "@oclif/plugin-autocomplete 2.1.9 (core)",
    "@oclif/plugin-commands 2.2.14 (core)",
    "@oclif/plugin-help 5.2.9 (core)",
    "@oclif/plugin-not-found 2.3.23 (core)",
    "@oclif/plugin-plugins 2.4.7 (core)",
    "@oclif/plugin-search 0.0.16 (core)",
    "@oclif/plugin-update 3.1.13 (core)",
    "@oclif/plugin-version 1.3.3 (core)",
    "@oclif/plugin-warn-if-update-available 2.0.36 (core)",
    "@oclif/plugin-which 2.2.19 (core)",
    "@salesforce/cli 1.77.1 (core)",
    "apex 2.2.14 (core)",
    "auth 2.7.14 (core)",
    "data 2.3.12 (core)",
    "deploy-retrieve 1.8.18 (core)",
    "info 2.6.8 (core)",
    "limits 2.3.13 (core)",
    "login 1.2.7 (core)",
    "org 2.7.2 (core)",
[...]

PR #4941, opened 2023-06-22
fix: changed export style of componentSetUtils.ts and added unit tests back in

jeffb-sfdc:
### What does this PR do?
This PR adds the unit tests for componentSetUtils.ts back in.  In order to do so:
1. The way function are exported in componentSetUtils.ts was changed
2. How the functions in componentSetUtils are imported was changed to reflect the new export style
3. The unit tests for componentSetUtils was aded back in.

https://github.com/forcedotcom/salesforcedx-vscode/issues/4938, @W-10273533@

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -630,20 +630,29 @@
       "internalConsoleOptions": "neverOpen"
     },
     {
-      "name": "Debug Tests - VSCode Core - Unit Tests",
-      "cwd": "${workspaceRoot}/packages/salesforcedx-vscode-core",
-      "type": "node",
-      "request": "launch",
-      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-vscode-core/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
-      "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
-    },
-    {
       "name": "Debug Tests - SObjects Faux Generator - Unit Tests",
       "cwd": "${workspaceRoot}/packages/salesforcedx-sobjects-faux-generator",
       "type": "node",
       "request": "launch",
       "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-sobjects-faux-generator/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen"
+    },
+    {
+      "name": "Debug Tests - Utils VSCode - Unit Tests",
+      "cwd": "${workspaceRoot}/packages/salesforcedx-utils-vscode",
+      "type": "node",
+      "request": "launch",
+      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-utils-vscode/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen"
+    },
+    {
+      "name": "Debug Tests - VSCode Core - Unit Tests",
+      "cwd": "${workspaceRoot}/packages/salesforcedx-vscode-core",
+      "type": "node",
+      "request": "launch",
+      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-vscode-core/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     }
--- a/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
@@ -33,7 +33,7 @@ import { TELEMETRY_METADATA_COUNT } from '../constants';
 import { WorkspaceContext, workspaceContextUtils } from '../context';
 import { handleDeployDiagnostics } from '../diagnostics';
 import { nls } from '../messages';
-import { setApiVersion, setSourceApiVersion } from '../services/sdr/componentSetUtils';
+import { componentSetUtils } from '../services/sdr/componentSetUtils';
 import { DeployQueue, sfdxCoreSettings } from '../settings';
 import { SfdxPackageDirectories } from '../sfdxProject';
 import { createComponentCount, formatException } from './util';
@@ -62,8 +62,8 @@ export abstract class DeployRetrieveExecutor<
 
     try {
       const components = await this.getComponents(response);
-      await setApiVersion(components);
-      await setSourceApiVersion(components);
+      await componentSetUtils.setApiVersion(components);
+      await componentSetUtils.setSourceApiVersion(components);
 
       this.telemetry.addProperty(
         TELEMETRY_METADATA_COUNT,
--- a/packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
+++ b/packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
@@ -10,7 +10,7 @@ import { ComponentSet } from '@salesforce/source-deploy-retrieve';
 import { WorkspaceContext } from '../../context/workspaceContext';
 import { SfdxProjectConfig } from '../../sfdxProject';
 
-export async function setApiVersion(componentSet: ComponentSet): Promise<void> {
+async function setApiVersion(componentSet: ComponentSet): Promise<void> {
   // For a listing (and order of precedence) of how to retrieve the value of apiVersion,
   // see "apiVersion: Order of Precedence" in the "How API Version and Source API Version
   // Work in Salesforce CLI" doc.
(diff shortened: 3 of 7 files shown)


packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
```diff
@@ -67,3 +67,9 @@ export async function getOrgApiVersion(): Promise<string> {
 
   return apiVersion;
 }
+
+module.exports = {
```
gbockus-sf, line 71: q: why are we using module.exports?  I had to google it to understand what it does different (and I learned the other exports in this file are ignored) Using the pattern found in https://github.com/forcedotcom/salesforcedx-vscode/blob/fb6111619ca8c6cbf2dc24100b75c626900dee1a/packages/salesforcedx-utils-vscode/src/workspaces/workspaceUtils.ts#L24-L29 seems like it'd be more consistent with the rest of the code base.

jeffb-sfdc: The problem I'm trying to solve is when a file exports several functions, with one of the functions being a dependency of the other, and then mocking it.
```
export function a() {
  b()
}

export function b() {
}
```

(note: I don't think the syntax here isn't the issue, as I've also tried the CommnJS style of exporting, as well as the style you referenced in workspaceUtils.ts but still not able to mock b())

In my unit tests for this file, I wish to test a(), and mock b()
(and then later when I test b(), to use the actual impl of b(), but that's a different issue)

I ran into problems with this, and in searching around I found this posting which recommended using `exports.`:
https://medium.com/welldone-software/jest-how-to-mock-a-function-call-inside-a-module-21c05c57a39f

This works w/o issue when I run or debug locally, but fails when using a VSIX file, so I'm looking for a different solution.

(I also tried "solution 2" here: https://dev.to/edwinwong90/how-to-mock-function-that-called-inside-module-7f7 but this too did not work)

Do we have test within our codebase which mock inner file dependencies, like the example above with b() being consumed by a()?

jeffb-sfdc: As an example (using workspaceUtils.ts as example, first I modified workspaceUtils.ts just to make it easier to flush out a solution:
```
import { workspace, WorkspaceFolder } from 'vscode';

function hasRootWorkspace(ws: typeof workspace = workspace) {
  debugger;
  // this should not get called if it has been mocked

  // return ws && ws.workspaceFolders && ws.workspaceFolders.length > 0;
  return false;
}

function getRootWorkspace(): WorkspaceFolder {
  // return hasRootWorkspace()
  //   ? workspace.workspaceFolders![0]
  //   : ({} as WorkspaceFolder);

  debugger;

  return hasRootWorkspace()
    ? ({ name: 'was true'  } as WorkspaceFolder)
    : ({ name: 'was false' } as WorkspaceFolder);
}

function getRootWorkspacePath(): string {
  return getRootWorkspace().uri ? getRootWorkspace().uri.fsPath : '';
}

export const workspaceUtils = {
  hasRootWorkspace,
  getRootWorkspace,
  getRootWorkspacePath
};
```


Next, I created a corresponding workspaceUtils.test.ts file:
```
// import {
//   workspaceUtils
// } from '../../../src/';

// import {
//   workspaceUtils
// } from '../../../src/workspaces';

import {
  workspaceUtils
} from '../../../src/workspaces/workspaceUtils';


  it('should return a path to the state folder if the project has a root workspace', () => {

    // mock hasRootWorkspace
    const hasRootWorkspaceSpy = jest.spyOn(workspaceUtils, 'hasRootWorkspace');
    hasRootWorkspaceSpy.mockReturnValue(true);

    debugger;

    // then call getRootWorkspace
    const rootWorkspace = workspaceUtils.getRootWorkspace();

    debugger;

    expect(rootWorkspace.name).toBe('was true');

    debugger;
  });
});
```

...but alas, everytime `getRootWorkspace()` is called, the actual impl of `hasRootWorkspace()` is called.

gbockus-sf: gotcha. Good ol' javascript scoping fun :). The issue here is how does func1 call func2 in such a way that we have access to mock it b/c is using the same reference as the test file. 

I think you have a few options.  
1.  Have one function call the other via the grouping object.  So in the above workspaceUtils change `return hasRootWorkspace()` to `return workspaceUtils.hasRootWorkspace()`.
2. Another option (which I always have to exercise and can get hairy) is to use this when calling the other function in the same module ala `return this.hasRootWorkspace()`.  I think this in this case will refer back to the workspaceUtils object giving you the right access for mocking. 
3. You can refactor to have just a single export ```

.vscode/launch.json
```diff
@@ -630,20 +630,29 @@
...
       "request": "launch",
-      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-vscode-core/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
+      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/packages/salesforcedx-sobjects-faux-generator/node_modules/.bin/jest", "--runInBand", "--coverage", "false"],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     },
     {
-      "name": "Debug Tests - SObjects Faux Generator - Unit Tests",
-      "cwd": "${workspaceRoot}/packages/salesforcedx-sobjects-faux-generator",
+      "name": "Debug Tests - Utils VSCode - Unit Tests",
```
jeffb-sfdc, line 642: Added a launch config for salesforcedx-utils-vscode

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -33,7 +33,7 @@ import { TELEMETRY_METADATA_COUNT } from '../constants';
 import { WorkspaceContext, workspaceContextUtils } from '../context';
 import { handleDeployDiagnostics } from '../diagnostics';
 import { nls } from '../messages';
-import { setApiVersion, setSourceApiVersion } from '../services/sdr/componentSetUtils';
+import { componentSetUtils } from '../services/sdr/componentSetUtils';
```
jeffb-sfdc, line 36: Changed the way the functions in componentSetUtils.ts are exported in order to write unit tests for them.

packages/salesforcedx-vscode-core/test/jest/services/sdr/componentSetUtils.test.ts
```diff
@@ -8,7 +8,7 @@
 import { ConfigUtil } from '@salesforce/salesforcedx-utils-vscode';
 import { ComponentSet } from '@salesforce/source-deploy-retrieve';
 import { WorkspaceContext } from '../../../../src/context/workspaceContext';
-import * as componentSetUtils from '../../../../src/services/sdr/componentSetUtils';
+import { componentSetUtils } from '../../../../src/services/sdr/componentSetUtils';
```
jeffb-sfdc, line 11: Added the unit tests for componentSetUtils.ts back in.

packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
```diff
@@ -54,16 +54,22 @@ export async function setSourceApiVersion(componentSet: ComponentSet): Promise<v
...
 
-export async function getOrgApiVersion(): Promise<string> {
+async function getOrgApiVersion(): Promise<string> {
   const connection = await WorkspaceContext.getInstance().getConnection();
   const apiVersion = connection.getApiVersion();
 
   return apiVersion;
 }
+
+export const componentSetUtils = {
```
jeffb-sfdc, line 71: Changed the way the functions in componentSetUtils.ts are exported in order to write unit tests for them.

packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
```diff
@@ -54,16 +54,22 @@ export async function setSourceApiVersion(componentSet: ComponentSet): Promise<v
 
   // Next, if it still is not set, set it to the highest API version supported by the target org.
   if (!sourceApiVersion) {
-    const orgApiVersion = await getOrgApiVersion();
+    const orgApiVersion = await componentSetUtils.getOrgApiVersion();
```
jeffb-sfdc, line 57: Call `componentSetUtils.getOrgApiVersion()` now so that `getOrgApiVersion()` can be mocked in unit tests.

packages/salesforcedx-vscode-core/src/services/sdr/componentSetUtils.ts
jeffb-sfdc: Thanks, option `#1` is working out.  I've updated the PR and the other places where componentSetUtils is used.

@RitamAgrawal and @CristiCanizales I've d/l and installed the VSIX files were were generated from this PR, and after installing I was able to deploy, so things are looking for for this PR.  Please review & test, and let me know if either of you discover any issues.

gbockus-sf: nice breakdown of the problem btw 👍

CristiCanizales (approved):
LGTM :) I verified deploy/retrieve/manifest/create/etc ops are working as expected✅

Issue #4904, opened 2023-06-03
Issue on problem panel not get clear after successful deployment


vikaskumar171020:
Hi i the latest release there is issue i found sometime after deploying code it show error for unsuccessfully code deployment and the error show even after the code deployed successfully which is not clear the problem in the file.

FOR BUGS RELATED TO THE SALEFORCE CLI, please use this repository: https://github.com/forcedotcom/cli

### Summary

_Short summary of what is going on or to provide context_.

### Steps To Reproduce:

1.  This is step 1.
1.  This is step 2. All steps should start with '1.'

### Expected result

problem need to clear after successful deployment 
### Actual result

problem still appear after successful deployment 

**Salesforce Extension Version in VS Code**:
Salesforce CLI Integration version "v58.0.1"

**SFDX CLI Version**:
sfdx-cli/7.203.6 win32-x64 node-v18.15.0

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/67593607/0e5c0ca7-1fea-4afd-b6ac-cf24ccc877e3)

YukiOkumura:
downgrading following extensions from 58.0.1 to 57.15.0 worked for me:

Apex
Apex Replay Debugger
Aura Components
Lightning Web Components
Salesforce CLI Integration
Salesforce Extension Pack

how you can choose a previous version:
https://github.com/forcedotcom/salesforcedx-vscode/issues/3901#issuecomment-1064188043

vikaskumar171020:
yeah already did that but due to auto update feature all extension gets updated which cause this issue come over and over again

yuki18886:
@vikaskumar171020 
You may consider to trun off auto update settings, in case you are in the situation where you can manage a personal settings.

see also: 
[Extension auto-update](https://code.visualstudio.com/docs/editor/extension-marketplace#_extension-autoupdate)
[Update an extension manually](https://code.visualstudio.com/docs/editor/extension-marketplace#_update-an-extension-manually)

akalatksy:
➕  Also affected by this. 
Downgrade of extensions mentioned by @yuki18886 helped. Thanks @yuki18886 ! 👍

vikaskumar171020:
yeah thanks that work for me @yuki18886

bobalicious:
Any update on this? This is a major obstacle for development - it gets in the way constantly - big downgrade in functionality.

AndrewStopchenko-SO:
[...]

PR #4962, opened 2023-07-05
fix: fix for problems tab not clearing errors

CristiCanizales:
### What does this PR do?
fixes issue that displayed errors in the problems tab even after a successful deployment

@W-13553883@
closes https://github.com/forcedotcom/salesforcedx-vscode/issues/4904

### Functionality Before
- errors in problems tab not cleared after successful deployment

### Functionality After
- errors in problems tab cleared after successful deployment

### DIFF
--- a/packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
+++ b/packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
@@ -569,6 +569,62 @@ describe('Base Deploy Retrieve Commands', () => {
         expect(appendLineStub.calledOnce).to.equal(true);
         expect(appendLineStub.firstCall.args[0]).to.equal(expectedOutput);
       });
+
+      xit('should report any diagnostics if deploy failed', async () => {
+        const executor = new TestDeploy();
+
+        const mockDeployResult = new DeployResult(
+          {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/test/vscode-integration/commands/baseDeployRetrieve.test.ts
```diff
@@ -569,62 +569,6 @@ describe('Base Deploy Retrieve Commands', () => {
         expect(appendLineStub.calledOnce).to.equal(true);
         expect(appendLineStub.firstCall.args[0]).to.equal(expectedOutput);
       });
-
```
RitamAgrawal, line 572: Since we are not fixing the mocha test or replacing it with jest test, let's not remove the test. Eith 'x' it or comment out the test. 
`xit('should......')` will prevent it from running.

RitamAgrawal:
QA Notes
✅ Removes the problem after successful deployment
✅ only removes the problem that has been  fixed, remaining problems stay there for reference.

Changes have been made in DeployExecutor and jest tests is a heavy lift for the code change at this point in time. So we will cover the test as a separate [W-13736554](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001VpH55YAF/view)

RitamAgrawal (approved):

PR #4963, opened 2023-07-06
Fix SFDX ISV Debugging: Update command line arguments for debugger

ChrisBland:
### What does this PR do?
This PR updates the command line arguments for sfdx isv debugging by changing the arguments to the new ones. SFDX updated the CLI so that --rootdir changed to --root-dir, see attached screenshot.

### Functionality Before
image

### Functionality After
🚀


CristiCanizales:
Hi! I'll close this PR as it has no effect on current code.

PR #4993, opened 2023-07-27
feat: replace sfdx to sf style flag name

RitamAgrawal:
### What does this PR do?
Upgrades deprecated cli flags to new flags.
Removes warnings while setting up the ISV debugger

#<Insert GitHub Issue>, @W-13834204@

### Changes
| Old style  | New style |
|--------|--------|
| --outputdir | --output-dir |
| --projectname | --name |
| --targetusername | --target-org |
| --retrievetargetdir | --target-metadata-dir |
| --unpackaged | --manifest |
|--packagenames | --package-name |
isvDebuggerSid | org-isv-debugger-sid
isvDebuggerUrl |org-isv-debugger-url
instanceUrl |org-instance-url

### Functionality After
Cli warnings should go away

Docs for reference: https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_mdapi.htm#cli_reference_force_mdapi_retrieve

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
@@ -79,9 +79,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
   public buildCreateProjectCommand(data: IsvDebugBootstrapConfig): Command {
     return new SfdxCommandBuilder()
       .withDescription(nls.localize('isv_debug_bootstrap_step1_create_project'))
-      .withArg('force:project:create')
-      .withFlag('--projectname', data.projectName)
-      .withFlag('--outputdir', data.projectUri)
+      .withArg('project generate')
+      .withFlag('--name', data.projectName)
+      .withFlag('--output-dir', data.projectUri)
       .withFlag('--template', 'standard')
       .withLogName('isv_debug_bootstrap_create_project')
       .build();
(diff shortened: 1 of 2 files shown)


CristiCanizales (commented):
Nice!😄 `bootstrapCmd.test.ts` should be updated accordingly

packages/salesforcedx-vscode-core/test/vscode-integration/commands/isvdebugger/bootstrapCmd.test.ts
```diff
@@ -202,7 +202,7 @@ describe('ISV Debugging Project Bootstrap Command', () => {
         }
       );
       expect(command.toCommand()).to.equal(
-        `sfdx force:data:soql:query --query SELECT NamespacePrefix FROM Organization LIMIT 1 --targetusername ${SESSION_ID} --json --loglevel fatal`
+        `sfdx data query --query SELECT NamespacePrefix FROM Organization LIMIT 1 --target-org ${SESSION_ID} --json --loglevel fatal`
```
CristiCanizales, line 205: And in that case change/remove it here too

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -109,9 +109,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
           'isv_debug_bootstrap_step2_configure_project_retrieve_namespace'
         )
       )
-      .withArg('force:data:soql:query')
+      .withArg('data query')
       .withFlag('--query', 'SELECT NamespacePrefix FROM Organization LIMIT 1')
-      .withFlag('--targetusername', data.sessionId)
+      .withFlag('--target-org', data.sessionId)
       .withJson()
       .withLogName('isv_debug_bootstrap_configure_project_retrieve_namespace')
```
CristiCanizales, line 116: It seems like we should also change/get rid of this flag...
"The loglevel flag is no longer in use on this command. You may use it without error, but it will be ignored.\nSet the log level using the SFDX_LOG_LEVEL environment variable."

CristiCanizales (commented):
LGTM! Just a small comment regarding a flag

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
RitamAgrawal: Yes! I'm not clear why are we using the --loglevel flag in these commands So not touching them for now. And having the flag --loglevel does not throwing any warnings so we should just keep it for now.

CristiCanizales (approved):
🚢

PR #5014, opened 2023-08-03
chore: remove pinned vsc version from int test run

RitamAgrawal:
### What does this PR do?
Replaces the `1.74.3` pinned version of vscode with `stable` in IntegrationTestsWindows.yml
Now it allows the win integration tests to run on latest stable version of vscode.
 @W-12498824@

### Functionality Before
Win Integration tests were run on vscode v1.74.3

### Functionality After
Win Integration tests run on latest stable version

Reference: #[VSCode Issue](https://github.com/microsoft/vscode/issues/181528)

### DIFF
--- a/.github/workflows/integrationTestsWindows.yml
+++ b/.github/workflows/integrationTestsWindows.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: windows-latest
     timeout-minutes: 60
     env:
-      CODE_VERSION: 1.74.3
+      CODE_VERSION: stable
     steps:
       - run: git config --global core.autocrlf false
       - uses: actions/checkout@v3


RitamAgrawal:
The reason I'm not removing the env variable CODE_VERSION is that in various implementation of function `runIntegrationTests()` the first parameter `version` is not passed and the value is coming from process.env.CODE_VERSION.
Such implementations are the in scripts: `run-tests-with-recipes.js` , `run-vscode-integration-tests-with-top-level-extension.js`  & `run-vscode-integration-tests.js`

.github/workflows/integrationTestsWindows.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: windows-latest
     timeout-minutes: 60
     env:
-      CODE_VERSION: 1.74.3
+      CODE_VERSION: stable
```
gbockus-sf, line 14: cool..this is the default if the value isn't provider so should be good to go.

gbockus-sf (approved):
LGTM

peternhale (approved):

PR #5042, opened 2023-08-24
fix: Suppress Output Channel with sObject Refresh in a Container Env

jonnyhork:
### What does this PR do?

This PR will prevent the output channel from opening when the `forceRefreshSObjects` command is run if the extension is running in a container environment. 

[This change is requested](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001YiA7fYAF/view) by @smaddox-sf because the output channel can interfere with the onboarding experience in Code Builder. Ideally we would only suppress the channel from opening at start up, however, as far as I can tell that would push the scope of this work too far.

With this change, sObject will still be refreshed at start up, and when run form the command palette, users can also see the output from the CLI, they will just have to open the output panel manually. 

[@W-13970705@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001YiA7fYAF/view)
### Functionality Before

**With the output channel at startup**
with_output_channel

### Functionality After

**Without the output channel at startup**

without_output_channel

**Notifications with manual sObject refresh (current functionality)**

manually_refresh_sObjects


packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
```diff
@@ -108,7 +109,7 @@ export class ForceRefreshSObjectsExecutor extends SfdxCommandletExecutor<{}> {
 
     channelService.streamCommandOutput(execution);
 
-    if (this.showChannelOutput) {
+    if (this.showChannelOutput && !isSFContainerMode()) {
```
jonnyhork, line 112: This seemed like the best place to get the behavior we are looking for while scoping the impact to just this one command for now. I think it might be nice in the future if there are mechanisms in place to turn up or down the noise (output, terminal, toasts, etc) thrown at the user from our extensions. Especially at startup.

jgellin-sf (approved):
💯

CristiCanizales (approved):
LGTM :)
🚢

packages/salesforcedx-vscode-core/src/commands/forceRefreshSObjects.ts
RitamAgrawal: At startup are you running any particular command like, `SFDX: Create Project`?

jonnyhork: In the context of testing this out? No, to test it locally I toggle the env `SF_CONTAINER_MODE` in the launch configuration for `Launch Extensions`, then open an SFDX project. I am happy to show you if you want.

RitamAgrawal (approved):
QE Notes:
Running  SFDX: Refresh SObject Definitions with SF_CONTAINER_MODE= true

* Output panel doesn’t pop up at launch ✅ 
* focus doesn’t shift from terminal to output panel ✅ 

Running  SFDX: Refresh SObject Definitions with SF_CONTAINER_MODE= false

* Output Panel appears as expected ✅ 
* focus shifts from terminal to output panel ✅ 

SFDX: Create Project with standard template refreshes SObject Definitions in the process.

* in non-container, desktop mode, focus shifts to output panel ✅ 
* in container mode, focus doesn’t shift to output panel after refreshing sobjects. ✅

PR #5092, opened 2023-09-13
Update job names in e2e.yml

daphne-sfdc:
Update job names in e2e.yml


CristiCanizales (approved):
Thank you for updating these names :)

Issue #4865, opened 2023-05-17
Deploy Source to org command takes several minutes to complete


WuggyX2:
### Summary

We have noticed that when using the "deploy source to org" command from the command palette, the notification that the action in running does not close even though the deployment has already completed in the backgroud. This give a wrong impression that the deployment takes several several minutes to complete. Some times the deployed to source message gets printed twice even though the command was run only once.

### Steps To Reproduce:

1. Create a new project or take an existing one and connect it to an org of your choosing.
1. Do a change to a lightning web component
1. Run the "deploy source to org" command

### Expected result

Success message for the deployment and the list of deployed components should appear in a second or two

### Actual result

Success message takes several minutes to appear. Sometimes the command in progress notificaiton keeps running even though list of deployed components has already been printed.

### Additional information
![kuva](https://github.com/forcedotcom/salesforcedx-vscode/assets/51646225/662e8d4e-8dda-4c29-8cbc-507170164aa3)

Here's an example when deployed to source gets printed twitch even though the command is run only once
![kuva](https://github.com/forcedotcom/salesforcedx-vscode/assets/51646225/f9265c54-8174-4165-95f6-ca28b67d14b9)

**Salesforce Extension Version in VS Code**:
v57.13.1

**SFDX CLI Version**:
@salesforce/cli/1.78.0 darwin-x64 node-v18.15.0

**OS and version**:
MacOS Ventura 13.3.1

randi274:
@WuggyX2 I've noticed this recently as well - if you cancel the first deploy and then run the command again, do you also find it's much faster on the second run?

WuggyX2:
@randi274 Unfortunately canceling the first deploy does not seem to increase the speed of the second deploy. However, removing the `org-metadata-rest-deploy`config variable for the old deploy/retrieve commands seems to have helped for now at least for me. That variable also breaks the default org switcher in the status bar.

mnunezdm:
In my case, if i go to the environment I find that deployment is done like normally but vscode keeps polling

randi274:
@WuggyX2 or others - what type of orgs are you working against?

WuggyX2:
@randi274 Me and my team are working against developer sandboxes.

JavierDAP:
I don't know if it is related but I noticed that now .sf\orgs\ is getting populated with the source tracking. With each deploy that I do, the folders gets bigger and bigger and makes things slower.
[...]

PR #5115, opened 2023-09-20
feat: Update STL Implementation to Speed Up Deploy/Retrieve Operations

klewis-sfdc:
### What does this PR do?
* intends to speed up deploy/retrieve operations when Source Tracking is involved by 
  * only ever creating a single instance of STL per project and connection per VS Code session
  * upgrading /source-tracking library to the latest which requires also upgrading:
    *  /core
    * /source-deploy-retrieve
    * /apex-node

@W-14160535@
Re: #4865

### Functionality Before
 * multiple instances of STL could be created per project, each listening for deployment hooks and running their own post-deploy processes
 * stl was on an older version (4.1.3) 

### Functionality After
 * only a single instance of STL per project and connection per VS Code session
 * stl updated to latest (4.2.14)
 * core updated to latest (5.3.0) 
 * sdr updated to latest (9.7.18)
 * apex-node updated to latest (2.1)


packages/salesforcedx-utils-vscode/src/providers/sourceTrackingProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Creates an instance of SourceTracking with options
+   * configured to work in VSCE.
+   * Since SourceTracking is initialized with an SfProject, which
+   * contains the project path, and PR #4643 made it so that VSCE is
+   * running with process.cwd set as the project root, there
+   * is no need to call process.chdir here as has been done in VSCE
+   * with other core types like Config and ConfigAggregator.
+   */
```
gbockus-sf, line 57: ❤️

packages/salesforcedx-utils-vscode/src/providers/sourceTrackingProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+    connection: Connection
+  ): Promise<SourceTracking> {
+    const project = await SfProject.resolve(projectPath);
+    const org = await Org.create({ connection });
+    const options: SourceTrackingOptions = {
+      org,
+      project,
+      ignoreLocalCache: true,
+      subscribeSDREvents: true,
+      ignoreConflicts: true
```
gbockus-sf, line 69: would it be worth a code comment to call out why it's configured this way? Seems important since the first instinct @WillieRuemmele had was to change the localCache setting.

packages/salesforcedx-vscode-apex-replay-debugger/package.json
```diff
@@ -24,8 +24,8 @@
     "Debuggers"
   ],
   "dependencies": {
-    "@salesforce/apex-node": "1.6.2",
-    "@salesforce/core": "4.3.11",
+    "@salesforce/apex-node": "2.1",
```
gbockus-sf, line 27: nit: 2.1.0

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -24,11 +24,11 @@
     "Programming Languages"
   ],
   "dependencies": {
-    "@salesforce/apex-node": "1.6.2",
+    "@salesforce/apex-node": "2.1",
```
gbockus-sf, line 27: same nit: 2.1.0

gbockus-sf (commented):
changes look great 👍

packages/salesforcedx-vscode-apex-replay-debugger/package.json
gbockus-sf: It's interesting that this worked when the version of core in apex-node is 5.2.0 and we have 5.3.0 here.

gbockus-sf: The overall extension size went from 173MB -> 176MB.  I wonder if the extra version of core is the size diff 🤷

gbockus-sf (commented):
I took a pass at QE.  Looking at the changes it appeared that doing a high level test of deploy/retrieve operations with STL enabled was the right path. Please checkout the QE doc

https://salesforce.quip.com/jLWCAQSEF6w9

The only odd error I ran into was I was getting a failure when attempting to open the org from my dev hub originally.  The problem went away on a scratch org, then worked fine after switching back to the dev hub.  Probably worth someone else verifying that open default org button works as expected. 

Please review the QE doc and let me know of any other test cases I should add.

gbockus-sf (approved):
Verified the changes commands & new labels as discussed and updated QE doc. 
Everything looks good to go 👍

PR #5117, opened 2023-09-21
fix: stop listeners from piling up, increase speed with large changes

WillieRuemmele:
### What does this PR do?

after retrieving `CustomObjects` from the org browser, deploying via the right click option became super slow for even a single class. This clears multiple listeners, increases STL performance by changing the options, and bumps the STL version to the latest (which had the largest impact)

@W-14160535@

### QA Setup

1. create a new org
2. from the Org Browser, retrieve all "Custom Objects"
3. from your local project (dreamhouse-lwc), right click on a class (FileUtilities can be deployed individually) and `SFDX: deploy source to org`
4. look at the time it took to complete

### Functionality Before
when a project had 2k+ local changes (from the COs) deploys were taking 5+ mins

### Functionality After
now they take closer to 5s, which matches how long they take via the CLI in the same scenario

```
10:48:06.776 Starting SFDX: Deploy Source to Org

=== Deployed Source
STATE      FULL NAME      TYPE       PROJECT PATH                                             
─────────  ─────────────  ─────────  ─────────────────────────────────────────────────────────
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls         
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls-meta.xml

10:48:13.922 ended SFDX: Deploy Source to Org
10:48:18.660 Starting SFDX: Deploy Source to Org

=== Deployed Source
STATE      FULL NAME      TYPE       PROJECT PATH                                             
─────────  ─────────────  ─────────  ─────────────────────────────────────────────────────────
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls         
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls-meta.xml

10:48:23.784 ended SFDX: Deploy Source to Org
10:48:26.933 Starting SFDX: Deploy Source to Org

=== Deployed Source
STATE      FULL NAME      TYPE       PROJECT PATH                                             
─────────  ─────────────  ─────────  ─────────────────────────────────────────────────────────
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls         
Unchanged  FileUtilities  ApexClass  force-app/main/default/classes/FileUtilities.cls-meta.xml

10:48:31.818 ended SFDX: Deploy Source to Org
```

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
+++ b/packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
@@ -9,6 +9,7 @@ import { Connection, Org, SfProject } from '@salesforce/core';
 import { RetrieveResult } from '@salesforce/source-deploy-retrieve';
 import {
   SourceTracking,
+  SourceTrackingOptions,
   StatusOutputRow
 } from '@salesforce/source-tracking';
 import { WorkspaceContextUtil } from '../context/workspaceContextUtil';
@@ -34,13 +35,13 @@ export class SourceTrackingService {
   ): Promise<SourceTracking> {
     const project = await SfProject.resolve(projectPath);
     const org = await Org.create({ connection });
-
-    const sourceTracking = await SourceTracking.create({
+    const options: SourceTrackingOptions = {
       org,
       project,
       subscribeSDREvents: true,
       ignoreConflicts: true
-    });
+    };
+    const sourceTracking = await SourceTracking.create(options);
     return sourceTracking;
   }
 
--- a/packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
+++ b/packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
@@ -9,7 +9,6 @@ import { Connection, Org, SfProject } from '@salesforce/core';
 import { RetrieveResult } from '@salesforce/source-deploy-retrieve';
 import {
   SourceTracking,
-  SourceTrackingOptions,
   StatusOutputRow
 } from '@salesforce/source-tracking';
 import { WorkspaceContextUtil } from '../context/workspaceContextUtil';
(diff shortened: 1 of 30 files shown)


CristiCanizales:
Currently we're on @salesforce/core v4 and @salesforce/source-tracking v4.2.13 has a dependency on @salesforce/core v5. The latest version of @salesforce/source-tracking that uses @salesforce/core v4 is v4.2.9, could we use that one instead?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -120,6 +121,8 @@ export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T> {
     components: ComponentSet,
     token: vscode.CancellationToken
   ): Promise<DeployResult | undefined> {
+    Lifecycle.getInstance().removeAllListeners('scopedPostDeploy');
+    Lifecycle.getInstance().removeAllListeners('scopedPreDeploy');
```
klewis-sfdc, line 125: I'm pretty sure that removing these listeners will have a negative effect on subsequent calls to "SFDX: View Local Changes" and the other "View Changes" commands.  STL will get out of sync and report an incorrect status.  I am going to verify and also explore another option, managing a single instance of STL, which is being explored in [this PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/5115).

packages/salesforcedx-utils-vscode/src/services/sourceTrackingService.ts
```diff
@@ -38,7 +38,6 @@ export class SourceTrackingService {
     const options: SourceTrackingOptions = {
       org,
       project,
-      ignoreLocalCache: true,
```
klewis-sfdc, line 41: My hunch is that this will also have an effect on the accuracy of calls to "View Changes" commands.  IIRC this was here because without it STL would get out of sync when creating new files locally/on the server, and STL doesn't expose a command to "refresh" or "re-index" it on demand.  Will investigate.

ImJohnMDaniel:
curiously watching progress 😎

gbockus-sf:
address in https://github.com/forcedotcom/salesforcedx-vscode/pull/5115

PR #5119, opened 2023-09-21
chore: set default aiKey value in TelemetryService

floralan:
### What does this PR do?
As a developer, to avoid copying the `aikey` to every new extension that utilizes the Telemetry Service from Core Library we should set a default key that the Telemetry Service will utilize if none is provided on instantiation.

@W-13485911@

### Functionality Before
Require to pass in aiKey when instantiating the Telemetry Service 

### Functionality After
aiKey is an optional param and it's set with a default value on instantiation of Telemetry Service


packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      // create vscode extensionContext
+      mockExtensionContext = new MockExtensionContext(
+        true,
+        ExtensionMode.Production
+      );
+
+      await telemetryService.initializeService(
+        mockExtensionContext,
+        'ext_name',
+        undefined,
```
floralan, line 196: Oddly, this test is failing on my end due to this exception:`test/jest/telemetry/index.test.ts:196:9 - error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'string'.`. However, the aiKey in this function can be string or undefined. Do you have a clue why it's failing?

floralan: But, it passed with GitHub actions 🤔

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -56,7 +56,7 @@ export class TelemetryService {
   private static instance: TelemetryService;
   private extensionContext: ExtensionContext | undefined;
   private reporter: TelemetryReporter | undefined;
-  private aiKey: string = '';
+  private aiKey: string = 'ec3632a4-df47-47a4-98dc-8134cacbaf7e';
```
lahernandezb, line 59: In response to your failing test... try the following. 


```suggestion
  private aiKey: string | undefined;
```

I don't think your current implementation is working because from what I can see we never call `new TelemetryService` so the default you're setting here is not being applied since there is no instance of this class. Instead were using this class statically via the method `initializeService`. At least this is how I understand things. Feel free to comment if you disagree.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -79,12 +79,12 @@ export class TelemetryService {
   public async initializeService(
     extensionContext: ExtensionContext,
     extensionName: string,
-    aiKey: string,
+    aiKey: string | undefined,
```
lahernandezb, line 82:
```suggestion
    aiKey = extensionPackage.aiKey,
```

alternatively you could just use the string value but I think leaving it in the package json of sfdx-utils-vscode is the way to go.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -79,12 +79,12 @@ export class TelemetryService {
...
     version: string
   ): Promise<void> {
     this.extensionContext = extensionContext;
     this.extensionName = extensionName;
-    this.aiKey = aiKey;
+    this.aiKey = aiKey || this.aiKey;
```
lahernandezb, line 87:
```suggestion
    this.aiKey = aiKey;
```

packages/salesforcedx-utils/package.json
```diff
@@ -3,7 +3,6 @@
   "displayName": "Node Utilities for VS Code",
   "version": "58.15.0",
   "description": "Provides node specific utilities for the Salesforce VSCode extensions",
-  "aiKey": "ec3632a4-df47-47a4-98dc-8134cacbaf7e",
```
lahernandezb, line 6: Keep the value in the package.json.

lahernandezb: Ok looking at this again. I think I might have mislead you with the comments above. Per our slack discussion I think you're on the right path. Sorry about the mix up.

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
floralan: It passed in my local. 👍

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -79,12 +79,12 @@ export class TelemetryService {
   public async initializeService(
     extensionContext: ExtensionContext,
     extensionName: string,
-    aiKey: string,
+    aiKey: string | undefined,
     version: string
```
lahernandezb, line 83: Flora, sorry I missed this the first time around but what do you think about using named parameters in this function? This would also remove the need of having to pass `undefined` when it's called.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -79,12 +79,12 @@ export class TelemetryService {
   public async initializeService(
     extensionContext: ExtensionContext,
     extensionName: string,
-    aiKey: string,
-    version: string
+    version: string,
```
gbockus-sf, line 82: Howdy @floralan 
What you think about making the following update? 

```suggestion
    extensionContext: ExtensionContext
  ): Promise<void> {
    const { name, version, aiKey } = extensionContext.extension.packageJSON;
    this.extensionContext = extensionContext;
    this.extensionName = name;
```
This would allow us to avoid syncing the change across extensions (the extra parameters would just be ignored), but would require some more touches in this repo? I took a look and it appears most places with couple of exceptions where name is hard coded are doing the pull from the packageJSON property on the context anyway so this make this more consistent across extensions. 

It would probably be worthwhile to also add some defensive coding to verify that the expected JSON properties are found and throw a clear error if that's not that case.

gbockus-sf: note this would remove all parameters but the context.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -77,15 +77,19 @@ export class TelemetryService {
    * @param extensionName extension name
    */
   public async initializeService(
-    extensionContext: ExtensionContext,
-    extensionName: string,
-    aiKey: string,
-    version: string
+    extensionContext: ExtensionContext
```
floralan, line 80: As Gordon suggested, we removed other params and only kept the extensionContext. In this way, the EFD extension is not required to be updated, and can still call this function as the extra params will be ignored. This can avoid "users have to update the core extension" issue.

klewis-sfdc (approved):
LGTM!  Thank you, @floralan!

Passes my QA:
I built this PR and installed into Insiders via vsix.  Then I installed Einstein for Developers from the marketplace and confirmed basic functionality of the EfD extension. Was able to generate completions and copy/clear etc. ✅

PR #5125, opened 2023-09-27
fix: updates std sobjects in minSObjects.json

peternhale:
@W-13871991@
Updates min sobject set properties and includes a couple of new in the set.


packages/salesforcedx-sobjects-faux-generator/scripts/regenerateMinsSObjects.ts
```diff
@@ -44,10 +44,13 @@ async function createConnection(aliasOrUsername: string) {
 async function generateLocalSobjectJSON(connection: Connection) {
   const sobjectNames = [
     'Account',
+    'AccountHistory',
```
mingxuanzhangsfdx, line 47: what source refers to the 3 sobjectNames?

packages/salesforcedx-sobjects-faux-generator/scripts/regenerateMinsSObjects.ts
```diff
@@ -79,7 +82,7 @@ async function generateLocalSobjectJSON(connection: Connection) {
     JSON.stringify({
       typeNames: output.getTypeNames(),
       standard: output.getStandard().map(removeCustomFields)
-    })
+    }, undefined, 2)
```
mingxuanzhangsfdx, line 85: Why is it necessary to add indentation?

packages/salesforcedx-sobjects-faux-generator/src/data/minSObjects.json
mingxuanzhangsfdx, line 1: Does this file generated by `npx ts-node ./scripts/regenerateMinsSObjects.ts <token> <instance_url>`?
what is token and instance_url then?

mingxuanzhangsfdx, line 1: Is this generated by `npx ts-node ./scripts/regenerateMinsSObjects.ts <token> <instance_url>`?
If so what are <token> and <instance_url>?

mingxuanzhangsfdx: "token" and instance_url

peternhale: The CLI has the ability to use an access token (encoded string, unique to a given user) and a instance URL, which is the instance URL of the org used to fetch sobjects from. The parameters are just passed through to the sfdx commands used to get the sobject data.

packages/salesforcedx-sobjects-faux-generator/scripts/regenerateMinsSObjects.ts
peternhale: None that we own. This set of sobejcts is a "educated guess" of what the minimum set might be, with this major release adding three more.

packages/salesforcedx-sobjects-faux-generator/src/data/minSObjects.json
mingxuanzhangsfdx: Would you like to point me what the exact sfdx command you used? I am just curious - in case I need to make such changes in future.

mingxuanzhangsfdx (approved):

PR #5136, opened 2023-10-09
fix: only depbot changes to both pkg json and lock

peternhale:
@W-14260306@
Modify dependabot.yml to allow only dep changes that affect package.json.
This change should prevent PRs from being created when the module being bump is transient dependency.


gbockus-sf (approved):
nice 🚢 
Will need to create an associated WI to get past the validate PR check.

PR #5140, opened 2023-10-09
fix: update eslint to version 8.32.0 across extensions

peternhale:
@W-14260664@

I was looking at a dependabot PR that was doing a minor version bump to eslint. Upon examining the change I discovered that one package had already upgraded to version 8 of eslint.

I then chose to upgrade all eslint version found in the mono-repo to version 8 of eslint.

One should see. no changes in functionality from this PR.


gbockus-sf (approved):
Looks good.  Feels like a if it builds it builds PR but if you disagree let me know 
What was the change that resulted in a major version bump? (from 7-8) 
✅ 

@peternhale mind updating the PR description with the details you shared in standup about why you needed to do the PR instead of dependabot? Maybe keep the parts of the usual template that make sense (description/WI addressed).

PR #5151, opened 2023-10-12
Mingxuanzhang/248 synchronize init tasks

mingxuanzhangsfdx:
### What does this PR do?
To add an option on if apex language client should wait until init jobs including indexing are done.

@W-14194698@

### Functionality Before
Client does not wait for initializing jobs including indexing so it could send requests to Apex LS before indexing is done, which causes mess like DB deadlock.

### Functionality After
Client will not send request until initializing jobs including indexing are done. And this setting could also be switched back through Salesforce Apex Configuration in Settings tab of VS Code.


peternhale (commented):
I like simple!

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
```diff
@@ -0,0 +1,19 @@
+import * as vscode from 'vscode';
```
peternhale, line 1: @mingxuanzhangsfdx This test is checking if `vscode.workspace.getConfiguration()` is working, which we have to assume is true, given it is provided by a third party. What is important is how we are using vscode settings to attain the desired behavior.

Please consider writing a test that answers the quest "does the code that modifies the server initialization options properly apply the custom settings `salesforcedx-vscode-apex.wait-init-jobs`?"

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,24 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+      await getTestOutlineProvider().refresh();
+      languageServerReady();
+    });
+
+    // client will check status with server after initialization
+    const indexerStatus = await languageClient.sendRequest('indexer/status');
+    if (indexerStatus === ServiceState.RUNNING) { // 2 is RUNNING state
+      await getTestOutlineProvider().refresh();
+      languageServerReady();
+    } else if ([ServiceState.STOPPING, ServiceState.FAILED, ServiceState.TERMINATED].includes(indexerStatus as number)) {
```
peternhale, line 329: @mingxuanzhangsfdx failures in language server startup or shutting down are handled in [apexErrorhandler.ts](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-apex/src/apexErrorHandler.ts). Can you tell me why you felt it necessary to cover these two states here as well?

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,24 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+
+    // The listener should be set after languageClient is ready
+    // Language client will get notified once async init jobs are done
+    languageClient.onNotification('indexer/done', async () => {
+      await getTestOutlineProvider().refresh();
+      languageServerReady();
+    });
+
+    // client will check status with server after initialization
+    const indexerStatus = await languageClient.sendRequest('indexer/status');
```
peternhale, line 325: Nice!

packages/salesforcedx-vscode-apex/src/index.ts
mingxuanzhangsfdx: My guts are that there is no callback from indexer service in sync mode, so language server has no idea if indexer service is healthy after start it async'ly. The situation that language server is ready but indexer failed is not what users want.

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
mingxuanzhangsfdx: Does this test deserve to implement and look good to check if vscode.workspace.getConfiguration() is working?

mingxuanzhangsfdx: Do you mean writing codes like this - 
`it('value should be updated after modifying it', () => {
    const config = vscode.workspace.getConfiguration();
    // expect(config.has('salesforcedx-vscode-apex.wait-init-jobs')).toBe(true);
    config.update('salesforcedx-vscode-apex.wait-init-jobs', false);
    expect(config.get<boolean>('salesforcedx-vscode-apex.wait-init-jobs', true)).toBe(false);
  });`

I found `config.update('salesforcedx-vscode-apex.wait-init-jobs', false);` not working in this file

diyer:
@mingxuanzhangsfdx @peternhale I had few questions regarding this story -
- when would the user want to disable the config to not wait for the indexer? 
- how much does the indexer add to the start up time ?
- what if it fails, are there retries? and what if the final state is failure.

also, it will be good to have a test plan for this change as seems to involve a lot of state transition

Issue #5154, opened 2023-10-16
Aura Components Overrides VS Code basic Highlighting for PHP files


Hantonson:
### Summary

Aura Components extension on the Salesfroce Extension Pack for VS Code overrides PHP configuration, making impossible highlightinh html inside a php file .

### Steps To Reproduce:

1.  Install Saleforce Extension Pack extension on VS Code.
1.  Create a new file (Ctrl + n) and select language PHP
1.  Write "<?php echo "Hello, World!";   ?> <p>Hello, World!</p>

### Expected result

The php keywords "<?php", "echo" and "?> and the String "Hello, World!" should be highlighted. The html tags <p> and </p> should also be highlighted.

### Actual result

The php part of the code is highlighted correctly. However, the html tags are not, what makes the code creation and management difficult.

### Additional information
Code above:
![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/62595545/308240f3-c978-44d0-ae31-c8964c1c3800)

More HTML:

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/62595545/8b00d5f4-af99-49b4-825e-8404f8a27e6e)

With the Aura Components Extension disabled:

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/62595545/80f1fb6e-66e4-408a-a177-1d49fa9dc4ca)

**Salesforce Extension Version in VS Code**:

**SFDX CLI Version**:

**OS and version**:

**VS Code version**:

gbockus-sf:
This is the same issue found in https://github.com/forcedotcom/salesforcedx-vscode/issues/1988 which has an associated WI (W-10934996)

I tracked it down the the [html grammar file](https://github.com/forcedotcom/salesforcedx-vscode/blob/93df86c8ff378182c0fdcf98fdf090c790aad051/packages/salesforcedx-vscode-lightning/syntaxes/html.tmLanguage.json#L2) in the lighting (aura) extension, which hasn't been altered in a very long time. 

Turned out this was a scope of the grammar issue.  I found a simple fix and opened a PR https://github.com/forcedotcom/salesforcedx-vscode/pull/5159

gbockus-sf:
PR merged.  Will be included in the next release 👍

Hantonson:
Thanks a lot!

PR #5159, opened 2023-10-17
fix: update aura html grammar scope

gbockus-sf:
### What does this PR do?
The html grammar in the Aura extension had a three part name that forced it to take precedence when looking at html associated files.  (text.html.basic).  This has been reported as an issue a couple of times for php which has an html grammar scope of (html.php). 

I altered the scope name to be simpler (aura.html) and now php html files are correctly parsed and visually marked up by the php grammar file. 

@W-10934996@
#5154
#1988

### Functionality Before
image

### Functionality After
Aura html file still works properly
image

php files also works properly 
image

### DIFF
--- a/packages/salesforcedx-vscode-lightning/package.json
+++ b/packages/salesforcedx-vscode-lightning/package.json
@@ -116,7 +116,7 @@
     "grammars": [
       {
         "language": "html",
-        "scopeName": "text.html.basic",
+        "scopeName": "aura.html",
         "path": "./syntaxes/html.tmLanguage.json"
       }
     ],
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-lightning/package.json
```diff
@@ -116,7 +116,7 @@
     "grammars": [
       {
         "language": "html",
-        "scopeName": "text.html.basic",
+        "scopeName": "aura.html",
```
gbockus-sf, line 119: By making this name have 2 parts instead of 3 it keeps it from overriding the styles incorrectly for php html.

PR #5160, opened 2023-10-17
feat: add detection of orphaned apex ls at startup

peternhale:
Adds the ability for the apex extension to detect orphaned Apex language servers and then
* alerts the user of the orphaned server
* allows the user to show those process in the Apex output channel
* allows the user to terminate the orphaned processes
* allows the use to ignore the orphaned processes

The check and handling of orphaned processes is independent of the Apex language server startup for the current workspace.

@W-14046348@


gbockus-sf (approved):
Great stuff @peternhale 

I liked the approach of defensive coding around the windows command line callout. 

[QE](https://salesforce.quip.com/WlLPAzrvFTyU#temp:C:NWbb230817c95dd46ea95f09ebb7): 
* Verify dialog shows up if there are orphans detected ✅
* Verify The terminate path. ✅
* Verify the skip path ✅
* Verify Show processes ✅
* Verify no dialog if there are no orphans ✅
* Windows
    * Turns out it’s a lot harder to create Orphaned processes on window.  I tried killing every level of the parent stack with a force flag on windows and never ended up getting an orphan process.  After chatting with Pete we decided that we’ll move forward and track usage in appinsights.

Issue #5156, opened 2023-10-11
Modal Message Displays Placeholder Text


jjsfdc:
**Describe the bug**
A modal message appears with placeholder text

**To Reproduce**
1. In Code Builder, close your project with "Hamburger" -> File -> Close Folder
2. Click the blue "Connect an Org" button on the screen that appears
3. Modal message appears in right lower corner, "`!!! MISSING LABEL !!! predicates_no_folder_opened_t...`"
If you expand the modal, the full message appears as:
```
!!! MISSING LABEL !!!
predicates_no_folder_opened_text

Source: Salesforce CLI Integration (Extension)
```

**Expected behavior**
Correct user-facing text would appear in the modal message view.

**Screenshots**
If applicable, add screenshots to help explain your problem.  They really help!

This is the message, as it first appears:
<img width="1929" alt="open-org-1" src="https://github.com/forcedotcom/try-code-builder-feedback/assets/12158392/564b5f54-97b8-4347-9efb-b95c406d1152">

This is the message, expanded:
<img width="1930" alt="open-org-2" src="https://github.com/forcedotcom/try-code-builder-feedback/assets/12158392/695d531b-5735-4d90-9cb4-2f07e404abec">

**Desktop (please complete the following information):**
 - OS: any
 - Browser: any

**Smartphone or other device (please complete the following information):**
N/A

**Additional context**
N/A

gbockus-sf:
@smaddox-sf I fired up the latest from code build to understand what's happening here and I'm not sure where to go with it.

The CLI extension is getting activated by the sfdx.org.login.web command (SFDX: Authorize an Org), but that command is only valid when in an existing SF project

```
        {
          "command": "sfdx.org.login.web.dev.hub",
          "when": "sfdx:project_opened"
        },
```

So it's an odd state where I can't try to repo as it's not a valid state to run the command.  
The obvious change to get around this issue would be to have the button on the default page create or open a project as that's the usual first step for users that have just installed the extensions. 

Let me know how you'd like to move forward on this issue.

gbockus-sf:
Ok so I started testing and did validate that that message is missing from the core extension.  PR incoming to address the missing label problem.

gbockus-sf:
Fixed!

PR #5162, opened 2023-10-17
fix: add missing label

gbockus-sf:
### What does this PR do?
CodeBuilder found a flow where we had a missing label.  I verified by altering the local code to always print the label and found it was indeed missing from the core extension.  

@W-14284449@
#5156

### Functionality Before
image

### Functionality After
image


peternhale (approved):

diyer:
@gbockus-sf how is this change tested? can automation be added for this?

gbockus-sf:
@diyer This is an odd corner case as it's only exposed when an external extension calls a command in our extension.  This is fairly unusal (currently anyway).  
We'd typically catch this through QE or our automated testing, but the lack of that taking place in Jan 2021 when the problem code was written is a problem we're working on :).

PR #5163, opened 2023-10-17
Mingxuanzhang/248 synchronize init tasks

mingxuanzhangsfdx:
### What does this PR do?

To add an option on if apex language client should wait until init jobs including indexing are done. Default is to wait.

@W-14194698@

### Functionality Before

Client does not wait for initializing jobs including indexing so it could send requests to Apex LS before indexing is done, which causes mess like DB deadlock.

### Functionality After

Client will not send request until initializing jobs including indexing are done. And this setting could also be switched back through Salesforce Apex Configuration in Settings tab of VS Code.


packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,24 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+
+    // The listener should be set after languageClient is ready
+    // Language client will get notified once async init jobs are done
+    languageClient.onNotification('indexer/done', async () => {
+      await getTestOutlineProvider().refresh();
+      languageServerReady();
+    });
+
+    // client will check status with server after initialization
+    const indexerStatus = await languageClient.sendRequest('indexer/status');
```
peternhale, line 325: I would like to "hide" the details of new endpoints. What do you think about adding a new function in ApexLanguageClient that provides access to the endpoint, i.e. ApexLanguageClient#getServerState() and possible a convienence function of ApexLanguageClient#isRunning(). This will make the call available to others that might be concerned with the language server state.

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,26 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+      // The listener should be set after languageClient is ready
+      // Language client will get notified once async init jobs are done
+      languageClient.onNotification(API.doneIndexing, async () => {
+        await getTestOutlineProvider().refresh();
+        languageServerReady();
+      });
+    } else {
+      // Client will check status with server after initialization (only for sync mode)
+      // In async mode, indexerStatus is never RUNNING at the point
+      const indexerStatus = await languageClient.sendRequest(API.indexerStatus);
```
peternhale, line 328: @mingxuanzhangsfdx Given
* we reach here only in sync mode
* which means indexer should have done its work (running)

Why do we need to verify that it is in running?

peternhale (changes requested):

peternhale (approved):

packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -10,6 +10,7 @@
   "java_home_description": "Specifies the folder path to the Java 11 or Java 17 runtime used to launch the Apex Language Server (for example, /Library/Java/JavaVirtualMachines/openjdk-11.jdk/Contents/Home).",
   "java_memory_description": "Specifies the amount of memory allocation to the Apex Language Server in MB, or null to use the system default value.",
   "apex_semantic_errors_description": "Allow Apex Language Server to surface semantic errors",
+  "apex_wait_init_jobs_description": "Allow Apex Language Client not to send requests until the initializing jobs such as indexing are done.",
```
gbockus-sf, line 13: @sbudhirajadoc Can we set signoff on this setting text? 
I personally lean towards 
`Do note allow Apex Language Client to send requests until after initialization.`

I think we have an experimental section.  I kinda think this goes under there as they user will be odd results if they turn it off.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -185,3 +186,7 @@ function isSOQLExtensionInstalled() {
   const soqlExtension = vscode.extensions.getExtension(soqlExtensionName);
   return soqlExtension !== undefined;
 }
+
+export const enableSyncInitJobs: boolean = vscode.workspace
```
gbockus-sf, line 190: This logic should go in [settings.ts](https://github.com/forcedotcom/salesforcedx-vscode/blob/399189a6ea09ef9e963f152fe0011509d8d64d77/packages/salesforcedx-vscode-apex/src/settings.ts#L11)

The languageServer shouldn't be in the business of reading vscode settings.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,22 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+      // The listener should be set after languageClient is ready
+      // Language client will get notified once async init jobs are done
+      languageClient.onNotification(API.doneIndexing, async () => {
+        await getTestOutlineProvider().refresh();
+        languageServerReady();
+      });
+    } else {
+      // indexer must be running at the point
+      await getTestOutlineProvider().refresh();
+      languageServerReady();
```
gbockus-sf, line 328: So previously we always waited on the ready notification to execute this logic. Now we'll be executing it prior to the doneIndexing event.  Won't this set the status to ready, then have it immediately set it back to indexing on line 335 below? 

It seems like this piece of the logic should go below setting the status to indexing.

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
```diff
@@ -0,0 +1,33 @@
+import * as sinon from 'sinon';
+import * as vscode from 'vscode';
+
+describe('Should synchronize init jobs in server by default', () => {
+  const sandbox = sinon.createSandbox();
+  beforeEach(() => {
+    sandbox.stub(vscode.workspace, 'getConfiguration').returns({
+      get: () => true
+    });
```
gbockus-sf, line 9: sandboxes and sinon are are a mocha thing.  You shouldn't use either of those in jest tests.

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
```diff
@@ -0,0 +1,33 @@
+import * as sinon from 'sinon';
```
gbockus-sf, line 1: So your jest unit test files should 1-1 map to the code under test.  I would expect to see a `jest/languageServer.test.ts` file in this PR with tests the cover the code additions there.

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
gbockus-sf, line 9: Testing default values of vscode setting is outside the scope of what a unit test should accomplish.

packages/salesforcedx-vscode-apex/test/jest/views/settings/configurations.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    sandbox.restore();
+    sandbox.stub(vscode.workspace, 'getConfiguration').returns({
+      get: () => false
+    });
+    const enableSyncInitJobs: boolean = vscode.workspace
+      .getConfiguration()
+      .get<boolean>('salesforcedx-vscode-apex.wait-init-jobs', true);
+
+    expect(enableSyncInitJobs).toBe(false);
+  });
```
gbockus-sf, line 32: I think these tests are going to be something you do on settings.test.ts after you move that logic out of the languageServer.  Although in that case you'll be mocking the vscode.workspace.getConfuration.get response to just ensure it does what we expect with what vscode gives us back (note all of these are mocked responses in jest unit tests)

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -315,6 +312,22 @@ async function createLanguageClient(extensionContext: vscode.ExtensionContext) {
...
+    }
```
gbockus-sf, line 329: We need jest unit test coverage for this update.

gbockus-sf (dismissed):
Nice first extension PR 👍 
Have a few requests in there.  Please hit me up if I can clarify the approach on jest unit testing.

scripts/setup-jest.ts
```diff
@@ -91,8 +133,41 @@ const getMockVSCode = () => {
...
+    },
+    SymbolInformation: class {
+      public constructor(name: string, kind: any, range: Range, uri?: Uri, containerName?: string) {}
+    },
+    InlayHint: class {
+      public constructor(position: any, label: any, kind?: any) {}
+    },
+    CancellationError: class {
+      public constructor() {}
+    },
```
gbockus-sf, line 168: Is there a reason all these don't just point to a jest.fn()?  One downside of having a bunch of classes in here is if/when the VSCode API changes we'll have to stay in sync.

scripts/setup-jest.ts
```diff
@@ -6,6 +6,50 @@ class EventEmitter {
...
+const mockExtension = {
+  id: '1123',
+  extensionUri: Uri,
+  extensionPath: '123',
+  isActive: true,
+  packageJson: jest.fn(),
+  extensionKind: jest.fn(),
+  exports: jest.fn(),
+  activate: jest.fn()
+};
```
gbockus-sf, line 51: hum...I'm a little worried about having these be outside teh getMockVSCode function.  Do this jest mocks get reset for each load of the mock vscode module?  I'll verify in QE

gbockus-sf (commented):
Good stuff.  I'll QE later this evening

packages/salesforcedx-vscode-apex/test/jest/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { extensionUtils } from '../../src/languageUtils/extensionUtils';
+
+describe('indexDoneHandler', () => {
+  let setStatusSpy: jest.SpyInstance;
+  let onNotificationSpy: jest.SpyInstance;
+  let mockLanguageClient: any;
+  let setClientReadySpy: jest.SpyInstance;
+
+  beforeEach( () => {
+    setStatusSpy = jest.spyOn(languageClientUtils, 'setStatus').mockImplementation(jest.fn());
```
gbockus-sf, line 13: I don't think you need to use `mockImplementation` here.  It can just be `jest.spyOn(languageClientUtils, 'setStatus').mockReturnValue(undefined);` This is true for all .mockImplementation calls in this test. 

mockImplementation should only be used when you need to execute some code when a mock'd function is called.

gbockus-sf: One important note.  If you don't set `mockReturnValue` or for async `mockResolvedValue` for mocks it will call through to the original function that is mocked.

packages/salesforcedx-vscode-apex/test/jest/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    expect(setStatusSpy).toHaveBeenCalled();
+    expect(setStatusSpy).toHaveBeenCalledWith(1, '');
+    expect(onNotificationSpy).toHaveBeenCalled();
+    expect(onNotificationSpy).toHaveBeenCalledWith(API.doneIndexing, expect.any(Function));
+  });
+
+  it('should call setClientReady when enableSyncInitJobs is true', async () => {
+    await index.indexerDoneHandler(true, mockLanguageClient as any);
+    expect(setClientReadySpy).toHaveBeenCalled();
+  });
```
gbockus-sf, line 36: woot tests look great

packages/salesforcedx-vscode-apex/package.nls.json
diyer: > I kinda think this goes under there as they user will be odd results if they turn it off.
@gbockus-sf would it make sense to add this disclaimer somewhere?

diyer:
it will be good to test out these changes thoroughly and not release them this week. thoughts?

packages/salesforcedx-vscode-apex/package.nls.json
sbudhirajadoc: If this is a checkbox, does checking the box stop requests from being sent to it or the other way around?  Two choices, depending on the checkbox logic. 

1. Don't allow requests to the Apex Language Server until its done initializing. OR
2. Allow requests to the Apex Language Server only until its done initializing.

packages/salesforcedx-vscode-apex/test/jest/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should call languageClientUtils.setStatus and set up event listener when enableSyncInitJobs is false', async () => {
+    await index.indexerDoneHandler(false, mockLanguageClient as any);
+    expect(setStatusSpy).toHaveBeenCalled();
+    expect(setStatusSpy).toHaveBeenCalledWith(1, '');
```
gbockus-sf, line 28: For these two expect just having the `toHaveBeenCalledWith` is sufficient.

CristiCanizales:
QE ✅ following [test plan](https://salesforce.quip.com/3Vg1AGMQGahp)

packages/salesforcedx-vscode-apex/src/languageUtils/extensionUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+  ClientStatus,
+  languageClientUtils
+} from './index';
+
+const setClientReady = async (languageClient: ApexLanguageClient, languageServerStatusBarItem: ApexLSPStatusBarItem) => {
+  await getTestOutlineProvider().refresh();
+  languageServerStatusBarItem.ready();
+  languageClientUtils.setStatus(ClientStatus.Ready, '');
+  languageClient?.errorHandler?.serviceHasStartedSuccessfully();
+};
```
gbockus-sf, line 14: do we have unit test coverage for this?

scripts/setup-jest.ts
gbockus-sf: turns out they do get reset so good to go 👍

packages/salesforcedx-vscode-apex/src/settings.ts
```diff
@@ -13,3 +13,7 @@ export function retrieveTestCodeCoverage(): boolean {
     .getConfiguration(SFDX_CORE_CONFIGURATION_NAME)
     .get<boolean>('retrieve-test-code-coverage', false);
 }
+
+export const enableSyncInitJobs: boolean = vscode.workspace
```
gbockus-sf, line 17: This should be wrapped in a function like `retrieveTestCodeCoverage`.  Otherwise doesn't it run when the module it loaded?

gbockus-sf: also we should write a unit test for it

gbockus-sf (commented):
Since we're waiting on a decision for the setting I had a couple more requests for unit test coverage and I think we're good to go.

packages/salesforcedx-vscode-apex/test/jest/settings.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    } as any);
+
+    const result = retrieveEnableSyncInitJobs();
+    expect(result).toBe(true);
+    expect(getConfigurationMock).toHaveBeenCalledWith();
+    expect(getFn).toHaveBeenCalledWith(
+      'salesforcedx-vscode-apex.wait-init-jobs',
+      true
+    );
+  });
```
gbockus-sf, line 47: nice

gbockus-sf (approved):
Great stuff.  Did QE around the setting changes.  I think you're good to go. 

If you want bonus points it's be nice to create a doc in the contributing dir to track these settings that exists that aren't in the package.json.  Just the name and how it's used to accomplish.

PR #5174, opened 2023-10-20
chore: use node 18 for e2e test workflow

daphne-sfdc:
Update runE2ETest.yml to download Node 18 instead of Node 16 so that E2E tests can be run after the upgrade to Node 18.

@W-13209367@

### DIFF
--- a/.github/workflows/runE2ETest.yml
+++ b/.github/workflows/runE2ETest.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
         nodeVersion:
-          - 16.15.1
+          - 18.11.0
         vscodeVersion:
           - stable
     steps:


CristiCanizales:
Run: [End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/6593314130)

.github/workflows/runE2ETest.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
         nodeVersion:
-          - 16.15.1
+          - 18.11.0
```
gbockus-sf, line 23: can we remove the matrix thing since it's just the one version and use the nvmrc to pick the right version?       - ```uses: actions/setup-node@v3
        with:
          node-version-file: '.nvmrc'``` from `buildAll.yml`

gbockus-sf: actually nevermind ...with the duel repo thing it gets complicated

gbockus-sf (approved):
LGTM.  Can you update the description of this PR to just use the Node18 upgrade WI W-13209367

PR #5180, opened 2023-10-23
Run e2e against vsixes

CristiCanizales:
### What does this PR do?
Refactors e2e testing framework to test against vsixes and not against the source code.
- Unschedules e2e workflows and makes e2e.yml run after Nightly Build Develop runs, which creates the vsixes e2e.yml tests. This can also be triggered on demand for a given branch
- Removes file testSlackWorkflow.yml which is not used
- Updates runE2ETest.yml to use vsixes from given run instead of source code

@W-14182445@

### Functionality Before
- Testing against source code

### Functionality After
- Testing against vsixes from given run

## Notes:
This will go in together with the PR in the e2e repo: https://github.com/forcedotcom/salesforcedx-vscode-automation-tests/pull/53


.github/workflows/buildAll.yml
```diff
@@ -1,5 +1,17 @@
...
+  workflow_dispatch:
+    inputs:
+      branch:
+        type: string
+        required: false
+        default: 'develop'
+      label:
+        type: string
+        required: false
+        default: 'VS Code Extensions'
```
gbockus-sf, line 14: Instead of overloading the buildAll Action let's create a separate `nightlyE2EDevelop.yml` file where we have the cron and the two jobs: buildAll & e2e-tests

This action should have one job.

.github/workflows/coreE2E.yml
```diff
@@ -1,8 +1,6 @@
 name: Core End to End Tests
 
 on:
```
gbockus-sf, line 3: good stuff on these changes 👍

.github/workflows/runE2ETest.yml
```diff
@@ -29,6 +33,17 @@ jobs:
         with:
           path: ./salesforcedx-vscode
           ref: ${{ github.event.ref }}
+      - name: Download extension vsixes
+        run: |
+          mkdir ./extensions
+          pwd
+          gh run download ${{ inputs.runId }} -D ./extensions
```
gbockus-sf, line 40: Is runId a GHA thing or did we create the term?

.github/workflows/testBuildAndRelease.yml
```diff
@@ -21,9 +21,10 @@ jobs:
   build-and-test:
     uses: ./.github/workflows/buildAndTest.yml
     needs: [get-version]
-    with: 
+    with:
       branch: 'main'
       label: ${{ needs.get-version.outputs.RELEASE_VERSION }}
+      runId: ${{ github.run_id }}
```
gbockus-sf, line 27: oh there is it.  
Is this a sha or what?  should we prefix it with the branch or something to make it more easily identifiable?

gbockus-sf (commented):
Good stuff.  Couple of suggestions for making runId identifiable and pulling the nightly logic out of the buildAll GHA.

.github/workflows/testBuildAndRelease.yml
CristiCanizales: We need only the id of the run that created the vsixes to then use it with `gh run download` command, so we can't add anything else

gbockus-sf: I think we can name it whatever we use if we use the gh cli flag `gh run download -n <name1>`. I'm not that hung up on it.  If it's a problem we can address later

.github/workflows/nightlyE2EDevelop.yml
```diff
@@ -0,0 +1,26 @@
...
+      label:
+        type: string
+        required: false
+        default: 'VS Code Extensions'
+jobs:
+  build-all:
+    uses: ./.github/workflows/buildAll.yml
+    with:
+      branch: ${{ inputs.branch || 'develop' }}
+      label: ${{ inputs.label || 'Nightly' }}
```
gbockus-sf, line 20: Nit: can we make this more specific 'Nightly E2E' maybe?

gbockus-sf (approved):
Good stuff.  Thanks for the updates 👍 

I see this as a wait and see if it works in GHA from a QE perspective.  Lemme know if you want me to validate it anyway before merge.

.github/workflows/runE2ETest.yml
gbockus-sf: Just calling out for future us using gh here means the run with the artifacts has to have already been completed.

gbockus-sf (approved):
🚢

PR #5182, opened 2023-10-25
chore: fix soql LSP location for vsix and fix GHA for node 18

gbockus-sf:
### What does this PR do?
With the node18 upgrade we moved to npm workspaces which changed where dependencies  are located in the project.  For development they are pulled up to the root node_modules folder when shared across packaged, but still have to be in the package node_modules folder when running as a VSIX. 

I missed this update for ensuring we get the SOQL LSP from the right location in vsix. 

GHA updates: 
The node LST version changed from 18->20 on 10/24/2023.  
Updated our GHA that were using lts/* to use the nvm file for getting the node version

Update the created release branch GHA to pin the npm and lerna version to the same version as in the repo. 

@W-13209367@

### Functionality Before
SOQL LSP fails to start when on used via vsix. 

### Functionality After
SOQL LSP behaves normally when used both in dev and vsix. 
GHA use node 18
create release branch GHA works

### DIFF
--- a/.github/workflows/createReleaseBranch.yml
+++ b/.github/workflows/createReleaseBranch.yml
@@ -45,9 +45,9 @@ jobs:
         with:
           email: ${{ secrets.IDEE_GH_EMAIL }}
       - name: Set NPM at the correct version for Lerna
-        run: npm install -g npm@8.12.1
+        run: npm install -g npm@@9.8.1
       - run: npm ci
-      - run: npm install -g shelljs && npm install -g lerna
+      - run: npm install -g shelljs && npm install -g lerna@5.5.4
       - name: Create and Push the Release Branch
         id: create_step
         run: |
--- a/packages/salesforcedx-vscode-soql/package.json
+++ b/packages/salesforcedx-vscode-soql/package.json
@@ -118,6 +118,13 @@
     ],
     "packageUpdates": {
       "main": "dist/index.js",
+      "serverPath": [
+        "node_modules",
+        "@salesforce",
+        "soql-language-server",
+        "lib",
+        "server.js"
+      ],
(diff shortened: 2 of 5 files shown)


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -118,6 +118,13 @@
     ],
     "packageUpdates": {
       "main": "dist/index.js",
+      "serverPath": [
+        "node_modules",
+        "@salesforce",
+        "soql-language-server",
+        "lib",
+        "server.js"
+      ],
```
gbockus-sf, line 127: This updates where we get the server file from when running as a vsix.  You can see the dev version in line 100 of this file.

.github/workflows/createReleaseBranch.yml
```diff
@@ -45,9 +45,9 @@ jobs:
         with:
           email: ${{ secrets.IDEE_GH_EMAIL }}
       - name: Set NPM at the correct version for Lerna
-        run: npm install -g npm@8.12.1
+        run: npm install -g npm@@9.8.1
       - run: npm ci
-      - run: npm install -g shelljs && npm install -g lerna
+      - run: npm install -g shelljs && npm install -g lerna@5.5.4
```
gbockus-sf, line 50: Can see identical changes and passing GHA run in my fork 
https://github.com/forcedotcom/salesforcedx-vscode/commit/da5f94338e62fd0ced8652fe26384aa8ccfdfc68

https://github.com/gbockus-sf/salesforcedx-vscode/actions/runs/6643056792

CristiCanizales (approved):
Thanks for these fixes!
🚢

PR #5185, opened 2023-10-26
feat: validate CLI installation and version during activation of CLI Integration extension

daphne-sfdc:
### What does this PR do?
This PR includes new functionality to check whether a supported CLI version is installed on the user's computer.  There are 4 main cases that are being caught: (1) CLI not installed (2) Outdated version of SFDX CLI that is incompatible with our extensions (3) Only SF v1 is installed (4) Both SFDX v7 and SF v2 are installed.  In each of those 4 cases, the CLI Integration extension will not activate, and the user will not be able to see SFDX commands in the command palette.  The user will receive an error notification about the unsupported/missing CLI when the activation of the CLI Integration extension fails.

@W-14196710@

### Functionality Before
As a result of the recent switch from SFDX -> SF-style variables to get rid of warnings in the Output tab, users with a CLI version below SFDX v7.193.2 are no longer able to use the extensions because older CLI versions do not support SF-style variables.  The extensions will activate, but the older version is caught at runtime.
Screenshot 2023-10-24 at 7 00 13 PM

### Functionality After
If the user does not have CLI installed on their computer or their installed CLI version is no longer supported, they will get an error notification when they first create/open a Salesforce Project and the CLI Integration extension will not activate.  Based on the case, one of these 4 error messages will pop up:
Screenshot 2023-11-28 at 9 37 21 AM


packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
```diff
@@ -250,6 +250,8 @@ export const messages = {
     'sfdx-project.json ファイル (%s) をパースできませんでした。パースエラー: %s',
   sfdx_cli_not_found:
     'Salesforce CLI がインストールされていません。[%s](%s) からインストールしてください。',
```
daphne-sfdc, line 252: I updated the English error message in i18n.ts but the Japanese error message is still the old one. @sbudhirajadoc

packages/salesforcedx-vscode-core/src/messages/i18n.ja.ts
```diff
@@ -250,6 +250,8 @@ export const messages = {
...
+  sfdx_cli_not_supported:
+    '!!! Your installed CLI version is no longer supported. Please uninstall your CLI and reinstall the latest version using this link: [%s](%s)',
```
daphne-sfdc, line 254: This is a new error message.  The English version is in i18n.ts but I don't have a Japanese translation so I temporarily copied it over and added 3 exclamation points in front. @sbudhirajadoc

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -290,7 +290,9 @@ export const messages = {
   error_parsing_sfdx_project_file:
     "Couldn't parse sfdx-project.json file (%s). Parse error: %s",
   sfdx_cli_not_found:
-    'Salesforce CLI is not installed. Install it from [%s](%s)',
+    'Either Salesforce CLI is not installed or only SF v1 is present. Install the latest CLI version from [%s](%s)',
```
daphne-sfdc, line 293: I changed the error message because SF v1 does not map sf commands to sfdx and thus cannot be used by our extensions.  The isCliInstalled() function in cliConfiguration.ts uses "which sfdx" for the checking, which means having only SF v1 is caught at the same time as having no CLI installed.

packages/salesforcedx-vscode-core/src/messages/i18n.ts
CristiCanizales, line 1: As discussed on huddle, could you make these 2 messages a bit more similar for consistency?

CristiCanizales (changes requested):
LGTM :) Waiting on the tests for approval. Thanks Daphne! 🥳

packages/salesforcedx-vscode-core/src/messages/i18n.ts
daphne-sfdc: Fixed

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { SfdxCommandBuilder } from './sfdxCommandBuilder';
+import { CliCommandExecutor } from './cliCommandExecutor';
+import { CommandOutput } from './commandOutput';
+
+export class CheckCliVersion {
```
gbockus-sf, line 12: unit test me

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+        return 'validCli';
+      }
+      else {
+        return 'cliNotSupported';
+      }
+    }
+
+    // Case 3: Neither SFDX CLI nor SF CLI is installed
+    else {
+      return 'cliNotInstalled';
```
gbockus-sf, line 64: How about using an enum for these return values?   Can be used here, in the code using the return value, and in test

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (sfdxVersionNumber[0] >= 7 && sfdxVersionNumber[1] >= 193 && sfdxVersionNumber[2] >= 2) {
+        return 'validCli';
+      }
+      else {
+        return 'cliNotSupported';
+      }
+    }
+
+    // Case 2: SF CLI is installed
+    else if (sfMatch) {
```
gbockus-sf, line 50: have you run prettier against this?  The if/else with spaces in between seems....non-ideal.

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const sfdxMatch = sfdxPattern.exec(result);
+    const sfMatch = sfPattern.exec(result);
+
+    // Case 1: SFDX CLI is installed
+    if (sfdxMatch) {
+      const sfdxVersion = sfdxMatch[1];
+      const sfdxVersionNumber = sfdxVersion.split('.').map(Number);
+      // The last working version of SFDX is v7.193.2
+      if (sfdxVersionNumber[0] >= 7 && sfdxVersionNumber[1] >= 193 && sfdxVersionNumber[2] >= 2) {
```
gbockus-sf, line 41: Let's use constants (in this file is fine) for the cli version numbers.

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+        return 'cliNotSupported';
+      }
+    }
+
+    // Case 2: SF CLI is installed
+    else if (sfMatch) {
+      const sfVersion = sfMatch[1];
+      const sfVersionNumber = sfVersion.split('.').map(Number);
+      // SF v1 does not map sf commands to sfdx commands and is not supported
+      if (sfVersionNumber[0] >= 2 && sfVersionNumber[1] >= 0 && sfVersionNumber[2] >= 0) {
```
gbockus-sf, line 54: same story..constants for sf version numbers

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,67 @@
...
+        return 'validCli';
+      }
+      else {
+        return 'cliNotSupported';
+      }
+    }
+
+    // Case 3: Neither SFDX CLI nor SF CLI is installed
+    else {
```
gbockus-sf, line 63: Yeah I'd def go with 
`} else { // Case 3: ......`

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -555,6 +555,22 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
...
+    showCLINotInstalledMessage();
+  }
+  const cliVersionCheckResult = await new CheckCliVersion().validateCliVersion();
+  if (cliVersionCheckResult === 'cliNotSupported') {
+    showCLINotSupportedMessage();
+    throw new Error();
+  } else if (cliVersionCheckResult === 'cliNotInstalled') {
+    showCLINotInstalledMessage();
+    throw new Error();
+  }
```
gbockus-sf, line 572: So keeping all the logic out of the activation method is a good thing (think separation of concerns).  Can we refactor this to be a stand alone utility function so all we see here is `await validateDependancies();`?  It will throw if all deps aren't correct and resolve happily otherwise.

gbockus-sf: Also throwing empty errors is sadness (it's gets reported to telemetry).  We should throw a custom error or an error with a valid message

gbockus-sf (commented):
some feedback for your consideration

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
diyer: should we check both the versions instead of an else if for the issue we are discussing here https://salesforce-internal.slack.com/archives/C041ZPJH317/p1699052602748759?thread_ts=1699051206.563149&cid=C041ZPJH317

diyer: have we considered addressing having both sfdx and sf installed in this change?

daphne-sfdc: @diyer I can try to make the change now and revert if @gbockus-sf thinks it is unnecessary.

diyer: lets wait for @gbockus-sf suggestion Daphne. It would be nice to address that issue as well IMHO

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
daphne-sfdc: @gbockus-sf I just ran Prettier against that file and Prettier actually didn't complain about the if/else with spaces in between.  I tried switching to the formatting that you suggested in this comment but when I saved the file, Prettier moved the comment to the next line (see attached screenshot for Option 3).  What would you prefer that I do?

1. Keep the original formatting with the spaces between the if/else statements
Screenshot 2023-11-09 at 4 48 00 PM

2. Turn off Prettier for the file and reformat to what you suggested
Screenshot 2023-11-09 at 4 49 09 PM

3. Let Prettier auto-format the comments and keep them inside the brackets on the next line
Screenshot 2023-11-09 at 4 42 34 PM

CristiCanizales: I like either 1 or 3 🙂

CristiCanizales:
QE notes [here](https://salesforce.quip.com/5YyEA6PetzcG)

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,125 @@
...
+        sfdxCliVersionArray[2] >= minSFDXVersionArray[2]
+      ) {
+        return CheckCliEnum.validCli;
+      } else {
+        return CheckCliEnum.outdatedSFDXVersion;
+      }
+    }
+
+    // Case 5: SF v2 is installed
+    return CheckCliEnum.validCli;
```
daphne-sfdc, line 123: I didn't do another check for SF versions because for SF v2 anything above 2.0.0 is supported.  The case of only SF v1 has already been caught in Case 2.

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,125 @@
...
+      return CheckCliEnum.cliNotInstalled;
+    }
+
+    // Case 2: Only SF CLI (v1) is installed (SF v1 cannot be used because it does not map sf to sfdx)
+    if (sfdxCliVersionArray[0] === -1 && sfCliVersionArray[0] === 1) {
+      return CheckCliEnum.onlySFv1;
+    }
+
+    // Case 3: Both SFDX CLI (v7) and SF CLI (v2) are installed at the same time
+    if (sfCliVersionArray[0] >= minSFVersionArray[0] && !sfdxCliVersionArray.every((val, index) => val === sfCliVersionArray[index])) {
```
daphne-sfdc, line 105: This is checking that the SF version is v2 and the SFDX version is not the same as the SF version.  Because SF v2 maps sf to sfdx, the results of `sfdx --version` and `sf --version` should be identical.  If they are different, that means the user has both SFDX v7 and SF v2 installed simultaneously.

CristiCanizales (changes requested):
LGTM 😄

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
diyer: @daphne-sfdc are we not taking care of both sfdx and sf installed?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -690,3 +693,41 @@ export function deactivate(): Promise<void> {
...
+  } else if (cliInstallationResult === CheckCliEnum.onlySFv1) {
+    showErrorNotification('sf_v1_not_supported', [SFDX_CLI_DOWNLOAD_LINK, SFDX_CLI_DOWNLOAD_LINK]);
+    throw Error('Only SF v1 installed');
+  } else if (cliInstallationResult === CheckCliEnum.outdatedSFDXVersion) {
+    showErrorNotification('sfdx_cli_not_supported', [SFDX_CLI_DOWNLOAD_LINK, SFDX_CLI_DOWNLOAD_LINK]);
+    throw Error('Outdated SFDX CLI version that is no longer supported');
+  } else if (cliInstallationResult === CheckCliEnum.bothSFDXAndSFInstalled) {
+    showErrorNotification('both_sfdx_and_sf', []);
+    throw Error('Both SFDX v7 and SF v2 are installed');
+  } else {
```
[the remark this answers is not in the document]
diyer: +1

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -699,3 +702,41 @@ export function deactivate(): Promise<void> {
...
+    throw Error('No Salesforce CLI installed');
+  } else if (cliInstallationResult === CheckCliEnum.onlySFv1) {
+    showErrorNotification('sf_v1_not_supported', [SFDX_CLI_DOWNLOAD_LINK, SFDX_CLI_DOWNLOAD_LINK]);
+    throw Error('Only SF v1 installed');
+  } else if (cliInstallationResult === CheckCliEnum.outdatedSFDXVersion) {
+    showErrorNotification('sfdx_cli_not_supported', [SFDX_CLI_DOWNLOAD_LINK, SFDX_CLI_DOWNLOAD_LINK]);
+    throw Error('Outdated SFDX CLI version that is no longer supported');
+  } else if (cliInstallationResult === CheckCliEnum.bothSFDXAndSFInstalled) {
+    showErrorNotification('both_sfdx_and_sf', []);
+    throw Error('Both SFDX v7 and SF v2 are installed');
```
diyer, line 730: are these user facing errors? if yes then can they also suggest the right course of action to take ?

daphne-sfdc: No, those are for telemetry purposes and are not user facing.  The user will see the error in the **showErrorNotification()** message, which you can find the text in **i18n.ts**.

packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -14,7 +14,7 @@ export const APEX_CODE_DEBUG_LEVEL = 'FINEST';
 export const VISUALFORCE_DEBUG_LEVEL = 'FINER';
 export const ENV_SF_DISABLE_TELEMETRY = 'SF_DISABLE_TELEMETRY';
 export const SFDX_CLI_DOWNLOAD_LINK =
```
diyer, line 16: nit: can this variable name be also updated to SF_CLI_...?

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,125 @@
...
+      return CheckCliEnum.bothSFDXAndSFInstalled;
+    }
+
+    // Case 4: SFDX CLI is installed - need to validate if the version is above the minimum supported version
+    if (sfCliVersionArray[0] < minSFVersionArray[0] && sfdxCliVersionArray[0] !== -1) {
```
diyer, line 110: what is this check for sfCliVersionArray[0] < minSFVersionArray[0] ?

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (sfdxCliVersionArray[0] === -1 && sfCliVersionArray[0] === -1) {
+      return CheckCliEnum.cliNotInstalled;
+    }
+
+    // Case 2: Only SF CLI (v1) is installed (SF v1 cannot be used because it does not map sf to sfdx)
+    if (sfdxCliVersionArray[0] === -1 && sfCliVersionArray[0] === 1) {
+      return CheckCliEnum.onlySFv1;
+    }
+
+    // Case 3: Both SFDX CLI (v7) and SF CLI (v2) are installed at the same time
```
diyer, line 104: cant this be simplified to following cases:
1. sfdx and sf both not installed
2. sf v1 installed
3. if sfdx installed, check if sfdx version is supported
4. if sf v2 is installed, check sf version is supported

daphne-sfdc: Resolved during our 1:1 on Monday 12/4.

Summary of the discussion: Case 3 (Both SFDX CLI (v7) and SF CLI (v2) are installed at the same time) is necessary for semantic reasons.  If the user has an outdated version of SFDX v7 that they installed simultaneously with a valid version of SF v2, the VSCode Terminal prefers the outdated version of SFDX v7, even though it is incompatible with the extensions.  If we remove Case 3, the user will see the error message that they have an outdated version of SFDX but get confused when they log a Github issue and tell us that they have SF v2 yet the extensions are not working.  This extra check for semantics is to remove that confusion by giving a clear error message to the user about what is happening with their system.

sbudhirajadoc (commented):
Update based on discussions here -- https://docs.google.com/document/d/1lE-xJMhEXfcrenaTU2LYSR6HB0HnGP6P-VW7KfPZ0tI/edit

sbudhirajadoc (approved):

packages/salesforcedx-vscode-core/src/util/cliConfiguration.ts
```diff
@@ -34,8 +34,8 @@ export function isCLIInstalled(): boolean {
 export function showCLINotInstalledMessage() {
   const showMessage = nls.localize(
     'sfdx_cli_not_found',
```
CristiCanizales, line 36:
```suggestion
    'salesforce_cli_not_found',
```
I think we should update this too

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -541,6 +543,11 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
   // thus avoiding the potential errors surfaced when the libs call
   // process.cwd().
   ensureCurrentWorkingDirIsProjectPath(rootWorkspacePath);
+  const platform = process.platform.toLowerCase();
+  const isWindows = platform === 'win32';
+  if (!isWindows) {
```
CristiCanizales, line 548: I'm not sure about this 💀 because we have so many users on windows... and we would be 100% skipping the validation for them. Is this something we want? It seems like half a feature

packages/salesforcedx-utils/test/jest/cli/checkCliVersion.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  const noSF_string = 'No SF CLI';
+
+  const sfV2_parsed = '2.15.9';
+  const sfdxV7_valid_parsed = '7.209.6';
+  const sfV1_parsed = '1.87.0';
+  const sfdxV7_outdated_parsed = '7.183.1';
+  const noSFDX_parsed = '0.0.0';
+  const noSF_parsed = '0.0.0';
+
+  it('Should create instance.', () => {
```
peternhale, line 26: This test is not necessary given class does not have a formal constructor

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -707,3 +714,44 @@ export function deactivate(): Promise<void> {
   disposeTraceFlagExpiration();
   return turnOffLogging();
 }
+
+export async function validateCliInstallationAndVersion(): Promise<void> {
```
peternhale, line 718: This function does not need promises

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export enum CheckCliEnum {
+  validCli = 1,
+  outdatedSFDXVersion = 2,
+  onlySFv1 = 3,
+  cliNotInstalled = 4,
+  bothSFDXAndSFInstalled = 5
+}
+
+export class CheckCliVersion {
```
peternhale, line 19: Consider renaming this class w/o a verb in the name, such as CliVersion. Class names tend to have noun based names. Same would apply to the enum.

packages/salesforcedx-utils/test/jest/cli/checkCliVersion.test.ts
```diff
@@ -0,0 +1,130 @@
+/*
```
peternhale, line 1: I don't see tests for the getSfdxCliVersion or get SfCliVersion functions.

daphne-sfdc: I originally tried to write tests for getSfdxCliVersion() and getSfCliVersion() functions but found that they're dependent on the version installed on the computer and thus I can't put definitive values to check for.  For example, suppose I have SF v2.14.1 installed on my computer.  In order to get the tests to pass, both getSfdxCliVersion() and getSfCliVersion() need to check for SF v2.14.1.  But the CLI team made a release and the next person to run the tests has SF v2.15.0.  Because v2.14.1 ≠ v2.15.0, the second person has two failing tests, even though they have a valid CLI version.  That's just one example - a third person can have SFDX v7 installed, which is another case.  It's difficult to test getSfdxCliVersion() and getSfCliVersion() functions unless we just test that _something_ is returned, regardless of the value.

diyer: @daphne-sfdc cant you use mocks?

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
```diff
@@ -0,0 +1,71 @@
...
+      return CheckCliEnum.onlySFv1;
+    }
+
+    // Case 3: Both SFDX CLI (v7) and SF CLI (v2) are installed at the same time
+    if (semver.satisfies(sfCliVersionString, '2.x') && !semver.satisfies(sfdxCliVersionString, sfCliVersionString)) {
+      return CheckCliEnum.bothSFDXAndSFInstalled;
+    }
+
+    // Case 4: Outdated SFDX CLI version is installed
+    const minSFDXVersion = '7.193.2';
```
diyer, line 63: do we have to run this version number by cli team? I know you ran some manual tests to make sure this version is supported, but is that a valid way to test this is indeed the minimum supported version or is there other authoritative way.

packages/salesforcedx-utils/test/jest/cli/checkCliVersion.test.ts
daphne-sfdc: @diyer I didn't think of that.  Let me check the other unit test classes to see how they use mocks and try to implement mocking in this test class.

peternhale (changes requested):

packages/salesforcedx-utils/src/cli/checkCliVersion.ts
daphne-sfdc: @diyer 

Some background info -
In the last several SFDX v7 versions, the CLI team has been gradually updating the CLI to allow the use of SF-style commands.  For example, the SF-style command for authorizing an org is `sfdx org login web`.  If you have a CLI version that is SFDX v7.191.1 or newer, you are able to run this command.  If you have an older version of the CLI, you will get the error message `Warning: org login web is not a sfdx command.` and you are prevented from running the command.

Our VSCode extensions need to use these new SF-style commands because the CLI team is gradually deprecating the SFDX-style commands.  For example, in the latest version SF v2.21.8, the CLI team has removed support for the SFDX-style command for authorizing an org: `sfdx force auth login web`.  If you have SF v2.21.8 and try to run that command, you get the error message `Warning: force auth login web is not a sf command.` and you are blocked from running the command.  In order to support newer CLI versions, we must keep up with the updates from SFDX -> SF style.

To answer the question -
As we make more and more SFDX -> SF style changes in our VSCode extensions, the last supported CLI version will gradually increase.  From my testing of some common commands, such as deploy/retrieve, auth, run Apex tests, etc. the current last supported version is SFDX v7.193.2.  I suspect that this minimum version will eventually not be supported because we are seeing more PRs for SFDX -> SF changes, so it would be good to reevaluate this number once every few months to make sure it shouldn't be raised, and also to recommend to customers that they make the leap to SF v2.

If we are unsure about whether SFDX v7.193.2 is the correct minimum version and there is still the possibility of seeing SF-style commands not being supported at runtime, I would say it's better to be safe than sorry and choose a higher version than the what we think is the minimum.  So for example, if we think SFDX v7.193.2 _might_ be supported but we're not 100% sure, I would put the minimum as a higher version like v7.209.6 (the latest SFDX v7 version).  That way, users with older CLI versions will all receive the message that their CLI is no longer supported, and we force them to upgrade to SF v2.  If we do this, we're guaranteed to not run into any SFDX-style command issues but some customers will have CLI versions that tell them are outdated but which really don't break the system.  It's the tradeoff of prematurely forcing customers to upgrade to SF v2 vs. possibly still seeing runtime errors and getting Github issues about them.

As this is VSCode extensions-specific, I don't think we need to run this number by CLI team - we can read the release notes for SFDX here: https://github.com/forcedotcom/cli/blob/main/releasenotes/sfdx/README.md.  From my testing, I think the minimum supported version is v7.193.2 as I put in the code, but feel free to do further testing before this PR gets merged if you think it is necessary and/or we can come back and reevaluate this minimum version in a few months.

packages/salesforcedx-utils/src/cli/cliVersionStatus.ts
```diff
@@ -0,0 +1,76 @@
...
+  outdatedSFDXVersion = 2,
+  onlySFv1 = 3,
+  cliNotInstalled = 4,
+  bothSFDXAndSFInstalled = 5,
+  SFDXv7Valid = 6
+}
+
+export class CliVersionStatus {
+
+  public getSfdxCliVersion(): string {
```
peternhale, line 22: @daphne-sfdc this function and the next have the same logic, with diff in the command being run. See if this can be reduced to one function that manages the execSync and then use the current functions to call into the new function passing command to run.

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -292,6 +292,14 @@ export const messages = {
...
   sfdx_cli_not_found:
     'Salesforce CLI is not installed. Install it from [%s](%s)',
+  sfdx_cli_not_supported:
+    'Your installed Salesforce CLI version is no longer supported. Uninstall Salesforce CLI and install the latest version from [%s](%s).',
+  sf_v1_not_supported:
+    'Your installed Salesforce CLI version is incompatible with all versions of our extensions. Uninstall Salesforce CLI and install the latest version from [%s](%s).',
+  both_sfdx_and_sf:
+    'You have both SFDX(v7) and SF(v2) of Salesforce CLI installed. Uninstall SFDX(v7) by running `npm uninstall sfdx-cli --global` in your terminal.',
+  sfdx_v7_deprecation:
+    'Our team will remove support for SFDX(v7) of Salesforce CLI in a few months. If you are able, uninstall SFDX(v7) and install SF(v2) from [%s](%s).',
```
diyer, line 302: As we talked in slack, lets make the tone of this message bit subtle :) https://salesforce-internal.slack.com/archives/C054SJJAB24/p1703132241482919

CristiCanizales (approved):
All good from QA perspective both in mac and windows! ✅ 
After the text updates I'd say :shipit: 
Nice work Daphne! 🥳 
QA doc [here](https://salesforce.quip.com/5YyEA6PetzcG)

packages/salesforcedx-utils/src/cli/cliVersionStatus.ts
```diff
@@ -0,0 +1,73 @@
...
+      if (isSfdx) {
+        result = execSync('sfdx --version');
+      } else {
+        result = execSync('sf --version');
+      }
+      return result.toString();
+    } catch {
+      return 'No CLI';
+    }
+  }
```
peternhale, line 34:
``` suggestion
  public getCliVersion(isSfdx: boolean): string {
    try {
      const result = execSync(`${isSfdx ? 'sfdx' : 'sf'} --version`);
      return result.toString();
    } catch {
      return 'No CLI';
    }
  }
```

peternhale (changes requested):

peternhale (approved):

sbudhirajadoc:
The text for the message recommending the move to SF from SFDX is [here](https://docs.google.com/document/d/1lE-xJMhEXfcrenaTU2LYSR6HB0HnGP6P-VW7KfPZ0tI/edit).

Issue #5187, opened 2023-10-26
Release 59.2.0 Core extension fails to activate on Windows


gbockus-sf:
### Summary

In release test of the v59.2.0 release it was found that the core extension will not activate when: 
- On Windows 
- Using VSCode version 1.83.1 (works on 1.81.1) 

### Steps To Reproduce:
- Install core extension from the [v59.2.0](https://github.com/forcedotcom/salesforcedx-vscode/releases/tag/v59.2.0) Release 

### Additional information

![Screenshot 2023-10-26 at 12 46 20 AM](https://github.com/forcedotcom/salesforcedx-vscode/assets/76090802/1be4cc9a-d6fc-4f5f-8c77-4596187c5a3e)

**Salesforce Extension Version in VS Code**: v59.2.0

**SFDX CLI Version**: N/A

**OS and version**: Windows 

**VS Code version**: 1.83.1

gbockus-sf:
Tested in two other windows environments and was not found to be an issue.  Appears to be something odd about my environment on windows.

PR #5189, opened 2023-10-26
chore: fix core vsix on windows

gbockus-sf:
### What does this PR do?
Add the older jsforce as a dep of the core extension.  This is a sad work around to the issue found where VSCode 1.83.1 isn't loading modules correctly

Downsides: 
- one more install of jsforce in the core extension 
- Inflates the core extension by another 5mb or so

@W-14374832@ 
#5187

### Functionality Before
vsix fails to start on windows

### Functionality After
vsix works on windows


packages/salesforcedx-vscode-core/package.json
```diff
@@ -92,6 +92,7 @@
         "@salesforce/source-tracking": "4.2.14",
         "@salesforce/templates": "^58.0.1",
         "@salesforce/source-deploy-retrieve": "9.7.18",
+        "jsforce": "1.11.0",
```
gbockus-sf, line 95: Note this is in the packageUpdates property.  These are deps that are directly installed b/c they can't be bundled.  In this case we are ensuring that the version of sf-core in the @heroku/functions/core dependancy finds the right version of jsforce.

daphne-sfdc (approved):
Thanks for the workaround Gordon!

gbockus-sf:
ouch

gbockus-sf:
This was tested on other Windows and it appears to be an issue with my env.  Closing

PR #5203, opened 2023-11-02
fix: migrate from tslint to eslint

peternhale:
Migrates extensions to use eslint.

Modified deps to include needed eslint libraries
Created root eslint config to old common directives and rules
Created root rules for test lint (more relacxed)
Created package level rules in files named `fix-these-rules.json` that changes originally discovered errors to warnings.
Created package level `.eslingrc.json` files that extend base config and fix rules
Created package /test level `.eslintrc.json` files that extend package lint config and base test lint rules

@W-7210928@


base-eslintrc.json
```diff
@@ -0,0 +1,261 @@
+{
```
gbockus-sf, line 1: how would you feel about putting this in the configs directory since it's not getting picked up by default anyway?

common-test-lint-rules.json
```diff
@@ -0,0 +1,22 @@
+{
```
gbockus-sf, line 1: same story...configs directory

gbockus-sf (approved):
I like it.  I the plan to spawn a WI for each fix-these-rules file and handle individually? 

Still would be good to remove refs to tslint in the developers docs and recommended extensions 
image

base-eslintrc.json
peternhale: I like the idea. I will make the change

gbockus-sf (approved):
Thanks Pete!

every future dev of this repo thanks you.

QE: 
For this build task if it passes it passes.

PR #5204, opened 2023-11-02
fix: remove heroku functions support

gbockus-sf:
### What does this PR do?
Remove all commands and dependancies for supporting heroku functions.

This was done as part of the effort to improve the slow loading time of the extensions. It reduces the overall size of the core extension by around 25% and brings the total number of copies of jsforce/sfdx-core shipped with the extension to 1. 

@W-14418804@

### Functionality Before
There were command for interaction with Heroku functions.
Screenshot 2023-11-03 at 10 48 18 AM

### Functionality After
No longer support functions command 
Screenshot 2023-11-03 at 10 48 28 AM

30% decrease in size and even more then that in complexity 

### Doc Writer
- @sbudhirajadoc Need to update docs to removed Functions docs and references.  Change log should include recommendation to use CLI for functions or to stay on the previous version of the SF Extensions for VSCode


diyer:
@gbockus-sf regarding this , 

> Change log should include recommendation to use CLI for functions or to stay on the previous version of the SF Extensions for VSCode

will there be any downsides in asking the user to use previous version of SF extensions or should we just ask them to use cli for functions always?

gbockus-sf:
> @gbockus-sf regarding this ,
> 
> > Change log should include recommendation to use CLI for functions or to stay on the previous version of the SF Extensions for VSCode
> 
> will there be any downsides in asking the user to use previous version of SF extensions or should we just ask them to use cli for functions always?

@diyer The downside being that if they stick an old version of the extensions they will not longer continue to get the releases we push with updates.  It's really up to the end user.  The cli gives folks an option to work with functions outside of VSCode, but if we have some folks that love that functionality I just wanted to call out all the options.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -9,7 +9,7 @@ import {
   ChannelService,
   getRootWorkspacePath,
   SFDX_CORE_CONFIGURATION_NAME,
-  TelemetryService
+  TelemetryService,
```
gbockus-sf, line 12: hum...I was previously the eslint/prettier upgrade branch and I think a new formatter stuck around when I switched branched...I'll fix the comma issues.

mingxuanzhangsfdx (approved):
* The 5 functions and corresponding tests are deleted
* The 5 commands are clear

PR #5205, opened 2023-11-03
feat: generate jar

mingxuanzhangsfdx:
### What does this PR do?
Update the 246 jorje jar, with the support of async/sync init feature

@W-14194698@

### Functionality Before
Does not support sync init of Apex LS

### Functionality After
Support sync init of Apex LS by default


gbockus-sf (approved):
Passes my QE: 
* Verify LSP starts in async mode ✅
* Verify LSP starts in sync mode ✅
* Install vsix from GHA build to verify behaviour is correct when packaged ✅
* Verify jar file hasn’t changed in size significantly ✅
    * +487 Bytes

diyer (approved):

PR #5217, opened 2023-11-07
fix: call language client stop during deactivation

peternhale:
This PR fixes the issue of language servers being orphaned. 

This change adds a call to the language client stop function during the apex extension deactivate call.

@W-14446753@


packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -284,6 +284,7 @@ async function registerTestView(): Promise<vscode.Disposable> {
 }
 
 export async function deactivate() {
+  await languageClient?.stop();
```
gbockus-sf, line 287: sweet.  We already have a unit test suite for this class.  Mind adding a unit test for the deactivate function?  I'll QE in the mean time.

gbockus-sf (approved):
Thanks for the quick fix!
QE passed. 

Test Cases

* Verify when run via a vsix the Apex LSP exits when VSCode is shut down via the X ✅
* Verify when run via a vsix the Apex LSP exits when VSCode is shut down via quit Visual Studio Code ✅
* Verify when run via a vsix the Apex LSP exits when VSCode is shut down when running in dev mode ✅

packages/salesforcedx-vscode-apex/src/index.ts
peternhale: Will do.

packages/salesforcedx-vscode-apex/src/index.ts
```diff
@@ -284,20 +284,25 @@ async function registerTestView(): Promise<vscode.Disposable> {
 }
 
 export async function deactivate() {
-  await languageClient?.stop();
+  await getClient()?.stop();
```
gbockus-sf, line 287: Why not just `languageClientUtils.getClientInstance().stop()`?

packages/salesforcedx-vscode-apex/test/jest/index.test.ts
```diff
@@ -62,3 +69,23 @@ describe('indexDoneHandler', () => {
     expect(apexLSPStatusBarItemMock).toHaveBeenCalledTimes(1);
   });
 });
+
+class MockApexLanguageClient {
+  constructor() { }
+  start() { }
+  stop() { }
+}
```
gbockus-sf, line 77: I'm not sure why this is necessary? 

Would 
```
stopSpy = jest.fn();
jest.spyOn(languageClientUtils, 'getClientInstance').mockReturnValue({stop: stopSpy}); 
```
achieve the same thing without the extra class here and exported function in index?

gbockus-sf (approved):
one q about the test. Good to go.

gbockus-sf (approved):
🚢

PR #5226, opened 2023-11-09
Mz/telemetry service refactor

mingxuanzhangsfdx:
### What does this PR do?
TelemetryService used to be a singleton in core. This caused mess in App Insights of telemetry when it was exported to extensions that are not part of core (e.g. Einstein for Devlopers). This PR refactors TelemetryService by exposing TelemetryProvider to provide TelemetryService instances according to the extension context instead of using a singleton.

@W-14250302@

### Functionality Before
image
A non-related command is associated with Einstein for Developers extension.

### Functionality After
The issue gets fixed. There is no abnormal telemetry event being observed.

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
+++ b/packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
@@ -60,7 +60,7 @@ export class TelemetryProvider {
   private static instances = new Map<string, TelemetryService>();
   public static getInstance(extensionName?: string): TelemetryService {
     // If there is no parameter, it is in the context of core extension
-    if (typeof extensionName == undefined) {
+    if (!extensionName) {
       if (!TelemetryProvider.instances.has(SFDX_CORE_EXTENSION_NAME)) {
         TelemetryProvider.instances.set(
           SFDX_CORE_EXTENSION_NAME,
@@ -68,15 +68,12 @@ export class TelemetryProvider {
         );
       }
       return TelemetryProvider.instances.get(
-        SFDX_CORE_CONFIGURATION_NAME
+        SFDX_CORE_EXTENSION_NAME
       ) as TelemetryService;
     } else {
       // If there is parameter, it is called by a external extension
       if (!TelemetryProvider.instances.has(extensionName as string)) {
-        TelemetryProvider.instances.set(
-          SFDX_CORE_EXTENSION_NAME,
-          new TelemetryService()
-        );
+        TelemetryProvider.instances.set(extensionName, new TelemetryService());
       }
       return TelemetryProvider.instances.get(
         extensionName as string
--- a/packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
+++ b/packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
@@ -7,7 +7,11 @@
 
 import * as util from 'util';
 import { env, ExtensionContext, ExtensionMode, workspace } from 'vscode';
-import { DEFAULT_AIKEY, SFDX_CORE_CONFIGURATION_NAME } from '../constants';
+import {
+  DEFAULT_AIKEY,
+  SFDX_CORE_CONFIGURATION_NAME,
+  SFDX_CORE_EXTENSION_NAME
+} from '../constants';
 import { disableCLITelemetry, isCLITelemetryAllowed } from './cliConfiguration';
 import { TelemetryReporter } from './telemetryReporter';
 
@@ -52,8 +56,36 @@ export class TelemetryBuilder {
   }
 }
 
+export class TelemetryProvider {
+  private static instances = new Map<string, TelemetryService>();
+  public static getInstance(extensionName?: string): TelemetryService {
+    // If there is no parameter, it is in the context of core extension
+    if (typeof extensionName == undefined) {
+      if (!TelemetryProvider.instances.has(SFDX_CORE_EXTENSION_NAME)) {
+        TelemetryProvider.instances.set(
+          SFDX_CORE_EXTENSION_NAME,
+          new TelemetryService()
+        );
+      }
+      return TelemetryProvider.instances.get(
+        SFDX_CORE_CONFIGURATION_NAME
+      ) as TelemetryService;
+    } else {
+      // If there is parameter, it is called by a external extension
+      if (!TelemetryProvider.instances.has(extensionName as string)) {
+        TelemetryProvider.instances.set(
+          SFDX_CORE_EXTENSION_NAME,
+          new TelemetryService()
+        );
+      }
+      return TelemetryProvider.instances.get(
+        extensionName as string
+      ) as TelemetryService;
+    }
+  }
+}
+
 export class TelemetryService {
-  private static instance: TelemetryService;
   private extensionContext: ExtensionContext | undefined;
   private reporter: TelemetryReporter | undefined;
   private aiKey = DEFAULT_AIKEY;
@@ -63,13 +95,6 @@ export class TelemetryService {
    */
   private cliAllowsTelemetryPromise?: Promise<boolean> = undefined;
   public extensionName: string = 'unknown';
-
-  public static getInstance() {
-    if (!TelemetryService.instance) {
-      TelemetryService.instance = new TelemetryService();
-    }
-    return TelemetryService.instance;
-  }
 
   /**
    * Initialize Telemetry Service during extension activation.
--- a/packages/salesforcedx-vscode-apex/src/telemetry/index.ts
+++ b/packages/salesforcedx-vscode-apex/src/telemetry/index.ts
@@ -4,5 +4,5 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode';
-export const telemetryService = TelemetryService.getInstance();
+import { TelemetryProvider } from '@salesforce/salesforcedx-utils-vscode';
+export const telemetryService = TelemetryProvider.getInstance();
--- a/packages/salesforcedx-vscode-core/src/index.ts
+++ b/packages/salesforcedx-vscode-core/src/index.ts
@@ -9,7 +9,7 @@ import {
   ChannelService,
   getRootWorkspacePath,
   SFDX_CORE_CONFIGURATION_NAME,
-  TelemetryService
+  TelemetryProvider
 } from '@salesforce/salesforcedx-utils-vscode';
 import * as vscode from 'vscode';
 import { channelService } from './channels';
@@ -164,11 +164,10 @@ function registerCommands(
     'sfdx.force.source.deploy.in.manifest',
     forceSourceDeployManifest
   );
-  const forceSourceDeployMultipleSourcePathsCmd =
-    vscode.commands.registerCommand(
-      'sfdx.force.source.deploy.multiple.source.paths',
-      forceSourceDeploySourcePaths
-    );
+  const forceSourceDeployMultipleSourcePathsCmd = vscode.commands.registerCommand(
+    'sfdx.force.source.deploy.multiple.source.paths',
+    forceSourceDeploySourcePaths
+  );
   const forceSourceDeploySourcePathCmd = vscode.commands.registerCommand(
     'sfdx.force.source.deploy.source.path',
     forceSourceDeploySourcePaths
@@ -432,22 +431,20 @@ function registerInternalDevCommands(
     forceInternalLightningAppCreate
   );
 
-  const forceInternalLightningComponentCreateCmd =
-    vscode.commands.registerCommand(
-      'sfdx.internal.lightning.component.create',
-      forceInternalLightningComponentCreate
-    );
+  const forceInternalLightningComponentCreateCmd = vscode.commands.registerCommand(
+    'sfdx.internal.lightning.component.create',
+    forceInternalLightningComponentCreate
+  );
 
   const forceInternalLightningEventCreateCmd = vscode.commands.registerCommand(
     'sfdx.internal.lightning.event.create',
     forceInternalLightningEventCreate
   );
 
-  const forceInternalLightningInterfaceCreateCmd =
-    vscode.commands.registerCommand(
-      'sfdx.internal.lightning.interface.create',
-      forceInternalLightningInterfaceCreate
-    );
+  const forceInternalLightningInterfaceCreateCmd = vscode.commands.registerCommand(
+    'sfdx.internal.lightning.interface.create',
+    forceInternalLightningInterfaceCreate
+  );
 
   const forceInternalLightningLwcCreateCmd = vscode.commands.registerCommand(
     'sfdx.internal.lightning.lwc.create',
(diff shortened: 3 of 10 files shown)


packages/salesforcedx-vscode-soql/src/telemetry/index.ts
```diff
@@ -4,11 +4,11 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode';
+import { TelemetryProvider } from '@salesforce/salesforcedx-utils-vscode';
 import { JsonMap } from '@salesforce/ts-types';
 import * as vscode from 'vscode';
 
-export const telemetryService = TelemetryService.getInstance();
+export const telemetryService = TelemetryProvider.getInstance();
```
diyer, line 11: dont all these other non core extensions need to pass the right extension name to TelemetryProvider.getInstance() now? IOW dont these other extensions also have the same issue as the egpt extension?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -619,7 +616,7 @@ export async function activate(extensionContext: vscode.ExtensionContext) {
     telemetryService,
     services: {
       ChannelService,
-      TelemetryService,
+      TelemetryProvider,
```
gbockus-sf, line 619: I think this is a fine approach.  We'll need to discuss with the E4D team and see if they want the clean break here in the API, or if we have to be backwards compatible.

diyer:
do we need a follow up PR in EGPT repo?

packages/salesforcedx-vscode-soql/src/telemetry/index.ts
mingxuanzhangsfdx: @gbockus-sf do you have idea about this? I think that is about the definition of "core". Does "core" refer to all extensions in the mono repo or just "salesforce-vscode-core"? My guts is that the issue is only caused by external extensions so far. So I keep the same instance of the internal extensions.

mingxuanzhangsfdx:
> do we need a follow up PR in EGPT repo?

@diyer This is a [preliminary PR](https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/pull/181) to give some insights. More considerations regarding compatibility between these two is required.

packages/salesforcedx-vscode-soql/src/telemetry/index.ts
gbockus-sf: They don't have the same issue b/c they aren't sharing an _instance_ of the telemetry service.  This is unique to the use of the service via the API.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,8 +56,33 @@ export class TelemetryBuilder {
   }
 }
 
+export class TelemetryProvider {
+  private static instances = new Map<string, TelemetryService>();
+  public static getInstance(extensionName?: string): TelemetryService {
+    // If there is no parameter, it is in the context of core extension
```
diyer, line 62: defaulting it to core extension when extension name is not provided seems odd and leaves room for bugs when external extensions can call this without api name (if I am reading this code correctly)

mingxuanzhangsfdx: If I make the parameter mandatory, there will be a bunch of changes in core. And I only need to specify the [interface](https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/blob/4ea742bd3f9762386609bd168338e6cbc57dfe8d/src/types/TelemetryService.ts#L42) with the parameter in external extensions. What else do you think is necessary?

packages/salesforcedx-vscode-apex/src/telemetry/index.ts
```diff
@@ -4,5 +4,5 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode';
-export const telemetryService = TelemetryService.getInstance();
+import { TelemetryProvider } from '@salesforce/salesforcedx-utils-vscode';
+export const telemetryService = TelemetryProvider.getInstance();
```
klewis-sfdc, line 8: Instead of changing to export the Provider, could we put an optional param on the call to TelemetryService.getInstance(); ?  Like  TelemetryService.getInstance(context?: ExtensionContext);

Then internally the telemetry service getInstance method can use the Provider to return the correct instance. If we can make this work, then it seems like it could be backwards compatible and we can just change the way we make the call to send the context once the changes live in the core extension.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
diyer: same question as I asked in during demo. is there an validation needed on the extension name? is is possible to create multiple instances for the same extension?

mingxuanzhangsfdx: @diyer I do not see it is possible to create multiple instances for the same extension. getInstance is only called once in a extension. Could you give me an example?

packages/salesforcedx-vscode-apex/src/telemetry/index.ts
mingxuanzhangsfdx: Thanks @klewis-sfdc , brilliant suggestion! I just made changes based on your idea.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
diyer: I agree, was thinking more from the perspective of defensive programing.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,25 +57,47 @@ export class TelemetryBuilder {
...
+      ) as TelemetryService;
+    } else {
+      // If there is parameter, it is called by a external extension
+      if (!TelemetryProvider.instances.has(extensionName as string)) {
+        TelemetryProvider.instances.set(extensionName, new TelemetryService());
+      }
+      return TelemetryProvider.instances.get(
+        extensionName as string
+      ) as TelemetryService;
+    }
```
peternhale, line 82:
``` suggestion
    // default if not present
    const name = extensionName || SFDX_CORE_EXTENSION_NAME;
    if (!TelemetryProvider.instances.has(name)) {
      TelemetryProvider.instances.set(
        name,
        new TelemetryService()
      );
    }
    return TelemetryProvider.instances.get(name)!;
  }
```

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,25 +57,33 @@ export class TelemetryBuilder {
...
 export class TelemetryService {
-  private static instance: TelemetryService;
   private extensionContext: ExtensionContext | undefined;
   private reporter: TelemetryReporter | undefined;
   private aiKey = DEFAULT_AIKEY;
   private version: string = '';
+  // Temporary usage for earlier version of e4d
+  public static getInstance() {
+    return TelemetryProvider.getInstance(SFDX_E4D_EXTENSION_NAME);
+  }
```
klewis-sfdc, line 80: Instead of calling getInstance() with the E4D extension name here, I was envisioning the getInstance method would have an optional param to pass in the extension name.  This way E4D can get a unique Telem Serv instance, but so can other extensions:

klewis-sfdc: TelemService.`initializeService` takes the ExtensionContext as a parameter.  In an ideal world, the `getInstance` method is called by passing in the ExtensionContext, and then the service is initialized internally before handing the service instance back to the caller.  

I don't love the current setup because you have to call `getInstance` and then call `initializeService` before you can use the Telem Serv instance. This might be a larger refactor to look into later, but I'd love for it to look like this in core ext:





So that we can use it like this in E4D:

.prettierrc
```diff
@@ -1,5 +1,7 @@
 {
   "printWidth": 80,
   "tabWidth": 2,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none",
+  "arrowParens": "avoid"
```
peternhale, line 6: Remove these please. I know the lint results are very noisy right now, but we will be working through soon.

peternhale (changes requested):

peternhale (changes requested):
prettier fix

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
mingxuanzhangsfdx: Thanks @klewis-sfdc. Your idea about refactoring is good. My concern is that current `getInstance` in TelemetryService is only a temporary usage for old version of e4d, which we cannot modify in there. Even if we reconstruct it here, the old version still call initialize service.

mingxuanzhangsfdx: Do you want me to opt in this change in this PR? If so, now that we have telemetryProvider, I prefer to call telemetryProvider.getInstance instead of telemetryService. What is your idea?

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,25 +57,33 @@ export class TelemetryBuilder {
   }
 }
 
+export class TelemetryProvider {
```
klewis-sfdc, line 60: This should be renamed TelemetryServiceProvider, since it is now offering instances of TelemetryService.

packages/salesforcedx-utils-vscode/src/constants.ts
```diff
@@ -5,3 +5,4 @@ export const TARGET_DEV_HUB_KEY = 'target-dev-hub';
 export const TARGET_ORG_KEY = 'target-org';
 export const SF_CONFIG_DISABLE_TELEMETRY = 'disable-telemetry';
 export const DEFAULT_AIKEY = 'ec3632a4-df47-47a4-98dc-8134cacbaf7e';
+export const SFDX_E4D_EXTENSION_NAME = 'salesforcedx-einstein-gpt';
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: yeah it should not be aware of names of external extensions. The old method in e4d is Telemetry.getInstance() so it is defined here to be compatible. We need to decide when to opt out this. @klewis-sfdc

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,25 +56,38 @@ export class TelemetryBuilder {
...
-  public static getInstance() {
-    if (!TelemetryService.instance) {
-      TelemetryService.instance = new TelemetryService();
-    }
-    return TelemetryService.instance;
-  }
-
   /**
    * Initialize Telemetry Service during extension activation.
    * @param extensionContext extension context
```
klewis-sfdc, line 93: Also, on line 107 `this.extensionName` is set to `name`

Since this.extensionName is set in the getInstance method, and used as the key in the map in the Provider, then the extensionName property should probably not be renamed.  What do you think about a future work item, after both of these prs are merged and released, to throw an error if the Telem Service is initialized with a different context?  Something like:

klewis-sfdc (commented):
Hi @mingxuanzhangsfdx !  This is looking great!  I have one request - to update the name passed to the TelemetryReporter - and one consideration for a future WI.  Let me know what you think, or if you have any questions.

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
klewis-sfdc: Or maybe just this.extensionName:

```
      this.reporter = new TelemetryReporter(
        this.extensionName,
        this.version,
        this.aiKey,
        true
      );
```

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    it('should return "salesforcedx-vscode" when extensionName starts with "salesforcedx-vscode"', () => {
+      telemetryService.extensionName = 'salesforcedx-vscode-core';
+      const result = telemetryService.getTelemetryReporterName();
+      expect(result).toBe('salesforcedx-vscode');
+    });
+
+    it('should return the actual extensionName when it does not start with "salesforcedx-vscode"', () => {
+      telemetryService.extensionName = 'salesforcedx-einstein-gpt';
+      const result = telemetryService.getTelemetryReporterName();
+      expect(result).toBe(telemetryService.extensionName);
```
klewis-sfdc, line 91: 👍🏽

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -52,25 +57,38 @@ export class TelemetryBuilder {
...
+// export only for unit test
+export class TelemetryServiceProvider {
+  public static instances = new Map<string, TelemetryService>(); // public only for unit test
+  public static getInstance(extensionName?: string): TelemetryService {
+    // default if not present
+    const name = extensionName || SFDX_CORE_EXTENSION_NAME;
+    if (!TelemetryServiceProvider.instances.has(name)) {
+      TelemetryServiceProvider.instances.set(name, new TelemetryService());
+    }
+    return TelemetryServiceProvider.instances.get(name)!;
```
klewis-sfdc, line 69:
```suggestion
    return TelemetryServiceProvider.instances.get(name);
```

if an instance with the name is not found, it is created and put into the map, and then retrieved and returned.  There should not be any time when this method needs to return undefined, which is what I think the "!" at the end allows.  Can the "!" be removed here?

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
```diff
@@ -121,6 +138,18 @@ export class TelemetryService {
...
+   * Helper to get the name for telemetryReporter
+   * if the extension from extension pack, use salesforcedx-vscode
+   * otherwise use the extension name
+   * exported only for unit test
+   */
+  public getTelemetryReporterName(): string {
+    return this.extensionName.startsWith(SFDX_EXTENSION_PACK_NAME)
+      ? SFDX_EXTENSION_PACK_NAME
+      : this.extensionName;
+  }
```
klewis-sfdc, line 151: 👍🏽

packages/salesforcedx-utils-vscode/src/telemetry/telemetry.ts
mingxuanzhangsfdx: thanks for pointing this out. There is strict null checking for typescript in the project so it has to be added in order to bypass the rule.

klewis-sfdc (approved):
Thank you @mingxuanzhangsfdx !

peternhale (changes requested):
Just a couple of nit issues

peternhale (changes requested):
Just a couple of nit issues

peternhale (approved):

PR #5244, opened 2023-11-20
Mz/changelog autogeneration

mingxuanzhangsfdx:
### What does this PR do?
This PR adds a workflow that generates changelog after the workflow Create Release Branch finishes.

@W-14255738@

### Functionality Before
Changelog is generated manually on a local machine by the person who is on trust.

### Functionality After
Changelog is generated and committed to the release branch on the GA workflow Generate Changelog.

### DIFF
--- a/.github/workflows/generateChangelog.yml
+++ b/.github/workflows/generateChangelog.yml
@@ -1,0 +1,35 @@
+name: Generate Changelog
+
+on:
+  workflow_run:
+    workflows: ['Create Release Branch']
+    types:
+      - completed
+
+jobs:
+  generate_changelog:
+    name: 'Generate Changelog'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: 'develop'
+          ssh-strict: false
+          token: ${{ secrets.IDEE_GH_TOKEN }}
+          fetch-depth: '0'
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+      - uses: ./.github/actions/gitConfig
+        with:
+          email: ${{ secrets.IDEE_GH_EMAIL }}
+      - name: Set NPM at the correct version for Lerna
+        run: npm install -g npm@9.8.1
+      - run: npm ci
+      - run: npm install -g shelljs && npm install -g lerna@5.5.4
+      - name: Generate CHANGELOG
+        run: npm run changelog
+      - id: result
+        run: echo "result=${{ job.status }}" >> $GITHUB_OUTPUT
--- a/.github/workflows/generateChangelog.yml
+++ b/.github/workflows/generateChangelog.yml
@@ -14,7 +14,7 @@ jobs:
           ref: 'develop'
           ssh-strict: false
           token: ${{ secrets.IDEE_GH_TOKEN }}
-          fetch-depth: '0'
+          fetch-depth: '0' # Fetch all tags to prepare for tag retrieval
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
@@ -22,10 +22,7 @@ jobs:
       - uses: ./.github/actions/gitConfig
         with:
           email: ${{ secrets.IDEE_GH_EMAIL }}
-      - name: Set NPM at the correct version for Lerna
-        run: npm install -g npm@9.8.1
       - run: npm ci
-      - run: npm install -g shelljs && npm install -g lerna@5.5.4
       - name: Generate CHANGELOG
         run: npm run changelog
       - id: result
--- a/.github/workflows/generateChangelog.yml
+++ b/.github/workflows/generateChangelog.yml
@@ -14,7 +14,7 @@ jobs:
           ref: 'develop'
           ssh-strict: false
           token: ${{ secrets.IDEE_GH_TOKEN }}
-          fetch-depth: '0' # Fetch all tags to prepare for tag retrieval
+          fetch-depth: '0' # Fetching tags will fail if it is not 0
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/generateChangelog.yml
+++ b/.github/workflows/generateChangelog.yml
@@ -1,10 +1,7 @@
 name: Generate Changelog
 
 on:
-  workflow_run:
-    workflows: ['Create Release Branch']
-    types:
-      - completed
+  workflow_call:
 
 jobs:
   generate_changelog:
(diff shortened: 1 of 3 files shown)


.github/workflows/generateChangelog.yml
```diff
@@ -0,0 +1,35 @@
+name: Generate Changelog
+
+on:
+  workflow_run:
+    workflows: ['Create Release Branch']
+    types:
+      - completed
```
diyer, line 7: Does this also indicate the workflow was successful ?

diyer:
Nice! with this change, publishing.md also needs to be updated with the right set of instructions

.github/workflows/generateChangelog.yml
daphne-sfdc: @diyer **type:completed** means the workflow finished running, regardless of whether it is successful or not.  It is also used in e2e.yml to start an E2E tests job whenever VSIXs are available after _Nightly Build Develop_ or _Test, Build, and Release_.

Documentation here: https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run

NOTE: There are multiple ways for one workflow to call another - this is just one method, and a perfectly valid one that has proven to work for running E2E tests with VSIXs.  I can think of another method that would seem equally valid, using **jobs** at the bottom of the caller workflow.  Based on what I see in the Apex E2E Tests workflow, I believe that one can filter on success vs. failure of the _Create Release Branch_ workflow - although I did not test that, so this is just my speculation based on what I have been doing with the E2E test workflows in the past.  I'm sure @mingxuanzhangsfdx  has his reasoning for why he chose one method of calling workflows over another.

mingxuanzhangsfdx: Thank you @diyer @daphne-sfdc . I made changes to make sure the job create_branch succeeds and trigger the changelog generation job.

.github/workflows/generateChangelog.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: 'Generate Changelog'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: 'develop'
+          ssh-strict: false
+          token: ${{ secrets.IDEE_GH_TOKEN }}
+          fetch-depth: '0'
```
gbockus-sf, line 17: are the ssh and fetch-depth values required?  I didn't think we used those in other places so just curious why it's different here.

.github/workflows/generateChangelog.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+      - uses: ./.github/actions/gitConfig
+        with:
+          email: ${{ secrets.IDEE_GH_EMAIL }}
+      - name: Set NPM at the correct version for Lerna
+        run: npm install -g npm@9.8.1
```
gbockus-sf, line 26: was this required or did you carry this over from somewhere?

.github/workflows/generateChangelog.yml
```diff
@@ -0,0 +1,32 @@
...
+      - run: npm ci
+      - run: npm install -g shelljs && npm install -g lerna@5.5.4
```
gbockus-sf, line 28: Do we use lerna in the changelog script?

gbockus-sf (approved):
Cool.  Logic looks good.  had a few q's about specifics of the yml file but looks good to me

.github/workflows/generateChangelog.yml
mingxuanzhangsfdx: I followed what is defined in createReleaseBranch.yml for ssh-strict.

fetch-depth is necessary. During my test, I found there are always errors [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/7a03ff034d4d3983744d7c430230e7eb0dcacc78/scripts/create-release-notes.ts#L40).
Adding this line could fetch all history for all branches and tags. And the issue got fixed

.github/workflows/generateChangelog.yml
mingxuanzhangsfdx: I copied them from createReleaseBranch.yml. both npm and lerna are used in s`alesforcedx-vscode/scripts/create-release-branch.js`

.github/workflows/generateChangelog.yml
mingxuanzhangsfdx: `salesforcedx-vscode/scripts/create-release-branch.js`

.github/workflows/generateChangelog.yml
gbockus-sf: b/c we're running it as an npm script node has access to the project dependancies (including lerna and shell js). I think you can remove this and line 28.  Give it a try on your fork and if it fails weirdly I'll give up and just 👍

.github/workflows/generateChangelog.yml
gbockus-sf: cool.  Def would be good to add a comment (can we do that in yml files?) about why it's necessary.  It makes sense that it's running out of room to get the required information (sorta...why don't we hit this limitation locally)

mingxuanzhangsfdx: Done!

.github/workflows/generateChangelog.yml
```diff
@@ -14,18 +14,15 @@ jobs:
           ref: 'develop'
           ssh-strict: false
           token: ${{ secrets.IDEE_GH_TOKEN }}
-          fetch-depth: '0'
+          fetch-depth: '0' # Fetch all tags to prepare for tag retrieval
```
gbockus-sf, line 17: nit This should probably say the call to fetch all tags fails if we don't limit the depth 🤷

.github/workflows/generateChangelog.yml
```diff
@@ -14,18 +14,15 @@ jobs:
...
         uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
       - uses: ./.github/actions/gitConfig
         with:
           email: ${{ secrets.IDEE_GH_EMAIL }}
-      - name: Set NPM at the correct version for Lerna
-        run: npm install -g npm@9.8.1
       - run: npm ci
-      - run: npm install -g shelljs && npm install -g lerna@5.5.4
```
gbockus-sf: 👍

gbockus-sf (approved):
Good stuff 🚢

gbockus-sf (approved):
:shipit:

peternhale:
QA Notes

Using a fork of the repository the changelog for a the given release was properly generated and pushed to the release branch.

Nice work @mingxuanzhangsfdx

PR #5245, opened 2023-11-21
chore: build failure on vscode:package

peternhale:
Remove the `-P ./tsconfig.json` for target vscode:extension

@W-14536918@


gbockus-sf (approved):
Nice find on the fix.  Def an if it builds it builds situation 🚢

gbockus-sf:
updated PR comment to chore so it won't show up on the CL

PR #5248, opened 2023-11-24
chore: new setting for NODE_EXTRA_CA_CERTS environment value

CristiCanizales:
### What does this PR do?
- Adds new core setting for the NODE_EXTRA_CA_CERTS environment value

@W-11384659@

### Functionality Before
- NODE_EXTRA_CA_CERTS has no effect in deploy/retrieve operations

### Functionality After
- new setting for NODE_EXTRA_CA_CERTS so it has effect in deploy/retrieve operations

### DIFF
--- a/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
@@ -99,7 +99,10 @@ export class SfdxCoreSettings {
   }
 
   public getNodeExtraCaCerts(): string {
-    return this.getConfigValue(ENV_NODE_EXTRA_CA_CERTS, '');
+    return this.getConfigValue(
+      ENV_NODE_EXTRA_CA_CERTS,
+      process.env.NODE_EXTRA_CA_CERTS ?? ''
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -94,6 +98,10 @@ export class SfdxCoreSettings {
     return this.getConfigValue(SETTING_CLEAR_OUTPUT_TAB, false);
   }
 
+  public getNodeExtraCaCerts(): string {
+    return this.getConfigValue(ENV_NODE_EXTRA_CA_CERTS, '');
```
peternhale, line 102: @CristiCanizales I was thinking that we could also consider env var for NODE_EXTRA_CA_CERTS as a default if the customer does use the setting.
```suggestion
    return this.getConfigValue(ENV_NODE_EXTRA_CA_CERTS, process.env.NODE_EXTRA_CA_CERTS ?? '');
```

peternhale (changes requested):

peternhale (approved):

PR #5260, opened 2023-12-05
fix: bump templates to v59

peternhale:
@W-14515374@
Bumps @salesforce/templates to v59.x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6085,107 +6085,6 @@
         "node": ">=16.13.0"
       }
     },
-    "node_modules/@salesforce/apex-node/node_modules/@salesforce/core": {
-      "version": "5.3.18",
-      "resolved": "https://registry.npmjs.org/@salesforce/core/-/core-5.3.18.tgz",
-      "integrity": "sha512-/Ag7elFngTT13PRblSPJPB2Q+xk3jR2SX8bYa83fcQljVF7ApGB5qtFpauXmUv8lgRnN+F01HNqM16iszAMP9w==",
-      "dependencies": {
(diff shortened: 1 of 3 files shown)


.github/workflows/integrationTestsWindows.yml
```diff
@@ -21,7 +21,6 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - run: npm config set scripts-prepend-node-path true
```
CristiCanizales, line 24: Why are we getting rid of this?

packages/salesforcedx-vscode-core/test/vscode-integration/commands/forceProjectCreate.test.ts
```diff
@@ -326,7 +326,6 @@ describe('Force Project Create', () => {
         'triggers'
       ];
       const filestocopy = [
-        '.eslintignore',
```
CristiCanizales, line 329: same question here :)

peternhale: @CristiCanizales recall the change you made to templates that removed this from project sample files? Removing it here because our create project tests are verifying the file should be present. Removing here to match the work done in templates.

.github/workflows/integrationTestsWindows.yml
peternhale: When running windows integration tests, the npm command is failing stating the option `scripts-prepend-node-path` is not supported. I assume that the later version of npm have removed this setting.

PR #5271, opened 2023-12-12
Cristi/w 14366453/e2e in old vscode

CristiCanizales:
### What does this PR do?
- Adds a new nightly GHA that will run just the initial E2E test suite (verifying extension start) against the desired version of vscode

@W-14366453@

### Functionality after
- E2E anInitialSuite.e2e.ts Test run after Nightly build develop runs.
- E2E anInitialSuite.e2e.ts Test run against oldest vscode version we support (by default same as engines property in the source code) or any other we provide
- E2E anInitialSuite.e2e.ts Test is the default test we run but any other suite e.g deployAndRetrieve.e2e.ts can be run as well
image

### DIFF
--- a/.github/workflows/vscodeSupportE2E.yml
+++ b/.github/workflows/vscodeSupportE2E.yml
@@ -24,7 +24,6 @@ on:
       vscodeVersion:
         description: 'VS Code Version'
         required: false
-        default: '1.61.2'
         type: string
 
 jobs:
@@ -37,7 +36,7 @@ jobs:
         nodeVersion:
           - 18.11.0
         vscodeVersion:
-          - ${{ inputs.vscodeVersion }}
+          - ${{ inputs.vscodeVersion || '1.61.2'}}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -48,7 +47,7 @@ jobs:
         run: |
           mkdir ./extensions
           pwd
-          gh run download ${{ inputs.runId }} -D ./extensions
+          gh run download ${{ inputs.runId || github.event.workflow_run.id }} -D ./extensions
           mv ./extensions/*/* ./extensions/
         working-directory: salesforcedx-vscode
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/vscodeSupportE2E.yml
```diff
@@ -0,0 +1,111 @@
...
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        nodeVersion:
+          - 18.11.0
```
daphne-sfdc, line 37: Why Node v18.11.0 instead of v18.18.2?

.github/workflows/vscodeSupportE2E.yml
```diff
@@ -0,0 +1,111 @@
...
+        vscodeVersion:
+          - ${{ inputs.vscodeVersion || '1.61.2'}}
```
daphne-sfdc, line 39: According to the VSCode documentation, I see that 1.61.2 is the VSCode version from September 2021 and a new minor version is released every month.  Since we are going with the hard-coding the VSCode version for now, this does mean that we need to update this workflow once a month, right?

packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
```diff
@@ -49,7 +49,7 @@ describe('Replay debugger adapter - integration', () => {
       'apex-replay'
     );
     await dc.start();
-    dc.defaultTimeout = 10000;
+    dc.defaultTimeout = 15000;
```
daphne-sfdc, line 52: Why does the defaultTimeout need to be increased in this unigration test?

.github/workflows/vscodeSupportE2E.yml
CristiCanizales: I had just used the same version we're using in the `runE2ETest.yml` workflow, and hadn't realized we had updated the version in the `.nvmrc` file. I just updated both workflows with v18.18.2. Thanks!

packages/salesforcedx-apex-replay-debugger/test/integration/adapter.test.ts
CristiCanizales: I was just trying some things and didn't realized this got pushed. I reverted this change

daphne-sfdc (approved):
Approved! 🎉

New workflow looks good ✅

PR #5273, opened 2023-12-12
Cristi/config commands

CristiCanizales:
### What does this PR do?
- Updates config commands and flags to SF style in order to get rid of cli warnings

@W-14668110@

### Functionality Before
- config commands in sfdx style

### Functionality After
- config comands in sf style


daphne-sfdc (approved):
Approved! 🎉

- SFDX: List All Config Variables ✅
- SFDX: Set a Default Org ✅

PR #5274, opened 2023-12-13
fix: refactor to arrow functions

peternhale:
Apply arrow function refactor to apex extension
@W-14564471@

### DIFF
--- a/packages/salesforcedx-vscode-apex/fix-these-rules.json
+++ b/packages/salesforcedx-vscode-apex/fix-these-rules.json
@@ -1,5 +1,6 @@
 {
   "rules": {
+    "prefer-arrow/prefer-arrow-functions": ["error", {}],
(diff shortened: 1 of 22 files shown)


packages/salesforcedx-vscode-apex/fix-these-rules.json
```diff
@@ -1,5 +1,6 @@
 {
   "rules": {
+    "prefer-arrow/prefer-arrow-functions": ["error", {}],
```
gbockus-sf, line 3: Was wondering if this line should be deleted from fix-there-rules then checked and the base has 
`"prefer-arrow/prefer-arrow-functions": ["warn", {}],`.  If we're overriding it to warn in the fix files can it be set to error in the base and removed from here?

gbockus-sf (approved):
Thanks Pete!  LGTM.  Def seems to fall into the if it builds it builds category.

PR #5277, opened 2023-12-14
chore: vscode-utils arrow functions

peternhale:
@W-14564471@
Refactor vscode-utils to use arrow functions

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/date/format.ts
+++ b/packages/salesforcedx-utils-vscode/src/date/format.ts
@@ -5,7 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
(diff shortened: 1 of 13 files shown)


packages/salesforcedx-utils-vscode/src/date/format.ts
```diff
@@ -5,19 +5,21 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-export function getYYYYMMddHHmmssDateFormat(localUTCDate: Date): string {
+import * as util from 'util';
```
gbockus-sf, line 8: nit:  should we do `import { format } from 'util';

gbockus-sf: it really doesn't matter today, but if we ever go esm it'll matter.

gbockus-sf (approved):
🆒 1 nit to consider 🚢

PR #5300, opened 2023-12-28
chore: refactor vf LS to use arrow functions

peternhale:
@W-14564471@
refactor package VF language server to use arrow functions where applicable


CristiCanizales (approved):
LGTM! ✅ 

🚢

PR #5301, opened 2023-12-28
chore: refactor vf md LS to use arrow functions

peternhale:
@W-14564471@
refactor VF MD language server to use arrow functions as applicable


CristiCanizales (approved):
lgtm!

PR #5304, opened 2023-12-29
chore: refactor vscode apex replay to use arrow

peternhale:
@W-14564471@
Refactor vscode apex replay debugger to use arrow functions


CristiCanizales (approved):
LGTM✅

Tested with vsixes apex replay debugger works as expected

:shipit:

PR #5318, opened 2024-01-05
fix: default SF_LOG_LEVEL to fatal

CristiCanizales:
### What does this PR do?
- Sets SF_LOG_LEVEL to default in 'fatal'

[skip-validate-pr]


daphne-sfdc (approved):
SF_LOG_LEVEL is now pre-populated to `fatal` in both User and Workspace settings ✅

Issue #1178, opened 2019-03-21
Code coverage error


ntotten:
After last update Code Coverage shows correctly on Apex Classes, but an annoying notification appears every time a class with no code coverage is opened

This happens with Test Classes, maybe it could be great to ignore this files that does not require any line to test

![testCodeCoverage](https://user-images.githubusercontent.com/10410852/54773321-a2210380-4c09-11e9-8efa-63107c8754c7.gif)

Also, the org selector is now "Lightning" Fast

_Originally posted by @mnunezdm in https://github.com/forcedotcom/salesforcedx-vscode/issues/973#issuecomment-475335340_

ntotten:
I verified that I can reproduce this error. We'll get it fixed.

lcampos:
Yes, this is by design for the first version of the feature. We don't do any additional checks on the content of the file other than it being apex. This is good feedback and will address it in the follow up versions.

mnunezdm:
No plans in changing this @smaddox-sf ?

smaddox-sf:
hi @mnunezdm - I thought this had been resolved, but see now that's it's still happening.

jonnypetraglia:
Would like to bump this as a request. If anything I'd like this to just be a status bar thing next to the "Show Apex Code Coverage" to show the current status (e.g. "On/Off/Found/None"), but seeing the notification **every time** I change tabs back to the test file I'm writing tests for is.....well it's rather ridiculous. How difficult would it really be to NEVER display the notification for Apex files that are tests? The extension(s) obviously know it's a test file if it knows to show "Run All Tests | Debug All Tests".

Like I get that this is a non-impacting change but just how high of an effort can it be to be stale for a year? Not even asking for caching with non-tests, just Tests that....literally will NEVER have coverage to retrieve. Not showing hundreds of useless notifications every day is not an unreasonable ask.

jonnypetraglia:
Bumping on this in hopes of a response.

randi274:
Thanks for the bump on this @jonnypetraglia - I've also confirmed it's in our backlog, it just hasn't bubbled up yet.

jag-sfdc:
FYI, tracked internally as bug W-5971491.

diyer:
This issue has been fixed and released.

PR #5320, opened 2024-01-05
fix: move no coverage warning to to console

peternhale:
Fixes issue [#1178](https://github.com/forcedotcom/salesforcedx-vscode/issues/1178)

Previously when user has apex code coverage enabled, as one opened any file type, one would get a warning stating no coverage is available on the file, even when it is not an Apex artifact (cls or trigger). 

This change accomplishes the following:
- When a user opens a file, only Apex artifacts that can report coverage will be subject to a warning.
- If the lack of coverage is detected, the warning will be written to the Apex output channel, rather than opening a warning window.

@W-10288513@


fellotin (approved):

packages/salesforcedx-vscode-apex/package.nls.json
```diff
@@ -25,5 +25,6 @@
   "apex_trace_server_description": "Controls the level of detail for trace logs in the Apex Language Server",
   "apex_verbose_level_trace_description": "Output everything, including details about notifications and responses received by the client, and requests sent by the server.",
   "apex_messages_level_trace_description": "Only output high-level messages of notifications and responses received by the client, and requests sent by the server.",
-  "apex_off_level_trace_description": "Don't generate any output. Turn off all tracing."
+  "apex_off_level_trace_description": "Don't generate any output. Turn off all tracing.",
+  "apex_code_disable-warnings-for-missing-coverage": "Disable UI warning when apex test coverage missing."
```
daphne-sfdc, line 29:
```suggestion
  "apex_code_disable-warnings-for-missing-coverage": "Disable UI warning when Apex test coverage is missing."
```

packages/salesforcedx-vscode-apex/src/commands/forceAnonApexExecute.ts
```diff
@@ -249,7 +255,7 @@ export class AnonApexLibraryExecuteExecutor extends LibraryCommandletExecutor<Ap
   }
 }
 
-export const forceAnonApexExecute = async () => {
+export const forceAnonApexExecute =  async () => {
```
daphne-sfdc, line 258:
```suggestion
export const forceAnonApexExecute = async () => {
```
Extra space was added here.

packages/salesforcedx-vscode-apex/src/commands/forceApexLogGet.ts
```diff
@@ -129,7 +129,7 @@ export class ApexLibraryGetLogsExecutor extends LibraryCommandletExecutor<{
     const logPath = logResults[0].logPath;
     if (logPath) {
       const document = await vscode.workspace.openTextDocument(logPath);
-      vscode.window.showTextDocument(document);
+      void vscode.window.showTextDocument(document);
```
daphne-sfdc, line 132: What's the purpose of this `void` declaration?

peternhale: The call to `vscode.window.showTextDocument(document);` returns a promise and the run function is async. We have an eslint rules that currently warns of a situation where the code is not properly handling the called functions promise/then-able return. Since this is opening a text document and the previous code was not capturing the returned promise I assume here that the intent was to open the text document w/o concern of the result.

To resolve the eslint and more importantly acknowledge in the code to match the intent, one can use void to say I am not concerned about the result, so call the function and ignore the presence of the returned promise.

The called function will run and resolve on its own and the calling function will continue without having to handle the promise. General term would be "fire the event and forget the result."

daphne-sfdc: Got it!  Thank you for the explanation - I'm clear now 👍

daphne-sfdc (approved):
Approved! 🎉

Thank you Pete for this fixing this code coverage warning issue, and for the improvements you made while refactoring the code for running Apex tests!  The code coverage warnings and setting are now appearing as expected, and there is no change in the functionality of running Apex tests. 👍 

QE doc here: https://salesforce.quip.com/kwvzA1Yb3NkF

PR #5321, opened 2024-01-05
Wr/generate basic unit test

WillieRuemmele:
### What does this PR do?
adds a new command for generating a "basic" apex unit test, just the test class and annotation

 @W-14709113@

### Functionality Before
would have to generate an apex class with a large header comment, and test method

### Functionality After
just generate a ready-to-go apex unit test

requires: https://github.com/forcedotcom/salesforcedx-templates/pull/575


packages/salesforcedx-vscode-core/package.json
```diff
@@ -29,7 +29,7 @@
     "@salesforce/salesforcedx-utils-vscode": "59.13.0",
     "@salesforce/schemas": "^1.6.1",
     "@salesforce/source-deploy-retrieve": "9.8.4",
-    "@salesforce/templates": "^59.0.4",
+    "@salesforce/templates": "^60.1.0",
```
peternhale, line 32: @WillieRuemmele this library change is going to pull in `@salesforce/core` v6. Are the template changes tied to 248 release of core, cuz I see the major bump?

@daphne-sfdc is working on the core v5-v6 change now, so this PR might have to wait until that work is complete before this PR can progress.

packages/salesforcedx-vscode-core/package.nls.json
```diff
@@ -43,6 +43,12 @@
   "force_source_status_local_text": "SFDX: View Local Changes",
   "force_source_status_remote_text": "SFDX: View Changes in Default Org",
   "force_source_status_text": "SFDX: View All Changes (Local and in Default Org)",
+  "cancel_sfdx_command_text": "SFDX: Cancel Active Command",
+  "apex_generate_class_text": "SFDX: Create Apex Class",
```
peternhale, line 47: @WillieRuemmele this key already exists above, please reconcile these additions to see if they need to be added.

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateBasicTestClass.ts
```diff
@@ -0,0 +1,60 @@
+/*
```
peternhale, line 1: What's the difference between a basic test class and the original?

peternhale (changes requested):

packages/salesforcedx-vscode-core/package.json
WillieRuemmele: they're not related to a 248 core version, but that was just the version of `templates` that was released when I merged. I believe `60` was already published, and then I published `60.1` - but it was automated

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateBasicTestClass.ts
WillieRuemmele: the basic UT just is an apex test class outline, the standard one has a large comment at the top describing how to test, what should be test, etc...

peternhale: @WillieRuemmele is there any function difference between the two for the work you are currently doing. If not, please check with Ananya to see if we really need two variations of an Apex unit test template.

If both are needed then I would suggest that instead of adding a distinct command, that the current command to create a unit test be modified to prompt the user for the template name.

peternhale: Another option would be to reduce the verbosity of the current template, possible adding pointers to web docs/resources that cover all the details presented in the comment block.

WillieRuemmele: I'll look at modifying the current command to create this basic template. I wasn't sure if it would be easier to do that, or create this new command. It should ideally be invisible to customers to call/use and should only be used by E4D for now

packages/salesforcedx-vscode-core/package.json
daphne-sfdc: @WillieRuemmele I published v60.0.1 less than an hour before you published - I realized that the Core v5 -> v6 change in the templates repo did not trigger a publish after the PR was committed, so I ran the manual release workflow.  Since your change in the templates repo is on top of that one, I believe your change would have triggered a v60 publish regardless of whether I did the manual release.

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateBasicTestClass.ts
peternhale: @WillieRuemmele what is the reason for the new template?

WillieRuemmele: for when Test Case Generation generates a new unit test file completely, the original template had a large comment at the top, and an empty test method that wasn't required.

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001gtJt8YAE/view

WillieRuemmele: @peternhale I've put up this draft https://github.com/forcedotcom/salesforcedx-vscode/pull/5331 as an alternative, I believe it'll achieve the same results for E4D, and the same UX as the existing command

PR #5326, opened 2024-01-09
fix: fix entry not found when launching apex replay debugger from apex class 

peternhale:
### What does this PR do?
Corrects a bug when starting the apex replay debugger from an apex class file "Apex: Launch Apex Replay Debugger With Current File"
@W-12508503@

### Functionality Before
When launching Apex replay debugger with an Apex class the command would fail with a file system error stating the class name did not exist.

### Functionality After
The launch of Apex replay debugger is now using the full path to the current Apex class, rather than the name

### DIFF
--- a/config/base-eslintrc.json
+++ b/config/base-eslintrc.json
@@ -37,7 +37,7 @@
         "",
         {
           "pattern": " \\* Copyright \\(c\\) \\d{4}, salesforce\\.com, inc\\.",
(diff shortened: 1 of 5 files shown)


CristiCanizales (approved):
LGTM! Thanks for getting this back to finally work again 🥳 

I tested with the vsixes and the Apex Replay Debugger now works on Apex test files too!

:shipit:

PR #5331, opened 2024-01-11
chore: test using existing command

WillieRuemmele:
What does this PR do?
adds a new command for generating a "basic" apex unit test, just the test class and annotation

@W-14709113@

Functionality Before
would have to generate an apex class with a large header comment, and test method

Functionality After
just generate a ready-to-go apex unit test

requires: https://github.com/forcedotcom/salesforcedx-templates/pull/575
used by: https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/pull/325


peternhale:
@WillieRuemmele if I read this correctly, ED4will be calling the a command execute directly with the basic template name and if run from command palette, the command will use the decorated template.

If this is the case, I like it!

WillieRuemmele:
correct, here's the accompanying [PR](https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/pull/325) in E4D

looks like the command name has changed since we first started with it anyways

peternhale:
@WillieRuemmele Thanks for going the extra mile to avoid a new command. All that is missing are unit tests. Could you have a look [here](packages/salesforcedx-vscode-core/test/jest/commands/templates/apexGenerateUnitTestClass.test.ts).

WillieRuemmele:
@peternhale added tests, hard to get many assertions because everything was mocked. I thought this approach would have a lot of breaking changes, and be harder to integrate, glad it's easier this way

peternhale (approved):

peternhale (approved):

floralan (approved):
Good work! 

✅ QA has passed. 
1. This command should not be exposed in the command palette ✅
2. Confirm the unit test template generated by sfdx.apex.generate.unit.test.class doesn’t change ✅

PR #5334, opened 2024-01-12
chore: setup Git2Gus for User Story when issues are labeled with 'feature' tag

daphne-sfdc:
When Github issues are tagged with the `feature` tag, a new User Story is automatically created for the issue.  I modified `.git2gus/config.json` according to the instructions on the Git2Gus page: https://lwc-gus-bot.herokuapp.com/

[skip-validate-pr]

### DIFF
--- a/.git2gus/config.json
+++ b/.git2gus/config.json
@@ -25,11 +25,11 @@
     "area:xml": "a1aB0000000BPyKIAW"
   },
   "issueTypeLabels": {
-    "type:feature": "",
+    "feature": "USER STORY",
     "type:bug": "BUG P3",
     "type:security": "BUG P2",
     "type:feedback": "",
-    "type:duplicate": "",
+    "status:duplicate": "",
     "type:community-contrib": "USER STORY"
   },
   "defaultBuild": "offcore.tooling.58",


.git2gus/config.json
```diff
@@ -25,11 +25,11 @@
     "area:xml": "a1aB0000000BPyKIAW"
   },
   "issueTypeLabels": {
-    "type:feature": "",
+    "feature": "USER STORY",
     "type:bug": "BUG P3",
```
diyer, line 29: it will be good to have more granular types here that lets us create p0 p1 p2 p3 bugs. 
also can you explain why did you take out the word "type"?

diyer: something like
    "type:bug-p0": "BUG P0",
    "type:bug-p1": "BUG P1",
    "type:bug-p2": "BUG P2",

default can be P3 as it exists today

daphne-sfdc: These are the existing labels that we can tag Github issues with.  We tag features with the `feature` tag - there is no `type:feature` tag in our options.

As for bugs, we currently have the `type:bug` tag, which automatically creates a P3 bug.  I can make 3 new tags for P0 bug, P1 bug, and P2 bug and then update this PR to wire them to GUS.

.git2gus/config.json
```diff
@@ -25,11 +25,11 @@
...
     "type:security": "BUG P2",
     "type:feedback": "",
-    "type:duplicate": "",
+    "status:duplicate": "",
```
peternhale, line 32: @daphne-sfdc what does this change affect?

daphne-sfdc: @peternhale I believe the person who originally created this config.json file was thinking that these 5 custom labels was used to categorize issues.  So an issue belongs in one of these categories - it's either a feature request, bug, security issue, general feedback for us, duplicate of another issue, or community contribution.  Thus, there are labels listed under `issueTypeLabels` that are not mapped to anything because we don't want to create GUS WIs for issues with those types of labels.

As for why I made that change, it's because the tag for duplicate issues is called `status:duplicate`, not `type:duplicate`.  I guess the tag name got changed between when the tag was first created and today.  I didn't get rid of Line 32 because of my assumption of the above, but really Line 32 does nothing.

What do you think - should I get rid of Line 32?  If yes, then the same applies to Line 31 because the `type:feedback` tag isn't mapped to anything either.

peternhale (approved):

PR #5341, opened 2024-01-17
fix: startupTime telemetry metrics for apex debugger & replay debugger

mingxuanzhangsfdx:
### What does this PR do?
startupTime was not a metrics in apex debugger & replay debugger, which makes it impossible to measure in the two extensions. This PR makes startupTime a measurement so we are able to view the metrics in azure appinsights. 

@W-14834514@

### Functionality Before
startupTime in apex debugger & replay debugger is not a log metrics.
### Functionality After
startupTime in apex debugger & replay debugger is a log metrics.

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
+++ b/packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
@@ -37,10 +37,15 @@ export class TelemetryService {
   public sendExtensionActivationEvent(hrstart: [number, number]): void {
     if (this.reporter !== undefined && this.isTelemetryEnabled) {
       const startupTime = this.getEndHRTime(hrstart);
-      this.reporter.sendTelemetryEvent('activationEvent', {
-        extensionName: EXTENSION_NAME,
-        startupTime
-      });
+      this.reporter.sendTelemetryEvent(
+        'activationEvent',
+        {
+          extensionName: EXTENSION_NAME
+        },
+        {
+          startupTime
+        }
+      );
     }
   }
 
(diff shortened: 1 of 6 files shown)


CristiCanizales (approved):
LGTM! Thanks for this fix Mingxuan!

:shipit:

packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
```diff
@@ -52,8 +57,8 @@ export class TelemetryService {
     }
   }
 
-  private getEndHRTime(hrstart: [number, number]): string {
+  private getEndHRTime(hrstart: [number, number]): number {
```
diyer, line 60: what does this change do?

diyer: was this the root cause of the issue?

diyer:
Thanks Mingxuan! How can this be tested locally?

packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
mingxuanzhangsfdx: `public sendTelemetryEvent(
    eventName: string,
    properties?: { [key: string]: string },
    measurements?: { [key: string]: number }
  ): void`

startupTime should be passed in via measurements, so that it can be viewed as a log metrics. So I need to change the function to return a number type. So measurements has a parameter like this:
`{
'startupTime': 1
}`

mingxuanzhangsfdx:
> Thanks Mingxuan! How can this be tested locally?

@diyer We run extensions with the change, and we should be able to see the metrics on azure after a while, which means when we specify the metrics startupTime and we set up the filter for extensionName, apex debugger & replay debugger shows up. I will do that after training.

PR #5342, opened 2024-01-17
fix: exclude source-tracking from apex ext bundle

peternhale:
@W-14840136@
Exclude @salesforce/source-tracking from being bundled with apex extension.
When included in the bundle, Apex extension activation fails.


gbockus-sf (approved):
Looks good 👍

daphne-sfdc (approved):
Approved! 🎉 

- Able to install VSIXs and confirm that the Apex extension activates properly ✅ 
- E2E test run here: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7572204162 (reruns due to flappy tests) ✅

PR #5343, opened 2024-01-18
revert: comment out Verify CLI Version feature

daphne-sfdc:
Comment out call to `verifyCliInstallationAndVersion()` in `packages/salesforcedx-vscode-core/src/index.ts` to temporarily disable the Verify CLI Version on Startup feature.

[skip-validate-pr]


peternhale (approved):

smithgp:
Howdy, just saw this revert of the original feature. Just a head up that the validation was also breaking the [analyticsdx-vscode](https://github.com/forcedotcom/analyticsdx-vscode) extensions and tests. If the validation is going to be brought back, please know that my team and I have an extension that depends on salesforcedx-vscode-core currently as well, and are expecting the old behavior (i.e. don't fail activation if cli isn't installed, but do show a warning of some kind), so it would be good if we can be kept in the loop. Thanks!

Issue #5358, opened 2024-01-24
An unsupported Java version was detected. Java 17. Oracle OpenJDK


Blackbackroom:
I have installed Java 17 on a laptop, Java Home is set but I still receive "An unsupported Java version was detected." message.
![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/40137179/9fdf6e40-e9e2-49b7-816f-f7c5267a8434)

OS Details:
Edition	Windows 11 Enterprise
Version	21H2
Installed on	‎10/‎27/‎2023
OS build	22000.2652
Experience	Windows Feature Experience Pack 1000.22001.1000.0

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/40137179/c5c31322-834c-4dd2-88d6-c8ce1da8b83d)

I set Java Home and have updated Path variable.
In the VS Code Apex extension settings I tried to set up Java Home there in a different ways:
in settings
```
C:\Program Files\Java\jdk-17
C:\Program Files\Java\jdk-17\bin
```
In settings.json file it was represented with double backslashes.

Even if I remove this setting or leave it empty, the Apex extension doesn't use the one set in path or JAVA_HOME

```
**java -version**
java version "17" 2021-09-14 LTS
Java(TM) SE Runtime Environment (build 17+35-LTS-2724)
Java HotSpot(TM) 64-Bit Server VM (build 17+35-LTS-2724, mixed mode, sharing)

**Get-Command java**
CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     java.exe                                           17.0.0.0   C:\Program Files\Java\jdk-17\bin\java.exe
```

The installed java is Oracle OpenJDK(x64)_17.0_L_EN_01.
The laptop belongs to a company, so I can not install everything without bureaucracy and chain of approvals.

```
**sf --version**
@salesforce/cli/2.23.20 win32-x64 node-v20.10.0

**sfdx --version**
@salesforce/cli/2.23.20 win32-x64 node-v20.10.0
```
![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/40137179/1a59448a-3acd-481e-9df8-08213d10850f)

**Apex extension version v59.13.0**
![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/40137179/eb152184-968e-49d0-864c-f40c42651032)

**VS Code version 1.85.2**
[...]

PR #5363, opened 2024-01-25
fix: check java version was criteria was ambiguous

peternhale:
Modified the check to include java option `XshowSettings:properties`

@W-14888195@ #5358 

### Functionality Before
The validation of java version used a string match for `build 17.` and the customer's java version reported `build 17+` causing a false positive when check version during startup

### Functionality After
The validation function has been changed to run the java version command with an option that displays all of the assign System properties the JVM will be using. One of the is `java.version = nn[.nn[.nn]]`. This representation is more reliable that using the `build` string from the java version command.

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/requirements.ts
+++ b/packages/salesforcedx-vscode-apex/src/requirements.ts
@@ -102,14 +102,11 @@ const isLocal = (javaHome: string): boolean => {
 export const checkJavaVersion = async (javaHome: string): Promise<boolean> => {
   return new Promise((resolve, reject) => {
     cp.execFile(
-      javaHome + '/bin/java',
-      ['-version'],
+      javaHome + '/bin/javax',
+      ['-XshowSettings:properties', '-version'],
       {},
       (error, stdout, stderr) => {
-        if (
-          stderr.indexOf('build 11.') < 0 &&
-          stderr.indexOf('build 17.') < 0
-        ) {
+        if (!/java\.version\s*=\s*(?:11|17)/g.test(stderr)) {
           reject(nls.localize('wrong_java_version_text', SET_JAVA_DOC_LINK));
         } else {
           resolve(true);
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/src/requirements.ts
```diff
@@ -102,14 +102,11 @@ const isLocal = (javaHome: string): boolean => {
 export const checkJavaVersion = async (javaHome: string): Promise<boolean> => {
   return new Promise((resolve, reject) => {
     cp.execFile(
-      javaHome + '/bin/java',
-      ['-version'],
+      javaHome + '/bin/javax',
+      ['-XshowSettings:properties', '-version'],
```
peternhale, line 106: Adding this property to the command will cause java to emit to stderr all of the System properties set in the VM, including `java.version`. I kept the `-version` option so that java does not emit the default help text.

packages/salesforcedx-vscode-apex/src/requirements.ts
```diff
@@ -102,14 +102,11 @@ const isLocal = (javaHome: string): boolean => {
...
       {},
       (error, stdout, stderr) => {
-        if (
-          stderr.indexOf('build 11.') < 0 &&
-          stderr.indexOf('build 17.') < 0
-        ) {
+        if (!/java\.version\s*=\s*(?:11|17)/g.test(stderr)) {
```
peternhale, line 109: Use a regex to validate version numbers

packages/salesforcedx-vscode-apex/test/jest/languageUtils/requirements.test.ts
peternhale, line 1: Moved the test from integration to jest.

daphne-sfdc (approved):
Approved! 🎉 

Check Java Version worked for the following Java versions:
1. Salesforce-approved Java 11 use for Core development ✅ 
2. Zulu Java 17 ✅ 
3. Oracle Java 17 ✅

daphne-sfdc (approved):
Approved! 🎉 

Test Cases (run using VSIXs generated from Commit Workflow):
1. Leave the Java Home setting blank -> Java Home should take the value of the $JAVA_HOME environment variable set locally ✅ (this was working as expected for Pete but Daphne has a system issue)
2. Invalid Java version (Java 21) -> should get `wrong_java_version_text` error message ✅ 
3. Path to random empty folder -> should get `wrong_java_version_text` error message ✅ 
4. Salesforce-approved Java 11 use for Core development ✅ 
5. Zulu Java 17 ✅ 
6. Oracle Java 17 ✅

PR #5388, opened 2024-01-31
fix: update pull commands to sf style

CristiCanizales:
### What does this PR do?
- Updates force:source:pull command to SF style

@W-14717014@


packages/salesforcedx-vscode-core/src/constants.ts
```diff
@@ -62,7 +62,7 @@ export const FUNCTIONS_PATH = '/functions/';
 
 // Commands
 export const ORG_OPEN_COMMAND = 'sfdx.org.open';
-export const FORCE_SOURCE_PULL_LOG_NAME =
-  'force_source_pull_default_scratch_org';
-export const FORCE_SOURCE_PUSH_LOG_NAME =
+export const PROJECT_RETRIEVE_START_LOG_NAME =
+  'project_retrieve_start_default_scratch_org';
+export const PROJECT_DEPLOY_START_LOG_NAME =
```
daphne-sfdc, line 67: Good catch 👍

packages/salesforcedx-vscode-core/src/commands/projectRetrieveStart.ts
```diff
@@ -55,13 +55,13 @@ export class ForceSourcePullExecutor extends SfdxCommandletExecutor<{}> {
     const builder = new SfdxCommandBuilder()
       .withDescription(nls.localize(this.params.description.default))
       .withArg(this.params.command)
-      .withJson()
+      .withJson(false)
```
daphne-sfdc, line 58: What is the difference between `.withJson(false)` and just completely taking out this line?

CristiCanizales: The false in that method is to not include the --loglevel flag. If we remove the line we lose the --json flag as well

packages/salesforcedx-vscode-core/package.nls.ja.json
```diff
@@ -56,9 +53,12 @@
   "org_logout_all_text": "SFDX: すべての認証済み組織からログアウト",
   "org_logout_default_text": "SFDX: Log Out from Default Org",
   "org_open_default_scratch_org_text": "SFDX: デフォルトの組織を開く",
-  "ignore_conflicts_on_push_description": "保存時の project:deploy:start コマンド実行時に、常に --ignore-conflicts オプションを使用するかどうかを指定します。",
+  "project_deploy_start_default_org_text": "SFDX: デフォルトのスクラッチ組織へソースをプッシュ",
+  "project_deploy_start_ignore_conflicts_default_org_text": "SFDX: デフォルトのスクラッチ組織へソースをプッシュして競合を上書き",
   "project_generate_text": "SFDX: プロジェクトを作成",
   "project_generate_with_manifest_text": "SFDX: マニフェストファイルを使用してプロジェクトを作成",
+  "project_retrieve_start_default_org_text": "SFDX: デフォルトのスクラッチ組織からソースをプル",
+  "project_retrieve_start_ignore_conflicts_default_org_text": "SFDX: デフォルトのスクラッチ組織からソースをプルして競合を上書き",
```
daphne-sfdc, line 61: Old command for pull and override conflicts

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -100,8 +100,8 @@ export const messages = {
     'There was an unexpected error when processing the org open response.',
   org_open_container_mode_message_text:
     'Access org %s as user %s with the following URL: %s',
-  force_source_pull_default_org_text: 'SFDX: Pull Source from Default Org',
-  force_source_pull_force_default_org_text:
+  project_retrieve_start_default_org_text: 'SFDX: Pull Source from Default Org',
+  project_retrieve_start_ignore_conflicts_default_org_text:
     'SFDX: Pull Source from Default Org and Override Conflicts',
```
daphne-sfdc, line 105:
```suggestion
    'SFDX: Pull Source from Default Org and Ignore Conflicts',
```

packages/salesforcedx-vscode-core/test/jest/commands/util/data/testData.ts
```diff
@@ -6,4 +6,4 @@
...
+  '{\n  "status": 0,\n  "result": {\n    "files": [\n      {\n        "state": "Created",\n        "fullName": "F10",\n        "type": "ApexClass",\n        "filePath": "/Users/kenneth.lewis/scratchpad/NewProj1/force-app/main/default/classes/F10.cls"\n      },\n      {\n        "state": "Created",\n        "fullName": "F10",\n        "type": "ApexClass",\n        "filePath": "/Users/kenneth.lewis/scratchpad/NewProj1/force-app/main/default/classes/F10.cls-meta.xml"\n      },\n      {\n        "state": "Changed",\n        "fullName": "Admin",\n        "type": "Profile",\n        "filePath": "/Users/kenneth.lewis/scratchpad/NewProj1/force-app/main/default/profiles/Admin.profile-meta.xml"\n      }\n    ]\n  },\n  "warnings": [\n    "We plan to deprecate this command in the future. Try using the \\"project retrieve start\\" command instead.",\n    "The loglevel flag is no longer in use on this command. You may use it without error, but it will be ignored.\\nSet the log level using the `SFDX_LOG_LEVEL` environment variable."\n  ]\n}\n';
```
daphne-sfdc, line 9: The message says `"We plan to deprecate this command in the future. Try using the \\"project retrieve start\\" command instead."`  So it's outdated and no longer applies, right?

daphne-sfdc (approved):
Approved! 🎉

This is what I tested:

• SFDX: Pull Source from Default Org (pull Dreamhouse into an empty project successfully) ✅
• SFDX: Pull Source from Default Org (no changes made -> no results found, pull successfully) ✅
• SFDX: Pull Source from Default Org (change file in org, local file overridden by org file after pull) ✅
• SFDX: Pull Source from Default Org (change file in local, no results found for pull) ✅
_Note: This is the same case as push source - I expected my local file to be overridden by the version that’s in the org, as that is what I see when I run retrieve.  But I’m going to accept the fact that nothing got pulled because that’s the behavior in the currently published version._
Screenshot 2024-02-07 at 11 20 36 AM
• SFDX: Pull Source from Default Org (failed because of a conflict) ✅
• SFDX: Pull Source from Default Org (failed because of a non-conflict error) - could not test this case because you cannot save an Apex class in an org if the code doesn’t compile

• SFDX: Pull Source from Default Org and Ignore Conflicts (pull Dreamhouse into an empty project successfully) ✅
• SFDX: Pull Source from Default Org and Ignore Conflicts (no changes made -> no results found, pull successfully) ✅
• SFDX: Pull Source from Default Org and Ignore Conflicts (change file in org, local file overridden by org file after pull) ✅
• SFDX: Pull Source from Default Org and Ignore Conflicts (change file in local, no results found for pull) ✅
_Note: Same as above - not the behavior I would expect because it differs from retrieve but it’s what’s already there in the published version._
Screenshot 2024-02-07 at 11 37 32 AM
• SFDX: Pull Source from Default Org and Ignore Conflicts (overrode a conflict successfully) ✅
• SFDX: Pull Source from Default Org and Ignore Conflicts (failed because of a non-conflict error) - could not test this case because you cannot save an Apex class in an org if the code doesn’t compile

PR #5403, opened 2024-02-06
fix: Org Browser refresh works for custom objects with a namespace

CristiCanizales:
### What does this PR do?
- Fixes extractJsonObject() to handle right upon receiving a plain text
- Fixes buildComponentsList() to consider namespace when there is one

@W-14983578@

### Functionality Before
- Couldn't refresh Custom Objects with a namespace
image

### Functionality After
- Custom Objects with a namespace can be successfully refreshed


packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,9 +14,12 @@ export const isNullOrUndefined = (object: any): object is null | undefined => {
 };
 
 export const extractJsonObject = (str: string): any => {
-  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
+  const isPlainText = str.includes('{') ? false : true;
```
peternhale, line 17: @CristiCanizales why is this change necessary?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts
```diff
@@ -154,7 +154,7 @@ export class MetadataOutlineProvider
   }
 }
 
-export function parseErrors(error: any): Error {
+export const parseErrors = (error: any): Error => {
```
peternhale, line 157: thank you

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
```diff
@@ -149,6 +149,12 @@ describe('build metadata components list', () => {
           fullName: 'fakeName1',
           type: 'ApexClass',
           manageableState: 'unmanaged'
+        },
+        {
+          fullName: 'fakeName3',
+          type: 'ApexClass',
+          manageableState: 'unmanaged',
+          namespacePrefix: 'sf_namespace'
```
peternhale, line 157: I believe manageableState needs to be 'managed' when namespace is present. Are you seeing something different?

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
```diff
@@ -67,12 +67,14 @@ export class ComponentUtils {
...
         cmpArray = cmpArray instanceof Array ? cmpArray : [cmpArray];
         for (const cmp of cmpArray) {
-          const { fullName, manageableState } = cmp;
+          const { fullName, manageableState, namespacePrefix } = cmp;
           if (
             !isNullOrUndefined(fullName) &&
             validManageableStates.has(manageableState)
           ) {
-            components.push(fullName);
+            components.push(
```
peternhale, line 75: Is this going to be the label in the tree? If so, is there an alternative to keeping track of namespace and track namespace as part of the tree node?

peternhale (commented):
Questions for now.

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
CristiCanizales: Yes! When debugging I saw unmanaged

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
CristiCanizales: I didn't get the follow up question, but yes, it would be the new label in the tree

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
diyer: Does this mean unmanaged package cannot have a namespace?
From this https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_create_namespace.htm looks like unmanaged package can also have a namespace.

Question: what will be the behavior for org browser refresh for a custom object in case of 1) manage package and 2) unmanaged package

diyer:
Thanks @CristiCanizales for fixing this !

packages/salesforcedx-vscode-core/test/vscode-integration/orgBrowser/metadataCmp.test.ts
CristiCanizales: No, precisely what I'm saying is that when debugging, the example was unmanaged with a namespace. In the code, whatever between the following options would be valid
image

diyer: ok. Did we get a repro of this locally with other projects? We should definitely do that to test this fix and to make sure we dont regress other things.

CristiCanizales: Yes! I tested with dreamhouse-lwc with my dev org, with namespace "guariqueteo"
image

packages/salesforcedx-vscode-core/src/orgBrowser/nodeTypes.ts
```diff
@@ -76,9 +78,12 @@ export class BrowserNode extends vscode.TreeItem
     }
 
     fullNames.forEach(fullName => {
+      const hasNamespacePrefix = fullName.includes('__');
```
peternhale, line 81: @CristiCanizales this won't work
```
'namespace__Broker__c'.includes('__');
true
> 'Broker__c'.includes('__');
true
```

peternhale: Consider using a regex to find a namespace in the fullname.

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
peternhale: The original code's intent was to return a parsed json string.

As you state, this function is being called with an invalid json string.

While this change avoids parsing an invalid json string, it also changes the intent of the function.

The risk here is returning a string to a calling function that is expecting an object. If the caller is assuming an object and tries to access an expected property on an object and this function returns a string, the access will fail.

Have a look at alternatives on how to handle an invalid json string.

packages/salesforcedx-vscode-core/src/orgBrowser/metadataCmp.ts
peternhale: Do we still need this change?

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -14,9 +14,13 @@ export const isNullOrUndefined = (object: any): object is null | undefined => {
 };
 
 export const extractJsonObject = (str: string): any => {
-  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
-
-  return JSON.parse(jsonString);
+  const isJsonString = str.indexOf('{') !== -1 && str.lastIndexOf('}') !== -1;
```
peternhale, line 17: @CristiCanizales would you some unit tests for this change?

peternhale (changes requested):

peternhale (approved):

PR #5407, opened 2024-02-09
fix: logging out from default org only shows 1 notif

CristiCanizales:
### What does this PR do?
- Fixes an issue when logging out from default org, which would cause 2 notifications (1 for setting default org and another one for logging out) and not just the 2nd one

@W-12431853@

### Functionality Before
- successful default org set and successful logout from default org notifications show up

### Functionality After
- Only successful logout notification shows up


packages/salesforcedx-vscode-core/src/commands/auth/orgLogout.ts
```diff
@@ -113,7 +113,7 @@ export async function orgLogoutDefault() {
 }
 
 async function removeUsername(username: string) {
-  await configSet('');
+  await ConfigUtil.setDefaultUsernameOrAlias('');
```
peternhale, line 116: @CristiCanizales core Config class has an unset method. I suggest creating a new function in ConfigUtil to unset target org.

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/config/configUtil.ts
```diff
@@ -180,6 +180,33 @@ export class ConfigUtil {
     return info.aliases.getUsername(usernameOrAlias) || usernameOrAlias;
   }
 
+  public static async unsetTargetOrg(usernameOrAlias: string) {
+    const originalDirectory = process.cwd();
+    // In order to correctly setup Config, the process directory needs to be set to the current workspace directory
+    const workspacePath = workspaceUtils.getRootWorkspacePath();
+    try {
+      // checks if the usernameOrAlias is non-empty and active.
+      if (usernameOrAlias) {
```
peternhale, line 189: @CristiCanizales Thank you for creating a new function. I was wondering, given target  org is a config setting, why this check is necessary, or even needed as a parameter?. Throwing when the org is expired could be too much, given that unsetting target org is what is necessary. What do you think?

CristiCanizales: Yeah you're right. I was thinking you needed to have the right connection to the default org in order to logout, but that's handled apart so we don't need it in this new unset function. thanks!

peternhale (approved):

PR #5408, opened 2024-02-09
chore: add GHA for running script to report installs

gbockus-sf:
### What does this PR do?
- Add a script for gather install info from the MS & OPENVSX marketplace (via the cli for each) and report to appinsights
- Add a GHA scheduled to run hourly to execute the above script

This will enable us to graph the installations over time for each extension as well as the grand total. 

@W-14926512@

### Functionality Before
Had to go to marketplace for install info

### Functionality After
install info is in app insights. 

Ran this once to verify things would make it to appinsights. 
 Query: (02/09/2024)
```
customEvents
| where name has 'installStats'
```


.github/workflows/reportInstalls.yml
```diff
@@ -0,0 +1,22 @@
+name: Report Installs
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 * * * *'
```
gbockus-sf, line 5: Run every hour https://crontab.guru/every-1-hour

.github/workflows/reportInstalls.yml
```diff
@@ -0,0 +1,22 @@
...
+      
+jobs:
+  report-installs:
+    runs-on: ubuntu-latest
+    env: 
+      VSCE_PERSONAL_ACCESS_TOKEN: ${{ secrets.VSCE_PERSONAL_ACCESS_TOKEN }}
+      OVSX_PAT: ${{ secrets.OVSX_PAT }}
```
gbockus-sf, line 12: secrets needed for the CLIs.

.github/workflows/reportInstalls.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+      - run: npm install
```
gbockus-sf, line 19: Note: running npm install b/c I need dev deps to run the script (ts-node)

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { promisify } from 'util';
+import { exec } from 'child_process';
+import * as appInsights from 'applicationinsights';
+import { DEFAULT_AIKEY } from '../packages/salesforcedx-utils-vscode/src/constants';
+
+const promisifiedExec = promisify(exec);
```
gbockus-sf, line 12: Wanted a promisfied exec.  Lemme know if there's an better way.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+  averagerating: string;
+  trendingdaily: string;
+  trendingmonthly: string;
+  trendingweekly: string;
+  updateCount: string;
+  weightedRating: string;
+  downloadCount: string;
+}
+
+type TelemetryData = InstallTelemetryData | InstallTelemetryDataMS;
```
gbockus-sf, line 35: Tried to ensure we send the data we think we're sending and if we change it the type system will let us know.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+import { promisify } from 'util';
+import { exec } from 'child_process';
+import * as appInsights from 'applicationinsights';
+import { DEFAULT_AIKEY } from '../packages/salesforcedx-utils-vscode/src/constants';
+
+const promisifiedExec = promisify(exec);
+
+interface InstallTelemetryData {
+  // To align with the properties property for appInsightsClient.trackEvent
+  [key: string]: string;
```
gbockus-sf, line 16: kinda weird, but this if the type required by appinsights method call.  It at least ensures all the property values are strings.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+  trendingweekly: string;
+  updateCount: string;
+  weightedRating: string;
+  downloadCount: string;
+}
+
+type TelemetryData = InstallTelemetryData | InstallTelemetryDataMS;
+
+// Set to true for local development
+const DRY_RUN = false;
```
gbockus-sf, line 38: Added for local development to avoid spamming appinsights.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+
+execute()
+  .then(() => {
+    logger('reportInstalls script completed successfully');
+    process.exit(0);
+  })
+  .catch(err => {
+    logger('Error running reportInstalls script', err);
+    process.exit(1);
+  });
```
gbockus-sf, line 160: output includes all the stdout from the cli calls, so it's pretty verbose, but here's an example log

```
*** Reporting install count for salesforcedx-einstein-gpt {
  "name": "salesforcedx-einstein-gpt/installStats",
  "properties": {
    "install": "15312",
    "averagerating": "3.4000000953674316",
    "ratingcount": "5",
    "trendingdaily": "0.013083011709295479",
    "trendingmonthly": "12.827892980964217",
    "trendingweekly": "2.740890953097403",
    "updateCount": "99065",
    "weightedRating": "4.068803184661991",
    "downloadCount": "282",
    "marketplace": "MS"
  }
}
```

peternhale (approved):

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+  install: string;
+  ratingcount: string;
+  marketplace: string;
+  'common.extname': string;
+}
+
+// These properties match the values provided in the statistics array from the
+// vsce cli output
+interface InstallTelemetryDataMS extends InstallTelemetryData {
+  averagerating: string;
```
randi274, line 26: I know this isn't on us, but ugh to inconsistent casing

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+  weightedRating: string;
+  downloadCount: string;
+}
+
+type TelemetryData = InstallTelemetryData | InstallTelemetryDataMS;
+
+// Set to true for local development
+const DRY_RUN = false;
+const extensionsToTrack = [
+  'salesforcedx-vscode-core',
```
randi274, line 40: Probably a selfish ask, but can you put gpt first? I think that's the one that has the highest visibility, so if we're reporting them at all, we should make sure that that one goes in.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+  'salesforcedx-vscode-apex',
+  'salesforcedx-vscode-apex-debugger',
+  'salesforcedx-vscode-apex-replay-debugger',
+  'salesforcedx-vscode-lightning',
+  'salesforcedx-vscode-lwc',
+  'salesforcedx-vscode-visualforce',
+  'salesforcedx-vscode',
+  'salesforcedx-vscode-expanded',
```
randi274, line 48: I think this one might be causing errors - I'm not seeing anything past this in telemetry

randi274 (changes requested):
Looks good overall! However I think something's going wrong with '-expanded', and I'm not seeing GPT installs either. Locally it looks like everything is fine though. FWIW I think we should move GPT to be first in that list to make sure it's always being added if any install events are recorded.

Screenshot 2024-02-09 at 1 23 39 PM

Screenshot 2024-02-09 at 1 23 39 PM

.github/workflows/reportInstalls.yml
gbockus-sf: good call...we should be using the org secret.  I just used whatever I found in our existing GHA

scripts/reportInstalls.ts
gbockus-sf: Hopefully fixed by adding a flush to the app insight client before exit.

scripts/reportInstalls.ts
```diff
@@ -0,0 +1,160 @@
...
+const logger = (msg: string, obj?: any) => {
+  if (!obj) {
+    console.log(`*** ${msg}`);
+  } else {
+    console.log(`*** ${msg}`, JSON.stringify(obj, null, 2));
+  }
+};
+
+// Lifted from https://github.com/forcedotcom/salesforcedx-vscode/blob/361a12dcdcb779cfd6aca844758361cf4947d0d1/packages/salesforcedx-utils-vscode/src/telemetry/telemetryReporter.ts#L83
+const setupAppInsights = () => {
```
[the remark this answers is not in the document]
gbockus-sf: yeah this is a thorny one.  We probably need a whole assessment for what we use appinsights. I hesitate to create a WI as it'd be more like an epic :) Def a good callout but I lean towards doing it ok for now and handling more if there are further use cases for this. We'd also need to understand sharing between extension code and script code.

randi274 (approved):
There they are! Looks good 🚢

PR #5419, opened 2024-02-14
chore: add telemetry for when orgId is not assigned

gbockus-sf:
### What does this PR do?
We have some records that make it to appinsights where there is no org_id value associated with the telemetry records.  This occurs when we fail to retrieve the orgId in the workspaceContext handleConfigChange method.  

In tracking this down in appinsights I found a few specific users that were impacted for a period of time.  There wasn't any particular condition that appeared to cause the issue.  I attempted to find the error condition but was never able to repo the issue.  This change will report the error to appinsights so we have a starting point for tracking down when this situation occurs.

@W-14990581@

### Functionality Before
only logged when orgId could not be assigned on config change 

### Functionality After
Send exception to telemetry when orgId lookup fails.

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
+++ b/packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
@@ -7,7 +7,7 @@
 
 import { AuthInfo, Connection, StateAggregator } from '@salesforce/core';
 import * as vscode from 'vscode';
-import { ConfigAggregatorProvider } from '..';
+import { ConfigAggregatorProvider, TelemetryService } from '..';
 import { ConfigUtil } from '../config/configUtil';
 import { projectPaths } from '../helpers';
 import { nls } from '../messages';
@@ -16,6 +16,7 @@ export interface OrgUserInfo {
   alias?: string;
 }
 
+export const WORKSPACE_CONTEXT_ORG_ID_ERROR = 'workspace_context_org_id_error';
 /**
  * Manages the context of a workspace during a session with an open SFDX project.
  */
@@ -100,12 +101,15 @@ export class WorkspaceContextUtil {
       try {
         const connection = await this.getConnection();
         this._orgId = connection?.getAuthInfoFields().orgId;
-      } catch (error) {
+      } catch (error: unknown) {
         this._orgId = '';
-        console.log(
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          `There was an problem getting the orgId of the default org: ${error}`
-        );
+        if(error instanceof Error) {
+          console.log(
+            'There was an problem getting the orgId of the default org: '
+            , error
+          );
+          TelemetryService.getInstance().sendException(WORKSPACE_CONTEXT_ORG_ID_ERROR, `name: ${error.name}, message: ${error.message}`);
+        }
       }
     } else {
       this._username = undefined;
--- a/packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
@@ -7,8 +7,9 @@
 
 import { AuthInfo, Connection, StateAggregator } from '@salesforce/core';
 import * as vscode from 'vscode';
-import { ConfigAggregatorProvider, WorkspaceContextUtil } from '../../../src';
+import { ConfigAggregatorProvider, TelemetryService, WorkspaceContextUtil } from '../../../src';
 import { ConfigUtil } from '../../../src/config/configUtil';
+import { WORKSPACE_CONTEXT_ORG_ID_ERROR } from '../../../src/context/workspaceContextUtil';
 import { nls } from '../../../src/messages';
 jest.mock('@salesforce/core');
 jest.mock('../../../src/config/configUtil');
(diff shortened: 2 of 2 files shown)


packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -100,12 +101,15 @@ export class WorkspaceContextUtil {
...
-        console.log(
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          `There was an problem getting the orgId of the default org: ${error}`
-        );
+        if(error instanceof Error) {
+          console.log(
+            'There was an problem getting the orgId of the default org: '
+            , error
+          );
+          TelemetryService.getInstance().sendException(WORKSPACE_CONTEXT_ORG_ID_ERROR, `name: ${error.name}, message: ${error.message}`);
```
gbockus-sf, line 111: I wanted to ensure that we could identify these errors so I gave it a unique name instead of using the name on the error.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -100,12 +101,15 @@ export class WorkspaceContextUtil {
...
+      } catch (error: unknown) {
         this._orgId = '';
-        console.log(
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          `There was an problem getting the orgId of the default org: ${error}`
-        );
+        if(error instanceof Error) {
+          console.log(
+            'There was an problem getting the orgId of the default org: '
+            , error
```
gbockus-sf, line 109: pulled error out of the string to make eslint happy without the rule bypass.

packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
```diff
@@ -52,16 +56,18 @@ describe('WorkspaceContextUtil', () => {
       'clearInstance'
     );
 
-    mockFileSystemWatcher = (vscode.workspace
-      .createFileSystemWatcher as any).mockReturnValue(mockWatcher);
+    mockFileSystemWatcher = mockedVSCode.workspace
```
gbockus-sf, line 59: if we use the `jest.mocked` function we can add return/resolve calls to known mocked modules.

packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
```diff
@@ -52,16 +56,18 @@ describe('WorkspaceContextUtil', () => {
...
+      .createFileSystemWatcher.mockReturnValue(mockWatcher as any);
 
-    getUsernameOrAliasStub = (configUtilMock.getDefaultUsernameOrAlias as any).mockReturnValue(testAlias);
-    getUsernameStub = (configUtilMock.getUsernameFor as any).mockReturnValue(testUser);
+    getUsernameOrAliasStub = configUtilMock.getDefaultUsernameOrAlias.mockResolvedValue(testAlias);
+    getUsernameStub = configUtilMock.getUsernameFor.mockResolvedValue(testUser);
```
gbockus-sf, line 63: Noticed these were returning when they should be resolving.

packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
```diff
@@ -73,6 +79,7 @@ describe('WorkspaceContextUtil', () => {
       'cliConfigWatcher',
       mockWatcher
     );
+    expect(mockFileSystemWatcher).toHaveBeenCalled();
```
gbockus-sf, line 82: this mock wasn't being verified

packages/salesforcedx-utils-vscode/test/jest/context/workspaceContextUtil.test.ts
```diff
@@ -112,7 +119,8 @@ describe('WorkspaceContextUtil', () => {
     expect(workspaceContextUtil.username).toEqual(testUser);
     expect(workspaceContextUtil.alias).toEqual(testAlias);
     expect(workspaceContextUtil.orgId).toEqual('');
-    expect(logMock.mock.calls[0][0].includes(dummyErrorMessage)).toBe(true);
+    expect(logMock).toHaveBeenCalled();
```
gbockus-sf, line 122: The actual log message didn't seem overly important.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -100,12 +101,15 @@ export class WorkspaceContextUtil {
...
         const connection = await this.getConnection();
         this._orgId = connection?.getAuthInfoFields().orgId;
-      } catch (error) {
+      } catch (error: unknown) {
         this._orgId = '';
-        console.log(
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          `There was an problem getting the orgId of the default org: ${error}`
-        );
+        if(error instanceof Error) {
```
randi274, line 106:
```suggestion
        if (error instanceof Error) {
```

randi274: I'm kinda surprised prettier didn't catch this

randi274 (approved):
One small suggestion, otherwise looks good to me

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
gbockus-sf: Good eye..yeah I'm not sure how well prettier is integrated in this repo.  The chances are low, but error type validation is just a thing you have to do in Typescript.

peternhale (approved):

PR #5440, opened 2024-02-19
Cristi/w 14727780 and w 15063547

CristiCanizales:
### What does this PR do?
- Updates lwc commands from sfdx to sf style
- Deprecates local dev server commands

@W-14727780@
@W-15063547

### Testing notes
- We need manual testing for these commands as we don't have e2e tests for them


daphne-sfdc:
@CristiCanizales Can you get rid of [lwc-localdev-supported-modules.md](https://github.com/forcedotcom/salesforcedx-vscode/blob/cristi/w-14727780-and-w-15063547/docs/_articles/en/localdev/lwc-localdev-supported-modules.md) and its Japanese translation?  (This would also make the localdev folder empty for both English and Japanese so the folders can be removed too.)

CristiCanizales:
Thanks Daphne, for reviewing and the QA! Though we also need to wait for the announcement and for some time afterwards for users to assimilate the deprecation🫠 ⏲️

PR #5441, opened 2024-02-20
Cristi/push fix

CristiCanizales:
### What does this PR do?
- Fixes error when trying to generate the output table from all the files involved in a push and some of the files didn't have errors. Now the files considered for the output table are only those that failed

### Functionality Before
- no error messages shown upon errors

### Functionality After
- Correct displaying of error messages

[skip-validate-pr]


daphne-sfdc (approved):
Approved! 🎉

- SFDX: Push Source to Default Org - error case shows error messages in Output tab ✅
- SFDX: Push Source to Default Org and Ignore Conflicts - error case shows error messages in Output tab ✅

PR #5443, opened 2024-02-21
fix: eliminate "Cannot read properties of undefined (reading type)" error

daphne-sfdc:
### What does this PR do?
Eliminates the "Cannot read properties of undefined (reading type)" error that happens when a user cancels a retrieve.

@W-15091722@

### Functionality Before
When a user clicks the Cancel button while a retrieve is in progress, they see the "Cannot read properties of undefined (reading type)" error.
Screenshot 2024-02-21 at 10 43 10 AM

### Functionality After
When a retrieve is cancelled, there is no error message and the command ends gracefully.
Screenshot 2024-02-21 at 11 59 35 AM


packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -280,7 +280,7 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
   protected async postOperation(
     result: RetrieveResult | undefined
   ): Promise<void> {
-    if (result) {
+    if (result && result.response.fileProperties !== undefined) {
```
peternhale, line 283: Can be done with
``` suggestion
    if (result?.response?.fileProperties !== undefined) {
```

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -280,7 +280,7 @@ export abstract class RetrieveExecutor<T> extends DeployRetrieveExecutor<T> {
   protected async postOperation(
     result: RetrieveResult | undefined
   ): Promise<void> {
-    if (result) {
+    if (result?.response?.fileProperties !== undefined) {
```
diyer, line 283: not sure if my comment got lost, so reposting.
not knowing the code well enough, a naive question. why is check fileProperties !== undefined not before calling setPropertiesForFilesRetrieve?

Also can setPropertiesForFilesRetrieve be called in other places? which means does this check need to go in setPropertiesForFilesRetrieve ?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
daphne-sfdc: I put my condition with the other condition because that was where it originally was.  It's pretty old code so I just kept the status quo. 🙂

I would assume it's because putting the if statement earlier means Lines 284-288 wouldn't be run unnecessarily.

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
diyer: Also as per this fix, everything after line 283 will be skipped even if result is not null . is this expected?

daphne-sfdc: As of now, `setPropertiesForFilesRetrieve()` is not called anywhere else, according to a search.
Screenshot 2024-02-21 at 12 33 16 PM

peternhale: @diyer That is a good point. The check `if (result?.response?.fileProperties !== undefined) {` guarantees the block will not be executed if any property named is undefined. The previous block ran when result was not undefined, it might be useful to better understand which parts of result are used within the block, including the `createOutput` call

daphne-sfdc: There's a slight difference in the contents of the Output tab.  If the check is moved down to Line 289, there will be an extra blank line in the Output tab before the ended statement because output is an empty line and channelService appends it.

That's the only difference that I see - result is not used anywhere else in that block.

Screenshot 2024-02-21 at 12 53 09 PM
Screenshot 2024-02-21 at 12 52 37 PM

peternhale (approved):

PR #5444, opened 2024-02-21
fix: add sf log level to current runtime

peternhale:
Enables debug logging for libraries that consume the env var `SF_LOG_LEVEL` by pushing the extension setting into the current process.env.

One should expect commands such as deploy to org or retrieve from org show show debug log entries to the sf log files, when the log level is set to the correct level used by the libraries.

Added validation of ext setting to ensure the setting values match those provided by core logging config.

@W-15062541@


packages/salesforcedx-vscode-core/src/settings/sfdxCoreSettings.ts
```diff
@@ -109,7 +109,7 @@ export class SfdxCoreSettings {
   public getSfLogLevel(): string {
     return this.getConfigValue(
       ENV_SF_LOG_LEVEL,
-      process.env.SF_LOG_LEVEL ?? ''
+      process.env.SF_LOG_LEVEL ?? 'debug'
```
mingxuanzhangsfdx, line 112: Given that the above default is fatal, why don't you keep it here same?

mingxuanzhangsfdx (approved):
With the change, I could see the vscode setting SF_LOG_LEVEL has the 6 options that match sf cli log level. With the debug level, I was able to see the log records in $HOME/.sf/sf-<datetime>.log whereas it is empty when it is fatal.

mingxuanzhangsfdx (approved):

PR #5482, opened 2024-02-27
refactor: remove sdr dependency from vscode apex

peternhale:
@W-12623227@
replace use of component set to find file paths with custom implementation

Should be no visible changes to the current user experience


packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
```diff
@@ -153,6 +154,60 @@ export class ApexLibraryTestRunExecutor extends LibraryCommandletExecutor<{}> {
...
+        pkgDir.fullPath
+      );
+      return `${relativePath}/**/*.cls`;
+    });
+
+    const findFilesPromises = patterns.map(pattern => {
+      return vscode.workspace.findFiles(pattern, '**/node_modules/**');
+    });
+
+    const filesWithDuplicates = (await Promise.all(findFilesPromises)).flat();
```
gbockus-sf, line 192: These two seem like good candidates for going in a shared utilty

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -68,7 +67,7 @@
   "scripts": {
     "bundle:extension": "npm run bundle:extension:build && npm run bundle:extension:copy",
     "bundle:extension:copy": "cp ./out/apex-jorje-lsp.jar ./dist/",
-    "bundle:extension:build": "esbuild ./src/index.ts  --bundle --outfile=dist/index.js --format=cjs --platform=node --external:vscode  --external:@salesforce/core --external:applicationinsights --external:shelljs --external:@salesforce/source-deploy-retrieve --external:@salesforce/source-tracking --minify",
```
gbockus-sf, line 71: This is worth a shot, but may still require SDR to be not bundled b/c this extension depends on `salesforcedx-utils-vscode` -> depend on SDR

peternhale: I ran E2E with this branch's vsix w/o marking sdr as external and tests passed. I also tried w.o source tracking, but that crashed and burned.

packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
peternhale: I'll have a look

packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
```diff
@@ -153,6 +154,60 @@ export class ApexLibraryTestRunExecutor extends LibraryCommandletExecutor<{}> {
...
+    // Iterate over each file found
+    for (const file of files) {
+      // Get the base name of the file without the extension
+      const fileName = path.basename(file.fsPath, '.cls');
+
+      // If the file name is in the testNames set, add it to the correlatedArtifacts map
+      if (correlatedArtifacts.has(fileName)) {
+        correlatedArtifacts.set(fileName, file.fsPath);
+      }
+    }
```
gbockus-sf, line 208: fancy stuff.  Just a dumb q. Did you consider just searching for each test class with `vscode.workspace.findFiles`?

gbockus-sf (commented):
Good stuff :+1:

packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
peternhale: @gbockus-sf I believe that is what line 188-190 are doing ;)

gbockus-sf: Yeah that does the find for all apex cls files.  We could do a find withe more specific matching for each test file we're looking for.  Then we wouldn't have to parse though all the results.  I'm sort of thinking through this being an issue for larger projects with thousands of cls files.  I'm not sure it's better b/c it if we're doing this for a condition where we have 100s of tests is going to have perf issues the other way (run 100's of finds vs just a single larger find) 

I don't think this is a big concern just figured I'd ask :)

peternhale: I thought about extending the pattern so I would include one pattern for each test. Not sure what the tolerance for findFiles is for pattern length, complexity, etc. I would think that the dir scan would be the slowest part of this whole process and regardless of how fine tuned the patterns are each entry in the directories have to be visited to check for a match.

gbockus-sf: makes sense.

gbockus-sf (commented):
This passes my sniff test.  Sorry for the delay.  I'll leave actual approval for the QE.

CristiCanizales (approved):
LGTM!
QE notes:
- tests can be run from sidebar, command palette, editor view ✅ 
- anon apex can be executed ✅ 
- results look as before ✅

PR #5500, opened 2024-03-12
fix: report right command execution status

CristiCanizales:
### What does this PR do?
- Reports the command execution the right way

### Functionality Before
- steps would always show failure notification

### Functionality After
- steps show right notification

[skip-validate-pr]


packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -444,9 +443,9 @@ export class IsvDebugBootstrapExecutor extends SfdxCommandletExecutor<{}> {
   ) {
     channelService.streamCommandOutput(execution);
     channelService.showChannelOutput();
-    notificationService.reportExecutionError(
```
diyer, line 447: can you explain this change please :)

CristiCanizales: so the function we were calling before was 'reportExecutionError', which as the name says is used for reporting errors. I just changed it to the other function that is used to report general command execution

daphne-sfdc (approved):
Approved! 🎉

I ran `SFDX: Create and Set Up Project for ISV Debugging` and was able to go through all the steps without seeing any error notifications.  The project was created successfully.

peternhale (approved):

packages/salesforcedx-vscode-core/src/commands/isvdebugging/bootstrapCmd.ts
diyer: got it. so now we will get a notification on command execution status?

PR #5507, opened 2024-03-15
fix: support folder with symbol link

mingxuanzhangsfdx:
### What does this PR do?
The vsce has been using `realpathSync.native(filePath)` to get the canonical pathname of a given path. So for customers that are using sfdx project with a symbolic link, it fails to run deploy or retrieve commands with error: “Error deploying or retrieving source: The file or directory that you tried to deploy or retrieve isn’t in a package directory that’s specified in your sfdx-project.json file. Add this location to your “packageDirectories” value, or deploy or retrieve a different file or directory. For details about sfdx-project.json, see: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm"

The reason why it uses `realpathSync.native(filePath)`  is that there was a [previous issue fix](https://github.com/forcedotcom/salesforcedx-vscode/pull/4362). So this PR adds a check based on that fix to detect if the path refers to somewhere under a symbolic link. If there is one, then we will return the exact symbolic link, rather than the original folder. So there will not be a mismatch.

With this update, I also tested the [scenario](https://github.com/forcedotcom/salesforcedx-vscode/pull/4362) of the previous issue. And it is also working fine.

@W-15247490@

### Functionality Before
Customers who work within a symbolic link are not able to use VSCE to run deploy and retrieve commands. 

To establish the symbolic link, 
in mac/unix, run `ln -s /path/to/target /path/to/link`,
in windows, run `mklink /j C:\LinkToFolder C:\TargetFolder`.

e.g.

1. Open the project folder in the VS code. Set the Salesforce CLI Extension version to 60.2.3.
2. Create one APEX/LWC/ any other component and try to deploy it.
3. It will get deployed.
4. Now Open the folder that you have mapped to the original project in VS code.
5. Make sure that your Salesforce CLI Extension is set to user 60.2.3 version.
6. Create one APEX/LWC/ any other component, and try to deploy it.
7. You will face the same error as the customer is facing.

Also there is no error from the [former issue.](https://github.com/forcedotcom/salesforcedx-vscode/pull/4362) 

### Functionality After

The functions of deploying and retrieving work fine in the symbolic link.

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -5,8 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { realpathSync, lstatSync } from 'fs';
-import { basename, resolve, dirname } from 'path';
+import { basename } from 'path';
 import { telemetryService } from '../telemetry';
 
 export const isNullOrUndefined = (object: any): object is null | undefined => {
@@ -35,7 +34,11 @@ export const flushFilePath = (filePath: string): string => {
     return filePath;
   }
 
-  let nativePath = isSymbolicLink(filePath)? filePath: realpathSync.native(filePath);
+  // let nativePath = realpathSync.native(filePath);
+  // Above is the original assigned nativePath value.
+  // We found that filePath is the correct path and the stale name issue
+  // no longer exists.
+  let nativePath = filePath;
   if (/^win32/.test(process.platform)) {
     // The file path on Windows is in the form of "c:\Users\User Name\foo.cls".
     // When called, fs.realpathSync.native() is returning the file path back as
@@ -58,22 +61,6 @@ export const flushFilePath = (filePath: string): string => {
     });
   }
   return nativePath;
-};
-
-export const isSymbolicLink = (path: string) => {
-  try {
-    let currentPath = resolve(path);
-    // track down the path until it is a root path
-    while(currentPath !== dirname(currentPath)) {
-      const stats = lstatSync(currentPath);
-      const isLink = stats.isSymbolicLink();
-      if (isLink) return true;
-      currentPath = dirname(currentPath);
-    }
-    return false;
-  } catch (err) {
-      throw new Error('The path does not exist');
-  }
 };
 
 export const flushFilePaths = (filePaths: string[]): string[] => {
--- a/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/utils.ts
@@ -61,9 +61,11 @@ export const flushFilePath = (filePath: string): string => {
 };
 
 const isSymbolicLink = (path: string) => {
+  // regex of windows root path
+  const winRootDir = /^[A-Za-z]:\\$/;
   try {
     let currentPath = resolve(path);
-    while(currentPath !== '/') {
+    while(currentPath !== '/'  && !winRootDir.test(currentPath)) {
       const stats = lstatSync(currentPath);
       const isLink = stats.isSymbolicLink();
       if (isLink) return true;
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -59,6 +60,21 @@ export const flushFilePath = (filePath: string): string => {
   return nativePath;
 };
 
+const isSymbolicLink = (path: string) => {
```
peternhale, line 63: @mingxuanzhangsfdx, nice job tracking this bug down. I had not considered flushFilepath as the culprit.

I would suggest that you try to find a more efficient means to detect a symlinked directory, other than climbing the path iteratively.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -59,6 +60,21 @@ export const flushFilePath = (filePath: string): string => {
...
+  try {
+    let currentPath = resolve(path);
+    while(currentPath !== '/') {
```
peternhale, line 66: Not sure the while condition will work on windows.

peternhale (changes requested):
Needs tests

diyer:
@mingxuanzhangsfdx do we know which change introduced this regression?

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -59,6 +60,21 @@ export const flushFilePath = (filePath: string): string => {
...
+      const stats = lstatSync(currentPath);
+      const isLink = stats.isSymbolicLink();
+      if (isLink) return true;
+      currentPath = dirname(currentPath);
+    }
+    return false;
+  } catch (err) {
+      throw new Error('The path does not exist');
```
diyer, line 74: Can we have more descriptive error message please? Also where would this error show up? output tab or dev tools?

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
mingxuanzhangsfdx: Thanks I just updated. The root path of windows is not similar with mac.

peternhale (changes requested):
Fix unit tests and improve efficiency of sym link detection

packages/salesforcedx-utils-vscode/test/jest/helpers/utils.test.ts
```diff
@@ -4,18 +4,30 @@
...
-import { extractJsonObject, flushFilePath } from '../../../src/helpers/utils';
+import { extractJsonObject, flushFilePath, isSymbolicLink } from '../../../src/helpers/utils';
 
 describe('flushFilePath', () => {
   let teleSpy: jest.SpyInstance;
+  let originalPath = '';
   beforeEach(() => {
     jest.mock('fs');
     jest.mock('../../../src/context/workspaceContextUtil');
+    originalPath = './test.txt';
```
peternhale, line 19: Unit tests should not use real file I/O. Please look into applying the needed mocks to eliminate the real files.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -59,6 +60,22 @@ export const flushFilePath = (filePath: string): string => {
   return nativePath;
 };
 
+export const isSymbolicLink = (path: string) => {
+  try {
```
peternhale, line 64: @mingxuanzhangsfdx this function needs to be optimized. It is visited quite a bit in our code.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -29,12 +29,13 @@ export const extractJsonObject = (str: string): any => {
...
 // To get around this, fs.realpathSync.native() is called to get the
 // URI with the actual file name.
+
 export const flushFilePath = (filePath: string): string => {
   if (filePath === '') {
     return filePath;
   }
 
-  let nativePath = realpathSync.native(filePath);
+  let nativePath = isSymbolicLink(filePath)? filePath: realpathSync.native(filePath);
```
peternhale, line 38: @mingxuanzhangsfdx it would be nice to not use `realpathSync.native` and use `realpathSync` instead. There are some limitations with `native` function on windows with the primary being max file path is limited to 260 characters, which could become an issue.

peternhale (approved):

PR #5508, opened 2024-03-15
chore: show correct input values in Slack notifications for scheduled runs of Base E2E Suite

daphne-sfdc:
### What does this PR do?
Show the correct values for **VSCode version** and **E2E branch** inputs in Slack notifications for scheduled runs of Base E2E suite that are triggered by other workflows, instead of a blank value.

Blank value was being shown because the default VSCode version and E2E branch were being used, and default values were treated as blank instead of their actual value when displayed in human-readable format.

Screenshot 2024-03-15 at 11 15 26 AM

[skip-validate-pr]

### DIFF
--- a/.github/workflows/baseSuiteE2E.yml
+++ b/.github/workflows/baseSuiteE2E.yml
@@ -188,7 +188,7 @@ jobs:
         lwcLSP,
         deployAndRetrieve,
         apexLSP,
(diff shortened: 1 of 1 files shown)


CristiCanizales (approved):
lgtm!

:shipit:

PR #5513, opened 2024-03-20
Cristi/update adapter

CristiCanizales:
### What does this PR do?
- Updates debuggers with new vscode adapter

@W-15391003@


packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
```diff
@@ -596,12 +593,14 @@ export class ApexReplayDebug extends LoggingDebugSession {
...
+    if (msg?.length !== 0) {
       const event: DebugProtocol.OutputEvent = new OutputEvent(
         `${msg}${EOL}`,
         category
       );
-      event.body.source = sourceFile;
+      if (sourceFile) {
+        event.body.source = sourceFile;
+      }
       event.body.line = sourceLine;
```
daphne-sfdc, line 604: 1. Why did you add `is (sourceFile) {` around `event.body.source = sourceFile`?  What is the harm in setting the `sourceFile` variable when the `printToDebugConsole()` function is called without the `sourceFile` argument and `sourceFile` automatically gets the value of `undefined`?
2. If there is indeed an edge case covered here, why didn't you add `is (sourceLine) {` around `event.body.line = sourceLine`?  Why do `sourceFile` and `sourceLine` need to be treated differently?

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
```diff
@@ -48,7 +48,7 @@ describe('LogEntry event', () => {
     expect(stackFrame.id).to.equal(0);
     expect(stackFrame.name).to.equal('');
     expect(stackFrame.line).to.equal(context.getLogLinePosition() + 1);
-    expect(stackFrame.source.name).to.equal(context.getLogFileName());
-    expect(stackFrame.source.path).to.equal(context.getLogFilePath());
+    expect(stackFrame.source!.name).to.equal(context.getLogFileName());
+    expect(stackFrame.source!.path).to.equal(context.getLogFilePath());
```
daphne-sfdc, line 52: Can you explicitly check that `stackFrame.source` is not undefined instead of just telling Typescript that it is defined?

packages/salesforcedx-apex-replay-debugger/src/adapter/apexReplayDebug.ts
CristiCanizales: After discussion in slack I moved the validation to userDebugState.ts:35

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
CristiCanizales: It can actually be undefined so I changed the test to verify name and path if source is defined

daphne-sfdc (approved):
Approved! 🎉

Thanks Cristi for doing this long-overdue dependency upgrade for our debugger extensions!

QE doc here: https://salesforce.quip.com/UjqpAlOCxYIm ✅

packages/salesforcedx-apex-replay-debugger/src/core/logContext.ts
```diff
@@ -182,7 +182,7 @@ export class LogContext {
     if (
       topFrame &&
       topFrame.source &&
-      topFrame.source.name.toLowerCase().endsWith('.trigger')
+      topFrame.source.name?.toLowerCase().endsWith('.trigger')
```
peternhale, line 185: @CristiCanizales 
the conditional can be shortened to
```
if(topFrame?.source?.name?.toLowerCase().endsWith('.trigger'))
```

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
```diff
@@ -48,7 +48,9 @@ describe('LogEntry event', () => {
     expect(stackFrame.id).to.equal(0);
     expect(stackFrame.name).to.equal('');
     expect(stackFrame.line).to.equal(context.getLogLinePosition() + 1);
-    expect(stackFrame.source.name).to.equal(context.getLogFileName());
-    expect(stackFrame.source.path).to.equal(context.getLogFilePath());
+    if (stackFrame.source) {
```
peternhale, line 51: It seems to me that source should be present and if not the test should fail.

packages/salesforcedx-apex-replay-debugger/test/unit/states/userDebugState.test.ts
```diff
@@ -63,7 +63,8 @@ describe('User debug event', () => {
   it('Should link to anonymous specific frame', () => {
     const frame = {
       name: EXEC_ANON_SIGNATURE,
-      source: { name: 'foo.log' }
+      // eslint-disable-next-line no-global-assign
```
peternhale, line 66: Why is this eslint directive necessary here and not on line 91?

peternhale (changes requested):

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
CristiCanizales: Source is an optional parameter though

packages/salesforcedx-apex-replay-debugger/test/unit/states/userDebugState.test.ts
CristiCanizales: good catch. I forgot to remove it

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
peternhale: Yes, but the original test expects it to be there (I know assuming that it is present). The original test would fail with error "cannot access property name from undefined", if it were undefined. The change eliminates the test case. 

If you remove the guard and change expect to
```
expect(stackFrame.source?.name).to.equal(context.getLogFileName());
```
The test will pass as expected when it is not defined and will fail if it is undefined, which matches the "intent" (even if unintentional) of the test.

peternhale (changes requested):

packages/salesforcedx-apex-replay-debugger/test/unit/states/logEntryState.test.ts
CristiCanizales: Oh, that makes sense. Thanks Pete!

peternhale (approved):

PR #5519, opened 2024-03-22
feat: change status bar color when in production org setting

### LINKED ISSUE #5517: Add setting to change the StatusBar color when working in a production org
**Is your feature request related to a problem? Please describe.**

When developing in VS Code, we often switch between the dev org and the production org to retrieve and compare files.
It sometime happens that we push code by mistake to the wrong org. It's okay with Apex classes as it launch the test classes automatically in production, but with Visualforce components or LWC it uploads it right away.

**Describe the solution you'd like**

A setting that changes the StatusBar color if the default org is the production one (Checkbox).
Another setting to choose what color is used (Hexadecimal input) 

**Describe alternatives you've considered**

I've considered changing the ActivityBar into another color, or the StatusBarItem that displays the current org, but I think it would be too big and too subtle respectively.

**Additional context**

The idea comes from this extension: https://chromewebstore.google.com/detail/org-header-for-salesforce/bhnfbnhpmnfacebccnjjoeecgmnbhjam

I already implemented it to test if it could be done, I came up with this:

https://github.com/forcedotcom/salesforcedx-vscode/assets/37369168/9313f325-6b4b-4221-a212-7c129d6c11e1

Let me know if you would be interested by.a PR!

PierreVanobbergen:
fix #5517

### What does this PR do?
Create a setting to allow the status bar to change color when the default org is a production one

Check #5517 for details

### DIFF
--- a/packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
+++ b/packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as vscode from 'vscode';
+import { WorkspaceContext } from '../context';
+import { getDefaultUsernameOrAlias } from '../context/workspaceOrgType';
+import { OrgAuthInfo } from '../util';
+import { SfdxCoreSettings } from './sfdxCoreSettings';
+
+/**
+ * Change the color of the status bar when the default org is a production org
+ * @returns {Promise<boolean>} - returns true if the color was changed
+ */
+export const colorWhenProductionOrg = async () => {
+  const baseColorStatusBar = new vscode.ThemeColor('statusBar.background');
+
+  const colorWHenProductionOrgHandler = async () => {
+    const usernameOrAlias = await getDefaultUsernameOrAlias();
+    const settings = SfdxCoreSettings.getInstance();
+
+    const activated = settings.getColorWarningWhenProductionOrg();
+    const colorForProdOrg = settings.getColorWarningWhenProductionOrgColor();
+
+    if (!usernameOrAlias || !activated) {
+      return false;
+    }
+    const isProdOrg = await OrgAuthInfo.isAProductionOrg(
+      await OrgAuthInfo.getUsername(usernameOrAlias)
+    );
+    const colorCustomizations = {
+      'statusBar.background': isProdOrg ? colorForProdOrg : baseColorStatusBar
+    };
+
+    // Save the configuration to the global settings file
+    await vscode.workspace
+      .getConfiguration()
+      .update(
+        'workbench.colorCustomizations',
+        colorCustomizations,
+        vscode.ConfigurationTarget.Global
+      );
+    return true;
+  };
+
+  WorkspaceContext.getInstance().onOrgChange(() =>
+    colorWHenProductionOrgHandler()
+  );
+
+  /**
+   * Change the color of the status bar when the window state changes, avoiding the status bar color on others vscode windows
+   */
+  vscode.window.onDidChangeWindowState(async e => {
+    if (e.focused) {
+      await colorWHenProductionOrgHandler();
+    } else {
+      await vscode.workspace
+        .getConfiguration()
+        .update(
+          'workbench.colorCustomizations',
+          {},
+          vscode.ConfigurationTarget.Global
+        );
+    }
+  });
+  return await colorWHenProductionOrgHandler();
+};
--- a/packages/salesforcedx-vscode-core/src/util/authInfo.ts
+++ b/packages/salesforcedx-vscode-core/src/util/authInfo.ts
@@ -18,9 +18,8 @@ import { telemetryService } from '../telemetry';
 
 export class OrgAuthInfo {
   public static async getDevHubUsername() {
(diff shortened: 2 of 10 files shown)


PierreVanobbergen:
Some notes about the feature/PR:

- The code will change the User settings and not the Workspace setting, as it would modify the file under .vscode/settings.json that could be tracked in version control
- The Japanese translation was made using DeepL, I'm not fluent at all, so if it's important, it needs to be checked.
- I would like some feedback about the `OrgAuthInfo.isAProductionOrg`to see if it's the correct way to deduce if an org is a production one

Thank you :)

CristiCanizales:
Hi @PierreVanobbergen, thank you for this contribution! I just saw the video you shared in the issue, and the comment I have is that the color should apply to only the org picker (that element where the org alias is displayed), because we have other items in the bottom bar that have colors according to the state and we would miss that with how this is working now. Do you think you could limit the feature to just the org picker? Thanks!

packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
```diff
@@ -0,0 +1,70 @@
...
+import { SfdxCoreSettings } from './sfdxCoreSettings';
+
+/**
+ * Change the color of the status bar when the default org is a production org
+ * @returns {Promise<boolean>} - returns true if the color was changed
+ */
+export const colorWhenProductionOrg = async () => {
+  const baseColorStatusBar = new vscode.ThemeColor('statusBar.background');
+
+  const colorWHenProductionOrgHandler = async () => {
```
peternhale, line 21: typo colorWHenProductionOrgHandler -> colorWhenProductionOrgHandler

packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
```diff
@@ -0,0 +1,70 @@
...
+      .update(
+        'workbench.colorCustomizations',
+        colorCustomizations,
+        vscode.ConfigurationTarget.Global
+      );
+    return true;
+  };
+
+  WorkspaceContext.getInstance().onOrgChange(() =>
+    colorWHenProductionOrgHandler()
```
peternhale, line 50: fix typo

packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
```diff
@@ -0,0 +1,70 @@
...
+  );
+
+  /**
+   * Change the color of the status bar when the window state changes, avoiding the status bar color on others vscode windows
+   */
+  vscode.window.onDidChangeWindowState(async e => {
+    if (e.focused) {
+      await colorWHenProductionOrgHandler();
```
peternhale, line 58: And here as well

packages/salesforcedx-vscode-core/src/settings/colorWarningWhenProdOrg.ts
```diff
@@ -0,0 +1,70 @@
...
+      await vscode.workspace
+        .getConfiguration()
+        .update(
+          'workbench.colorCustomizations',
+          {},
+          vscode.ConfigurationTarget.Global
+        );
+    }
+  });
+  return await colorWHenProductionOrgHandler();
```
peternhale, line 69: And one more here

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -114,6 +114,15 @@ export class OrgAuthInfo {
     );
   }
 
+  public static async isAProductionOrg(username: string): Promise<boolean> {
```
peternhale, line 117: Please review https://forcedotcom.github.io/sfdx-core/classes/org_org.Org-1.html to determine this.

peternhale (changes requested):

PierreVanobbergen:
> Hi @PierreVanobbergen, thank you for this contribution! I just saw the video you shared in the issue, and the comment I have is that the color should apply to only the org picker (that element where the org alias is displayed), because we have other items in the bottom bar that have colors according to the state and we would miss that with how this is working now. Do you think you could limit the feature to just the org picker? Thanks!

I thought about just changing the color of the org picker as well, but I feel that it might not be enough of an eye catcher 👀 .
I would like to keep the status bar color change, but maybe having the setting to be where the change happens might be the way. Like this:
image

We might even add more element later in the picklist.

What do you think about it @CristiCanizales ?

packages/salesforcedx-vscode-core/src/util/authInfo.ts
randi274: @peternhale is there an affirmative case for "this is a production org"? I'm not seeing that in these docs, so it's looking like Production would be the fallback case.

peternhale: @randi274 Yes. production should be the fallback.

randi274: Got it! @interfaceconjurer and I were chatting about this detail today, it might be worth y'all discussing the possibility of "Production" being overly inclusive.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -114,6 +114,15 @@ export class OrgAuthInfo {
...
+    const authInfo = await AuthInfo.create({ username });
+    const authInfoFields = authInfo.getFields();
+    return Promise.resolve(
+      !authInfoFields.isSandbox &&
+        !authInfoFields.instanceUrl?.includes('sandbox.my.salesforce.com')
```
AndrewStopchenko-SO, line 122: Should this also exclude scratch orgs (`!authInfoFields.isScratch`)?

PR #5522, opened 2024-03-28
fix: ISV Debugger no longer throws project configuration error when 'Launch Apex Debugger' button is clicked

daphne-sfdc:
### What does this PR do?
Fixes an issue where the ISV Debugger fails to launch when the `Launch Apex Debugger` button is clicked.

@W-14182113@

### Functionality Before
This error message pops up when the `Launch Apex Debugger` button is clicked, even though the project is configured correctly.
Screenshot 2024-03-28 at 11 48 06 AM (2)

### Functionality After
ISV Debugger is able to start when the `Launch Apex Debugger` button is clicked, as expected.
Screenshot 2024-03-28 at 1 32 53 PM (2)


diyer:
yay 🎉 Nice job @daphne-sfdc !!

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -56,7 +56,7 @@
     "vscode-uri": "1.0.1"
   },
   "scripts": {
-    "bundle:debugger": "esbuild ./src/adapter/apexDebug.ts  --bundle --outfile=dist/apexdebug.js --format=cjs --platform=node --minify",
+    "bundle:debugger": "esbuild ./src/adapter/apexDebug.ts  --bundle --outfile=dist/apexdebug.js --format=cjs --platform=node --external:vscode --minify",
```
peternhale, line 59: Do any of our other modules externalize vscode?

daphne-sfdc, line 59: I added the `--external:vscode` tag here because my Commit Workflows were failing with the following error after I added calls to telemetry:
Screenshot 2024-03-29 at 11 13 10 AM

I did a search on this error message and found https://github.com/evanw/esbuild/issues/103, which says to make the `vscode` module an external module for esbuild.  I found that the other `package.json` files in the repo used this tag to make `vscode` an external module, so I added the tag to this file, and the Commit Workflow passes now.

peternhale (changes requested):

packages/salesforcedx-apex-debugger/package.json
daphne-sfdc: A lot of them 🙂
Screenshot 2024-03-29 at 11 24 01 AM

peternhale (approved):

peternhale (changes requested):

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -240,14 +248,47 @@ const notifyDebuggerSessionFileChanged = (): void => {
...
+  const customEventHandler = vscode.debug.onDidReceiveDebugSessionCustomEvent(
+    async event => {
+      if (event && event.session) {
+        const type = await getDebuggerType(event.session);
+        if (type !== DEBUGGER_TYPE) {
+          return;
+        }
+
+        if (event.event === SEND_METRIC_LAUNCH_EVENT && event.body) {
+          const metricLaunchArgs = event.body as MetricLaunch;
```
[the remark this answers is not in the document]
daphne-sfdc: I added a new type guard and replaced the type casting with a call to the type guard.  (See `isMetric()` in `packages/salesforcedx-apex-debugger/src/index.ts`.)

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -5,6 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import {
+  isMetric
+} from '@salesforce/salesforcedx-apex-debugger/out/src';
```
CristiCanizales, line 10: Can you combine these duplicated imports?

CristiCanizales (changes requested):
Looks good to me! Just one small request

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,6 +760,33 @@ export class ApexDebug extends LoggingDebugSession {
       }
     } catch (error) {
       this.tryToParseSfError(response, error);
+      // telemetry for expired session or invalid org-isv-debugger-sid (authentication error)
```
daphne-sfdc, line 763: This block covers two cases that display the same popup and error message:
1. The `forceIde://` URL of the current ISV Debugger session has expired (pretty common use case, as the `forceIde://` URL expires after 5 hours of use and after the user has been idle for too long)
2. The `org-isv-debugger-sid` config variable is set to an invalid value (edge case, as the config variable should be automatically set to the correct value during the bootstrap process)

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,6 +760,33 @@ export class ApexDebug extends LoggingDebugSession {
...
+      if (error === undefined) {
+        this.sendEvent(
+          new Event(SEND_METRIC_EVENT, {
+            subject: nls.localize('isv_debugger_session_authentication_invalid'),
+            type: 'startIsvDebuggerAuthenticationInvalid'
+          } as Metric)
+        );
+      }
+      // telemetry for invalid org-isv-debugger-url
```
daphne-sfdc, line 772: This block covers another edge case where the `org-isv-debugger-sid` config variable is set to an invalid value.  This is also a value that is supposed to be set correctly during the bootstrap process.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,6 +760,33 @@ export class ApexDebug extends LoggingDebugSession {
...
       this.tryToParseSfError(response, error);
+      // telemetry for expired session or invalid org-isv-debugger-sid (authentication error)
+      if (error === undefined) {
+        this.sendEvent(
+          new Event(SEND_METRIC_EVENT, {
+            subject: nls.localize('isv_debugger_session_authentication_invalid'),
+            type: 'startIsvDebuggerAuthenticationInvalid'
+          } as Metric)
+        );
+      }
```
daphne-sfdc, line 771: This block covers two cases that display the same popup and error message:
1. The `forceIde://` URL of the current ISV Debugger session has expired (pretty common use case, as the `forceIde://` URL expires after 5 hours of use and after the user has been idle for too long)
2. The `org-isv-debugger-sid` config variable is set to an invalid value (edge case, as the config variable should be automatically set to the correct value during the bootstrap process)

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -727,6 +760,33 @@ export class ApexDebug extends LoggingDebugSession {
...
+      // telemetry for invalid org-isv-debugger-url
+      else if (String(error) === "TypeError: Cannot read properties of undefined (reading 'pathname')") {
+        this.sendEvent(
+          new Event(SEND_METRIC_EVENT, {
+            subject: nls.localize('org_isv_debugger_url_invalid'),
+            type: 'startIsvDebuggerOrgIsvDebuggerUrlInvalid'
+          } as Metric)
+        );
+      }
```
daphne-sfdc, line 780: This block covers another edge case where the `org-isv-debugger-sid` config variable is set to an invalid value.  This is also a value that is supposed to be set correctly during the bootstrap process.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -688,6 +696,13 @@ export class ApexDebug extends LoggingDebugSession {
           typeof isvDebuggerUrl === 'undefined'
         ) {
           response.message = nls.localize('invalid_isv_project_config');
+          // telemetry for the case where the org-isv-debugger-sid and/or org-isv-debugger-url config variable is not set
+          this.sendEvent(
+            new Event(SEND_METRIC_EVENT, {
+              subject: nls.localize('invalid_isv_project_config'),
+              type: 'startIsvDebuggerConfigError'
+            } as Metric)
```
peternhale, line 704: The "as" assertions are unnecessary, please remove them in all locations with this same usage pattern

packages/salesforcedx-apex-debugger/src/index.ts
```diff
@@ -26,3 +26,19 @@ export interface WorkspaceSettings {
...
+// Type guard to check if the object conforms to Metric
+export const isMetric = (input: unknown): input is Metric => (
+  input !== null &&
+  typeof input === 'object' &&
+  Object.keys(input).length === 2 &&
+  'subject' in input &&
+  typeof input.subject === 'string' &&
+  'type' in input &&
+  typeof input.type === 'string'
+);
```
peternhale, line 44:
```suggestion
export const isMetric = (input: unknown): input is Metric => (
  !!input &&
  Object.keys(input).every(key => ['subject', 'type'].includes(key)) &&
  Object.values(input).every(value => typeof value === 'string')
);
```

packages/salesforcedx-vscode-apex-debugger/src/index.ts
```diff
@@ -240,14 +242,38 @@ const notifyDebuggerSessionFileChanged = (): void => {
...
+const registerDebugHandlers = (): vscode.Disposable => {
+  const customEventHandler = vscode.debug.onDidReceiveDebugSessionCustomEvent(
+    async event => {
+      if (event?.session) {
+        const type = await getDebuggerType(event.session);
+        if (type !== DEBUGGER_TYPE) {
+          return;
+        }
+
+        if (event.event === SEND_METRIC_EVENT && isMetric(event.body)) {
```
peternhale, line 254: Seems that event.event already establishes the event as a metric event and isMetric verifies the body. Feels redundant. How can this be changed to reduce this to a single check?

packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
```diff
@@ -65,4 +65,15 @@ export class TelemetryService {
     const hrend = process.hrtime(hrstart);
     return Number(util.format('%d%d', hrend[0], hrend[1] / 1000000));
   }
+
+  public sendEvent(msg: string, type: string): void {
```
peternhale, line 69: This function was added to specifically handle a metric event, why not embrace that. The other methods in this class are specializations as well that are handling a very specific need. Seems reasonable to apply the same here.

peternhale (changes requested):

packages/salesforcedx-apex-debugger/src/index.ts
daphne-sfdc: Wow, this is very consolidated compared to what was there before!  I really would not have thought this is how to do the improvement!  Thank you @peternhale!

packages/salesforcedx-vscode-apex-debugger/src/telemetry/telemetry.ts
daphne-sfdc: Thanks for the suggestion.  I changed the function name from `sendEvent()` to `sendMetricEvent()` to be more specific about what it does.

packages/salesforcedx-vscode-apex-debugger/src/index.ts
daphne-sfdc: I am currently debating how to either consolidate those checks into one or eliminate the isMetric type guard. I understand that the first check for SEND_METRIC_EVENT is necessary because that is the specific event we intentionally pass from the debugger adapter to the extension for telemetry.  There might be another event type that transmits a body with a structure identical to that of SEND_METRIC_EVENT. For example, in my test, I introduced a new event type named SEND_TRIAL_EVENT, which has the same body structure as SEND_METRIC_EVENT but sends a message to telemetry of the type launchApexDebuggerTrial instead. The screenshot shows that when the first check is removed, AppInsights records both launchApexDebuggerTrial and launchApexDebugger when the ISV Debugger is initiated.  (Code changes made in the "[chore: try SEND_TRIAL_EVENT](https://github.com/forcedotcom/salesforcedx-vscode/pull/5522/commits/6aec25e9f7206ddf6d91b111be812021095fd082)" commit and then reverted.)

Screenshot 2024-04-05 at 10 39 55 PM
_______________________________________________

When we invoke sendEvent() in apexDebug.ts, we control the event body that we pass to the event handler. We set the event body object to contain only subject and type attributes, ensuring that the body aligns with the Metric interface's structure. The isMetric() type guard is employed to verify that there are no missing or unexpected attributes in the body sent from the debugger adapter when sendEvent() is called. However, the only objects filtered by the type guard are the ones we create in apexDebug.ts.

I am hoping to hear your suggestion on how we can best handle the checks.

CristiCanizales (commented):
Looks good to me! QE notes [here](https://salesforce.quip.com/pXOuAeaklgP9).

I'd just delete all the tests that are commented out and then good to go!

Awesome work on this Daphne, thanks!
:shipit:

CristiCanizales (approved):

PR #5540, opened 2024-04-16
fix: prevent negative extension activations times

peternhale:
@W-15328509@
Prevent negative extension activation time from being included in telemetry data

Modified the calculations and logic that sends the extended activation data to telemetry to:
- Only calculate activation time when load start date is present and activate end date >= load start date
- Only send extended activation data if activation time > -1

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/helpers/activationTracker.ts
+++ b/packages/salesforcedx-utils-vscode/src/helpers/activationTracker.ts
@@ -60,7 +60,7 @@ export class ActivationTracker {
     let extensionActivationTime = -1;
     if (
       extensionInfo?.loadStartDate &&
-      activateEndDate.getTime() <= extensionInfo.loadStartDate.getTime()
+      activateEndDate.getTime() >= extensionInfo.loadStartDate.getTime()
     ) {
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-utils-vscode/src/helpers/activationTracker.ts
```diff
@@ -50,24 +50,36 @@ export class ActivationTracker {
...
-      ...extensionInfo,
-      extensionActivationTime,
-      activateEndDate,
-      markEndTime: hrEnd
-    };
-    this.telemetryService.sendActivationEventInfo(this.activationInfo);
+    let extensionActivationTime = -1;
+    if (
+      extensionInfo?.loadStartDate &&
+      activateEndDate.getTime() <= extensionInfo.loadStartDate.getTime()
```
mingxuanzhangsfdx, line 63: Why activateEndDate should be no more than loadStartDate? In my test when loadStartDate is defined, activateEndDate is always larger than loadStartDate, so the extensionActivationTime is a positive value, which makes sense.

peternhale: I have it backward, good catch

Issue #5534, opened 2024-04-11
Code coverage highlighting no longer displaying and returns error message on Apex v60.8.0


zfrench:
### Summary

Code coverage highlighting no longer displays and a pop error occurs

### Steps To Reproduce:

1.  Enable 'Highlight Apex Code Coverage'
1.  Select a unit test to run from a test class
1.  Click on the on class that you ran the code coverage for and an error message will be displayed.

### Expected result

Expected the code highlighting to display green or red lines depending on the code that ran.

### Actual result

No code highlighting occurred and error message was displayed.

### Additional information
Error Message:
Expected double-quoted property name in JSON at position 30195 (line 1 column 30196)

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/4147537/04c7ec7d-74ff-4955-8cf6-10e4b8416be6)

If I downgrade the extension from 60.8.0 to 60.7.0 then it will display the highlighted coverage correctly.

**Salesforce Extension Version in VS Code**: Apex v60.8.0

**SFDX CLI Version**: 2.36.8

**OS and version**: Windows 10 Enterprise

**VS Code version**: 1.88.0

alan-liu-ofx:
Same for me.
![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/62679825/dd0f46cf-ff09-4b29-b077-03b0873619b8)
@salesforce/cli/2.36.8 win32-x64 node-v20.11.1
VS Code 1.88.1 
Win 10 22H2

maximojoinansel:
Hi there! I am still getting this issue. I saw there is a work to fix it in progress but in the meantime Is there any workaround available to make this work? I really appreciate any help you can provide.

peternhale:
@maximojoinansel the only workaround would be to downgrade extensions to v60.7.0.

The work is in progress and we hope to have the fix out in our release next week.

jh480:
> Hi there! I am still getting this issue. I saw there is a work to fix it in progress but in the meantime Is there any workaround available to make this work? I really appreciate any help you can provide.

You can still get the first few uncovered lines from the Output section under "Apex Code Coverage by Class".

maximojoinansel:
@peternhale downgrading the extensions to v60.7.0. works! Thanks for your help here.

AhmedElAmory:
I thought I was the only one, thanks for reporting this. I'm waiting for the fix.

rvanaken:
[...]

PR #5543, opened 2024-04-18
fix: bump apex-node for dangling comma fix

peternhale:
pick up latest v4 version of apex-node to correct test result parsing error when no code coverage is in test results.

#5534 , @W-15530915@

### Functionality Before
When using the highlight code coverage feature and running apex tests without code coverage, extensions were displaying an message stating coverage results parsing failed. This is due to a bug introduced in apex-node streaming changes in v60.8.0.

### Functionality After
The fix to apex-node corrected the JSON results used for highlighting code coverage, when there is not code coverage present, no produces the correct warning that no code coverage is present in the latest test run for the apex class opened in the editor.


daphne-sfdc (approved):
Approved! 🎉 

Confirmed that there are no dangling commas in **test-result.json** files both when code coverage setting is checked and unchecked, and code coverage highlighting is present.  Works on both Mac and Windows. ✅ 

QE notes at the bottom of this doc: https://salesforce.quip.com/DXmnAhMNNOBZ

Issue #4294, opened 2022-07-19
When right-clicking to create an Apex Class, the new class should automatically go into the selected folder


randi274:
**Is your feature request related to a problem? Please describe.**
We recently fixed #3572 to allow creating an apex class in a subdirectory, but our users are still prompted to provide a directory when using the command! The prompt menu doesn't make much sense under the nested structure. 

**Describe the solution you'd like**

- If I right-click the `classes` directory to create an Apex class, it shouldn't prompt for the directory, and should automatically create the class there.
- If I right-click a subdirectory of `classes`, it shouldn't prompt for directory either, and should create the class in the correct subdirectory.
- It should work on all types of machines, as always 😄 
- If I use the Command Palette to create an Apex class, it should prompt for the directory. 

**Describe alternatives you've considered**
It works now, it's just a little clunky since I've already selected a directory with a right-click action.

kaustubhdapurkar:
Hi @randi274 , could you please take a look at this draft.
I already see multiple issues with the new code, but created this PR to get started on the enhancement -

1. I have created an new class that implements ParametersGatherer which could be avoided. I did this because changing the existing `SelectOutputDir` class had bigger impact.
2. I have removed the `initialized `variable, will need to incorporate it with my changes, but I don't know how
3. I am not really sure about the unused parameter `uris`, but this doesn't seem to work if I don't pass it. I couldn't find any supporting documentation on this.
4. Since the command only shows up when selecting the singular classes folder, I have not written this code in a way that handles when `sourceUri` is an array. This could be problematic.

There could be more issues that I am not seeing. Not very familiar with TS and vscode extension API.
Thanks in advance for your time 😊.

PR #4901, opened 2023-06-01
feat: create apex class in classes directory when right clicking to create class

kaustubhdapurkar:
feat: create apex class in classes directory when right clicking to create class

This is just draft PR and I am pretty sure my implementation is not the best. Would love to get any feedback on this. I strongly  believe that the code I have written resembles spaghetti code🤣.

#4294 

### What does this PR do?
When creating an apex class by right clicking on the classes directory, it will create the class directly in this classes directory without prompting the user for a file path.

### What issues does this PR fix or reference?
#4294 

### Functionality Before
When creating apex class by right clicking the classes directory, the extension still asks for a directory.

### Functionality After
When creating apex class by right clicking the classes directory, class is created automatically in the directory.

PR #5544, opened 2024-04-18
feat: do not prompt for output dir when called from context menu

aheber:
### What does this PR do?
Make the Apex Class and Apex Test Class context options take the output directory from context instead of prompting the user. 

#4294  

### Functionality Before
User is prompted for output directory
VSCode-4294-CreateApex-Before

### Functionality After
Output directory is taken from context
VSCode-4294-CreateApex-After

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateClass.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateClass.ts
@@ -9,13 +9,15 @@ import {
   LocalComponent,
   ParametersGatherer
 } from '@salesforce/salesforcedx-utils-vscode';
+import * as vscode from 'vscode';
 import {
   CompositeParametersGatherer,
   MetadataTypeGatherer,
   SelectFileName,
   SelectOutputDir,
   SfCommandlet,
-  SfWorkspaceChecker
+  SfWorkspaceChecker,
+  SimpleGatherer
 } from '../util';
 import { OverwriteComponentPrompt } from '../util/overwriteComponentPrompt';
 import { ApexTestTemplateGatherer } from '../util/parameterGatherers';
@@ -27,20 +29,20 @@ import {
   APEX_TEST_TEMPLATE
 } from './metadataTypeConstants';
 
-let initialized = false;
-let fileNameGatherer: ParametersGatherer<any>;
-let outputDirGatherer: ParametersGatherer<any>;
-let metadataTypeGatherer: ParametersGatherer<any>;
-let templateGatherer: ParametersGatherer<any>;
+let fileNameGatherer: ParametersGatherer<any> | undefined;
+let outputDirGatherer: ParametersGatherer<any> | undefined;
+let metadataTypeGatherer: ParametersGatherer<any> | undefined;
+let templateGatherer: ParametersGatherer<any> | undefined;
 
 export const getParamGatherers = () => {
-  if (!initialized) {
-    fileNameGatherer = new SelectFileName(APEX_CLASS_NAME_MAX_LENGTH);
-    outputDirGatherer = new SelectOutputDir(APEX_CLASS_DIRECTORY);
-    metadataTypeGatherer = new MetadataTypeGatherer(APEX_CLASS_TYPE);
-    templateGatherer = new ApexTestTemplateGatherer(APEX_TEST_TEMPLATE);
-    initialized = true;
-  }
+  fileNameGatherer =
+    fileNameGatherer ?? new SelectFileName(APEX_CLASS_NAME_MAX_LENGTH);
+  outputDirGatherer =
+    outputDirGatherer ?? new SelectOutputDir(APEX_CLASS_DIRECTORY);
+  metadataTypeGatherer =
+    metadataTypeGatherer ?? new MetadataTypeGatherer(APEX_CLASS_TYPE);
+  templateGatherer =
+    templateGatherer ?? new ApexTestTemplateGatherer(APEX_TEST_TEMPLATE);
   return {
     fileNameGatherer,
     outputDirGatherer,
@@ -49,8 +51,17 @@ export const getParamGatherers = () => {
   };
 };
 
-export const apexGenerateClass = async () => {
+// if called from a file's context menu, will deliver the clicked file URI,
+// ignoring an additional arg that is array of selected
+// if called from the command pallet args will be empty
+export const apexGenerateClass = async (sourceUri?: vscode.Uri) => {
   const gatherers = getParamGatherers();
+
+  if (sourceUri) {
+    gatherers.outputDirGatherer = new SimpleGatherer<{ outputdir: string }>({
+      outputdir: sourceUri.fsPath
+    });
+  }
 
   const createTemplateExecutor = new LibraryApexGenerateClassExecutor();
   const commandlet = new SfCommandlet(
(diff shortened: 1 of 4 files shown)


aheber:
This also changes the tests so they are correct in isolation as well as when run as a group. Had to leak implementation details a bit to make it all work. We can remove or change as needed.

There is risk here, I changed the order of the parameters on apexGenerateUnitTestClass to match the parameter order when invoked from a Context menu. As far as I can tell this exists for testing but if this is used elsewhere then we may need to undo that decision.

I also tested this from the Command Pallet at the behaviors are not modified, output directory is still prompted.

aheber:
@AnanyaJha @randi274 I think this would replace #4901 if accepted.

aheber:
If we're happy with this implementation I'm happy to look at the other context menu options like creating LWCs for a follow up PR.

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateClass.ts
```diff
@@ -49,9 +51,18 @@ export const getParamGatherers = () => {
...
 };
 
-export const apexGenerateClass = async () => {
+// if called from a file's context menu, will deliver the clicked file URI,
+// ignoring an additional arg that is array of selected
+// if called from the command pallet args will be empty
+export const apexGenerateClass = async (sourceUri?: vscode.Uri) => {
   const gatherers = getParamGatherers();
 
+  if (sourceUri) {
```
aheber, line 60: This is the only necessary change, this block alone enabled the feature. Everything else is actually to improve testing and typings.

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateUnitTestClass.ts
```diff
@@ -20,13 +21,31 @@ import { getParamGatherers } from './apexGenerateClass';
 import { LibraryApexGenerateUnitTestClassExecutor } from './executors/LibraryApexGenerateUnitTestClassExecutor';
 import { APEX_CLASS_TYPE } from './metadataTypeConstants';
 
+// if called from a file's context menu, will deliver the clicked file URI and an array of selected files
+// if called from the command pallet args will be empty
 export const apexGenerateUnitTestClass = async (
-  unitFileToCreate?: string,
-  unitFileDirectory?: string,
+  outputDirectory?: string | vscode.Uri,
```
aheber, line 27: This is the risky bit. I changed where I found usage but it looks like it is only here for testing. If it is somewhere that I didn't find then we can back this out.

The reason for this is it better matches the arg ordering when invoked from the Context Menu. Ignoring the second parameter's variation but still adding it to typings so we know it is possible.

packages/salesforcedx-vscode-core/test/jest/commands/templates/apexGenerateUnitTestClass.test.ts
```diff
@@ -77,14 +79,14 @@ describe('apexGenerateUnitTestClass Unit Tests.', () => {
...
-  it('Should prompt if the provided params are not valid.', async () => {
+  it('Should not prompt if called with a file URI', async () => {
     // This happens when the command is executed from the context menu in the explorer on the classes folder.
-    const notAString = { path: 'thing' };
-    const notAString2 = { path: 'thing2' };
-    await (apexGenerateUnitTestClass as any)(notAString, notAString2);
-    // Note there is a bad pattern in the exported getParamGatherers method that uses module state to cache the gatherers.
-    // This being the case the selectFileNameMocked check is invalid here when all tests are run.
-    expect(simpleGathererMocked).not.toHaveBeenCalled();
+    const selectedPathUri = {
```
aheber, line 84: Better example of the invocation but cheating on the type casting.

packages/salesforcedx-vscode-core/test/jest/commands/templates/apexGenerateUnitTestClass.test.ts
```diff
@@ -97,9 +99,10 @@ describe('apexGenerateUnitTestClass Unit Tests.', () => {
   it('Should used the passed parameters if provided.', async () => {
     const fileName = 'testFileName';
     const outputDir = 'testOutputDir';
-    await apexGenerateUnitTestClass(fileName, outputDir);
-    expect(selectFileNameMocked).not.toHaveBeenCalled();
-    expect(selectOutputDirMocked).not.toHaveBeenCalled();
+    await apexGenerateUnitTestClass(outputDir, fileName);
```
aheber, line 102: Changed because the test's now reset the module cache. This causes the various gatherers to still be initialized, not necessarily used but still initiated.

Production code is still doing the same thing, this is only a test behavior that better matches reality.

packages/salesforcedx-vscode-core/src/commands/templates/apexGenerateClass.ts
```diff
@@ -27,20 +29,20 @@ import {
...
 export const getParamGatherers = () => {
-  if (!initialized) {
-    fileNameGatherer = new SelectFileName(APEX_CLASS_NAME_MAX_LENGTH);
-    outputDirGatherer = new SelectOutputDir(APEX_CLASS_DIRECTORY);
-    metadataTypeGatherer = new MetadataTypeGatherer(APEX_CLASS_TYPE);
-    templateGatherer = new ApexTestTemplateGatherer(APEX_TEST_TEMPLATE);
-    initialized = true;
-  }
+  fileNameGatherer =
+    fileNameGatherer ?? new SelectFileName(APEX_CLASS_NAME_MAX_LENGTH);
```
peternhale, line 39:
```suggestion
  fileNameGatherer ??= new SelectFileName(APEX_CLASS_NAME_MAX_LENGTH);
```

aheber: Great! I've never used that one before. Thanks.

peternhale (changes requested):
@aheber The logic looks solid and I dug around a bit with different test scenarios, specifically projects with multiple "packageDirectories" in sfdx-project.json.

I made some additional changes and pushed them [here](https://github.com/forcedotcom/salesforcedx-vscode/tree/phale/5544)

The branch also includes some changes to get rid of ParameterGatherer<any>.

aheber:
@peternhale looks great! Appreciate the speedy review and changes.

peternhale:
@aheber FYI, we will need our product manager to review the functional changes to make sure they fit in well with other similar functionality, so the PR may "sit" for a bit before we see movement.

peternhale:
> @aheber The logic looks solid and I dug around a bit with different test scenarios, specifically projects with multiple "packageDirectories" in sfdx-project.json.
> 
> I made some additional changes and pushed them [here](https://github.com/forcedotcom/salesforcedx-vscode/tree/phale/5544)
> 
> The branch also includes some changes to get rid of ParameterGatherer<any>.

@aheber Product and our UX folks like this change so we can proceed. I see the suggestions I made have yet to be applied to your PR. Once those changes are made, I will move this PR forward. Thanks again for the contribution!

aheber:
Great news!

My fault on the suggestions, I misunderstood. I'll take care of it now.

aheber:
@peternhale did I brake it by cherry picking your `[skip-ci]` commit? I seem to remember the CI jobs normally starting but maybe that took intervention.

Hopefully that should do it but if I need to do something else please let me know.

peternhale (approved):

aheber:
Thank you!

Issue #5553, opened 2024-04-24
The Outline window gets broken if there is an abstract method in the class


gpervukhin:
### Summary

The Outline window gets broken if there is an abstract method in the class OR in any of its nested classes.

### Steps To Reproduce:

```
public abstract class MyTestClass
{
   public void method1() {}

   //public abstract void abstractMethod();
}
```

1. Create a new class using the above code;
2. Check that the Outline window shows the name of the class and its method1;
3. Uncomment the abstract method in the line # 5.

### Expected result

I expect that VS Code shows 2 methods in the Outline window now.

### Actual result

The Outline window gets broken and shows this phrase "No symbols found in document 'MyTestClass.cls'".

### Additional information

![image](https://github.com/forcedotcom/salesforcedx-vscode/assets/4255023/4a22ba5a-3a06-49ba-96ed-fca02273ce28)

**Salesforce Extension Version in VS Code**: v60.8.0

**SFDX CLI Version**: @salesforce/cli/2.20.6 win32-x64 node-v20.10.0

**OS and version**: Windows 10 Home 22H2

**VS Code version**: 1.85.0

diyer:
Closing this as the issue has been fixed and released. @gpervukhin please let us know if you see any other issues. Thank you for reporting this!

gpervukhin:
@diyer, it works fine now. Thank you.

PR #5555, opened 2024-04-25
fix: fix handling of abstract methods in outline

peternhale:
When document symbols are requested from the apex LS for an Apex class with abstract methods, it did not take into account an empty method body.

The LS update now handles this condition properly

@W-15593844@, #5553 

Given the Apex class
```
public abstract class my {
    public my() {

    }
    public String s1() {
        return 's1';
    }

    public abstract String s2();
}
```
### Functionality Before

Screenshot 2024-04-25 at 8 02 39 AM

### Functionality After

Screenshot 2024-04-25 at 8 07 47 AM


mingxuanzhangsfdx (approved):
Thanks for the fix! LGTM

PR #5639, opened 2024-06-12
chore: point to the right apex-pmd in Open VSX

mingxuanzhangsfdx:
### What does this PR do?
apex-pmd has different publishers in VS Code Marketplace and Open VSX Registry. Code Builder installs the extensions from Open VSX, so apex-pmd is not able to be installed in Code Builder. The entry for Open VSX is added. By adding the entry, installing from VS Code marketplace is not affected and no errors can be detected. There is also no error to install apex-pmd from Open VSX with the existence of the entry for Marketplace confirmed by CB team.

@W-15936982@

### Functionality Before
apex-pmd is not installed from Open VSX.

### Functionality After
apex-pmd is installed from Open VSX

### DIFF
--- a/packages/salesforcedx-vscode-expanded/package.json
+++ b/packages/salesforcedx-vscode-expanded/package.json
@@ -43,6 +43,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "chuckjonas.apex-pmd",
+    "pmd.apex-pmd",
     "financialforce.lana"
   ]
 }


diyer:
@mingxuanzhangsfdx curious to know how was this tested n Vs Code without publishing the pack

packages/salesforcedx-vscode-expanded/package.json
```diff
@@ -43,6 +43,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "chuckjonas.apex-pmd",
+    "pmd.apex-pmd",
```
peternhale, line 46: @mingxuanzhangsfdx is the assumption here that an "unknown" extension will fail silently regardless of registry

peternhale (approved):

diyer:
I think this is a bug open vsx. package.json seems to be the source of truth for creating vsixes. And if the same extension has different publishers in marketplace and open vsix, then how do we specify that in a single package.json

packages/salesforcedx-vscode-expanded/package.json
diyer: @mingxuanzhangsfdx nit: is it worth adding a comment here why there are duplicate entries for future readers?

PR #5641, opened 2024-06-13
fix: externalize bfj dependency to allow bundling

daphne-sfdc:
### What does this PR do?
Externalize the `bfj` dependency used in @salesforce/apex-node so that @salesforce/apex-node can be bundled without issues.

@W-15960945@

### Functionality Before
Apex extension does not activate and produces and error message.

### Functionality After
Apex extension works as expected.
https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9501952453


CristiCanizales (approved):
- [E2E Test run passing](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9501952453)✅
- Confirmed all extensions activate correctly locally✅

PR #5643, opened 2024-06-13
chore: needs approval before build beta

mingxuanzhangsfdx:
### What does this PR do?
Allow an async manual check before building extensions in the beta release branch. 

@W-16002826@

### Functionality Before
The building process is completely automated, which does not allow manual check. This might cause errors due to incompatible versions of dependencies.

### Functionality After
Developer can do updates manually if necessary before building beta extensions.


.github/workflows/createAndTestBetaReleaseBranch.yml
```diff
@@ -52,11 +52,19 @@ jobs:
       - id: branch
         run: echo "branch=release/v${{ steps.version.outputs.version }}" >> $GITHUB_OUTPUT
 
+  confirm_build:
+    environment: publish
+    runs-on: ubuntu-latest
+    needs: create_branch
+    name: 'Confirm Beta Release Branch Release'
+    steps:
+      - run: echo "Please confirm the beta release branch working"
```
diyer, line 61: how about updating the message to say something like - this is your change to update versions for any dependencies if required ?

diyer (approved):

.github/workflows/createAndTestBetaReleaseBranch.yml
```diff
@@ -52,11 +52,19 @@ jobs:
...
+
   run_build_and_test:
     if: ${{ needs.create_branch.result == 'success' }} # Only run this if the previous job is successful
     uses: ./.github/workflows/buildAndTest.yml
     name: 'Run Build and Unit Test'
-    needs: create_branch
+    needs: [create_branch, confirm_build]
```
diyer, line 67: nit: confirm_build is good enough here I believe as confirm_build needs create_branch

mingxuanzhangsfdx: The next line needs the output values from `create_branch`

PR #5650, opened 2024-06-19
fix: bump sdr to resolve the issue of deploying MD files

mingxuanzhangsfdx:
### What does this PR do?
Bump sdr and related dependencies to fix the issue that the action of deploying MD files forever deploys the first-time snapshot. 

@W-15952068@

### Functionality Before
The action of deploying MD files forever deploys the first-time snapshot. 

### Functionality After
It will deploy MD with the correct content.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5905,89 +5905,30 @@
       }
     },
     "node_modules/@oclif/core": {
-      "version": "3.27.0",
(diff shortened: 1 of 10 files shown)


diyer:
This is great @mingxuanzhangsfdx !!. However we should have test coverage with this change, either in SDR or our end to avoid having tech debt and not have this issue pop up again in future

peternhale (approved):

PR #5656, opened 2024-06-21
feat: opt in bundling @salesforce/core

mingxuanzhangsfdx:
### What does this PR do?
This PR replaces `@salesforce/core`, `@salesforce/source-deploy-retrieve`, `@salesforce/source-tracking` and `@salesforce/apex-node` with the bundle-able modules in all extensions and bundles the dependency `@salesforce/core-bundle`, `@salesforce/source-deploy-retrieve-bundle`, `@salesforce/source-tracking-bundle` that were previously externalized in the extensions. 

@W-15635320@

### Functionality Before
core, sdr, source-tracking are externalized during the esbuild bundle. The size of the extension pack is up to 100MB now.

### Functionality After
After opting in the dependencies with bundle-able versions and bundle the dependencies, the size of the extension pack is reduced to 61.7MB and a significant improvement in extension startup performance is expected.

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -37,7 +37,7 @@
     "sanitize-filename": "^1.6.1",
     "shelljs": "0.8.5",
     "@salesforce/core-bundle": "7.5.0",
-    "@salesforce/templates-bundle": "61.0.1",
+    "@salesforce/templates-bundle": "^61.0.1",
     "@salesforce/source-deploy-retrieve-bundle": "11.6.10"
   },
   "devDependencies": {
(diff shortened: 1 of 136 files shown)


packages/salesforcedx-vscode-core/package.json
```diff
@@ -25,20 +25,20 @@
...
     "@salesforce/vscode-service-provider": "1.0.4",
     "adm-zip": "0.5.10",
     "applicationinsights": "1.0.7",
     "glob": "^7.1.2",
     "rxjs": "^5.4.1",
     "sanitize-filename": "^1.6.1",
-    "shelljs": "0.8.5"
+    "shelljs": "0.8.5",
+    "@salesforce/core-bundle": "7.5.0",
+    "@salesforce/templates-bundle": "61.0.1",
```
mingxuanzhangsfdx, line 40: This should be with a caret

peternhale (approved):

PR #5674, opened 2024-07-03
fix: link lsp with node_modules in root folder

mingxuanzhangsfdx:
### What does this PR do?
the previous shellscript is not able to link the modules in node_modules in root directory, and makes the link not effect.  this PR is to update the script to link all lsp modules to the local one.

@W-16085908@

### Functionality Before
cannot link

### Functionality After
link all lsp modules under vsce to the local lsp project.


daphne-sfdc (approved):
Approved! This is the same command for **link-lsp** as the one I made for linking lightning-language-server and salesforcedx-vscode! Thanks for also updating the **unlink-lsp** script 😀

PR #5689, opened 2024-07-17
Unique user_id for appInsights reporter

RitamAgrawal:
### What does this PR do?
The PR adds logic to retrieve userId from extensionContext global state and use the same in Telemetry Service, where it calls `AppInsights()` to report usage data in App Insight.

### Functionality Before

- Earlier we were using `env.machineId` as `ai.user.id`, which is userId reported in appInsights.

### Functionality After
- Adds an additional parameter, userId to AppInsights constructor.

- Now we use telemetryUserId defined in globalState of extensionContext to report userId in appInsight
- If telemetryUserId is already defined in global state, it uses the same value for userId in appInsights.
- If not, we call `sf telemetry --json` at activation of extension. Extract `cliId` from the output and assign it to telemetryUserId which is stored in globalState of extensionContext.
- When cliId is also undefined, then it generates a random userId and stores in globalState.
- In case of error, it checks globalState for userId, if undefined generates a random Id and updates the global state.
- Random id used as UserId will be unique to each extension.


klewis-sfdc:
Looking great!  I would do some renaming of the methods you've put in place.  Below are some screenshots of my suggestions:

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,65 @@ export class TelemetryService {
...
+        await this.extensionContext?.globalState.update('UserId', globalStateUserId);
+      }
+    } finally {
+      if (globalStateUserId === undefined) {
+        globalStateUserId = env.machineId;
+      }
+    }
+
+    return globalStateUserId;
+  }
```
peternhale, line 448: Just a suggestion. Using finally to set a value is unreliable given the finally block is run when garbage collection recovers the memory. I also don't see a need to use machineId at all.
``` suggestion
  public async getUserId(): Promise<string> {
    // Defining UserId in globalState and using the same in appInsights reporter.
    // Assigns cliId to UserId when it's undefined in global state.
    // cliId is undefined when cli-telemetry variable disable-telemetry is true.
    let globalStateUserId = this.extensionContext?.globalState.get<
      string | undefined
    >('UserId');

    if (globalStateUserId) {
      return globalStateUserId;
    }

    globalStateUserId = await this.executeCliTelemetry(
      this.buildCliTelemetryCommand()
    )
      .then((getCliTelemetryData): string => {
        const cmdResult = JSON.parse(getCliTelemetryData) as {
          result?: { cliId: string };
        };
        return cmdResult?.result?.cliId ?? this.getRandomUserId();
      })
      .catch(error => {
        console.log(
          `Error: ${error} occurred in retrieving cliId, generating user-id ..`
        );
        return this.getRandomUserId();
      });

    await this.extensionContext?.globalState.update(
      'UserId',
      globalStateUserId
    );

    return globalStateUserId;
  }
```

packages/salesforcedx-utils-vscode/src/telemetry/reporters/appInsights.ts
```diff
@@ -94,7 +95,7 @@ export class AppInsights extends Disposable implements TelemetryReporter {
 
     this.appInsightsClient.commonProperties = this.getCommonProperties();
     if (this.uniqueUserMetrics && env) {
-      this.appInsightsClient.context.tags['ai.user.id'] = env.machineId;
+      this.appInsightsClient.context.tags['ai.user.id'] = this.userId;
```
peternhale, line 98: I think this userId would also benefit non E4D telemetry. Could this change include adding a new property to common for all to use?

If E4D requires that it use a unique property, then please add a common property and adjust that coe as well.

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/telemetry/reporters/appInsights.ts
gbockus-sf: FWIW I think this is the common one.  ai = appinsights not artificially intelligent :)

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
gbockus-sf: good suggestion 👍

packages/salesforcedx-utils-vscode/src/telemetry/reporters/appInsights.ts
RitamAgrawal: This changes how we report userId in appInsights irrespective of the extensions. @peternhale Is there anything else you would want to change in how we report?

peternhale: No.

peternhale (approved):

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -4,9 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { randomBytes } from 'crypto';
 import * as util from 'util';
-import { ExtensionContext, ExtensionMode, workspace } from 'vscode';
-import { ActivationInfo } from '..';
+import { env, ExtensionContext, ExtensionMode, workspace } from 'vscode';
```
gbockus-sf, line 9: env is unused

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+    return new SfCommandBuilder()
+      .withArg('telemetry')
+      .withJson()
+      .build();
+  }
+
+  private async executeCliTelemetry(
+    command: Command
+  ): Promise<string> {
+    const workspacepath = workspaceUtils.getRootWorkspacePath();
```
gbockus-sf, line 401: nit: workspacePath

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+    return result;
+  }
+
+  public async getUserId(): Promise<string> {
+    // Defining UserId in globalState and using the same in appInsights reporter.
+    // Assigns cliId to UserId when it's undefined in global state.
+    // cliId is undefined when cli-telemetry variable disable-telemetry is true.
+    let globalStateUserId = this.extensionContext?.globalState.get<
+      string | undefined
+    >('UserId');
```
gbockus-sf, line 415: Use a constant for the key value.  I'd go with something like `sfdxTelemetryUserId` following the pattern of the only existing global key.

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -91,4 +91,47 @@ describe('Telemetry', () => {
       expect(result).toBe(telemetryService.extensionName);
     });
   });
+
+  describe('Telemetry Service - getUserId', () => {
```
gbockus-sf, line 95: Nit: to following the established naming convention this should just be the method under test.

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -91,4 +91,47 @@ describe('Telemetry', () => {
...
+    const fakeCliIdUndefined = {
+      result: {
+        enabled: true
+      }
+    };
+    const randomId = 'sedtrfyv234dd';
+    beforeEach(() => {
+      telemetryService = new TelemetryService();
+      executeCliTelemetrySpy = jest.spyOn((telemetryService as any), 'executeCliTelemetry');
+      getRandomUserIdSpy = jest.spyOn((telemetryService as any), 'getRandomUserId');
```
gbockus-sf, line 115: Should the globalState call be mocked here?

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -91,4 +91,47 @@ describe('Telemetry', () => {
...
+
+    it('should generate random userId when globalState userId as well as cliId is undefined', async () => {
+      executeCliTelemetrySpy.mockResolvedValueOnce(JSON.stringify(fakeCliIdUndefined));
+      getRandomUserIdSpy.mockResolvedValueOnce(randomId);
+
+      const uId = await telemetryService.getUserId();
+
+      expect(getRandomUserIdSpy).toHaveBeenCalled();
+      expect(uId).toBe(randomId);
+    });
```
gbockus-sf, line 135: Need a test for when the id is found in the global state.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+
+    if (globalStateUserId) {
+      return globalStateUserId;
```
gbockus-sf, line 418: needs test case.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+        const cmdResult = JSON.parse(getCliTelemetryData) as {
+          result?: { cliId: string };
+        };
+        return cmdResult?.result?.cliId ?? this.getRandomUserId();
+      })
+      .catch(error => {
+        console.log(
+          `Error: ${error} occurred in retrieving cliId, generating user-id ..`
+        );
+        return this.getRandomUserId();
```
gbockus-sf, line 434: needs test case

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+
+  private async executeCliTelemetry(
+    command: Command
+  ): Promise<string> {
+    const workspacepath = workspaceUtils.getRootWorkspacePath();
+    const execution = new CliCommandExecutor(command, { cwd: workspacepath }
+    ).execute();
+    const cmdOutput = new CommandOutput();
+    const result = cmdOutput.getCmdResult(execution);
+    return result;
```
gbockus-sf, line 406: no test coverage

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+        return cmdResult?.result?.cliId ?? this.getRandomUserId();
+      })
+      .catch(error => {
+        console.log(
+          `Error: ${error} occurred in retrieving cliId, generating user-id ..`
+        );
+        return this.getRandomUserId();
+      });
+
+    await this.extensionContext?.globalState.update(
```
gbockus-sf, line 437: So it Turns out that globalState is sandboxed for each extension.  SO if we set a value in the core extension, it's not going to be set for E4D or any other extension.  I don't think this is a big deal as we're intending to use cliID and if cli telemetry is turned off we will never reach this code flow.  Might be worth adding a comment above this line that is the randonUserId value is used here it will be unique per extension

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +385,60 @@ export class TelemetryService {
...
+    command: Command
+  ): Promise<string> {
+    const workspacepath = workspaceUtils.getRootWorkspacePath();
+    const execution = new CliCommandExecutor(command, { cwd: workspacepath }
+    ).execute();
+    const cmdOutput = new CommandOutput();
+    const result = cmdOutput.getCmdResult(execution);
+    return result;
+  }
+
```
gbockus-sf, line 408: So it was bugging me that we had all these new imports and function that seemed out of the _concern_ of the telemetry module.  I tried creating a UserService class to handle it.  Check out my branch (off your branch) and see what you think. 
gbockus/W-16086212-suggestUserService

gbockus-sf (commented):
Some nits to consider and potentially moving the userId & related to a new service class.  

Ran through QE and functionally looking good.

gbockus-sf (commented):
1. Verify I get the cli userId ✅
2. Verify userId is set in AppInsights ✅
    1. Can see cliId found in previous test step showing up for all extension in app insights
        1. 6a3b46564c9e3fc87ae614f61776a467be868a2e
    2. Query
1. Verify we still have machineId in appInsights ✅
2. Verify it is reused across VSCode reloads ✅
3. Verify I get the same UserID for various extensions ✅
    1. core ✅
    2. apex✅
    3. soql ✅
    4. e4d✅
4. Verify I get a random UserID if CLI telemetry is disabled ✅
    1. Had to comment out using global value in order to hit this flow. 
5. Verify that if I get a randomID it is persisted across reloads. ✅
6. Verify that when I open a different project my userId is persisted ✅
7. There is a weird state where if we use the randomly generated id it will differ between extensions due to the globalState not being shared across extension contexts ✅

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
RitamAgrawal: Thank you for calling these nit out.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
RitamAgrawal: In order to avoid confusion around sfdx/ sf prefix in variable names, using `telemetryUserId`.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
RitamAgrawal: added a test where the function returns telemetryUserId if defined.

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
RitamAgrawal: mocked get & update function to achieve the same.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
RitamAgrawal: second test under `getuserId` describe block.

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -91,4 +91,68 @@ describe('Telemetry', () => {
...
+    it('should generate random userId when telemetryUserId as well as cliId is undefined', async () => {
+      executeCliTelemetrySpy.mockResolvedValueOnce(JSON.stringify(fakeCliIdUndefined));
+      getRandomUserIdSpy.mockResolvedValueOnce(randomId);
+
+      const uId = await telemetryService.getUserId();
+
+      expect(getRandomUserIdSpy).toHaveBeenCalled();
+      expect(uId).toBe(randomId);
+    });
+
```
gbockus-sf, line 146: The following test will cover the catch condition in the promise chain. 

```suggestion
   it('should generate random userId when cli command throws', async () => {
      executeCliTelemetrySpy.mockRejectedValueOnce(new Error('no cli for you'));
      getRandomUserIdSpy.mockResolvedValueOnce(randomId);

      const uId = await telemetryService.getUserId();

      expect(getRandomUserIdSpy).toHaveBeenCalled();
      expect(uId).toBe(randomId);
    });
```

packages/salesforcedx-utils-vscode/test/jest/telemetry/telemetry.test.ts
```diff
@@ -91,4 +91,68 @@ describe('Telemetry', () => {
...
+      fakeGet.mockReturnValueOnce(globalTelemetryUserId);
+
+      const uId = await telemetryService.getUserId();
+
+      expect(executeCliTelemetrySpy).not.toHaveBeenCalled();
+      expect(getRandomUserIdSpy).not.toHaveBeenCalled();
+      expect(uId).not.toBe(randomId);
+      expect(uId).toBe(globalTelemetryUserId);
+    });
+  });
```
gbockus-sf, line 157: Describe for covering the `executeCliTelemetry` method
```suggestion
  describe('executeCliTelemetry()', () => {
    const cliTelemetryData = {
      result: {
        enabled: true,
        cliId: 'aTotallyValidCli-id_forSur3'
      }
    };

    let executeSpy: jest.SpyInstance;
    let getCmdResultSpy: jest.SpyInstance;
    let getRootWorkspacePathSpy: jest.SpyInstance;

    beforeEach(() => {
      getRootWorkspacePathSpy = jest.spyOn(workspaceUtils, 'getRootWorkspacePath');
      executeSpy = jest.spyOn(CliCommandExecutor.prototype, 'execute');
      getCmdResultSpy = jest.spyOn(CommandOutput.prototype, 'getCmdResult');

    });

    it('should return cli telemetry data', async () => {
      const fakeExecution = 'FindMeSoCliIdValue';
      const fakeCommand = 'sfdx force:telemetry:status --json';
      const fakePath = '/some/path';

      getRootWorkspacePathSpy.mockReturnValueOnce(fakePath);
      executeSpy.mockReturnValueOnce(fakeExecution);
      getCmdResultSpy.mockResolvedValueOnce(cliTelemetryData);

      const telemetryService = new TelemetryService();

      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
      const result = await (telemetryService as any).executeCliTelemetry(fakeCommand);

      expect(result).toBe(cliTelemetryData);
      expect(getRootWorkspacePathSpy).toHaveBeenCalled();
      expect(executeSpy).toHaveBeenCalled();
    });
  });
```

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -383,4 +386,60 @@ export class TelemetryService {
...
+  private getRandomUserId = (): string => randomBytes(20).toString('hex');
+
+  private buildCliTelemetryCommand(): Command {
+    return new SfCommandBuilder()
+      .withArg('telemetry')
+      .withJson()
+      .build();
+  }
+
+  private async executeCliTelemetry(
```
peternhale, line 399: @RitamAgrawal given that this function sole purpose is to run the telemetry command I would suggest creating the Command inside executeCliTelemetry instead having two functions.

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
RitamAgrawal: @peternhale okay! I see that's how we are implementing execute in other classes as well. I have added tests for buildTelemetryCommand  as well. Do you think it's necessary to combine the two fn?

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { randomBytes } from 'crypto';
+import { ExtensionContext } from 'vscode';
+import { CliCommandExecutor, Command, CommandOutput, SfCommandBuilder, workspaceUtils } from '..';
+import { TELEMETRY_GLOBAL_USER_ID } from '../constants';
+
+export class UserService {
+  private static getRandomUserId = (): string => randomBytes(20).toString('hex');
+
+  private static buildCliTelemetryCommand(): Command {
```
gbockus-sf, line 16: this method can be removed right?

RitamAgrawal: yes! I missed that. Thanks.

gbockus-sf (approved):
We did some final smoke testing with vsix installs in appsights and verified that our userId was set correctly for both myself and Ritam.  

Calling this good to go. 
:ship:
Nice work!

PR #5705, opened 2024-07-31
Mz/bundle core n templates

mingxuanzhangsfdx:
### What does this PR do?
This PR further bundles sfdx-core in core extension and templates

@W-15736930@

### Functionality Before
sfdx-core and templates are not bundled in core extension.

### Functionality After
sfdx-core and templates are bundled.

### DIFF
--- a/packages/salesforcedx-vscode-core/esbuild.config.js
+++ b/packages/salesforcedx-vscode-core/esbuild.config.js
@@ -6,7 +6,7 @@
  */
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { build } = require('esbuild');
-// const fs = require('fs').promises;
+const fs = require('fs').promises;
 
 const sharedConfig = {
   bundle: true,
@@ -22,18 +22,18 @@ const sharedConfig = {
 };
 
 // copy core-bundle/lib/transformStream.js to dist if core-bundle is included
-// const copyFiles = async (src, dest) => {
-//   try {
-//     // Copy the file
-//     await fs.copyFile(src, dest);
-//     console.log(`File was copied from ${src} to ${dest}`);
-//   } catch (error) {
-//     console.error('An error occurred:', error);
-//   }
-// };
+const copyFiles = async (src, dest) => {
+  try {
+    // Copy the file
+    await fs.copyFile(src, dest);
+    console.log(`File was copied from ${src} to ${dest}`);
+  } catch (error) {
+    console.error('An error occurred:', error);
+  }
+};
 
-// const srcPath = '../../node_modules/@salesforce/core-bundle/lib/transformStream.js';
-// const destPath = './dist/transformStream.js';
+const srcPath = '../../node_modules/@salesforce/core-bundle/lib/transformStream.js';
+const destPath = './dist/transformStream.js';
 
 (async () => {
   await build({
@@ -43,6 +43,6 @@ const sharedConfig = {
   });
 })()
   .then(async () => {
-    // await copyFiles(srcPath, destPath);
+    await copyFiles(srcPath, destPath);
   })
   .catch(() => process.exit(1));
--- a/packages/salesforcedx-vscode-core/esbuild.config.js
+++ b/packages/salesforcedx-vscode-core/esbuild.config.js
@@ -16,7 +16,6 @@ const sharedConfig = {
     'vscode',
     'applicationinsights',
     'shelljs',
-    '@salesforce/core-bundle',
     '@salesforce/templates-bundle'
   ],
   minify: true
--- a/packages/salesforcedx-vscode-core/esbuild.config.js
+++ b/packages/salesforcedx-vscode-core/esbuild.config.js
@@ -6,6 +6,7 @@
  */
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { build } = require('esbuild');
+const esbuildPluginPino = require('esbuild-plugin-pino');
 const fs = require('fs').promises;
 
 const sharedConfig = {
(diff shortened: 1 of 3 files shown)


peternhale (approved):

packages/salesforcedx-vscode-core/package.json
```diff
@@ -96,11 +96,11 @@
       "dist"
     ],
     "packageUpdates": {
-      "main": "dist/index.js",
+      "main": "dist/src/index.js",
+      "salesforceApiVersion": "61",
```
mingxuanzhangsfdx, line 100: I added this config in package.json of the core vsix, in order to provide the missing info from bundling template. So with this setting, one should be aware to update the field when the version of templates is bumped here with a bump of salesforceApiVersion.

I would also like to make this dynamic within esbuild.config.js, so one does not need to update the field frequently, but this added config should be always taken care.

The reason why it works is that the relative path of package.json of [baseGenerator.ts](https://github.com/forcedotcom/salesforcedx-templates/blob/main/src/generators/baseGenerator.ts#L69) within templates library is identical to the relative path of package.json within vsix (of dist/src/index.js).

packages/salesforcedx-vscode-core/esbuild.config.js
```diff
@@ -15,35 +16,40 @@ const sharedConfig = {
...
-//     console.error('An error occurred:', error);
-//   }
-// };
+const srcPathTransformStream = '../../node_modules/@salesforce/core-bundle/lib/transformStream.js';
+const destPathTransformStream = './dist/src/transformStream.js';
 
-// const srcPath = '../../node_modules/@salesforce/core-bundle/lib/transformStream.js';
-// const destPath = './dist/transformStream.js';
+const srcTemplatesPath = '../../node_modules/@salesforce/templates/lib/templates';
+const destTemplatesPath = './dist/templates';
```
mingxuanzhangsfdx, line 42: I made the structure bundle artifacts into a /src folder and a /templates folder, so when /dist/src/index.js refers to the generators from templates, the path to templates folder is identical to the path from [baseGenerator](https://github.com/forcedotcom/salesforcedx-templates/blob/main/src/generators/baseGenerator.ts#L86) to /templates in the original library.

packages/salesforcedx-vscode-core/test/vscode-integration/commands/templates/apexGenerateClass.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { ConfigUtil } from '@salesforce/salesforcedx-utils-vscode';
-import { nls as templatesNls } from '@salesforce/templates-bundle/lib/i18n';
+import { nls as templatesNls } from '@salesforce/templates/lib/i18n';
```
mingxuanzhangsfdx, line 9: The usage like this should not work, but vscode-integration is deprecated now so we just ignore it.

packages/salesforcedx-vscode-core/package.json
```diff
@@ -96,11 +96,11 @@
       "dist"
     ],
     "packageUpdates": {
-      "main": "dist/index.js",
+      "main": "dist/src/index.js",
```
mingxuanzhangsfdx, line 99: changed the path of entry due to the reason from the other comments

packages/salesforcedx-vscode-core/esbuild.config.js
```diff
@@ -6,7 +6,8 @@
  */
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { build } = require('esbuild');
-// const fs = require('fs').promises;
+const esbuildPluginPino = require('esbuild-plugin-pino');
+const fs = require('fs-extra');
```
peternhale, line 10: Where is this module added as a dependency?

packages/salesforcedx-vscode-core/package.json
peternhale: Not a big fan of this. Having this here does not guarantee it is correct. This data should come from templates module.

peternhale (changes requested):

packages/salesforcedx-vscode-core/esbuild.config.js
```diff
@@ -7,7 +7,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { build } = require('esbuild');
 const esbuildPluginPino = require('esbuild-plugin-pino');
-const fs = require('fs-extra');
+const fs = require('fs');
```
peternhale, line 10: thank you

packages/salesforcedx-vscode-core/esbuild.config.js
peternhale: @mingxuanzhangsfdx could we get the constraints of coordinating file placement within a bundle and the corresponding needs documented some where?

peternhale: Having these mappings bothers me because these are things that need to be remembered.

Given we are consuming templates in this way, what can we do to templates (and other modules that need to be bundled) so we retain structure moving a->b?

mingxuanzhangsfdx:
@peternhale I create a [PR](https://github.com/forcedotcom/salesforcedx-templates/pull/596) in templates to resolve the concern you raised.

e2e test [passes](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10378089870).

packages/salesforcedx-vscode-core/esbuild.config.js
```diff
@@ -15,35 +16,45 @@ const sharedConfig = {
...
-// const destPath = './dist/transformStream.js';
+const srcTemplatesPath = '../../node_modules/@salesforce/templates/lib/templates';
+const destTemplatesPath = './dist/templates';
 
 (async () => {
   await build({
     ...sharedConfig,
     entryPoints: ['./src/index.ts'],
-    outfile: 'dist/index.js'
+    outdir: 'dist/src'
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: I put the bundle artifacts inside a src folder in order to align the relative path to /templates with the original one. I can put them directly under dist, but that will make /templates parallel to /dist. So I need to adjust this during vscode:package process, which I found it hard to maintain. So the best solution is to construct it during /dist generation imho.

peternhale (approved):

Issue #5692, opened 2024-07-19
SFDX: Rename Component command fails when there's no aura folder in project


pozil:
### Summary

When using the `SFDX: Rename Component` command on a LWC you get the following error if you don't have an `aura` folder in your project:
> 15:21:54.155 Starting SFDX: Rename Component
> ENOENT: no such file or directory, scandir '/USER_PATH/my-project/force-app/main/default/aura'

I don't expect the `aura` folder to be required especially in new projets with no Aura components.

### Steps To Reproduce:

1.  Create a new project
1.  Get rid of the aura folder
1. Create a new LWC
1. Try to rename the LWC with the `SFDX: Rename Component` command

### Expected result

The component is renamed. The presence of an `aura` folder should not be a requirement.

### Actual result

You get the following error:
> 15:21:54.155 Starting SFDX: Rename Component
> ENOENT: no such file or directory, scandir '/USER_PATH/my-project/force-app/main/default/aura'

### Additional information

**Salesforce Extension Version in VS Code**:
Salesforce Extension Pack 61.5.0

**Salesforce CLI Version**:
@salesforce/cli/2.50.6 darwin-arm64 node-v18.17.0

**OS and version**:
macOS Sonoma 14.5

**VS Code version**:
1.91.1

CristiCanizales:
Hello @pozil! Thanks for creating this issue, I was able to reproduce it and the above WI has been created to address it.

daphne-sfdc:
Hi @pozil, This is fixed and published in this week's release v61.10.0.  Please upgrade your Salesforce Extensions for VSCode to the latest version to get the changes. 🙂

PR #5718, opened 2024-08-09
fix: failure when missing aura folder

peternhale:
### What does this PR do?

- lwc component rename no longer no long fails when aura folder is no present and vice versa
- added checks when creating a new lwc or aura component to ensure that the name does not already exist, enforcing the rule that there cannot be a LWC and an Aura component with the same name.

#[5692](https://github.com/forcedotcom/salesforcedx-vscode/issues/5692), @W-16285103@

### Functionality Before

- When renaming a LWC component and the aura folder does not exist, the rename would fail with an aura folder not found error
- The rule, "There cannot be a LWC and an Aura component with the same name", was not being enforced when creating either a LWC or Aura component.

### Functionality After
- Rename of a component where either lwc or aura folder is missing no longer encounters a folder not found.
- The rule, "There cannot be a LWC and an Aura component with the same name", is now enforced during component create requests.


packages/salesforcedx-vscode-core/src/commands/util/lwcAuraDuplicateComponentCheckers.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Checks for existing component name between LWC and Aura during rename
+ */
+export class LwcAuraDuplicateComponentCheckerForRename implements PostconditionChecker<OneOrMany> {
+  constructor(private readonly sourceFsPath: string) { }
+  async check(inputs: ContinueOrCancel): Promise<ContinueOrCancel> {
+    if (!isContinue(inputs)) {
+      return Promise.resolve(inputs);
+    }
+    if (Array.isArray(inputs.data)) {
+      // TODO: add to nls
```
CristiCanizales, line 36: this is done already right? In line 37

CristiCanizales (commented):
Looks good to me! 🚀  Just a small nit

CristiCanizales:
- [End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10325978071) ✅ 
- Can create/rename LWC with Aura folder (not) present ✅ 
- Can create/rename Aura Component with LWC folder (not) present ✅ 
- Can't create/rename LWC if there's an Aura component with the same name ✅ 
- Can't create/rename Aura Component if there's a LWC with the same name ✅

CristiCanizales (approved):
:shipit:

PR #5584, opened 2024-05-07
chore: fire an event upon the completion of the refreshSObjects command to notify E4D

floralan:
### What does this PR do?
When the refreshSobjects command is completed successfully, it fires an event. The `CommandEventDispatcher` serves as a middleman to expose this event in APIs, so that E4D can subscribe to the event.  

### What issues does this PR fix or reference?
@W-15573955@

### Functionality Before
No command completion events fired or exported in APIs

### Functionality After
CommandEventDispatcher is added to export events on refreshSobjects command completion

PR #5720, opened 2024-08-13
feat: return deploy status from command 

angelamuliu:
### What does this PR do?
Additionally returns status of deploy to those who directly call the vscode extension command

@W-16420054@

### Functionality Before
`await vscode.commands.executeCommand('sf.deploy.current.source.file');`
Would wait for completion of deploy command, but not inform caller directly of success of failure.

### Functionality After
`const result = await vscode.commands.executeCommand('sf.deploy.current.source.file');`
Now additionally returns true/false if deploy was successful or a failure (still waits for completion)

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
@@ -80,7 +80,8 @@ export const deploySourcePaths = async (
       )
     );
 
-    await commandlet.run();
+    const result = await commandlet.run();
+    return result;
   }
 };
 
--- a/packages/salesforcedx-vscode-core/src/commands/source/sourceTrackingGetStatusExecutor.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/source/sourceTrackingGetStatusExecutor.ts
@@ -27,14 +27,17 @@ export class SourceTrackingGetStatusExecutor extends LibraryCommandletExecutor<
   }
 
   public async execute(): Promise<boolean> {
-    const sourceStatusSummary: string = await SourceTrackingService.getSourceStatusSummary(
-      this.options || {}
-    );
-    channelService.appendLine(nls.localize('source_status'));
-    channelService.appendLine(sourceStatusSummary);
-    channelService.showChannelOutput();
-
-    return true;
+    try {
+      const sourceStatusSummary: string = await SourceTrackingService.getSourceStatusSummary(
+        this.options || {}
+      );
+      channelService.appendLine(nls.localize('source_status'));
+      channelService.appendLine(sourceStatusSummary);
+      channelService.showChannelOutput();
+      return true;
+    } catch (e) {
+      return false;
+    }
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/packages/salesforcedx-vscode-core/src/commands/source/sourceTrackingGetStatusExecutor.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/source/sourceTrackingGetStatusExecutor.ts
@@ -26,13 +26,15 @@ export class SourceTrackingGetStatusExecutor extends LibraryCommandletExecutor<
     this.options = options;
   }
 
-  public async execute(): Promise<void> {
+  public async execute(): Promise<boolean> {
     const sourceStatusSummary: string = await SourceTrackingService.getSourceStatusSummary(
       this.options || {}
     );
     channelService.appendLine(nls.localize('source_status'));
     channelService.appendLine(sourceStatusSummary);
     channelService.showChannelOutput();
+
+    return true;
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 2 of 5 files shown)


angelamuliu:
@mingxuanzhangsfdx FYI! From our conversation - https://salesforce-internal.slack.com/archives/C4LRHH0AD/p1722556176573019

packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
```diff
@@ -80,7 +80,8 @@ export const deploySourcePaths = async (
       )
     );
 
-    await commandlet.run();
+    const result = await commandlet.run();
+    return result;
```
mingxuanzhangsfdx, line 84: The type of result here is void, which does not look right. You might need to update the type definition in `packages/salesforcedx-vscode-core/src/commands/util/commandletExecutor.ts` and `packages/salesforcedx-vscode-core/src/commands/util/sfCommandlet.ts`

packages/salesforcedx-vscode-core/src/commands/source/sourceTrackingGetStatusExecutor.ts
```diff
@@ -26,13 +26,15 @@ export class SourceTrackingGetStatusExecutor extends LibraryCommandletExecutor<
     this.options = options;
   }
 
-  public async execute(): Promise<void> {
+  public async execute(): Promise<boolean> {
```
mingxuanzhangsfdx, line 29: Now that the return type is boolean, I feel using a `try catch` and returning false instead of undefined will make more sense to me.

packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
angelamuliu: Yeah, there's an inconsistency here. However, if I touch commandletExecutor + sfCommandlet, it forces me to uplift the other 20+ commands. Would we be okay leaving a comment in here notifying that currently only 1 command properly returns boolean, and the rest will return undefined?

peternhale (changes requested):
@angelamuliu the changes you are proposing do not fit into the current architecture of these command-let executors. I know you are trying to see if the command succeeds or not and the class that runs the actual command is tangled up with pre-checkers, gathers and post-checkers, which complicates the change you need.

To continue along with the notion of getting a boolean back from the execute method, the return type change would need to be applied to all classes that implement `CommandletExecutor`, which defines the signature and return type of the execute function.

This also means any function that calls one of the command lets would need to be examined to determine what to do with the return type, if anything.

@gbockus-sf any thoughts on alternate paths for this need?

peternhale:
> @angelamuliu the changes you are proposing do not fit into the current architecture of these command-let executors. I know you are trying to see if the command succeeds or not and the class that runs the actual command is tangled up with pre-checkers, gathers and post-checkers, which complicates the change you need.
> 
> To continue along with the notion of getting a boolean back from the execute method, the return type change would need to be applied to all classes that implement `CommandletExecutor`, which defines the signature and return type of the execute function.
> 
> This also means any function that calls one of the command lets would need to be examined to determine what to do with the return type, if anything.
> 
> @gbockus-sf any thoughts on alternate paths for this need?

@angelamuliu what is the delivery pressure for this change?

angelamuliu:
> > @angelamuliu the changes you are proposing do not fit into the current architecture of these command-let executors. I know you are trying to see if the command succeeds or not and the class that runs the actual command is tangled up with pre-checkers, gathers and post-checkers, which complicates the change you need.
> > To continue along with the notion of getting a boolean back from the execute method, the return type change would need to be applied to all classes that implement `CommandletExecutor`, which defines the signature and return type of the execute function.
> > This also means any function that calls one of the command lets would need to be examined to determine what to do with the return type, if anything.
> > @gbockus-sf any thoughts on alternate paths for this need?
> 
> @angelamuliu what is the delivery pressure for this change?

@peternhale It needs to go in this release, because the initiative driving this change ends in 254. Just the deploy command, no others needed.
If not done, our extension (that'll call vscode.commands.executeCommand('sf.deploy.current.source.file')) would be forced proceed w/o knowing the results, leading to a bad DX experience.

diyer:
@angelamuliu Unfortunately, as @peternhale said we cannot accept changes to our repo without heads up. Any feature work needs to be discussed between the product owner's first. Release planning is a good time to create team dependencies if required for any kind of required from our team included QA'ing and reviewing the changes. In addition to that, we need design docs, test plans and security reviews done(if applicable).

angelamuliu:
Had a discussion with @peternhale and closing this PR without merging 👍 

Instead, we can pursue changes via EventEmitters as done by the einstein VSCode extension in the past. Eg:
Core changes - https://github.com/forcedotcom/salesforcedx-vscode/pull/5584
Consumer https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/blob/52469a0ede02d0163dd7c1d870e188b6752c8fac/src/types/CommandEventDispatcher.ts

That'll be a new PR.

PR #5722, opened 2024-08-13
fix: no infinite running command notification for diff file and folder commands

CristiCanizales:
### What does this PR do?
Gets rid of the Running SFDX: Diff File Against Org notification when the diff source fails.

@W-16006725@

### Functionality Before
- Running notification even when the command already failed
image

### Functionality After
- Only failure notifications when it fails
image

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
@@ -85,7 +85,7 @@ export const sourceFolderDiff = async (explorerPath: vscode.Uri) => {
     new FilePathGatherer(explorerPath),
     new MetadataCacheExecutor(
       username,
-      'Source Diff',
+      nls.localize('source_diff_folder_text'),
       'source-diff-loader',
       handleCacheResults
     )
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
```diff
@@ -109,7 +109,7 @@ export const handleCacheResults = async (
     }
   } else {
     const message = nls.localize('source_diff_components_not_in_org');
-    await notificationService.showErrorMessage(message);
+    notificationService.showErrorMessage(message);
```
peternhale, line 112: Please use `void notificationService.showErrorMessage(message);`

peternhale: Also, on line 108, the await is not necessary, given `differ.diffFolder` does not return a promise.

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
CristiCanizales: good catch! done

daphne-sfdc (approved):
Approved! 🎉 

1. Diff file not present in the org ✅ 
Screenshot 2024-08-14 at 8 08 17 AM

2. Diff folder with no files present in the org ✅ 
Screenshot 2024-08-14 at 1 20 53 PM

3. Diff folder with some files present in the org -> only the files that are present in the org are listed in the diff ✅

peternhale (approved):

Issue #1437, opened 2019-06-27
Don't List Expired Orgs in the Org Picker


ntotten:
The CLI now defaults to listing only active orgs when running `sfdx force:org:list`, you now need to specify the `--all` parameter to list everything. The org picker should use the same behavior and only list valid orgs.

PR #5729, opened 2024-08-20
fix: ignore expired orgs

CristiCanizales:
### What does this PR do?
- Fixes list of connected orgs to not show expired/deleted orgs

@W-16501566@, #1437 

### Functionality Before
- Expired/Deleted orgs listed in the org picker dropdown

### Functionality After
- Expired/Deleted orgs **not** listed in the org picker dropdown


packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -88,31 +92,28 @@ export class OrgList implements vscode.Disposable {
         // scratch orgs parented by other (non-default) devHub orgs
         continue;
       }
-      const isExpired =
-        authFields && authFields.expirationDate
-          ? today >= new Date(authFields.expirationDate)
-          : false;
+      const isExpired = authFields?.expirationDate
```
peternhale, line 95: @CristiCanizales the OrgAuthorization does have a `isExpired` field, why not use that?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -88,31 +92,28 @@ export class OrgList implements vscode.Disposable {
...
-
       authList.push(authListItem);
     }
     return authList;
   }
 
   public async updateOrgList(): Promise<string[]> {
     const orgAuthorizations = await this.getOrgAuthorizations();
-    if (orgAuthorizations && orgAuthorizations.length === 0) {
+    if (orgAuthorizations?.length === 0) {
```
peternhale, line 116: I know this is existing code, but I don't see how getOrgAuthorizations() returns an undefined and why are we checking for an empty array when the filterAuthInfo will do the same?

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
CristiCanizales: True! I tried it elsewhere looking at other thing and it was always unknown so didn't think about using it here, but I just tried it here and it works fine. Good catch, thanks!

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -88,34 +91,26 @@ export class OrgList implements vscode.Disposable {
...
-      const isExpired =
-        authFields && authFields.expirationDate
-          ? today >= new Date(authFields.expirationDate)
-          : false;
 
+      if (orgAuth.isExpired === true) {
+        // If the scratch org is expired we don't want to see it in the org picker
+        continue;
+      }
       const aliases = await ConfigUtil.getAllAliasesFor(orgAuth.username);
```
peternhale, line 99: I was digging around in the OrgAuthorization and it also has aliases, so maybe we don't have to repeat the fetch here.

See https://github.com/forcedotcom/sfdx-core/blob/da5483e7bbbdd1837f313d168fe5aeaa5e5ec016/src/org/authInfo.ts#L90

peternhale (changes requested):

daphne-sfdc (approved):
Approved! 🎉 

Before:
Screenshot 2024-08-22 at 3 18 16 PM

After:
Screenshot 2024-08-22 at 3 11 49 PM

All the expired orgs are no longer displayed ✅

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -88,34 +90,26 @@ export class OrgList implements vscode.Disposable {
...
     return authList;
   }
 
   public async updateOrgList(): Promise<string[]> {
-    const orgAuthorizations = await this.getOrgAuthorizations();
-    if (orgAuthorizations && orgAuthorizations.length === 0) {
-      return [];
-    }
-    const authUsernameList = await this.filterAuthInfo(orgAuthorizations);
+    const authUsernameList = await this.filterAuthInfo(
```
mingxuanzhangsfdx, line 110: You skipped the check if `orgAuthorizations` is empty here, which will always have a call of `filterAuthInfo`. In `filterAutoInfo`, there will always be a `const targetDevHub = await OrgAuthInfo.getDevHubUsername();` which might be dummy.

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
```diff
@@ -245,7 +243,7 @@ describe('orgList Tests', () => {
         expect(authList[0]).to.equal('alias1 - test-username1@example.com');
       });
 
-      it('should flag the org as expired if expiration date has passed', async () => {
+      it('should filter the list to ignore expired orgs', async () => {
```
mingxuanzhangsfdx, line 246: Why updating vscode-integration test? I think the test has been deprecated and there is no plan to use it again. I would suggest a jest unit test for the change you made in `orgList.ts`.

mingxuanzhangsfdx (changes requested):

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
CristiCanizales: Because if orgAuthorizations is empty filterAuthInfo will return an empty array anyways. See https://github.com/forcedotcom/salesforcedx-vscode/pull/5729#discussion_r1727107139

packages/salesforcedx-vscode-core/test/vscode-integration/orgPicker/orgList.test.ts
CristiCanizales: I know we're not running them in GHA, but I try to keep them up to date anyways because they can still be run locally and provide some direction.

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -88,34 +90,26 @@ export class OrgList implements vscode.Disposable {
...
-        authListItem += ` - ${nls.localize(
-          'org_expired'
-        )} ${String.fromCodePoint(0x274c)}`; // cross-mark
+      if (orgAuth.isExpired === true) {
+        // If the scratch org is expired we don't want to see it in the org picker
+        continue;
       }
+      const aliases = orgAuth.aliases || [];
+      const authListItem =
+        aliases?.length > 0
```
peternhale, line 100: Don't need the '?' now that aliases is always an array
```suggestion
        aliases.length > 0
```

peternhale (approved):
Just a nit suggestion

mshanemc:
closed, WI was never'd

PR #5768, opened 2024-08-27
chore: add eventEmitter on libraryCommand

angelamuliu:

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -130,6 +130,10 @@ export abstract class LibraryCommandletExecutor<T>
   protected showChannelOutput = true;
   protected readonly telemetry = new TelemetryBuilder();
 
+  public static readonly libraryCommandCompletionEventEmitter = new vscode.EventEmitter<boolean>();
+  public static readonly onLibraryCommandCompletion: vscode.Event<boolean> =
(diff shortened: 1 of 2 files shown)


peternhale (approved):

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -220,13 +224,15 @@ export abstract class LibraryCommandletExecutor<T>
         properties,
         measurements
       );
+      LibraryCommandletExecutor.libraryCommandCompletionEventEmitter.fire(!!success);
```
mingxuanzhangsfdx, line 227: Not sure if it is meaningful. LibraryCommandletExecutor is an abstract class to be extended by other real executors such as `orgLogoutDefault` or `orgLoginAccessToken`. How does the subscriber know the result from which command it is listening to with only a boolean value? Can you help me understand that? @angelamuliu @peternhale

peternhale: Great catch @mingxuanzhangsfdx, there is no way to know. If two commands are run concurrently and are listening for the completion event, knowing which event belongs any given listener is indeterminant.

peternhale (changes requested):

peternhale:
@angelamuliu This PR cannot move forward due to concurrency issues.

PR #5769, opened 2024-08-27
build: remove e2e cjs due to min vscode version

peternhale:
Given min vscode ersion is now 1.86.0, the CJS automation tests will no longer run (wdio constraint). Removing e2e-cjs workflow.
@W-16596055@


CristiCanizales (approved):
Thanks!

PR #5838, opened 2024-09-11
fix: show error message for push/pull when the error does not involve files

CristiCanizales:
### What does this PR do?
- Shows error message for push/pull when the error is not related to files

### Functionality Before
- no error message and command kept running forever
image

### Functionality After
- Error message displayed and no infinite notification for running push/pull
image

[skip-validate-pr]


daphne-sfdc (approved):
Approved! 🎉 

Test plan here: https://salesforce.quip.com/AaoSA8JSZkUe ✅

PR #5848, opened 2024-09-13
chore: update telemetry developer doc

gbockus-sf:
### What does this PR do?
When QE'ing the a recent PR I had to dig into how to enabled local telemetry logging for both Dev mode & VSIX mode.  This PR update the telemetry.md developer doc to capture the HOWTO. 

[#5849](https://github.com/forcedotcom/salesforcedx-vscode/issues/5849)
Related to: @W-16207092@

### Functionality Before
Doc was out of date

### Functionality After
Doc is more accurate and how HOWTO for local telemetry files.


contributing/telemetry.md
```diff
@@ -4,9 +4,12 @@ Similar to what vscode does, we are reporting on certain events that happen in t
 
 ## How it works
 
-All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is responsible for showing an information message to users and properly initializing the service for other extensions in this repository to leverage. Similar to VSCode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
```
gbockus-sf, line 7: This was no longer accurate so I removed it.

contributing/telemetry.md
```diff
@@ -4,9 +4,12 @@ Similar to what vscode does, we are reporting on certain events that happen in t
...
+All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. 
 
-Additionally, if telemetry is disabled for Salesforce CLI using the `disable-telemetry` configuration setting or the `SF_DISABLE_TELEMETRY` environment variable, the telemetryService doesn't send data.
+There are several ways to disable telemetry for all Salesforce extensions.
+- Turn off the Core Extension telemetry setting at the workspace `("salesforcedx-vscode-core.telemetry.enabled": false)`
+- Disable SF CLI telemetry 
+  - `sf config set disable-telemetry=true --global`
```
gbockus-sf, line 12: Added specific list of how to disable telemetry

contributing/telemetry.md
```diff
@@ -4,9 +4,12 @@ Similar to what vscode does, we are reporting on certain events that happen in t
 
 ## How it works
 
-All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data. The service in `salesforce-vscode-core` is responsible for showing an information message to users and properly initializing the service for other extensions in this repository to leverage. Similar to VSCode, we provide the user an opt out mechanism. We check that user settings `telemetry.enableTelemetry` and `salesforcedx-vscode-core.telemetry.enabled` are enabled before initializing the service.
+All extensions leverage the telemetryService in `salesforce-vscode-core` extension in order to send data.
```
diyer, line 7: There were some extensions like visualforce and debugger telemetry that did something different. Do these extensions also leverage the core extension telemetry?

contributing/telemetry.md
diyer: we have not called out internal users use case here. is that intentional?

diyer: wonder why we have 2 setting instead of just 1?

gbockus-sf: heh...we really have 3, but the top level vscode setting is no longer valid since VSCode made a change to how to manage this (when did this change happen? no idea... I'm pretty sure before I was ever working on this product so at least 3 years ago).

gbockus-sf: I view these docs as general and the internal use case is specific to just SFDC users.  Think we should call it out here?

contributing/telemetry.md
gbockus-sf: I just deleted the part I knew to be totally wrong.  AFAIK every extension in the monorepo has it's own telemetry service created via the shared utils package, and external extensions to the monorepo use the core extension to get access to creating a TelemetryService.  Can flush this out more.

contributing/telemetry.md
diyer: Unless you see any downside to it, it will be good to capture it here so we know what to expect for internal users?

contributing/telemetry.md
```diff
@@ -15,3 +18,23 @@ Additionally, if telemetry is disabled for Salesforce CLI using the `disable-tel
 - Create a telemetry service under src
 - Initialize the telemetry service on the extension's `activate` call and initialize it using `salesforce-vscode-core` telemetry service
 - Use the telemetry service where needed in the extension
+
+## Logging telemetry to a file
```
diyer, line 22: Btw, I just noticed this slack thread https://salesforce-internal.slack.com/archives/C4LRHH0AD/p1726781465323499?thread_ts=1726780954.506789&cid=C4LRHH0AD where Pete mentions this setting for local telemetry "salesforcedx-vscode-apex.advanced": {
      "localTelemetryLogging": "true"
  },

Are we missing it in the docs?

gbockus-sf: Wasn't aware of that one.   It's dev mod specific.  Will add it.

daphne-sfdc (approved):

PR #5856, opened 2024-09-18
feat: core extension command log

jgellin-sf:
### What does this PR do?
This PR logs commands, including exit codes and errors from CLI command executions, in the `workspaceState` `Memento` of the core extension. Additional arbitrary data can also be logged with the commands, but this is not currently used. It also allows dependent extensions to query the command log from the API. 

Command events are streamed using the new `CommandEventStream` class. A `START` event, including the `commandId` signals the beginning of logging for a given command, and an `END` event signals the end, at which point the command log entry is written to the log which includes duration information as well as any data, error, or exit code events logged between the `START` and `END`. For some long-running events, there is a chance of overlapping streams from multiple commands, so it is possible that data/errors/exit codes from one event are written to multiple events. The risk of incorrect data for overlapping commands is there, but a fully traced command log would be difficult to achieve without a major refactor. This solution attempts to achieve as much as possible as a general solution. 

@W-16486962@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
@@ -174,18 +174,14 @@ export class CompositeCliCommandExecution implements CommandExecution {
   }
 
   public successfulExit() {
-    CommandEventStream.getInstance().post({ type: CommandEventType.EXIT_CODE, exitCode: 0 });
     this.exitSubject.next(0);
   }
 
   public failureExit(e?: {}) {
     if (e) {
       // eslint-disable-next-line @typescript-eslint/no-base-to-string, @typescript-eslint/restrict-template-expressions
-      CommandEventStream.getInstance().post({ type: CommandEventType.ERROR, error: `${e}` });
-      // eslint-disable-next-line @typescript-eslint/no-base-to-string, @typescript-eslint/restrict-template-expressions
       this.stderr.next(`${e}${os.EOL}`);
     }
-    CommandEventStream.getInstance().post({ type: CommandEventType.EXIT_CODE, exitCode: 1 });
     this.exitSubject.next(1);
   }
 }
--- a/packages/salesforcedx-utils-vscode/test/jest/cli/commandExecutor.test.ts
+++ b/packages/salesforcedx-utils-vscode/test/jest/cli/commandExecutor.test.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { CliCommandExecutor } from "../../../src/cli/commandExecutor";
+import { CommandEventStream } from "../../../src/commands/commandEventStream";
+
+
+describe('CompositeCliCommandExecution', () => {
+  const commandEventStreamPostSpy = jest.spyOn(CommandEventStream.getInstance(), 'post');
+
+});
+
+describe('CliCommandExecutor', () => {
+  describe('patchEnv', () => {
+    it('should patch the environment with the base environment', () => {
+      const baseEnvironment = new Map<string, string>();
+      baseEnvironment.set('SFDX_ENV', 'test');
+      const options = { env: { TEST: 'test' } };
+      const patchedOptions = CliCommandExecutor.patchEnv(options, baseEnvironment);
+      expect(patchedOptions.env).toEqual({ SFDX_ENV: 'test', TEST: 'test', SFDX_TOOL: 'salesforce-vscode-extensions' });
+    });
+  });
+});


packages/salesforcedx-utils-vscode/src/commands/commandEventStream.ts
```diff
@@ -0,0 +1,67 @@
...
+export type CommandEventError = {
+  type: CommandEventType.ERROR;
+  error: string;
+};
+
+export type CommandEvent = CommandEventStart | CommandEventEnd | CommandEventExitCode | CommandEventData | CommandEventError;
+
+export class CommandEventStream {
+  private static instance: CommandEventStream;
+  private readonly eventEmitter = new vscode.EventEmitter<CommandEvent>();
```
gbockus-sf, line 47: should this be assigned in the constructor for testability?

packages/salesforcedx-vscode-core/src/commands/util/commandWrapper.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as vscode from 'vscode';
+
+export const registerCommand = (commandId: string, callback: (...args: any[]) => any, thisArg?: any): vscode.Disposable => {
+  return vscode.commands.registerCommand(commandId, wrapCommandCallback(commandId, callback), thisArg);
+};
+
+const wrapCommandCallback = (commandId: string, callback: (...args: any[]) => any): (...args: any[]) => any => {
+  return async (...args: any[]) => {
+    CommandEventStream.getInstance().post({ type: CommandEventType.START, commandId });
+    await callback(...args);
```
gbockus-sf, line 18: I think we need to return the result of this.  One thing I'm not clear about...what if the caller doesn't want to wait on the promise to complete.  

Should this be something like: 
```
...
const result = Promise.resolve(callback(...args)).then((result) => {
  CommandEventStream.getInstance().post({ type: CommandEventType.END, commandId });
  return result;
});
```
That way we always get the END event recorded, but the caller has control on awaiting the promise, and access to the result of the call when desired.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -673,7 +676,9 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
       TelemetryService,
       WorkspaceContext,
       CommandEventDispatcher
-    }
+    },
+    getCommandLog,
+    getLastCommandLogEntry
```
gbockus-sf, line 681: These seems they they should be wrapped in a service vs just exported as standalone function to following the existing shared services pattern.  CommandService on like 679 instead?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -140,257 +143,257 @@ const registerCommands = (
...
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const diffFolder = vscode.commands.registerCommand(
+  const diffFolder = registerCommand(
     'sf.folder.diff',
     sourceFolderDiff
   );
 
-  const forceRefreshSObjectsCmd = vscode.commands.registerCommand(
+  const forceRefreshSObjectsCmd = registerCommand(
```
gbockus-sf, line 385: How would you feel about only doing this for sobject refresh for the initial release, but expanding to other services after we've had a couple of week to ensure there are no surprises with just wrapping sobjectRefresh?

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -20,16 +22,21 @@ import { workspaceContextUtils } from '.';
...
     this.onOrgChange = workspaceContextUtil.onOrgChange;
     this.onOrgChange(this.handleCliConfigChange);
+    this.workspaceStore = {} as vscode.Memento;
   }
 
   public async initialize(extensionContext: vscode.ExtensionContext) {
     await WorkspaceContextUtil.getInstance().initialize(extensionContext);
+    this.workspaceStore = extensionContext.workspaceState;
+    CommandEventStream.getInstance().initialize(extensionContext);
+    CommandLog.getInstance().initialize();
```
gbockus-sf, line 39: Registering with another service doesn't seem like a business WorkspaceContext should be in.  
Move to the initializeProject function? https://github.com/forcedotcom/salesforcedx-vscode/blob/c7aa878a847eae8909f0ec6d561762e35672d700/packages/salesforcedx-vscode-core/src/index.ts#L714

packages/salesforcedx-vscode-core/src/commands/util/commandLogEntry.ts
```diff
@@ -0,0 +1,15 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export type CommandLogEntry = {
+  commandId: string;
+  timestamp: number;
+  duration: number;
+  exitCode?: number;
+  error?: string;
+  data?: any;
```
gbockus-sf, line 14: unknown would be better for consuming the data in a type safe way.

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
```diff
@@ -0,0 +1,111 @@
...
+
+import { CommandEvent, CommandEventStream, CommandEventType } from '@salesforce/salesforcedx-utils-vscode';
+import { WorkspaceContext } from '../../context';
+import { CommandLogEntry } from './commandLogEntry';
+
+export const getCommandLog = (commandIdFilter?: string, exitCodeFilter?: number): CommandLogEntry[] => {
+  return CommandLog.getInstance().getCommandLog(commandIdFilter, exitCodeFilter);
+};
+
+export const getLastCommandLogEntry = (commandIdFilter?: string, exitCodeFilter?: number): CommandLogEntry | undefined => {
```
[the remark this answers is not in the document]
gbockus-sf: Also what's the use case for getting the last commandLogEntry? Seems error prone as there can always be a bunch of command running in parallel

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { CommandEvent, CommandEventStream, CommandEventType } from '@salesforce/salesforcedx-utils-vscode';
+import { WorkspaceContext } from '../../context';
+import { CommandLogEntry } from './commandLogEntry';
+
```
gbockus-sf, line 11: The stand alone methods...should they be static methods on CommandLog?

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
```diff
@@ -0,0 +1,111 @@
...
+      data: this.inProgressData.data
+    };
+    this.commandLogEntries.push(entry);
+    if (this.commandLogEntries.length > CommandLog.MAX_LOG_ENTRIES) {
+      this.commandLogEntries.shift();
+    }
+    await this.updateCommandLog();
+  }
+
+  public getCommandLog(commandIdFilter?: string, exitCodeFilter?: number): CommandLogEntry[] {
```
gbockus-sf, line 61: is commandIdFilter purposely optional?   
I guess it comes with the benefit that I could get all command with -1 exit code, but I'm more concerned with code that would accidentally pass undefined as the commandId then react as if they got back command that's weren't expected. 
```
let startCmd = undefined;
let failedStartCommands = getCommandLog(startCmd, -1) 
console.log('All failed start commands', failedStartCommands);
```

packages/salesforcedx-vscode-core/src/index.ts
peternhale: This is a great idea @gbockus-sf

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
```diff
@@ -155,28 +156,36 @@ export class CompositeCliCommandExecution implements CommandExecution {
...
 
   public successfulExit() {
+    CommandEventStream.getInstance().post({ type: CommandEventType.EXIT_CODE, exitCode: 0 });
     this.exitSubject.next(0);
   }
 
   public failureExit(e?: {}) {
     if (e) {
+      // eslint-disable-next-line @typescript-eslint/no-base-to-string, @typescript-eslint/restrict-template-expressions
+      CommandEventStream.getInstance().post({ type: CommandEventType.ERROR, error: `${e}` });
```
gbockus-sf, line 184: Is there a safer way to do this without all the eslint bypasses?

packages/salesforcedx-vscode-core/src/commands/util/commandWrapper.ts
```diff
@@ -0,0 +1,21 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { CommandEventStream, CommandEventType } from '@salesforce/salesforcedx-utils-vscode';
+import * as vscode from 'vscode';
+
+export const registerCommand = (commandId: string, callback: (...args: any[]) => any, thisArg?: any): vscode.Disposable => {
+  return vscode.commands.registerCommand(commandId, wrapCommandCallback(commandId, callback), thisArg);
```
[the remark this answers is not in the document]
jgellin-sf: Correct

packages/salesforcedx-utils-vscode/src/commands/commandEventStream.ts
jgellin-sf: TS complains somewhere about the member not being initialized if I do that, when I write the tests I'll see if I can refactor to make everything happy.

packages/salesforcedx-utils-vscode/src/commands/commandEventStream.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public initialize(extensionContext: vscode.ExtensionContext) {
+    extensionContext.subscriptions.push(this.eventEmitter);
+  }
+
+  public readonly onCommandEvent: vscode.Event<CommandEvent> = this.eventEmitter.event;
+
+  public post(event: CommandEvent) {
+    this.eventEmitter.fire(event);
```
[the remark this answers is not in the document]
jgellin-sf: yes. good call

packages/salesforcedx-vscode-core/src/index.ts
jgellin-sf: Good idea I'll give that a whorl.

packages/salesforcedx-vscode-core/src/index.ts
jgellin-sf: 👍

packages/salesforcedx-vscode-core/src/commands/util/commandLogEntry.ts
jgellin-sf: 👍

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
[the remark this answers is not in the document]
jgellin-sf: Yeah it's really not necessary. It's easy enough for a client of this to find the last one. I'm thinking about just having `getCommandLog` and an options parameter that has the `commandIdFilter` and `exitCodeFilter` and maybe a timestamp sort order as well. Overkill?

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
jgellin-sf: My goal was to hide the `CommandLog` implementation from the API. If I move a new service to the API, I can get rid of the standalone methods.

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
jgellin-sf: Yeah I wanted to leave open the option for clients to get multiple types of commands. But parameter order makes it seem like it shouldn't be `undefined`. It seems to me like an argument for:

> Yeah it's really not necessary. It's easy enough for a client of this to find the last one. I'm thinking about just having getCommandLog and an options parameter that has the commandIdFilter and exitCodeFilter and maybe a timestamp sort order as well. Overkill?

packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
jgellin-sf: I'll see what I can do.

packages/salesforcedx-vscode-core/src/commands/util/commandLog.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export class CommandLog {
+  private static instance: CommandLog;
+  private static STORAGE_KEY = 'COMMAND_LOG';
+  private static MAX_LOG_ENTRIES = 1000;
+
+  private commandLogEntries: CommandLogEntry[] = [];
+  private inProgressCommands: Map<string, number> = new Map();
+  private inProgressData: any = {};
```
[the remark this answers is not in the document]
jgellin-sf: Not sure I understand the question completely. But the reason things are set up like this is because of where the code knows what the `commandId` is and where the code does not. The code only knows the `commandId` at command registration. But for other information that can get logged (errors, exit codes, misc data), that data is coming from implementation code where we don't have the `commandId`. The `inProgressCommands` keeps track for commands for which a `START` event has been issued but no `END` yet. The `inProgressData` keeps track of any `EXIT_CODE`, `DATA, `ERROR` events that are issued in between those events.

If the log simply logged the raw events, it would be harder to pull the data like `EXIT_CODE` and associate it with a given command.

PR #5786, opened 2024-08-28
fix: implement provider for telemetry

peternhale:
@W-16507598@

Registers a vscode core command that provides an instance of a telemetry service.

PR #5857, opened 2024-09-19
fix: telemetry user id is now properly filled

peternhale:
Telemetry user Id is now being populated correctly

### What does this PR do?
This PR corrects a bug introduced by #5786. The referenced PR did not take into account an extension's fully qualified 'id' when trying to fetch the extension context reference from vscode, resulting in a userId value of "unknown".

This fix no longer uses logic to lookup an extension context but uses an instance provided by the caller.

@W-16742418@

### Functionality Before
All telemetry records carried a user Id value of unknown

### Functionality After
Telemetry records will have a user Id, when the extension context is valid.


packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
```diff
@@ -5,8 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import { TELEMETRY_GLOBAL_USER_ID } from '@salesforce/salesforcedx-utils-vscode/src/constants';
```
gbockus-sf, line 8: is the import here intentional? 
@salesforce/salesforcedx-utils-vscode/src/constants -> @salesforce/salesforcedx-utils-vscode

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
```diff
@@ -95,7 +97,7 @@ describe('Telemetry', () => {
       expect(telemetryEnabled).toEqual(true);
 
       await showTelemetryMessage(mockExtensionContext);
-      expect(globalStateTelemetrySpy).toHaveBeenCalledTimes(2);
+      expect(globalStateTelemetrySpy).toHaveBeenCalledTimes(3);
```
gbockus-sf, line 100: Why'd this change?  Should we test for the third value below here like we do for the other 2?

daphne-sfdc (approved):
Approved! 🎉 

Test plan here: https://salesforce.quip.com/Ty2QA58suY10 ✅

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
peternhale: Yes

gbockus-sf: is that value not exported from the top of the package? 

I realize this is a test, but this a pattern that should be avoided as it causes bundling issue when outside of tests.

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
peternhale: The function `UserService.getTelemetryUserId` fetches the global state to get value for `TELEMETRY_GLOBAL_USER_ID`. The bug I introduced prevented the call from happening since extensionContext was always undefined.

Now that the bug is fixed, I needed to adjust this check since there is an additional call to global state `get`.

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
```diff
@@ -95,7 +97,7 @@ describe('Telemetry', () => {
...
       expect(globalStateTelemetrySpy).toHaveBeenCalledWith(TELEMETRY_GLOBAL_VALUE);
```
gbockus-sf, line 101: For consistency in the test. 
```suggestion
      expect(globalStateTelemetrySpy).toHaveBeenCalledWith(TELEMETRY_GLOBAL_USER_ID);
```

packages/salesforcedx-vscode-core/test/jest/telemetry/index.test.ts
gbockus-sf: I can never make those code suggestions work.

Can you add the following after the above change so we check for the three expected values? 
```
      expect(globalStateTelemetrySpy).toHaveBeenCalledWith(TELEMETRY_GLOBAL_USER_ID);
```

gbockus-sf (approved):
Couple of test follow ups for you to consider, but should be good to go 👍

gbockus-sf (approved):
🚢

PR #5865, opened 2024-09-20
chore: refactor initsobjectsdefinitions and send logname in telemetry

CristiCanizales:
### What does this PR do?
- Sends command log name to telemetry instead of just error.name so filtering can be more specific
- Refactors initSObjectDefinitions() to have a different source and sobject folder depending on the refresh setting

@W-16563596@

### Functionality After
exceptions in telemetry can be searched with logname


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -685,15 +691,10 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
...
-    if (sobjectRefreshStartup) {
-      initSObjectDefinitions(
-        vscode.workspace.workspaceFolders[0].uri.fsPath
-      ).catch(e => telemetryService.sendException(e.name, e.message));
-    } else {
-      checkSObjectsAndRefresh(
-        vscode.workspace.workspaceFolders[0].uri.fsPath
-      ).catch(e => telemetryService.sendException(e.name, e.message));
-    }
+    initSObjectDefinitions(
```
peternhale, line 694: Nice!

peternhale (approved):

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -219,7 +221,7 @@ export abstract class LibraryCommandletExecutor<T>
       );
     } catch (e) {
       if (e instanceof Error) {
-        telemetryService.sendException(e.name, e.message);
+        telemetryService.sendException(this.logName, e.message);
```
gbockus-sf, line 224: Is losing the error name here a big deal?   I'm not sure what kind of errors are getting handled today.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -685,15 +691,15 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
...
+      vscode.workspace.workspaceFolders[0].uri.fsPath,
```
gbockus-sf, line 695: Can workspaceFolders be empty here?  Nothing sadder then the ol `unable to read uri of undefined` or whatever

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -685,15 +691,15 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
...
+      sobjectRefreshStartup
+    ).catch(e =>
+      telemetryService.sendException(
+        e.name,
+        `Error: ${e.message} with sobjectRefreshStartup = ${sobjectRefreshStartup}`
+      )
```
gbockus-sf, line 701: Why not do this error better like you did the other one.  This makes it easy to find these errors in telemetry. Maybe
```
telemetryService.sendException(
        'initSObjectDefinitionsError', // should be a constant
        `Error: name = ${e.name} message = ${e.message} with sobjectRefreshStartup = ${sobjectRefreshStartup}`
      )
```

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
gbockus-sf: A prefix on this.logName would make these exceptions identifiable in appinsights
`LibraryCommandletExecutor-${this.logName}` maybe?

gbockus-sf (commented):
Couple of thoughts for you consideration...nothing blocking

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
diyer: what do we mean by error names here? Can you share some examples please?

packages/salesforcedx-vscode-core/src/index.ts
CristiCanizales: haha no, it can't. We do the validation a couple lines before

packages/salesforcedx-vscode-core/src/index.ts
CristiCanizales: That was the idea but I missed it. Thanks!

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
gbockus-sf: It can be literally any string. If the default Error is used it will be 'Error' 
image
image

PR #5900, opened 2024-10-15
fix: move resolve orphan apex lang server

peternhale:
Move resolveAnyFoundOrphanLanguageServers outside of client setup Run resolveAnyFoundOrphanLanguageServers in separate tick on main loop
@W-16953996@

### Functionality Before
Code to locate and resolve orphaned apex language servers was run inside the language client setup, with potential downside of impeding performance.

### Functionality After
Code to locate and resolve orphaned apex language servers now runs as the last element of the apex extension activate function. It is also run with a setImmediate so it will run on the next iteration of the main loop.


CristiCanizales (approved):
LGTM!
[Test run](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/11354852361) ✅ 
Manual testing looks good as well ✅

Issue #5903, opened 2024-10-18
Do not switch to the Output panel when deploy on save is enabled


tehnrd:
**Is your feature request related to a problem? Please describe.**
When the "Push-or-deploy-on-save" setting is enabled, I’m always frustrated when I save a file, and VSCode changes the panel to the Output panel when saving the record. This completely disrupted the development experiences.

**Describe the solution you'd like**
By default, when "Push-or-deploy-on-save" is enabled, the UI should not switch to the Output panel when the files are saved. A setting can override this default behavior and show the Output panel if the user wants this behavior.

**Describe alternatives you've considered**
I am unaware of any alternatives to changing the output panel behavior when “Push-or-deploy-on-save” is enabled.

**Additional context**
This feature has been requested in some form many times and has attempted to be addressed in PRs.
Issue: https://github.com/forcedotcom/salesforcedx-vscode/issues/1429
Issue: https://github.com/forcedotcom/salesforcedx-vscode/issues/1110
PR: https://github.com/forcedotcom/salesforcedx-vscode/pull/4337
PR: https://github.com/forcedotcom/salesforcedx-vscode/pull/4971
PR: https://github.com/forcedotcom/salesforcedx-vscode/pull/1865
PR: https://github.com/forcedotcom/salesforcedx-vscode/pull/1181

The PRs to address this issue have had regressions, which no longer work. Some of the feedback on other PRs was that a global setting for suppressing output panel switching was too aggressive, so this suggestion is for Push-or-deploy-on-save only.

Here is a video showing the before and after functionality:
https://www.loom.com/share/a59adaa191474e64a835e9d52121f6c3?sid=7d8250d5-8e2d-4904-afda-6f0d05888f79

tehnrd:
I think there is an argument to be made in which there should be no setting at all, and by default, the UI simply does not change to the Output panel at all with Deploy on save enabled.

mingxuanzhangsfdx:
Hi @tehnrd thanks for putting up the PR. it has been merged and will get into next week's release. I know there are some discussions still around. If you are still interested in further enhancement, please let us know so we could specify the requirements. Due to business need, we do not have enough bandwidth developing this area for now. We really appreciate your contributions.

PR #5904, opened 2024-10-18
feat(settings): control switching to Output panel when Push-or-deploy…

tehnrd:
### What does this PR do?
When Push-or-deploy-on-save is enabled, do not switch the UI to the Output panel when a file is saved.

https://github.com/forcedotcom/salesforcedx-vscode/issues/5903
#5903 @W-17206953@

### Functionality Before
https://www.loom.com/share/a59adaa191474e64a835e9d52121f6c3?sid=36357412-c0f6-48b8-93bd-ea2df8b6f16b

### Functionality After
https://www.loom.com/share/a59adaa191474e64a835e9d52121f6c3?t=279&sid=9bc3b0f9-7245-4026-aa22-fd3e64d05300 (same video with timestamp start at functionality after)


packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
```diff
@@ -19,8 +20,9 @@ import { CompositePostconditionChecker } from './util/compositePostconditionChec
 import { TimestampConflictChecker } from './util/timestampConflictChecker';
 
 export class LibraryDeploySourcePathExecutor extends DeployExecutor<string[]> {
-  constructor() {
+  constructor(showChannelOutput: boolean = true) {
```
tehnrd, line 23: This is one area I debated about. 1) passing in the indicator to show the output panel as a constructor argument or 2) creating a separate instance method like `setShowChannelOutput()` called after LibraryDeploySourcePathExecutor is instantiated.

Option 2 would make the ` new SfCommandlet<string[]>(` command below a little more complicated as you'd have to instantiate ` LibraryDeploySourcePathExecutor` call the method to set the output, then pass the object to the `SfCommandlet` constructor.

Open to thoughts on either approach.

mingxuanzhangsfdx:
The implementation looks good to me. @rsensharma would you like to take a look at the proposal?

brian-krynitsky:
@tehnrd I was recently looking for a way to turn off the auto focus on the output panel and came across this PR, thank you for contributing to this. Question about the behavior if you have the bottom panel hidden...assuming there are no errors in the save, does the panel stay hidden all together?

klouf-threshold:
This would be a huge QOL improvement. I would suggest 2 changes:

1. Apply to all deploy operations (or have options for both), not just push/deploy-on-save.
2. Have the option be configurable to suppress the panel open/switch as a select list with options, instead of a boolean:
    a. Always
    b. Success only
    c. Failure only
    d. Never

Personally (and I imagine there's a decent overall user split) I don't use deploy-on-save so this actually wouldn't benefit me as currently implemented. Additionally (and I would venture a guess that I'm probably not the only one here, although admittedly maybe more likely in the minority here), I don't use the Problems view since I have a chaotic workspace with dozens of files open and therefore a Problems tab with hundreds of warnings/errors at any given time. I actually didn't even realize that deployment errors were surfaced there 😆 . Now that I know that, I could potentially transition to use it instead of the Output tab, though.

tl;dr love it, but would like to see more flexibility/options 😄

tehnrd:
@klouf-threshold, I am definitely open to adding that, but I would like someone from Salesforce to provide some input on what they would like to see.

rsensharma:
> The implementation looks good to me. @rsensharma would you like to take a look at the proposal?

@mingxuanzhangsfdx As a first step, this looks good. We can add more flexibility in the future.

rsensharma:
@tehnrd - We currently don't have the bandwidth to work on this. Your initial use case/implementation looks fine. We can get that added. For adding flexibility and the other use cases, I can provide more clear requirements, if you are up for to do the development. Let me know. fyi - @mingxuanzhangsfdx

mingxuanzhangsfdx (approved):
Looks good! Thanks for the enhancement. If you would like to support more flexibility for the feature, please continue the discussion on Github Issue and tag me or @rsensharma .

tehnrd:
Amazing, thanks all. Unfortunately, I don't have the bandwidth to expand on this either, but if someone opens a new issue and tags me, I might be able to get to it someday.

PR #5932, opened 2024-10-31
ci: checks for bundling

mingxuanzhangsfdx:
### What does this PR do?
1. externalizes `jest-editor-support` in bundle of lwc because part of its artifacts cannot be bundled. 
2. disallows dynamic imports in codebase
3. detects errors such as unaware unbundleable dependencies during bundling in each commit.

@W-16304508@

### Functionality Before
Lack of bundling checks, so new changes might break bundling process with no awareness.

### Functionality After
More bundling checks in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "vscode:bundle": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension",
     "vscode:bundle:debugger": "lerna run bundle:debugger",
     "vscode:bundle:extension": "lerna run bundle:extension",
-    "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
+    "vscode:package": "node ./scripts/capture-esbuild-warnings.js && lerna run vscode:package --concurrency 1 && npm run reformat",
     "vsix:install": "find ./packages -name '*.vsix' -exec code --install-extension {} \\;",
     "vsix:install-insiders": "find ./packages -name '*.vsix' -exec code-insiders --install-extension {} \\;",
     "vscode:sha256": "node ./scripts/generate-sha256.js >> ./SHA256",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -96,7 +96,7 @@
     "vscode:bundle": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension",
     "vscode:bundle:debugger": "lerna run bundle:debugger",
     "vscode:bundle:extension": "lerna run bundle:extension",
-    "vscode:package": "npm run vscode:bundle:debugger && npm run vscode:bundle:extension && lerna run vscode:package --concurrency 1 && npm run reformat",
```
mingxuanzhangsfdx, line 99: running the command vscode:bundle wrapped within capture-esbuild-warnings.js to report any unbundle-able dependencies during bundling.

packages/salesforcedx-vscode-lwc/esbuild.config.js
```diff
@@ -19,13 +19,20 @@ const sharedConfig = {
     'applicationinsights',
     '@salesforce/lightning-lsp-common',
     '@salesforce/lwc-language-server',
-    '@babel/preset-typescript/package.json'
+    '@babel/preset-typescript/package.json',
+    'jest-editor-support'
```
mingxuanzhangsfdx, line 23: some artifacts in this dependency are not able to be bundled. It was never externalized because the extension still worked fine even if esbuild reports warnings every time. But I think we should move this out from now on.
image

mingxuanzhangsfdx: the extensions size of lwc increases from 10.2MB to 12.8MB by externalizing `jest-editor-support`

packages/salesforcedx-vscode-lwc/package.json
```diff
@@ -31,7 +31,7 @@
     "@salesforce/salesforcedx-utils-vscode": "62.3.1",
     "ajv": "6.12.6",
     "applicationinsights": "1.0.7",
-    "jest-editor-support": "^30.1.0",
```
mingxuanzhangsfdx, line 34: I pinned the version since 30.3.1 is the latest minor of v30. And the latest major is 32.0.1-beta so there will not be any update with v30.

scripts/capture-esbuild-warnings.js
mingxuanzhangsfdx, line 1: This script detects any bundling errors including unbundle-able dependencies among the extensions.

peternhale (approved):

PR #5937, opened 2024-11-04
fix: bump min v scode version to 1.90.0 for node 20

peternhale:
@W-16702202@

Bump vscode min version to 1.90.0 to bring in node 20 runtime via Electron v29

[End to End Tests - Redhat](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/11746802243) ✅ 
[End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/11746803912) ✅


mingxuanzhangsfdx (changes requested):
Thanks for the work! with eslint rules being extracted outside, I think the eslint-related dev dependencies defined in pjson file in each sub-repo can also be removed. Could you try to remove those dependencies as well and see if it could work?

mingxuanzhangsfdx (approved):
Looks good!

PR #5939, opened 2024-11-07
feat: add org shape and dev hub id to telemetry

mingxuanzhangsfdx:
### What does this PR do?
This PR adds the org shape and dev hub id field to the telemetry including customEvents and exceptions.

@W-17148345@

### Functionality Before
No fields of org shape and dev hub id.

### Functionality After
fields of org shape will show up if the project is connected to an org, and dev hub id will show up if the org is scratch.


packages/salesforcedx-vscode-core/src/util/authInfo.ts
```diff
@@ -104,6 +104,25 @@ export class OrgAuthInfo {
     );
   }
 
+  public static async isASandboxOrg(username: string): Promise<boolean> {
+    const authInfo = await AuthInfo.create({ username });
+
+    const authInfoFields = authInfo.getFields();
+    return authInfoFields.instanceUrl?.includes('sandbox') || false;
```
mingxuanzhangsfdx, line 111: This is the best way I found to identify a sandbox. I have tried other solutions such as [AuthInfo.isSandBox](https://forcedotcom.github.io/sfdx-core/types/org_authInfo.AuthFields.html#__type.isSandbox) and [Org.isSandbox()](https://forcedotcom.github.io/sfdx-core/classes/org_org.Org-1.html#isSandbox) but they did not work as expected. I found the sandbox instanceUrl always works for me.

packages/salesforcedx-vscode-core/src/context/workspaceContext.ts
```diff
@@ -26,6 +29,7 @@ export class WorkspaceContext {
     const workspaceContextUtil = WorkspaceContextUtil.getInstance();
     this.onOrgChange = workspaceContextUtil.onOrgChange;
     this.onOrgChange(this.handleCliConfigChange);
+    this.onOrgChange(this.handleOrgShapeChange);
```
mingxuanzhangsfdx, line 32: registered a subscription here in the workspaceContext initialization when core starts up.

packages/salesforcedx-vscode-core/src/util/authInfo.ts
peternhale: @mingxuanzhangsfdx what do you mean by did not work as expected? Searching URLs for sandbox is not guaranteed over time.

packages/salesforcedx-vscode-core/src/context/workspaceOrgShape.ts
peternhale, line 1: @mingxuanzhangsfdx this file should be in utils-vscode.

Also consider using a union type instead of enum.

I know it is here due to proximity to workspace utils, so maybe move the type to utils-vscode and retain logic here.

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/util/authInfo.ts
mingxuanzhangsfdx: the isSandbox returns false when I literally uses a sandbox org.

CristiCanizales (approved):
LGTM! I was able to verify new data in appinsights ✅

packages/salesforcedx-vscode-core/src/util/authInfo.ts
peternhale: The sandbox property is a bit complicated. These CLI commands
```
org create sandbox                │ Create a sandbox org.                                                                                                                                                                                
org delete sandbox               │ Delete a sandbox.                                                                                                                                                                                    
org refresh sandbox              │ Refresh a sandbox org using the sandbox name.                                                                                                                                                        
org resume sandbox     
```
and when managing sandboxes the property is only set when these commands are used.

Another option that might be used in combination with isSandbox is to query the [Organization](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_organization.htm) in the org in question.

It has a field called isSandbox, which is set during sandbox creation.

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/context/workspaceOrgShape.ts
```diff
@@ -0,0 +1,26 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { OrgShape } from '@salesforce/salesforcedx-utils-vscode';
+import { OrgAuthInfo, workspaceUtils } from '../util';
+
+
+export const getOrgShape = async (username: string): Promise<OrgShape> => {
+  if (workspaceUtils.hasRootWorkspace()) {
+    if (await OrgAuthInfo.isAScratchOrg(username)) {
```
mingxuanzhangsfdx, line 14: check if it is a scratch org first since the implementation of isASandboxOrg return true when it is a scratch org.

peternhale (approved):

PR #5950, opened 2024-11-18
feat: create apex action from selected method/this class

CristiCanizales:
### What does this PR do?
- Creates new Central Class for Apex Action Creation
- Creates shell for Metadata Orchestrator
- Adds new command SFDX: Create Apex Action from Selected Method
- Adds new command SFDX: Create Apex Action from this Class

@ W-17236284@, @W-17236576@, @W-17236448@, @W-17236508@


packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -0,0 +1,198 @@
...
+    const lines = document.getText().split('\n');
+    const metadataList: MethodMetadata[] = [];
+    let currentMethodSignature = '';
+    let isAuraEnabled = false;
+
+    for (let line of lines) {
+      line = line.trim();
+
+      // Detect @AuraEnabled annotation
+      if (line.includes('@AuraEnabled')) {
```
mingxuanzhangsfdx, line 111: I think there is no need to capture the method metadata here. What we need is the file URI and location (if available), and the language server will return the method document symbol with the eligibility to the client side.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    const methodMetadata = this.parseMethodSignature(methodSignature, isAuraEnabled);
+    if (!methodMetadata.isAuraEnabled) {
+      throw new Error(nls.localize('not_aura_enabled', methodMetadata.name));
+    }
+    return methodMetadata;
+  };
+
+  public extractAllMethodsMetadata = (): MethodMetadata[] | undefined => {
+    const editor = vscode.window.activeTextEditor;
```
mingxuanzhangsfdx, line 90: I think there is a use case that the active text editor is not available when we right click on one apex source file in the file explorer.

packages/salesforcedx-vscode-apex/src/commands/createApexAction.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export const createApexActionFromMethod = async (): Promise<void> => {
+  // Call Controller
+  await controller.createApexAction(false);
+};
+
+/**
+ * Creates Apex Actions from all methods in the current class.
+ */
+export const createApexActionFromClass = async (): Promise<void> => {
```
mingxuanzhangsfdx, line 24: As per my understanding, there should be an optional parameter `sourceUri: vscode.Uri | vscode.Uri[] | undefined,`. There will be a value when you select a file from the file explorer.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
CristiCanizales: Yes, but we still don't have any of that logic so I thought it'd be nice to keep this one until then to be able to demo if needed

packages/salesforcedx-vscode-apex/src/commands/createApexAction.ts
CristiCanizales: True, I forgot for this command we can right click from file explorer

package.json
```diff
@@ -57,6 +57,7 @@
     "markdown-link-check": "^3.9.3",
     "ncp": "^2.0.0",
     "nyc": "15.1.0",
+    "openapi-types": "12.1.3",
```
peternhale, line 60: @CristiCanizales please verify the 3pp status of the new modules

peternhale (commented):
@CristiCanizales aside from the comments from Mingxuan and the 3pp check, very nice job of laying the foundation of the work that follows!

mingxuanzhangsfdx (approved):
Looks good!

PR #5958, opened 2024-11-22
Mz/metadata eligibility

mingxuanzhangsfdx:

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
@@ -6,7 +6,14 @@
  */
 import { notificationService } from '@salesforce/salesforcedx-utils-vscode';
 import * as vscode from 'vscode';
+import { languageClientUtils } from '../languageUtils';
 import { nls } from '../messages';
+import {
+  ApexClassOASEligibleRequest,
+  ApexClassOASEligibleResponses,
+  ApexOASEligiblePayload
+} from '../openApiUtilities/schemas';
+import { getTelemetryService } from '../telemetry/telemetry';
 
 /**
  * Interface representing the metadata of a method.
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
         return 'string';
     }
   };
+
+  private eligibilityDelegate = async (
+    requests: ApexOASEligiblePayload,
+    maxRetries: number = 3
+  ): Promise<ApexClassOASEligibleResponses> => {
+    const telemetryService = await getTelemetryService();
+    let response = {};
+    let attempt = 0;
+    const languageClient = languageClientUtils.getClientInstance();
+    if (languageClient) {
+      while (attempt < maxRetries) {
+        try {
+          attempt++;
+          response = await languageClient?.sendRequest('apexoas/isEligible', requests);
+          telemetryService.sendEventData('isEligibleResponseSucceeded', {
+            classNumbers: requests.payload.length.toString()
+          });
+          break;
+        } catch (error) {
+          telemetryService.sendException('isEligibleResponseFailed', `${error} after trying ${attempt} times.`);
+
+          if (attempt >= maxRetries) {
+            // fallback TBD after we understand it better
+            throw new Error(nls.localize('cannot_get_apexoaseligibility_response'));
+          }
+        }
+      }
+    }
+    return response as ApexClassOASEligibleResponses;
+  };
+
+  public validateEligibility = async (
+    sourceUri: vscode.Uri | vscode.Uri[],
+    isMethodSelected: boolean = false
+  ): Promise<ApexClassOASEligibleResponses> => {
+    const telemetryService = await getTelemetryService();
+    if (Array.isArray(sourceUri)) {
+      throw new Error('We do not consider list of src files now');
+    }
+    let cursorPosition;
+    if (isMethodSelected) {
+      const editor = vscode.window.activeTextEditor;
+      if (editor && this.getFileExtension(editor.document) === 'cls') {
+        cursorPosition = editor.selection.active;
+      } else {
+        telemetryService.sendException('activeTextEditorNotApex', nls.localize('active_text_editor_not_apex'));
+        throw new Error(nls.localize('invalid_active_text_editor'));
+      }
+    }
+    // generate the payload
+    const request: ApexClassOASEligibleRequest = {
+      resourceUri: sourceUri.toString(),
+      includeAllMethods: !isMethodSelected,
+      includeAllProperties: !isMethodSelected,
+      positions: cursorPosition ? [cursorPosition] : null,
+      methodNames: [],
+      propertyNames: []
+    };
+    const responses = await this.eligibilityDelegate({ payload: [request] });
+    return responses;
+  };
+
+  private getFileExtension(document: vscode.TextDocument): string {
+    const filePath = document.fileName; // Full file path
+    return filePath.substring(filePath.lastIndexOf('.') + 1).toLowerCase(); // Extract the file extension
+  }
 }
--- a/packages/salesforcedx-vscode-apex/src/languageServer.ts
+++ b/packages/salesforcedx-vscode-apex/src/languageServer.ts
@@ -106,8 +106,7 @@ const createServer = async (extensionContext: vscode.ExtensionContext): Promise<
       args.push(
         '-Dtrace.protocol=false',
         `-Dapex.lsp.root.log.level=${LANGUAGE_SERVER_LOG_LEVEL}`,
-        `-agentlib:jdwp=transport=dt_socket,server=y,suspend=${SUSPEND_LANGUAGE_SERVER_STARTUP ? 'y' : 'n'
-        },address=*:${JDWP_DEBUG_PORT},quiet=y`
+        `-agentlib:jdwp=transport=dt_socket,server=y,suspend=${SUSPEND_LANGUAGE_SERVER_STARTUP ? 'y' : 'n'},address=*:${JDWP_DEBUG_PORT},quiet=y`
       );
       if (process.env.YOURKIT_PROFILER_AGENT) {
         if (SUSPEND_LANGUAGE_SERVER_STARTUP) {
--- a/packages/salesforcedx-vscode-apex/src/languageServer.ts
+++ b/packages/salesforcedx-vscode-apex/src/languageServer.ts
@@ -55,6 +55,37 @@ const createServer = async (extensionContext: vscode.ExtensionContext): Promise<
       .getConfiguration()
       .get<boolean>('salesforcedx-vscode-apex.advanced.enable-completion-statistics', false);
 
+    // Configurations of the definitions of eligible apex classes/methods/properties
+    const classAccessModifiers: string[] = vscode.workspace
+      .getConfiguration()
+      .get<string[]>('salesforcsalesforcedx-vscode-apex.apexoas.eligibility.class.access-modifiers', ['public']);
+
+    const classDefinitionModifiers: string[] = vscode.workspace
+      .getConfiguration()
+      .get<
+        string[]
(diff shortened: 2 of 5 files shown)


packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
...
+      includeAllProperties: !isMethodSelected,
+      positions: cursorPosition ? [cursorPosition] : null,
+      methodNames: [],
+      propertyNames: []
+    };
+    const responses = await this.eligibilityDelegate({ payload: [request] });
+    return responses;
+  };
+
+  private getFileExtension(document: vscode.TextDocument): string {
```
peternhale, line 282: @mingxuanzhangsfdx this function is not needed.
It can be replaced with endsWith function call.

@CristiCanizales Why don't we condition the context and the this class commands using when property, then the supporting code can only run when active editor is a cls file.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
...
+  public eligibilityDelegate = async (
+    requests: ApexOASEligiblePayload,
+    maxRetries: number = 3
+  ): Promise<ApexClassOASEligibleResponses> => {
+    const telemetryService = await getTelemetryService();
+    let response = {};
+    let attempt = 0;
+    const languageClient = languageClientUtils.getClientInstance();
+    if (languageClient) {
+      while (attempt < maxRetries) {
```
peternhale, line 230: I am still not convinced that retries are necessary.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
...
+        try {
+          attempt++;
+          response = await languageClient?.sendRequest('apexoas/isEligible', requests);
+          telemetryService.sendEventData('isEligibleResponseSucceeded', {
```
peternhale, line 234: What other telemetry info is worth capturing here?

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
...
+    }
+    return response as ApexClassOASEligibleResponses;
+  };
+
+  public validateEligibility = async (
+    sourceUri: vscode.Uri | vscode.Uri[],
+    isMethodSelected: boolean = false
+  ): Promise<ApexClassOASEligibleResponses> => {
+    const telemetryService = await getTelemetryService();
+    if (Array.isArray(sourceUri)) {
```
peternhale, line 256: Why not just add support for multiples? If the plumbing is already present then hooking the command up to do multiples becomes even simpler.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +217,70 @@ export class MetadataOrchestrator {
...
+        telemetryService.sendException('activeTextEditorNotApex', nls.localize('active_text_editor_not_apex'));
+        throw new Error(nls.localize('invalid_active_text_editor'));
+      }
+    }
+    // generate the payload
+    const request: ApexClassOASEligibleRequest = {
+      resourceUri: sourceUri.path,
+      includeAllMethods: !isMethodSelected,
+      includeAllProperties: !isMethodSelected,
+      positions: cursorPosition ? [cursorPosition] : null,
```
peternhale, line 274: This must be an array or null, since all requests are going to the same endpoint. Java is strongly typed and does not allow union types. The other choice would be to change the position to only a single value. Let's discuss further.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -55,6 +55,37 @@ const createServer = async (extensionContext: vscode.ExtensionContext): Promise<
       .getConfiguration()
       .get<boolean>('salesforcedx-vscode-apex.advanced.enable-completion-statistics', false);
 
+    // Configurations of the definitions of eligible apex classes/methods/properties
```
peternhale, line 58: Look into passing these values as part of LanguageClientOptions section initializationOptions, just like you did for feature enableSynchronizedInitJobs.

packages/salesforcedx-vscode-apex/src/languageServer.ts
```diff
@@ -89,6 +118,15 @@ const createServer = async (extensionContext: vscode.ExtensionContext): Promise<
 
     args.push(APEX_LANGUAGE_SERVER_MAIN);
 
+    args.push(
```
peternhale, line 121: I would prefer values passed via initializationOptions

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
CristiCanizales: Ohh. I just realized the when clause is for right clicking in file explorer and editor but it's missing for command palette. Can you update that in this PR @mingxuanzhangsfdx ? It's line 261 in package.json

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
mingxuanzhangsfdx: Yes I still feel that way. Let me remove it first and see how it will work.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
mingxuanzhangsfdx: Regarding the request payload, I think maybe we can capture if the command is for folders or just one class/classes. But I am not sure if we want to capture the info here compared to the command dispatch function, since the request does not have to be successful. But I will add that here first.

I don't think we need any info worth capturing in the response, since the info is about the eligibility of each method. not about the use case.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
mingxuanzhangsfdx: Got it. I still cannot figure out what use case leads to multiple cursor positions, since there is only one cursor.

peternhale: @mingxuanzhangsfdx I am on the fence with this too.

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -50,8 +58,8 @@ export class MetadataOrchestrator {
    */
   public extractMethodMetadata = (): MethodMetadata | undefined => {
     const editor = vscode.window.activeTextEditor;
-    if (!editor) {
-      notificationService.showErrorMessage(nls.localize('no_active_editor'));
+    if (!editor || !editor.document.fileName.endsWith('cls')) {
```
CristiCanizales, line 61: The command won't show up anyways if it's not a .cls file so I think we don't need this

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -102,8 +110,8 @@ export class MetadataOrchestrator {
       lines = fileText.split('\n');
     } else {
       const editor = vscode.window.activeTextEditor;
-      if (!editor) {
-        notificationService.showErrorMessage(nls.localize('no_active_editor'));
+      if (!editor || !editor.document.fileName.endsWith('cls')) {
```
CristiCanizales, line 113: same here

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -16,6 +16,7 @@
  * If omitted, we will assume _message.
  */
 export const messages = {
+  active_text_editor_not_apex: 'The active text editor is not on an apex class file',
```
CristiCanizales, line 19:
```suggestion
  active_text_editor_not_apex: 'The active text editor is not an Apex Class file',
```

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -25,6 +26,7 @@ export const messages = {
   apex_language_server_failed_activate: 'Unable to activate the Apex Language Server',
   apex_language_server_quit_and_restarting: 'Apex Language Server has stopped. Restarting… %d of 5',
   cannot_determine_workspace: 'Unable to determine workspace folders for workspace',
+  cannot_get_apexoaseligibility_response: 'Fail to get response through apexoas/isEligible from apex language server',
```
CristiCanizales, line 29:
```suggestion
  cannot_get_apexoaseligibility_response: 'Failed to get response through apexoas/isEligible from Apex Language Server',
```

packages/salesforcedx-vscode-apex/src/messages/i18n.ts
```diff
@@ -104,5 +105,6 @@ export const messages = {
   terminate_processes_title: 'Terminate Orphaned Processes',
   orphan_process_advice:
     "The list of processes below are Apex Language Server instances that didn't properly shutdown. These\nprocesses can be stopped from the warning message that brought you here, or you can handle this\ntask yourself. If you choose to terminate these processes yourself, refer to relevant documentation\nto stop these processes.",
-  unknown: 'Unknown'
+  unknown: 'Unknown',
+  invalid_active_text_editor: 'The active text editor is missing or on an invalid file.'
```
CristiCanizales, line 109:
```suggestion
  invalid_active_text_editor: 'The active text editor is missing or is an invalid file.'
```

CristiCanizales (changes requested):

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
mingxuanzhangsfdx: Agreed but just for double check in case extractMethodMetadata is used in other commands in future.

CristiCanizales, line 61:
```suggestion
    if (!editor || !editor.document.fileName.endsWith('.cls')) {
```

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
CristiCanizales, line 113:
```suggestion
      if (!editor || !editor.document.fileName.endsWith('.cls')) {
```

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -210,4 +218,89 @@ export class MetadataOrchestrator {
...
+          positions: null,
+          propertyNames: []
+        } as ApexClassOASEligibleRequest;
+        requests.push(request);
+      }
+    } else {
+      let cursorPosition;
+      if (isMethodSelected) {
+        const editor = vscode.window.activeTextEditor;
+        if (editor && editor.document.fileName.endsWith('cls')) {
```
CristiCanizales, line 272:
```suggestion
        if (editor && editor.document.fileName.endsWith('.cls')) {
```

CristiCanizales (changes requested):

CristiCanizales (approved):
:shipit:

PR #5962, opened 2024-11-30
feat: shell for deploy and retrieve oas

CristiCanizales:
### What does this PR do?
- Creates structure for deploying/retrieving oas files

@W-17341339@, @W-17341336@


packages/salesforcedx-vscode-core/package.json
```diff
@@ -194,15 +194,15 @@
       "editor/context": [
         {
           "command": "sf.retrieve.current.source.file",
-          "when": "sf:project_opened && resourceLangId != 'forcesourcemanifest' && sf:has_target_org"
+          "when": "sf:project_opened && resourceLangId != 'forcesourcemanifest' && resourceLangId != 'yaml' && sf:has_target_org"
```
mingxuanzhangsfdx, line 197: From my understanding, you need to define the resourceLangId yaml [here](https://github.com/forcedotcom/salesforcedx-vscode/blob/feat/apex-oas/packages/salesforcedx-vscode-core/package.json#L1005) in order to use it.

CristiCanizales: I believe it's not needed. I tested it just like this and works fine

mingxuanzhangsfdx: Thanks for the confirmation!

packages/salesforcedx-vscode-core/package.json
```diff
@@ -194,15 +194,15 @@
...
         },
         {
           "command": "sf.retrieve.in.manifest",
           "when": "sf:project_opened && resourceLangId == 'forcesourcemanifest' && sf:has_target_org"
         },
         {
           "command": "sf.deploy.current.source.file",
-          "when": "sf:project_opened && resourceLangId != 'forcesourcemanifest' && sf:has_target_org"
+          "when": "sf:project_opened && resourceLangId != 'forcesourcemanifest' && resourceLangId != 'yaml' && sf:has_target_org"
```
peternhale, line 205: @CristiCanizales I am starting to get the feeling that the introduction of a new deployable resource (yaml files) is starting to splinter the cohesion of the notion o deploy/retrieve.

Do we really need separate commands to do a deploy/retrieve for or should we use a single command and then let the command's impl dispatch the deploy/retrieve according to what was picked.

CristiCanizales: I don't think the user will tell the difference. It's like with manifest files, there's also a separate command to deploy their contents. User only sees one deploy command in each of the different categories (stuff in force-app, xml files and yml files) so there's no confusion and in the end they're in fact just deploying to org

peternhale: I understand your point and it has worked well for us. I am just wondering if there might be a way to reduce command proliferation.

Maybe not a good discussion to have now.

peternhale (approved):

PR #5969, opened 2024-12-06
Mz/disable apex ls error telemetry

mingxuanzhangsfdx:
# The PR needs to be with an updated jorje jar. Do NOT merge until jar is updated.
### What does this PR do?
This PR exposes a user setting that user can choose to turn on the telemetry for error reporting from Apex Language server. The setting is off by default, so error telemetry is not reported to save the storage on azure.

@W-17380991@

### Functionality Before
Apex Language Server reports telemetry about errors.

### Functionality After
Apex Language Server does not report error telemetry by default, and user can toggle the setting on for us to understand the error.


peternhale (commented):
@mingxuanzhangsfdx why so many files in the PR?

peternhale (approved):

PR #5976, opened 2024-12-13
chore: eligibility criteria for class and method annotations

mingxuanzhangsfdx:
### What does this PR do?
Update the configurations for class and method annotations.

@W-17429259@

### Functionality Before
there were only general annotations, which does not fit the need.

### Functionality After
as above


CristiCanizales (commented):
Looks good! But I see the jar is not included in the PR, is that on purpose?

peternhale (approved):

PR #5988, opened 2024-12-31
feat: gather context

CristiCanizales:
### What does this PR do?
- Updates jar and creates util to gather context

@W-17260440@


packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
```diff
@@ -140,6 +160,7 @@ export class MetadataOrchestrator {
       requests.push(request);
     }
 
+    const response = this.gatherContext(sourceUri);
```
peternhale, line 163: @CristiCanizales I was thinking that context gathering would occur after eligibility is determined and the class is eligible for generation.

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/commands/metadataOrchestrator.ts
CristiCanizales: Yes! Right now it's just like a placeholder so we can play with the response when doing prompt engineering

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -44,10 +45,14 @@ export class ApexActionController {
           if (!metadata) {
```
peternhale, line 45: @CristiCanizales can we find a better name than "metadata"?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -44,10 +45,14 @@ export class ApexActionController {
...
             throw new Error(nls.localize('extraction_failed', type));
```
peternhale, line 46: This label name makes me think of going to the dentist ;). A different name related to the operation just performed?

peternhale (approved):
@CristiCanizales just a couple of var/label changes around eligibility. Could you do the noted changes in your next PR?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
CristiCanizales: lol. Sure

PR #5993, opened 2025-01-06
feat: create/update an ESR record with generated OAS spec

CristiCanizales:
### What does this PR do?
- Modifies vscode apex extension to create/update an ESR record with generated OAS spec
- Prompts user to select/type named credential for ESR record

@W-17482259@, @W-17543634@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
@@ -10,6 +10,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { URL } from 'url';
 import * as vscode from 'vscode';
+import { create } from 'xmlbuilder2';
 import { parse, stringify } from 'yaml';
 import { workspaceContext } from '../context';
 import { nls } from '../messages';
@@ -139,23 +140,7 @@ export class ApexActionController {
     }
     const namedCredential = await this.showNamedCredentialsQuickPick();
 
-    const updatedContent = existingContent
-      ? existingContent.replace(/<schema>([\s\S]*?)<\/schema>/, `<schema>${oasSpec.replaceAll('"', '&apos;')}</schema>`)
-      : [
-          '<?xml version="1.0" encoding="UTF-8"?>',
-          '<ExternalServiceRegistration xmlns="http://soap.sforce.com/2006/04/metadata">',
-          `\t<description>${path.basename(fullPath).split('.')[0]} External Service</description>`,
-          `\t<label>${path.basename(fullPath).split('.')[0]}</label>`,
-          `\t<namedCredentialReference>${namedCredential}</namedCredentialReference>`,
-          '\t<registrationProviderType>Custom</registrationProviderType>',
-          `\t<schema>${oasSpec.replaceAll('"', '&apos;')}</schema>`,
-          '\t<schemaType>OpenApi3</schemaType>',
-          '\t<schemaUploadFileExtension>yaml</schemaUploadFileExtension>',
-          `\t<schemaUploadFileName>${path.basename(fullPath).split('.')[0].toLowerCase()}_openapi</schemaUploadFileName>`,
-          '\t<status>Complete</status>',
-          '\t<systemVersion>5</systemVersion>',
-          '</ExternalServiceRegistration>'
-        ].join('\n');
+    const updatedContent = this.buildESRXml(existingContent, fullPath, namedCredential, oasSpec);
     try {
       // Step 3: Write File
       fs.writeFileSync(fullPath, updatedContent);
@@ -217,7 +202,7 @@ export class ApexActionController {
   };
   private showNamedCredentialsQuickPick = async (): Promise<string | undefined> => {
     let namedCredentials;
-    let selectedNamedCredential;
+    let selectedNamedCredential: string | undefined;
     let finalNamedCredential: string | undefined;
     try {
       const rawQueryData = await (
@@ -239,7 +224,7 @@ export class ApexActionController {
       });
     }
 
-    if (selectedNamedCredential === nls.localize('enter_new_nc')) {
+    if (!selectedNamedCredential || selectedNamedCredential === nls.localize('enter_new_nc')) {
       finalNamedCredential = await vscode.window.showInputBox({
         prompt: nls.localize('enter_nc_name')
       });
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -52,17 +54,30 @@ export class ApexActionController {
...
+          // Step 3: Determine filename
           name = isClass
             ? path.basename(eligibilityResult.resourceUri, '.cls')
             : eligibilityResult?.symbols?.[0]?.docSymbol?.name;
-          const openApiFileName = `${name}_openapi.yml`;
+          const openApiFileName = `${name}.externalServiceRegistration-meta.xml`;
+          let fullPath;
+          // Step 4: Check if the file already exists
+          try {
+            fullPath = await this.pathExists(openApiFileName);
```
peternhale, line 65: @CristiCanizales if we are going to move forward throwing an error if the file already exists, please log a new work item to handle overwrite/cancel.

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +130,127 @@ export class ApexActionController {
...
+
+  private saveOasAsErsMetadata = async (oasSpec: string, fullPath: string): Promise<void> => {
+    // Replace the schema section in the ESR file if it already exists
+    let existingContent;
+    if (fs.existsSync(fullPath)) {
+      existingContent = fs.readFileSync(fullPath, 'utf8');
+    }
+    const namedCredential = await this.showNamedCredentialsQuickPick();
+
+    const updatedContent = existingContent
```
peternhale, line 142: @CristiCanizales Would this be better handled using a XML document, rather than doing all of this using string manipulation?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +130,127 @@ export class ApexActionController {
...
+    const response = await vscode.window.showWarningMessage(
+      nls.localize('file_exists'),
+      { modal: true },
+      nls.localize('overwrite'),
+      nls.localize('cancel')
+    );
+    return response === nls.localize('overwrite');
+  };
+
+  private getFolderForArtifact = async (): Promise<string | undefined> => {
```
peternhale, line 203: Is it not possible to get the default location for ESR from SDR?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +130,127 @@ export class ApexActionController {
...
     return stringify(parse(theDoc));
   }
+
+  private saveOasAsErsMetadata = async (oasSpec: string, fullPath: string): Promise<void> => {
+    // Replace the schema section in the ESR file if it already exists
+    let existingContent;
+    if (fs.existsSync(fullPath)) {
+      existingContent = fs.readFileSync(fullPath, 'utf8');
+    }
+    const namedCredential = await this.showNamedCredentialsQuickPick();
```
peternhale, line 140: @CristiCanizales I see that `showNamedCredentialsQuickPick` throws an error if something goes sideways on the query. What happens if there is not a default org available? Should we issue a message and then ask the user to enter a named credential?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
CristiCanizales: I think we can using [src/registry/metadataRegistry.json](https://github.com/forcedotcom/source-deploy-retrieve/blob/7a75847af83777a2673165c909672857c32f90e8/src/registry/metadataRegistry.json#L273) ? I'll look at it! Thanks!

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +132,173 @@ export class ApexActionController {
...
+      registryAccess.getTypeByName('ExternalServiceRegistration').directoryName
+    );
+    const folderUri = await vscode.window.showInputBox({
+      prompt: nls.localize('enter_esr_path'),
+      value: defaultESRFolder
+    });
+
+    return folderUri ? path.resolve(folderUri) : undefined;
+  };
+  private showNamedCredentialsQuickPick = async (): Promise<string | undefined> => {
```
peternhale, line 206: Should we consider the use case where the named cred only exist in the workspace?

We can follow up with a new work item to handle finding named creds in the workspace.

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +132,173 @@ export class ApexActionController {
...
+      finalNamedCredential = await vscode.window.showInputBox({
+        prompt: nls.localize('enter_nc_name')
+      });
+    } else {
+      finalNamedCredential = selectedNamedCredential;
+    }
+    return finalNamedCredential;
+  };
+
+  private flattenQueryRecords(rawQueryRecords: JsonMap[]) {
```
peternhale, line 240: Why is this necessary?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +132,173 @@ export class ApexActionController {
...
+      .txt('yaml')
+      .up()
+      .ele('schemaUploadFileName')
+      .txt(`${baseName}_openapi`)
+      .up()
+      .ele('status')
+      .txt('Complete')
+      .up()
+      .ele('systemVersion')
+      .txt('5')
```
peternhale, line 298: According to the ESR [spec](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_externalserviceregistration.htm?q=external) this should be in the range of 1-3, not sure which but let's start with 3

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -132,4 +132,173 @@ export class ApexActionController {
...
+      if (schemaElement) {
+        schemaElement.txt(safeOasSpec);
+      } else {
+        throw new Error('The <schema> element was not found in the provided XML.');
+      }
+
+      return doc.end({ prettyPrint: true });
+    }
+
+    const newDoc = create({ version: '1.0', encoding: 'UTF-8' })
```
peternhale, line 268: As we discussed yesterday, the `xmlbuilder2` module is fairly large, > 1MB. What other features does this offer aside from using it to build the xml doc for ESR?

peternhale (changes requested):
Mostly questions and one small change

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
CristiCanizales: Actually it is not. Thanks!

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
peternhale: @CristiCanizales let leave it as is for now. Maybe revisit the xml building in Feb.

peternhale (approved):

PR #6003, opened 2025-01-10
fix: hide commands outside of forceapp main default

CristiCanizales:
### What does this PR do?
- Makes the constraints in package.json stricter so that `SFDX: Delete This from Project and Org`, `SFDX: Deploy This Source to Org`, `SFDX: Diff File Against Org` and `SFDX: Retrieve This Source from Org` commands don't show up in the context menu for files that aren't meant to be deployed to the org.

@W-17568713@

### Functionality After
image

### DIFF
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -241,7 +241,7 @@
         },
         {
           "command": "sf.folder.diff",
-          "when": "explorerResourceIsFolder && sf:project_opened && sf:has_target_org"
+          "when": "explorerResourceIsFolder && sf:project_opened && sf:has_target_org && resourcePath =~ /force-app\\/main\\/default/"
         },
         {
           "command": "sf.analytics.generate.template",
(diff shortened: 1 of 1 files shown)


daphne-sfdc (approved):
Approved! 🎉 

Commands no longer appear in context menu in both Editor View and Explorer View for files and folders outside `forceapp/main/default` ✅

aheber:
I know this was just merged in. I'm concerned about this change and maybe I don't actually understand it. Not all projects use `force-app` as a project directory and `main/defalut` are optional/recommended folders structures only. The SDR library is much more flexible in what folder structures it supports.

If you can't check the `sfdx-project.json` file for registered folders then how do you know what the folder structure of this project is? Is it intentional that those commands only work for projects that haven't customized?

Here is one example project that probably defines each of the expectations of this PR, lots of different directories and different folder structures for different purposes. https://github.com/SFDO-Community/declarative-lookup-rollup-summaries

daphne-sfdc:
Hi Anthony, Thanks a lot for the comment and for addressing your concerns.  For the time being, a PR has been created to revert this change: #6004.

aheber:
Thanks! Sorry to just drop in. Really appreciate the quick response.

PR #6017, opened 2025-01-22
Madhur/w 17619393

madhur310:
feat: adding spectral library and validation based on OAS3 spec

### What does this PR do?
adding spectral library and validation based on OAS3 spec, note currently this just logs the validation errors. We still need a way to display it to user.

[@W-17619393@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE0000285LGMYA2/view)

### Functionality Before
No OAS specific validation existed.

### Functionality After
Validations are happening on OAS docs generated by LLM.


package.json
```diff
@@ -35,7 +38,7 @@
     "@vscode/test-electron": "2.3.0",
     "@vscode/vsce": "2.21.1",
     "acorn": "8.8.2",
-    "ajv": "6.12.6",
+    "ajv": "8.17.1",
```
peternhale, line 41: @madhur310 we need to check other packages to see if bumping ajv to current version. Unit tests are on the light side across the repo

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -69,7 +69,7 @@
     "lint:fix": "npm run lint -- --fix",
     "watch": "tsc -watch -p .",
     "clean": "shx rm -rf node_modules && cd out && node ../../../scripts/clean-all-but-jar.js && shx rm -rf coverage && shx rm -rf .nyc_output",
-    "copy:files": "shx cp -R ../../node_modules/@salesforce/apex-tmlanguage/grammars .",
+    "copy:files": "shx cp -R ../../node_modules/@salesforce/apex-tmlanguage/grammars . && cp ./src/commands/*.yaml ./out/src/commands/",
```
peternhale, line 72: @madhur310 why is this change necessary?

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -114,6 +120,18 @@ export class ApexActionController {
     telemetryService.sendException(telemetryEvent, errorMessage);
   };
 
+  private async validateOpenApiDocument(openapiDocument: string) {
```
peternhale, line 123:
``` suggestion
  private async validateOpenApiDocument(openapiDocument: string): Promise<void> {
```

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -114,6 +120,18 @@ export class ApexActionController {
...
+    const spectral = new Spectral();
+
+    const ruleset = await bundleAndLoadRuleset(join(__dirname, './ruleset.spectral.yaml'), { fs, fetch });
```
peternhale, line 126: @madhur310 building paths using `__dirname` is an anti-pattern for bundling. @mingxuanzhangsfdx can provide advice on how to remediate this.

peternhale (changes requested):

packages/salesforcedx-vscode-apex/package.json
madhur310: The ruleset.yaml file needs to be pulled in and bundled for spectral to find it. yaml files are not copied by default to `out` directory, the other way I could find was to use a webpack - this seemed like the easier path to make sure .yaml file is present in the `out` directory where the code is running from..

package.json
```diff
@@ -23,6 +23,9 @@
     "@octokit/core": "4.2.4",
     "@octokit/webhooks-definitions": "^3.67.3",
     "@salesforce/dev-config": "^3.1.0",
+    "@stoplight/spectral-core": "1.19.4",
+    "@stoplight/spectral-functions": "1.9.3",
```
mingxuanzhangsfdx, line 27: where are the two dependencies used? @stoplight/spectral-functions and @stoplight/spectral-ruleset-bundler?

package.json
```diff
@@ -23,6 +23,9 @@
     "@octokit/core": "4.2.4",
     "@octokit/webhooks-definitions": "^3.67.3",
     "@salesforce/dev-config": "^3.1.0",
+    "@stoplight/spectral-core": "1.19.4",
```
mingxuanzhangsfdx, line 26: If the dependency is only used in apex, I think we just include it in its package.json. The package.json in root is mostly for the whole monorepo

packages/salesforcedx-vscode-apex/src/commands/ruleset.spectral.yaml
mingxuanzhangsfdx, line 1: Is this file still needed?

mingxuanzhangsfdx (changes requested):
There is also something wrong with CI due to incompatibility of dependencies and node version.

package.json
madhur310: I added it in rev2 per Peter's recommendation, to keep pin the dependency to specific version so 1. package size doesn't get bloated 2. no odd behavior.

package.json
madhur310: ah my miss, I don't need these any more.

packages/salesforcedx-vscode-apex/src/commands/ruleset.spectral.yaml
madhur310: no I thought I have deleted it, will remove it.

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
madhur310: Updated the code to use typescript based ruleset instead of yaml, after discussing with @mingxuanzhangsfdx

peternhale (approved):

PR #6018, opened 2025-01-22
Mz/apply generation strategy

mingxuanzhangsfdx:
### What does this PR do?
make some adjustments before we can develop strategies.

@W-17587223@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
@@ -4,6 +4,8 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+/* eslint-disable prettier/prettier */
+
 import { notificationService, WorkspaceContextUtil, workspaceUtils } from '@salesforce/salesforcedx-utils-vscode';
 import { RegistryAccess } from '@salesforce/source-deploy-retrieve-bundle';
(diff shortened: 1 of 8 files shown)


packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -68,13 +70,12 @@ export class ApexActionController {
           // Step 4: Check if the file already exists
           const fullPath = await this.pathExists(openApiFileName);
           if (!fullPath) throw new Error(nls.localize('full_path_failed'));
-          // Step 5: Generate OpenAPI Document
-          progress.report({ message: nls.localize('generate_openapi_document') });
-          const openApiDocument = await this.generateOpenAPIDocument(eligibilityResult, context);
-
-          // Step 6: Write OpenAPI Document to File
-          progress.report({ message: nls.localize('write_openapi_document_to_file') });
-          await this.saveOasAsErsMetadata(openApiDocument, fullPath);
```
CristiCanizales, line 77: can we leave this here and not make it part of the strategy? Because I think the storing is independent of the strategy

CristiCanizales: It also looks like all of this will change in [W-17587198](https://gus.lightning.force.com/a07EE000027w9p8YAA), so let's leave it as is now.

CristiCanizales (approved):
🚀

PR #6041, opened 2025-01-28
feat: put OAS doc in separate YAML file if sfdx-project.json contains "decomposeExternalServiceRegistrationBeta"

daphne-sfdc:
### What does this PR do?
1. Reads the sfdx-project.json file of the Salesforce project for the "decomposeExternalServiceRegistrationBeta" sourceBehaviorOption
2. If the setting is present, then separate the ESR into 2 files: (1) the OpenAPI v3 specification in the YAML file (2) everything else in the XML file
a. After the generation, both the YAML file and the XML file are opened in the Editor View
3. If the setting is not present, then keep the OpenAPI v3 specification under the \<schema\> tag in the ESR XML file

@W-17670044@, @W-17684701@


packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -87,8 +88,18 @@ export class ApexActionController {
               'vscode.diff',
               vscode.Uri.file(fullPath[0]),
               vscode.Uri.file(fullPath[1]),
-              'Manual Diff of ESR Files'
+              'Manual Diff of ESR XML Files'
```
peternhale, line 91: @daphne-sfdc can you move all of this diff code into a separate function

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -390,4 +447,34 @@ export class ApexActionController {
 
     return operations;
   };
+
+  /**
+   * Reads sfdx-project.json and checks if decomposeExternalServiceRegistrationBeta is enabled.
+   * @returns boolean - true if sfdx-project.json contains decomposeExternalServiceRegistrationBeta
+   */
+  private isESRDecomposed = async (): Promise<boolean> => {
```
CristiCanizales, line 455: I see you call this in multiple places. Can you call it just once in createApexAction and then just pass it along?

peternhale (changes requested):
Works in both composed and decomposed. Just a few asks.

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -35,6 +36,7 @@ export class ApexActionController {
     let context;
     let name;
     const telemetryService = await getTelemetryService();
+    const isESRDecomposed = await this.isESRDecomposed();
```
peternhale, line 39: @daphne-sfdc make it a field in the class, then you will not have to change method signatures to pass the value around.

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -83,12 +90,16 @@ export class ApexActionController {
...
-              'vscode.diff',
-              vscode.Uri.file(fullPath[0]),
-              vscode.Uri.file(fullPath[1]),
-              'Manual Diff of ESR Files'
-            );
+            this.openDiffFile(fullPath[0], fullPath[1], 'Manual Diff of ESR XML Files');
+
+            // If sfdx-project.json contains decomposeExternalServiceRegistrationBeta, also open a diff for the YAML OAS docs
+            if (this.isESRDecomposed) {
+              this.openDiffFile(
```
peternhale, line 97: @daphne-sfdc be explicit dealing with calls that return promises
```suggestion
              void this.openDiffFile(
```

peternhale (changes requested):

peternhale (approved):

PR #6057, opened 2025-02-05
fix: exclude jsonc-parser from apex bundling

peternhale:
@W-17751384@
exclude jsonc-parser from apex ext bundle


daphne-sfdc (approved):
Approved! 🎉 

I installed the Apex extension from the VSIX in the Commit Workflow.  The extension is able to activate in my Salesforce project with no errors. ✅

PR #6071, opened 2025-02-10
feat: determine operationId and request path

mingxuanzhangsfdx:
### What does this PR do?
This PR makes operationId and request path deterministic.

The request path will always be either UrlMapping + parameters in path (RestResource cls for tdx) or ClassName + parameters in path (for after tdx).
- required params are always ahead of optional params

The operationId will always be present and the same as the corresponding method name in the apex class

@W-17775794@
@W-17777277@

### Functionality Before
operationId and request path are generated by LLM, which might not make sense.

### Functionality After
operationId and request path are deterministic by the rule above.

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/oas/generationStrategy/generationStrategy.ts
+++ b/packages/salesforcedx-vscode-apex/src/oas/generationStrategy/generationStrategy.ts
@@ -15,11 +15,11 @@ import {
 export abstract class GenerationStrategy {
   abstract metadata: ApexClassOASEligibleResponse;
   abstract context: ApexClassOASGatherContextResponse;
-  abstract prompts: string[];
+  abstract prompts: string[] | Map<string, string>; // Map<methodName, prompt>
   abstract strategyName: string;
   abstract callCounts: number;
   abstract maxBudget: number;
-  abstract llmResponses: string[];
+  abstract llmResponses: string[] | Map<string, string>; // Map<methodName, yaml response>
   abstract bid(): PromptGenerationStrategyBid;
   abstract generate(): PromptGenerationResult; // generate the prompt(s) to be sent to the LLM
   abstract callLLMWithPrompts(): Promise<string[]>;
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-apex/src/oas/generationStrategy/generationStrategy.ts
```diff
@@ -15,11 +15,11 @@ import {
 export abstract class GenerationStrategy {
   abstract metadata: ApexClassOASEligibleResponse;
   abstract context: ApexClassOASGatherContextResponse;
-  abstract prompts: string[];
+  abstract prompts: string[] | Map<string, string>; // Map<methodName, prompt>
```
peternhale, line 18: Changing the type in the base class forces subclasses to adopt that change as well. If a given strategy requires a different shape, and are only used within the subclass, then I would suggest moving strategy specific need to each subclass and remove these proeprties from the base class.

packages/salesforcedx-vscode-apex/src/oas/generationStrategy/methodByMethodStrategy.ts
```diff
@@ -38,26 +39,104 @@ export class MethodByMethodStrategy extends GenerationStrategy {
...
+                  optionalParameters.push(param.name);
+                }
+              }
+            });
+          }
+        }
+      }
+    }
+    parametersArray.push(...requiredParameters, ...optionalParameters);
+    return parametersArray;
```
peternhale, line 125:
```suggestion
   return (
      JSONPath<OpenAPIV3.ParameterObject[]>({
        path: '$..parameters[?(@.in=="path")]',
        json: oas
      })
        // sort by param.required, with true coming first
        .sort((param1, param2) => {
          return param1.required === param2.required ? 0 : param1.required ? -1 : 1;
        })
        // extract parameter.name
        .map(param => param.name)
    );
```

mingxuanzhangsfdx: This is so elegant🙂 updated!

peternhale (approved):

PR #6079, opened 2025-02-13
fix: more tolerance for errors during parsing LLM response

mingxuanzhangsfdx:
### What does this PR do?
- generating as many methods as possible based on legit LLM responses rather than interrupting generation when there is a failed response.
- dealing with corner cases of LLM response

@W-17813723@

### Functionality Before
When there is a failed response, OAS doc will not be generated and the error message cannot be understood. (we are using method-by-method strategy).

### Functionality After
When there is a failed response, OAS doc will still be generated based on the legit responses, and missing methods from failed responses can be noticed to user in the Problems tab.

Parsing OpenAPI doc is more tolerant since all characters before "openapi" in the response is removed.


mingxuanzhangsfdx:
image
If there is a failed response, the OAS doc will be generated based on the legit responses. One can tell the missing method in Problems tab and check the details of the error in developer console.

PR #6085, opened 2025-02-14
[W-17764181] fix: log generation strategy stream

peternhale:
@W-17764181 
## Add local capture of generation strategy stream 
When SF_LOG_LEVEL is set to any other value than `fatal` the OAS generation is captured to local directory `.llmLogs`

Add internal config for strategy selection
Added internal config to set generation strategy


daphne-sfdc (approved):
Thanks @peternhale for adding the logging! 🥇 

Just a few nits and then it's good to merge.

PR #6095, opened 2025-02-17
[W-17830514] fix: adjust prompts and handlers for json response

peternhale:
- Modify base prompts to request generated OAS docs be provided in JSON format
- Add setting to indicate that OAS schema should be included as guided json in requests
- Modify call to LLM to include guided json, if configured to do so

@W-17830514@


packages/salesforcedx-vscode-apex/src/oas/generationStrategy/generationStrategy.ts
```diff
@@ -22,6 +24,13 @@ export abstract class GenerationStrategy {
   abstract generate(): PromptGenerationResult; // generate the prompt(s) to be sent to the LLM
   abstract callLLMWithPrompts(): Promise<string[]>;
   abstract generateOAS(): Promise<string>; // generate OAS with the generated prompt(s)
+  openAPISchema: string;
+  includeOASSchema: boolean | undefined;
+
+  constructor() {
+    this.openAPISchema = JSON.stringify(openAPISchema_v3_0, undefined, 2);
```
mingxuanzhangsfdx, line 31: I am curious if we still need spaces given limited token size.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
```diff
@@ -12,14 +12,26 @@ export const isNullOrUndefined = (object: any): object is null | undefined => {
...
   }
-  throw new Error(`The string "${str}" is not a valid JSON string.`);
+  const jsonString = extractJsonString(str);
+  if (!jsonString) {
+    throw new Error(`The string "${str}" is undefined.`);
+  }
+  return JSON.parse(jsonString) as Record<string, unknown>;
+};
+
+export const isJsonString = (str: string): boolean => {
```
mingxuanzhangsfdx, line 26: I am a bit worried about if it is a sufficient check. It only checks if brackets exist in the response, what if a non-json text contains brackets? 

How about "{" must be the first char, and "}" must be the last char if it is a legit json?

packages/salesforcedx-vscode-apex/src/oasUtils.ts
```diff
@@ -77,3 +78,17 @@ export const checkIfESRIsDecomposed = async (): Promise<boolean> => {
 
   return false;
 };
+
+export const cleanupGeneratedDoc = (doc: string): string => {
```
mingxuanzhangsfdx, line 82: This only deals with the json response, but looks missing our previous yaml response.

packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
```diff
@@ -116,17 +118,4 @@ export class PromptGenerationOrchestrator {
   getWholeClass(bids: Map<GenerationStrategy, PromptGenerationStrategyBid>): GenerationStrategy {
     return GenerationStrategy.WHOLE_CLASS;
   }
-
```
mingxuanzhangsfdx, line 119: If we have not ruled out our original yaml response, I don't think we need to remove functions. Or the PR should not be merged until we finally have good quality with the new model.

mingxuanzhangsfdx (commented):
Thanks for the quick and huge change for the incoming new model. Since we want to make comparison between incoming and current designs with the vscode config setting, I think we do not want to get rid of the original prompt and parsing logics.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
madhur310: or use something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse within a try catch.

packages/salesforcedx-vscode-apex/src/oasUtils.ts
```diff
@@ -77,3 +78,17 @@ export const checkIfESRIsDecomposed = async (): Promise<boolean> => {
...
+  if (isJsonString(doc)) {
+    const json = extractJsonString(doc);
```
madhur310, line 84: nit: technically we don't need the second if right? extractJsonString will always return a non-undefined as it also checks for isJsonString?

packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
```diff
@@ -116,17 +118,4 @@ export class PromptGenerationOrchestrator {
...
-  private cleanupYaml(doc: string): string {
```
madhur310, line 120: should we hide this behind a FF or some setting flag to toggle between the two?

packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/propertyCorrectionStep.ts
```diff
@@ -10,11 +10,11 @@ import { ProcessorInputOutput, ProcessorStep } from './processorStep';
 
 export class PropertyCorrectionStep implements ProcessorStep {
   process(input: ProcessorInputOutput): Promise<ProcessorInputOutput> {
-    let fixedYaml = this.ensureServersIsPresent(input.yaml);
+    let fixedYaml = this.ensureServersIsPresent(input.openAPIDoc);
```
madhur310, line 13: nit: `fixedYaml` -> `fixedOASDoc` ?

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
peternhale: @mingxuanzhangsfdx @madhur310

Possibilities:
`{...` isJsonString check returns false
`...}` isJsonString check returns false
`...`  isJsonString check returns false
`}...{` isJsonString check returns true, this can be fixed by adding first index < last index
`{...}` isJsonString check returns true
`some text{...}more text` isJsonString check returns true

Given that a string can contain anything, I am not sure what else can be done. I can change the name to containsJsonString and add a cautionary note.

The other choice is to remove the function and inline the check where needed.

packages/salesforcedx-vscode-apex/src/oas/generationStrategy/generationStrategy.ts
peternhale: I can load w/o.

packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
peternhale: Yes this definitely needs to be thoroughly vetted as the solution moving forward.

packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
peternhale: Not sure how the new model will land, but I understand your concern about abandoning the previous  generation result type.

packages/salesforcedx-vscode-apex/src/oasUtils.ts
peternhale: I think I will remove the is json check and inline the conditional in both extract methods.

packages/salesforcedx-utils-vscode/src/helpers/utils.ts
madhur310: What I was thinking was something like:

```
function isJsonString(str) {
    try {
        JSON.parse(str);
    } catch (e) {
        return false;
    }
    return true;
}
```

I know you mentioned that JSON.parse is used to actually parse so these issues would be caught later. Or we could have a safeJsonParse method, which catches odd strings and throws/logs the error without blowing up.

peternhale: @madhur310 This is the what the extract string is resolving for us. When we get a response from the LLM, be it json or yaml, it is surrounded with data that is not part of the document. Handling of the string requires that the text around the doc string must be removed before it can be parsed.

``` javascript
console.log(JSON.parse('{"a": "aa","b": "bbb"}'))
console.log(JSON.parse('```json\n{"a": "aa","b": "bbb"}\n```'));
```

mingxuanzhangsfdx (approved):
I happened to try our previous methodbymethod (without guided_json) strategy with the new model. And the response looks bad as well. yaml cannot be generated with failures. The code looks nothing wrong with me though.

packages/salesforcedx-vscode-apex/src/commands/apexActionController.ts
```diff
@@ -54,6 +54,7 @@ export class ApexActionController {
           cancellable: true
         },
         async progress => {
+          // Step0
```
daphne-sfdc, line 57: There's no additional step that is supposed to be added here, right?

```suggestion
```

packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/methodValidationStep.ts
```diff
@@ -25,10 +25,10 @@ export class MethodValidationStep implements ProcessorStep {
     }
 
     MethodValidationStep.diagnosticCollection.clear();
-    const cleanedupYaml = this.validateMethods(input.yaml, input.eligibilityResult);
+    const cleanedupYaml = this.validateMethods(input.openAPIDoc, input.eligibilityResult);
```
daphne-sfdc, line 28: Change `cleanedupYaml` to `cleanedUpOpenAPIDoc`?

mingxuanzhangsfdx (approved):
Tested with CaseManager and Class6-9, looks good!

PR #6099, opened 2025-02-18
[W-17854317] fix: diff source now works for ____.object-meta.xml files

daphne-sfdc:
@W-17854317@

### Functionality Before
**SFDX: Diff Source Against Org** does not work for ____.object-meta.xml files.  The command completes successfully but the diff does not popup in Editor View.

### Functionality After
**SFDX: Diff Source Against Org** works for ____.object-meta.xml files.  The diff pops up in Editor View like all other diffs.

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
@@ -84,7 +84,17 @@ export const sourceFolderDiff = async (explorerPath: vscode.Uri) => {
 export const handleCacheResults = async (username: string, cache?: MetadataCacheResult): Promise<void> => {
   if (cache) {
     if (cache.selectedType === PathType.Individual && cache.cache.components) {
-      await differ.diffOneFile(cache.selectedPath, cache.cache.components[0], username);
+      if (cache.selectedPath.endsWith('object-meta.xml')) {
+        // object-meta.xml files are listed as the last item in the cache.cache.components array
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-core/src/commands/sourceDiff.ts
```diff
@@ -84,7 +84,17 @@ export const sourceFolderDiff = async (explorerPath: vscode.Uri) => {
 export const handleCacheResults = async (username: string, cache?: MetadataCacheResult): Promise<void> => {
   if (cache) {
     if (cache.selectedType === PathType.Individual && cache.cache.components) {
-      await differ.diffOneFile(cache.selectedPath, cache.cache.components[0], username);
+      if (cache.selectedPath.endsWith('object-meta.xml')) {
+        // object-meta.xml files are listed as the last item in the cache.cache.components array
+        await differ.diffOneFile(
+          cache.selectedPath,
+          cache.cache.components[cache.cache.components.length - 1],
```
peternhale, line 91: @daphne-sfdc I have never looked at the cache code before, but how does one know for certain that the last entry will always be object-metadata.xml?

daphne-sfdc: That's a good question - we should dig into the cache code to know for sure.  I proposed this change based on my observations - I ran the debugger on both a standard object and a custom object and saw that the last entry was object-metadata.xml for both of them.

PR #6100, opened 2025-02-18
chore: screenshots in E2E tests specify VSCode version in filename

daphne-sfdc:
### What does this PR do?
E2E screenshots filename now includes VSCode version of the job.

[skip-validate-pr]

### Functionality Before
The nightly build and release E2E tests run on both the latest and minimum supported VSCode versions.  If the same test on the same OS fails for both VSCode versions, only one set of screenshots is produced, and the other job throws an error, because the filenames of the screenshot zip files are the same.  See this run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/13378208445

### Functionality After
E2E screenshots filename now includes VSCode version of the job.  Both jobs can produce screenshots without errors.


mingxuanzhangsfdx (approved):
good catch!

PR #6123, opened 2025-03-03
@W-17902076@ fix: fix the condition to prevent a false positive of the "paths.<method>.description is required" warning

daphne-sfdc:
### What does this PR do?
Fixes the condition to prevent a false positive of the `paths.<method>.description is required` warning.

@W-17902076@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/ruleset.spectral.ts
+++ b/packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/ruleset.spectral.ts
@@ -120,7 +120,7 @@ const ruleset = {
     },
     'paths-method-description': {
       description: 'paths.<method>.description is required',
-      given: '$.paths.*',
+      given: '$.paths[*][get,post,put,delete,patch]',
       message: 'paths.<method>.description is required',
       then: {
         field: 'description',


CristiCanizales (approved):

packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/ruleset.spectral.ts
```diff
@@ -120,7 +120,7 @@ const ruleset = {
     },
     'paths-method-description': {
       description: 'paths.<method>.description is required',
-      given: '$.paths.*',
+      given: '$.paths[*][get,post,put,delete,patch]',
```
madhur310, line 123: This changes makes this rule identical to 'operations-description'. Why is this change needed?

madhur310 (changes requested):

packages/salesforcedx-vscode-apex/src/oas/documentProcessorPipeline/ruleset.spectral.ts
peternhale: I think there is some terminology confusion happening here.

* paths - The "paths" key
* method - The key expressed by the path to the endpoint, i.e. `/a/b/{c}`
* operation - the Http method (`get,post,put,delete,patch`) to use when calling the method

Now the OpenAPI spec has different names, which, IMO, make more sense.

* paths - The "paths" key
* path - The key expressed by the path to the endpoint, i.e. `/a/b/{c}`
* operation - the Http method (`get,post,put,delete,patch`) to use when calling the method

PR #6139, opened 2025-03-10
fix(W-18002235): update org browser labels

CristiCanizales:
### What does this PR do?
- Adds missing labels to org browser

@W-18002235@

### Functionality Before
image

### Functionality After
image


daphne-sfdc (changes requested):
Please also make the same changes to `i18n.ts`.

daphne-sfdc (approved):

PR #6140, opened 2025-03-10
@W-18006149 - fix: error name is undefined after bundling in apex

mingxuanzhangsfdx:
### What does this PR do?
This PR fixes the error that OAS generation results in error: __name is undefined.
https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/13775473723

@W-18006149@

### Functionality Before
OAS function failed to generate the result with an error: a is undefined

### Functionality After
OAS function works


daphne-sfdc (approved):
Approved!  I am able to generate, validate, and deploy OAS docs without any errors. ✅ Thanks @mingxuanzhangsfdx for the fix! 🎉

peternhale (approved):

PR #6149, opened 2025-03-14
[DRAFT] feat: Apex Code Coverage - configurable `red_green` and `red_blue` color themes

AndrewStopchenko-SO:
### What does this PR do?
This is a very raw implementation of the configurable "color theme" for Apex code coverage (per discussion #5807):
- `red_green` (default): uncoveredLines: **red**, coveredLines: **green**
- `red_blue` ("Developer Console" alike): uncoveredLines: **red**, coveredLines: **blue**

This ColorTheme selection is exposed as `salesforcedx-vscode-apex.apex-code-coverage-color-theme` configuration property.

See discussion #5807.

### Functionality Before
Apex code coverage colors are static: uncoveredLines: **red**, coveredLines: **green**.

### Functionality After
Apex coverage colors can be switched between `red_green`/`red_blue` options from the Settings page.
Example `red_green` (same as current colors):
ApexCodeCoverageColorTheme_redGreen

Example `red_blue`:
ApexCodeCoverageColorTheme_redBlue

---

### Working Items ###
- [x] **FEAT:** Impl `red_green` and `red_blue` color themes, base functionality.
  - // DISCLAIMER: I don't have much experience with JS/TS programming; basically tried to mimic what already was there or around.
- [ ] **I18N:** `packages/salesforcedx-vscode-apex/package.nls.ja.json` Japanese description labels to be translated.
- [ ] **TESTS:** Impl unit tests to validate how decorations are being applied.
  - **// DISCLAIMER: I have no experience with JS/TS testing, so this is completely absent at the moment. _Any_ help is highly appreciated.**
  - Note: from a brief research, this doesn't seem like an easy task, because VSCode API does not allow retrieving currently applied decorations. [This article](https://github.com/microsoft/vscode/issues/136164#issuecomment-956027228) suggests adding an "injection pattern" for testing to inspect what TextEditorDecorationType values are passed - but I have no experience with JS/TS testing, so that is way above my skills.
- [ ] **FEAT:** Impl dynamically applying selectedColorTheme's colors when switching between `red_green` and `red_blue` options on the _Settings_ page.
  - I tried coverting `coveredLinesDecorationType`/`uncoveredLinesDecorationType` from constants into functions, so that every time the `CodeCoverageHandler.setCoverageDecorators(...)` runs, it reads the current color theme setting and returns the new TextEditorDecorationType instance with selected theme's colors.
    - It didn't work, those **dynamically-created TextEditorDecorationTypes do not get auto-disposed** when we are applying different TextEditorDecorationType (probably for [this reason](https://github.com/microsoft/vscode-extension-samples/issues/22#issuecomment-321555992)). As a result, it was impossible to toggle the decorations off, and when we switch to other color theme option, colors got applied **on top of** the previous colors - not looking good at all.
    - Because of that, I ended up building the decorations just once. This means that switching the color theme requires reloading extensions host/window.

### DIFF
--- a/packages/salesforcedx-vscode-apex/package.json
+++ b/packages/salesforcedx-vscode-apex/package.json
@@ -519,6 +519,19 @@
           "type": "number",
           "default": 750,
           "description": "%apex_generation_output_token_limit%"
+        },
+        "salesforcedx-vscode-apex.apex-code-coverage-color-theme": {
+          "type": "string",
+          "enum": [
+            "red_green",
+            "red_blue"
+          ],
+          "default": "red_green",
+          "description": "%apex_code_coverage_color_theme%",
+          "enumDescriptions": [
+            "%apex_code_coverage_color_theme-red_green_description%",
+            "%apex_code_coverage_color_theme-red_blue_description%"
+          ]
         }
       }
     },
--- a/packages/salesforcedx-vscode-apex/package.nls.ja.json
+++ b/packages/salesforcedx-vscode-apex/package.nls.ja.json
@@ -36,5 +36,8 @@
   "run_tests_title": "テストを実行",
   "show_error_title": "エラーを表示",
   "test_view_container_title": "テストビューア",
-  "test_view_name": "Apex テスト"
+  "test_view_name": "Apex テスト",
+  "apex_code_coverage_color_theme": "Select the colors theme for Apex Tests code coverage",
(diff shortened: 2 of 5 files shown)


peternhale:
@AndrewStopchenko-SO , thank for you for the contribution. We are currently examining options to solve this issue in a more general way, to avoid custom settings and code. We are focusing on how to use the selected theme/theme entries to find a combination of existing keys that make sense in this use case.

In this way, each user can select the theme of their choosing and then be able to customize the color choices even further through vscode suppled theme editors.

PR #6158, opened 2025-03-19
W-14314586: Add restart function /clean db to command palette and status bar item

kylewalke:
### What does this PR do?
Adds a restart function to the Apex Language Server via Command Palette and the ApexStatusBarItem
Includes ability to clean the apex.db and related artifacts out of the .sfdx/tooling dir if wanted. 

@W-12345678@ 

### Functionality Before
No restart ability was implemented and deleting the files manually following a manual refresh of the entire workspace was required. 

### Functionality After
feature implemented as described above.


packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
```diff
@@ -104,3 +119,137 @@ export const getExceptionBreakpointInfo = async (): Promise<{}> => {
...
+    const wsrf = vscode.workspace.workspaceFolders[0].uri;
+    const toolsUri = vscode.Uri.joinPath(wsrf, '.sfdx', 'tools');
+    try {
+      const entries = await vscode.workspace.fs.readDirectory(toolsUri);
+      const releaseFolders = entries
+        .filter(([name, type]) => type === vscode.FileType.Directory && /^\d{3}$/.test(name))
+        .map(([name]) => name);
+      for (const folder of releaseFolders) {
+        const folderUri = vscode.Uri.joinPath(toolsUri, folder);
+        await vscode.workspace.fs.delete(folderUri, { recursive: true, useTrash: true });
```
mshanemc, line 170: I'm used to having the `no-await-in-loop` rule in place so that you parallelize everything by default.  

I'm guessing these are the 252, 254 etc folders so there's not too many of them that it would matter.

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
```diff
@@ -104,3 +119,137 @@ export const getExceptionBreakpointInfo = async (): Promise<{}> => {
...
+      languageClient.errorHandler?.addListener('restarting', (count: number) => {
+        languageServerStatusBarItem.error(
+          nls.localize('apex_language_server_quit_and_restarting').replace('$N', `${count}`)
+        );
+      });
+      languageClient.errorHandler?.addListener('startFailed', () => {
+        languageServerStatusBarItem.error(nls.localize('apex_language_server_failed_activate'));
+      });
+
+      // TODO: the client should not be undefined. We should refactor the code to
```
mshanemc, line 203: yeah, I agree with this TODO.  you could  do the errors and return if no languageClient and then you don't have the giant conditonal block.

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
```diff
@@ -104,3 +119,137 @@ export const getExceptionBreakpointInfo = async (): Promise<{}> => {
...
+  } catch (e) {
+    let errorMessage = '';
+    if (typeof e === 'string') {
+      errorMessage = e;
+    } else if (e instanceof Error) {
+      errorMessage = e.message ?? nls.localize('unknown_error');
+    }
+    if (errorMessage.includes(nls.localize('wrong_java_version_text', SET_JAVA_DOC_LINK))) {
+      errorMessage = nls.localize('wrong_java_version_short');
+    }
```
mshanemc, line 231: I think that's big enough to be its own function.  `createLanguageClient` is a bit long

getErrorMessage (e:unknown):string

packages/salesforcedx-vscode-apex/package.json
```diff
@@ -200,6 +200,10 @@
         }
       ],
       "commandPalette": [
+        {
+          "command": "sf.apex.languageServer.restart",
+          "when": "resource =~ /.\\.(cls)?$/"
```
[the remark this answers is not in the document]
kylewalke: For the command to be visible in the command palette it requires that the .cls file be open but if you want to restart while selecting a different file, restarting via the LSP status bar item is always available. It is a non-blocking visibility only condition here.

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
kylewalke: Yeah I think unless there is something that's gone horribly wrong, users would have at most 2 folders (an old release and the new release) where we could blow both away and start fresh with a regenerated folder from the LSP. This shouldn't be handling more than couple at any given time.

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
```diff
@@ -1,23 +1,30 @@
 /*
- * Copyright (c) 2017, salesforce.com, inc.
+ * Copyright (c) 2025, salesforce.com, inc.
```
CristiCanizales, line 2: just a nit, we don't usually modify this for old files, but just add current year for new ones.

peternhale (approved):

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
```diff
@@ -104,3 +119,135 @@ export const getExceptionBreakpointInfo = async (): Promise<{}> => {
...
+      //delete the apex DB file if we need a clean slate
+      await removeApexDB();
+    }
+    //Need a half second timeout for the creation of a new language client so that we give the previous LS a moment to complete shutdown
+    setTimeout(createLanguageClient, 500, extensionContext, statusBarInstance);
+  }
+};
+
+const removeApexDB = async () => {
+  if (vscode.workspace.workspaceFolders) {
```
CristiCanizales, line 158: another nit, we have hasRootWorkspace() in workspaceUtils

packages/salesforcedx-vscode-apex/test/jest/__mocks__/vscode.ts
CristiCanizales, line 1: check scripts/setup-jest.ts

packages/salesforcedx-vscode-apex/src/languageUtils/languageClientUtils.ts
kylewalke: I have added it in but I have to also use the direct check for the tsconfig to be okay with me referencing it a few lines later that way.

mshanemc:
QA notes:
installed only the `-apex` vsix, working on dreamhouse

📓  I don't see the action until I open an apex class.
created 4 errors
✅  I see the action with a class open
✅  clicking `restart only` briefly clears the errors before they come back (that's how eslint and ts restarts work)
✅ do it again but `clean apex db`.  Takes a little longer.  I do see it delete `254` and its children and then recreate them
did all those from the palette.  The status bar seems to work fine, too.

tried to do evil things (create a very large number of errors across lots of file, restart the server while tests are running).  
✅ all seems cool

PR #6160, opened 2025-03-24
chore: W-17684179 - remove yaml strategies

CristiCanizales:
### What does this PR do?
- Removes methodByMethod and wholeClass YAML strategies

@W-17684179@


peternhale (changes requested):
@CristiCanizales The strategies also need to be removed from package.json settings

PR #6163, opened 2025-03-25
fix: [W-14667562] remove dups from Apex problems

peternhale:
@W-14667562@
Both LS and org report the same Apex compile issues. This removes the dups.

From:

Screenshot 2025-03-25 at 06 59 52

To:

Screenshot 2025-03-25 at 07 01 02


packages/salesforcedx-utils-vscode/test/jest/helpers/utils.test.ts
```diff
@@ -96,4 +96,52 @@ describe('utils tests', () => {
...
+      const input = '   Error occurred   ';
+      const result = fixupError(input);
+      expect(result).toBe('Error occurred');
+    });
+
+    it('should handle error messages without the line/column pattern', () => {
+      const input = 'General error';
+      const result = fixupError(input);
+      expect(result).toBe('General error');
+    });
```
daphne-sfdc, line 139: This unit test case looks to be testing the same thing as the first test case "should return an error with the correct message". Good to remove it?

daphne-sfdc (approved):
Thanks @peternhale for this fix!  Duplicate problems are no longer existing for me. ✅ 

Before:
Screenshot 2025-03-25 at 10 46 27 AM

After:
Screenshot 2025-03-25 at 10 48 01 AM

packages/salesforcedx-vscode-core/src/diagnostics/diagnostics.ts
```diff
@@ -34,48 +34,47 @@ export const handlePushDiagnosticErrors = (
...
-    const fileUri = vscode.Uri.file(defaultErrorPath);
-    const range = getRange('1', '1');
     const diagnostic = {
-      message: errors.message,
+      message: fixupError(error),
       severity: vscode.DiagnosticSeverity.Error,
-      source: errors.name,
+      source: type,
       range
     } as vscode.Diagnostic;
```
mshanemc, line 50: you can change to `const diagnostic: vscode:Diagnostic` gets rid of the type assertion and lets the type checker check.

since fixupError always returns a string, it should satisfy the type

daphne-sfdc (approved):

PR #6194, opened 2025-04-07
W-18196743: Change restart cmd to use palette instead of notification

kylewalke:
### What does this PR do?
Changes restart command to use the command palette instead of the notification window for deciding whether or not to delete the apex artifacts. 

@W-18196743@


CristiCanizales (approved):
Thanks for these fixes! 🚀 

Test notes:
- Restart only --> only restarts ✅ 
- Restart and clean --> erases DB and restarts, db is back ✅ 
- Options now show up as prompts from Cmd Palette ✅

PR #6200, opened 2025-04-09
W-18205398 Fix: fallout from changing LSP restart function

kylewalke:
Notification modal is dead to me, cmd palette quick pick is my new best friend

### What does this PR do?
Fixes the now broken E2E tests for the LSP restart since they no longer use the notification modal but instead use the quick pick bar from the command palette to decide if the artifacts will be cleaned out from the .sfdx folder or not. 

@W-18205398@


daphne-sfdc (approved):
Approved! 🎉 

Passing E2E test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/14385950134 ✅

PR #6219, opened 2025-04-16
@W-18291188 remove jsforce direct dep

mshanemc:
### What does this PR do?
remove jsforce direct dep.  It still exists as part of core-bundle.

this is a types-only change, no runtime effects should occur (verify this during code review) and "it compiled+linted" should be all the validation required.  

@W-18291188@


kylewalke (approved):
I was able to compile and run lint without any issues. LGTM

PR #6222, opened 2025-04-16
refactor: remove admzip

mshanemc:
### What does this PR do?
from what I can tell, this was used ONLY to copy dirs of test data to another location.
maybe this code is older than node16's fs.cp ?  Only good reason I could think of.  

[skip-validate-pr]


kylewalke (approved):
Nice find. Kaboot it!

PR #6233, opened 2025-04-21
chore: @W-18321869 - ignore threshold to immediately upload events

smuduganti-sf:
### What does this PR do?
This change disables the ignoreThreshold to ensure that all events for the A4D extension use case are captured and uploaded immediately, instead of waiting for the event buffer to reach the 50,000-byte threshold. In certain scenarios, this delay results in an incomplete view of events until the buffer fills up, which is not ideal for time-sensitive or low-volume use cases.

By disabling the threshold, events will be uploaded as they are reported, providing more accurate and timely visibility.

We can revisit this approach in the future and define an optimal upload frequency or threshold value based on usage patterns and performance considerations.

#<Insert GitHub Issue>, @W-18321869@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>


smuduganti-sf:
Hi @peternhale  and @mshanemc 
Following up on this PR — would appreciate your review and approval when you have a moment. Thanks!

peternhale (approved):

PR #6234, opened 2025-04-22
@W-18337323 refactor: node-prefixed imports

mshanemc:
### What does this PR do?
eslint rule to enforce the pattern
autofix all existing
update ut mocks

now we have an inventory of direct node dependencies that need to removed/bundled for an ext to work on the web

@W-18337323@


CristiCanizales (approved):
[End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/14595345780) ✅

PR #6245, opened 2025-04-28
W-18382518: eslint no-extraneous-dependencies

mshanemc:
### What does this PR do?
new rule: if you use it in /src, it must be in dependencies [unless you specify `type` on the import and it really is type-only]
if you use it only in /test it can be in dev dependencies.
remove more integration testsi

@W-18382518@

cleans up the resulting errors found.


kylewalke (approved):
lgtm!

PR #6249, opened 2025-05-01
@W-18406865 refactor: vscode-uri everywhere.  fix tests

mshanemc:
### What does this PR do?
these are types-only `import * from vscode`
so if you want to test with them, you have to mock things (they're fine at runtime)

vscode-uri is the real implementation published as a package

@W-18406865@

### DIFF
--- a/packages/salesforcedx-visualforce-markup-language-server/src/services/htmlLinks.ts
+++ b/packages/salesforcedx-visualforce-markup-language-server/src/services/htmlLinks.ts
@@ -5,27 +5,21 @@
 'use strict';
 
 import { Range, TextDocument } from 'vscode-languageserver-types';
-import Uri from 'vscode-uri';
+import { URI } from 'vscode-uri';
 import { DocumentContext, DocumentLink } from '../htmlLanguageService';
 import { createScanner, TokenType } from '../parser/htmlScanner';
 
 function stripQuotes(url: string): string {
-  return url
-    .replace(/^'([^']*)'$/, (substr, match1) => match1)
-    .replace(/^"([^"]*)"$/, (substr, match1) => match1);
+  return url.replace(/^'([^']*)'$/, (substr, match1) => match1).replace(/^"([^"]*)"$/, (substr, match1) => match1);
 }
 
 function getWorkspaceUrl(
-  modelAbsoluteUri: Uri,
+  modelAbsoluteUri: URI,
   tokenContent: string,
   documentContext: DocumentContext,
   base: string
 ): string {
-  if (
-    /^\s*javascript\:/i.test(tokenContent) ||
-    /^\s*\#/i.test(tokenContent) ||
-    /[\n\r]/.test(tokenContent)
-  ) {
+  if (/^\s*javascript\:/i.test(tokenContent) || /^\s*\#/i.test(tokenContent) || /[\n\r]/.test(tokenContent)) {
     return null;
   }
   tokenContent = tokenContent.replace(/^\s*/g, '');
@@ -57,7 +51,7 @@ function createLink(
   endOffset: number,
   base: string
 ): DocumentLink {
-  const documentUri = Uri.parse(document.uri);
+  const documentUri = URI.parse(document.uri);
   const tokenContent = stripQuotes(attributeValue);
   if (tokenContent.length === 0) {
     return null;
@@ -66,37 +60,31 @@ function createLink(
     startOffset++;
     endOffset--;
   }
-  const workspaceUrl = getWorkspaceUrl(
-    documentUri,
-    tokenContent,
-    documentContext,
-    base
-  );
+  const workspaceUrl = getWorkspaceUrl(documentUri, tokenContent, documentContext, base);
   if (!workspaceUrl || !isValidURI(workspaceUrl)) {
     return null;
   }
   return {
-    range: Range.create(
-      document.positionAt(startOffset),
-      document.positionAt(endOffset)
-    ),
+    range: Range.create(document.positionAt(startOffset), document.positionAt(endOffset)),
     target: workspaceUrl
   };
 }
 
 function isValidURI(uri: string) {
   try {
-    Uri.parse(uri);
+    // reject URIs with bare % not followed by 2 hex digits
+    // I don't know if this was a requirement, but it was in the UT
+    if (uri.includes('%') && !uri.match(/%[0-9A-Fa-f]{2}/)) {
+      return false;
+    }
+    URI.parse(uri);
     return true;
   } catch (e) {
     return false;
   }
 }
 
-export function findDocumentLinks(
-  document: TextDocument,
-  documentContext: DocumentContext
-): DocumentLink[] {
+export function findDocumentLinks(document: TextDocument, documentContext: DocumentContext): DocumentLink[] {
   const newLinks: DocumentLink[] = [];
 
   const scanner = createScanner(document.getText(), 0);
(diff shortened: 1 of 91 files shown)


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -42,7 +42,8 @@
     "@salesforce/soql-model": "1.0.0",
     "debounce": "^1.2.0",
     "papaparse": "^5.3.0",
-    "vscode-languageclient": "^9.0.1"
+    "vscode-languageclient": "^9.0.1",
+    "vscode-uri": "^3.0.8"
```
daphne-sfdc, line 46: I noticed that in some package.json files, the version of `vscode-uri` is `^3.1.0`.  Is there a reason you chose `^3.0.8` instead here?

packages/salesforcedx-visualforce-markup-language-server/src/services/htmlLinks.ts
```diff
@@ -66,37 +60,31 @@ function createLink(
...
 
 function isValidURI(uri: string) {
   try {
-    Uri.parse(uri);
+    // reject URIs with bare % not followed by 2 hex digits
+    // I don't know if this was a requirement, but it was in the UT
+    if (uri.includes('%') && !uri.match(/%[0-9A-Fa-f]{2}/)) {
+      return false;
+    }
+    URI.parse(uri);
```
daphne-sfdc, line 80:
```suggestion
    URI.parse(uri, true);
```

I went to check the [original code](https://github.com/microsoft/vscode-uri/blob/main/src/uri.ts) for the `parse()` function.  The check for `%` is already done in that function, so we don't need to add an additional check here.

The function does provide a strict mode, which defaults to **false**.  I changed it to **true** so that invalid URIs would throw an error and reach our catch block.

packages/salesforcedx-vscode-apex/src/oas/ExternalServiceRegistrationManager.ts
```diff
@@ -63,22 +71,19 @@ export class ExternalServiceRegistrationManager {
...
-    if (fs.existsSync(this.newPath)) {
-      existingContent = fs.readFileSync(this.newPath, 'utf8');
-    }
-
+    const existingContent = fs.existsSync(this.newPath) ? fs.readFileSync(this.newPath, 'utf8') : undefined;
     //Step 1: Let user choose or enter a named credential
-    if (!this.isVersionGte(orgVersion, '63.0')) namedCredential = await this.showNamedCredentialsQuickPick();
+    const namedCredential = this.isVersionGte(orgVersion, '63.0')
+      ? undefined
+      : await this.showNamedCredentialsQuickPick();
```
daphne-sfdc, line 86: This step is actually no longer needed, as all orgs are now on v63 and don't require the user to select a named credential from a picklist anymore.

packages/salesforcedx-vscode-soql/package.json
mshanemc: they resolve to the same, but yeah, I can make it more consistent.  I blame cursor 🤪

packages/salesforcedx-vscode-apex/test/jest/oas/externalServiceRegistrationManager.test.ts
```diff
@@ -93,20 +99,23 @@ describe('ExternalServiceRegistrationManager', () => {
...
       expect(esrHandler['newPath']).toBe(fullPath[1]);
     });
 
     it('should initialize with overwrite set to true when paths are the same', async () => {
-      await esrHandler['initialize'](false, processedOasResult, ['/path/to/file.xml', '/path/to/file.xml', true]);
+      await esrHandler['initialize'](false, processedOasResult, ['/path/to/file.xml', '/path/to/file.xml']);
       expect(esrHandler['overwrite']).toBe(true);
     });
+    it('should initialize with overwrite set to false when paths differ only by extension', async () => {
+      await esrHandler['initialize'](false, processedOasResult, ['/path/to/original.xml', '/path/to/new.pdf']);
```
daphne-sfdc, line 113:
```suggestion
      await esrHandler['initialize'](false, processedOasResult, ['/path/to/file.xml', '/path/to/file.pdf']);
```

The paths were originally differing by more than just the extension.

packages/salesforcedx-vscode-apex/test/jest/oas/externalServiceRegistrationManager.test.ts
```diff
@@ -144,11 +153,13 @@ describe('ExternalServiceRegistrationManager', () => {
   describe('promptNamedCredentialSelection', () => {
```
daphne-sfdc, line 153: Named credential tests are no longer needed, as the code is going to be removed.

packages/salesforcedx-vscode-apex/src/oas/ExternalServiceRegistrationManager.ts
daphne-sfdc, line 1: Don't know why this class name starts with uppercase... can you please rename it to `externalServiceRegistrationManager.ts`?

packages/salesforcedx-vscode-soql/test/jest/editor/soqlEditorProvider.test.ts
```diff
@@ -7,7 +7,8 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as vscode from 'vscode';
-import { BUILDER_VIEW_TYPE, SOQL_BUILDER_WEB_ASSETS_PATH } from '../../../src/constants';
```
daphne-sfdc, line 10: After this removal, `SOQL_BUILDER_WEB_ASSETS_PATH` is actually no longer used anywhere in the code, so the constant can be removed from **constants.ts**.

packages/salesforcedx-visualforce-markup-language-server/src/services/htmlLinks.ts
mshanemc: I like the research, but it does cause one of our previously passing UT to fail.

Maybe the UT is invalid, but I'd rather just leave it

packages/salesforcedx-vscode-apex/src/oas/ExternalServiceRegistrationManager.ts
mshanemc: good catch...let's never have this happen again

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md

fixed a few others

packages/salesforcedx-vscode-apex/src/oas/ExternalServiceRegistrationManager.ts
mshanemc: so it's not needed at all?  no named credentials are used anymore and we can delete all the other referneces to the named credential?

mshanemc: please check the logic of those changes

packages/salesforcedx-vscode-apex/src/oas/ExternalServiceRegistrationManager.ts
```diff
@@ -233,8 +186,7 @@ export class ExternalServiceRegistrationManager {
     className: string,
     safeOasSpec: string,
     operations: any,
-    orgVersion: string,
-    namedCredential: string | undefined
+    orgVersion: string
```
daphne-sfdc, line 189: `orgVersion` was only used for checking named credentials, so this parameter can now be removed.

daphne-sfdc (approved):
Approved! 🎉 

Passing E2E test run here: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/14893536221 ✅ (the failure is a known flapper)

PR #6255, opened 2025-05-07
@W-18480731 refactor: fewer assertions

mshanemc:
### What does this PR do?
remove a LOT of ` as `
eslint rule to prevent future assertions
eslint disable comments for things that are either
- going away
- to complex to mess with

some zod parsing to replace some assertions
update the find-java-home (recent versions are TS)
remove unused DiffResultParser (check this)
minSobjects.json was had errors in it.  The TS assertion was hiding them (you can import a json file and it'll be typed checked unless your assertion tells it not to)
removed a few enums (prefer string unions) to clean up type stuff

@W-18480731@


packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
daphne-sfdc, line 1: FYI I reverted Line 159 to say `if (result.response.status === RequestStatus.Succeeded) {`.  After the `===` was changed to `!==`, a successful deploy would show an error with no explanation in the Output Tab.

Screenshot 2025-05-24 at 7 56 53 AM

PR #6257, opened 2025-05-07
chore: W-18379395 - use vscode api instead of node:fs

CristiCanizales:
### What does this PR do?
- use vscode api instead of node:fs

@W-18379395@

[Core End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/14941645046). ✅


kylewalke (approved):
Nice idea to put it in the util.

packages/salesforcedx-sobjects-faux-generator/src/generator/typingGenerator.ts
```diff
@@ -20,33 +20,31 @@ export class TypingGenerator implements SObjectGenerator {
...
+  public async generateTypes(sobjects: SObject[], targetFolder: string): Promise<void> {
+    if (!(await folderExists(targetFolder))) {
+      await createDirectory(targetFolder);
     }
 
     for (const sobj of sobjects) {
       if (sobj.name) {
         const sobjDefinition = this.declGenerator.generateSObjectDefinition(sobj);
-        this.generateType(targetFolder, sobjDefinition);
+        await this.generateType(targetFolder, sobjDefinition);
```
mshanemc, line 36: why do we do these on at a time?

packages/salesforcedx-sobjects-faux-generator/src/generator/typingGenerator.ts
```diff
@@ -20,33 +20,31 @@ export class TypingGenerator implements SObjectGenerator {
...
+  public async generateType(folderPath: string, definition: SObjectDefinition): Promise<string> {
     const typingPath = path.join(folderPath, `${definition.name}${TYPESCRIPT_TYPE_EXT}`);
-    if (fs.existsSync(typingPath)) {
-      fs.unlinkSync(typingPath);
+    if (await folderExists(typingPath)) {
+      await deleteFile(typingPath);
     }
 
-    fs.writeFileSync(typingPath, this.convertDeclarations(definition), {
-      mode: 0o444
```
mshanemc, line 48: was this important for some reason?  Is there another way to ensure that these are read-only?

CristiCanizales: I tried workarounds but looks like there's no way to create read-only files using vscode api

PR #6259, opened 2025-05-08
[W-18338281] refactor: turn off apex debug log for replay debugger without calling the CLI command

daphne-sfdc:
### What does this PR do?
Refactors the code for **SFDX: Turn Off Apex Debug Log for Replay Debugger** to not call the CLI command.

@W-18338281@


packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -1,90 +1,63 @@
...
-    if (resultJson.status === 0) {
-      return Promise.resolve();
-    } else {
-      return Promise.reject('Restoring the debug levels failed.');
-    }
-  }
-};
+    console.log('Developer log trace flag is active');
+    try {
+      await vscode.window.withProgress(
```
mshanemc, line 22: I don't know the UI patterns well enough to know the right choice...but does this still take long enough that you'd want a progress indicator, especially when it's not going to do anything?

packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -1,90 +1,63 @@
...
-    execution.processExitSubject.subscribe(async data => {
-      this.logMetric(execution.command.logName, startTime);
-      if (data !== undefined && String(data) === '0') {
-        developerLogTraceFlag.turnOffLogging();
-        hideTraceFlagExpiration();
-      }
-    });
-  }
-}
+const command = 'SFDX: Turn Off Apex Debug Log for Replay Debugger';
```
mshanemc, line 12: there is a nls message for this string that should be used

packages/salesforcedx-utils-vscode/src/helpers/traceFlagsRemover.ts
```diff
@@ -50,4 +50,8 @@ export class TraceFlagsRemover {
       }
     }
   }
+
+  public async removeTraceFlag(traceFlagId: string) {
```
mshanemc, line 54: I don't think there's much value of having this as a method on the class since it's not interacting with the rest of the class's stuff.

export const removeTraceFlag (conn: Connection, traceFlagId:string) and then use that.  then you don't have to go through the whole getInstance() just to do the simple thing.

while we're in here, I see `removeNewTraceFlags` is unused.

mshanemc: actually, a lot from this looks unused.

packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -1,90 +1,63 @@
...
+          location: vscode.ProgressLocation.Notification,
+          title: command,
+          cancellable: false
+        },
+        async progress => {
+          progress.report({ message: 'Running...' });
+          const nonNullTraceFlag = developerLogTraceFlag.getTraceFlagId()!;
+          const connection = await WorkspaceContext.getInstance().getConnection();
+          await TraceFlagsRemover.getInstance(connection).removeTraceFlag(nonNullTraceFlag);
+          telemetryService.sendCommandEvent('stop_apex_debug_logging');
```
mshanemc, line 33: do we think we'll probably always have telemetry in these?  Maybe the stuff you have in channelUtils should also have telemetry in it?

packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -1,90 +1,63 @@
...
+      );
+    } catch (error) {
+      console.error('Error in turnOffLogging(): ', error);
+      telemetryService.sendException('stop_apex_debug_logging', error);
+      await handleFinishCommand(
+        channelService,
+        command,
+        false
+      );
+      return Promise.reject('Restoring the debug levels failed.');
```
mshanemc, line 50: instead of returning a reject, what happens if we throw a new Error?  As is, it's the equivalent of throwing a string instead of an error.

mshanemc (changes requested):

mshanemc (approved):

PR #6262, opened 2025-05-10
chore: W-18379537 - use vscode api instead of fs in lwc package

CristiCanizales:
### What does this PR do?
- Updates code to use vscode api instead of fs in lwc package

@W-18379537@

Test run:

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
+++ b/packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
@@ -14,8 +14,8 @@ import {
   toMinimalSObject
 } from '@salesforce/salesforcedx-sobjects-faux-generator';
 import { projectPaths } from '@salesforce/salesforcedx-utils-vscode';
-import * as fs from 'node:fs';
 import * as path from 'node:path';
+import * as vscode from 'vscode';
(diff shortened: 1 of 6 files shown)


packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
```diff
@@ -47,11 +47,26 @@ export class FileSystemOrgDataSource implements OrgDataSource {
     const standardsFolder = path.join(soqlMetadataPath, STANDARDOBJECTS_DIR);
 
     const files: string[] = [];
-    if (fs.existsSync(standardsFolder)) {
-      files.push(...(await fs.promises.readdir(standardsFolder)));
+    try {
+      const standardsUri = vscode.Uri.file(standardsFolder);
+      const standardsStat = await vscode.workspace.fs.stat(standardsUri);
```
mshanemc, line 52: what's the value of the stat here?  Won't the `catch` block handle the "doesn't exist" scenario if readdir fails?

CristiCanizales: You're right

mshanemc (approved):

PR #6264, opened 2025-05-11
chore: W-18379555 - use vscode api instead of fs in utils and utils-vscode packages

CristiCanizales:
### What does this PR do?
- Updates code to use vscode api instead of fs in utils and utils-vscode packages

@W-18379555@

Test run: [End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15146587562) ✅


packages/salesforcedx-apex-debugger/package.json
```diff
@@ -42,7 +42,7 @@
     "vscode-uri": "^3.1.0"
   },
   "scripts": {
-    "bundle:debugger": "esbuild ./src/adapter/apexDebug.ts  --bundle --outfile=dist/apexdebug.js --format=cjs --platform=node --minify",
+    "bundle:debugger": "esbuild ./src/adapter/apexDebug.ts  --bundle --outfile=dist/apexdebug.js --format=cjs --platform=node --minify --external:vscode",
```
mshanemc, line 45: what made the `external` necessary here?  isn't `vscode` just a bunch of Types?

CristiCanizales: `npm run vscode:package` was failing without it, but I'm not sure if it's ok to use it

packages/salesforcedx-vscode-lwc/src/testSupport/workspace/getLwcTestRunnerExecutable.ts
```diff
@@ -38,23 +36,17 @@ export const getLwcTestRunnerExecutable = async (cwd: string) => {
     });
     if (lwcTestRunnerExecutable) {
       try {
-        const uri = vscode.Uri.file(lwcTestRunnerExecutable);
-        const stat = await vscode.workspace.fs.stat(uri);
-        if (stat) {
-          return lwcTestRunnerExecutable;
-        }
+        await stat(lwcTestRunnerExecutable);
```
mshanemc, line 39: while we're here, look through this code to see if there's any way to make it less duplicative.

ex: can we 
1. get the executable (using a conditional for its path)
2. throw if not present
3. return

and maybe there's at the beginning to check for `lwc_test_no_lwc_testrunner_found` ?

packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
```diff
@@ -79,20 +76,12 @@ export class FileSystemOrgDataSource implements OrgDataSource {
 
     let filePath = path.join(soqlMetadataPath, STANDARDOBJECTS_DIR, sobjectName + '.json');
     try {
-      const fileUri = vscode.Uri.file(filePath);
-      const fileStat = await vscode.workspace.fs.stat(fileUri);
-      if (!fileStat) {
-        filePath = path.join(soqlMetadataPath, CUSTOMOBJECTS_DIR, sobjectName + '.json');
-      }
-    } catch {
+      await stat(filePath);
```
mshanemc, line 79: is this also an existence check?  

I like try/catch for handling errors, but this seems to be more of a "may or may not be there, fine either way" and I like your `fileOrFolderExists` with if/else for that.

or are we more concerned about `catch` for the "readFile might not work"?

If it's that, maybe just call readFile and let the catch handle it not existing?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -384,7 +384,7 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
   // commands are run with the project path returned from process.cwd(),
   // thus avoiding the potential errors surfaced when the libs call
   // process.cwd().
-  ensureCurrentWorkingDirIsProjectPath(rootWorkspacePath);
+  void ensureCurrentWorkingDirIsProjectPath(rootWorkspacePath);
```
mshanemc, line 387: maybe avoid void

packages/salesforcedx-vscode-apex-replay-debugger/src/adapter/debugConfigurationProvider.ts
```diff
@@ -52,7 +52,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     }
 
     if (vscode.workspace && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders[0]) {
-      config.projectPath = vscode.workspace.workspaceFolders[0].uri.fsPath;
+      config.projectPath = vscode.workspace.workspaceFolders[0].uri.toString();
```
mshanemc, line 55: check that `toString` is correct, and not `path` or `fsPath`

mshanemc (approved):

PR #6292, opened 2025-05-19
feat: [@W-18497015@,@W-18499135@] add restart setting to apex language restart command

peternhale:
Modifies the current Apex language server restart command to retain how the user wants to run the restart; with an Apex db cleanup before restart or without the cleanup.

The setting has been added to apex extension with name "Restart Behavior." 

The setting has three choices:
- prompt (current behavior and default for backwards compatibility)
- restart - restart the language server with no other intervention.
- reset - remove the existing artifacts found in `.sfdx/tools/<version>` folder. These artifacts are installed into a workspace during startup and are not already present.

Current behavior is preserved via default setting value of `prompt`. This is to preserve backwards compatibility.

There is a slight difference in behavior when starting the restart from the status bar versus the command palette. This difference is due to the ability to change the text in the status bar dynamically. 

When started via the status bar the command will always follow the setting:
- setting of `prompt` will result in the choices in the top input box, with "Restart only", selected in the list
- a setting of `restart` will result in the Apex language server restarting immediately
- a setting of `reset` will result in the removal of the folder ".sfdx/tools/252" and then Apex language server restarting

When running the Apex language server restart command from the command palette, the user will always be prompted for their choice. The prompt will contain the same values when running with setting set to "prompt"; "Restart Only" or "Clean Apex DB and Restart" with highlighted entry aligned with the setting;

When `restart` is the setting, the order of the prompt will have "Restart Only" first and the value will be selected. When  the setting is reset, the order of the prompt will have "Clean Apex DB and Restart" first and the value will be selected. This allows the user to simply hit enter to accept the restart that matches their setting.

@W-18497015@ restart setting
@W-18499135@ restart telemetry


packages/salesforcedx-vscode-apex/package.nls.ja.json
```diff
@@ -7,6 +7,11 @@
   "apex_oas_general_method_access-modifiers": "The general eligibility of method access modifiers for the generation of Apex Open API Specs.",
   "apex_oas_general_property_access-modifiers": "The general eligibility of property access modifiers for the generation of Apex Open API Specs.",
   "apex_log_get_text": "SFDX: Apex デバッグログを取得",
+  "apex_language_server_restart": "SFDX: Apex 言語サーバを再起動",
+  "apex_language_server_restart_behavior_description": "Apex 言語サーバの再起動コマンドの動作を制御します。",
```
kylewalke, line 11: naisu

kylewalke (approved):
Looks good to me

PR #6316, opened 2025-05-29
fix: [@W-18656877@] require find-java-home instead of import

peternhale:
@W-18656877@

using import for find-java-home is causing bundling induced runtime errors


daphne-sfdc (approved):
In my Windows VM, when no `JAVA_HOME` value is set ✅ 
Screenshot 2025-05-29 at 10 10 52 AM

daphne-sfdc (approved):

PR #6318, opened 2025-05-30
[W-18589337] refactor: turn on apex debug log for replay debugger without calling the CLI command

daphne-sfdc:
### What does this PR do?
Refactors the code for **SFDX: Turn On Apex Debug Log for Replay Debugger** to not call the CLI command.

@W-18589337@


packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
```diff
@@ -4,100 +4,64 @@
...
-      }).execute(this.cancellationToken);
-      this.attachSubExecution(execution);
-      const resultPromise = new CommandOutput().getCmdResult(execution);
-      const result = await resultPromise;
-      return JSON.parse(result);
+    const expirationDate = new Date(Date.now() + 30 * 60 * 1000); // 30 minutes from now
+    const traceFlag = {
+      TracedEntityId: await getUserId(connection),
+      LogType: 'DEVELOPER_LOG',
+      StartDate: '',
```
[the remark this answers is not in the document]
mshanemc: rather than pass `''` you can just omit it, right?

packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -15,11 +15,20 @@ const command = 'stop_apex_debug_logging';
 export const turnOffLogging = async (): Promise<void> => {
   handleStartCommand(command);
 
-  if (developerLogTraceFlag.isActive()) {
+  const connection = await WorkspaceContext.getInstance().getConnection();
+
+  // Check if a DebugLevel with DeveloperName 'ReplayDebuggerLevels' already exists
+  const replayDebuggerLevels = await connection.tooling.query(
+    "SELECT id, logtype, startdate, expirationdate, debuglevelid, debuglevel.apexcode, debuglevel.visualforce FROM TraceFlag WHERE logtype='DEVELOPER_LOG'"
```
peternhale: @daphne-sfdc no need to return all of these named fields if all that is needed is "id"
``` suggestion
    "SELECT id FROM TraceFlag WHERE logtype='DEVELOPER_LOG'"
```

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
```diff
@@ -112,131 +86,10 @@ export const getUserId = async (projectPath: string): Promise<string> => {
...
-          this.userId
-        }' logtype=developer_log debuglevelid=${developerLogTraceFlag.getDebugLevelId()} StartDate='' ExpirationDate='${developerLogTraceFlag
-          .getExpirationDate()
-          .toUTCString()}`
-      )
-      .withArg('--use-tooling-api')
-      .withJson()
-      .withLogName('create_trace_flag')
-      .build();
+  const result = await connection.singleRecordQuery(`SELECT Id FROM User WHERE Username = '${username}'`);
```
peternhale: I believe this throws if the record is not found

peternhale (changes requested):

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
daphne-sfdc: Good catch 🙂 The documentation says "Will throw if either zero records are found OR multiple records are found."

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -499,6 +500,35 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
...
+  const currentTime = new Date();
+  const expiredTraceFlagExists = traceFlags.records.filter(
+    (flag: any) => flag.ExpirationDate && new Date(flag.ExpirationDate) < currentTime
+  ).length > 0;
+
+  if (expiredTraceFlagExists) {
+    const traceFlagId = typeof traceFlags.records[0].Id === 'string'
+      ? traceFlags.records[0].Id
+      : '';
+    await connection.tooling.delete('TraceFlag', traceFlagId);
```
mshanemc: what happens when it tries to run `delete` with an empty string?  isn't that going to throw?

Another way to say it would be 
`const [expiredTraceFlag] = traceFlags.records.filter(
    (flag: any) => flag.ExpirationDate && new Date(flag.ExpirationDate) < currentTime
  )`.

Then if there is a record there, it'll be expiredTraceFlag and otherwise expiredTraceFlag is undefined.  And only try to delete if there is one (and grab its ID)

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -499,6 +500,35 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
...
   void activateTracker.markActivationStop();
   MetricsReporter.extensionPackStatus();
+
+  console.log('DOING TRACE FLAG STUFF');
+
+  const connection = await WorkspaceContext.getInstance().getConnection();
+
+  // If an expired TraceFlag exists, delete it
+  const traceFlags = await connection.tooling.query(
+    "SELECT Id, ExpirationDate FROM TraceFlag WHERE LogType = 'DEVELOPER_LOG'"
```
mshanemc: that would grab any trace flag from any user.  I think the way this is currently written you'd be mixing them between users.

I think you need traceEntityId from TraceFlag to be the current user's ID
https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_traceflag.htm

mshanemc (changes requested):
isolate trace flags by user (see comments)

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
```diff
@@ -86,7 +86,7 @@ export const turnOnLogging = async (extensionContext: vscode.ExtensionContext):
   }
 };
 
-const getUserId = async (connection: Connection): Promise<string> => {
+export const getUserId = async (connection: Connection): Promise<string> => {
```
mshanemc: nit (non-blocking): I think getUserId belongs on OrgAuthInfo and not inside one of the commands.

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -499,6 +500,32 @@ export const activate = async (extensionContext: vscode.ExtensionContext) => {
 
   void activateTracker.markActivationStop();
   MetricsReporter.extensionPackStatus();
+
+  console.log('DOING TRACE FLAG STUFF');
+
+  const connection = await WorkspaceContext.getInstance().getConnection();
+
+  // If an expired TraceFlag exists, delete it
```
mshanemc: I'd rather not have all of this twice.

 "clean up trace flag" could be a function that index#activate and `startApexDebugLogging` both use.

mshanemc (approved):

PR #6322, opened 2025-06-03
feat: [@W-18572237@] implement aura enabled oas gen

peternhale:
Changes:

- Reenabled Aura Enabled annotations for jorje LS config
- Create new Aura Enabled generation strategy
  - Used API version guard to reject Aura Enable generation requests against an org whose version does not support the request
- Modified bid process to use smallest bid by default
- Add new correction step that is used to inject a "x-betaInfo" property into info block, if the strategy defines a betaInfo string

@W-18572237@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/oas/generationStrategy/json/auraEnabledStrategy.ts
+++ b/packages/salesforcedx-vscode-apex/src/oas/generationStrategy/json/auraEnabledStrategy.ts
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ConfigUtil } from '@salesforce/salesforcedx-utils-vscode';
+import * as fs from 'node:fs';
+import { SUM_TOKEN_MAX_LIMIT, IMPOSED_FACTOR } from '..';
+import { workspaceContext } from '../../../context';
+import {
+  ApexClassOASEligibleResponse,
+  ApexClassOASGatherContextResponse,
+  PromptGenerationStrategyBid
+} from '../../schemas';
+import { buildClassPrompt } from '../buildPromptUtils';
+import { GenerationStrategy } from '../generationStrategy';
+import { openAPISchema_v3_0_guided } from '../openapi3.schema';
+
+const MIN_ORG_VERSION = 65.0;
+
+export class AuraEnabledStrategy extends GenerationStrategy {
+  private isDefaultOrg: boolean;
+  private isOrgVersionCompatible: boolean;
+
+  constructor(metadata: ApexClassOASEligibleResponse, context: ApexClassOASGatherContextResponse) {
+    super(
+      metadata,
+      context,
+      'AuraEnabled',
+      0,
+      SUM_TOKEN_MAX_LIMIT * IMPOSED_FACTOR,
+      'OpenAPI documents generated from Apex classes using @AuraEnabled annotations are in beta.'
+    );
+    this.servicePrompts = new Map();
+    this.serviceResponses = new Map();
+    this.serviceRequests = new Map();
+    this.sourceText = fs.readFileSync(new URL(this.metadata.resourceUri.toString()), 'utf8');
+    this.classPrompt = buildClassPrompt(this.context.classDetail);
+    this.oasSchema = JSON.stringify(openAPISchema_v3_0_guided);
+    this.isDefaultOrg = false;
+    this.isOrgVersionCompatible = false;
+  }
+
+  get openAPISchema(): string {
+    return this.oasSchema;
+  }
+
+  private async checkOrgVersion(): Promise<void> {
+    try {
+      const targetOrg = await ConfigUtil.getTargetOrgOrAlias();
+      this.isDefaultOrg = targetOrg !== undefined;
+
+      if (this.isDefaultOrg) {
+        const connection = await workspaceContext.getConnection();
+        const apiVersion = connection.getApiVersion();
+        // Convert API version to number by handling decimal format (e.g. "58.0")
+        const numericVersion = parseFloat(apiVersion);
+        this.isOrgVersionCompatible = numericVersion >= MIN_ORG_VERSION;
+      }
+    } catch (err) {
+      console.error('Failed to initialize org checks:', err);
+      this.isDefaultOrg = false;
+      this.isOrgVersionCompatible = false;
+    }
+  }
+
+  public async bid(): Promise<PromptGenerationStrategyBid> {
+    // Initialize org checks
+    await this.checkOrgVersion();
+
+    // Check if any method has @AuraEnabled annotation
+    const hasAuraEnabled = this.context.methods.some(method =>
+      method.annotations.some(annotation => annotation.name === 'AuraEnabled')
+    );
+
+    // Only bid if we have Aura-enabled methods AND we're in the default org AND the org version is compatible AND the class exists in org
+    const shouldBid = hasAuraEnabled && this.isDefaultOrg && this.isOrgVersionCompatible;
+
+    return {
+      result: {
+        maxBudget: shouldBid ? this.maxBudget : 0,
+        callCounts: shouldBid ? 1 : 0
+      }
+    };
+  }
+
+  public async generateOAS(): Promise<string> {
+    const responses: string[] = [];
+
+    // Get the connection and make the API call
+    const connection = await workspaceContext.getConnection();
+    const className = this.context.classDetail.name;
+    const apiVersion = connection.getApiVersion();
+    const endpoint = `${connection.instanceUrl}/services/data/v${apiVersion}/specifications/oas3/apex/${className}`;
+
+    try {
+      const result = await connection.request({
+        method: 'GET',
+        url: endpoint
+      });
+      responses.push(JSON.stringify(result));
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+      throw new Error(`Failed to fetch OAS specification from org: ${errorMessage}`);
+    }
+    this.oasSchema = responses.join('\n');
+    return this.oasSchema;
+  }
+}
--- a/packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
+++ b/packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
@@ -7,10 +7,11 @@
 import { nls } from '../messages';
 import { cleanupGeneratedDoc } from '../oasUtils';
 import GenerationInteractionLogger from './generationInteractionLogger';
+import { GenerationStrategy } from './generationStrategy/generationStrategy';
 import {
-  GenerationStrategy,
-  GenerationStrategyFactory,
-  Strategy
+  GenerationStrategyType,
+  initializeAndBid,
+  StrategyTypes
 } from './generationStrategy/generationStrategyFactory';
 import {
   ApexClassOASEligibleResponse,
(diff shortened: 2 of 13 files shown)


packages/salesforcedx-vscode-apex/src/oas/promptGenerationOrchestrator.ts
```diff
@@ -7,84 +7,130 @@
...
+    const result = this.getStrategyByCallCount(bids, () => true);
+    return result ? result[0] : undefined;
+  }
+
+  private getMostCallsStrategy(
+    bids: Map<GenerationStrategyType, PromptGenerationStrategyBid>
+  ): GenerationStrategyType | undefined {
+    const result = this.getStrategyByCallCount(bids, bid => bid.result.callCounts > 0);
+    return result ? result[0] : undefined;
+  }
```
daphne-sfdc: It seems like these two functions are equivalent - they both select the strategy with the minimum call count.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,448 @@
...
+      classDetail,
+      methods,
+      properties: [],
+      relationships: new Map()
+    };
+    const result = hasValidRestAnnotations(context);
+    expect(result).toBe(false);
+  });
+
+  it('should return true when class has RestResource annotation and at least one method has HTTP REST annotation', () => {
```
daphne-sfdc, line 151: This is the case where a @RestResource class contains both an @HttpGet method and an @AuraEnabled method.  I remember we discussed during a standup a few months ago that a class that is both @RestResource and @AuraEnabled should not be eligible for OAS doc generation.  Have the requirements changed?

daphne-sfdc: It looks like this case is treated like a regular @RestResource case, unlike `'should return false when class has annotations even if methods have AuraEnabled annotations'` which is treated as an @AuraEnabled case since there are no @Http____ methods available there to pair with the @RestResource.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,448 @@
...
+          comment: ''
+        }
+      ]
+    };
+
+    const result = hasAuraEnabledMethods(context as ApexClassOASGatherContextResponse);
+    expect(result).toBe(true);
+  });
+
+  it('should return false when class has annotations even if methods have AuraEnabled annotations', () => {
```
daphne-sfdc, line 313: I would like to see the annotation for `TestClass` be something other than @RestResource, since this could also be treated as a @RestResource class with no eligible @Http____ methods.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
httpget: waveball

packages/salesforcedx-vscode-apex/src/oasUtils.ts
```diff
@@ -251,3 +257,89 @@ export const getCurrentTimestamp = (): string => {
...
+ */
+export const hasValidRestAnnotations = (context: ApexClassOASGatherContextResponse): boolean => {
+  // Check for class-level RestResource annotation
+  if (!hasRestResourceAnnotation(context)) {
+    return false;
+  }
+
+  // Check for at least one method with HTTP REST annotation
+  return hasHttpRestAnnotations(context);
+};
```
daphne-sfdc: Can simplify to the same format as `hasAuraFrameworkCapability()` 🙂

```suggestion
export const hasValidRestAnnotations = (context: ApexClassOASGatherContextResponse): boolean =>
  // Check for class-level RestResource annotation and at least one method with HTTP REST annotation
  hasRestResourceAnnotation(context) && hasHttpRestAnnotations(context);
```

packages/salesforcedx-vscode-apex/src/oas/generationStrategy/json/auraEnabledStrategy.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  public async bid(): Promise<PromptGenerationStrategyBid> {
+    // Initialize org checks
+    await this.checkOrgVersion();
+
+    // Check if any method has @AuraEnabled annotation
+    const hasAuraEnabled = hasAuraFrameworkCapability(this.context);
+
+    // Only bid if we have Aura-enabled methods AND we're in the default org AND the org version is compatible AND the class exists in org
+    const shouldBid = hasAuraEnabled && this.isDefaultOrg && this.isOrgVersionCompatible;
```
daphne-sfdc: The comment says we want to check if the class exists in the org during the bidding process, but the code doesn't contain this check.  Right now, if you try to generate an OAS doc on an @AuraEnabled class that isn't deployed to the org, you get this error:
Screenshot 2025-06-10 at 10 24 26 AM

Should the comment be changed to remove this condition, or should the condition be added to the code?

packages/salesforcedx-vscode-apex/src/oas/externalServiceRegistrationManager.ts
```diff
@@ -333,6 +371,55 @@ export class ExternalServiceRegistrationManager {
...
+
+    // Clean all description fields in the document
+    const descriptionPaths = [
+      '$.info.description',
+      '$.paths[*][*].description',
+      '$.paths[*].description',
+      '$.paths[*][*].responses[*].description',
+      '$.paths[*][*].parameters[*].description',
+      '$.paths[*][*].requestBody.description',
+      '$.components.schemas[*].description'
```
daphne-sfdc, line 392: If I recall correctly, we decided that we would inline all the properties instead of putting them in a `components.schemas` block for better handling in the Core backend.  Thus a generated OAS doc would never have a `components.schemas` block, and Line 392 isn't needed.

packages/salesforcedx-vscode-apex/src/oas/externalServiceRegistrationManager.ts
```diff
@@ -126,11 +160,15 @@ export class ExternalServiceRegistrationManager {
...
-    const replaceElement = this.isESRDecomposed ? "'" : '&apos;';
-    const safeOasSpec = stringify(this.oasSpec ?? {})
-      .replaceAll('"', replaceElement)
-      .replaceAll('type: Id', 'type: string');
+    // Clean the OpenAPI document before stringifying
+    const cleanedOasSpec = this.oasSpec ? this.cleanOasDocument(this.oasSpec) : {};
+
+    const safeOasSpec = stringify(cleanedOasSpec, null, {
+      singleQuote: false, // Disable single quotes entirely
+      doubleQuotedAsJSON: false, // Use JSON-safe escaping for double quotes
```
daphne-sfdc: The comment and code are contradictory.  JSON-safe escaping is `doubleQuotedAsJSON: true`.  (The default value is false, so if you did actually mean to keep it as false, this line can be removed entirely.)

packages/salesforcedx-vscode-apex/test/jest/oas/promptGenerationOrchestrator.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    it('should return undefined when all strategies have zero calls', () => {
+      const bids = new Map<GenerationStrategyType, PromptGenerationStrategyBid>([
+        ['ApexRest', { result: { callCounts: 0, maxBudget: 100 } }],
+        ['AuraEnabled', { result: { callCounts: 0, maxBudget: 100 } }]
+      ]);
+
+      const result = orchestrator['getMostCallsStrategy'](bids);
+      expect(result).toBeUndefined();
+    });
```
daphne-sfdc, line 110: Also add this test case to the `describe('getLeastCallsStrategy', () => {` block.

packages/salesforcedx-vscode-apex/test/jest/oas/promptGenerationOrchestrator.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  describe('getMostCallsStrategy', () => {
+    it('should return strategy with highest calls among those with at least one call', () => {
+      const bids = new Map<GenerationStrategyType, PromptGenerationStrategyBid>([
+        ['ApexRest', { result: { callCounts: 5, maxBudget: 100 } }],
+        ['AuraEnabled', { result: { callCounts: 0, maxBudget: 100 } }]
+      ]);
+
+      const result = orchestrator['getMostCallsStrategy'](bids);
+      expect(result).toBe('ApexRest');
+    });
```
daphne-sfdc: This test case doesn't match the description.  The `callCounts` value for AuraEnabled should be nonzero.

packages/salesforcedx-vscode-apex/src/oas/externalServiceRegistrationManager.ts
peternhale: This is included just in case the new OAS doc for Aura Enabled produces components.

packages/salesforcedx-vscode-apex/src/oas/generationStrategy/json/auraEnabledStrategy.ts
peternhale: Good catch, I am letting the org handle that error condition.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
peternhale: This is so I can test to see if the rest qualifier code can handle the aura enabled annotation being present.

The other things we look for are mixed rest and aura enabled.

I added test cases below for those.

packages/salesforcedx-vscode-apex/test/jest/oas/promptGenerationOrchestrator.test.ts
peternhale: fixed

packages/salesforcedx-vscode-apex/test/jest/oas/promptGenerationOrchestrator.test.ts
peternhale: Done

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,896 @@
...
+      },
+      methods: [],
+      properties: [],
+      relationships: new Map()
+    };
+    expect(hasNoClassAnnotations(context)).toBe(false);
+  });
+});
+
+describe('hasAuraEnabledMethods', () => {
```
daphne-sfdc: This entire section is already covered by the section in Lines 293-462, as `hasAuraFrameworkCapability()` directly calls `hasAuraEnabledMethods()`.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,896 @@
...
+        properties: [],
+        relationships: new Map()
+      };
+      const result = hasValidRestAnnotations(context);
+      expect(result).toBe(true);
+    }
+  });
+});
+
+describe('hasAuraEnabledMethods', () => {
```
daphne-sfdc: This section is testing the `hasAuraFrameworkCapability()` function.

```suggestion
describe('hasAuraFrameworkCapability', () => {
```

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,896 @@
...
+      },
+      methods: [],
+      properties: [],
+      relationships: new Map()
+    };
+    expect(hasAuraEnabledMethods(context)).toBe(false);
+  });
+});
+
+describe('hasRestResourceAnnotation', () => {
```
daphne-sfdc: This section is covered by the `hasValidRestAnnotations()` section in Lines 25-291.

packages/salesforcedx-vscode-apex/test/jest/oas/oasUtils.test.ts
```diff
@@ -0,0 +1,896 @@
...
+      },
+      methods: [],
+      properties: [],
+      relationships: new Map()
+    };
+    expect(hasRestResourceAnnotation(context)).toBe(false);
+  });
+});
+
+describe('hasHttpRestAnnotations', () => {
```
daphne-sfdc: This section is also completely covered by the `hasValidRestAnnotations()` section in Lines 25-291.

PR #6330, opened 2025-06-04
@W-18520689 no core bundle in faux gen

mshanemc:
#### main things
- remove core-bundle references from faux-generator
  - now references the library itself as a devDep (needs types)
  - Connection is passed in by the  refreshSobjects action.  It manages what version should be used (that was the most complicated thing that was in faux-generator that was a core-bundle dep) 

#### big things
- the helpers/fs uses vscode-uri module where there was previously a mixture of that and vscode.uri.
- most of faux-generator code is now functions instead of [generators, selectors, retrievers, transformers, factory] so it's much simpler to test

#### little things
- gets rid of some enums so they disappear at compile time
- rather than running all fs writes sequentially for [apex fauxClasses from objects, metadata .json for SOQL, typeNames.json,  
- eslint rule (no useless constructor), applied various places
- added some more tests scenarios to improve coverage
- bump the core bundle
- reduce some barrel files

#### Notes
there are some TS expect-error comments because the bundling process messes with the type of Logger in Connection so they technically don't match.

#### testing ideas
- making sure the api versioning stuff is correct (config > project > default)
- deleting and recreating all the objects
- make sure soql builder is happy

@W-18520689@

e2e https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15458033470


packages/salesforcedx-utils-vscode/src/helpers/fs.ts
CristiCanizales, line 1: I see in the PR description that we're using vscode-uri now because previously there was a mixture of that and vscode, but we weren't even using it before. Why do we need to add it then?

packages/salesforcedx-sobjects-faux-generator/test/jest/generator/fauxClassGenerator.test.ts
```diff
@@ -1,42 +1,40 @@
...
 jest.mock('../../../src/messages');
 
 const vscodeMocked = jest.mocked(vscode);
 const nlsMocked = jest.mocked(nls);
 
-const declarationGeneratorMocked = jest.mocked(DeclarationGenerator);
-
 describe('FauxClassGenerator Unit Tests.', () => {
-  const fakePath = './this/is/a/path';
+  // const fakePath = './this/is/a/path';
```
CristiCanizales:
```suggestion
```

packages/salesforcedx-sobjects-faux-generator/test/jest/generator/fauxClassGenerator.test.ts
```diff
@@ -131,86 +96,67 @@ describe('FauxClassGenerator Unit Tests.', () => {
...
-      const fauxClassGeneratorInst = new FauxClassGenerator(SObjectCategory.STANDARD, fakePath);
-      await fauxClassGeneratorInst.generate(fakeOutput as SObjectRefreshOutput);
-
-      expect(fakeOutput.getStandard).toHaveBeenCalled();
-      expect(fakeOutput.getCustom).not.toHaveBeenCalled();
-      expect(declarationGeneratorMocked.prototype.generateSObjectDefinition).toHaveBeenCalledWith(fakeSObject);
-      expect(generateFauxClassMock).toHaveBeenCalledWith(expectedFolderPath, fakeSobjectDef);
+      // @ts-expect-error - partial mock
+      await generateFauxClasses({ standard: [standardMock], custom: [] });
+      // the actual declarationGenerated is tested separately, so we just check that it is being was called
```
CristiCanizales:
```suggestion
      // the actual declarationGenerated is tested separately, so we just check that it is being called
```

packages/salesforcedx-sobjects-faux-generator/test/jest/generator/fauxClassGenerator.test.ts
```diff
@@ -131,86 +96,67 @@ describe('FauxClassGenerator Unit Tests.', () => {
...
-      const fauxClassGeneratorInst = new FauxClassGenerator(SObjectCategory.CUSTOM, fakePath);
-      await fauxClassGeneratorInst.generate(fakeOutput as SObjectRefreshOutput);
-
-      expect(fakeOutput.getStandard).not.toHaveBeenCalled();
-      expect(fakeOutput.getCustom).toHaveBeenCalled();
-      expect(declarationGeneratorMocked.prototype.generateSObjectDefinition).toHaveBeenCalledWith(fakeSObject);
-      expect(generateFauxClassMock).toHaveBeenCalledWith(expectedFolderPath, fakeSobjectDef);
+      // @ts-expect-error - partial mock
+      await generateFauxClasses({ standard: [], custom: [customMock] });
+      // the actual declarationGenerated is tested separately, so we just check that it is being was called
```
CristiCanizales:
```suggestion
      // the actual declarationGenerated is tested separately, so we just check that it is being called
```

packages/salesforcedx-sobjects-faux-generator/test/jest/generator/fauxClassGenerator.test.ts
```diff
@@ -131,86 +96,67 @@ describe('FauxClassGenerator Unit Tests.', () => {
...
+      expect(declarationGeneratorSpy).toHaveBeenCalledWith(customMock);
+    });
+
+    it('Should process both standard and custom sobjects.', async () => {
+      // @ts-expect-error - partial mock
+      await generateFauxClasses({ standard: [standardMock], custom: [customMock] });
+      // the actual declarationGenerated is tested separately, so we just check that it is being was called
```
CristiCanizales:
```suggestion
      // the actual declarationGenerated is tested separately, so we just check that it is being called
```

packages/salesforcedx-sobjects-faux-generator/src/transformer/sobjectFilter.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
```
CristiCanizales:
```suggestion
 * Copyright (c) 2025, salesforce.com, inc.
```

CristiCanizales (commented):
Changes look good! Just one question and small comments

packages/salesforcedx-utils-vscode/src/helpers/fs.ts
mshanemc: it helps by being testable (they're the same at runtime, but `vscode` is types only so you have to heavily mock everything in it).  in `vscode-uri` they publish the real module `vscode` uses so it can do UT.

For example, doing path joins with it or turning a local windows/mac/unix filename into a URI.  Makes the tests a bit more realistic and likely to catch something useful.

CristiCanizales (approved):
[QA doc](https://salesforce.quip.com/q3c9AudcbfZE) ✅

PR #6331, opened 2025-06-05
@W-18520714 feat: soql without core-bundle

mshanemc:
### What does this PR do?
soql ext uses sfdx-core library for types only, no direct dependency on core-bundle

#### other
- update `debounce` lib to v2 which has types and can be imported in the normal way.
- restructure the `withSFConnection` fn wrapper (return types is wrong, could have been a generic but not really necessary)
- adds unit test for the CSV table logic.  It's...not great but subqueries aren't supported anyway so 🤷🏻 .  Probably shouldn't even be part of the tests or the code then?
- remove the `reformat` command/script/hooks
- redo how bundling works (see slack demo)
- serve locally from bundled output

#<Insert GitHub Issue>, @W-18520714@

e2e: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15478398681


packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
```
CristiCanizales:
```suggestion
 * Copyright (c) 2025, salesforce.com, inc.
```
```suggestion
 * Copyright (c) 2024, salesforce.com, inc.
```
```suggestion
 * Copyright (c) 2024, salesforce.com, inc.
```

PR #6333, opened 2025-06-05
refactor: [@W-18705732@] make message handling type safe

peternhale:
# Message Type Safety Improvements

## Overview
Enhanced the localization system with comprehensive type safety for printf-style format specifiers, providing compile-time validation and improved developer experience.

### New Type System
- **Created `packages/salesforcedx-utils/src/types/localization/messageArgs.ts`**
  - `PlaceholderToType<S>` - Maps format specifiers to TypeScript types
  - `ExtractArgs<S>` - Recursively extracts argument types from message strings  
  - `MessageArgs<K, T>` - Main type for compile-time argument validation

### Format Specifier Support
Added type-safe support for all Node.js `util.format()` specifiers:
- `%s` → `string`
- `%d` → `number` (decimal)
- `%i` → `number` (integer) 
- `%f` → `number` (float)
- `%j` → `any` (JSON)
- `%%` → literal percent (no argument consumed)

## 🧪 Testing
**Enhanced `packages/salesforcedx-utils/test/jest/i18n/message.test.ts`**
- ✅ All format specifiers (`%s`, `%d`, `%i`, `%f`, `%j`)
- ✅ Mixed format specifier combinations
- ✅ Literal percent signs (`%%`) edge cases
- ✅ Partial argument population scenarios
- ✅ Node.js `util.format()` behavior validation

## 🔗 Integration
The new `MessageArgs` type is integrated across **15+ packages**:
```typescript
// Example usage in package message files
localize: <K extends MessageKey>(key: K, ...args: MessageArgs<K, typeof enMessages>): string
```

**Packages Updated:**
- `@salesforce/salesforcedx-vscode-core`
- `@salesforce/salesforcedx-vscode-apex`
- `@salesforce/salesforcedx-vscode-lightning`
- `@salesforce/salesforcedx-vscode-lwc`
- `@salesforce/salesforcedx-apex-debugger`
- `@salesforce/salesforcedx-apex-replay-debugger`
- And 9+ more packages

## ✨ Benefits
- **🛡️ Compile-time safety**: Catch formatting errors at build time
- **🔍 IntelliSense support**: Type hints and autocompletion in IDEs
- **🔄 Backward compatible**: No breaking changes to existing APIs
- **📝 Better DX**: Clear error messages for mismatched argument types

## Example
```typescript
// Before: No type checking
localize('userStats', 'John', 'invalid'); // Runtime error

// After: Compile-time validation
localize('userStats', 'John', 150); // ✅ string, number
localize('userStats', 'John', 'invalid'); // ❌ TypeScript error
```
@W-18705732@


packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1544,4 +1544,11 @@ export class ApexDebug extends LoggingDebugSession {
   }
 }
 
-DebugSession.run(ApexDebug);
+// Only run the debug session if not in test mode
+// Check for Jest environment using multiple indicators
+const isTestEnvironment =
+  process.env.NODE_ENV === 'test' || typeof jest !== 'undefined' || process.env.JEST_WORKER_ID !== undefined;
```
mshanemc: do we know if e2e do anything with process.env.NODE_ENV === 'test'

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,163 @@
...
+
+class LocalizationConfig {
+  private static instance: LocalizationConfig | undefined;
+
+  private constructor() {}
+
+  public static getInstance(instanceName: string): LocalizationConfig {
+    if (!LocalizationConfig.instance) {
+      LocalizationConfig.instance = new LocalizationConfig();
+    }
```
mshanemc: nullish assignment
```suggestion
    LocalizationConfig.instance ??= new LocalizationConfig();
```

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,163 @@
...
+    return LocalizationConfig.instance;
+  }
+
+  public isLocaleSupported(locale: unknown): boolean {
```
mshanemc, line 40: are there really not types for it?

packages/salesforcedx-vscode-core/src/messages/index.ts
```diff
@@ -5,31 +5,30 @@
...
-  const procNlsConfig = process.env.VSCODE_NLS_CONFIG;
-  return procNlsConfig ? JSON.parse(procNlsConfig) : undefined;
+export const nls = {
+  localize: <K extends MessageKey>(key: K, ...args: MessageArgs<K, typeof enMessages>): string =>
+    localizationService.localize(key, ...args)
 };
 
-export const nls = new Localization(loadMessageBundle(getNlsConfig()));
+export function coerceMessageKey(key: string): MessageKey {
+  return isValidMessageKey(key) ? key : 'missing_label';
```
mshanemc: how hard would it be to get telemetry events from this? 

we had pattern in CLI libraries where there was a Lifecycle singleton, so anyone could instantiate it and drop messages on it.

then the outermost (sfCommand) would create the listener and send it in.  That might be vscode-core in our case (most likely to be needed and running).  

I'm just thinking we'd want to know if there are runtime missing messages.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
peternhale: For e2e, we are running from vsix, which typically means env is prod. But I will run e2e against this PR

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
peternhale: Union `export type Locale = 'en' | 'ja';`
I can add an isLocale function, instead the of the old code.

packages/salesforcedx-vscode-core/src/messages/index.ts
peternhale: For this instance of missing labels, we could fetch telemetry instance for vscode-core to report the problem. 

The general implementation of message handling eventuallt gets to the code in utils, where the same thing can happen; label not found, replace it with missing label.

I'll add telemetry message here. The other use case deserves a larger discussion.

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
mshanemc: I think it's fine.  It was more about what's coming in...can it really be anything, or is there not an API we trust?

packages/salesforcedx-utils/src/types/localization/messageArgs.ts
```diff
@@ -0,0 +1,29 @@
...
+    : S extends '%i'
+      ? number
+      : S extends '%f'
+        ? number
+        : S extends '%j'
+          ? any
+          : unknown;
+
+// Recursively extracts argument types from a message string
+export type ExtractArgs<S extends string, Acc extends any[] = []> = S extends `${infer _}%${infer P}${infer Rest}`
```
mshanemc, line 22: there's this linter warning `'_' is defined but never used.`

Why is the `_` required?  What does it do?

packages/salesforcedx-utils/src/i18n/message.ts
```diff
@@ -34,7 +34,9 @@ export class Message implements LocalizationProvider {
 
     let labelArgs = args;
     if (args.length >= 1) {
-      const expectedNumArgs = possibleLabel.split('%s').length - 1;
+      // Count all printf-style format specifiers: %s, %d, %i, %f, %j, %%, etc.
+      const formatSpecifiers = possibleLabel.match(/%[sdifj%]/g);
```
mshanemc, line 38: why are the runtime arg length checks still here if TS is checking arg now?

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+    return instance;
+  }
+
+  public get nls(): Localization {
+    if (!this._nls) {
+      try {
+        const messageBundle = this.messageBundleManager.loadMessageBundle();
+        this._nls = new Localization(messageBundle);
+      } catch (error) {
```
mshanemc, line 151:
```suggestion
      } catch {
```

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,170 @@
...
+          `LocalizationService: No messages registered for instance '${this.instanceName}', using fallback localization`
+        );
+        const emptyMessage = new Message({});
+        this._nls = new Localization(emptyMessage);
+      }
+    }
+    return this._nls;
+  }
+
+  public localize<K extends string>(key: K, ...args: any[]): string {
```
mshanemc, line 162: is `args` really an array of `any` ?  `unknown` would be better, but even better would be restrict it to the things we support ex: `string | number | boolean | Record<string,unknown> (object)` etc.

this happens in several places but I think fixing it in `nls.localize` would be ideal.

another option would be to use TS magic to say that l10nService.localize has the type of Localization.localize and not have to maintain it everywhere.

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,170 @@
...
+ */
+const isLocale = (value: unknown): value is Locale => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+  return SUPPORTED_LOCALES.includes(value as Locale);
+};
+
+class LocalizationConfig {
+  private static instance: LocalizationConfig | undefined;
```
mshanemc, line 31: I don't think this class is very useful (singleton might help with avoiding some expensive setup compute or initializing a bunch of immutable stuff only once).

mshanemc: I'd just call the isLocale fn in loadMessageBundle

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,170 @@
...
+  public static getInstance(instanceName: string): LocalizationConfig {
+    LocalizationConfig.instance ??= new LocalizationConfig();
+    return LocalizationConfig.instance;
+  }
+
+  public isLocaleSupported(locale: unknown): boolean {
+    return isLocale(locale);
+  }
+
+  public getDefaultLocale(): Locale {
```
mshanemc, line 44: this is never used

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+
+class MessageBundleManager {
+  private static instances = new Map<string, MessageBundleManager>();
+  private baseMessages: Message | null = null;
+  private localeMessages = new Map<Locale, Message>();
+  private messageBundles = new Map<string, AdvancedMessageBundle[]>();
+
+  private constructor() {}
```
mshanemc, line 56:
```suggestion
```

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,169 @@
...
+/** exported only for test */
+export class MessageBundleManager {
+  private static instances = new Map<string, MessageBundleManager>();
+  private baseMessages: Message | null = null;
+  private localeMessages = new Map<Locale, Message>();
+  private messageBundles = new Map<string, AdvancedMessageBundle[]>();
+
+  private constructor() {}
+
+  public static getInstance(instanceName: string): MessageBundleManager {
```
mshanemc, line 58: I also don't see how someone would end up with another instance of MessageBundleManager.  It's only used in l10nService, which is already a singleton.  I think it's redundant to nest them like that.

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -0,0 +1,169 @@
...
+    const instance = LocalizationService.instances.get(instanceName);
+    if (!instance) {
+      const newInstance = new LocalizationService(instanceName);
+      LocalizationService.instances.set(instanceName, newInstance);
+      return newInstance;
+    }
+    return instance;
+  }
+
+  public get nls(): Localization {
```
mshanemc, line 146: why does this need to be public?  Nothing is consuming it AFAIK.

PR #6354, opened 2025-06-16
fix: @W-18793417 - modify o11yService to support extension specific telemetry configuration

smuduganti-sf:
### What does this PR do?

Modify O11yService to support separate telemetry configurations for each extension by maintaining a dedicated instance per extension.

Changes Made:

- Replaced the static instance variable with a Map<string, O11yService> to store instances per extension ID.
- Updated getInstance() to accept an extensionId and return the corresponding instance from the map.
- Updated the initialize method to set up extension-specific properties and shared resources.
- Introduced o11yReporterInstance support in determineReporters.ts to ensure each extension gets its own O11y reporter instance with the correct routing and attribution.

Shared Instrumentation App (O11y SDK Limitation):
Introducing extension-specific O11yService instances initially led to the registration of multiple instrumented apps. However, the O11y SDK has a limitation and allows only one instrumented app per process, resulting in the following error:
Error: An app has already been registered with instrumentation

Solution:

- Created shared static resources (sharedInstrumentation, sharedInstrApp, etc.)
- First extension initializes the shared instrumentation; subsequent extensions reuse it.
- Extension-specific metadata (e.g., name, endpoint) is automatically included in telemetry events to preserve attribution.

Expected Outcome:

- Each extension will have:

1. Its own O11yService instance and dedicated o11yUploadEndpoint, ensuring telemetry is correctly isolated per extension.
2. A corresponding o11yReporterInstance created via determineReporters.ts, enabling proper attribution and routing of telemetry data.

- Shared instrumentation resolves the SDK constraint while supporting multiple extensions efficiently in the same process.

#<Insert GitHub Issue>, @W-18793417@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>


packages/salesforcedx-utils-vscode/src/services/o11yService.ts
```diff
@@ -32,55 +32,93 @@ export class O11yService {
...
+      O11yService.sharedO11yModules!;
+
+    // Create a single shared instrumentation
+    const sharedInstrumentationName = 'salesforce-vscode-extensions-instrumentation';
+    console.log('Creating shared instrumentation with name:', sharedInstrumentationName);
+    O11yService.sharedInstrumentation = getInstrumentation(sharedInstrumentationName);
+
+    // Create a single shared app
+    const sharedAppName = 'salesforce-vscode-extensions';
+    console.log('Registering shared app with name:', sharedAppName);
```
gbockus-sf: Dunno if we always want these console log messages.  Going to be like ~9 logs on every startup, but as this expands to all extensions going to be pretty noisy.

packages/salesforcedx-utils-vscode/src/services/o11yService.ts
```diff
@@ -32,55 +32,93 @@ export class O11yService {
...
   readonly environment: Record<string, string> = {};
-  private o11yModules: Awaited<ReturnType<typeof loadO11yModules>> | null = null;
-  private static instance: O11yService | null = null;
+  private static instances: Map<string, O11yService> = new Map();
+
+  // Shared instrumentation and app across all extensions
+  private static sharedInstrumentation: Instrumentation | null = null;
+  private static sharedInstrApp: InstrumentedAppMethods | null = null;
+  private static sharedO11yModules: Awaited<ReturnType<typeof loadO11yModules>> | null = null;
+  private static sharedProtoEncoderFunc: ProtoEncoderFuncType | null = null;
```
gbockus-sf, line 43: Hopefully with the move to the shared lib these types can be tightened up.

packages/salesforcedx-utils-vscode/src/telemetry/reporters/determineReporters.ts
```diff
@@ -65,29 +66,30 @@ export const initializeO11yReporter = async (
...
+  o11yReporterInstances.set(extName, o11yReporterInstance);
+  setO11yInitializationPromise(extName, initPromise);
   await initPromise;
 };
 
 const addO11yReporter = (reporters: TelemetryReporter[], extName: string): void => {
-  if (o11yReporterInstance) {
-    reporters.push(o11yReporterInstance);
+  if (o11yReporterInstances.has(extName)) {
+    reporters.push(o11yReporterInstances.get(extName)!);
```
gbockus-sf, line 92: IS the unit test coverage for this file still good?

gbockus-sf (commented):
LGTM.  Will ping IDE Experience team for approval.

packages/salesforcedx-utils-vscode/src/services/o11yService.ts
smuduganti-sf: Agreed - I'll address this as part of the ongoing improvements in the shared library. Once the types are finalized and exported there, I’ll revisit this file to tighten the types accordingly.

packages/salesforcedx-utils-vscode/src/services/o11yService.ts
smuduganti-sf: I’ll remove the console.log statements to avoid unnecessary noise, especially as more extensions begin using this shared service.

From what I see, the shared initialization logic (including these logs) only runs once — on the first extension that calls initialize(). So while we wouldn’t get 9 logs per extension, it could still be misleadingly noisy during startup.

packages/salesforcedx-utils-vscode/src/telemetry/reporters/determineReporters.ts
smuduganti-sf: I've added unit tests to improve coverage for this file. To support dynamic imports required for mocking, I had to include @ts-nocheck and some ESLint disables, given the known limitations in the TypeScript + Jest + ESM setup.

packages/salesforcedx-utils-vscode/src/services/o11yService.ts
```diff
@@ -28,59 +28,93 @@
...
+
+  // Extension-specific properties
+  private extensionName: string = '';
```
peternhale: Why empty string as an initial value?

smuduganti-sf: The empty string serves as a clear indicator that the service hasn't been properly initialized yet, making the initialization state explicit and predictable. 

This creates a necessary separation between instance creation and configuration.

TypeScript requires class properties to be initialized. The empty string provides a safe default that satisfies the type system without causing runtime errors.

peternhale: I think undefined would serve the same purpose and this aligns with recent style related changes we would like all contributors to adhere to.

smuduganti-sf: Got it. I'll update the initialization to use undefined instead of empty string to align with modern TypeScript practices and recent style guidelines.

smuduganti-sf: Done.

smuduganti-sf: This will be replaced with the @salesforce/o11y-reporter npm package in a follow-up pull request after this is merged.

peternhale (approved):

PR #6357, opened 2025-06-17
fix: add 'No components retrieved' message in Output Tab if a manifest that contains no valid components is retrieved

daphne-sfdc:
### What does this PR do?
Provides a "No components retrieved" message in the Output Tab if a user tries to retrieve a manifest containing no valid components.

@W-18811418@

### Functionality Before
Screenshot 2025-06-17 at 1 32 55 PM

### Functionality After
Screenshot 2025-06-17 at 3 00 30 PM (2)


CristiCanizales (approved):
Thanks for the quick fix! 🚀

PR #6372, opened 2025-06-24
@W-14564740 refactor: eslint rules

mshanemc:
### What does this PR do?
see eslint.config.mjs for full list of new rules

main things are highlighted in PR comments.

@W-14564740@


packages/salesforcedx-apex-debugger/src/commands/baseDebuggerCommand.ts
```diff
@@ -14,7 +14,7 @@ export class BaseDebuggerCommand extends BaseCommand {
   private readonly debuggerApiPath = 'services/debug/v41.0';
   private readonly request: DebuggerRequest | undefined;
 
-  public constructor(commandName: string, debuggedRequestId: string, queryString?: string, request?: DebuggerRequest) {
+  constructor(commandName: string, debuggedRequestId: string, queryString?: string, request?: DebuggerRequest) {
```
mshanemc, line 18: methods and props have access modifiers.  constructors only need them when they're NOT public.

packages/salesforcedx-apex-debugger/src/core/breakpointService.ts
```diff
@@ -35,14 +35,14 @@ export class BreakpointService {
   }
 
   public isApexDebuggerBreakpointId(id: string): boolean {
-    return id != null && id.startsWith(DEBUGGER_BREAKPOINT_ID_PREFIX);
+    return id?.startsWith(DEBUGGER_BREAKPOINT_ID_PREFIX);
```
mshanemc, line 38: optional chaining instead of logical &&

packages/salesforcedx-apex-debugger/src/core/streamingService.ts
```diff
@@ -75,7 +75,7 @@ export class StreamingService {
 
     await this.systemEventClient.subscribe();
     await this.userEventClient.subscribe();
-    return Promise.resolve(this.isReady());
+    return this.isReady();
```
mshanemc, line 78: don't return a Promise.resolve, just return the value

packages/salesforcedx-apex-debugger/test/jest/core/breakpointService.test.ts
```diff
@@ -313,19 +313,19 @@ describe('breakpointService Unit Tests.', () => {
       (breakpointService as any).lineBreakpointCache.set(fakeUri, [
         {
           line: 1,
-          breakpointId: bpId + '1'
+          breakpointId: `${bpId}1`
```
mshanemc, line 316: use template literals everywhere for strings instead of doing math with them.

packages/salesforcedx-sobjects-faux-generator/src/generator/soqlMetadataGenerator.ts
```diff
@@ -15,8 +15,8 @@ const outputFolderPath = path.join(projectPaths.toolsFolder(), SOQLMETADATA_DIR)
 export const generateAllMetadata = async (sobjects: SObjectsStandardAndCustom) => {
   await createDirectory(outputFolderPath);
 
-  await Promise.all([
-    ...Object.entries(sobjects)
+  await Promise.all(
+    Object.entries(sobjects)
```
mshanemc, line 19: prevents unnecessary spread (helps you know when a function takes an interable without needing to be spread first)

packages/salesforcedx-utils-vscode/src/helpers/activationTracker.ts
```diff
@@ -24,7 +24,7 @@ export class ActivationTracker {
     };
   }
 
-  async markActivationStop(activationEndDate?: Date): Promise<void> {
+  public async markActivationStop(activationEndDate?: Date): Promise<void> {
```
mshanemc, line 27: when adding these, I looked around for usage.  if I couldn't find any, I marked them private.  Most needed to go to public.

packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
```diff
@@ -131,7 +131,7 @@ export class TraceFlags {
 
   private calculateExpirationDate(expirationDate: Date): Date {
     if (!this.isValidDateLength(expirationDate)) {
-      expirationDate = new Date(Date.now() + this.LOG_TIMER_LENGTH_MINUTES * this.MILLISECONDS_PER_MINUTE);
+      return new Date(Date.now() + this.LOG_TIMER_LENGTH_MINUTES * this.MILLISECONDS_PER_MINUTE);
```
mshanemc, line 140: this is one of my favorite rules because it causes really weird bugs...fn/methods shouldn't modify their parameters (the consumer is usually not expecting that side effect).  I did exempt a few places where I couldn't tell if the side effect was important/intentional.

mshanemc: [logic: everything that's not a primite in JS is pass-by-ref]

packages/salesforcedx-utils-vscode/src/services/o11yService.ts
```diff
@@ -28,14 +28,14 @@
...
+const O11Y_UPLOAD_THRESHOLD_BYTES = 50_000;
 export class O11yService {
-  O11Y_UPLOAD_THRESHOLD_BYTES = 50_000;
-  o11yUploadEndpoint: string | undefined;
-  instrumentation: Instrumentation;
-  _instrApp: InstrumentedAppMethods;
-  protoEncoderFunc: ProtoEncoderFuncType | null = null;
-  a4dO11ySchema: unknown;
-  readonly environment: Record<string, string> = {};
+  private o11yUploadEndpoint: string | undefined;
```
mshanemc, line 33: olly stuff is a prime example of why these rules will help...there's stuff that was relying on "default public" used only for tests.

I want people to think a bit more about what they expose and least-privilege class stuff.

packages/salesforcedx-utils/src/cli/configGet.ts
```diff
@@ -23,15 +23,11 @@ export class ConfigGet {
...
-      results.forEach(entry => configMap.set(entry.key, entry.value));
-      return Promise.resolve(configMap);
-    } catch (e) {
-      return Promise.reject(e);
-    }
+    const configMap = new Map<string, string>();
+    // Will be removed as part of removing CLI calls
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    const results = JSON.parse(result).result as any[];
+    results.forEach(entry => configMap.set(entry.key, entry.value));
```
mshanemc, line 30: ex of no-useless-catch

you can just let the error happen

packages/salesforcedx-utils/src/i18n/advancedLocalization.ts
```diff
@@ -148,7 +148,7 @@ export class LocalizationService {
       try {
         const message = this.messageBundleManager.loadMessageBundle();
         this._nls = new Localization(message);
-      } catch (error) {
+      } catch {
```
mshanemc, line 151: ex of prefer-optional-catch-binding (if you aren't using the error, you can omit it from the catch).  Helps you think about whether you needed anything from the error or not.

packages/salesforcedx-utils/src/requestService/requestService.ts
```diff
@@ -35,7 +35,7 @@ export class RequestService {
   private _connectionTimeoutMs: number = DEFAULT_CONNECTION_TIMEOUT_MS;
 
   public getEnvVars(): NodeJS.ProcessEnv {
-    const envVars = Object.assign({}, process.env);
+    const envVars = { ...process.env };
```
mshanemc, line 38: object spread has been around since 2018!  (prefer-object-spread)

packages/salesforcedx-visualforce-language-server/src/modes/embeddedSupport.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-param-reassign */
```
mshanemc, line 1: I left all the behavior here in this copy-paste 3PP code

packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
```diff
@@ -242,7 +242,7 @@ const resolveTestMethodParam = async (testMethod: string): Promise<string> => {
     // value not provided for re-run invocations
     // apply cached value, if available
     if (apexTestRunCacheService.hasCachedMethodTestParam()) {
```
mshanemc, line 244: I **think** these were ok with `testMethod` primitive (no side effect worries)

packages/salesforcedx-vscode-apex-replay-debugger/src/index.ts
```diff
@@ -65,18 +65,12 @@ const registerCommands = (): vscode.Disposable => {
   const launchFromLogFileCmd = vscode.commands.registerCommand(
     'sf.launch.replay.debugger.logfile',
     (editorUri: URI) => {
```
mshanemc, line 73: optional chaining makes some nice things possible that were hard before

packages/salesforcedx-vscode-core/src/commands/retrieveMetadata/typeNodeDescriber.ts
```diff
@@ -11,27 +11,16 @@ import { NodeDescriber } from './nodeDescriber';
 export class TypeNodeDescriber extends NodeDescriber {
   public buildMetadataArg(data?: LocalComponent[]): string {
     if (data) {
-      const dedupe = new Set<string>(); // filter dupes caused by cli bug. See buildOutput in parent class
-      data.forEach(c => dedupe.add(`${c.type}:${c.fileName}`));
+      const dedupe = new Set<string>(data.map(c => `${c.type}:${c.fileName}`)); // filter dupes caused by cli bug. See buildOutput in parent class
       if (dedupe.size < this.node.children!.length) {
```
mshanemc, line 15: this was a very strange way to make a comma separated list

packages/salesforcedx-vscode-core/src/commands/deployManifest.ts
```diff
@@ -39,18 +36,14 @@ class LibraryDeployManifestExecutor extends DeployExecutor<string> {
...
-      const errorMessage = nls.localize('deploy_select_manifest');
-      telemetryService.sendException('deploy_with_manifest', errorMessage);
-      void notificationService.showErrorMessage(errorMessage);
-      channelService.appendLine(errorMessage);
-      channelService.showChannelOutput();
-      return;
-    }
+  const resolved =
+    manifestUri ??
+    (await getUriFromActiveEditor({
```
mshanemc, line 41: there's a bunch like this, so I made a shared fn for them.  One of the very few human refactorings in this PR.

packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
```diff
@@ -36,18 +36,17 @@ export const deploySourcePaths = async (
   uris: URI[] | undefined,
   isDeployOnSave?: boolean | undefined
```
mshanemc, line 37: this is mostly about not mutating those inputs

packages/salesforcedx-vscode-core/src/commands/describeMetadata.ts
```diff
@@ -11,10 +11,6 @@ import * as path from 'node:path';
 import { SfCommandletExecutor } from './util';
 
 class DescribeMetadataExecutor extends SfCommandletExecutor<string> {
-  public constructor() {
```
mshanemc, line 21: linter is pretty smart.  no-useless-constructor looked and said, "oh, the parent class doesn't have a constructor, so this isn't actually doing anything"

eslint.config.mjs
```diff
@@ -288,7 +309,10 @@ export default [
       'no-fallthrough': 'off',
       'no-invalid-this': 'off',
       'no-new-wrappers': 'error',
+      'no-param-reassign': 'error',
```
kylewalke, line 306: nice

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -201,7 +201,7 @@ export class ApexVariable extends Variable {
   }
 
   private static extractNumber(s: string): string {
-    if (s[0] === '[' && s[s.length - 1] === ']') {
+    if (s[0] === '[' && s.at(-1) === ']') {
```
kylewalke, line 201: 👍

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -400,10 +380,10 @@ export class MapTupleContainer implements VariableContainer {
     }
 
     const idsToFetch = [];
-    if (this.tuple.key && this.tuple.key.ref) {
+    if (this.tuple.key?.ref) {
```
kylewalke, line 380: safe navigation operators are so much nicer to read

packages/salesforcedx-apex-debugger/src/core/breakpointService.ts
```diff
@@ -116,12 +116,12 @@ export class BreakpointService {
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
       const breakpointId = JSON.parse(result).result.id as string;
       if (this.isApexDebuggerBreakpointId(breakpointId)) {
-        return Promise.resolve(breakpointId);
+        return breakpointId;
       } else {
-        return Promise.reject(result);
+        throw result;
```
kylewalke, line 121: much cleaner

packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
kylewalke: I suppose that's the result of getting used to doing. 

```
Array x = [1,2,3];
Add4(x);
```
instead of 

```
Array x = [1,2,3];
x = Add4(x);
```

or worse, doing both at different times. As a consistent pattern, if we keep to not modifying params I think it should be a good cleanup/improvement long term.

packages/salesforcedx-utils-vscode/src/output/table.ts
```diff
@@ -52,15 +52,17 @@ export class Table {
             } else {
               // If the cell is multiline, add an additional line to the table
               // and pad it to the beginning of the current column
-              outputRow +=
-                '\n' +
-                this.fillColumn('', rowWidth, COLUMN_FILLER, isLastCol) +
-                this.fillColumn(line, rowWidth, COLUMN_FILLER, isLastCol);
+              outputRow += `\n${this.fillColumn(
```
kylewalke, line 55: Old way seemed a little easier to read imo.

eslint.config.mjs
```diff
@@ -73,7 +73,21 @@ export default [
     },
     rules: {
       'local/no-duplicate-i18n-values': 'error',
+      'unicorn/consistent-empty-array-spread': 'error',
+      'unicorn/consistent-function-scoping': 'error',
+      'unicorn/explicit-length-check': 'error',
+      'unicorn/no-instanceof-builtins': 'error',
```
kylewalke, line 73: Why do we not want instanceof?

mshanemc: [docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-instanceof-builtins.md) link to the [logic](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms) 

for us, that's going to possibly be 
1. web stuff
2. workers vs. main thread

maybe it's overly cautious, but if the linter can change it to `typeof` etc that seems good

another ex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#using_instanceof_with_string

[stuff like this is why people hate on JS; you almost need a linter rule to help steer you away from some of the problem features of the language]

packages/salesforcedx-utils-vscode/src/output/table.ts
mshanemc: oh, interesting.  I feel the opposite way (the multiline string concat always fools me into thinking they'll be on different lines when they aren't), and the "spaces for clarity" `'x' + 'y'` trick me into thinking there'll be actual spaces until I see the output.

guess it matters how long you've been looking at template literals vs. the original way.

packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
mshanemc: yeah, they've added a lot of non-mutable array methods lately (ex: toSorted, toSpliced) to help avoid some of the unintended mutations.

mshanemc: for your 2nd example, `const y = x.map(add4)` is my preference.  I'll almost never use `let`, big fan of immutability everywhere.

eslint.config.mjs
kylewalke: I think that makes sense for the sake of avoiding any future issues once we have the web stuff especially since it's called out explicitly as an example in the docs. After reading the string examples, I can barely understand why anyone would use it at all. That kind of unexpected behavioral error would be hell to track down.

packages/salesforcedx-utils-vscode/src/output/table.ts
kylewalke: Yeah I think coming from Apex land I naturally pattern match more easily to the old way but realistically I should just acclimate to the new interpolated way.

eslint.config.mjs
mshanemc: > I can barely why anyone would use it

because JS isn't their primary language (think Core Java dev, OSS contributor who looks at Apex or some other language all day, etc) and they expect JS to be reasonable but it's only "mostly reasonable"

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -305,33 +305,23 @@ export class ObjectReferenceContainer implements VariableContainer {
 }
 
 export class CollectionReferenceContainer extends ObjectReferenceContainer {
-  public async expand(session: ApexDebug, filter: FilterType, start?: number, count?: number): Promise<ApexVariable[]> {
+  public async expand(session: ApexDebug, filter: FilterType, start = 0, count?: number): Promise<ApexVariable[]> {
```
daphne-sfdc, line 308: The changes to this function 🤩

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -952,14 +932,7 @@ export class ApexDebug extends LoggingDebugSession {
   }
 
   private hasStackFrames(response: DebuggerResponse): boolean {
```
daphne-sfdc, line 934: This function can be simplified further to
```
private hasStackFrames(response: DebuggerResponse): boolean {
  return response.stateResponse?.state?.stack?.stackFrame?.length > 0 ? true : false;
}
```

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1224,7 +1191,7 @@ export class ApexDebug extends LoggingDebugSession {
   }
 
   protected printToDebugConsole(msg?: string, sourceFile?: Source, sourceLine?: number): void {
-    if (msg && msg.length !== 0) {
+    if (msg && msg?.length > 0) {
```
daphne-sfdc: Can reduce redundancy by changing to
```
if (msg?.length > 0) {
```

Same for Lines 1204 and 1210

packages/salesforcedx-apex-debugger/test/unit/adapter/apexDebug.test.ts
```diff
@@ -86,6 +86,28 @@ describe('Interactive debugger adapter - unit', () => {
...
+        // Ignore disconnect errors in tests
+      }
+
+      adapter.clearIdleTimers();
+    }
+  });
+
   afterAll(() => {
     if (adapter) {
       adapter.clearIdleTimers();
```
daphne-sfdc: Since we're doing the cleanup after each test case now, we can get rid of this `afterAll()` cleanup.

packages/salesforcedx-vscode-apex/src/oas/generationStrategy/json/apexRest.ts
```diff
@@ -34,21 +34,11 @@ import { openAPISchema_v3_0_guided } from '../openapi3.schema';
...
 export class ApexRestStrategy extends GenerationStrategy {
-  methodsList: string[];
-  methodsDocSymbolMap: Map<string, DocumentSymbol>;
-  methodsContextMap: Map<string, ApexOASMethodDetail>;
-  urlMapping: string;
-  servicePrompts: Map<string, string>;
-  serviceResponses: Map<string, string>;
-  serviceRequests: Map<string, Promise<string>>;
-  classPrompt: string; // The prompt for the entire class
-  oasSchema: string;
```
daphne-sfdc, line 45: Thanks for cleaning up all these unused properties! 🎉

packages/salesforcedx-vscode-apex/src/apexErrorHandler.ts
```diff
@@ -44,7 +44,8 @@ export class ApexErrorHandler extends EventEmitter implements ErrorHandler {
       this.emit('restarting', this.restarts.length);
       return { action: CloseAction.Restart };
     } else {
-      const diff = this.restarts[this.restarts.length - 1] - this.restarts[this.restarts.length - 5];
+      // assertions: we know the length is > 5
```
daphne-sfdc:
```suggestion
      // assertions: we know the length is >= 5
```

packages/salesforcedx-vscode-apex/test/jest/oas/externalServiceRegistrationManager.test.ts
```diff
@@ -206,7 +206,7 @@ describe('ExternalServiceRegistrationManager', () => {
         .mockImplementation(() => ({ directoryName: mockDirectoryName }));
       (vscode.window.showInputBox as jest.Mock).mockResolvedValue(mockFolderPath);
 
-      const result = await esrHandler.getFolderForArtifact();
+      const result = await esrHandler['getFolderForArtifact']();
```
daphne-sfdc: This change can be reverted because `getFolderForArtifact()` is a public method.

```suggestion
      const result = await esrHandler.getFolderForArtifact();
```

packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
```diff
@@ -74,9 +74,6 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
       } satisfies WorkspaceSettings;
     }
 
-    if (!this.salesforceApexExtension?.isActive) {
-      await this.salesforceApexExtension?.activate();
-    }
```
daphne-sfdc, line 79: Why isn't this needed?  I see that it's still present in **debugConfigurationProvider.ts** of **salesforcedx-vscode-apex-replay-debugger**.

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -674,7 +674,7 @@ export const processBreakpointChangedForCheckpoints = async (
 
   for (const bp of breakpointsChangedEvent.changed) {
     const breakpointId = bp.id;
-    if (bp.condition && bp.condition.toLowerCase().indexOf(CHECKPOINT) >= 0 && bp instanceof vscode.SourceBreakpoint) {
+    if (bp.condition && bp.condition.toLowerCase().includes(CHECKPOINT) && bp instanceof vscode.SourceBreakpoint) {
```
daphne-sfdc:
```suggestion
    if (bp.condition?.toLowerCase().includes(CHECKPOINT) && bp instanceof vscode.SourceBreakpoint) {
```

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -697,7 +697,7 @@ export const processBreakpointChangedForCheckpoints = async (
   }
 
   for (const bp of breakpointsChangedEvent.added) {
-    if (bp.condition && bp.condition.toLowerCase().indexOf(CHECKPOINT) >= 0 && bp instanceof vscode.SourceBreakpoint) {
+    if (bp.condition && bp.condition.toLowerCase().includes(CHECKPOINT) && bp instanceof vscode.SourceBreakpoint) {
```
daphne-sfdc:
```suggestion
    if (bp.condition?.toLowerCase().includes(CHECKPOINT) && bp instanceof vscode.SourceBreakpoint) {
```

packages/salesforcedx-vscode-core/src/commands/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+/** retrieve the active */
+export const getUriFromActiveEditor = async ({
+  message,
+  exceptionKey
+}: {
+  message: MessageKey;
+  exceptionKey: string;
+}): Promise<URI | undefined> => {
+  const editor = vscode.window.activeTextEditor;
+  if (editor && editor.document.languageId !== 'forcesourcemanifest') {
```
daphne-sfdc:
```suggestion
  if (editor?.document.languageId !== 'forcesourcemanifest') {
```

packages/salesforcedx-vscode-core/src/commands/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    return editor.document.uri;
+  }
+  const errorMessage = nls.localize(message);
+  telemetryService.sendException(exceptionKey, errorMessage);
+  await notificationService.showErrorMessage(errorMessage);
+  channelService.appendLine(errorMessage);
+  channelService.showChannelOutput();
+  return undefined;
+};
```
daphne-sfdc: Isn't this a duplicate of the function in `getUriFromActiveEditor.ts`?  Why do we need both?

packages/salesforcedx-vscode-core/src/commands/deploySourcePath.ts
```diff
@@ -58,15 +57,11 @@ export const deploySourcePaths = async (
...
-      uris = sourceUri;
-    } else {
-      uris = [sourceUri];
-    }
-  }
+
+  // When "Push-or-deploy-on-save" is enabled, the first parameter
+  // passed in (sourceUri) is actually an array and not a single URI.
+
+  const resolvedUris = uris?.length ? uris : Array.isArray(resolvedSourceUri) ? resolvedSourceUri : [resolvedSourceUri];
```
daphne-sfdc, line 64: 🚀

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
```diff
@@ -49,7 +49,7 @@ class StartApexDebugLoggingExecutor extends SfCommandletExecutor<{}> {
       const userId = await getUserId(workspaceUtils.getRootWorkspacePath());
 
       let resultJson = await this.subExecute(new QueryTraceFlag().build(userId));
-      if (resultJson && resultJson.result && resultJson.result.totalSize >= 1) {
+      if (resultJson?.result && resultJson.result.totalSize >= 1) {
```
daphne-sfdc:
```suggestion
      if (resultJson?.result?.totalSize >= 1) {
```

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
```diff
@@ -118,17 +118,13 @@ export const getUserId = async (projectPath: string): Promise<string> => {
...
   const cmdOutput = new CommandOutput();
   const result = await cmdOutput.getCmdResult(execution);
-  try {
-    const orgInfo = JSON.parse(result).result.records[0].Id;
-    return Promise.resolve(orgInfo);
-  } catch {
-    return Promise.reject(result);
-  }
+  const orgInfo = JSON.parse(result).result.records[0].Id;
+  return orgInfo;
```
daphne-sfdc: We don't want to return `result` anymore if it's not able to be parsed?

packages/salesforcedx-vscode-core/src/salesforceProject/salesforcePackageDirectories.ts
```diff
@@ -72,6 +72,6 @@ export default class SalesforcePackageDirectories {
         throw e;
       }
     }
-    return packageDirs && packageDirs.length ? packageDirs[0] : undefined;
+    return packageDirs && packageDirs.length > 0 ? packageDirs[0] : undefined;
```
daphne-sfdc:
```suggestion
    return packageDirs?.length > 0 ? packageDirs[0] : undefined;
```

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -64,7 +64,7 @@ export class QueryDataViewService {
         const errorType = 'data_view_post_message';
         const message = nls.localize('error_unknown_error', errorType);
         channelService.appendLine(message);
-        trackErrorWithTelemetry(errorType, err);
+        await trackErrorWithTelemetry(errorType, err);
```
daphne-sfdc: `trackErrorWithTelemetry()` isn't an async function.

```suggestion
        trackErrorWithTelemetry(errorType, err);
```

packages/salesforcedx-vscode-core/src/salesforceProject/salesforcePackageDirectories.ts
mshanemc: I have a better idea

packages/salesforcedx-vscode-core/src/commands/utils.ts
mshanemc: there's a subtle difference with !==

In your change, if there's no editor, the conditional would be true (undefined !== 'forcesourcemanifest')
in the original, the conditional is false if there's no editor.

That's why the linter didn't flag it or autofix it; it's a real thing.

packages/salesforcedx-vscode-core/src/commands/utils.ts
mshanemc: good catch!

packages/salesforcedx-vscode-core/src/commands/utils.ts
mshanemc: the return value would be the same (undefined) but you'd miss the error handling stuff

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
mshanemc: it is (it returns a promise!), it's just not marked correctly.  

Notice that the linter didn't fuss about using `await` here because it knows that it should be awaited!

mshanemc: I think the only reason it was was to enable `.catch(console.error)` in queryDataViewService.  

so I wrapped it in another try catch so the console.error happens with it having to be that weird faux-async

packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts
mshanemc: Promise.reject isn't returning anything.  In the original, the caller would receive the command result as the reason for the failure (which is the same as throw).  

that's drop down to https://github.com/forcedotcom/salesforcedx-vscode/blob/daf0684e728cc186439aeb7a5746adbfbbfe0b6a/packages/salesforcedx-vscode-core/src/commands/startApexDebugLogging.ts#L83 

I think it's the same error (whatever the CLI error was).

All this (querying a userId via CLI) is going away soon in another WI, so it's probably moot anyway.

packages/salesforcedx-vscode-core/src/commands/utils.ts
mshanemc: fixed it (only the input param types were different)

packages/salesforcedx-vscode-apex-debugger/src/adapter/debugConfigurationProvider.ts
mshanemc: yeah, I think I'll do those the way I've done in other places, too (one helper method that returns an activate extension so consumers don't have to fiddle with it)

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
daphne-sfdc: @mshanemc I realized that `if (msg?.length > 0) {` actually produces an error `'msg.length' is possibly 'undefined'.` because the parameter is defined as `msg?: string`.  But I can do `if (msg?.length) {`.  I know you had made some changes to be more explicit about adding `> 0` to comparisons, so I'm not sure if you'd like this change over keeping it as `if (msg && msg.length > 0) {`.

packages/salesforcedx-vscode-apex-debugger/src/context/isvContext.ts
```diff
@@ -35,3 +36,14 @@ export const registerIsvAuthWatcher = (extensionContext: vscode.ExtensionContext
...
+export const getActiveApexExtension = async (): Promise<vscode.Extension<ApexVSCodeApi>> => {
+  const salesforceApexExtension = vscode.extensions.getExtension<ApexVSCodeApi>('salesforce.salesforcedx-vscode-apex');
+  if (!salesforceApexExtension) {
+    throw new Error('Apex extension not found');
+  }
+  if (!salesforceApexExtension.isActive) {
+    await salesforceApexExtension.activate();
+  }
+  return salesforceApexExtension;
+};
```
daphne-sfdc: Why is this helper function placed in **isvContext.ts**?  It's a general utility function unrelated to ISV debugger.

mshanemc: good point.  I wanted it in a "context" kind of place and probably should create a new one.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
mshanemc: good comments, I went with `msg?.length` for the conditional.

daphne-sfdc (approved):
Manually tested the following:
1. Delete source from project and org ✅ 
2. Run LWC tests ✅ 
3. Debug LWC tests ✅ 

E2E test runs:
1. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15974402755
2. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16033658803

PR #6373, opened 2025-06-24
W-18807561 refactor: simplify-bundling

mshanemc:
> # requires https://github.com/forcedotcom/lightning-language-server/pull/608 and a bump here to consume it

#### primary
- bundle from compiled source
- bundle DX libraries like any other dependency, not their `-bundle` counterpart
- package inside the directory instead of temp dir
- use `*` for monorepo dependencies
- inter-extension dependencies via typed export (just apex and core ext).  affects devDeps and tsconfigs
- special handling for non-explicitly-depended resources (ex: templates, aura resources, tern)
- remove various path modifications (serverPath)
- remoe all package-time package.json modifications (`packaging` stanza)
- esbuild using mjs 

#### other
- gha-configurable node version (remove `.nvmrc`
- optional chaining
- clean up some unused scripts

@W-18807561@


mshanemc:
big problems:

lightning-language-server has a bunch of dynamic file reference issues (possibly fixable, but a big PR on its own!)
json-path/jsonc-parser aren't bundleable.

mshanemc:
big problems

lightning-language-server dynamic file loading (tern, aura/LWC resources). Can be done statically, that's a whole separate PR
json-path, jsonc-parser are also doing unbundle-able things. as is bfj that we use for json stream parsing

PR #6381, opened 2025-06-25
test: show the test name

mshanemc:
[skip-validate-pr]

what if...you could see which test was running

see https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15887534777
Screenshot 2025-06-25 at 4 26 15 PM

### DIFF
--- a/.github/workflows/runE2ETest.yml
+++ b/.github/workflows/runE2ETest.yml
@@ -32,7 +32,7 @@ on:
         type: string
 jobs:
   build:
-    name: ${{ inputs.testToRun }}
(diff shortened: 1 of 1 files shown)


.github/workflows/runE2ETest.yml
```diff
@@ -32,6 +32,7 @@ on:
         type: string
 jobs:
   build:
+    name: ${{ inputs.testToRun }} (${{ matrix.os }})
```
madhur310: shall we add the node and vscode version too?

madhur310 (approved):

PR #6386, opened 2025-06-28
[W-18589337] refactor: turn on apex debug log for replay debugger without calling the CLI command

daphne-sfdc:
### What does this PR do?
1. Refactors the code for **SFDX: Turn On Apex Debug Log for Replay Debugger** to not call the CLI command.
2. Corrects existing bugs in the behavior when switching between orgs.  The following use cases were tested in addition to the sunny day cases:
a. Reload the VSCode window
b. Handle expired trace flags
c. Set a default org
d. Authorize to an org
e. Logout from an org
f. Create a scratch org
g. Delete a scratch org

Test plan here: https://salesforce.quip.com/J7TIA322cOdj

@W-18589337@


packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -78,6 +78,34 @@ export class WorkspaceContextUtil {
     return connection;
   }
 
+  /**
+   * Creates a connection for the current target org, bypassing any caching.
```
peternhale: @daphne-sfdc while I understand the reason for adding this, it is a ["code smell"](https://en.wikipedia.org/wiki/Code_smell), pointing toward something missing when handling either an init problem at extension start or in the handling of a new target-org.

packages/salesforcedx-vscode-core/src/commands/auth/orgLoginWeb.ts
```diff
@@ -129,6 +134,10 @@ export class OrgLoginWebContainerExecutor extends SfCommandletExecutor<AuthParam
 class OrgLoginWebExecutor extends SfCommandletExecutor<AuthParams> {
   protected showChannelOutput = false;
 
+  constructor(extensionContext?: vscode.ExtensionContext) {
```
mshanemc: this is going to get flagged by that eslint rule for no-useless-constructor.  SfCommandletExecutor doesn't have a constructor so this does nothing

packages/salesforcedx-vscode-core/src/commands/auth/orgLoginWeb.ts
```diff
@@ -141,9 +150,44 @@ class OrgLoginWebExecutor extends SfCommandletExecutor<AuthParams> {
 
     return command.build();
   }
+
```
mshanemc: I don't understand why this is added.  Isn't it the same as what's inherited from the SfCommandletExecutor?

peternhale (approved):
Just an observation

mshanemc:
QA: turned on debugging, 
✅  I see the bottom bar element.
✅ logs are generated

turn off debug logs via command
✅ element disappears
✅ `sf data:query --query "SELECT Id FROM ApexDebuggerSession WHERE Status = 'Active' LIMIT 1" --use-tooling-api --json` produces no results

turn on debugging
delete scratch org
✅ element disappears

seems good!

mshanemc (approved):

PR #6391, opened 2025-07-02
fix: W-18939319 - manifest generation is cancelled when escaping

CristiCanizales:
### What does this PR do?
This pull request refactors and improves the `GenerateManifestExecutor` class and related functionality in the `projectGenerateManifest.ts` file. The changes enhance user input handling, simplify the code by removing unused methods, and add a new user-facing message for cancellation scenarios.

### Refactoring and Simplification:
* Removed the `openUntitledDocument` method, as it is no longer needed. The logic now ensures that the manifest is either saved with a user-provided name or a default name.
* Updated the `responseText` property in the `GenerateManifestExecutor` class to be non-optional, simplifying its usage and ensuring it is always a string.

### Improved User Input Handling:
* Added a check to handle user cancellation (e.g., pressing Escape) when entering a filename. If the input is canceled, a warning message is displayed, and the process stops.
* Adjusted the logic for saving the document to handle cases where the user provides an empty string as the filename, using a default name in such cases.

### User Messaging:
* Added a new localized message, `manifest_generation_cancelled`, to inform users when the manifest generation process is canceled.

@W-18939319@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/projectGenerateManifest.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/projectGenerateManifest.ts
@@ -28,8 +28,8 @@ const MANIFEST_SAVE_PROMPT = 'manifest_input_save_prompt';
 
 class GenerateManifestExecutor extends LibraryCommandletExecutor<string> {
   private sourcePaths: string[];
-  private responseText: string | undefined;
-  constructor(sourcePaths: string[], responseText: string | undefined) {
+  private responseText: string;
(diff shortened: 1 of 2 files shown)


daphne-sfdc (approved):
1. Generate manifest with a filename ✅ 
2. Generate manifest without inputting a filename -> manifest created with default name "package.xml" ✅ 
3. Try to generate manifest with name that already exists -> get error message ✅ 
4. Cancel the command by hitting the Escape key ✅ 
Screenshot 2025-07-02 at 10 51 08 AM

PR #6396, opened 2025-07-03
@W-18962079 refactor: unused-vars

mshanemc:
### primary:
- enable eslint rule to error unused vars
- there's an allowance for underscore-prefixed args `_foo` that are **intentionally** unused.  Usually, it's because we have to match an interface for OOP `extends` reasons, but should keep us from introducing more of them unintentionally.
- clean up some truly unused things

### other
dedupe: core ext uses SfWorkspaceChecker from utils instead of its own local copy
removed the Predicate stuff
removed the unused stuff related to `getTelemetryData` on sfCommandletExecutor (also verified a4d ext isn't using it)
fix some promise handling (async vs. not)

@W-18962079@


packages/salesforcedx-utils/test/jest/i18n/advancedLocalization.test.ts
```diff
@@ -31,35 +31,35 @@ describe('Advanced Localization Unit Tests', () => {
...
       expect(config1).toBe(config2);
     });
 
     it('should return the same instance for different instance names', () => {
-      const config1 = LocalizationConfig.getInstance('test1');
-      const config2 = LocalizationConfig.getInstance('test2');
+      const config1 = LocalizationConfig.getInstance();
+      const config2 = LocalizationConfig.getInstance();
       expect(config1).toBe(config2);
     });
```
daphne-sfdc: This test case is no longer relevant, as it is identical to the "should create a singleton instance" case above.

packages/salesforcedx-utils-vscode/src/commands/preconditionCheckers.ts
```diff
@@ -5,34 +5,28 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import * as fs from 'node:fs';
```
daphne-sfdc: We should use the equivalent VSCode function instead of node:fs here, right?

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
```diff
@@ -479,15 +479,15 @@ export class CheckpointNode extends BaseNode {
   private actionObjectId: string | undefined;
 
   constructor(
-    breapointIdInput: string,
+    breakpointIdInput: string,
```
daphne-sfdc, line 482: Good catch on the typo!

daphne-sfdc (commented):
E2E test run at this point: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16058138138

packages/salesforcedx-vscode-apex-replay-debugger/src/breakpoints/checkpointService.ts
mshanemc: credit goes to https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

packages/salesforcedx-utils-vscode/src/commands/preconditionCheckers.ts
mshanemc: yes, great catch.  small windows between now and Cristi's linter PR going in, but that's why we need that rule!

daphne-sfdc (approved):
**E2E test run:** https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16060105919

**Manual testing:**

Mac:
- Deploy and Retrieve ✅ 
- Push and Pull ✅ 

Ubuntu:
- Create OAS Doc ✅ 
- Deploy and Retrieve ✅ 
- Push and Pull ✅ 

Windows:
- Create OAS Doc ✅ 
- Apex LSP ✅ 
- Debug Apex Tests ✅ 
- Run Apex Tests ✅ 
- Deploy and Retrieve ✅ 
- Manifest Builder ✅ 
- Push and Pull ✅ 
- LWC LSP ✅ 
- Debug LWC Tests ✅ 
- Run LWC Tests ✅

PR #6397, opened 2025-07-03
@W-18759649 feat: less eager ext activations

mshanemc:
### What does this PR do?
see activationEvents.md

this will affect e2e tests.  Instead of broadly asserting activation of all extensions, tests can assert activation of anything they are concerned with and should be agnostic to the rest.

#<Insert GitHub Issue>, @W-18759649@


packages/salesforcedx-vscode-automation-tests/test/utils/copilotHidingHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+import { Duration, log, pause } from '@salesforce/salesforcedx-vscode-test-tools/lib/src/core';
+import {
+  clickButtonOnModalDialog,
+  executeQuickPick
+} from '@salesforce/salesforcedx-vscode-test-tools/lib/src/ui-interaction';
+
+export const tryToHideCopilot = async (): Promise<void> => {
+  try {
+    await executeQuickPick('Chat: Hide Copilot');
```
peternhale, line 15: @mshanemc I looked for this command in vscode with chat extension enabled. I cannot find it.
I believe the chat view, by default appears in the "secondary side bar". The command "View: Hide Secondary Side Bar" might work.
The command is only available in the palette when the view is open.

madhur310: Yeah - it is only available when the view is open, hence I put it inside a try catch so if it fails I still try and run the tests. I didn't want to use the secondary sidebar, because I didn't want to open the side bar if its not already open and cause some inadvertent failures

madhur310: The command "View: Hide Secondary Side Bar" is also only available when the sidebar is open, so behaves quite similarly. Any specific reason to use that for "Chat: Hide Copilot"?

madhur310: There is a toggle side bar command that is always open, which would cause side bar to open.

PR #6400, opened 2025-07-04
feat: W-18338854 - execute soql query commands don't rely on CLI anymore

CristiCanizales:
### What does this PR do?
This pull request introduces significant enhancements to the SOQL query execution functionality in the `dataQuery.ts` file. The changes include refactoring the `DataQueryExecutor` class to use the `LibraryCommandletExecutor`, improving error handling, introducing CSV export capabilities, and updating i18n messages for better user feedback.

### Enhancements to SOQL Query Execution:

* Refactored `DataQueryExecutor` to use `LibraryCommandletExecutor` for improved handling of SOQL queries. Added methods for executing queries, displaying results in a table format, converting results to CSV, and saving them to a file. Enhanced error handling to provide specific messages for common issues, such as session expiration or malformed queries.

### Improvements to API Type Handling:

* Updated `ApiType` enum to use uppercase naming (`TOOLING` instead of `Tooling`) for consistency.
* Adjusted references to `ApiType.Tooling` in the `GetQueryAndApiInputs` class to align with the updated enum naming convention.

### Updates to i18n Messages:

* Added new localized messages for success and error scenarios, including detailed error descriptions for cases like invalid login credentials, insufficient access rights, and connection issues. These messages improve user feedback during query execution.

@W-18338854@

### DIFF
--- a/packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
@@ -4,6 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { ConfigAggregator } from '@salesforce/core-bundle';
 import { Connection, Tooling } from '@salesforce/core-bundle/org/connection';
 import {
   CancelResponse,
@@ -31,10 +32,11 @@ interface QueryResult {
 }
 
 class DataQueryExecutor extends LibraryCommandletExecutor<QueryAndApiInputs> {
-  private readonly MAX_FETCH = 50_000;
-
   constructor() {
     super(nls.localize('data_query_input_text'), 'data_soql_query_library', OUTPUT_CHANNEL);
+    // Disable automatic success notifications since we show our own custom success notification
+    // Keep failure notifications enabled for automatic error handling
+    this.showSuccessNotifications = false;
   }
 
   public async run(response: ContinueResponse<QueryAndApiInputs>): Promise<boolean> {
@@ -45,32 +47,13 @@ class DataQueryExecutor extends LibraryCommandletExecutor<QueryAndApiInputs> {
       const connection = await WorkspaceContext.getInstance().getConnection();
 
       // Execute query using the appropriate API
-      const queryResult = await this.runSoqlQuery(api === ApiType.TOOLING ? connection.tooling : connection, query);
+      const queryResult = await this.runSoqlQuery(api === 'TOOLING' ? connection.tooling : connection, query);
 
       // Display results in table format
       this.displayTableResults(queryResult);
 
-      // Convert results to CSV and save to file
-      const csvContent = this.convertToCSV(queryResult);
-
-      // Generate filename with timestamp
-      const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
-      const fileName = `soql-query-${timestamp}.csv`;
-
-      // Get workspace path and create output directory
-      const outputDir = path.join(workspaceUtils.getRootWorkspacePath(), '.sfdx', 'data');
-      const filePath = path.join(outputDir, fileName);
-
-      // Write CSV file
-      await writeFile(filePath, csvContent);
-
-      // Open the file in editor
-      const document = await vscode.workspace.openTextDocument(filePath);
-      await vscode.window.showTextDocument(document);
-
-      // Show success message
-      const recordCount = queryResult.totalSize;
-      channelService.appendLine(nls.localize('data_query_success_message', recordCount.toString(), filePath));
+      // Save results to CSV file and show notification
+      await this.saveResultsToCSV(queryResult);
 
       return true;
     } catch (error) {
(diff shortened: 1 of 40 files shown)


packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+      const connection = await WorkspaceContext.getInstance().getConnection();
+
+      // Execute query using the appropriate API
+      const queryResult = await this.runSoqlQuery(api === ApiType.TOOLING ? connection.tooling : connection, query);
+
+      // Display results in table format
+      this.displayTableResults(queryResult);
+
+      // Convert results to CSV and save to file
+      const csvContent = this.convertToCSV(queryResult);
```
peternhale: @CristiCanizales as with displayTableResults, the CSV file activities should be in a distinct method.

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+      // Get workspace path and create output directory
+      const outputDir = path.join(workspaceUtils.getRootWorkspacePath(), '.sfdx', 'data');
+      const filePath = path.join(outputDir, fileName);
+
+      // Write CSV file
+      await writeFile(filePath, csvContent);
+
+      // Open the file in editor
+      const document = await vscode.workspace.openTextDocument(filePath);
+      await vscode.window.showTextDocument(document);
```
peternhale: Running queries is an async task, do we really want to force a focus change in the UI?

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+
+      // Show success message
+      const recordCount = queryResult.totalSize;
+      channelService.appendLine(nls.localize('data_query_success_message', recordCount.toString(), filePath));
```
peternhale:
```suggestion
      channelService.appendLine(nls.localize('data_query_success_message', queryResult.totalSize, filePath));
```

packages/salesforcedx-vscode-core/src/messages/i18n.ts
```diff
@@ -166,6 +166,20 @@ export const messages = {
   debugger_stop_none_found_text: 'No Apex Debugger session found.',
   data_query_input_text: 'SFDX: Execute SOQL Query...',
   data_query_selection_text: 'SFDX: Execute SOQL Query with Currently Selected Text',
+  data_query_success_message: 'Query executed successfully. Found %s records. Results saved to: %s',
```
peternhale:
```suggestion
  data_query_success_message: 'Query executed successfully. Found %d records. Results saved to: %s',
```

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
peternhale: The localize now accepts all of the primitive types, so no need to toString() a number.

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+
+    const result = await connection.query(query, {
+      autoFetch: true,
+      maxFetch: this.MAX_FETCH,
+      scanAll: false
+    });
+
+    if (result.records.length > 0 && result.totalSize > result.records.length) {
+      const missingRecords = result.totalSize - result.records.length;
+      channelService.appendLine(
```
peternhale, line 83: Should the text be added to i18n?

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+        `Warning: The query result is missing ${missingRecords} records due to a ${this.MAX_FETCH} record limit. ` +
+          `Increase the number of records returned by setting the config value "org-max-query-limit" or the environment variable "SF_ORG_MAX_QUERY_LIMIT" to ${result.totalSize} or greater than ${this.MAX_FETCH}.`
+      );
+    }
+
+    channelService.appendLine(`Query complete with ${result.totalSize} records returned`);
```
peternhale: i18n

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -81,7 +297,7 @@ type QueryAndApiInputs = {
 
 enum ApiType {
```
peternhale: unions over enums

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+      return false;
+    }
+  }
+
+  private async runSoqlQuery(connection: Connection | Tooling, query: string): Promise<QueryResult> {
+    channelService.appendLine('Running query...');
+
+    const result = await connection.query(query, {
+      autoFetch: true,
+      maxFetch: this.MAX_FETCH,
```
peternhale: While testing this change with sf config value "org-max-query-limit" to 2000, the query returned 50_000
```
sf config list
List Config
┌─────────────────────┬──────────┬───────────────┐
│ Name                │ Location │ Value         │
├─────────────────────┼──────────┼───────────────┤
│ org-max-query-limit │ Local    │ 2000          │

```
Account29791                  

Query executed successfully. Found 51003 records. Results saved to: /Users/peter.hale/git/dreamhouse-lwc/.sfdx/data/soql-query-2025-07-07T12-50-28-022Z.csv

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+}
+
+class DataQueryExecutor extends LibraryCommandletExecutor<QueryAndApiInputs> {
+  private readonly MAX_FETCH = 50_000;
+
+  constructor() {
+    super(nls.localize('data_query_input_text'), 'data_soql_query_library', OUTPUT_CHANNEL);
+  }
+
+  public async run(response: ContinueResponse<QueryAndApiInputs>): Promise<boolean> {
```
peternhale, line 39: @CristiCanizales this change needs to consider large queries. We give the user full control of max result size and they are able to write any query that can be run via APIs.

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
mshanemc: I think you can omit maxFetch.  sfdx-core/Connection is going to read the env/config and that's what we want.

mshanemc: alternative: connection.autoFetchQuery that has the built-in handling of the config/env 
https://github.com/forcedotcom/sfdx-core/blob/4110e5e3b87cd9d68e966edf7bc0b2649e4b49a4/src/org/connection.ts#L356-L357

for run, subscribe to lifecycle warnings and get the message stuff for free
https://github.com/forcedotcom/sfdx-core/blob/4110e5e3b87cd9d68e966edf7bc0b2649e4b49a4/src/org/connection.ts#L374

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+      return '';
+    }
+    if (typeof value === 'object') {
+      // Handle nested objects (like relationship fields)
+      return JSON.stringify(value);
+    }
+    return String(value);
+  }
+
+  private formatFieldValueForDisplay(value: any): string {
```
mshanemc: private methods like this can be simple functions since they don't refer to stuff from the class's properties.  That makes it easier to catch dead code in the future.

then if you want to write UT for them (and you should!) then it's easy to export the fn and call it from the test.  

ex: with a little refactoring, you could have a fn that returns the table output.  then pass it some data and assert the table output or use a jest snapshot.  

then you just have to get the table output back and do the appendline.

Same with csv and its escaping.

it's a lot more unit-testable than having to mock `channelService` etc.

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,23 +4,239 @@
...
+    }));
+
+    // Create table rows
+    const rows: Row[] = records.map((record: { [x: string]: any }) => {
+      const row: Row = {};
+      fields.forEach(field => {
+        const value = record[field];
+        row[field] = this.formatFieldValueForDisplay(value);
+      });
+      return row;
```
mshanemc: see Object.fromEntries

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
mshanemc: good point. `preserveFocus:false` would open the document but not steal focus?  Or are you thinking of something else?

CristiCanizales: I was thinking to show the success notification and a "Open file" button in case the user wants to review it. Wdyt?

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
CristiCanizales: You're totally right. Thanks Shane!

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
mshanemc: sounds good

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -4,34 +4,159 @@
...
+    }
+
+    // Priority 2: Check environment variable as fallback (SF_ORG_MAX_QUERY_LIMIT)
+    const envValue = process.env.SF_ORG_MAX_QUERY_LIMIT;
+    if (envValue) {
+      const parsed = parseInt(envValue, 10);
+      if (!isNaN(parsed) && parsed > 0) {
+        return parsed;
+      }
+    }
```
mshanemc: ConfigAggregator is already doing this.

it's also supporting the old SFDX_ORG_MAX_QUERY_LIMIT.


```suggestion
```

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -96,3 +218,210 @@ export const dataQuery = (): void => {
...
+      return '';
+    }
+    return fields
+      .map(field => {
+        const value = record[field];
+        return escapeCSVField(formatFieldValue(value));
+      })
+      .join(',');
+  });
+
```
mshanemc, line 259: as is, it would leave a blank row.  I couldn't reproduce this by getting bad data from the server and I don't think it's a real scenario, so I just skipped that row instead of leaving it blank

mshanemc (commented):
code: 
❌ linter warnings are no good.  I'm going to fix those with some PR/types changes.

QA notes:
✅ queries fine when entering the query
✅ good error when I do an invalid query
❌  see comment about blank line in output being non-reproducible
✅ did `SELECT COUNT() FROM Property__c` from highlighted text in dreamhouse repo, got fine result

PR #6403, opened 2025-07-07
[W-18337597] refactor: remove deprecated ConfigGet

daphne-sfdc:
### What does this PR do?
Replaces the deprecated `ConfigGet` with `ConfigAggregator` from `@salesforce/core-bundle`.  No change in functionality.

@W-18337597@

### DIFF
--- a/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
+++ b/packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
@@ -148,7 +148,7 @@ export class ApexVariable extends Variable {
   }
 
   public static valueAsString(value: Value): string {
-    if (typeof value.value === 'undefined' || value.value === null) {
+    if (value.value === undefined || value.value === null) {
       // We want to explicitly display null for null values (no type info for strings).
       return 'null';
     }
@@ -1149,7 +1149,7 @@ export class ApexDebug extends LoggingDebugSession {
     requestId: string,
     apexId: number | undefined
   ): Promise<number | undefined> {
-    if (typeof apexId === 'undefined') {
+    if (apexId === undefined) {
       return;
     }
     if (!this.variableContainerReferenceByApexId.has(apexId)) {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -71,6 +71,7 @@ export default [
       'unicorn/consistent-function-scoping': 'error',
       'unicorn/explicit-length-check': 'error',
       'unicorn/no-instanceof-builtins': 'error',
+      'unicorn/no-typeof-undefined': 'error',
       'unicorn/no-useless-fallback-in-spread': 'error',
       'unicorn/no-useless-length-check': 'error',
       'unicorn/no-useless-promise-resolve-reject': 'error',
--- a/packages/salesforcedx-apex-debugger/test/unit/adapter/apexDebugVariablesHandling.test.ts
+++ b/packages/salesforcedx-apex-debugger/test/unit/adapter/apexDebugVariablesHandling.test.ts
@@ -872,7 +872,7 @@ export const newStringValue = (name: string, value = 'value', slot?: number): Va
     nameForMessages: 'String',
     value
   };
-  if (typeof slot !== 'undefined') {
+  if (slot !== undefined) {
     result.slot = slot;
   }
   return result;
--- a/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
+++ b/packages/salesforcedx-utils-vscode/src/cli/commandExecutor.ts
@@ -38,7 +38,7 @@ export class CliCommandExecutor {
     }
 
     // then specific environment from Spawn Options
-    if (typeof options.env !== 'undefined') {
+    if (options.env !== undefined) {
       Object.assign(env, options.env);
     }
 
--- a/packages/salesforcedx-utils-vscode/src/services/telemetry.ts
+++ b/packages/salesforcedx-utils-vscode/src/services/telemetry.ts
@@ -168,7 +168,7 @@ export class TelemetryService implements TelemetryServiceInterface {
   }
 
   public async checkCliTelemetry(): Promise<boolean> {
-    if (typeof this.cliAllowsTelemetryPromise !== 'undefined') {
+    if (this.cliAllowsTelemetryPromise !== undefined) {
       return this.cliAllowsTelemetryPromise;
     }
     this.cliAllowsTelemetryPromise = isCLITelemetryAllowed();
--- a/packages/salesforcedx-utils/src/cli/cliCommandExecutor.ts
+++ b/packages/salesforcedx-utils/src/cli/cliCommandExecutor.ts
@@ -32,7 +32,7 @@ export class CliCommandExecutor {
     }
 
     // then specific environment from Spawn Options
-    if (typeof options.env !== 'undefined') {
+    if (options.env !== undefined) {
       Object.assign(env, options.env);
     }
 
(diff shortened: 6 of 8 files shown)


.vscode/settings.json
```diff
@@ -33,5 +33,6 @@
   "prettier.useTabs": false,
   "prettier.printWidth": 120,
   "prettier.arrowParens": "avoid",
-  "editor.rulers": [120]
+  "editor.rulers": [120],
+  "cSpell.words": ["debugadapter", "debugprotocol", "salesforcedx", "STREAMINGAPI", "testsupport"]
```
daphne-sfdc: As @mshanemc suggested, I installed the **Code Spell Checker** VSCode extension.  These are some words in our code that had blue squiggly lines, which I added to the dictionary.  As I come across more blue squigglies in our code, I'll continue to populate this list, and I'm hoping to see contributions from the team as we visit files across the codebase in our refactoring. 🙂

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -567,13 +567,9 @@ export class ApexDebug extends LoggingDebugSession {
...
     try {
       if (args.connectType === CONNECT_TYPE_ISV_DEBUGGER) {
-        const config = await new ConfigGet().getConfig(
-          args.salesforceProject,
-          SF_CONFIG_ISV_DEBUGGER_SID,
-          SF_CONFIG_ISV_DEBUGGER_URL
-        );
-        const isvDebuggerSid = config.get(SF_CONFIG_ISV_DEBUGGER_SID);
-        const isvDebuggerUrl = config.get(SF_CONFIG_ISV_DEBUGGER_URL);
+        const localConfig = await Config.create({ isGlobal: false });
```
mshanemc: the library supports both global and project level configs and we want to respect those.  

this code will ignore global setting and only read the local.

We want to use ConfigAggregator (see the extensions ConfigUtil for example).  It properly merges global and local and lets you get the data.

it might involve creating a new method on ConfigUtil for getting these ISV props off of it.  OR you could use ConfigAggregator directly.

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -14,6 +14,7 @@
   ],
   "dependencies": {
     "@salesforce/salesforcedx-utils": "64.4.0",
+    "@salesforce/salesforcedx-utils-vscode": "64.4.0",
```
gbockus-sf: So this is the danger zone.  `salesforcedx-utils-vscode` is the module that depends on the vscode import and can be utilized by extension code.   The debugger is compiled down to a stand alone script and runs outside the extension host.   

Unless things have changed I think this will fail at runtime b/c apexDebug -> configAggregatorProvider -> workspaceUtils -> vscode 💥 

This apex-debugger package should really only depend on salesforcedx-utils.

daphne-sfdc: Yes this does fail at runtime.  The ISV Debugger fails to start when I try to launch a debugging session.  I had just checked in the code at this point for recordkeeping.

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -567,13 +567,11 @@ export class ApexDebug extends LoggingDebugSession {
...
-          args.salesforceProject,
-          SF_CONFIG_ISV_DEBUGGER_SID,
-          SF_CONFIG_ISV_DEBUGGER_URL
-        );
-        const isvDebuggerSid = config.get(SF_CONFIG_ISV_DEBUGGER_SID);
-        const isvDebuggerUrl = config.get(SF_CONFIG_ISV_DEBUGGER_URL);
+        const configAggregator: ConfigAggregator = await ConfigAggregator.create({
+          projectPath: args.salesforceProject
+        });
+        const isvDebuggerSid = JSON.stringify(configAggregator.getPropertyValue(SF_CONFIG_ISV_DEBUGGER_SID));
```
mshanemc: getPropertyValue takes a generic type Param.  so If you know the 2 props you want are string, you can say `getPropertyValue<string>(SF_)`.  the stringify isn't necessary.

packages/salesforcedx-apex-debugger/src/context/isvContextUtil.ts
```diff
@@ -18,13 +18,11 @@ export class IsvContextUtil {
...
-        projectWorkspacePath,
-        SF_CONFIG_ISV_DEBUGGER_SID,
-        SF_CONFIG_ISV_DEBUGGER_URL
-      );
-      const isvDebuggerSid = config.get(SF_CONFIG_ISV_DEBUGGER_SID);
-      const isvDebuggerUrl = config.get(SF_CONFIG_ISV_DEBUGGER_URL);
+      const configAggregator: ConfigAggregator = await ConfigAggregator.create({
+        projectPath: projectWorkspacePath
+      });
+      const isvDebuggerSid = JSON.stringify(configAggregator.getPropertyValue(SF_CONFIG_ISV_DEBUGGER_SID));
```
mshanemc: it's subtle, but `JSON.stringify('foo')` => `"'foo'"`
I think it might not matter if you run it through the env, which is another topic.

packages/salesforcedx-apex-debugger/src/context/isvContextUtil.ts
```diff
@@ -18,13 +18,11 @@ export class IsvContextUtil {
...
+      const isvDebuggerUrl = JSON.stringify(configAggregator.getPropertyValue(SF_CONFIG_ISV_DEBUGGER_URL));
 
       if (typeof isvDebuggerSid !== 'undefined' && typeof isvDebuggerUrl !== 'undefined') {
         // set auth context
```
mshanemc: what do you think the point of `// set auth context` is?

hint: peek at how `setIsvDebuggerContext` is called from `setupGlobalDefaultUserIsvAuth`.  What would happen if those weren't set?

would the reset still be necessary?

packages/salesforcedx-apex-debugger/src/context/isvContextUtil.ts
```diff
@@ -18,13 +18,11 @@ export class IsvContextUtil {
   public async setIsvDebuggerContext(projectWorkspacePath: string) {
     let isvDebugProject = false;
```
mshanemc: could it return true or false instead of using `let` ?

packages/salesforcedx-apex-debugger/src/context/isvContextUtil.ts
mshanemc: and the perennial question...why is this a class?

mshanemc: how simple could this code be if it only needs to know if the two props are defined to return a boolean?

mshanemc: and does it really belong in the debugger, or in the ext that needs it?

packages/salesforcedx-apex-debugger/src/context/isvContextUtil.ts
daphne-sfdc: I did notice that - which is why I had added the double quotes in **apexDebug.test.ts** to make the tests pass.  But now that I'm using `getPropertyValue<string>(SF_)` I don't need the `JSON.stringify()` anymore and thus removed the double quotes.

mshanemc (approved):
Screenshot 2025-07-10 at 3 46 16 PM
good ratio!

PR #6405, opened 2025-07-08
W-18989201 fix: file parsing bug

mshanemc:
### What does this PR do?
fix a types bug that **probably** causes apex test diagnostics to not be applied to a file.  I don't know how to reproduce this or exactly what it should look like, but I know the code is wrong!

@W-18989201@


packages/salesforcedx-vscode-apex/src/commands/apexTestRunCodeAction.ts
```diff
@@ -153,7 +153,10 @@ export class ApexLibraryTestRunExecutor extends LibraryCommandletExecutor<{}> {
               .map(pkgDir => `${path.relative(workspaceFolder.uri.fsPath, pkgDir.fullPath)}/**/*.cls`)
               .flatMap(pattern => vscode.workspace.findFiles(pattern, '**/node_modules/**'))
           )
-        ).map(file => file.toString())
+        )
+          .flat()
+          // parsing to string for set to dedupe, then back to URI
+          .map(uri => uri.toString())
```
mshanemc, line 159: this slipped by the compiler because without `flat` you have `URI[]` instead of `URI`

BUT array has a `toString` method so the compiler thought this was fine.

PR #6407, opened 2025-07-09
refactor: fs

mshanemc:
pr feedback


madhur310 (approved):
just documenting, like we discussed in slack we should delete the removeFolder method itself too

PR #6412, opened 2025-07-12
[W-19015026] fix: deploy/retrieve works on manifests in command palette

daphne-sfdc:
### What does this PR do?
Fixes a bug where **SFDX: Deploy Source in Manifest to Org** and **SFDX: Retrieve Source in Manifest from Org** were producing errors when run in command palette.

@W-19015026@

### Functionality Before
Screenshot 2025-07-11 at 12 47 30 PM

### Functionality After
Commands can be run successfully.


CristiCanizales (approved):
Thanks for the fix! I verified deploy and retrieve from manifest now work as expected ✅ 

:shipit:

PR #6415, opened 2025-07-15
[W-19004812] fix: "SFDX: Update Checkpoints in Org" works when the default org is set as a global config

daphne-sfdc:
### What does this PR do?
Fixes a bug where **SFDX: Update Checkpoints in Org** was producing an error when the default org is set as a global config instead of a local config.  There should be no noticeable difference for the global vs. local `target-org` config variable.

@W-19004812@

### Functionality Before
Screenshot 2025-07-14 at 4 42 57 PM

### Functionality After
Screenshot 2025-07-14 at 9 19 05 PM


packages/salesforcedx-utils/src/cli/orgDisplay.ts
```diff
@@ -16,34 +16,36 @@ export class OrgDisplay {
...
       return this.username;
     }
 
     // Try to get username from project config
     try {
-      const config = await Config.create(Config.getDefaultOptions());
-      const username = config.get('target-org');
-      if (username && typeof username === 'string') {
-        return username;
+      const configAggregator: ConfigAggregator = await ConfigAggregator.create();
```
mshanemc: seems like a great place to pass in the project root instead of relying on the default cwd

packages/salesforcedx-vscode-automation-tests/test/specs/trailApexReplayDebugger.e2e.ts
```diff
@@ -86,8 +86,7 @@ describe('"Find and Fix Bugs with Apex Replay Debugger" Trailhead Module', () =>
     expect(outputPanelText).to.contain('Expected: CRM, Actual: SFDC');
   });
 
-  // This is broken and will be fixed with W-19004812 https://gus.lightning.force.com/a07EE00002Ht4taYAB
-  it.skip('Set Breakpoints and Checkpoints', async () => {
```
mshanemc, line 91: nice!

mshanemc (approved):
nice UT

packages/salesforcedx-utils/src/cli/orgDisplay.ts
daphne-sfdc: I did try to pass in the project root like I previously did in **apexDebug.ts**, but this class doesn’t offer any class-level variable for me to pass in. **apexDebug.ts** offered `args.salesforceProject`, which doesn’t have anything similar in this class. Would you suggest I make such a variable?

mshanemc: getUsername could take an optional string param.  That way any consumer that has one could use it.

packages/salesforcedx-utils/src/cli/orgDisplay.ts
```diff
@@ -15,39 +15,43 @@ export class OrgDisplay {
...
+      if (configUsernameOrAlias && typeof configUsernameOrAlias === 'string') {
+        usernameOrAlias = configUsernameOrAlias;
       }
     } catch {
       // Ignore config errors
     }
 
-    // Try to get from state aggregator
+    if (!usernameOrAlias) {
+      throw new Error('No username provided and no default username found in project config or state');
```
CristiCanizales: can you move this to i18n file please?

scripts/vsce-bundled-extension.ts
```diff
@@ -87,15 +87,25 @@ logger(`Now in ${cwd}`);
 logger('executing npm install');
 execSync('npm install', { stdio: 'inherit', cwd });
 
+// Clean up any existing VSIX files from previous builds
+logger('cleaning up existing VSIX files');
```
CristiCanizales, line 91: 🚀

daphne-sfdc: I made that change to the script because I was having trouble running `npm run vscode:package`.  I was getting weird errors whenever I tried to build VSIXs to run the E2E tests locally until I isolated the issue to `lerna run vscode:package --concurrency 1`.  Cursor was smart enough to tell me that I had an older set of VSIXs with the previous version v64.4.0 that was preventing new VSIXs with version v64.5.0 from being generated. 🤡

CristiCanizales (approved):

CristiCanizales (changes requested):
I can see some e2e tests failing, can you please take a look at them?

daphne-sfdc:
> I can see some e2e tests failing, can you please take a look at them?

Looks like a flapper in the `orgLoginSfdxUrl` CLI command.  I reran the failing E2E tests and they all pass.

CristiCanizales (approved):
Thanks for fixing this! 😄 
- SFDX: Update Checkpoints in Org ✅ 
- SFDX: Display Org Details for Default Org ✅ 
- SFDX: Display Org Details... ✅

PR #6426, opened 2025-07-21
W-19047827: fix: dependabot e2e fix

madhur310:
### What does this PR do?
Update runE2E yml to use github.token to download artifacts since we are doing so from current repo/run. 

@W-19047827@

passing e2e tests for dependabot in ci-testing repo https://github.com/forcedotcom/salesforcedx-vscode-ci-testing/pull/9

### DIFF
--- a/.github/workflows/runE2ETest.yml
+++ b/.github/workflows/runE2ETest.yml
@@ -161,7 +161,7 @@ jobs:
         with:
           max_attempts: 5
           command: bash download-vscode.sh "${{ matrix.vscodeVersion }}"
-          retry_wait_seconds: 30
+          retry_wait_seconds: 300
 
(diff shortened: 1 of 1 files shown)


.github/workflows/runE2ETest.yml
```diff
@@ -161,7 +161,7 @@ jobs:
         with:
           max_attempts: 5
           command: bash download-vscode.sh "${{ matrix.vscodeVersion }}"
-          retry_wait_seconds: 30
+          retry_wait_seconds: 300
```
madhur310, line 164: This is not technically required, but I felt it helped in some network issues where retry would trigger too soon.

daphne-sfdc (approved):

PR #6427, opened 2025-07-21
[W-19101015] ci: delete scratch org infos for deleted orgs in svcideebot org automatically once per day to avoid storage limit exceeded error

daphne-sfdc:
### What does this PR do?
Every time a new scratch org is created from the svcideebot dev hub, a scratch org info entry is created in the svcideebot org.  Each E2E test run creates and deletes a scratch org, generating over 50 scratch org info entries.  In the past, we have been manually running a query to delete the scratch org infos from the org whenever we get the "storage limit exceeded" error, but this is becoming a much more frequent chore now that E2E runs on every commit.  Therefore, I created a new GHA workflow that automatically runs the query to do the cleanup for us once per day.

Here's a successful run of the workflow: https://github.com/daphne-sfdc/salesforcedx-vscode-cleanup-deleted-scratch-orgs/actions/runs/16427936296

@W-19101015@

### Functionality Before
Screenshot 2025-07-21 at 4 38 39 PM (2)

### Functionality After
Screenshot 2025-07-21 at 4 56 18 PM (2)

### DIFF
--- a/.github/workflows/cleanupDeletedScratchOrgs.yml
+++ b/.github/workflows/cleanupDeletedScratchOrgs.yml
@@ -53,7 +53,7 @@ jobs:
           sf data delete bulk \
             --sobject ScratchOrgInfo \
             --where "Status = 'Deleted'" \
-            --target-org cleanup-org \
+            --target-org vscodeOrg \
             --wait 10
           echo "✅ Successfully deleted scratch org records"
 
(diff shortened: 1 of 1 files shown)


CristiCanizales (approved):
This is awesome Daphne! Thanks for working on this 🚀  Please create a WI to account for your effort 💪

PR #6434, opened 2025-07-23
W-19122147: ci: update dependabot run schedule to off hours

madhur310:
### What does this PR do?
update dependabot run schedule to off hours

@W-19122147@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,9 +13,10 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/' # Location of package manifests
     schedule:
-      # Run daily at 6pm Pacific time to align with automerge schedule
-      interval: 'daily'
-      time: '18:00'
+      # After the weekly release branch cut-off on Monday
+      interval: 'weekly'
+      day: 'wednesday'
+      time: '06:00'
       timezone: 'America/Los_Angeles'
     ignore:
       - dependency-name: '*'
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -13,10 +13,9 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/' # Location of package manifests
     schedule:
-      # After the weekly release branch cut-off on Monday
-      interval: 'weekly'
-      day: 'wednesday'
-      time: '06:00'
+      # Run daily at 6pm Pacific time to align with automerge schedule
+      interval: 'daily'
+      time: '18:00'
```
madhur310: when I tried to have it run more frequently, I get error: https://github.com/forcedotcom/salesforcedx-vscode/runs/46606455363.

mshanemc: my mental model: 
that `schedule` configs "how often should dependabot make a list of new dependencies."  You could run that 1x/week.

During the week, it keeps opening PRs (up to the limit) from that list until there are none left of the list.   As the automerge closes PRs, dependabot will open more (but only from the list of things from the last `schedule`).  If a new dependency version is released between `schedule` it won't get added to the list until the next `schedule`.

madhur310: Ahh! You are right, just read the docs https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#frequency-of-dependabot-pull-requests. I will revert this file then.

mshanemc (approved):

PR #6438, opened 2025-07-25
feat: W-18991174 - pull operations now use SDR instead of CLI 

CristiCanizales:
### What does this PR do?
This pull request introduces new utility functions for error handling, enhances directory management features, updates dependencies, and improves the reliability of end-to-end tests. Below is a categorized summary of the most significant changes:

### Error Handling Enhancements:
* Added utility functions in `errorUtils.ts` to handle errors more robustly, including `errorToString`, `isError`, `isErrorString`, `getErrorMessage`, and `getErrorStack`. These functions provide safe error conversion, type guards, and stack trace extraction.
* Exported the new error utility functions in `index.ts` for broader usage across the codebase.
* Added comprehensive unit tests for the error utility functions in `errorUtils.test.ts`.

### Directory Management Improvements:
* Introduced an `isEmptyDirectory` function in `fs.ts` to check if a directory is empty or contains only empty subdirectories. This enhances file system utilities.
* Exported the `isEmptyDirectory` function in `index.ts` for external use.

### Dependency Updates:
* Updated the `@salesforce/source-deploy-retrieve-bundle` dependency from version `^12.21.1` to `^12.21.5` in both `salesforcedx-utils-vscode` and `salesforcedx-vscode-core` packages. [[1]](diffhunk://#diff-37f4e48b0cfae5c1a2a05ff67b5534296ef4ad77e910d4d92fc0115223090b33L27-R27) [[2]](diffhunk://#diff-50be6bcc9e6b22d0fbc39fb93b55729543fa2eefdaf011ac5d57c598ca90901eL32-R32)

### End-to-End Test Reliability:
* Added `beforeEach` hooks in multiple test suites (`createOasDoc.e2e.ts`, `debugApexTests.e2e.ts`, `deployAndRetrieve.e2e.ts`, `pushAndPull.e2e.ts`, `runApexTests.e2e.ts`, `trailApexReplayDebugger.e2e.ts`) to skip tests if any previous test in the suite has failed. This prevents cascading failures. [[1]](diffhunk://#diff-350f07c2d8d85083165033a09f13001744c190ebb8032ebec221e31cfe8ea0a7R140-R145) [[2]](diffhunk://#diff-0b53e01f1248cf5a1c5b9b4f17b6d93f3c1379b82abf8835e32e72ae3db0f067R75-R80) [[3]](diffhunk://#diff-d2c62eeec8d3699854a5e66f4bcd3c6a3d047832fa911c2c8212779fb58b363cR89-R94) [[4]](diffhunk://#diff-4dded13cbf81a0f506bf1372fdafbb7554e795ce2efc333e0daf16b1746473bcR49-R54) [[5]](diffhunk://#diff-76ddf698a442379fd89d70d2a2c469fba7266dad5d10d9b44b4c91b2e2b61fe5R93-R98) [[6]](diffhunk://#diff-979b86e7dcf1b9e97bf9ec140ae64e5397c3e2a4ca972443e67eed66019de325R67-R72)
* Improved test scenarios in `deployAndRetrieve.e2e.ts` by refining notifications and adding steps to disable multiple settings. [[1]](diffhunk://#diff-d2c62eeec8d3699854a5e66f4bcd3c6a3d047832fa911c2c8212779fb58b363cL280-R292) [[2]](diffhunk://#diff-d2c62eeec8d3699854a5e66f4bcd3c6a3d047832fa911c2c8212779fb58b363cR343-R345) [[3]](diffhunk://#diff-d2c62eeec8d3699854a5e66f4bcd3c6a3d047832fa911c2c8212779fb58b363cL401-R413) [[4]](diffhunk://#diff-d2c62eeec8d3699854a5e66f4bcd3c6a3d047832fa911c2c8212779fb58b363cR535-R540)

### Minor Code Enhancements:
* Added `join` import in `fs.ts` to support new file path operations.
* Simplified and cleaned up test code in `pushAndPull.e2e.ts` and `runApexTests.e2e.ts`. [[1]](diffhunk://#diff-4dded13cbf81a0f506bf1372fdafbb7554e795ce2efc333e0daf16b1746473bcR144-R146) [[2]](diffhunk://#diff-4dded13cbf81a0f506bf1372fdafbb7554e795ce2efc333e0daf16b1746473bcL301) [[3]](diffhunk://#diff-76ddf698a442379fd89d70d2a2c469fba7266dad5d10d9b44b4c91b2e2b61fe5L302-R308)

@W-18991174@


packages/salesforcedx-vscode-core/src/commands/util/timestampConflictChecker.ts
```diff
@@ -27,11 +27,18 @@ export class TimestampConflictChecker implements PostconditionChecker<string> {
   private isManifest: boolean;
   private messages: ConflictDetectionMessages;
   private isPushOperation: boolean;
+  private isPullOperation: boolean;
```
peternhale: Having these two booleans implies an additional state is being handled in the class.
The caller has full control of third state by passing false for both in the c'tor or letting them default.

What is the third state and is it mutually exclusive of push or pull?

I there a way to express the needed states in a simpler way, 'push' | 'pull' | 'thrird' | ...

peternhale (changes requested):

packages/salesforcedx-utils-vscode/src/providers/sourceTrackingProvider.ts
```diff
@@ -37,6 +37,17 @@ export class SourceTrackingProvider {
     return sourceTracker;
   }
 
+  /**
+   * Clears the cached source tracking instance for a specific project and connection.
+   * This is useful to force a fresh source tracking instance, particularly for pull operations
+   * where stale remote tracking data might cause issues.
+   */
+  public clearSourceTracker(projectPath: string, connection: Connection): void {
```
mshanemc, line 49: I'm sure this works, but the bigger question is, 
"why are the extensions managing multiple sourceTrackers?  What's the scenario where you'd want to be running tracking for several orgs at once?"  

Why not limit it to just the current org, and reinstantiate when the user switches orgs?  Is it a scenario where you retrieve from one trackng sandbox and deploy to another?

Or was this a workaround for some bug that the library isn't handling well?

I think at this point we've got
* singleton-by-path around remote tracking in STL
* singleton-by-path around local tracking in STL
* singleton-by path wrapped around STL instances in the Provider
* basic singleton in the Service.

mshanemc: can't comment on is because github, but why is `ignoreConflicts` true below?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -87,6 +92,106 @@ export abstract class DeployRetrieveExecutor<
...
+      }
+
+      const sourceTracking = await SourceTrackingService.getSourceTracking(projectPath, connection);
+      if (!sourceTracking) {
+        throw new Error(nls.localize('error_source_tracking_service_failed'));
+      }
+
+      // Get local changes for conflict detection using the proper method
+      const localComponentSets = await sourceTracking.localChangesAsComponentSet(false);
+      const localComponentSet = localComponentSets.length > 0 ? localComponentSets[0] : new ComponentSet();
```
mshanemc:
```suggestion
      const localComponentSet = localComponentSets[0] ?? new ComponentSet();
```

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -87,6 +92,106 @@ export abstract class DeployRetrieveExecutor<
...
+
+      // Populate changedFilePaths for conflict detection from local changes
+      changedFilePaths.length = 0;
```
mshanemc: did you mean `this.changedFilePaths` ?

you should NOT change the length of an array like this, just reset the array like `this.changedFilePaths = []`

OR don't do mutation at all and say (ex)

this.changedFilePaths = Array.from(localComponentSet.getSourceComponents()).filter(c => c.content).map(c => nodePath.isAbsolute(component.content) ? c.content : nodePath.resolve(projectPath, c.content)

so that `this.filePaths` is always correct

packages/salesforcedx-vscode-core/src/commands/projectRetrieveStart.ts
```diff
@@ -5,193 +5,117 @@
...
+  ): Promise<boolean> {
+    // Get components to determine changed files
+    const components = await this.getComponents(response);
+
+    // Check for conflicts if:
+    // 1. We're not ignoring conflicts
+    // 2. Conflict detection is enabled
+    // 3. There are changed files
+    if (
+      !this.ignoreConflicts &&
```
mshanemc: checking my understanding:
vscode completely disables the conflict detection stuff from SDR/STL and always uses its own timestamp-based conflict logic

packages/salesforcedx-vscode-core/src/conflict/messages.ts
```diff
@@ -33,6 +34,22 @@ export const getConflictMessagesFor = (logName: string): ConflictDetectionMessag
...
+    ],
+    [
+      'retrieve_with_sourcepath',
+      {
+        warningMessageKey: retrieveWarningMessageKey,
+        commandHint: inputs =>
+          (Array.isArray(inputs) ? inputs : [inputs])
+            .map(input =>
+              new SfCommandBuilder()
+                .withArg('project:retrieve:start')
```
mshanemc, line 64: what's this for?

packages/salesforcedx-vscode-core/src/conflict/conflictUtils.ts
```diff
@@ -18,3 +28,105 @@ export const filesDiffer = async (one: string, two: string): Promise<boolean> =>
...
+    // Create a single cache operation for all changed files
+    // If we have changedFilePaths (source tracking enabled), use those specific files
+    // If we don't have changedFilePaths (source tracking disabled), use the entire project
+    const componentPath =
+      changedFilePaths.length > 0
+        ? changedFilePaths.length === 1
+          ? changedFilePaths[0]
+          : changedFilePaths
+        : projectPath;
+    const result = await cacheService.loadCache(componentPath, projectPath, false);
```
mshanemc: this would be a lot simpler of loadCache always took an array, and you could pass it either `changedFilePaths` or, if empty, [projectPath]

and then everything else that's passing in a string passes in `[whatever]`

--

if that's too much, I think you can do without the length===1 check since it would handle an array of one element just fine?

packages/salesforcedx-vscode-core/src/conflict/messages.ts
CristiCanizales: It's like a a conflict messaging system that gives the user warning messages and CLI command hints that can be run to resolve conflicts when conflicts are detected during deploy/retrieve

packages/salesforcedx-vscode-core/src/commands/projectRetrieveStart.ts
CristiCanizales: yes! Good catch. I'm currently trying some things to now use SDR/STL conflict detection like CLI commands do tho

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -84,8 +92,100 @@ export abstract class DeployRetrieveExecutor<
...
+    }
+  }
+
+  /**
+   * Shared method to handle empty ComponentSet scenarios
+   * This eliminates duplication between deploy and retrieve executors
+   * @param operationType The type of operation ('push' | 'deploy' | 'pull' | 'retrieve')
+   * @param isSuccess Whether the operation was successful
+   */
+  protected handleEmptyComponentSet(operationType: OperationType, isSuccess: boolean = true): void {
```
mshanemc: neither `handleEmptyComponentSet` nor `getLocalChanges` are using class members.  so they don't really belong on the abstract base class.

I think they'd be better as exported functions that deploy or retrieve or anything else that needs it (ex: `SFDX View Local Changes`) can compose with, whether they relate to Deploy/Retrieve or not.

hint: if a method doesn't use `this` it probably shouldn't be a method.

`handleEmptyComponentSet` might need it to access the errorCollection stuff?

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -84,8 +92,100 @@ export abstract class DeployRetrieveExecutor<
...
+    const output =
+      operationType === 'push' || operationType === 'deploy'
+        ? createDeployOrPushOutput([], [], isSuccess, operationType)
+        : createRetrieveOrPullOutput([], [], operationType);
+
+    channelService.appendLine(output);
+
+    // Clear any existing errors since this is a successful "no changes" scenario
+    DeployRetrieveExecutor.errorCollection.clear();
```
mshanemc: why error collection at both levels?

packages/salesforcedx-vscode-core/src/commands/projectDeployStart.ts
```diff
@@ -9,45 +9,30 @@ import {
...
       : nls.localize('project_deploy_start_default_org_text');
     super(localizedCommandName, PROJECT_DEPLOY_START_LOG_NAME);
     this.showChannelOutput = showChannelOutput;
-    this.isPushOp = true;
     this.ignoreConflicts = ignoreConflicts;
   }
 
-  public getChangedFilePaths(): string[] {
-    return this.changedFilePaths;
+  protected getOperationType(): DeployRetrieveOperationType {
```
mshanemc: what's the value in having these be a method?  Would a public, read-only property work?

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
```diff
@@ -9,17 +9,41 @@ import { ComponentSet, DeployResult } from '@salesforce/source-deploy-retrieve-b
...
+  }
+
+  private deploy = async (
+    components: ComponentSet,
+    connection: any,
+    token: vscode.CancellationToken,
+    enableIgnoreConflicts: boolean = false
+  ): Promise<DeployResult> => {
+    if (enableIgnoreConflicts) {
+      // Retry with ignoreConflicts: true
```
mshanemc: I don't understand the `retry` comment

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
```diff
@@ -101,7 +143,6 @@ export abstract class DeployExecutor<T> extends DeployRetrieveExecutor<T, Deploy
     const isSuccess =
       result.response.status === RequestStatus.Succeeded || result.response.status === RequestStatus.SucceededPartial;
 
-    const operationType = this.isPushOperation() ? 'push' : 'deploy';
-    return createDeployOrPushOutput(result.getFileResponses(), relativePackageDirs, isSuccess, operationType);
+    return createDeployOrPushOutput(result.getFileResponses(), relativePackageDirs, isSuccess, this.getOperationType());
```
mshanemc: is `createDeployOrPushOutput` useful?  Isn't it just a redirect to `createOperationOutput`

packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -94,6 +94,14 @@ export abstract class LibraryCommandletExecutor<T> implements CommandletExecutor
     this.outputChannel = outputChannel;
   }
 
+  protected getLogName(): string {
```
mshanemc: rather than wrap these private props in accessor methods, could you make the props protected (but still readonly) so that subclasses can access them?

packages/salesforcedx-vscode-core/src/conflict/conflictUtils.ts
```diff
@@ -18,3 +33,237 @@ export const filesDiffer = async (one: string, two: string): Promise<boolean> =>
...
+ * Handles conflicts by showing VS Code's conflict UI and allowing user interaction
+ * @param conflicts Array of conflicts from SourceTracking.getConflicts() or SDR error
+ * @param conflictMessageType The type of conflict messages to use ('deploy_with_sourcepath', 'retrieve_with_sourcepath', etc.)
+ * @param operationType The type of operation ('deploy', 'retrieve', 'push', 'pull')
+ * @param operation The original operation function
+ * @param retryOperation Optional function to retry the operation with ignoreConflicts: true
+ * @returns Promise<T> The result of operation, or undefined if user cancels
+ */
+export const handleConflictsWithUI = async <T>(
+  conflicts: any[],
```
mshanemc: let's not use `any`. `conflicts` is typed in STL (it's `ChangeResult[]`).

If that's not exported, you can use `ReturnType<>` to get it from the public `getConflicts` method.  LMK if you need help with that.

packages/salesforcedx-vscode-core/src/conflict/messages.ts
```diff
@@ -33,6 +35,36 @@ export const getConflictMessagesFor = (logName: string): ConflictDetectionMessag
...
+      }
+    ],
+    [
+      'project_deploy_start_default_scratch_org',
+      {
+        warningMessageKey,
+        commandHint: () => nls.localize('push_conflicts_error')
+      }
     ]
   ]);
```
mshanemc: this is a really strange fn.

Why not return the appropriate conflictMessage based on the log name?

instead of doing all the localization and commandBuilding, then building a map of all the possibilities, and then choosing to return only one of them?

And when could it ever return `undefined`?  It's going to throw otherwise?  But you could probably eliminate that at compile time by restricting `logName` to be one of a union type (`project_retrieve_start_default_scratch_org | ...`) so that there's no possibility of someone using it with an invalid key, so there's not this weird runtime error `No conflict messages found` that the user wouldn't be able to understand.

packages/salesforcedx-vscode-core/src/conflict/messages.ts
```diff
@@ -33,6 +35,36 @@ export const getConflictMessagesFor = (logName: string): ConflictDetectionMessag
...
+      {
+        warningMessageKey: retrieveWarningMessageKey,
+        commandHint: () => nls.localize('pull_conflicts_error')
+      }
+    ],
+    [
+      'project_deploy_start_default_scratch_org',
+      {
+        warningMessageKey,
+        commandHint: () => nls.localize('push_conflicts_error')
```
mshanemc, line 81: I'm trying to follow all the layers of abstraction here.  What I *think* i see is that only `retrieve` is going to do anything with the "commandHint" but we're creating them on all the other message types?

packages/salesforcedx-vscode-core/src/commands/projectRetrieveStart.ts
```diff
@@ -5,193 +5,87 @@
...
+  SourceTrackingService,
+  workspaceUtils
 } from '@salesforce/salesforcedx-utils-vscode';
+import { ComponentSet } from '@salesforce/source-deploy-retrieve-bundle';
 import * as vscode from 'vscode';
-import { channelService } from '../channels';
-import { PersistentStorageService } from '../conflict';
 import { PROJECT_RETRIEVE_START_LOG_NAME } from '../constants';
-import { coerceMessageKey, nls } from '../messages';
-import { CommandParams, FlagParameter, SfCommandlet, SfCommandletExecutor } from './util';
```
mshanemc, line 25: removing all these made an orphan of CommandParam so we can get rid of it

mshanemc (commented):
mostly stuff in the category of code cleanup.

Also recommended: run `npx knip` on this.  It'll find some more dead code, unused exports, etc.  Anytime you do a big rearrangement like this, you'll probably have a bit of dead code cleanup.  It produces a few false positives and sometimes misses things that are exported for use only by tests, so you'll have to look at them like "cleanup suggestions"

packages/salesforcedx-vscode-core/src/commands/util/timestampConflictChecker.ts
```diff
@@ -23,15 +23,17 @@ import { DeployQueue, salesforceCoreSettings } from '../../settings';
 import { telemetryService } from '../../telemetry';
 import { ConflictDetectionMessages } from './conflictDetectionMessages';
 
+export type ConflictOperationType = 'push' | 'pull' | 'deploy' | 'retrieve';
```
mshanemc: it's also the same as `DeployRetrieveOperationType`.  we should only have 1

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
```diff
@@ -167,7 +267,7 @@ export const createOutputTable = (
 /**
  * Operation type for output configuration
  */
-export type OperationType = 'deploy' | 'retrieve' | 'push';
+export type OperationType = 'deploy' | 'retrieve' | 'push' | 'pull';
```
mshanemc: that's a duplicate of DeployRetrieveOperationType from packages/salesforcedx-vscode-core/src/util/types.ts

packages/salesforcedx-vscode-core/src/commands/baseDeployRetrieve.ts
CristiCanizales: One is for push/pull and the other one for deploy/retrieve, and we want them all cleared when a new operation starts

packages/salesforcedx-vscode-core/src/commands/projectDeployStart.ts
CristiCanizales: it would! Thanks!

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
CristiCanizales: It's just that we're reaching this in the retry operation we're providing for handleConflictsWithUI, where we're setting ignoreConflicts to true to just continue with the deploy. I'll remove if it's confusing!

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
CristiCanizales: nope it isn't 🤡 good catch

packages/salesforcedx-utils-vscode/src/providers/sourceTrackingProvider.ts
CristiCanizales: > was this a workaround for some bug that the library isn't handling well?

Yes. I tried removing this and pull stops working (Always getting no results found even when I've made changes in the org)

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
mshanemc: I get it now.  `enableIgnoreConflicts` is only used by the retry method of handleConflictsWithUi.

packages/salesforcedx-vscode-core/src/commands/deployExecutor.ts
mshanemc: did you forget to push some commits?  I see a lot of stuff marked `resolved` like this one but no fixes for them.

I had started this https://github.com/forcedotcom/salesforcedx-vscode/pull/6477/files to help move things along but maybe you already did all that but not pushed it?

mshanemc:
QA notes:

run pull command
✅ no results

create a new permset, run pull command
✅ retrieved

same permset, edit the description in org, and add a different description locally
❌ should have been a Conflict but wasn't caught.  This is reproducible.  And STL knows it's a conflict, so we're missing the conflict check 
Screenshot 2025-08-17 at 6 00 33 PM

mshanemc (approved):

PR #6451, opened 2025-07-31
[W-19111839] fs.writeFile() for E2E tests ; [W-19111477] global beforeEach() for E2E tests ; [W-19109744] fix bug in TraceFlag creation

daphne-sfdc:
### [W-19111839] refactor: use `fs.writeFile()` instead of `textEditor.setText()` to populate file contents
Corresponds to https://github.com/forcedotcom/salesforcedx-vscode-automation-tests-redhat/pull/82

This PR updates our E2E tests to enhance reliability by switching from `textEditor.setText()` and `textEditor.setTextAtLine()` to `fs.writeFile()` for populating file contents.

Previously, our E2E tests would open files in an Editor View and use UI-based methods to replace their contents. This approach proved unreliable and led to flappers. These failures were due to the inherent instability of UI-driven file manipulation within the testing environment.

By adopting `fs.writeFile()`, we now directly interact with the filesystem to create or overwrite file contents. This change significantly improves the stability of our E2E tests, as it bypasses the less predictable UI operations. Our primary goal for these E2E tests is not to test file creation – that's already covered in our **templates.e2e.ts** – but rather to validate broader system functionality.

This refactor ensures our E2E tests are more robust and less prone to environmental inconsistencies, allowing us to focus on testing the core features they were designed for.

---

### [W-19111477] refactor: move `beforeEach()` from each *.e2e.ts file to a Mocha configuration file

This PR moves a common `beforeEach()` block from individual test files to a centralized Mocha configuration.

Previously, every E2E test file contained the following `beforeEach()` hook:
```
beforeEach(function () {
  if (this.currentTest?.parent?.tests.some(test => test.state === 'failed')) {
    this.skip();
  }
});
```

This block's purpose is to ensure that E2E test steps run sequentially and that subsequent tests are skipped if a preceding test in the same suite fails. This "fail-fast" behavior is crucial for efficient debugging of E2E test failures.

With this change, this duplicate code has been extracted and moved to **globalHooks.ts**. This global hook is now integrated into our Mocha test runner via **.mocharc.json**.

All existing and newly added E2E tests will automatically inherit the desired sequential execution and fail-fast logic without needing to explicitly include the `beforeEach()` block in each file.

---

#### Problem
The **TrailApexReplayDebugger** E2E test was failing intermittently on Ubuntu with empty `quickPicks` arrays when trying to retrieve debug logs. The tests would pass the "SFDX: Turn On Apex Debug Log for Replay Debugger" and "SFDX: Run Apex Tests" steps successfully, but no debug logs were actually being generated, causing the subsequent "SFDX: Get Apex Debug Logs" command to return no results.

#### Root Cause
The issue was in the **TraceFlags** class where the `StartDate` field was being set to `Date.now()` (a numeric timestamp) while the `ExpirationDate` was correctly formatted using `toUTCString()`. This inconsistency caused platform-specific issues:
```// ❌ Problematic code
const traceFlag = {
  tracedentityid: userId,
  logtype: 'developer_log', 
  debuglevelid: debugLevelId,
  StartDate: Date.now(),                    // Raw number - unreliable
  ExpirationDate: expirationDate.toUTCString()  // Proper string format
};
```
The Salesforce Tooling API expects date fields in specific string formats (e.g., "yyyy-MM-ddTHH:mm:ss.SSSZ"). While the API has some tolerance for numeric timestamps, this tolerance varies based on platform-specific JSON serialization behavior, making it unreliable on Ubuntu/Linux systems.

#### Solution
Updated both `createTraceFlag()` and `updateTraceFlag()` methods to use consistent string formatting for the `StartDate` field:
```
// ✅ Fixed code
const traceFlag = {
  tracedentityid: userId,
  logtype: 'developer_log',
  debuglevelid: debugLevelId, 
  StartDate: new Date().toUTCString(),      // Consistent string format
  ExpirationDate: expirationDate.toUTCString()  // Already correct
};
```

#### Passing E2E Test Runs

After implementing the fix, I made 5 runs of the **TrailApexReplayDebugger** E2E test, which all passed. 🚀
1. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16655162118
2. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16655165375
3. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16655168482
4. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16655171807
5. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/16655176222


packages/salesforcedx-vscode-automation-tests/test/specs/lwcLsp.e2e.ts
```diff
@@ -73,28 +75,31 @@ describe('LWC LSP', () => {
...
-
-      // Go to definition through F12
-      await executeQuickPick('Go to Definition', Duration.seconds(2));
-
-      // Verify 'Go to definition' took us to the definition file
-      const editorView = workbench.getEditorView();
-      const activeTab = await editorView.getActiveTab();
-      const title = await activeTab?.getTitle();
-      expect(title).to.equal('lwc1.js');
+      await executeQuickPick('View: Close All Editors', Duration.seconds(1));
```
daphne-sfdc, line 79: I added this line so that the `lwc1.html` tab remains visible in Ubuntu, where the VSCode window is smaller and fewer tabs fit.

packages/salesforcedx-vscode-automation-tests/test/specs/deployAndRetrieve.e2e.ts
```diff
@@ -71,28 +70,7 @@ describe('Deploy and Retrieve', () => {
...
-      Duration.TEN_MINUTES
-    );
-    expect(successNotificationWasFound).to.equal(true);
-
-    const outputPanelText = await attemptToFindOutputPanelText(
-      'Salesforce CLI',
-      'Finished SFDX: Create Apex Class',
-      10
-    );
-    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
daphne-sfdc, line 86: No Output Tab message and success notification produced when the Apex class is created using `fs.writeFile()`.

packages/salesforcedx-vscode-automation-tests/test/specs/trailApexReplayDebugger.e2e.ts
```diff
@@ -173,48 +175,77 @@ describe('"Find and Fix Bugs with Apex Replay Debugger" Trailhead Module', () =>
 
   it('SFDX: Get Apex Debug Logs', async () => {
     logTestStart(testSetup, 'SFDX: Get Apex Debug Logs');
+    await executeQuickPick('View: Close All Editors', Duration.seconds(1));
```
daphne-sfdc, line 181: I added this line to ensure the log file's tab remains visible in Ubuntu.  Same issue as **lwcLsp.e2e.ts** where the VSCode window size in Ubuntu is smaller than Mac and Windows.

packages/salesforcedx-vscode-automation-tests/test/specs/debugApexTests.e2e.ts
```diff
@@ -56,14 +57,14 @@ describe('Debug Apex Tests', () => {
 
     // Create Apex class 1 and test
     await retryOperation(
-      () => createApexClassWithTest('ExampleApexClass1'),
+      () => createApexClassWithTest('ExampleApexClass1', path.join(testSetup.projectFolderPath!, 'force-app', 'main', 'default', 'classes')),
```
CristiCanizales: can you please move this to a constant before these calls ans use it instead?

packages/salesforcedx-vscode-automation-tests/test/specs/debugLwcTests.e2e.ts
```diff
@@ -60,10 +60,10 @@ describe('Debug LWC Tests', () => {
     await closeAllEditors();
 
     // Create LWC1 and test
-    await createLwc('lwc1');
+    await createLwc('lwc1', path.join(testSetup.projectFolderPath!, 'force-app', 'main', 'default', 'lwc'));
```
CristiCanizales: same here and everywhere we're doing this

packages/salesforcedx-vscode-automation-tests/test/specs/sfdxProjectJson.e2e.ts
```diff
@@ -74,9 +72,8 @@ const createSfdxProjectJsonWithAllFields = async (testSetup: TestSetup): Promise
     '\t"sourceBehaviorOptions": ["decomposeCustomLabelsBeta", "decomposePermissionSetBeta", "decomposeWorkflowBeta", "decomposeSharingRulesBeta"]',
     '}'
   ].join('\n');
-  await openFile(path.join(testSetup.projectFolderPath!, 'sfdx-project.json'));
-  const textEditor = await getTextEditor(workbench, 'sfdx-project.json');
-  await textEditor.setText(sfdxConfig);
-  await textEditor.save();
+
+  const sfdxProjectPath = path.join(testSetup.projectFolderPath!, 'sfdx-project.json');
+  await fs.writeFile(sfdxProjectPath, sfdxConfig, 'utf8');
```
CristiCanizales: I think we should have an utility in redhat repo to do this and use it instead. Wdyt?

packages/salesforcedx-vscode-automation-tests/test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -190,19 +193,19 @@ describe('Apex Replay Debugger', () => {
     expect(executionFinished).to.be.greaterThanOrEqual(1);
   });
 
-  it('SFDX: Launch Apex Replay Debugger with Last Log File', async () => {
-    logTestStart(testSetup, 'ApexReplayDebugger - SFDX: Launch Apex Replay Debugger with Last Log File');
+  it('SFDX: Launch Apex Replay Debugger with Current File - log file', async () => {
+    logTestStart(testSetup, 'ApexReplayDebugger - SFDX: Launch Apex Replay Debugger with Current File - log file');
```
daphne-sfdc, line 206: I found that about 1/3 of attempts for **apexReplayDebugger.e2e.ts** were failing in recent runs.  Digging into the issue, I found that we were running **SFDX: Launch Apex Replay Debugger with Last Log File** immediately after getting Apex debug logs.  Because we hadn't run Apex Replay Debugger with that log file yet, there was no "last log file", and thus we get the error in the below screenshot and a test failure.
Apex Replay Debugger SFDX Launch Apex Replay Debugger with Current File - log file

The solution is to run the **SFDX: Launch Apex Replay Debugger with Current File - log file** step first.  That way, when it's time to run **SFDX: Launch Apex Replay Debugger with Last Log File**, the log file that we use as the "last log file" has already been used to run Apex Replay Debugger.

With this change, the Apex Replay Debugger E2E test is now able to run on Windows and Ubuntu too! 🚀

.github/workflows/apexE2E.yml
CristiCanizales, line 1: 🚀 🥳

packages/salesforcedx-vscode-automation-tests/test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -56,12 +57,13 @@ describe('Apex Replay Debugger', () => {
     log('ApexReplayDebugger - Set up the testing environment');
     testSetup = await TestSetup.setUp(testReqConfig);
     projectFolderPath = testSetup.projectFolderPath!;
+    classesFolderPath = path.join(testSetup.projectFolderPath!, 'force-app', 'main', 'default', 'classes');
```
CristiCanizales: this should be a const

packages/salesforcedx-vscode-automation-tests/test/specs/auraLsp.e2e.ts
```diff
@@ -37,24 +39,18 @@ describe('Aura LSP', () => {
   before('Set up the testing environment', async () => {
     log('AuraLsp - Set up the testing environment');
     testSetup = await TestSetup.setUp(testReqConfig);
+    auraFolderPath = path.join(testSetup.projectFolderPath!, 'force-app', 'main', 'default', 'aura');
```
CristiCanizales: same. const

CristiCanizales (changes requested):
Thanks for addressing my comments. Last one, I think all those paths could be in a separate constants.ts file in salesforcedx-vscode-automation-tests package

PR #6465, opened 2025-08-07
[W-19276014] fix: trace flags created in Developer Console should not appear in VSCode

daphne-sfdc:
### What does this PR do?
Every time a user opens Developer Console, they generate a trace flag in their org.  That trace flag has DeveloperName **SFDC_DevConsole** and should not be used by VSCode.  This PR corrects a regression in the v64.5.1 release where the trace flags created by Developer Console were being propagated into VSCode.

@W-19276014@

### Functionality Before
If you create a trace flag by opening Developer Console, you get a "Recording detailed logs until ___" message in the status bar of VSCode.
Screenshot 2025-08-07 at 12 23 18 PM

You can delete that trace flag by running **SFDX: Turn Off Apex Debug Logs for Replay Debugger** in the command palette.
Screenshot 2025-08-07 at 12 32 42 PM

### Functionality After
Trace flags created by opening Developer Console display nothing in VSCode.
Screenshot 2025-08-07 at 12 53 45 PM

Running **SFDX: Turn Off Apex Debug Logs for Replay Debugger** produces a "No active trace flag found" notification because there was no corresponding trace flag created for Apex Replay Debugger.
Screenshot 2025-08-07 at 12 58 07 PM


packages/salesforcedx-utils-vscode/src/helpers/traceFlags.ts
```diff
@@ -153,9 +154,9 @@ export class TraceFlags {
 
   public async getTraceFlagForUser(userId: string): Promise<TraceFlagRecord | undefined> {
     const traceFlagQuery = `
-      SELECT id, logtype, startdate, expirationdate, debuglevelid, debuglevel.apexcode, debuglevel.visualforce
+      SELECT id, logtype, startdate, expirationdate, debuglevelid, debuglevel.apexcode, debuglevel.visualforce, debuglevel.developername
```
mshanemc: fyi, soql is not case sensitive so it works, but you should use the real names of these fields instead of this weird lowercased version of them.  See how `TracedEntityId` id properly cased in the where clause?

then you don't have to mess with spell checker stuff.

https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_traceflag.htm

packages/salesforcedx-vscode-core/src/commands/stopApexDebugLogging.ts
```diff
@@ -27,7 +27,7 @@ export const turnOffLogging = async (extensionContext: vscode.ExtensionContext):
   // Check if a TraceFlag already exists for the current user
   const myTraceFlag = await traceFlags.getTraceFlagForUser(userId);
 
-  if (myTraceFlag?.Id) {
+  if (myTraceFlag?.Id && myTraceFlag.DebugLevel.DeveloperName === 'ReplayDebuggerLevels') {
```
mshanemc: the `Id` check is unnecessary

PR #6467, opened 2025-08-08
https://gus.lightning.force.com/a07EE00002Jl9iVYAR

mshanemc:
#<Insert GitHub Issue>, @W-19282034@

### DIFF
--- a/packages/salesforcedx-utils/src/cli/orgDisplay.ts
+++ b/packages/salesforcedx-utils/src/cli/orgDisplay.ts
@@ -87,7 +87,7 @@ export class OrgDisplay {
 
     // Get organization details via SOQL
     const orgQuery = await connection.singleRecordQuery<OrgQueryResult>(
-      'SELECT Id, Name, CreatedDate, CreatedBy.Username, OrganizationType, InstanceName, IsSandbox FROM Organization'
+      'SELECT Id, Name, CreatedDate, CreatedBy.Username, OrganizationType, InstanceName, NamespacePrefix, IsSandbox FROM Organization'
     );
 
     // For scratch orgs, get detailed information from the dev hub
@@ -143,7 +143,7 @@ export class OrgDisplay {
       }
     }
 
-    const orgInfo: OrgInfo = {
+    return {
       username,
       devHubId: authFields.devHubUsername ?? '',
       id: authFields.orgId ?? orgQuery.Id,
@@ -152,6 +152,7 @@ export class OrgDisplay {
       expirationDate: scratchOrgInfo.expirationDate,
       status,
       edition,
+      ...(orgQuery.NamespacePrefix ? { namespace: orgQuery.NamespacePrefix } : {}),
       orgName: scratchOrgInfo.orgName,
       accessToken: authFields.accessToken ?? '',
       instanceUrl: authFields.instanceUrl ?? '',
@@ -161,7 +162,5 @@ export class OrgDisplay {
       connectionStatus,
       password: scratchOrgInfo.password ?? ''
     };
-
-    return orgInfo;
   }
 }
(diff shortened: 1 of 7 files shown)


packages/salesforcedx-vscode-apex/src/namespaceLensRewriter.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+import * as vscode from 'vscode';
+
+const singleTest = ['Run Test', 'Debug Test'];
+const allTests = ['Run All Tests', 'Debug All Tests'];
+
+// Jorje sticks the namespace from sfdx-project.json (namespaceFromProject) on the arguments, like ns.class.method
+// org may or may not actually use the namespace (ex: scratch org with --no-namespace)
+// namespaceFromOrg represents the namespace that came from auth files (ie, when the org is created/auth'd)
+export const rewriteNamespaceLens =
```
peternhale, line 15: @mshanemc is a curried function needed in this use case? Typically the first is making some sort of change that is passed along to the next. Am I missing something?

packages/salesforcedx-vscode-apex/src/namespaceLensRewriter.ts
```diff
@@ -0,0 +1,41 @@
...
+  (namespaceFromOrg?: string) => (namespaceFromProject?: string) => (lens: vscode.CodeLens) => {
+    if (namespaceFromOrg !== undefined || !lens.command?.title || namespaceFromProject === undefined) {
+      // if the org is a namespaces, we preserve the namespace from the LS.
+      // if the project has no namespace, we want to use what the LS provides (its use of the namespace is the cause of https://github.com/forcedotcom/salesforcedx-vscode/issues/6458 )
+      return lens;
+    }
+
+    if (singleTest.includes(lens.command.title)) {
+      // namespace.class.method => class.method
+      console.log(`provideCodeLenses Middleware > Single test originally: ${lens.command.arguments}`);
```
peternhale: Are the console.log call still needed? Maybe change them console.debug?

packages/salesforcedx-vscode-apex/src/namespaceLensRewriter.ts
mshanemc: a little nicer to be able to use a point-free map.
quite a bit nicer to load the 2 partially applied params in the UT instead of supplying x,y,z every call.

packages/salesforcedx-vscode-apex/src/namespaceLensRewriter.ts
mshanemc: I haven't figured out the pattern of logging in the extensions.  I'll go with `debug`.

packages/salesforcedx-vscode-apex/src/namespaceLensRewriter.ts
mshanemc: I don't think a curried fn is *ever* needed, just easier to work with.

peternhale: Ok, this may be a new topic for other on the team. I am sure they would like to learn about curried functions; when to use, advantages, disadvantages, etc.

CristiCanizales (approved):
org display changes look good to me! 
I do would like to learn about curried functions and all that Pete mentioned 😄

daphne-sfdc (changes requested):
These 2 commands are still producing the original error:
1. Launch Apex replay debugger with current (Apex test class) file
2. Single class from sidebar

Screenshot 2025-09-09 at 10 55 51 PM

Please fix the error so that the Apex Replay Debugger is able to run in these two scenarios.

PR #6482, opened 2025-08-14
[W-19176017] refactor: telemetry display without CLI

daphne-sfdc:
### What does this PR do?
Getting the telemetry ID from users to send telemetry to AppInsights no longer requires running the `sf telemetry` CLI command.  Instead, the telemetry ID is calculated as a hashed version of (orgID + userID).  This is in preparation for running the extensions on the web, where users won't have access to the CLI.

Test plan here: https://salesforce.quip.com/rKQ3AcfmDPbb

@W-19176017@

### Functionality Before
Telemetry ID is the user's CLI ID, taken by running the `sf telemetry` CLI command.

### Functionality After
Telemetry ID is a hashed version of (orgID + userID), which works in a web environment.


packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -187,6 +187,81 @@ export class TelemetryService implements TelemetryServiceInterface {
...
+    this.extensionContext?.subscriptions.push(...this.reporters);
+
+    console.log(`Telemetry reporters refreshed for ${name} with new user ID:`, userId);
+  }
+
+  /**
+   * Refreshes telemetry reporters for ALL extension instances when org authorization changes.
+   * This ensures that all extensions (Core, Apex, etc.) use the updated hashed user ID.
+   */
+  public static async refreshAllExtensionReporters(coreExtensionContext: ExtensionContext): Promise<void> {
```
madhur310: This seems like a stand alone function, should we move it to a utils/helper?

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -187,6 +187,81 @@ export class TelemetryService implements TelemetryServiceInterface {
     return this.reporters;
   }
 
+  /**
+   * Refreshes telemetry reporters with the latest user ID when org authorization changes.
+   * This ensures that telemetry events use the correct user ID (hashed orgId + userId)
+   * instead of the initial random ID that was set during extension activation.
+   */
+  public async refreshReporters(extensionContext: ExtensionContext): Promise<void> {
```
madhur310, line 207: we need unit tests for the new methods

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -187,6 +187,81 @@ export class TelemetryService implements TelemetryServiceInterface {
...
+
+    // Get the updated user ID
+    const userId = await UserService.getTelemetryUserId(extensionContext);
+
+    // Dispose existing reporters
+    for (const reporter of this.reporters) {
+      try {
+        await reporter.dispose();
+      } catch (error) {
+        console.log('Error disposing telemetry reporter:', error);
```
madhur310: who is supposed to see this error?

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
```diff
@@ -187,6 +187,81 @@ export class TelemetryService implements TelemetryServiceInterface {
...
+      }
+    }
+    this.reporters = [];
```
madhur310: nit: I think `this.reporters.length = 0` is more efficient.

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -5,50 +5,95 @@
...
+      console.log(`Failed to get shared telemetry user ID: ${String(error)}`);
+    }
+    return undefined;
   }
 
+  /**
+   * Creates a one-way hash of orgId and userId for telemetry compliance.
+   * This ensures customer data cannot be decoded while maintaining user distinction.
+   */
+  private static hashUserIdentifier(orgId: string, userId: string): string {
```
madhur310: Might be better as utils/helper method?

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -5,50 +5,95 @@
...
+    return createHash('sha256').update(`${orgId}-${userId}`).digest('hex');
+  }
+
+  /**
+   * Retrieves or generates a telemetry user ID for the current VS Code extension context.
+   * The returned user ID is used for telemetry purposes and is determined as follows:
+   *
+   * 1. First, attempts to get a shared telemetry user ID from the Core extension if available.
```
madhur310: Nice comments!
may be adding some indentation could help readability a little more.
```
   * 1. First, attempts to get a shared telemetry user ID from the Core extension if available.
   * 2. If shared ID is not available, falls back to extension-specific behavior:
   *  a. If org authorization data (orgId and userId) is available:
   *    - If no user ID exists in global state, or if the existing user ID is the anonymous user ID, a deterministic SHA-256 hash of orgId and userId is generated, stored, and returned.
   *    - If a non-anonymous user ID already exists in global state, it is returned as-is.
   *  b. If org authorization data is not available:
   *    - If a user ID exists in global state, it is returned.
   *  c. Otherwise, the anonymous user ID is returned.
```

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -5,50 +5,95 @@
...
-    const workspacePath = workspaceUtils.getRootWorkspacePath();
-    const execution = new CliCommandExecutor(command, { cwd: workspacePath }).execute();
-    const cmdOutput = new CommandOutput();
-    const result = cmdOutput.getCmdResult(execution);
-    return result;
+  /**
+   * Attempts to get the shared telemetry user ID from the Core extension.
+   * Returns undefined if the Core extension is not available or doesn't have the method.
+   */
+  private static async getSharedTelemetryUserId(): Promise<string | undefined> {
```
madhur310: why not utils/helper method?

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -5,50 +5,95 @@
...
+   * @param extensionContext - The VS Code extension context, used to access global state.
+   * @returns The telemetry user ID, either shared, hashed, or the anonymous user ID.
+   */
   public static async getTelemetryUserId(extensionContext: ExtensionContext): Promise<string> {
-    // Defining UserId in globalState and using the same in appInsights reporter.
-    // Assigns cliId to UserId when it's undefined in global state.
-    // cliId is undefined when cli-telemetry variable disable-telemetry is true.
-    let globalStateUserId = extensionContext?.globalState.get<string | undefined>(TELEMETRY_GLOBAL_USER_ID);
+    // First, try to get the shared telemetry user ID from the Core extension
+    // Only check for shared user ID if this is not the Core extension itself (to avoid infinite loop)
```
madhur310: if the method `getSharedTelemetryUserId` becomes a helper/utils method out of this class -> then we don't need to worry about loop? or may be I am misunderstanding the comment...

packages/salesforcedx-utils-vscode/src/services/userService.ts
```diff
@@ -5,50 +5,95 @@
...
+    const globalStateUserId = extensionContext?.globalState.get<string | undefined>(TELEMETRY_GLOBAL_USER_ID);
+
+    const context = WorkspaceContextUtil.getInstance();
+    const orgId = context.orgId;
+    const userId = context.username;
+
+    // If we have org authorization data available (orgId + userId)
+    if (orgId && userId) {
+      // If globalStateUserId is undefined or is the anonymous user ID, replace it with the hashed value
+      if (!globalStateUserId || globalStateUserId === 'UNAUTHENTICATED_USER') {
```
madhur310: `UNAUTHENTICATED_USER` should be a constant, just to reduce any human error

packages/salesforcedx-utils-vscode/src/services/userService.ts
daphne-sfdc: I actually had it indented originally, but then I get this ESLint error that prevents me from pushing:
Screenshot 2025-08-19 at 4 26 37 PM

So I removed the indentation to get rid of the error.

packages/salesforcedx-utils-vscode/src/services/telemetry.ts
daphne-sfdc: You're right, the customer doesn't need to see the error.  I removed the console.log() statement.

packages/salesforcedx-utils-vscode/src/services/userService.ts
daphne-sfdc: No, it's still the same function with the same relationship to `getTelemetryUserId()`, just moved to a different place.  The if statement must remain to avoid the infinite loop.

Suppose we are currently trying to activate the Apex extension.  The Core extension has already been activated.

The `activate()` function calls `getTelemetryUserId(extensionContext: ExtensionContext)` with the Apex extension.  Since the current extension is not Core, we go into the if statement and call `await getSharedTelemetryUserId();`.

As the Core extension is available, we will reach the line `return await coreExtension.exports.getSharedTelemetryUserId();`.  According to the Core extension's **index.ts** Line 479, `getSharedTelemetryUserId()` refers to `await UserService.getTelemetryUserId(extensionContext)`.

This means we are now calling `getTelemetryUserId(extensionContext: ExtensionContext)` with the Core extension.  Because the if statement exists, we can reach the rest of the function and return the telemetry ID that we had previously calculated for the Core extension.

If the if statement were removed, calling `getTelemetryUserId(extensionContext: ExtensionContext)` with the Core extension would call back to `await getSharedTelemetryUserId();`, which would then call back to `await UserService.getTelemetryUserId(extensionContext)` with the Core extension, ... creating an infinite loop.

madhur310: Why not use depependency injection? TelemetryService should allow instantiators to specify optional `generateTelemetryUserId`:

```
export class TelemetryService implements TelemetryServiceInterface {
  private generateTelemetryUserId: () => Promise<string> | undefined; 

...

public async initializeService(extensionContext: ExtensionContext, generateTelemetry?: ...): Promise<void> {
...
    if (this.reporters.length === 0 && (await this.isTelemetryEnabled())) {
      const userId = this.extensionContext ? await UserService.getTelemetryUserId(this.extensionContext, generateTelemetryUserId) : 'unknown';
...
    this.extensionContext?.subscriptions.push(...this.reporters);
  }
```

Core will pass in the helper method to generateTelemetryUserId, so core telemetry continues to generate the id. 

All places other than core will just pass undefined and use the shared.

Why?:
- avoids the check and no circular dependency issue.
- no hard coded string, lesser of an issue since don't think core's name is changing anytime soon.
- cleaner IMO.





Kinda tangential -- User service seems redundant IMO, both of the method can be helper methods, since both are static without any class level context.

PR #6500, opened 2025-08-25
feat: W-19158987 - remove deleted and expired orgs without cli

CristiCanizales:
### What does this PR do?
This pull request introduces significant improvements to the org cleanup and display functionality in the Salesforce VSCode extension. The main changes involve refactoring the org list clean command to more accurately identify and remove expired or deleted orgs, improving the precision of expiration checks by using timestamps, and enhancing the user feedback with detailed messages and a table display of remaining orgs. Additionally, related utility functions and messages have been updated for clarity and consistency.

**Org cleanup and display improvements:**

* Refactored the `orgList` command to use a new `OrgListCleanExecutor` class, which removes expired and deleted orgs with more precise timestamp checks and displays the remaining orgs in a formatted table, including status and expiration details.
* Added detailed i18n messages for org cleanup actions, including success, errors, and legend explanations for default orgs and DevHubs.

**Expiration check precision and consistency:**

* Updated all expiration checks in `orgPicker/orgList.ts` and related utilities to compare exact timestamps rather than just dates, ensuring orgs are only removed or marked expired at the correct time of day. [[1]](diffhunk://#diff-2b72c15ab3c2e3d66f269772e80aed9902d2d85d9acb2a462116d03541484d4fL60-R72) [[2]](diffhunk://#diff-2b72c15ab3c2e3d66f269772e80aed9902d2d85d9acb2a462116d03541484d4fL91-R101) [[3]](diffhunk://#diff-af4d2d3517e97641bc998017a352d3e3f949438d2a5a411e584e69252182c34bL41-R41) [[4]](diffhunk://#diff-af4d2d3517e97641bc998017a352d3e3f949438d2a5a411e584e69252182c34bL66-R66)

**User experience enhancements:**

* Improved output formatting for command completion messages by adding a newline for better readability in the command output channel.
* Enhanced the org filtering logic to optionally show expired orgs and provide more informative list items, including alias handling and expiration indicators. [[1]](diffhunk://#diff-2b72c15ab3c2e3d66f269772e80aed9902d2d85d9acb2a462116d03541484d4fL60-R72) [[2]](diffhunk://#diff-2b72c15ab3c2e3d66f269772e80aed9902d2d85d9acb2a462116d03541484d4fL91-R101)

These changes together provide a more robust, user-friendly, and accurate org management experience in the extension.

@W-19158987@


packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+    }
+
+    await org.refreshAuth();
+    return 'Connected';
+  } catch (err) {
+    return authErrorHandler(err, org?.getUsername() ?? username);
+  }
+};
+
+const authErrorHandler = (err: unknown, _username: string): string => {
```
madhur310: nit: the method name is misleading a bit, the method is not handling the error but rather getting error message as string. We should rename it for readability.

madhur310: getConnectionStatusFromError

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+  if (err instanceof Error) {
+    // Orgs under maintenance return html as the error message
+    if (err.message.includes('maintenance')) return 'Down (Maintenance)';
+
+    // Handle other potential html responses
+    if (err.message.includes('<html>') || err.message.includes('<!DOCTYPE HTML>')) return 'Bad Response';
+
+    // Handle specific error codes/messages
+    if (
+      err.message.includes('expired access/refresh token') ||
```
madhur310: Is it possible to avoid error message based checks, seems little fragile -- what if core decided to rename the message text? Do we have codes that we could use?

madhur310: Found this https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/src/orgBrowser/metadataOutlineProvider.ts#L132.

Also if error message is the only way we have, we should extract and share the logic in all places that check based on error message.

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+      return 'Unable to refresh session: expired access/refresh token';
+    }
+    return err.message;
+  }
+
+  return String(err);
+};
+
+/** Format time difference into human-readable string */
+export const formatTimeDifference = (timeDiffInMs: number): string => {
```
madhur310: This method seems quite verbose, can we simplify it? How about something like:

export const formatTimeDifference = (timeDiffInMs: number): string => {
  const days = Math.floor(timeDiffInMs / (1000 * 60 * 60 * 24));
  const hours = Math.floor((timeDiffInMs % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60));
  const minutes = Math.floor((timeDiffInMs % (1000 * 60 * 60)) / (1000 * 60));

  const parts = [];
  if (days > 0) parts.push(pluralize(days, 'day'));
  if (hours > 0) parts.push(pluralize(hours, 'hour'));
  if (minutes > 0) parts.push(pluralize(minutes, 'minute'));

  return parts.length > 0 ? parts.join(', ') : '0 minutes';
};

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+      return `${totalHours} hour${totalHours > 1 ? 's' : ''}, ${remainingMinutes} minute${remainingMinutes > 1 ? 's' : ''}`;
+    } else {
+      return `${totalHours} hour${totalHours > 1 ? 's' : ''}`;
+    }
+  } else {
+    return `${totalMinutes} minute${totalMinutes !== 1 ? 's' : ''}`;
+  }
+};
+
+class OrgListCleanExecutor extends LibraryCommandletExecutor<{}> {
```
madhur310: I get the reason for extending the executor, but why embed it in the orgList.ts and not create a separate file? I personally like classes being there own files for ease of finding and reuse for later (if applicable).

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+    const removedOrgs: string[] = [];
+    const now = new Date().getTime();
+
+    try {
+      const orgAuthorizations = await AuthInfo.listAllAuthorizations();
+      if (!orgAuthorizations || orgAuthorizations.length === 0) {
+        return removedOrgs;
+      }
+
+      const authRemover = await AuthRemover.create();
```
madhur310: Does it need to be created for each invocation of this method?

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+      const orgData: Row[] = [];
+      const now = new Date().getTime();
+
+      // Use ConfigAggregator for proper default org detection and resolve to usernames
+      const configAggregator = await ConfigAggregator.create();
+      const defaultDevHubProperty = configAggregator.getPropertyValue(OrgConfigProperties.TARGET_DEV_HUB);
+      const defaultOrgProperty = configAggregator.getPropertyValue(OrgConfigProperties.TARGET_ORG);
+
+      // Resolve aliases to usernames for comparison
+      const defaultDevHubUsername = defaultDevHubProperty
```
madhur310: This would do the same:

const defaultDevHubUsername = defaultDevHubProperty ?? (await ConfigUtil.getUsernameFor(String(defaultDevHubProperty)));

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+
+          // Get aliases
+          const aliases = await ConfigUtil.getAllAliasesFor(orgAuth.username);
+          const alias = aliases?.length > 0 ? aliases[0] : '';
+
+          // Determine status by actually testing the connection
+          let status: string;
+          if (authFields.expirationDate) {
+            const expirationDate = new Date(authFields.expirationDate).getTime();
+            if (expirationDate < now) {
```
madhur310: You don't need this variable, Date.now() would yield the same result

packages/salesforcedx-vscode-core/src/commands/orgList.ts
```diff
@@ -5,26 +5,352 @@
...
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      throw new Error(nls.localize('org_list_clean_general_error', errorMessage));
+    }
+
+    return removedOrgs;
+  }
+
+  /** Display remaining orgs in a table format */
+  private async displayRemainingOrgs(): Promise<void> {
```
madhur310: This method seems quite big, can we split it up?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -57,19 +57,19 @@ export class OrgList implements vscode.Disposable {
...
     const username = await ConfigUtil.getUsernameFor(targetOrgOrAlias);
     const authFields = await getAuthFieldsFor(username);
-    const today = new Date();
+    const now = new Date().getTime();
     let expirationDate;
     if (authFields.expirationDate) {
       expirationDate = new Date(authFields.expirationDate);
     }
-    return expirationDate ? expirationDate < today : false;
+    return expirationDate ? expirationDate.getTime() < now : false;
```
madhur310: don't need variable - Date.now()?

packages/salesforcedx-vscode-core/src/orgPicker/orgList.ts
```diff
@@ -57,19 +57,19 @@ export class OrgList implements vscode.Disposable {
...
   }
 
-  public async filterAuthInfo(orgAuthorizations: OrgAuthorization[]): Promise<string[]> {
+  public async filterAuthInfo(orgAuthorizations: OrgAuthorization[], showExpired: boolean = false): Promise<string[]> {
     const targetDevHub = await OrgAuthInfo.getDevHubUsername();
 
     const authList = [];
-    const today = new Date();
+    const today = new Date().getTime();
```
madhur310: same

packages/salesforcedx-vscode-core/src/commands/orgList.ts
CristiCanizales: It had always been like that, in the same file. I can make the refactor 🙌

packages/salesforcedx-vscode-core/src/commands/orgList.ts
CristiCanizales: Yes, it's only called once per command run though.

madhur310 (approved):

PR #6301, opened 2025-05-23
fix: update vscodeVersionRegex to support minor releases with 3-digits (e.g. "v1.100.2")

AndrewStopchenko-SO:
### What does this PR do?
VS Code minor release number raised into 3 digits now (current latest version is _v1.**100**.2_).
Updated the `vscodeVersionRegex` to allow minor releases with 3 digits.

### What issues does this PR fix or reference?
None.

### Functionality Before
- Validations are passing only for VS Cpde Versions having 2-digits in the minor release: _"v1.99.1"_
- Reporting issues with _"**VS Code Version:** v1.100.2"_ value would fail the validation check, and issue gets marked with the `label:missing required information` label (see [this job run results](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/15215888713/job/42801186936)).

### Functionality After
Validation should pass for _"**VS Code Version:** v1.100.2"_ value.

[skip-validate-pr]

PR #6501, opened 2025-08-25
chore: recompile /validate-issue/lib/src/index.js GitHub action script

AndrewStopchenko-SO:
### What does this PR do?
Re-compiled `validate-issue` GitHub action's javascript action so that fix for VSCode Version introduced to the `src/index.ts` file (see #6301) is applied.

[skip-validate-pr]

### Functionality Before
Without recompilation, when `validate-issue` action runs for new/updated issues where the additional information provides the _VS Code version:_ value, those issues still get `missing-required-information` label added on them because the regex for VSCodeVersions still does not support 3 digits in the VSCode's minor version number in the action's `/lib/src/index.js` script file.

### Functionality After
Validation for VSCode versions should pass if the information is provided in expected format in the issue's description or comments.


daphne-sfdc:
@AndrewStopchenko-SO Thanks for the contribution!

PR #6525, opened 2025-09-02
chore: script + dep cleanup

mshanemc:
### What does this PR do?
in prep for having Eric work on our build/release process, I went through and cleaned up things that didn't seem to be used.

Please verify that nothing in here is a problem, and that there are no scripts that aren't referred to be code/gha/docs but that you personally use for convenience.

[skip-validate-pr]


kylewalke (approved):
Nice! Looks like a decent amount of red. Didnt see anything removed that I ever used.

gbockus-sf (approved):
Don't see anything concerning 👍

PR #6527, opened 2025-09-02
feat: bundle vf packages W-19480954

mshanemc:
### What does this PR do?
following the soql ext's patterns
- move copy/file and bundle-server to the esbuild process for the 2 VF servers
- package in the package dir using only vsce package
- bump css LS to current
- stop using the deprecated TextDocument and use the recommended one

other
- put all types/node devDeps at the project-level

@W-19480954@

QA:
do some validation with a vf page.  I tested snippets, css highlighting, and some minimal js script.


CristiCanizales (approved):
Changes look good, as well as local testing!

Issue #6536, opened 2025-09-04
"Execute SOQL Query with Currently Selected Text" shows "[Object]" for parent fields


jh480:
### Summary

Marking a SOQL string and then selecting "Show and Rund Commands" > "Execute SOQL Query with Currently Selected Text" will list "[Object]" for the columns of parent (relationship) fields.

### Steps To Reproduce:

1.  Mark the text in quotes "SELECT Createdby.Name FROM Account LIMIT 1"
1.  Call "Show and Rund Commands"
2. Select "Execute SOQL Query with Currently Selected Text".
3. Select "REST API"
1.  Mark the text in quotes "SELECT Createdby.Name FROM Layout LIMIT 1"
1.  Call "Show and Rund Commands"
2. Select "Execute SOQL Query with Currently Selected Text".
3. Select "Tooling API"

### Expected result

Output similar to
`sf data query -q "SELECT Createdby.Name FROM Account LIMIT 1"`
and
`sf data query -q "SELECT Createdby.Name FROM Layout LIMIT 1" --use-tooling-api`

### Actual result

_=== Query Results
CreatedBy
─────────
[Object]_ 

### Additional information

**Salesforce Extension Version in VS Code**: salesforcedx-vscode-core 64.11.1

**Salesforce CLI Version**: @salesforce/cli/2.103.7 darwin-arm64 node-v20.19.4

**OS and version**: macOS 13.7.7

**VS Code version**:  1.103.2 (Universal); Electron: 37.2.3

**Most recent version of the extensions where this was working**: salesforcedx-vscode-core 64.3.0

daphne-sfdc:
Released in v64.14.0

PR #6539, opened 2025-09-06
[W-19562256] fix: SFDX: Execute SOQL Query With Currently Selected Text correctly displays the value for parent relationship fields

daphne-sfdc:
### What does this PR do?
Fixes **SFDX: Execute SOQL Query With Currently Selected Text** to display the correct value for columns with parent relationship fields instead of displaying them as `[Object]`.  This matches the behavior of the `sf data query -q` CLI command.

Test plan here: https://salesforce.quip.com/Sp2PAVIjjyRi

#6536, @W-19562256@

### Functionality Before
Columns of parent relationship fields have values displayed as `[Object]`
Screenshot 2025-09-06 at 12 13 14 AM

### Functionality After
Columns of parent relationship fields have their value correctly displayed
Screenshot 2025-09-06 at 12 14 09 AM

### DIFF
--- a/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
+++ b/packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
@@ -136,7 +136,7 @@ export abstract class LibraryCommandletExecutor<T> implements CommandletExecutor
           return this.run(response, progress, token);
         }
       );
-      channelService.showCommandWithTimestamp(`\n${nls.localize('channel_end')} ${this.executionName}`);
+      channelService.showCommandWithTimestamp(`${nls.localize('channel_end')} ${this.executionName}`);
 
       if (this.showChannelOutput) {
         channelService.showChannelOutput();
--- a/packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
+++ b/packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
@@ -181,25 +181,99 @@ export const generateTableOutput = (records: QueryResult['records'], title: stri
     return '';
   }
 
-  const fields = Object.keys(firstRecord).filter(key => key !== 'attributes');
-  // If no fields after filtering attributes, return empty string
-  if (fields.length === 0) {
+  // Flatten nested objects into separate columns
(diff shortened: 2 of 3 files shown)


packages/salesforcedx-utils-vscode/src/commands/commandletExecutors.ts
```diff
@@ -136,7 +136,7 @@ export abstract class LibraryCommandletExecutor<T> implements CommandletExecutor
           return this.run(response, progress, token);
         }
       );
-      channelService.showCommandWithTimestamp(`\n${nls.localize('channel_end')} ${this.executionName}`);
+      channelService.showCommandWithTimestamp(`${nls.localize('channel_end')} ${this.executionName}`);
```
daphne-sfdc, line 139: This change caused the "Ended" message for all SFDX commands to appear on the next line after the timestamp.  I have reverted it to the original version, where the entire message appears on the same line.

Incorrect version:
```
00:13:07.523
Ended SFDX: Execute SOQL Query...
```

Correct version:
```
00:13:07.523 Ended SFDX: Execute SOQL Query...
```

packages/salesforcedx-vscode-core/src/commands/dataQuery.ts
```diff
@@ -261,8 +335,23 @@ export const formatFieldValueForDisplay = (value: unknown): string => {
...
-    return 'Id' in value ? String(value.Id) : '[Object]';
+  if (typeof value === 'object' && isRecord(value)) {
+    // Handle nested objects (like relationship fields)
+    const obj = value;
+
+    // Filter out attributes field and collect all values
+    const values = Object.entries(obj)
+      .filter(([key]) => key !== 'attributes')
+      .map(([_key, val]) => String(val))
+      .join(', ');
```
daphne-sfdc, line 346: In nested objects, the values are displayed as a comma-separated string.

For example:
```
{
  Id: '001XX0000001',
  Level1: {
    Id: '002XX0000001',
    Level2: {
      Id: '003XX0000001',
      Name: 'Deep Value'
    }
  }
}
```

In Level2, the values would become the concatenated string `'003XX0000001, Deep Value'`.

PR #6548, opened 2025-09-16
[W-18831011] feat: add hover documentation and autocompletion support for metadata XML files for RedHat XML extension

daphne-sfdc:
### What does this PR do?
We have the RedHat XML extension in our Salesforce Extension Pack Expanded.  It's meant to be used for metadata XML files, but previously it wasn't working as expected, producing unintelligible errors.  In this PR, I have removed the errors and added proper hover documentation and autocompletion to the metadata XML files.

@W-18831011@

### Functionality Before
This error appears for all metadata XML files when the RedHat XML extension is installed and activated.
Screenshot 2025-09-16 at 3 41 23 PM

### Functionality After
Proper hover documentation and autocompletion for all metadata types and internal fields according to [metadata.wsdl](https://github.com/forcedotcom/wsdl/blob/main/resources/metadata.wsdl).  Additional metadata types added according to manual testing using [NPSP](https://github.com/SalesforceFoundation/NPSP), [dreamhouse-lwc](https://github.com/trailheadapps/dreamhouse-lwc), and [ebikes-lwc](https://github.com/trailheadapps/ebikes-lwc).
Screenshot 2025-09-16 at 3 46 08 PM
Screenshot 2025-09-16 at 3 46 23 PM


packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export interface MetadataFieldInfo {
```
mshanemc: let's use type, not interface.

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+ */
+export type MetadataFieldInfo = {
+  name: string;
+  type: string;
+  description: string;
+  required?: boolean;
+};
+
+export type MetadataFieldDocumentation = {
+  name: string;
```
mshanemc: don't duplicate 4 fields just to add 1

= MetadataFieldInfo & {validValues?: string[]}

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+  public getAllMetadataTypes(): string[] {
+    return Array.from(this.documentationMap.keys()).sort();
+  }
+
+  /**
+   * Load metadata documentation from the official Salesforce Developer Guide
+   * This references the authoritative documentation at:
+   * https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_types_list.htm
+   */
+  private async loadMetadataDocumentation(): Promise<void> {
```
mshanemc: see linter warning

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -522,6 +524,21 @@ const initializeProject = async (extensionContext: vscode.ExtensionContext) => {
...
+  // Initialize metadata XML support
+  const metadataXmlSupport = MetadataXmlSupport.getInstance();
+  await metadataXmlSupport.initializeMetadataSupport(extensionContext);
+
+  // Initialize metadata hover provider
+  const metadataHoverProvider = new MetadataHoverProvider();
+  await metadataHoverProvider.initialize();
+
+  // Register hover provider for XML files
+  const hoverProviderDisposable = vscode.languages.registerHoverProvider(
```
mshanemc, line 536: Why was this necessary to register a new provider, as opposed to providing the built xsd to the built-in xml extension?

mshanemc: via, ex: https://github.com/redhat-developer/vscode-xml/blob/main/docs/Preferences.md#file-associations

packages/salesforcedx-vscode-core/src/metadataSupport/metadataXmlSupport.ts
```diff
@@ -0,0 +1,95 @@
...
+import { channelService } from '../channels';
+import { nls } from '../messages';
+
+/**
+ * Provides XML schema support for Salesforce metadata files using RedHat XML extension
+ */
+export class MetadataXmlSupport {
+  private static instance: MetadataXmlSupport;
+
+  public static getInstance(): MetadataXmlSupport {
```
mshanemc, line 18: is all the singleton boilerplate needed if there's only one place (vscode core) that instantiates it?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataXmlSupport.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+    return listOfPaths;
+  }
+
+  /**
+   * Setup RedHat XML extension with metadata schema support
+   * @param inputCatalogs - list of catalog file paths
+   * @param inputFileAssociations - list of file associations
+   */
+  private async setupRedhatXml(
```
mshanemc, line 38: does this really need to be a class at all?  can the core activate method call the 1 exported fn (initialize)

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+    const metadataType = this.extractMetadataType(lineText, word, position.character);
+    if (metadataType) {
+      const documentation = this.documentationService.getDocumentation(metadataType);
+      if (documentation) {
+        const markdownContent = new vscode.MarkdownString();
+        markdownContent.appendCodeblock(metadataType, 'xml');
+        markdownContent.appendMarkdown(documentation.description);
+
+        if (documentation.fields && documentation.fields.length > 0) {
+          markdownContent.appendMarkdown('\n\n**Key Fields:**\n');
```
mshanemc, line 60: oh, maybe this is the reason for needing a custom hover?  because just the `documentation` is too big/unusable?

My preference is that the PR body explains some of the non-obvious "why"

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+      fileName.includes('.workflow') ||
+      fileName.includes('.validationRule') ||
+      fileName.includes('.customField')
+    );
+  }
+
+  /**
+   * Extract metadata type or field from the current line and cursor position
+   */
+  private extractMetadataType(lineText: string, word: string, cursorPosition: number): string | null {
```
mshanemc, line 161: here and elsewhere: if it doesn't use `this` it shouldn't be a method.  put a fn in module scope.  Then if you want you can write unit tests for it without having to create a class, easily refactor, etc.

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+    return null;
+  }
+
+  /**
+   * Extract field information for internal tags within metadata
+   */
+  private extractFieldInfo(
+    document: vscode.TextDocument,
+    position: vscode.Position
+  ): { metadataType: string; fieldName: string } | null {
```
mshanemc, line 197: here and elsewhere unless required by the API, prefer `undefined` over `null`.  That'll let you omit the `return` statements at the end, and use only `return` on the early exits.

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+      const line = document.lineAt(i).text;
+      const xmlElementRegex = /<([\w:]+)(\s|>)/g;
+      let match;
+
+      while ((match = xmlElementRegex.exec(line)) !== null) {
+        const elementName = match[1];
+        const cleanElementName = elementName.includes(':') ? elementName.split(':')[1] : elementName;
+
+        // Check if this looks like a metadata type (starts with capital letter)
+        if (/^[A-Z]/.test(cleanElementName)) {
```
mshanemc, line 247: hmm, if we have the list of all valid metadata types (ex: from SDR registry), would that be useful vs. "any uppercase letter"?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+      fileName.includes('.cls') ||
+      fileName.includes('.component') ||
+      fileName.includes('.page') ||
+      fileName.includes('.email') ||
+      fileName.includes('.report') ||
+      fileName.includes('.dashboard') ||
+      fileName.includes('.resource') ||
+      fileName.includes('.workflow') ||
+      fileName.includes('.validationRule') ||
+      fileName.includes('.customField')
```
mshanemc, line 154: doesn't the `-meta.xml` cover these as long as the metadata is in source format?  Otherwise, that's a very small subset of metadata types to work for

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    // If no hardcoded definition, try to extract from XSD patterns
+    return Promise.resolve(this.extractFieldFromXSDPatterns(metadataType, fieldName));
+  }
+
+  /**
+   * Get all available metadata types
+   */
+  public getAllMetadataTypes(): string[] {
```
mshanemc: unused, right?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+      return commonFields[fieldName];
+    }
+
+    return null;
+  }
+
+  /**
+   * Extract field information from XSD patterns (fallback method)
+   */
+  private extractFieldFromXSDPatterns(metadataType: string, fieldName: string): MetadataFieldDocumentation | null {
```
mshanemc, line 655: also probably shouldn't be a method if it can be a pure function.

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+      masterLabel: {
+        name: 'masterLabel',
+        type: 'string',
+        description: 'The display label for the metadata component. This is what users see in the Salesforce UI.',
+        required: false
+      }
+    };
+
+    // Metadata type specific fields
+    const typeSpecificFields: Record<string, Record<string, MetadataFieldDocumentation>> = {
```
mshanemc, line 501: a big immutable const like this can go in its own file and then be imported.  The class is getting really, really long and scrolling through this (and some of the other consts like known urls, commonFields, etc isn't important )

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+        developerGuideUrls: this.getDeveloperGuideUrls(typeName)
+      });
+    }
+  }
+
+  /**
+   * Get official metadata types from the Salesforce Metadata API Developer Guide
+   * Source: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_types_list.htm
+   */
+  private getOfficialMetadataTypes(): Record<string, Partial<MetadataTypeDocumentation>> {
```
mshanemc, line 91: why only these?  what's "official" about them vs. the others?

definitely alphabetize so they match the docs (which are alphabetized)

mshanemc: this should probably be its own file, and somewhere in this PR should be the code/script that turns the docs page in your `Source:` link into this programatically for the next time it needs updating.

[which we could do, programmatically...ex: that's how SDR maintains https://github.com/forcedotcom/source-deploy-retrieve/blob/main/METADATA_SUPPORT.md]

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,720 @@
...
+        name: 'masterLabel',
+        type: 'string',
+        description: 'The display label for the metadata component. This is what users see in the Salesforce UI.',
+        required: false
+      }
+    };
+
+    // Metadata type specific fields
+    const typeSpecificFields: Record<string, Record<string, MetadataFieldDocumentation>> = {
+      CustomObject: {
```
mshanemc, line 502: again, what made you choose these fields instead of doing them all programatically?  And where's the code the scrapes the docs for each type to get its fields?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
daphne-sfdc: Actually I learned that there are some metadata XML files that don't end in `*-meta.xml`.  See the `unpackaged` folder in the NPSP project for lots of examples.

PR #6576, opened 2025-10-01
[W-19669685] feat: add hover documentation and autocompletion support for metadata XML files for RedHat XML extension

daphne-sfdc:
### What does this PR do?
We have the RedHat XML extension in our Salesforce Extension Pack Expanded.  It's meant to be used for metadata XML files, but previously it wasn't working as expected, producing unintelligible errors.  In this PR, I have removed the errors and added proper hover documentation and autocompletion to the metadata XML files.

@W-19669685@

### Functionality Before
This error appears for all metadata XML files when the RedHat XML extension is installed and activated.
Screenshot 2025-09-16 at 3 41 23 PM

### Functionality After
Proper hover documentation and autocompletion for all metadata types and internal fields according to the [Metadata API Developer Guide](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_intro.htm).
Screenshot 2025-09-16 at 3 46 08 PM
Screenshot 2025-09-16 at 3 46 23 PM


scripts/json_to_xsd_converter_clean.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python3
```
mshanemc, line 1: is this just a local script that's run as part of creating the xsd?  If so, it shouldn't be in resources, it should be in a separate folder like `scripts` and `.vscodeignore` so we don't ship stuff in the vsix that users don't need.

mshanemc: future PR: I'd like this to eventually be TS instead of python (avoiding extra local and CI setup).

Any reason it needed to be python in the first place?

packages/salesforcedx-vscode-core/resources/metadata-catalog.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE catalog PUBLIC "-//OASIS//DTD Entity Resolution XML Catalog V1.0//EN" "http://www.oasis-open.org/committees/entity/release/1.0/catalog.dtd">
+<catalog prefer="system" xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
+  <uri name="http://soap.sforce.com/2006/04/metadata" uri="./salesforce_metadata_api_clean.xsd"/>
```
mshanemc, line 4: ooh, nice.  I see how this fits together now.

scripts/json_to_xsd_converter_clean.py
mshanemc: metadata+types_map.json should also not be in the vsix...there's something weird about the file so that I can't comment on it from inside vscode, plus a warning about "many ambiguous unicode characters"

daphne-sfdc: yes let me move it to `scripts` and add it to `.vscodeignore`

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -349,6 +351,21 @@ const initializeProject = async (extensionContext: vscode.ExtensionContext) => {
   registerPushOrDeployOnSave();
   await decorators.showOrg();
 
+  // Initialize metadata XML support
+  const metadataXmlSupport = MetadataXmlSupport.getInstance();
+  await metadataXmlSupport.initializeMetadataSupport(extensionContext);
```
mshanemc: are these 2 dependent on each other?  any reason that can't be run in parallel?

concern here is core ext startup time

packages/salesforcedx-vscode-core/src/metadataSupport/metadataHoverProvider.ts
```diff
@@ -0,0 +1,308 @@
...
+import { VALID_METADATA_TYPES } from '../constants';
+import { MetadataDocumentationService } from './metadataDocumentationService';
+
+/**
+ * Provides hover documentation for Salesforce metadata types in XML files
+ */
+export class MetadataHoverProvider implements vscode.HoverProvider {
+  private documentationService: MetadataDocumentationService;
+
+  constructor() {
```
mshanemc, line 206: I think it's awkward to have both a `new` (that kinda does nothing) and then a required `initialize`.

What if `provideHover` first checked for initialization and then initialized the docService only if it wasn't already?

That way, you defer that big file load until the first time someone actually hovers?

Alternatively, if you really want to do everything as soon as it loads, we could make the ctor private and have the `initialize` method return the class instance so that that's the only way to instantiate it.

packages/salesforcedx-vscode-core/src/index.ts
daphne-sfdc: you want `await MetadataXmlSupport.getInstance().initializeMetadataSupport(extensionContext);` right?

packages/salesforcedx-vscode-core/test/jest/metadataSupport/metadataHoverProvider.test.ts
```diff
@@ -0,0 +1,860 @@
...
+      return {
+        start: { line: position.line, character: start },
+        end: { line: position.line, character: end }
+      } as vscode.Range;
+    }),
+    lineAt: jest.fn((line: number) => ({
+      text: content.split('\n')[line] || '',
+      lineNumber: line
+    }))
+  }) as any;
```
mshanemc, line 94: nah, I think there's an intended type here (ex TextDocument?)

packages/salesforcedx-vscode-core/test/jest/metadataSupport/metadataHoverProvider.test.ts
```diff
@@ -0,0 +1,860 @@
...
+    jest.clearAllMocks();
+  });
+
+  describe('isMetadataFile', () => {
+    it('should identify metadata files correctly', () => {
+      const metadataDoc = createMockDocument('test-meta.xml', '<ApexClass>');
+      const regularDoc = createMockDocument('test.txt', 'some text');
+
+      // Use reflection to access private method for testing
+      const isMetadataFile = (hoverProvider as any).isMetadataFile;
```
mshanemc: the reason you have to do all the TS workarounds is because these are private methods.  but many (including isMetadatafile) shouldn't have been methods in the first place (write a fn, export it, pass it inputs and assert the result).  Then it doesn't have to have the class instantiation, or mocking of the parts it doesn't care about, etc.  Just a pure function.

rule: if it doesn't use `this` it's not a method.

packages/salesforcedx-vscode-core/test/jest/metadataSupport/metadataHoverProvider.test.ts
```diff
@@ -0,0 +1,860 @@
...
+        '<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">',
+        'ApexClass',
+        5
+      );
+
+      expect(result).toBe('ApexClass');
+    });
+
+    it('should handle namespaced elements', () => {
+      const extractMetadataType = (hoverProvider as any).extractMetadataType;
```
mshanemc: I know it's a test, but `any` opens up cans of worms.  this also didn't need to be a method

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,323 @@
...
+   * Get documentation for a specific metadata type
+   */
+  public getDocumentation(metadataType: string): MetadataTypeDocumentation | null {
+    return this.documentationMap.get(metadataType) ?? null;
+  }
+
+  /**
+   * Get documentation for a specific field within a metadata type
+   */
+  public getFieldDocumentation(metadataType: string, fieldName: string): Promise<MetadataFieldDocumentation | null> {
```
mshanemc: why did this need to be async?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,323 @@
...
+      console.warn('Error extracting fields from complex type:', error);
+    }
+
+    return fields;
+  }
+
+  /**
+   * Get hardcoded field definitions for common metadata fields
+   */
+  private getFieldDefinitions(metadataType: string, fieldName: string): MetadataFieldDocumentation | null {
```
mshanemc: why is metadataType unused?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -0,0 +1,323 @@
...
+        elementsContainer = sequence;
+      } else if (choice?.['xsd:element']) {
+        elementsContainer = choice;
+      }
+
+      if (!elementsContainer?.['xsd:element']) {
+        return fields;
+      }
+
+      const elements = Array.isArray(elementsContainer['xsd:element'])
```
mshanemc, line 208: there's a lot of danger here with all of it being `any` (hopefully you see all the yellow warnings).

that would be one advantage of using TS to create the xsd--you'd have types for the JS equivalent of the xsd...then when you parse it back here, you'd have types for that and it would be a lot safer.

mshanemc: so that should be part of the "use TS, not python" WI

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
daphne-sfdc: It was a relic from when the XSD used to come from metadata.wsdl.  I had reused that branch and replaced that XSD with my current one.

This function used to contain fields for specific metadata types as a hard-coded fallback.  However, with the new XSD generated from the Metadata XML Developer Guide, the fallback is no longer needed, as the fields are taken from the documentation and parsed into the XSD.

mshanemc:
QA:
this has 2 known bugs
1. https://github.com/forcedotcom/salesforcedx-vscode/pull/6577#issuecomment-3357302829 (non-unique nested children of metadata types all get the hover doc for the first of that name)
2. the python script uses `in` (stringA in stringB) such that any type that contains the string `date` is considered a date. We found this in `Flow.UpdateRecords`.  Those need to be much less greedy than `in`
```elif 'date' in field_type:
        if 'time' in field_type:
            return 'xsd:dateTime'
        else:
            return 'xsd:date'
 ```

PR #6577, opened 2025-10-01
chore: only activate if not already active, type the rh ext api

mshanemc:


mshanemc:
QA using current non-insiders vscode + vsix for all these ext

✅ looking at an LWC meta-xml in dreamhouse
✅ LWC stuff looks fine (hover, goto def from html => js method)

📓 ux: I still don't love the "key fields" thing.  They're really just "the first fields on the list" right?  

❓ what's the "anyType" about?  Is that a complex type or just something we're missing (refers to LayoutSection[] but that's not in the types_map)
 
Screenshot 2025-10-01 at 11 52 22 AM

Can you confirm that even though the key is Layout.behavior (metadataType.xmlElementName) that we're not going to get collisions for very complex types with generic child names ex:
- `Foo.Bar.Name` (would show up as Foo.Name)
- `Foo.Baz.Name` (would also show up as Foo.Name)
[it could be that only the display name is affected and that they're unique by "full xml path" and there's nothing to worry about here]

Screenshot 2025-10-01 at 11 55 05 AM


actually, from spot checking this looks like it's a bug
repro: dreamhouse, using the Create_property flow (they're all getting the first `value`'s hover
Screenshot 2025-10-01 at 12 01 12 PM
Screenshot 2025-10-01 at 12 00 42 PM

PR #6613, opened 2025-10-27
feat: web-compatible org browser W-19881349

mshanemc:
### What does this PR do?
behind a new config (default false), enable the new web+desktop compatible org browser [experimental]
ships the new services extension for runtime use by other extensions outside the monorepo

#<Insert GitHub Issue>, @W-19881349@

### DIFF
--- a/packages/salesforcedx-vscode-services/src/observability/applicationInsightsWebExporter.ts
+++ b/packages/salesforcedx-vscode-services/src/observability/applicationInsightsWebExporter.ts
@@ -11,11 +11,11 @@
  * We want to send spans that have no parent span as CustomEvents, and capture their properties from the span.
  * Additionally, we want to send any failed spans as Exceptions (the parent span may not finish if a child span fails)
  */
-import * as Match from 'effect/Match';
 import { SpanKind, SpanStatusCode } from '@opentelemetry/api';
 import { ExportResult, ExportResultCode } from '@opentelemetry/core';
 import { ReadableSpan, SpanExporter } from '@opentelemetry/sdk-trace-base';
 import { TelemetryReporter } from '@vscode/extension-telemetry';
+import * as Match from 'effect/Match';
 import { workspace } from 'vscode';
 import { DEFAULT_AI_CONNECTION_STRING } from './appInsights';
 // TODO: should this be in Effect?


packages/salesforcedx-utils-vscode/src/providers/sourceTrackingProvider.ts
```diff
@@ -6,7 +6,10 @@
  */
 
 import { Connection, Org, SfProject } from '@salesforce/core';
-import { SourceTracking, SourceTrackingOptions } from '@salesforce/source-tracking';
+import type {
+  SourceTrackingOptions,
+  SourceTracking
+} from '@salesforce/source-tracking' with { 'resolution-mode': 'import' };
```
mshanemc, line 12: these import changes work with the existing cjs STL package, but will ALSO work when STL goes ESM (future TBD).

it might also defers loading STL until/if needed (depending on how all these classes are created)

packages/salesforcedx-vscode-core/test/jest/mock-source-tracking.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// this file somehow became necessary when STL moved to ESM module because of mocking differences.
+
+export const SourceTracking = {
```
mshanemc, line 10: leaving this to avoid doing it later when STL goes ESM.  currently unnecessary

packages/salesforcedx-vscode-org-browser/test/playwright/fixtures/desktopFixtures.ts
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// This is Node.js test infrastructure, not extension code
+import { test as base, _electron as electron, type ElectronApplication, type Page } from '@playwright/test';
```
mshanemc, line 9: a bunch of these don't really belong in orgBrowser, but probably in the services extension's test folder or a shared testing pkg for playwright.

scripts/bundling/buffer-global.js
mshanemc, line 1: here and elsewhere in bundling/polyfills stuff:
these are probably excessive.  
Now that we have some e2e test coverage, it might be possible to back out some of the polyfills and find out which ones really mattered.

scripts/bundling/url-polyfill.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  URL: globalThis.URL,
```
mshanemc, line 2: publicly available polyfills didn't have some of the recent URL changes node made

packages/salesforcedx-vscode-core/src/commands/deleteSource.ts
```diff
@@ -28,7 +28,7 @@ import {
   RequestStatus,
   SourceComponent
 } from '@salesforce/source-deploy-retrieve';
-import { ChangeResult, SourceTracking, deleteCustomLabels } from '@salesforce/source-tracking';
+import type { ChangeResult, SourceTracking } from '@salesforce/source-tracking' with { 'resolution-mode': 'import' };
```
peternhale, line 31: Why static import here and then dynamic imports in code?

packages/salesforcedx-vscode-org-browser/package.json
```diff
@@ -0,0 +1,144 @@
...
+  },
+  "repository": {
+    "url": "https://github.com/forcedotcom/salesforcedx-vscode"
+  },
+  "icon": "images/VSCodeBundle.png",
+  "galleryBanner": {
+    "color": "#ECECEC",
+    "theme": "light"
+  },
+  "version": "64.4.0",
```
peternhale:
```suggestion
  "version": "65.2.0",
```

packages/salesforcedx-vscode-org-browser/package.json
```diff
@@ -0,0 +1,144 @@
...
+  },
+  "categories": [
+    "Other"
+  ],
+  "dependencies": {
+    "effect": "^3.18.4",
+    "vscode-uri": "^3.1.0"
+  },
+  "devDependencies": {
+    "@salesforce/core": "^8.23.1",
```
peternhale:
```suggestion
    "@salesforce/core": "^8.23.3",
```

packages/salesforcedx-vscode-org-browser/package.json
```diff
@@ -0,0 +1,144 @@
...
+    "@salesforce/source-deploy-retrieve": "^12.25.0",
+    "salesforcedx-vscode-services": "64.4.0",
```
peternhale:
```suggestion
    "salesforcedx-vscode-services": "65.2.0",
```

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
```diff
@@ -0,0 +1,104 @@
...
+export class MetadataDescribeService extends Effect.Service<MetadataDescribeService>()('MetadataDescribeService', {
+  effect: Effect.gen(function* () {
+    // a task that can be cached - uses the key parameter for caching
+    const cacheableDescribe = (
+      _key: string = 'cached'
+    ): Effect.Effect<readonly DescribeMetadataObject[], Error, DescribeContext> =>
+      Effect.flatMap(ConnectionService, svc => svc.getConnection).pipe(
+        Effect.flatMap(conn =>
+          Effect.tryPromise({
+            try: () => conn.metadata.describe(),
```
peternhale, line 32: How can I use this service for an explicit version?

packages/salesforcedx-vscode-services/src/core/projectService.ts
```diff
@@ -0,0 +1,57 @@
...
+const resolveSfProject = (fsPath: string): Effect.Effect<SfProject, Error, never> =>
+  Effect.tryPromise({
+    try: () => SfProject.resolve(fsPath),
+    catch: error => new Error('Project Resolution Error', { cause: error })
+  }).pipe(Effect.withSpan('resolveSfProject', { attributes: { fsPath } }));
+
+// Global cache - created once at module level, not scoped to any consumer
+const globalSfProjectCache = Effect.runSync(
+  Cache.make({
+    capacity: 10, // Maximum number of cached SfProject instances
```
peternhale, line 24: Why capacity of ten, feels like a singleton that is refreshed when changed

packages/salesforcedx-vscode-services/src/core/configService.ts
```diff
@@ -0,0 +1,58 @@
...
+const createConfigAggregator = (projectPath: string): Effect.Effect<ConfigAggregator, Error, never> =>
+  Effect.tryPromise({
+    try: () => ConfigAggregator.create({ projectPath }),
+    catch: (error: unknown) => new Error(`Failed to get ConfigAggregator at ${projectPath}: ${String(error)}`)
+  }).pipe(Effect.withSpan('createConfigAggregator (cache miss)', { attributes: { projectPath } }));
+
+// Global cache - created once at module level, not scoped to any consumer
+const globalConfigCache = Effect.runSync(
+  Cache.make({
+    capacity: 5, // Maximum number of cached ConfigAggregators
```
peternhale, line 27: why more than one copy?

packages/salesforcedx-vscode-services/src/observability/applicationInsightsWebExporter.ts
```diff
@@ -0,0 +1,124 @@
...
+    ? 'INTERNAL'
+    : kind === SpanKind.SERVER
+      ? 'SERVER'
+      : kind === SpanKind.CLIENT
+        ? 'CLIENT'
+        : kind === SpanKind.PRODUCER
+          ? 'PRODUCER'
+          : kind === SpanKind.CONSUMER
+            ? 'CONSUMER'
+            : 'UNKNOWN';
```
peternhale: Uses effect Match
```suggestion
const getSpanKindName = (kind: SpanKind): string =>
  Match.value(kind).pipe(
    Match.when(SpanKind.INTERNAL, () => 'INTERNAL'),
    Match.when(SpanKind.SERVER, () => 'SERVER'),
    Match.when(SpanKind.CLIENT, () => 'CLIENT'),
    Match.when(SpanKind.PRODUCER, () => 'PRODUCER'),
    Match.when(SpanKind.CONSUMER, () => 'CONSUMER'),
    Match.orElse(() => 'UNKNOWN')
  );
```

peternhale (approved):
Share, LGTM, a few questions and suggestions

packages/salesforcedx-vscode-core/src/commands/deleteSource.ts
mshanemc: types only static imports evaporate after compile.
dynamic imports are required for when it goes to ESM since this module is CJS.  [also defers loading STL until the first time it's used]

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
mshanemc: got a use case for overriding it?  [default is to let connection stuff gather than from project, api, org, etc the "standard" way the library does it]

peternhale: For metadata requests, probably not. Maybe a better question in the connection service. In the OAS features where we had to guard code for a given org max version. I guess getting a ref to connection one can call `retrieveMaxApiVersion`

packages/salesforcedx-vscode-org-browser/package.json
mshanemc: can't wait for these to be `*`

packages/salesforcedx-vscode-org-browser/package.json
mshanemc: am going to let lerna manage these.  These are behind because they've not been published in main

packages/salesforcedx-vscode-services/src/core/configService.ts
mshanemc: each different project will produce an aggregator that's {...global, ...local}.  So when you switch projects you get the one that was there for that project.

I'm fine with `1` if you prefer.

Was trying not to box us into the "only one sfdx project at a time" we have now

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
mshanemc: or getConnection, then inspect it for apiVersion (or change it if needed)

packages/salesforcedx-vscode-services/src/core/projectService.ts
mshanemc: same reason as above.  Maybe we never get there, but maybe we do.

both config and project are scoped by the path.

looking at this again, it probably needs a filewatcher to invalidate an entry when the project file changes (I did a few others like that)

packages/salesforcedx-vscode-services/src/core/configService.ts
peternhale: Thanks for the explanation. I'm good

PR #6619, opened 2025-10-29
fix: don't recreate telemetryReporters, just update their user/web IDs @W-20017422

mshanemc:
### What does this PR do?
don't recreate telemeryReporters, just update their user/web IDs
- only applies to o11y/appInsights since they're the only ones the care about the update

did some other linter cleanup around error typing (less linter warnings, more type safety) and simplified the telemetry classes (bye, TelemetryBuilder)

@W-20017422@

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-vscode-lwc/package.json
+++ b/packages/salesforcedx-vscode-lwc/package.json
@@ -40,7 +40,7 @@
     "which": "1.3.1"
   },
   "devDependencies": {
-    "salesforcedx-vscode-core": "65.2.0",
+    "salesforcedx-vscode-core": "65.3.0",
     "@types/uuid": "^3.4.8",
     "@types/which": "^1.3.1",
     "esbuild": "0.25.10",


packages/salesforcedx-vscode-lwc/src/commands/lightningLwcStart.ts
```diff
@@ -113,6 +114,12 @@ export class LightningLwcStartExecutor extends SfCommandletExecutor<{}> {
           );
         }
 
+        const coreExtension = vscode.extensions.getExtension<SalesforceVSCodeCoreApi>(
+          'salesforce.salesforcedx-vscode-core'
+        );
+        if (coreExtension) {
+          coreExtension.exports.telemetryService.sendCommandEvent(execution.command.logName, startTime);
+        }
         this.logMetric(execution.command.logName, startTime);
```
daphne-sfdc, line 123: The same telemetry is actually sent twice when this command is run!  We can delete Line 123?

daphne-sfdc: Also, I see that you're using this new method for sending telemetry in **lightningLwcStart.ts**, but the other 3 commands in this area (**lightningLwcOpen.ts**, **lightningLwcStop.ts**, **lightningLwcPreview.ts**) are still using the old method.  You're planning to migrate the other 3 commands to the new version later, right?

PR #6620, opened 2025-10-29
W-20064193: move lls and remove fs references

madhur310:
@W-20064193@
move lls and remove fs references

### DIFF
--- a/packages/salesforcedx-aura-language-server/src/tern-server/ternAura.ts
+++ b/packages/salesforcedx-aura-language-server/src/tern-server/ternAura.ts
@@ -5,8 +5,9 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 // @ts-nocheck as this is a third party library
-import { extractJsonFromImport, Logger } from '@salesforce/salesforcedx-lightning-lsp-common';
+import { extractJsonFromImport, Logger, normalizePath } from '@salesforce/salesforcedx-lightning-lsp-common';
 import * as walk from 'acorn-walk';
+import * as path from 'node:path';
 import * as infer from '../tern/lib/infer';
 import * as tern from '../tern/lib/tern';
 import * as auraTypesJsonImport from './aura_types.json';
@@ -42,21 +43,20 @@ const forAllPropsPurgeable = infer.constraint({
 });
 
 const getFilename = (filename: string): string => {
-  if (server.options.projectDir.endsWith('/')) {
-    return server.options.projectDir + filename;
-  }
-  return `${server.options.projectDir}/${filename}`;
+  return path.join(server.options.projectDir, filename);
 };
 
 const isBlocklisted = (filename: string): boolean => {
-  let ret = filename.endsWith('/scrollerLib/bootstrap.js');
-  ret = ret ?? filename.endsWith('ExportSymbolsHelper.js');
+  // Normalize path separators for cross-platform compatibility
+  const normalized = normalizePath(filename);
+  let ret = normalized.endsWith('/scrollerLib/bootstrap.js');
+  ret = ret ?? normalized.endsWith('ExportSymbolsHelper.js');
   return ret;
 };
 
 const readFile = async (filename: string): Promise<string> => {
   let normalized = filename;
-  if (!normalized.startsWith('/')) {
+  if (!path.isAbsolute(normalized)) {
     normalized = getFilename(normalized);
   }
 
@@ -71,11 +71,6 @@ const readFile = async (filename: string): Promise<string> => {
     // Handle file not found or other errors
     return '';
   }
-};
-
-const baseName = (path: string): string => {
-  const lastSlash = path.lastIndexOf('/');
-  return lastSlash === -1 ? path : path.slice(lastSlash + 1);
 };
 
 const trimExt = (path: string): string => {
@@ -134,7 +129,7 @@ const findAndBindComponent = (type: any, _server: any, cx: any, infr: any): void
 const findAndBindHelper = (type: any, servr: any, modules: any, file: any): void => {
   const helperFile = getHelper(file.name);
 
-  const bn = trimExt(baseName(helperFile));
+  const bn = trimExt(path.basename(helperFile));
   const r = servr.findFile(helperFile);
   if (!r) {
     servr.addFile(helperFile);
@@ -205,7 +200,7 @@ const connectModule = async (file: any, out: any): Promise<void> => {
       Logger.log(`Process helper exports ${file.name}`);
       let outObj;
       if (!out.getType()) {
-        const type = baseName(file.name).replace(/.js$/, '');
+        const type = path.basename(file.name).replace(/.js$/, '');
         outObj = new infer.Obj(true);
         outObj.origin = file.name;
         outObj.originNode = file.ast;
@@ -234,7 +229,7 @@ const connectModule = async (file: any, out: any): Promise<void> => {
                 //    });
                 // -- would have worked, but didnt'
                 // delete all types, and re-add...
-                const target = outObj.defProp(baseName(file.name).replace(/.js$/, ''));
+                const target = outObj.defProp(path.basename(file.name).replace(/.js$/, ''));
                 const types = target.types;
                 types.length = 0;
                 //note: propogate calls addType on the target
@@ -421,7 +416,8 @@ tern.defineQueryType('cleanup-file', {
   run: (_server: any, query: any) => {
     const files = query.files;
     files.forEach(f => {
-      let ff = f;
+      let ff = normalizePath(f);
+      // Remove leading slash to normalize absolute paths
       if (ff.startsWith('/')) {
         ff = ff.slice(1);
       }


eslint.config.mjs
```diff
@@ -376,6 +379,7 @@ export default [
     files: [
       'packages/salesforcedx**/test/jest/**/*',
       'packages/salesforcedx**/test/unit/**/*',
+      'packages/salesforcedx**/src/__tests__/**/*',
```
mshanemc: this should cover both `src/__tests__` AND `packages/salesforcedx-lwc-language-server/src/javascript/__tests__/fixtures/**`, right?
```suggestion
      'packages/salesforcedx**/src/**/__tests__/**/*',
```

eslint.config.mjs
```diff
@@ -391,7 +395,14 @@ export default [
       '@typescript-eslint/no-unsafe-member-access': 'off',
       '@typescript-eslint/no-unsafe-call': 'off',
       '@typescript-eslint/no-unsafe-assignment': 'warn',
-      '@typescript-eslint/no-unused-expressions': 'warn',
+      '@typescript-eslint/no-unused-expressions': [
+        'warn',
+        {
```
mshanemc: why?  https://eslint.org/docs/latest/rules/no-unused-expressions#options

defaults are false for these, right?

package.json
```diff
@@ -104,10 +104,8 @@
     "vscode:sha256": "node ./scripts/generate-sha256.js >> ./SHA256",
     "vscode:publish": "lerna run vscode:publish --concurrency 1",
     "watch": "lerna run --parallel watch",
-    "link-lsp": "yarn link @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common && lerna exec yarn link @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lightning && lerna exec yarn link @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lwc",
```
mshanemc: nice!

packages/salesforcedx-aura-language-server/.npmignore
mshanemc, line 1: do you really need npm ignore if not publishing npm packages?

[and if we do decide to npm publish these, which we can, I'd rather use `pjson.files` to specify what I want instead of having to maintain an ignore file

packages/salesforcedx-aura-language-server/.prettierrc
mshanemc, line 1: why is this necessary?  Why not have it work like the rest of the packages in the monorepo?

Or is the file required because there's a siblinb .prettierignore? (if so, maybe put the prettierIgnore stuff in the top-level one?

packages/salesforcedx-aura-language-server/.vscode/launch.json
```diff
@@ -0,0 +1,46 @@
+{
```
mshanemc: cool, I didn't know you could do these at the package level!

packages/salesforcedx-aura-language-server/.vscode/settings.json
mshanemc, line 1: are these diffrent enough from the top-level ones in the monorepo to need their own file?

packages/salesforcedx-aura-language-server/.vscode/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+    "files.exclude": {
+      "**/.git": true,
+      "**/.svn": true,
+      "**/.hg": true,
+      "**/CVS": true,
+      "**/.DS_Store": true
+    },
+    "search.exclude": {
+      "**/lib": true,
```
mshanemc: I'd like to align these new packages with how the rest of the packages in the mono repo are structured, for consistency and less mental overhead

1. compile src => `out` (js)
1. bundle `out` => `dist`

so not `lib`

packages/salesforcedx-aura-language-server/.vscode/settings.json
```diff
@@ -0,0 +1,20 @@
...
+      "**/node_modules": true,
+      "**/bin": true
+    },
+    "eslint.packageManager": "yarn",
```
mshanemc: and no yarn

packages/salesforcedx-aura-language-server/CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
+# Requirements
+
+* Node 8+
```
mshanemc: please read this or delete it

packages/salesforcedx-aura-language-server/package.json
```diff
@@ -0,0 +1,95 @@
...
+    "acorn-walk": "^6.0.0",
+    "line-column": "^1.0.2",
+    "@types/vscode": "^1.90.0",
+    "vscode-html-languageservice": "^5.0.0",
+    "vscode-languageserver": "^9.0.1",
+    "vscode-languageserver-textdocument": "^1.0.12",
+    "vscode-languageserver-types": "3.17.5",
+    "vscode-uri": "1.0.6"
+  },
+  "devDependencies": {
```
mshanemc, line 79: make sure these are not duplicating devDeps from the top-level.

only do that if using the top-level version would break something

packages/salesforcedx-aura-language-server/scripts/build.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+const shell = require('shelljs');
```
mshanemc: let's get rid of shelljs in the 3 build scripts.  basic node fs can do all these things just fine.

packages/salesforcedx-aura-language-server/scripts/transform-aura-system.js
```diff
@@ -0,0 +1,31 @@
+// NOTE: This script is not used in the build.js, but needs to be run manually to update the aura-system.json file
```
mshanemc, line 1: do we use this?

ex: it's possible that the "original" `aura-system.json` comes from some other place (ex: core, or some other closed repo).  If so, we probably want to use that as the checked in file and have `transformed-aura-system.json` as gitignored, and then have the transform run automatically somewhere in the compile step

packages/salesforcedx-aura-language-server/src/aura-indexer/__tests__/indexer.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+const normalize = (start: string, p: string): string => {
+    // Fix relative paths on windows
+    let normalizedStart = start;
+    let normalizedP = p;
+    if (normalizedStart.includes('\\')) {
+        normalizedStart = normalizedStart.replace(/\\/g, '/');
+    }
+    if (normalizedP.includes('\\')) {
+        normalizedP = normalizedP.replace(/\\/g, '/');
+    }
```
mshanemc: we have `string.replaceAll` these days so we don't have to do regex or do weird stuff for the beginning.

there's also path.posix.normalize if you're just trying to get that

packages/salesforcedx-aura-language-server/src/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export * from './util/componentUtil';
```
mshanemc: let's export them by name so we really mean them to be expored outside the LS pkg and if something stops using them we'll know

packages/salesforcedx-aura-language-server/src/markup/auraTags.ts
```diff
@@ -0,0 +1,64 @@
...
+import AuraIndexer from '../aura-indexer/indexer';
+
+let indexer: AuraIndexer;
+
+const getAuraTags = (): Map<string, TagInfo> => {
+    if (indexer) {
+        return indexer.getAuraTags();
+    }
+    return new Map();
+};
```
mshanemc: etc.
```suggestion
const getAuraTags = (): Map<string, TagInfo> => indexer?.getAuraTags() ?? new Map();
```

packages/salesforcedx-aura-language-server/src/resources/lwc-standard.json
mshanemc, line 1: maybe we only need one of these?  the LWC one seems to be more accurate (aria-described-by)?  and maybe that one gets copied/transformed as needed by scripts?

then there's only one place to update it? [I'd suggest having aura LS grab it from LWC ls since it's LWC's stuff]

packages/salesforcedx-aura-language-server/src/tern/.no-sublime-package
mshanemc, line 1: ?

packages/salesforcedx-aura-language-server/src/tern/.travis.yml
mshanemc, line 1: ?

packages/salesforcedx-aura-language-server/package.json
```diff
@@ -0,0 +1,95 @@
...
+  "version": "64.16.1",
+  "description": "Language server for Aura components.",
+  "publisher": "salesforce",
+  "engines": {
+    "vscode": "^1.90.0"
+  },
+  "main": "out/src/index.js",
+  "typings": "out/src/index.d.ts",
+  "sideEffects": false,
+  "exports": {
```
mshanemc, line 13: are there really consumers of all these things outside the LS package?

packages/salesforcedx-aura-language-server/tsconfig.json
```diff
@@ -0,0 +1,50 @@
+{
+    "compilerOptions": {
```
mshanemc: only override what's necessary but doesn't match the common top-level one.  recommend 
  "extends": "../../../tsconfig.common.json",
like all the other ones do

packages/salesforcedx-vscode-automation-tests/test/specs/createOasDoc.e2e.ts
```diff
@@ -328,7 +328,7 @@ describe('Create OpenAPI v3 Specifications', () => {
 
           // Get all open tabs
           const openTabs = await editorView.getOpenTabs();
-          if (!openTabs || openTabs.length === 0) {
+          if (openTabs?.length === 0) {
```
mshanemc: I think the optionalChain is unnecessary since getOpenTabs says it return an array

packages/salesforcedx-lwc-language-server/src/javascript/compiler.ts
```diff
@@ -0,0 +1,191 @@
...
+        for (const member of metadata.classMembers) {
+            if (member.type === memberType) {
+                if (!memberDecorator || member.decorator === memberDecorator) {
+                    members.push(member);
+                }
+            }
+        }
+    }
+    return members;
+};
```
mshanemc: etc
```suggestion
export const getClassMembers = (metadata: Metadata, memberType: string, memberDecorator?: string): ClassMember[] =>
    (metadata.classMembers ?? []).filter((member) => member.type === memberType && (!memberDecorator || member.decorator === memberDecorator));
```

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -0,0 +1,646 @@
...
+};
+
+export default class Server {
+    public readonly connection: Connection = createConnection();
+    public readonly documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
+    private context: LWCWorkspaceContext;
+    private workspaceFolders: WorkspaceFolder[];
+    private workspaceRoots: string[];
+    public componentIndexer: ComponentIndexer;
+    private typingIndexer: TypingIndexer;
```
mshanemc: I wonder why that's not being used?

packages/salesforcedx-lwc-language-server/package.json
```diff
@@ -0,0 +1,128 @@
...
+  "version": "64.16.1",
+  "description": "Language server for Lightning Web Components.",
+  "publisher": "salesforce",
+  "engines": {
+    "vscode": "^1.90.0"
+  },
+  "main": "out/src/indexer.js",
+  "typings": "out/src/indexer.d.ts",
+  "sideEffects": false,
+  "exports": {
```
mshanemc, line 13: same question here...only export what's imported by something else (keeps the API smaller and reduces the chances for having to make changes in 2 packages)

packages/salesforcedx-lwc-language-server/package.json
```diff
@@ -0,0 +1,128 @@
...
+    "@lwc/metadata": "12.2.0",
+    "@lwc/template-compiler": "8.16.0",
+    "@salesforce/apex": "0.0.21",
+    "@salesforce/label": "0.0.21",
+    "@salesforce/salesforcedx-lightning-lsp-common": "64.16.1",
+    "@salesforce/resourceurl": "0.0.21",
+    "@salesforce/schema": "0.0.21",
+    "babel-types": "^6.26.0",
+    "camelcase": "^6.0.0",
+    "change-case": "^4.1.1",
```
mshanemc: if we have change-case, why not use that to do camel-case and drop a dep?

packages/salesforcedx-lwc-language-server/src/baseIndexer.ts
```diff
@@ -0,0 +1,37 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { IFileSystemProvider } from '@salesforce/salesforcedx-lightning-lsp-common';
+import * as path from 'node:path';
+
+// Utility function to resolve workspace root
+export const getWorkspaceRoot = (workspaceRoot: string): string => path.resolve(workspaceRoot);
+
+// Utility function to get SFDX configuration
+export const getSfdxConfig = async (root: string, fileSystemProvider: IFileSystemProvider): Promise<any> => {
```
mshanemc: not for this PR, but I think the client should be responsible for doing this, and then providing all the config the LS needs to do its job.

packages/salesforcedx-vscode-lightning/esbuild.config.mjs
```diff
@@ -6,10 +6,64 @@
  */
 import { nodeConfig } from '../../scripts/bundling/node.mjs';
 import { build } from 'esbuild';
+import { copyFileSync, mkdirSync, existsSync } from 'fs';
+import { join, dirname } from 'path';
```
mshanemc: what's with the unused stuff?

packages/salesforcedx-vscode-lightning/esbuild.config.mjs
```diff
@@ -6,10 +6,64 @@
...
 
 await build({
   ...nodeConfig,
-  external: ['vscode', 'applicationinsights', '@salesforce/lightning-lsp-common', '@salesforce/aura-language-server'],
+  external: ['vscode', 'applicationinsights', '@salesforce/lightning-lsp-common'],
```
mshanemc: I think the current version of nodeConfig has an `external` property (just `vscode` for now)

You can spread that and then add what you need.  Check if `applicationinsights` is necessary...most pkgs don't have to do that.

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import equal from 'deep-equal';
+import * as jsonc from 'jsonc-parser';
```
mshanemc: not for this PR, but we'll need to find a way to get rid of this since it won't bundle

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -0,0 +1,221 @@
...
+export const isAuraWatchedDirectory = async (context: BaseWorkspaceContext, uri: string): Promise<boolean> => {
+    const file = toResolvedPath(uri);
+    return await context.isFileInsideAuraRoots(file);
+};
+
+/**
+ * @return true if changes include a directory delete
+ */
+// TODO This is not waiting for the response of the promise isLWCWatchedDirectory, maybe we have the same problem on includesDeletedAuraWatchedDirectory
+export const includesDeletedLwcWatchedDirectory = async (context: BaseWorkspaceContext, changes: FileEvent[]): Promise<boolean> => {
```
mshanemc: I don't think this is "real code" but is a test helper based on its references.

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -0,0 +1,221 @@
...
+    for (const event of changes) {
+        if (event.type === FileChangeType.Deleted && !event.uri.includes('.') && (await isLWCWatchedDirectory(context, event.uri))) {
+            return true;
+        }
+    }
+    return false;
+};
+
+export const includesDeletedAuraWatchedDirectory = async (context: BaseWorkspaceContext, changes: FileEvent[]): Promise<boolean> => {
```
mshanemc: also looks unused

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -0,0 +1,221 @@
...
+export const isAuraRootDirectoryCreated = (context: BaseWorkspaceContext, changes: FileEvent[]): boolean => {
+    for (const event of changes) {
+        if (event.type === FileChangeType.Created && isAuraDirectory(context, event.uri)) {
+            return true;
+        }
+    }
+    return false;
+};
+
+export const unixify = (filePath: string): string => filePath.replace(/\\/g, '/');
```
mshanemc: this only looks to be used locally and in the index.ts.

It'd be worth reviewing the barrel (top-level index.ts) to see if there are exports from common not being used by either LS.  As I poked around here, I see a lot of dead code.

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -0,0 +1,221 @@
...
+export const getSfdxResource = (resourceName: string): string => join(__dirname, RESOURCES_DIR, 'sfdx', resourceName);
+
+export const getCoreResource = (resourceName: string): string => join(__dirname, RESOURCES_DIR, 'core', resourceName);
+
+/**
+ * Deep merges the 'from' object into the 'to' object
+ * (assumes simple JSON config objects)
+ * @return true if the 'to' object was modified, false otherwise
+ */
+export const deepMerge = (to: Record<string, unknown>, from: Record<string, unknown>): boolean => {
```
mshanemc: also looks unused.

packages/salesforcedx-lightning-lsp-common/src/utils.ts
mshanemc: probably something like https://github.com/fabiospampinato/tiny-jsonc

packages/salesforcedx-vscode-lwc/esbuild.config.mjs
```diff
@@ -14,10 +14,30 @@ await build({
...
   ],
   entryPoints: ['./src/index.ts'],
   outdir: 'dist'
 });
+
+// Bundle the external LWC language server package
+await build({
+  ...nodeConfig,
+  loader: { '.node': 'file' },
+  external: [
```
mshanemc, line 31: meta: have you tried without some of these?  I'm curious how bad it's going to be trying to get this to bundle (esp babel because it seems integral to lwc)

packages/salesforcedx-aura-language-server/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  extends: ['../../.eslintrc.js'],
+  rules: {
+    // Allow Node.js APIs in language server packages
```
mshanemc: just for now, right?

packages/salesforcedx-lwc-language-server/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  extends: ['../../.eslintrc.js'],
+  rules: {
+    // Allow Node.js APIs in language server packages
```
mshanemc: are you sure this is needed?

packages/salesforcedx-vscode-lwc/esbuild.config.mjs
madhur310: lsp-common one should be in there for sure. will try removing others too.

packages/salesforcedx-lwc-language-server/.eslintrc.js
madhur310: not needed, cursor intermidiate code which I missed cleaning up.

packages/salesforcedx-aura-language-server/.eslintrc.js
madhur310: not needed, cursor intermidiate code which I missed cleaning up.

packages/salesforcedx-lwc-language-server/src/baseIndexer.ts
madhur310: the method is pulling from fileSystemProvider, which is populated by the client on start up for now.

packages/salesforcedx-lwc-language-server/package.json
madhur310: good point, let me make the change

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
madhur310: I actually refactored the typingIndexer a bit, old remnant will fix it.

packages/salesforcedx-vscode-automation-tests/test/specs/createOasDoc.e2e.ts
madhur310: yep you are right, shouldn't have missed it.

packages/salesforcedx-aura-language-server/src/tern/.travis.yml
madhur310: legacy from tern, checked and its not used. Will delete it.

packages/salesforcedx-aura-language-server/src/tern/.no-sublime-package
madhur310: damn it, cursor...

packages/salesforcedx-aura-language-server/.npmignore
madhur310: good catch - yeah remnant from the move

packages/salesforcedx-aura-language-server/scripts/transform-aura-system.js
madhur310: yes this is supposed to be used, but don't think it has been updated for years now. I don't know the source of `aura-system.json`. I would like to keep this as is and do what you suggested in [W-19346602](https://gus.my.salesforce.com/a07EE00002KDFmdYAH)

packages/salesforcedx-aura-language-server/src/resources/lwc-standard.json
madhur310: Will look into it with the [ticket](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002KDFmdYAH/view).

packages/salesforcedx-lightning-lsp-common/src/utils.ts
mshanemc: make a WI for that

packages/salesforcedx-aura-language-server/src/tern/doc/manual.html
mshanemc, line 1: why do we need the demo/manual stuff in our LS ?

packages/salesforcedx-aura-language-server/src/tern/plugin/angular.js
mshanemc, line 1: what would happen if we didn't have all the tern plugins?

packages/salesforcedx-aura-language-server/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "extends": "../../tsconfig.common.json",
+  "compilerOptions": {
+    "outDir": "./out",
+    "target": "ES2022",
```
mshanemc: why can't these use the base values?  what happens when you try?

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+    updateDirectoryListing(uri: string, entries: DirectoryEntry[]): void;
+    updateFileStat(uri: string, stat: FileStat): void;
+    updateWorkspaceConfig(config: WorkspaceConfig): void;
+}
+
+/**
+ * Manages file system data received from the LSP client
+ * This replaces direct file system access in the language server
+ */
+export class FileSystemDataProvider implements IFileSystemProvider {
```
mshanemc, line 31: If you haven't already, I'd like you to get Pete's verdict on this.  It was my understanding the that the LS would need to run as its own process via a worker and what's passed in would be primitives.

Perhaps I misunderstood.  Or maybe these all run "inside" the extension rather than as a separate process and it's just "different" from apex that way?

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -0,0 +1,649 @@
...
+  }
+
+  public async isAuraMarkup(document: TextDocument): Promise<boolean> {
+    const extension = utils.getExtension(document);
+    const isAuraExtension = AURA_EXTENSIONS.includes(extension);
+    const languageIdMatches = document.languageId === 'html' || (isAuraExtension && !document.languageId);
+    return languageIdMatches && isAuraExtension && (await this.isInsideAuraRoots(document));
+  }
+
+  public async isLWCTemplate(document: TextDocument): Promise<boolean> {
```
mshanemc: only used by lwcServer via LWCWorkspaceContext

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -0,0 +1,649 @@
...
+  }
+
+  public async isFileInsideModulesRoots(file: string): Promise<boolean> {
+    const namespaceRoots = await this.findNamespaceRootsUsingTypeCache();
+    for (const root of namespaceRoots.lwc) {
+      if (utils.pathStartsWith(file, root)) {
+        return true;
+      }
+    }
+    return false;
```
mshanemc: here and elsewhere.   
```suggestion
    return (await this.findNamespaceRootsUsingTypeCache()).some(root=>utils.pathStartsWith(file, root))
    
```

hopefully not a method eventually

packages/salesforcedx-lightning-lsp-common/src/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Connection } from 'vscode-languageserver';
+
+/**
+ * Intercepts global console logging methods and redirects them to the LSP connection.
+ * This allows language server log messages to appear in the client's output panel
+ * (e.g., VS Code's Output view) instead of the server's local console.
+ *
+ * @param connection - The LSP connection to redirect console output to
+ */
+export const interceptConsoleLogger = (connection: Connection): void => {
```
mshanemc: make a WI to get rid of this and have the LS use https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage and the clients listen for those and do the right thing with them.

packages/salesforcedx-lwc-language-server/src/decorators/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export * from './lwcDecorators';
```
mshanemc: is importing them from `/decorators` any better than importing from `/lwcDecorators` ?  I think these barrel files are silly

packages/salesforcedx-lwc-language-server/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../../tsconfig.common.json",
+  "compilerOptions": {
+    "outDir": "./out",
```
mshanemc, line 4: same question...is there a need for all these overrides?

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
peternhale: @mshanemc The LSP protocol uses JSON for request payloads, so anything that can be properly converted to JSON can be used.

The client side of LLS is using function `serialize` to ensure only the data is included in the payload.

peternhale: The custom endpoint that Kyle created to pass the std apex class zip from client to server gives us evidence shows that streaming is used on the client/server connection. Payload size may not be an issue for the connection, but transferring contents of workspace during init may impact perceived time to ready. 

The approach in the new LS is to use std LSP didOpen requests for each file found in the workspace. Madhur and I have discussed that option as well.

As you know, that approach may not be the correct one given the perf issues we have been discussing.

packages/salesforcedx-lightning-lsp-common/src/indexer/attributeInfo.ts
```diff
@@ -0,0 +1,40 @@
...
+    location?: Location;
+    detail?: string;
+};
+
+export type MemberType = 'PROPERTY' | 'METHOD';
+
+export type DecoratorType = 'API' | 'TRACK';
+
+// Factory function to replace constructor usage
+export const createAttributeInfo = (
```
mshanemc: this doesn't seem useful. If the Attribute info is already typed, why not just create the object directly instead of this?

packages/salesforcedx-lightning-lsp-common/src/indexer/tagInfo.ts
```diff
@@ -0,0 +1,119 @@
...
+    location?: Location;
+    documentation?: string;
+    name?: string;
+    namespace?: string;
+    properties?: ClassMember[];
+    methods?: ClassMember[];
+};
+
+// Factory function to create TagInfo objects
+export const createTagInfo = (
```
mshanemc: this one does "do a default" but you could do that by making `documentation` required (so people have to pass in an empty string), or (if we control all consumers) having them handle there not being a tag.

otherwise, it's not doing much

packages/salesforcedx-lightning-lsp-common/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+} from './baseContext';
+
+// Re-export from shared
+export { WorkspaceType, isLWC, getSfdxProjectFile, detectWorkspaceHelper, detectWorkspaceType } from './shared';
+
+// Re-export from indexer
+export {
+  TagInfo,
+  createTagInfo,
+  getAttributeInfo,
```
mshanemc: it's worth reviewing whether these are used somewhere.  I think it escapes `knip` because it's reexported and assumed to part of the packagek's API, but if nothing in our repo is using ex: getAttributeInfo then it can be removed.

packages/salesforcedx-lightning-lsp-common/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  getHover,
+  getComponentLibraryLink,
```
mshanemc: getComponentLibraryLink needs to exist, but only within the pkg, don't export it if there are no consumers outside the `common` pkg

packages/salesforcedx-lightning-lsp-common/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export {
+  toResolvedPath,
+  isLWCRootDirectoryCreated,
+  relativePath,
+  pathStartsWith,
```
mshanemc, line 11: no need to export.



In general, please review unnecessary exports from the top-level index files where knip doesn't handle the task for you.  good AI task

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
madhur310: @mshanemc yes you are right - this is a temp approach I implemented to get everything else working. Currently I am sending the entire workspace as initParam which won't work for bigger workspaces and have perceivable start up delays. 

> but transferring contents of workspace during init may impact perceived time to ready

@peternhale agreed on that, the current approach in this PR is temperory & is following:
- Client populates & serializes entires workspaces -> add to init options -> server deserializes and uses it.

I am working through you workspaceLoader poc, that would make it:
- Client initializes server (without any fileSystemProvider) -> Client fires a bunch of onTextDocumentOpen -> Server uses it to populate the fileSystemDataProvider object it needs

packages/salesforcedx-lwc-language-server/src/baseIndexer.ts
mshanemc: is this really an `any` ?

we should have a type for the project.json somewhere (sfdx-core for example) that could imported as a devDep

mshanemc: and if the LSes need the project, maybe it should be provided to them by the client as part of their config (initializationOptions) or whatever the need to know about the project could be provided there (ex: the project isn't important, they only think is needs from it is the pkgDirs so that becomes part of LS init options)

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -0,0 +1,307 @@
...
+    public workspaceType: WorkspaceType = 'UNKNOWN';
+    public readonly tags: Map<string, Tag> = new Map();
+    public readonly fileSystemProvider: IFileSystemProvider;
+
+    constructor(private readonly attributes: ComponentIndexerAttributes) {
+        this.workspaceRoot = getWorkspaceRoot(attributes.workspaceRoot);
+        this.fileSystemProvider = attributes.fileSystemProvider;
+    }
+
+    public async sfdxConfig(root: string): Promise<any> {
```
mshanemc: can you go through the public methods and make sure they're used by something?

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -0,0 +1,307 @@
...
+            return null;
+        }
+    }
+
+    public findTagByURI(uri: string): Tag | null {
+        const uriText = uri.replace('.html', '.js');
+        return Array.from(this.tags.values()).find((tag) => getTagUri(tag) === uriText) ?? null;
+    }
+
+    public async loadTagsFromIndex(): Promise<void> {
```
mshanemc: why public?

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -0,0 +1,307 @@
...
+                sfdxTsConfig.compilerOptions.paths = sfdxTsConfig.compilerOptions.paths ?? {};
+                for (const filePath of filePaths) {
+                    const { dir, name: fileName } = path.parse(filePath);
+                    const componentName = `c/${fileName}`;
+                    const componentFilePath = path.join(dir, fileName);
+                    const tsConfigFilePaths: string[] = (sfdxTsConfig.compilerOptions.paths[componentName] ??= []);
+                    const hasExistingPath = tsConfigFilePaths.includes(componentFilePath);
+                    if (!hasExistingPath) {
+                        tsConfigFilePaths.push(componentFilePath);
+                    }
```
mshanemc: why does it matter if you push them to tsConfigFilePaths?  I don't see it used to do/decide anything.

Maybe there's a bug in this?

mshanemc: This entire for loop might not be doing anything?

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -0,0 +1,307 @@
...
+                    this.fileSystemProvider.updateFileContent(uri, JSON.stringify(sfdxTsConfig, null, 2));
+                }
+            } catch (err) {
+                console.error(err);
+            }
+        }
+    }
+
+    public async getTsConfigPathMapping(): Promise<TsConfigPaths> {
+        const files: TsConfigPaths = {};
```
mshanemc: I also don't understand this one.  how does it not return `{}` if `files` is never modified?

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -0,0 +1,307 @@
...
+        staleTags.forEach((tag) => {
+            if (tag) {
+                this.tags.delete(getTagName(tag));
+            }
+        });
+
+        await this.persistCustomComponents();
+    }
+
+    public async reindex(): Promise<void> {
```
mshanemc: nothing uses this!

packages/salesforcedx-lwc-language-server/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@salesforce/salesforcedx-lwc-language-server",
+  "displayName": "LWC Language Server",
+  "version": "64.16.1",
+  "description": "Language server for Lightning Web Components.",
+  "publisher": "salesforce",
+  "engines": {
+    "vscode": "^1.90.0"
+  },
+  "main": "out/src/indexer.js",
```
mshanemc: I don't see how this points to something.  Where's `indexer` coming from?  should `server` be the main export?

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -0,0 +1,649 @@
...
+            jsconfigExists = true;
+          }
+        } catch {
+          // jsconfig.json doesn't exist
+        }
+
+        if (jsconfigExists) {
+          const existingConfigContent = this.fileSystemProvider.getFileContent(jsconfigPath);
+          if (!existingConfigContent) {
+            throw new Error('Existing config not found');
```
mshanemc: why? line 356 checked that it was there?

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+        workspaceConfig: WorkspaceConfig | null;
+    } {
+        return {
+            fileContents: Object.fromEntries(this.fileContents),
+            directoryListings: Object.fromEntries(this.directoryListings),
+            fileStats: Object.fromEntries(this.fileStats),
+            workspaceConfig: this.workspaceConfig,
+        };
+    }
+}
```
mshanemc, line 209: unused.  If you have to use classes, please check carefully that you don't have dead code

mshanemc, line 209: unused

mshanemc, line 209: unsued

mshanemc, line 209: unused

mshanemc, line 209: unused

mshanemc, line 209: unused

mshanemc, line 209: you can tell `| undefined` is the wrong API if all the consumers have to either `?? []` or check for its existence after calling `getDirectoryListing`.  If there's not any, why not return `[]`.  Then if its empty, they can all iterate `[]` without having to do checks

mshanemc: so then you don't need the class prop, or its type, etc.  This feels like unchallenged AI slop

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export interface FileStat {
+    type: 'file' | 'directory';
+    ctime: number;
+    mtime: number;
+    size: number;
+    exists: boolean;
+}
+
+export interface WorkspaceConfig {
```
mshanemc, line 28: none of this seems to be used for anything!

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    /** JavaScript configuration */
+    jsConfig?: any;
+    /** Force ignore patterns */
+    forceIgnore?: string[];
+}
+
+/**
+ * LSP notification types for file system operations
+ */
+export namespace FileSystemNotifications {
```
mshanemc: let's avoid namespaces.  If you really like the feel of `foo.bar` use an object {}

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    export const FILE_CONTENT_CHANGED = 'lwc/fileContentChanged';
+    export const DIRECTORY_CHANGED = 'lwc/directoryChanged';
+    export const WORKSPACE_CONFIG_CHANGED = 'lwc/workspaceConfigChanged';
+    export const TYPING_FILES_REQUEST = 'lwc/typingFilesRequest';
```
mshanemc: unused !

mshanemc: please check all of these, looks like you only used 1.  Use of namespaces was hiding the deadness.

I think it would be better to just have constant names like `FileSystemNotification_DIRECTORY_CHANGED` and keep it flat OR make 
FileSystemNotification a string union that could be `'lwc/fileContentChanged' | ...`

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    export const GET_FILE_STAT = 'lwc/getFileStat';
+    export const CREATE_TYPING_FILES = 'lwc/createTypingFiles';
+    export const DELETE_TYPING_FILES = 'lwc/deleteTypingFiles';
+    export const UPDATE_COMPONENT_INDEX = 'lwc/updateComponentIndex';
+}
+
+/**
+ * Request/Response types for LSP communication
+ */
+export interface FileContentRequest {
```
mshanemc: unused, please review all of these.

packages/salesforcedx-lightning-lsp-common/src/indexer/tagInfo.ts
```diff
@@ -0,0 +1,119 @@
...
+    documentation: documentation ?? '',
+    name,
+    namespace,
+    properties,
+    methods,
+});
+
+// Utility function to get attribute info by name
+export const getAttributeInfo = (tagInfo: TagInfo, attribute: string): AttributeInfo | null => {
+    const lowerAttribute = attribute.toLowerCase();
```
mshanemc: isn't that just `.find` ?

and then return `| undefined` so we can avoid `null`

mshanemc: oh, wait, this is unused anyway

packages/salesforcedx-aura-language-server/package.json
madhur310: shredded it down

packages/salesforcedx-lightning-lsp-common/src/utils.ts
madhur310: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002OpPFyYAN/view

packages/salesforcedx-aura-language-server/src/tern/doc/manual.html
madhur310: just copied over code from LSP package, but good call out - opportunity to clean up.

packages/salesforcedx-aura-language-server/src/tern/plugin/angular.js
madhur310: tried deleting the unused ones, things seem to work fine. Will keep them deleted.

packages/salesforcedx-aura-language-server/tsconfig.json
madhur310: I went through all of them and leaving comments as to why they are needed or removed them.

packages/salesforcedx-lightning-lsp-common/src/logger.ts
madhur310: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002Oq7oyYAB/view

packages/salesforcedx-lightning-lsp-common/src/index.ts
madhur310: good catch, updated and deleted unused types from fileSystemTypes

packages/salesforcedx-lightning-lsp-common/src/index.ts
madhur310: pathStartsWith is now used by lwcContext, but I will clean up as much as possible.

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
madhur310: I went through some of these, missed this one. Thanks for the catch.

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
madhur310: The loop mutates sfdxTsConfig by reference. will update and add comment.

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
madhur310: yeah its a little confusing because update via references, will update it.

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
madhur310: Thanks for the catch, will clean this up.

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
madhur310: yes making it flat sound best to me, will update.

packages/salesforcedx-lightning-lsp-common/src/indexer/tagInfo.ts
madhur310: yep - deleted

packages/salesforcedx-lightning-lsp-common/src/types/fileSystemTypes.ts
madhur310: updated to remove to unused ones.

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
madhur310: Good point, let me update.

madhur310: I cleaned up fileSystemDataProvider, thanks for the catches on unused code.

packages/salesforcedx-aura-language-server/src/aura-indexer/__tests__/__snapshots__/indexer.spec.ts.snap
```diff
@@ -117,12 +117,9 @@ Map {
     ],
     "documentation": "An app is a special top-level component whose markup is in a .app file. The markup looks similar to HTML and can contain components as well as a set of supported HTML tags. The .app file is a standalone entry point for the app and enables you to define the overall application layout, style sheets, and global JavaScript includes. It starts with the top-level <aura:application> tag, which contains optional system attributes. These system attributes tell the framework how to configure the app.",
     "file": null,
-    "location": undefined,
```
mshanemc: why is the snapshot changing so much?

mshanemc: there may be some things in tern/plugin that are needed...maybe some are important but not all?

packages/salesforcedx-aura-language-server/tsconfig.json
```diff
@@ -2,12 +2,8 @@
   "extends": "../../tsconfig.common.json",
   "compilerOptions": {
     "outDir": "./out",
-    "target": "ES2022",
-    "lib": ["es2022", "dom"],
-    "noImplicitAny": false,
+    "noImplicitAny": false, // needed for ternAura.ts & ternServer.ts to avoid type errors
```
mshanemc: if it's only a few places, it's better to have the rule inline-disabled or file-level disabled so that we don't have `any` stuff introduced elsewhere

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
mshanemc: pro tip: if you have stuff you don't intend to keep around, it'll save us some time if you put that either in pr description or a code comment `//TODO` or both) so I know not to pull on that thread because we both know it needs changing.

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -0,0 +1,721 @@
...
+
+/**
+ * LSP notification types for file system operations
+ */
+const FILE_CONTENT_CHANGED = 'lwc/fileContentChanged';
+
+/**
+ * LSP request types for file system operations
+ */
+const GET_FILE_CONTENT = 'lwc/getFileContent';
```
mshanemc: checking my understanding: these are requests that the LC is going to send to the client?  And the client doesn't have any code that relates to them because you haven't built it yet?

but the implementations look like server endpoints that a client would call?  

I'm not understan this well enough.

packages/salesforcedx-aura-language-server/tsconfig.json
madhur310: makes good sense, will add it to specific files.

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
madhur310: I think it would be cleaner if I add these as needed with next PR i.e. implementing the workspaceLoader. Let me remove them from here, they are not currently used.

madhur310: sorry to answer your question, after discussing with Pete I think it makes sense to not add non-lsp end points. We should be updating the server's fileSystemProvider using the existing end points.

packages/salesforcedx-aura-language-server/src/aura-indexer/__tests__/__snapshots__/indexer.spec.ts.snap
madhur310: yeah I removed the tern/plugins you pointed out and cleaned up best I could. I have tested locally and things still seem to work. IIRC this was a side effect of those clean ups.

packages/salesforcedx-aura-language-server/src/aura-indexer/indexer.ts
```diff
@@ -183,6 +183,7 @@ export default class AuraIndexer implements Indexer {
     if (auraSystem && typeof auraSystem === 'object') {
       for (const tag in auraSystem) {
         if (auraSystem.hasOwnProperty(tag) && typeof tag === 'string') {
+          // @ts-ignore - auraSystem is typed as any
```
mshanemc: pro-tip: use ts-expect-error (so that if the error goes away for some reason, it gets flagged).  the comment is good

madhur310: ty - will make the change

packages/salesforcedx-aura-language-server/src/tern-server/ternAura.ts
```diff
@@ -0,0 +1,486 @@
...
+    }
+  },
+  addSource(source) {
+    this.sources ??= [];
+    this.sources.push(source);
+  }
+});
+
+const getFilename = (filename: string): string => {
+  if (server.options.projectDir.endsWith('/')) {
```
CristiCanizales: does this and following code looking for "/" work for all os's?

packages/salesforcedx-aura-language-server/src/tern/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (C) 2013 by Marijn Haverbeke and others <marijnh@gmail.com>
```
CristiCanizales, line 1: is this right?

packages/salesforcedx-aura-language-server/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@salesforce/salesforcedx-aura-language-server",
+  "displayName": "Aura Language Server",
+  "version": "64.16.1",
```
CristiCanizales, line 4: shouldn't this be same as latest extensions? so 65.9.0?

mshanemc: it's ok either way. To the extension that needs them, they're just `*` and these aren't going to publish to npm anymore.

packages/salesforcedx-aura-language-server/src/tern/LICENSE
mshanemc: yeah, is someone else's code that Aura uses

packages/salesforcedx-aura-language-server/src/tern-server/ternAura.ts
mshanemc: 🤷🏻

PR #6637, opened 2025-11-07
test: e2e playwright on windows W-20153124

mshanemc:
### What does this PR do?
os-specific snapshots because of differeing shortcut keys (that sucks, I might consider doing this in an an assertion-based way instead of aria-snapshots in the next round of e2e tests).

playwright e2e for windows

#<Insert GitHub Issue>, @W-20153124@

proof that it runs (took a few tries, ai is good at this) 
https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/19152212788 (on the test-results for windows you can open the videos from the 5 tests)

### DIFF
--- a/packages/salesforcedx-vscode-org-browser/test/playwright/specs/orgBrowser.customObject.headless.spec.ts-snapshots/desktop-linux/customobject-found
+++ b/packages/salesforcedx-vscode-org-browser/test/playwright/specs/orgBrowser.customObject.headless.spec.ts-snapshots/desktop-linux/customobject-found
@@ -1,4 +1,0 @@
-- treeitem "CustomObject" [level=1]:
-  - toolbar:
-    - button "Refresh Type"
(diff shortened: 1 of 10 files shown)


packages/salesforcedx-vscode-org-browser/test/playwright/specs/orgBrowser.customObject.headless.spec.ts-snapshots/desktop-linux/customobject-found
```diff
@@ -0,0 +1,4 @@
+- treeitem "CustomObject" [level=1]:
```
madhur310: how come we have linux/darwin and windows snapshots - but the e2e run linked https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/19152212788/job/54744983166 doesn't have linux?

madhur310 (approved):

PR #6646, opened 2025-11-12
[W-20175076 ] chore: apex tests discovery api

CristiCanizales:
### What does this PR do?
This pull request introduces a new Apex Test Discovery feature using the Salesforce Tooling REST API, along with supporting types, error handling, and comprehensive test coverage. The main changes include the implementation of the `discoverTests` function, new type definitions for test discovery data structures, and a suite of Jest tests to validate the new functionality.

**Apex Test Discovery Feature Implementation**

* Added `discoverTests` function in `testDiscovery.ts` to fetch Apex test classes and methods from the Tooling REST API, supporting paging, options for filtering, error handling, and telemetry reporting.
* Defined new TypeScript types in `schemas.ts` for test discovery results, including `ToolingTestMethod`, `ToolingTestClass`, `TestDiscoveryResult`, and `ToolingTestsPage`, matching the Tooling REST API response shapes.

**Error Handling and Localization**

* Added a new localized error message key `error_no_connection_found_message` in `i18n.ts` for cases where no Salesforce connection is available.
* Fixed formatting of the `launch_apex_replay_debugger_with_selected_file` message in `i18n.ts` for consistency.

**Testing and Validation**

* Introduced a comprehensive Jest test suite in `testDiscovery.test.ts` to cover successful discovery, error handling, option passing, telemetry emission, and edge cases for the new test discovery logic.

@W-20175076@

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscovery.ts
+++ b/packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscovery.ts
@@ -4,6 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+/* eslint-disable jsdoc/check-indentation */
 
 import type { ToolingTestClass, TestDiscoveryResult, ToolingTestsPage } from './schemas';
 import { getVscodeCoreExtension } from '../coreExtensionUtils';
@@ -17,6 +18,22 @@ import { getTelemetryService } from '../telemetry/telemetry';
 
 const minApiVersion = 65.0;
 
+/**
+ * Query options mirroring the Tooling REST Test Discovery API.
+ *
+ * Notes from the remote API:
+ * - showAllMethods:
+ *   - When true (our default), retrieve all methods; when false (Tooling default), only visible methods.
+ *   - Visibility depends on namespace, package origin, class and method access modifiers and user permissions.
+ *   - Remote default is false; this client defaults to true to surface more tests by default.
+ * - namespacePrefix:
+ *   - Omit (undefined) to retrieve tests in all namespaces (Apex and Flow).
+ *   - Use 'FlowTesting' or 'FlowTesting.<Namespace>' to filter to flow tests (per docs).
+ *   - Use '<Namespace>' to filter to a specific Apex namespace.
+ *   - Supplying '' (empty string) is treated as omitted by this client and won’t be sent.
+ * - pageSize:
+ *   - Number of classes per page; remote default is 1000; maximum 10000.
+ */
 export type DiscoverTestsOptions = {
   showAllMethods?: boolean;
   namespacePrefix?: string;


packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { ToolingTestClass, TestDiscoveryResult, ToolingTestsPage } from './schemas';
+import type { Connection } from '@salesforce/core';
+import { getVscodeCoreExtension } from '../coreExtensionUtils';
+import { getTelemetryService } from '../telemetry/telemetry';
+
+/**
+ * Service that retrieves Apex test classes and methods from the Tooling REST Test Discovery API.
+ * Docs: https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/intro_rest_resources_testing_discovery.htm
+ */
+export class TestDiscoveryService {
```
mshanemc: why make it a class?

Are there scenarios where you'd want to create it but not call discover and just pass in the connection?  [alternatively, is discover the only thing you can do and therefore it should just be a fn]



should the ctor private (like you want consumers to only create it via `create` and not to `new` it directly?)

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
```diff
@@ -0,0 +1,109 @@
...
+      const baseUrl = `${this.connection.instanceUrl}/services/data/v${apiVersion}/tooling/tests`;
+
+      const classes: ToolingTestClass[] = [];
+
+      // paginate until no nextRecordsUrl
+      let nextUrl: string | undefined = baseUrl;
+      while (nextUrl) {
+        let page: ToolingTestsPage | undefined;
+        try {
+          page = await this.connection.request<ToolingTestsPage>({ method: 'GET', url: nextUrl });
```
mshanemc: I think this is short-term fine (using the raw request with a url).

as part of the epic, we should add this endpoint to jsforce so everyone can benefit.

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
```diff
@@ -0,0 +1,109 @@
...
+  public async discover(): Promise<TestDiscoveryResult> {
+    const telemetry = getTelemetryService();
+    const startTime = Date.now();
+    telemetry.sendEventData('apexTestDiscoveryStart', {
+      apiVersion: this.connection.getApiVersion()
+    });
+
+    try {
+      const apiVersion = this.connection.getApiVersion();
+      const baseUrl = `${this.connection.instanceUrl}/services/data/v${apiVersion}/tooling/tests`;
```
mshanemc: the instanceUrl is optional because the conn already knows about it.  request urls can start with `/services` (check my on that by trying it)

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
```diff
@@ -0,0 +1,109 @@
...
+  private connection: Connection;
+
+  constructor(connection: Connection) {
+    this.connection = connection;
+  }
+
+  public static async create(): Promise<TestDiscoveryService | undefined> {
+    try {
+      const core = await getVscodeCoreExtension();
+      const connection = await core.exports.services.WorkspaceContext.getInstance().getConnection();
```
mshanemc: you'll want to force the connection to a high enough API since this service is only available >=65

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
```diff
@@ -0,0 +1,109 @@
...
+      // Gracefully degrade to empty result
+      return { classes: [] };
+    }
+  }
+
+  // no safeRequest helper needed; using typed connection.request<T>() directly above
+}
+
+/**
+ * Convenience function for consumers that don't want to manage service lifecycle.
```
mshanemc: are there consumers that do?!

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
mshanemc: Cristian is the jsforce guru and should review that.  For example, jsforce usually does some nice (paraphrasing from memory) `{queryAll: boolean, limit: number}` kinda stuff for things like queries so consumers don't all have to implement the pagination and can have a nice interface.

and of course the types belong there with that move

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
peternhale: +1
Possibly use effect?

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
peternhale: @mshanemc I have the same need in the OAS PR, getting max api version from the org. There is a prop on authInfo (instanceApiVersion) that seems to fit the bill. Do you recall if that property is guaranteed to be org version?

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
peternhale: In apex-node when fetching test results, I ran into limitations using query all and max record limit which was query all wasn't really all, it was all records up the given limit in one call.
The solution was to deal with jsut like Cristi is doing here. Not sure if the limitations in jsforce are still present, but this pattern will work.

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
CristiCanizales: I thought about making it a class to later include the stuff to create the suites, but I guess those can be functions as well 👍

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
peternhale: @CristiCanizales The "tooling/tests" also takes query [parameters](https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/intro_rest_resources_testing_discovery.htm). I think we will need at least one of them.

I know this targeted for use in Apex extension, but if this is meant to be ageneral purpose API/service, then we might want to expose the query parameters as options to the discover.

I ran the following and results are empty (dreamhouse-lwc)
```
sf api request rest /services/data/v65.0/tooling/tests                      
{
  "apexTestClasses": [],
  "message": null,
  "nextRecordsUrl": null,
  "size": 0,
  "testSetSignature": ""
}
```
and with the following I get some results. Not sure why the first example did not produce any results. Given the docs, I assume the public test classes for dreamhouse-lwc should have been listed.
```
sf api request rest '/services/data/v65.0/tooling/tests?showAllMethods=true'
{
  "apexTestClasses": [
    {
      "id": "01pSG000004dzxO",
      "name": "FileUtilitiesTest",
      "namespacePrefix": "",
      "testMethods": [
        {
          "name": "createFileFailsWhenIncorrectBase64Data"
        },
        {
          "name": "createFileFailsWhenIncorrectFilename"
        },
        {
          "name": "createFileFailsWhenIncorrectRecordId"
        },
        {
          "name": "createFileSucceedsWhenCorrectInput"
        }
      ]
    },
    {
      "id": "01pSG000004dzxQ",
      "name": "GeocodingServiceTest",
      "namespacePrefix": "",
      "testMethods": [
        {
          "name": "blankAddress"
        },
        {
          "name": "errorResponse"
        },
        {
          "name": "successResponse"
        }
      ]
    },
    {
      "id": "01pSG000004dzxU",
      "name": "TestPropertyController",
      "namespacePrefix": "",
      "testMethods": [
        {
          "name": "testGetPagedPropertyList"
        },
        {
          "name": "testGetPicturesNoResults"
        },
        {
          "name": "testGetPicturesWithResults"
        }
      ]
    },
    {
      "id": "01pSG000004dzxV",
      "name": "TestSampleDataController",
      "namespacePrefix": "",
      "testMethods": [
        {
          "name": "importSampleData"
        }
      ]
    }
  ],
  "message": null,
  "nextRecordsUrl": null,
  "size": 4,
  "testSetSignature": "de8bb81d3756ad3fc48b5c9d4981aeb2"
}
```

peternhale (changes requested):

packages/salesforcedx-vscode-apex/src/testDiscovery/testDiscoveryService.ts
mshanemc: > In apex-node when fetching test results, I ran into limitations using query all and max record limit which was query all wasn't really all, it was all records up the given limit in one call. The solution was to deal with jsut like Cristi is doing here. Not sure if the limitations in jsforce are still present, but this pattern will work.

yeah, it'll definitely work. 

Test discovery is NOT tooling sobject records like ApexTestResult etc that you can query the "normal" way.    jsforce doesn't have this endopint at all yet (to my knowledge) so that's why I want it added.  Then you can put the "queryAll-like" option on that API and do the "loop the requests" logic there to make it nicer.

PR #6654, opened 2025-11-17
ci: package from bundle using only vsce W-20248461

mshanemc:
### What does this PR do?
package org plugin the new way

@W-20248461@


contributing/decisions/why-all-all-secrets.md
CristiCanizales, line 1: What's the purpose of this file?

mshanemc: record of why I decided to do what I did, since it's hard to put comments and links in a package.json file like I would have in code.

And then when vsce gets it fixed, we or the AIs will know why that flag was there and that it's safe to remove.

PR #6666, opened 2025-11-20
W-19345728 : Madhur/fix windows ut

madhur310:
@W-19345728@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-aura-language-server/src/context/auraContext.ts
+++ b/packages/salesforcedx-aura-language-server/src/context/auraContext.ts
@@ -95,10 +95,7 @@ export class AuraWorkspaceContext extends BaseWorkspaceContext {
       case 'STANDARD_LWC':
       case 'MONOREPO':
       case 'UNKNOWN': {
-        let depth = 6;
-        if (this.type === 'MONOREPO') {
-          depth += 2;
-        }
+        const depth = this.type === 'MONOREPO' ? 8 : 6;
         const unknownroots = await findNamespaceRoots(this.workspaceRoots[0], this.fileSystemProvider, depth);
         roots.lwc.push(...unknownroots.lwc);
         return roots;
--- a/packages/salesforcedx-lightning-lsp-common/src/__tests__/context.test.ts
+++ b/packages/salesforcedx-lightning-lsp-common/src/__tests__/context.test.ts
@@ -143,7 +143,7 @@ const verifyCoreSettings = (settings: any): void => {
 describe('WorkspaceContext', () => {
   it('WorkspaceContext', async () => {
     let context = new WorkspaceContext(SFDX_WORKSPACE_PATH, sfdxFileSystemProvider);
-    await context.initialize();
+    context.initialize('SFDX');
     expect(context.type).toBe('SFDX');
     expect(context.workspaceRoots[0]).toBeAbsolutePath();
 
@@ -156,7 +156,7 @@ describe('WorkspaceContext', () => {
     ).toBe(3);
 
     context = new WorkspaceContext(STANDARD_WORKSPACE_PATH, standardFileSystemProvider);
-    await context.initialize();
+    context.initialize('STANDARD_LWC');
     expect(context.type).toBe('STANDARD_LWC');
 
     expect(
@@ -166,7 +166,7 @@ describe('WorkspaceContext', () => {
     ).toEqual([]);
 
     context = new WorkspaceContext(CORE_WORKSPACE_PATH, coreFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_ALL');
     expect(context.type).toBe('CORE_ALL');
 
     expect(
@@ -178,7 +178,7 @@ describe('WorkspaceContext', () => {
     ).toBe(3);
 
     context = new WorkspaceContext(CORE_PROJECT_ROOT, coreProjectFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_PARTIAL');
     expect(context.type).toBe('CORE_PARTIAL');
 
     expect(
@@ -188,7 +188,7 @@ describe('WorkspaceContext', () => {
     ).toEqual([normalizePath(path.join(context.workspaceRoots[0], 'modules'))]);
 
     context = new WorkspaceContext(CORE_MULTI_ROOT, coreMultiFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_ALL');
     expect(context.workspaceRoots.length).toBe(2);
 
     const modulesDirs = await getModulesDirs(context.type, context.workspaceRoots, coreMultiFileSystemProvider, () =>
@@ -226,7 +226,7 @@ describe('WorkspaceContext', () => {
 
   it('configureSfdxProject()', async () => {
     const context = new WorkspaceContext(SFDX_WORKSPACE_PATH, sfdxFileSystemProvider);
-    await context.initialize();
+    context.initialize('SFDX');
     const jsconfigPathForceApp = path.resolve(FORCE_APP_ROOT, 'lwc', 'jsconfig.json');
     const jsconfigPathUtilsOrig = path.resolve(UTILS_ROOT, 'lwc', 'jsconfig-orig.json');
     const jsconfigPathUtils = path.resolve(UTILS_ROOT, 'lwc', 'jsconfig.json');
@@ -342,7 +342,7 @@ describe('WorkspaceContext', () => {
 
   it('configureCoreProject()', async () => {
     const context = new WorkspaceContext(CORE_PROJECT_ROOT, coreProjectFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_PARTIAL');
     const jsconfigPath = path.join(CORE_PROJECT_ROOT, 'modules', 'jsconfig.json');
     const typingsPath = path.join(CORE_ALL_ROOT, '.vscode', 'typings', 'lwc');
     const settingsPath = path.join(CORE_PROJECT_ROOT, '.vscode', 'settings.json');
@@ -396,7 +396,7 @@ describe('WorkspaceContext', () => {
 
   it('configureCoreMulti()', async () => {
     const context = new WorkspaceContext(CORE_MULTI_ROOT, coreMultiFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_ALL');
 
     const jsconfigPathGlobal = `${context.workspaceRoots[1]}/modules/jsconfig.json`;
     const tsconfigPathForce = `${context.workspaceRoots[0]}/tsconfig.json`;
@@ -423,7 +423,7 @@ describe('WorkspaceContext', () => {
 
   it('configureCoreAll()', async () => {
     const context = new WorkspaceContext(CORE_ALL_ROOT, coreFileSystemProvider);
-    await context.initialize();
+    context.initialize('CORE_ALL');
     const jsconfigPathGlobal = path.join(CORE_ALL_ROOT, 'ui-global-components', 'modules', 'jsconfig.json');
     const jsconfigPathForce = path.join(CORE_ALL_ROOT, 'ui-force-components', 'modules', 'jsconfig.json');
 
--- a/packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
+++ b/packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
@@ -13,7 +13,7 @@ import { FileSystemDataProvider, IFileSystemProvider } from './providers/fileSys
 import * as jsconfigCoreTemplateJson from './resources/core/jsconfig-core.json';
 import * as settingsCoreTemplateJson from './resources/core/settings-core.json';
 import * as jsconfigSfdxTemplateJson from './resources/sfdx/jsconfig-sfdx.json';
-import { WorkspaceType, detectWorkspaceType, getSfdxProjectFile } from './shared';
+import { WorkspaceType, getSfdxProjectFile } from './shared';
 import * as utils from './utils';
 import { NormalizedPath } from './utils';
 
@@ -230,8 +230,8 @@ export abstract class BaseWorkspaceContext {
   /**
    * Initialize the workspace context asynchronously
    */
-  public async initialize(): Promise<void> {
-    this.type = await detectWorkspaceType(this.workspaceRoots, this.fileSystemProvider);
+  public initialize(workspaceType: WorkspaceType): void {
+    this.type = workspaceType;
     if (this.type === 'SFDX') {
       void this.initSfdxProjectConfigCache();
     }
@@ -508,11 +508,11 @@ export abstract class BaseWorkspaceContext {
 
         updateConfigFile(jsconfigPath, jsconfigContent, this.fileSystemProvider);
       } catch (error) {
-        const errorMessage = error instanceof Error ? error.message : String(error);
-        const errorStack = error instanceof Error ? error.stack : '';
-        console.error(`writeSfdxJsconfig: Error reading/writing jsconfig: ${errorMessage}`);
-        if (errorStack) {
-          console.error(`Stack: ${errorStack}`);
+        console.error(
+          `writeSfdxJsconfig: Error reading/writing jsconfig: ${error instanceof Error ? error.message : String(error)}`
+        );
+        if (error instanceof Error) {
+          console.error(`Stack: ${error.stack}`);
         }
         throw error;
       }
--- a/packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
+++ b/packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
@@ -17,10 +17,6 @@ import {
   NormalizedPath
 } from '@salesforce/salesforcedx-lightning-lsp-common';
 import { snakeCase, camelCase } from 'change-case';
-// minimatch is correctly listed in this package's package.json dependencies,
-// but eslint-plugin-import's no-extraneous-dependencies rule doesn't properly detect
-// dependencies in monorepo setups (it checks the root package.json instead of the package's own)
-// eslint-disable-next-line import/no-extraneous-dependencies
 import { minimatch as minimatchFn } from 'minimatch';
 import * as path from 'node:path';
 
@@ -137,12 +133,10 @@ const findFilesWithGlob = (pattern: string, fileSystemProvider: IFileSystemProvi
       }
     }
 
-    // Check if file matches any of the patterns using minimatch
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    const matchesRelative = patterns.some(p => minimatchFn(relativePath, p, { dot: true }));
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    const matchesAbsolute = patterns.some(p => minimatchFn(fileUri, p, { dot: true }));
-    const matches = matchesRelative || matchesAbsolute;
+    // Check if file matches any of the patterns using minimatc
+    const matches = patterns.some(
+      p => minimatchFn(relativePath, p, { dot: true }) || minimatchFn(fileUri, p, { dot: true })
+    );
 
     if (matches) {
       const fileStat = fileSystemProvider.getFileStat(fileUri);
--- a/packages/salesforcedx-lwc-language-server/src/lwcServer.ts
+++ b/packages/salesforcedx-lwc-language-server/src/lwcServer.ts
@@ -15,7 +15,8 @@ import {
   syncDocumentToTextDocumentsProvider,
   scheduleReinitialization,
   normalizePath,
-  NormalizedPath
+  NormalizedPath,
+  WorkspaceType
 } from '@salesforce/salesforcedx-lightning-lsp-common';
 import { basename, dirname, parse } from 'node:path';
 import {
@@ -134,6 +135,8 @@ export default class Server {
   public lwcDataProvider!: LWCDataProvider;
   public fileSystemProvider: FileSystemDataProvider;
   private textDocumentsFileSystemProvider: FileSystemDataProvider;
+  private workspaceType: WorkspaceType;
+  private isDelayedInitializationComplete = false;
 
   constructor() {
     this.fileSystemProvider = new FileSystemDataProvider();
@@ -147,12 +150,14 @@ export default class Server {
     this.connection.onShutdown(() => this.onShutdown());
     this.connection.onDefinition(params => this.onDefinition(params));
     this.connection.onDidChangeWatchedFiles(params => void this.onDidChangeWatchedFiles(params));
-
+    this.workspaceType = 'UNKNOWN';
     this.documents.listen(this.connection);
   }
 
   public async onInitialize(params: InitializeParams): Promise<InitializeResult> {
     this.workspaceFolders = params.workspaceFolders ?? [];
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+    this.workspaceType = params.initializationOptions?.workspaceType ?? 'UNKNOWN';
     // Normalize workspaceRoots at entry point to ensure all paths are consistent
     // This ensures all downstream code receives normalized paths
     this.workspaceRoots = this.workspaceFolders.map(folder => normalizePath(URI.parse(folder.uri).fsPath));
@@ -162,13 +167,16 @@ export default class Server {
     this.documents.onDidChangeContent(changeEvent => this.onDidChangeContent(changeEvent));
     this.documents.onDidSave(changeEvent => this.onDidSave(changeEvent));
 
+    // Create context but don't initialize yet - wait for files to be loaded via onDidOpen
     this.context = new LWCWorkspaceContext(this.workspaceRoots, this.fileSystemProvider);
 
+    // Create component indexer with fileSystemProvider (will be re-initialized after delayed init)
     this.componentIndexer = new ComponentIndexer({
       workspaceRoot: this.workspaceRoots[0],
       fileSystemProvider: this.fileSystemProvider
     });
 
+    // Create data providers (will be re-initialized after delayed init)
     this.lwcDataProvider = new LWCDataProvider({ indexer: this.componentIndexer });
     this.auraDataProvider = new AuraDataProvider({ indexer: this.componentIndexer });
     await TypingIndexer.create({ workspaceRoot: this.workspaceRoots[0] }, this.fileSystemProvider);
@@ -176,12 +184,6 @@ export default class Server {
       customDataProviders: [this.lwcDataProvider, this.auraDataProvider],
       useDefaultDataProvider: false
     });
-
-    await this.context.initialize();
-    this.context.configureProject();
-
-    // Initialize componentIndexer to get workspace structure
-    await this.componentIndexer.init();
 
     return this.capabilities;
   }
@@ -212,7 +214,7 @@ export default class Server {
     const hasTsEnabled = await this.isTsSupportEnabled();
     if (hasTsEnabled) {
       await this.context.configureProjectForTs();
-      await this.componentIndexer.updateSfdxTsConfigPath();
+      this.componentIndexer.updateSfdxTsConfigPath();
     }
   }
 
@@ -319,6 +321,10 @@ export default class Server {
   }
 
   public async onHover(params: TextDocumentPositionParams): Promise<Hover | null> {
+    if (!params?.textDocument || !params.position) {
+      return null;
+    }
+
     const {
       position,
       textDocument: { uri }
@@ -331,21 +337,19 @@ export default class Server {
 
     const htmlDoc: HTMLDocument = this.languageService.parseHTMLDocument(doc);
 
-    const isLWCTemplate = await this.context.isLWCTemplate(doc);
-    const isAuraMarkup = await this.context.isAuraMarkup(doc);
-
-    if (isLWCTemplate) {
+    if (await this.context.isLWCTemplate(doc)) {
       this.auraDataProvider.activated = false;
       this.lwcDataProvider.activated = true;
-    } else if (isAuraMarkup) {
+      const hover = this.languageService.doHover(doc, position, htmlDoc);
+      return hover;
+    } else if (await this.context.isAuraMarkup(doc)) {
       this.auraDataProvider.activated = true;
       this.lwcDataProvider.activated = false;
+      const hover = this.languageService.doHover(doc, position, htmlDoc);
+      return hover;
     } else {
       return null;
     }
-
-    const hover = this.languageService.doHover(doc, position, htmlDoc);
-    return hover;
   }
 
   /**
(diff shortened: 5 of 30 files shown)


packages/salesforcedx-aura-language-server/src/tern/defs/chai.json
madhur310, line 1: This files are not used in old or new LS.

packages/salesforcedx-lightning-lsp-common/src/__tests__/context.test.ts
```diff
@@ -350,22 +353,6 @@ describe('WorkspaceContext', () => {
     expect(apexContents).not.toContain('declare type');
   });
 
-  /*
```
madhur310, line 353: not used

packages/salesforcedx-lightning-lsp-common/src/__tests__/documentSync.test.ts
```diff
@@ -22,8 +24,9 @@ describe('syncDocumentToTextDocumentsProvider', () => {
   it('should normalize file:// URI to fsPath and sync document', async () => {
     const uri = 'file:///workspace/src/file.js';
     const content = 'console.log("test");';
+    const normalizedPath = normalizePath(URI.parse(uri).fsPath);
```
madhur310, line 27: Most normalization now happens on entry i.e. server.ts files. We still have some sporadic normalization in the code to handle windows/capitalization issues.

packages/salesforcedx-lightning-lsp-common/src/logger.ts
```diff
@@ -4,33 +4,127 @@
...
+                ? 'debug'
+                : 'log';
+
+      console[consoleMethod](...args);
+      return;
     }
-};
+
+    const formattedMessage = formatMessage(...args);
+    void Logger.connection.sendNotification(LogMessageNotification.type, {
```
madhur310, line 80: easy use logger to help implement windows/logMessage in LSP.

.vscode/tasks.json
```diff
@@ -62,6 +62,34 @@
...
+        "isDefault": false
+      },
+      "dependsOn": ["Compile"],
+      "command": "npm",
+      "type": "shell",
+      "presentation": {
+        "focus": false,
+        "panel": "dedicated"
+      },
+      "args": ["run", "vscode:bundle:extension"],
```
mshanemc: I know it's hard with the pace of change, but FYI this is about to be `vscode:bundle` in the wireit branch that Cristi's reviewing for me.

madhur310: Thanks for reminder, yeah I was expecting this to change. I will wait for your change to go in and update it.

packages/salesforcedx-aura-language-server/src/tern/defs/chai.json
mshanemc: woohoo!

packages/salesforcedx-lightning-lsp-common/src/__tests__/context.test.ts
mshanemc: love it!

packages/salesforcedx-lightning-lsp-common/src/__tests__/documentSync.test.ts
```diff
@@ -6,6 +6,8 @@
  */
 import { syncDocumentToTextDocumentsProvider } from '../documentSync';
 import { FileSystemDataProvider } from '../providers/fileSystemDataProvider';
+import { normalizePath } from '../utils';
```
mshanemc, line 10: my local linter was flagging import order

packages/salesforcedx-lightning-lsp-common/src/__tests__/documentSync.test.ts
mshanemc: makes sense, I think that makes the test name weird.  

There are a couple nice ways to use the compiler to avoid mistakes around the (now required) path normalization.

One would be a "branded type" (string & brand), have normalizePath return that type, and have all your methods/fn require that.  Effect has it built in if you want that https://effect.website/docs/code-style/branded-types/ (also has a good explanation of the problem and the solution).  But it's pretty easy to do in vanilla TS anyway.

packages/salesforcedx-lightning-lsp-common/src/__tests__/testUtils.ts
```diff
@@ -79,40 +83,37 @@ const readFileContent = (filePath: string): string => {
...
+        const content = readFileContent(fullPath);
+        if (content) {
+          files[relativePath] = content;
+        }
+      }
+    }
+  } catch {
+    // Ignore errors when reading directories
+  }
+};
```
mshanemc, line 108: I haven't thought deeply about it, but what do you think is missing from readdir with `recursive` option (with file types), then filter out the dirs and map that to get content and build a giant object out of?

fine if you don't like it, since it's test code, but I'd like to get us in the habit of using all the current node features that people used to have to do work like this for

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
```diff
@@ -62,46 +55,108 @@ export class FileSystemDataProvider implements IFileSystemProvider {
...
+                name: fileName,
+                type: stat?.type ?? 'file',
+                uri: fileUri
+              });
+            }
+          }
+        }
+      }
+
+      if (entries.length > 0) {
```
mshanemc: why not just return entries on line 107, so that it handles the empty and non-empty case without extra conditionals?

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
```diff
@@ -62,46 +55,108 @@ export class FileSystemDataProvider implements IFileSystemProvider {
...
+
+    // Infer directory existence from files: if any file path starts with this directory path,
+    // the directory must exist. Ensure we check with a trailing slash to avoid partial matches.
+    const dirPathWithSlash = normalizedUri.endsWith('/') ? normalizedUri : `${normalizedUri}/`;
+    for (const fileUri of this.fileStats.keys()) {
+      if (fileUri.startsWith(dirPathWithSlash)) {
+        return true;
+      }
+    }
+    return false;
```
mshanemc: Array.some

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -119,24 +119,27 @@ export const getModulesDirs = (
   fileSystemProvider: IFileSystemProvider,
   getSfdxProjectConfig: () => SfdxProjectConfig
 ): string[] => {
```
mshanemc: and certain things like this probably return the branded type too

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -209,7 +216,11 @@ export abstract class BaseWorkspaceContext {
    * @return BaseWorkspaceContext representing the workspace with workspaceRoots
    */
   constructor(workspaceRoots: string[] | string, fileSystemProvider: FileSystemDataProvider) {
-    this.workspaceRoots = typeof workspaceRoots === 'string' ? [path.resolve(workspaceRoots)] : workspaceRoots;
+    // Normalize workspaceRoots to ensure consistent path format (especially Windows drive letter casing)
+    this.workspaceRoots =
```
mshanemc: and then you could use branded type on props like workspaceRoots, etc

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -245,11 +257,17 @@ export abstract class BaseWorkspaceContext {
...
+    // Normalize file path to ensure consistent format (especially Windows drive letter casing and path separators)
+    const normalizedFile = utils.normalizePath(file);
+    return (await this.findNamespaceRootsUsingTypeCache()).lwc.some(root => utils.pathStartsWith(normalizedFile, root));
   }
 
   public async isFileInsideAuraRoots(file: string): Promise<boolean> {
-    return (await this.findNamespaceRootsUsingTypeCache()).aura.some(root => utils.pathStartsWith(file, root));
+    // Normalize file path to ensure consistent format (especially Windows drive letter casing and path separators)
+    const normalizedFile = utils.normalizePath(file);
+    return (await this.findNamespaceRootsUsingTypeCache()).aura.some(root =>
```
mshanemc, line 268: good use of some to return a boolean

packages/salesforcedx-lightning-lsp-common/src/documentSync.ts
```diff
@@ -34,16 +34,17 @@ const ensureDirectoryTracked = async (
   });
 
   // Get or create directory listing
-  const entries = provider.getDirectoryListing(dirUri) ?? [];
+  const entries = provider.getDirectoryListing(dirPath) ?? [];
```
mshanemc: if we know getDirectoryListing always returns an array, do we really need a `??`

packages/salesforcedx-lightning-lsp-common/src/documentSync.ts
```diff
@@ -52,15 +53,14 @@ const ensureDirectoryTracked = async (
...
  */
 const addFileToDirectoryListing = async (
-  fileUri: string,
+  filePath: string,
   provider: FileSystemDataProvider,
   workspaceRoots: string[]
 ): Promise<void> => {
-  // fileUri is already normalized to fsPath format, but handle both formats for safety
-  const filePath = fileUri.startsWith('file://') ? URI.parse(fileUri).fsPath : fileUri;
+  // filePath is already normalized to fsPath format
```
mshanemc: a branded type would make that comment explicit

packages/salesforcedx-lightning-lsp-common/src/namespaceUtils.ts
```diff
@@ -48,22 +49,26 @@ const traverse = async (
...
   const dirs = [];
   if (entries) {
     for (const entry of entries) {
       if (entry.type === 'directory') {
-        dirs.push(path.join(candidate, entry.name));
+        // Normalize the joined path to ensure consistent format (especially Windows drive letter casing)
+        dirs.push(normalizePath(path.join(normalizedCandidate, entry.name)));
       }
     }
   }
```
mshanemc: avoid mutation
const dirs = entries.filter(e => e.type === 'directory').map(normalize etc)

packages/salesforcedx-lightning-lsp-common/src/namespaceUtils.ts
```diff
@@ -85,9 +90,11 @@ export const findNamespaceRoots = async (
   };
 
   try {
-    const stat = fileSystemProvider.getFileStat(root);
+    // Normalize root path before calling getFileStat to ensure path format consistency
```
mshanemc: I'm a bit confused on what this is doing and could use some comments.
(ex: what are namespace roots? what do they do?  Something in Core only, right?)

and does it ever find an Aura root?

and then this might be better as a recursive readdir and filter on `lwc` ?  I don't know because I don't understand the domain well enough to know what it's doing and why

packages/salesforcedx-lightning-lsp-common/src/utils.ts
```diff
@@ -47,7 +47,19 @@ export const isLWCRootDirectoryCreated = (context: BaseWorkspaceContext, changes
   return false;
 };
 
-export const unixify = (filePath: string): string => filePath.replace(/\\/g, '/');
+const unixify = (filePath: string): string => filePath.replace(/\\/g, '/');
+
```
mshanemc: prefer string.replaceAll over global regex.  There's a lint rule for that, I think, in develop these days.

packages/salesforcedx-lwc-language-server/src/context/lwcContext.ts
```diff
@@ -18,7 +18,8 @@ import {
   getExtension,
   toResolvedPath,
   pathStartsWith,
-  unixify
+  normalizePath,
+  Logger
 } from '@salesforce/salesforcedx-lightning-lsp-common';
 import baseTsConfigJsonImport from '@salesforce/salesforcedx-lightning-lsp-common/resources/sfdx/tsconfig-sfdx.base.json';
```
mshanemc: you should export these (use pjson.exports) from the other package and import them the legit way.

packages/salesforcedx-lwc-language-server/src/context/lwcContext.ts
```diff
@@ -192,18 +196,22 @@ export class LWCWorkspaceContext extends BaseWorkspaceContext {
...
+    const file = normalizePath(toResolvedPath(document.uri));
+
     for (const ws of this.workspaceRoots) {
-      if (pathStartsWith(file, ws)) {
-        return this.isFileInsideModulesRoots(file);
+      const startsWith = pathStartsWith(file, ws);
+      if (startsWith) {
+        const isInside = await this.isFileInsideModulesRoots(file);
+        return isInside;
       }
```
mshanemc, line 207: I think I liked the version without all the extra temporary consts

packages/salesforcedx-lwc-language-server/src/baseIndexer.ts
```diff
@@ -23,15 +23,31 @@ interface SfdxProjectConfig {
 }
 
 // Utility function to resolve workspace root
-export const getWorkspaceRoot = (workspaceRoot: string): string => path.resolve(workspaceRoot);
+// Normalizes the path to ensure consistency (path.resolve() may reintroduce backslashes on Windows)
+// On Windows, path.isAbsolute() returns false for Windows-style paths on non-Windows platforms
+// So we check for Windows drive letter pattern explicitly
+export const getWorkspaceRoot = (workspaceRoot: string): string => {
```
mshanemc: would return the branded type here, too

packages/salesforcedx-lwc-language-server/src/baseIndexer.ts
```diff
@@ -46,10 +62,7 @@ const getSfdxConfig = (root: string, fileSystemProvider: IFileSystemProvider): S
 };
 
 /** Get SFDX package directories pattern from sfdx-project.json */
-export const getSfdxPackageDirsPattern = async (
-  workspaceRoot: string,
-  fileSystemProvider: IFileSystemProvider
-): Promise<string> => {
+export const getSfdxPackageDirsPattern = (workspaceRoot: string, fileSystemProvider: IFileSystemProvider): string => {
```
mshanemc: I'd make that use the normalized path for workspaceRoot so that getSfdxConfig doesn't have to worry about it.  Or, push the line out closer to the edge

mshanemc: the branded type would help enforce where the line is

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -62,31 +180,35 @@ export default class ComponentIndexer {
...
-        files = sync(sfdxSource, {
-          stats: true
-        });
-        return files.filter((item: Entry): boolean => {
+        // workspaceRoot is already normalized by getWorkspaceRoot()
+        const packageDirsPattern = await this.getSfdxPackageDirsPattern();
+        // Pattern matches: {packageDir}/**/*/lwc/**/*.js
+        // The **/* before lwc requires at least one directory level (e.g., main/default/lwc or meta/lwc)
+        const sfdxPattern = `${packageDirsPattern}/**/*/lwc/**/*.js`;
+        files = await findFilesWithGlob(sfdxPattern, this.fileSystemProvider, this.workspaceRoot);
```
mshanemc: it's not async

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -62,31 +180,35 @@ export default class ComponentIndexer {
...
-        const sfdxSource = normalize(
-          `${this.workspaceRoot}/${await this.getSfdxPackageDirsPattern()}/**/*/lwc/**/*.js`
-        );
-        files = sync(sfdxSource, {
-          stats: true
-        });
-        return files.filter((item: Entry): boolean => {
+        // workspaceRoot is already normalized by getWorkspaceRoot()
+        const packageDirsPattern = await this.getSfdxPackageDirsPattern();
+        // Pattern matches: {packageDir}/**/*/lwc/**/*.js
```
mshanemc: hmm, why wouldn't you just want the packageDirs, and then use those to match stuff, intead of all the glob/regex complexity?

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -62,31 +180,35 @@ export default class ComponentIndexer {
...
+          const dirEndsWithName = data.dir.endsWith(data.name);
+          return dirEndsWithName;
         });
+        return filteredFilesDefault;
     }
   }
 
   public getCustomData(): Tag[] {
-    return Array.from(this.tags.values());
+    const tags = Array.from(this.tags.values());
```
mshanemc: but why?

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -226,20 +231,27 @@ export default class Server {
 
     const htmlDoc: HTMLDocument = this.languageService.parseHTMLDocument(doc);
 
-    if (await this.context.isLWCTemplate(doc)) {
+    const isLWCTemplate = await this.context.isLWCTemplate(doc);
```
mshanemc: I don't understand why you'd do all three of these async calculations if you're only going to use 1 of them and they're mutually exclusive?  Like what was wrong with having them in the `if` ?

mshanemc: the idea is that if the first one (isLWCTemplate) is true, then you never had to wait on the rest of them.  And you're not using it for anything other than the conditionals.

** more inline, less consts sitting around waiting for the outer scope to close **

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -433,35 +466,41 @@ export default class Server {
     const uri = document.uri;
     const content = document.getText();
 
-    // Sync to TextDocuments FileSystemDataProvider
-    await syncDocumentToTextDocumentsProvider(uri, content, this.textDocumentsFileSystemProvider, this.workspaceRoots);
+    // Normalize URI to fsPath before syncing (entry point for path normalization)
+    const normalizedPath = normalizePath(URI.parse(uri).fsPath);
+    void syncDocumentToTextDocumentsProvider(
```
mshanemc: await, in case of error.

void should be last resort, and if you're going to do it, you having to be absolutely certain it can't throw or have handling for that

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -433,35 +466,41 @@ export default class Server {
...
 
-    await this.connection.sendNotification(ShowMessageNotification.type, {
+    void this.connection.sendNotification(ShowMessageNotification.type, {
       type: MessageType.Info,
       message: 'LWC Language Server shutting down'
     });
   }
 
-  public async onExit(): Promise<void> {
+  public onExit(): void {
```
mshanemc: were these not able to be async for some reason?

mshanemc: I think that either needs a comment or some explanation in the PR body.  I can't tell if this was a conscious decision for a reason I don't understand yet, or AI slop, or something else.

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -630,5 +669,11 @@ export default class Server {
       customDataProviders: [this.lwcDataProvider, this.auraDataProvider],
       useDefaultDataProvider: false
     });
+
+    // send notification that re-initialization is complete with new FileSystemProvider
+    void this.connection.sendNotification(ShowMessageNotification.type, {
+      type: MessageType.Info,
+      message: 'LWC Language Server is ready'
```
mshanemc: cool!

packages/salesforcedx-lwc-language-server/typings/glob-to-regexp.d.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+declare module 'glob-to-regexp' {
```
mshanemc: a new added library that doesn't have TS support is a bad sign.  I looked after seeing its lack of TS, and it's last published 7 years ago, archived, not got a valid license (probably not ok to use)

I think this library needs to go and use somethign else.

packages/salesforcedx-vscode-automation-tests/test/specs/lwcLsp.e2e.ts
```diff
@@ -36,22 +38,33 @@ describe('LWC LSP', () => {
     },
     isOrgRequired: false,
     testSuiteSuffixName: 'LwcLsp',
-    extensionConfigs: defaultExtensionConfigs
+    extensionConfigs: [
+      ...defaultExtensionConfigs,
+      {
+        extensionId: 'salesforcedx-vscode-lwc',
+        shouldVerifyActivation: false, // We don't activate until we have an LWC
```
mshanemc, line 45: yay!

packages/salesforcedx-vscode-automation-tests/test/specs/runLwcTests.e2e.ts
```diff
@@ -70,6 +71,9 @@ describe('Run LWC Tests', () => {
 
     // Install Jest unit testing tools for LWC
     await installJestUTToolsForLwc(testSetup.projectFolderPath);
+
+    // wait for server initialization to complete
+    await verifyNotificationWithRetry(/LWC Language Server is ready/, Duration.seconds(10));
```
mshanemc, line 82: this is good use of the ui to make testing easier!

packages/salesforcedx-vscode-lightning/src/index.ts
```diff
@@ -178,8 +197,21 @@ const createAuraResourcesProvider = async (
   const provider = new FileSystemDataProvider();
 
   // Load Aura framework resources from extension
+  // In packaged extension: dist/resources/aura (copied during bundling)
```
mshanemc, line 200: fine for now, but I hope eventually this goes away when everything consistently runs from dist.

mshanemc (changes requested):
that regex library has to go.  Probably worth thinking through what that code's goal is, and if there's a simpler way to get there.

I think the branded type idea solves a LOT of things for anyone who has to touch these packages, esp us.  Can be a separate WI but I wouldn't wait too long on it becuase there's probably SOME bug in here where something should have been normalized but isn't, and probably some unnecessary normalization, too.

Other than that, it's just code style nits suggestions.

packages/salesforcedx-lwc-language-server/typings/glob-to-regexp.d.ts
madhur310: I saw 31M weekly downloads and went for it. Let me try to find an alternative.

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
madhur310: ah did this for some logging, missed reverting back. Thanks for the catch.

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
madhur310: debug logs, missed reverting it.

packages/salesforcedx-lightning-lsp-common/src/__tests__/documentSync.test.ts
madhur310: it just flagged for me too, odd it didn;t do it before. Fixed it.

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
madhur310: 107 is outside of the if, to do that I would need to define the variable outside of if. But I see your point that this condition is unnecessary. I will refactor

packages/salesforcedx-lightning-lsp-common/src/namespaceUtils.ts
madhur310: you are right it never populates aura, great catch. Let me revisit this.

packages/salesforcedx-aura-language-server/src/context/auraContext.ts
```diff
@@ -64,30 +62,32 @@ export class AuraWorkspaceContext extends BaseWorkspaceContext {
...
           }
-          const auraDir = path.join(ws, 'components');
-          if (this.fileSystemProvider.fileExists(auraDir)) {
-            const subroots = await findNamespaceRoots(path.join(ws, 'components'), this.fileSystemProvider, 2);
-            roots.aura.push(...subroots.lwc);
+          const auraDir = normalizePath(path.join(ws, 'components'));
+          if (this.fileSystemProvider.directoryExists(auraDir)) {
+            // The components directory itself is the Aura namespace root
+            // (findNamespaceRoots only detects LWC, not Aura)
+            roots.aura.push(auraDir);
```
madhur310, line 90: This makes sense right?

packages/salesforcedx-lightning-lsp-common/src/__tests__/testUtils.ts
madhur310: I just kept it as as since it is test code.

.vscode/tasks.json
mshanemc: it's in

madhur310: does this https://github.com/forcedotcom/salesforcedx-vscode/pull/6666/files#diff-7d76d7533653c23b753fc7ce638cf64bdb5e419927d276af836d3a03fdf1745aR77 not look right?

packages/salesforcedx-lightning-lsp-common/src/__tests__/workspaceContext.ts
```diff
@@ -11,80 +11,95 @@ import { BaseWorkspaceContext } from '../baseContext';
...
-            case 'SFDX':
-                // For SFDX workspaces, check for both lwc and aura directories
-                for (const root of this.workspaceRoots) {
-                    const forceAppPath = path.join(root, 'force-app', 'main', 'default');
-                    const utilsPath = path.join(root, 'utils', 'meta');
-                    const registeredEmptyPath = path.join(root, 'registered-empty-folder', 'meta');
+  /**
+   * @returns string list of all lwc and aura namespace roots
+   */
+  protected async findNamespaceRootsUsingType(): Promise<{ lwc: string[]; aura: string[] }> {
```
mshanemc, line 17: I'm trying to trace through what this is for?

It's only used for testing?  But it has to exist to satisfy the "base" interface and because other parts of the Base context (ex: ctor) are going to call it, and we're trying to test that?

And this avoids us having to test contextType being correct in either LWC or Aura consumer?  But since both of those have their own implementation of `findNamespaceRootsUsingType` do these tests prove anything useful?

it smells weird, like giant mock.

If there really are useful things to test on the base class (ex: that we need `context.type` or workspaceRoots set) maybe it's simpler to make those public properties so tests can set them rather than having to have all this.

I haven't dug in enough because OOP hurts my brain so I'll defer to your judgment.

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -119,25 +121,28 @@ export const getModulesDirs = (
...
+): NormalizedPath[] => {
+  // Normalize workspaceRoots at the start to ensure consistent path format
+  // This ensures all path operations use normalized paths
+  const normalizedWorkspaceRoots = workspaceRoots.map(root => utils.normalizePath(root));
+  const modulesDirs: NormalizedPath[] = [];
   switch (workspaceType) {
     case 'SFDX':
       const { packageDirectories } = getSfdxProjectConfig();
       for (const pkg of packageDirectories) {
         // Check both new SFDX structure (main/default) and old structure (meta)
```
mshanemc, line 133: `meta` have never heard of this.

if it doesn't show up in public github, maybe we never needed it?

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -119,25 +121,28 @@ export const getModulesDirs = (
...
-        const newPkgDir = path.join(workspaceRoots[0], pkg.path, 'main', 'default');
-        const oldPkgDir = path.join(workspaceRoots[0], pkg.path, 'meta');
+        const newPkgDir = path.join(normalizedWorkspaceRoots[0], pkg.path, 'main', 'default');
```
mshanemc, line 134: that's kind of our default path, but there's never been a requirement to use `main/default` and some people don't.

Are we really just looking for "is there a dir named `lwc` in each packageDir" ?  If so, this is a clunky way to find it (doing a recursive readdir of them or using a fs.glob pattern) would be smarter.

mshanemc: and then maybe we don't care about meta or main/default

packages/salesforcedx-lwc-language-server/src/__tests__/componentIndexer.test.ts
```diff
@@ -55,15 +54,17 @@ describe('ComponentIndexer', () => {
...
           'utils/meta/lwc/todo_utils/todo_utils.js'
-        ].map(item => path.join(componentIndexer.workspaceRoot, item));
+          // Note: todo_util and todo_utils are now found with the updated pattern **/lwc/**/*.js
+        ].map(item => normalizePath(path.join(componentIndexer.workspaceRoot, item)));
 
         const componentEntries = await componentIndexer.getComponentEntries();
-        const paths = componentEntries.map(entry => path.resolve(entry.path)).sort();
+        const paths = componentEntries.map(entry => normalizePath(path.resolve(entry.path))).sort();
 
         expect(paths).toEqual(expectedComponents.sort());
```
mshanemc, line 69: see linter note, here and elsewhere

packages/salesforcedx-lwc-language-server/src/componentIndexer.ts
```diff
@@ -12,37 +12,156 @@ import {
   SfdxTsConfig,
   TsConfigPaths,
   IFileSystemProvider,
-  unixify
+  normalizePath,
+  Logger,
+  NormalizedPath
 } from '@salesforce/salesforcedx-lightning-lsp-common';
 import { snakeCase, camelCase } from 'change-case';
```
mshanemc, line 19: this is not true

mshanemc: I removed it without any issues

PR #6668, opened 2025-11-24
ci: eslint rule to enforce i18n for vscode window messages W-20311793

mshanemc:
### What does this PR do?
flag text in vscode window messages that should use localization
@W-20311793@

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6664,7 +6664,7 @@
       "dev": true,
       "license": "BSD-3-Clause"
     },
-    "node_modules/@salesforce/eslint-plugin-local-rules": {
+    "node_modules/@salesforce/eslint-plugin-vscode-extensions": {
       "resolved": "packages/eslint-local-rules",
       "link": true
     },
@@ -38061,7 +38061,7 @@
       "license": "MIT"
     },
     "packages/eslint-local-rules": {
-      "name": "@salesforce/eslint-plugin-local-rules",
+      "name": "@salesforce/eslint-plugin-vscode-extensions",
       "version": "65.6.0",
       "license": "BSD-3-Clause",
       "devDependencies": {
--- a/packages/eslint-local-rules/README.md
+++ b/packages/eslint-local-rules/README.md
@@ -1,4 +1,4 @@
-# @salesforce/eslint-plugin-local-rules
+# @salesforce/eslint-plugin-vscode-extensions
 
 Custom ESLint rules for Salesforce VSCode extensions.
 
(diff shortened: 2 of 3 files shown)


eslint-local-rules/no-vscode-message-literals.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * ESLint rule to enforce that vscode.window.show*Message calls
+ * use nls.localize() or variables, not string literals
+ */
+
+module.exports = {
```
madhur310: is it possible to add ut for this? just to make sure others don't break it down the line.

packages/salesforcedx-vscode-org-browser/src/messages/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { LocalizationService, type MessageArgs } from '@salesforce/vscode-i18n';
+import { messages as enMessages, isValidMessageKey, type MessageKey } from './i18n';
+
+const INSTANCE_NAME = 'salesforcedx-vscode-org-browser';
```
madhur310: is this intentionally different than `packages/salesforcedx-vscode-org-browser/src/constants.ts` `EXTENSION_NAME`?

packages/salesforcedx-vscode-org-browser/src/commands/retrieveMetadata.ts
```diff
@@ -96,8 +97,7 @@ const confirmOverwrite = (localComponents: ComponentSet, target: MetadataMember)
   Effect.promise(async () => {
     if (localComponents.size === 0) return true;
     const answer = await vscode.window.showWarningMessage(
```
madhur310, line 99: should this be `void` too? like other show..messages?

madhur310 (approved):

packages/salesforcedx-vscode-org-browser/src/commands/retrieveMetadata.ts
daphne-sfdc: I think it's not void because there's a choice between Yes and No here?

mshanemc: no, those are only void because they can't await.  await (or fork from Effect) is definitely better, just in case some error does happen and you get unhandled stuff

eslint-local-rules/no-vscode-message-literals.js
mshanemc: yeah, but it'll need to become a real package.  It's fine since I was going to do that anyway

packages/salesforcedx-vscode-org-browser/src/messages/index.ts
mshanemc: no, good idea!

PR #6675, opened 2025-11-26
[W-20328559][W-20328751] feat: scrape Metadata API Developer Guide metadata types for metadata XML hover documentation using Playwright + GHA workflow to do this once per week

daphne-sfdc:
### What does this PR do?
1. Switch the mechanism of gathering metadata types for hover documentation on metadata XML files from using Swarm to web scraping via Playwright.  This allows a more comprehensive list of the metadata types, as the Metadata API Developer Guide is the source of truth for metadata types and contains info for all subtypes.
2. Create a GHA workflow that scrapes the Metadata API Developer Guide and updates the XSD that integrates with the RedHat XML extension weekly.  We want to do this weekly rather than once per major release because the teams that own the metadata types can update their page on the developer guide anytime, and we want to provide the most up-to-date information to our customers in our hover documentation.

@W-20328559@, @W-20328751@


scripts/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
...
+    cleaned = 'field_' + cleaned;
+  }
+
+  // Remove multiple consecutive underscores
+  cleaned = cleaned.replace(/_+/g, '_');
+
+  // Remove trailing underscores
+  cleaned = cleaned.replace(/_+$/, '');
+
+  return cleaned || 'unknown_field';
```
mshanemc: I think we have the `??`  as a lint rule, but it probably doesn't apply to scripts.  A good habit to get into everywhere, though.

scripts/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
...
+  const lowerFieldType = fieldType.toLowerCase().trim();
+
+  // Basic types
+  if (lowerFieldType.endsWith('[]')) {
+    return 'xsd:anyType'; // Arrays - treated as anyType
+  } else if (lowerFieldType.includes('string') || lowerFieldType.includes('text')) {
+    return 'xsd:string';
+  } else if (lowerFieldType.includes('boolean')) {
+    return 'xsd:boolean';
+  } else if (lowerFieldType.includes('int') || lowerFieldType.includes('integer')) {
```
mshanemc: I don't understand why you'd want both the conditionals here, since the first is a subset of the second?

I guess the other question, about this entire fn, is why use `includes` and not just `===` to match them more exactly?  Maybe it'll make more sense when I look at the json.

scripts/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * Escape XML special characters.
+ */
+const escapeXml = (text: string): string => {
+  if (!text) {
+    return '';
+  }
+
+  let escaped = String(text);
```
mshanemc: isn't it already a string?

scripts/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  // Basic XML escaping
+  escaped = escaped.replace(/&/g, '&amp;');
```
mshanemc: here, you might consider 2 simplifications
1. just replace a string using ex: escaped.replaceAll('<', '&lt;') instead of getting regex involved
2. chain the replaces like

```
return escaped.replaceAll('foo', 'bar').replaceAll('next', 'thing') etc. instead of having to use `let`

File: scripts/jsonToXsdConverter.ts [hunk 5]
```diff
@@ -0,0 +1,291 @@
...
+    '  </xsd:choice>',
+    ' </xsd:complexType>'
+  );
+
+  // Process each metadata type (sorted alphabetically)
+  const sortedTypes = Object.keys(metadataTypes).sort();
+
+  for (const typeName of sortedTypes) {
+    // Skip the base Metadata type to avoid duplicate definition
+    if (typeName === 'Metadata') {
```
mshanemc: simpler: filter on 148, before the sort, instead of having to deal with it inside a loop?

scripts/xsd/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  // Close schema
+  xsdLines.push('</xsd:schema>');
+
+  // Write to file
+  fs.writeFileSync(outputFilePath, xsdLines.join('\n') + '\n', 'utf-8');
+
+  return numTypes;
+};
```
mshanemc, line 273: this is a very long fn; if it were runtime code instead of a script it'd probably be a good idea to break it up.  I'm indifferent because of where it is.

scripts/scrape-single-page.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (results.length === 0) {
+      console.log(`\n❌ No metadata extracted`);
+      if (isVisible) {
+        console.log(`\nBrowser will stay open for 30 seconds for inspection...`);
+        await page.waitForTimeout(30000);
+      }
+    } else {
+      // Save results
+      const output: any = {};
```
mshanemc: cleaner: Object.fromEntries from `result.map`

scripts/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,430 @@
...
+
+    return [];
+  }
+
+  return results;
+};
+
+/** Discovers metadata types from the Salesforce Metadata API JSON documentation endpoint */
+const discoverMetadataTypes = async (page: Page): Promise<Array<{ name: string; url: string }>> => {
+  const JSON_DOC_URL = 'https://developer.salesforce.com/docs/get_document/atlas.en-us.api_meta.meta';
```
mshanemc: in general: it's better to put consts like this, that never change, outside of fns, since they're not really scoped to the fn.

scripts/xsd/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,430 @@
...
+      '--disable-blink-features=AutomationControlled',
+      '--disable-dev-shm-usage',
+      '--disable-web-security',
+      '--disable-features=IsolateOrigins,site-per-process',
+      '--no-first-run',
+      '--no-default-browser-check'
+    ]
+  });
+
+  const context = await browser.newContext({
```
mshanemc: I think I've seen these launch and context args before, let's deduplicate them

mshanemc: there's a tool called `jscpd` that helps find this stuff before it gets to code review

mshanemc: or, make it part of your Cursor habits to have it review open changes for duplicate code.

package.json
```diff
@@ -136,6 +138,8 @@
     "link-lsp": "yarn link @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common && lerna exec yarn link @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lightning && lerna exec yarn link @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lwc",
     "unlink-lsp": "yarn unlink @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common && lerna exec yarn unlink @salesforce/aura-language-server @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lightning && lerna exec yarn unlink @salesforce/lwc-language-server @salesforce/lightning-lsp-common --scope salesforcedx-vscode-lwc",
     "report:installs": "ts-node scripts/reportInstalls.ts",
+    "scrape:all:pages": "ts-node scripts/scrape-all-metadata-pages.ts",
+    "scrape:single:page": "ts-node scripts/scrape-single-page.ts",
```
mshanemc: checking my understanding: this was only used for dev/test purposes?  It's not used by the automation?

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -57,9 +57,37 @@ export class MetadataDocumentationService {
...
+      const layerField = layerTypeDoc.fields.find(f => f.name === layerName);
+      if (!layerField) {
+        break;
+      }
+
+      // Get the type of this layer field and strip array notation if present
+      let layerType = layerField.type;
+      if (layerType.endsWith('[]')) {
+        layerType = layerType.slice(0, -2);
+      }
```
mshanemc:
```suggestion
      const layerType = layerField.type.endsWith('[]') ? layerField.type.slice(0, -2) : layerField.type
```

avoid let

packages/salesforcedx-vscode-core/src/metadataSupport/metadataDocumentationService.ts
```diff
@@ -57,9 +57,37 @@ export class MetadataDocumentationService {
...
+
+      // Update currentType to this resolved type for the next iteration
+      currentType = layerType;
```
mshanemc: or just assign currentType to what I wrong above instead of having so many vars to juggle?

scripts/jsonToXsdConverter.ts
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/env node
+/**
+ * Convert metadata_types_map_scraped.json to XSD file compatible with the RedHat XML extension.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+type FieldInfo = {
```
mshanemc: this is about the 4th time I've seen some of these.

suggestion: new folder `scripts/xsd` then put all your stuff in there.  and then export the types from this nice utils module so the other files don't have to repeat them.

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,850 @@
...
+      if (!pageLevelDescription && mainHeading) {
+        const allParagraphs = Array.from(document.querySelectorAll('p, dd'));
+        for (const p of allParagraphs) {
+          // Only consider paragraphs that come after the heading in DOM order
+          if (mainHeading.compareDocumentPosition(p) & Node.DOCUMENT_POSITION_FOLLOWING) {
+            const text = p.textContent?.trim() || '';
+            const textLower = text.toLowerCase();
+
+            const isSubstantial = text.length > 50;
+            const isNotNavigation =
```
mshanemc: this is repeated

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,850 @@
...
+    // Helper to extract type name from array notation (e.g., "AssignmentRule[]" -> "AssignmentRule")
+    const extractArrayTypeName = (fieldType: string): string | null => {
+      // Remove zero-width characters that might be in the type name
+      const cleanType = fieldType.replace(/[\u200B-\u200D\uFEFF]/g, '');
+      const match = cleanType.match(/^([A-Z][a-zA-Z0-9_]+)\[\]$/);
+      return match ? match[1] : null;
+    };
+
+    // Helper to extract complex type names (non-primitive types without [])
+    const extractComplexTypeName = (fieldType: string): string | null => {
```
mshanemc: this (`extractComplexTypeName`) is a reasonably-sized fn.  does it need to be inside that other fn? maybe it needs another param (cleanType?)

That's the type of thing that would help make `extractMetadataFromPage` smaller.

I think `extractArrayTypeName` (maybe there's others) should be it own module-scoped fn.

There shouldn't really be units of code (fns, methods) that go beyond 100 lines.  There's usually **something** inside there that becomes its own fn.

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,850 @@
...
+    } catch {
+      // No cookie banner
+    }
+
+    // Strategy 3: Wait for network to settle
+    await page.waitForLoadState('networkidle', { timeout: 15000 }).catch(() => {});
+
+    // Strategy 4: Wait for JavaScript to fully render
+    console.log(`${indent}⏳ Waiting 10 seconds for content to fully load...`);
+    await page.waitForTimeout(10000);
```
mshanemc: I think all this `waitForTimeout` is why these take so long.  Is there some dom element you can wait for to know that the page is done loading all its scripts, etc?

I'd review this, maybe with the single-page script, and just have the AI repeat it and with the goal of seeing what's really necessary.  Otherwise, it just keeps piling more 💩 on, and never asks, "was all of that really necessary? can I back some of it out and still be ok?"
 
ex: what happens if you don't accept the cookies?  does the page still load?  Is there, in fact, lazy-loaded content where scrolling is necessary?  are some of the "fallbacks" never actually used?

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,850 @@
...
+
+export type MetadataType = {
+  fields: MetadataField[];
+  short_description: string;
+  url: string;
+};
+
+/**
+ * Improved page loading with multiple strategies
+ * Returns the frame containing the actual content (might be an iframe)
```
mshanemc: suggestion: I'd have Cursor read https://playwright.dev/docs/api/class-framelocator and then review this code to see if there's better ways to handle the iframe stuff.

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,850 @@
...
+    }
+
+    console.log(`${indent}✓ Using frame: ${contentFrame.url() || 'main'}`);
+
+    // Wait for tables in the iframe
+    console.log(`${indent}⏳ Waiting for tables to stabilize...`);
+    let lastTableCount = 0;
+    let stableIterations = 0;
+
+    for (let i = 0; i < 15; i++) {
```
mshanemc: I'd have AI read https://playwright.dev/docs/locators#locate-in-shadow-dom and then review this file, too.  I'm wondering whether this is necessary?

mshanemc: fyi, getByRole('table') and filter() might be useful
https://playwright.dev/docs/locators#filter-by-text

scripts/xsd/scrapeUtils.ts
mshanemc: it's probably worth just asking for this file to be cleaned up and "create a plan to reduce code duplication"

scripts/jsonToXsdConverter.ts
daphne-sfdc: Here's an example:
```
      {
        "Description": "Required. The visual specification for the visualization.",
        "Field Name": "visualSpecification",
        "Field Type": "base64Binary"
      },
```

vs.

```
      {
        "Description": "The Apex class definition. Base 64-encoded binary data. Before making an API call, client applications must encode the binary attachment data as base64. Upon receiving a response, client applications must decode the base64 data to binary. This conversion is handled for you by a SOAP client. This field is inherited from the MetadataWithContent component.",
        "Field Name": "content",
        "Field Type": "base64"
      },
```

Even thought one is `base64Binary` and the other is `base64`, they should both be mapped onto this category:
```
else if (lowerFieldType.includes('base64')) {
    return 'xsd:base64Binary';
  }
```

So we need to use `includes` instead of `===`.

package.json
daphne-sfdc: I do use `scrape:all:pages` in the new GHA workflow **generateXsdForMetadataTypes.yml**.

Screenshot 2025-12-01 at 6 17 29 PM

scripts/xsd/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,740 @@
...
+      'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
+    viewport: { width: 1920, height: 1080 },
+    locale: 'en-US',
+    timezoneId: 'America/Los_Angeles',
+    extraHTTPHeaders: {
+      'Accept-Language': 'en-US,en;q=0.9',
+      Accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8',
+      'Accept-Encoding': 'gzip, deflate, br'
+    }
+  });
```
mshanemc: this all exists in 2 places

scripts/xsd/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,740 @@
...
+
+  const discoveryContext = await createBrowserContext(discoveryBrowser);
+  const discoveryPage = await discoveryContext.newPage();
+
+  // Comprehensive anti-detection for discovery page
+  await discoveryPage.addInitScript(() => {
+    // Hide webdriver
+    Object.defineProperty(navigator, 'webdriver', { get: () => undefined });
+
+    // Mock chrome object
```
mshanemc: this is duplicated.

scripts/xsd/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,740 @@
...
+  console.log('   • Fast browsers will process more tasks than slow ones\n');
+
+  // Step 1: Launch a single browser for discovery
+  console.log('🔍 Launching discovery browser...');
+  const discoveryBrowser = await chromium.launch({
+    headless: !isVisible,
+    args: BROWSER_LAUNCH_ARGS
+  });
+
+  const discoveryContext = await createBrowserContext(discoveryBrowser);
```
mshanemc, line 501: let's try not to use `any`.  discoveryBrowser is nicely typed coming from chromium.launch, so why have fn `createBrowserContext` take `any`?  it should match the type `Browser`

scripts/xsd/scrape-all-metadata-pages.ts
```diff
@@ -0,0 +1,740 @@
...
+    return metadataTypes.length > 0 ? metadataTypes : [];
+  } catch (error) {
+    console.error('   ❌ Discovery failed:', error);
+    console.log('   📋 Using fallback list...\n');
+    return [];
+  }
+};
+
+/** Context pool for dynamic work distribution */
+class ContextPool {
```
mshanemc, line 245: please don't change this, just fyi if you ever need something like this again
https://effect.website/docs/concurrency/semaphore/#withpermits
https://effect.website/docs/resource-management/scope/#acquirerelease

scripts/xsd/scrapeUtils.ts
```diff
@@ -0,0 +1,1264 @@
...
+      return countTablesIncludingShadowDOM(document);
+    };
+  }
+};
+
+/**
+ * Extract metadata from a loaded page (or iframe)
+ * Returns an array because some pages have multiple tables representing different types
+ */
+export const extractMetadataFromPage = async (
```
mshanemc, line 175: that is quite the large function.  Any natural places to split up or reuse into something easier to fit in a human or AI context window?

mshanemc: for example, move helper functions outside this function (dont' nest them).  Can even put them in their own files inside `/helpers` or whatever.

I know it's just scripts, but I need you to be repulsed by code that looks like this and want to fix it when you look at it.

scripts/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "outDir": "./dist",
     "noEmit": true,
     "resolveJsonModule": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "lib": ["es2021", "dom"]
```
mshanemc: why not use use 2023, or inherit from up higher?  Was there something special that needed overriding?

package.json
```diff
@@ -134,6 +139,11 @@
     "vscode:publish": "npm run vscode:publish -ws --if-present",
     "watch": "npm run compile --watch",
     "report:installs": "ts-node scripts/reportInstalls.ts",
+    "scrape:all:pages": "ts-node scripts/xsd/scrape-all-metadata-pages.ts",
+    "scrape:single:page": "ts-node scripts/xsd/scrape-single-page.ts",
```
mshanemc, line 143: is this used/useful outside of local test/debug?  I'd be ok running the script without a pjson script to wrap it

mshanemc (approved):
approved with non-blocking comments.

package.json
daphne-sfdc: `scrape:all:pages` is used in **generateXsdForMetadataTypes.yml**.
`scrape:single:page` is run only manually.
Screenshot 2026-01-02 at 10 22 30 AM

scripts/xsd/scrapeUtils.ts
daphne-sfdc: I actually did try to move the helper functions out from `extractMetadataFromPage()` but was unable to do so.  It's because the helper functions must stay within the `evaluate()` block on Line 164 in order to have access to the DOM.

scripts/tsconfig.json
daphne-sfdc: Yes, `replaceAll()` needs a newer version.
```
scripts/xsd/scrapeUtils.ts:1259:25 - error TS2550: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.

1259   cleanType = cleanType.replaceAll(/[\u200B-\u200D\uFEFF]/g, '');
```

I've updated **scripts/tsconfig.json** to use ES2023 instead of ES2021.

PR #6685, opened 2025-12-01
ci: wireit W-20391187

mshanemc:
### What does this PR do?
replace lerna/nx with wireit
- gha caching (ci runs that are on same infra/os with same code/lockfile don't have to be repeated!)
- express cross-package deps more granularly
- rely on explicit dep mgmt instead
- more file/output cacheability (opens up things like "don't re-run e2e that have already passed and nothing has changed")
- faster

no intended functionality changes 
@W-20391187@

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,6 +183,7 @@
         "./packages/salesforcedx-vscode-apex-debugger:lint",
         "./packages/salesforcedx-vscode-apex-replay-debugger:lint",
         "./packages/salesforcedx-vscode-apex-oas:lint",
+        "./packages/salesforcedx-vscode-apex-testing:lint",
         "./packages/salesforcedx-vscode-lwc:lint",
         "./packages/salesforcedx-vscode-lightning:lint",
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-services/src/observability/spansWeb.ts
```diff
@@ -16,7 +16,7 @@ export const WebSdkLayer = WebSdk.layer(() => ({
   resource: {
     serviceName: 'salesforcedx-vscode-services',
     //manually bump this to cause rebuilds/bust cache
-    serviceVersion: '2025-08-15T20:49:30.000Z',
```
mshanemc, line 19: unimportant, used to force re-builds when I was QAing this

CristiCanizales (commented):
Looks good overall! 
I'm just curious why we're downgrading to versions from 65.7.0 65.6.0 and o11y-reporter from 1.6.0 to 1.4.0

PR #6712, opened 2025-12-11
feat: logging-auth-errors-for-CB W-20509584

mshanemc:
### What does this PR do?
CB gets weird auth errors and needs more info about the failure.  Our message "invalid/expired token' is true but not helpful as to why (actually expired? IP restrictions? connectedApp policies?)

So we're logging the actual error.

The problem is that we didn't have an OutputChannel (AFAICT) within utils-vscode, so I stuck this the org extensions's channel.  Open to feedback on which channel to use since we don't really know who the caller is?

The uglier option would be to allow consumers of WCU pass in a channel?

We want this out for next week's release.

@W-20509584@


CristiCanizales (approved):
I guess it's ok to use org management channel, since it's auth/connection related, while we figure out the issue. Should we send telemetry too with a log name to track it?

PR #6727, opened 2025-12-17
feat: move debugger commands to debugger W-20163556

mshanemc:
### What does this PR do?
2 commands to old apex debugger
- isv bootstrap
- stop debugger

@W-20163556@

### DIFF
--- a/packages/salesforcedx-vscode-apex-debugger/package.json
+++ b/packages/salesforcedx-vscode-apex-debugger/package.json
@@ -39,7 +39,6 @@
     "salesforcedx-vscode-core": "*"
   },
   "extensionDependencies": [
-    "salesforce.salesforcedx-vscode-apex",
     "salesforce.salesforcedx-vscode-core"
   ],
   "scripts": {
@@ -84,7 +83,8 @@
       "files": [
         "esbuild.config.mjs",
         "../../scripts/bundling/node.mjs",
-        "out/**"
+        "out/**",
+        "package.json"
       ],
       "output": [
         "dist"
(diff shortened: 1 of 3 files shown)


packages/salesforcedx-vscode-apex-debugger/src/commands/isvdebugging/bootstrapCmd.ts
```diff
@@ -51,7 +50,83 @@ export const ISVDEBUGGER = 'isvdebuggermdapitmp';
...
+/** Prompts user for input and returns trimmed value */
+const getFormattedString = async (prompt: string, value?: string) => {
+  const input = await vscode.window.showInputBox({
+    prompt,
+    value
+  });
+  return input ? input.trim() : input;
+};
+
+class SelectProjectName implements ParametersGatherer<ProjectName> {
```
mshanemc, line 72: some of this duplicates the project creation stuff in core.  I expect that to change and would prefer not to have to touch this ext

packages/salesforcedx-vscode-apex-debugger/package.json
```diff
@@ -29,14 +29,16 @@
...
   },
   "devDependencies": {
     "esbuild": "0.25.10",
-    "salesforcedx-vscode-apex": "*"
+    "salesforcedx-vscode-apex": "*",
+    "salesforcedx-vscode-core": "*"
   },
   "extensionDependencies": [
-    "salesforce.salesforcedx-vscode-apex",
     "salesforce.salesforcedx-vscode-core"
```
mshanemc, line 42: remove apex is intentional.  the ISV debug command expects to create a new project and doesn't need to be run from inside a project.

that causes problems because extensionDependencies tries to activate apex (and without a project it fails).

Given that people install from the pack, and if they have one,they probably have the other, I chose the lesser evil.  It still activates apex for the breakpoint stuff.

PR #6774, opened 2026-01-15
[W-20175153] feat: new test controller by default and apex testing ext on web

CristiCanizales:
### What does this PR do?
This pull request refactors the `salesforcedx-vscode-apex-testing` package to improve modularity, enable browser compatibility, and remove legacy dependencies. Key changes include replacing dependencies on `salesforcedx-utils-vscode` and core extension APIs with direct service calls, adding support for web (browser) builds, updating configuration and activation events, and cleaning up unused settings and views.

**Build and packaging improvements:**
* Added browser build support in `esbuild.config.mjs` and updated `package.json` to include browser entry points and extension kind for web compatibility. [[1]](diffhunk://#diff-80894ec0a9dcb81b6350423f9f584f3f7c15cb1096eff0d4b1cdcd2a4ced97adR9-R12) [[2]](diffhunk://#diff-80894ec0a9dcb81b6350423f9f584f3f7c15cb1096eff0d4b1cdcd2a4ced97adR21-R31) [[3]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL138-R147)

**Dependency and API refactoring:**
* Removed dependencies on `@salesforce/salesforcedx-utils-vscode` and related legacy APIs, replacing them with direct service calls and new helper modules. Updated `tsconfig.json` references and cleaned up imports throughout the codebase. [[1]](diffhunk://#diff-32fba7b75e5b93edf230d34afe99f95f194abb6c0be858390f8b249eb51cebcfL18-R25) [[2]](diffhunk://#diff-a94837b46297bfccc8fb84b2baba5bc81b46b940f3f56291bb22d8b6c6d16ee2L20-R26) [[3]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL28-R29) [[4]](diffhunk://#diff-3dc0d7fecd96b81c658415ed346e72c8f64158898b4cf45af4b76491eee817e1L7-R23) [[5]](diffhunk://#diff-9d9dd63caba597ae453c9242ce43d702915ce6d58e8e799c5f22eb2086793425R9-L39) [[6]](diffhunk://#diff-8462d9dcdb1998a03548bc5fa79661c6c7d5745f8d7c254a669a4a1e67df5286L12-R21) [[7]](diffhunk://#diff-7b88d8644e7c65de9f22a5681a30f875e354f9faa0041800f150379b6969cd81R9-R14) [[8]](diffhunk://#diff-a2cde531d7259951e2c9e0bbb43f18577904f96a89de665bbc47fc6c968b60d0L10-R30)

**Configuration and activation updates:**
* Updated `package.json` to add new activation events (`onStartupFinished`), browser entry points, and extension kind. Removed unused extension dependencies and configuration options, such as test explorer and discovery source settings. [[1]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL40-R43) [[2]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL138-R147) [[3]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL153-R160) [[4]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL364-L382)

**Feature additions and settings:**
* Added a new configuration option to control retrieval of test code coverage results, with corresponding localization strings. [[1]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcR410-R414) [[2]](diffhunk://#diff-23413ccdba8b30ad27dc021c13bd84ee8d29bbffc9149df2c678b262f862942bL36-R39)

**Code maintenance and cleanup:**
* Removed unused view contributions and configuration properties, streamlined output channel service, and improved file system handling for browser compatibility. [[1]](diffhunk://#diff-7ad1fdbda23343a27171d13f26f69e175758672ed5465b00920e4c10b0d9d5dcL153-R160) [[2]](diffhunk://#diff-3dc0d7fecd96b81c658415ed346e72c8f64158898b4cf45af4b76491eee817e1L7-R23) [[3]](diffhunk://#diff-a2cde531d7259951e2c9e0bbb43f18577904f96a89de665bbc47fc6c968b60d0L172-R187)

These changes collectively modernize the extension, improve maintainability, and prepare it for running in both desktop and browser environments.

@W-20175153@

### DIFF
--- a/packages/salesforcedx-vscode-apex-testing/src/index.ts
+++ b/packages/salesforcedx-vscode-apex-testing/src/index.ts
@@ -133,8 +133,8 @@ const setupTestResultsFileWatcher = (
     yield* Effect.forkDaemon(
       Stream.fromPubSub(fileWatcherService.pubsub).pipe(
         Stream.filter(isTestResultJsonFile),
-        Stream.runForEach(event => {
-          const filePath = event.uri.fsPath ?? event.uri.path;
+        Stream.map(event => event.uri.fsPath ?? event.uri.path),
+        Stream.runForEach(filePath => {
           // Extract the apex test results directory from the file path (handle both / and \ separators)
           const lastSepIndex = Math.max(filePath.lastIndexOf('/'), filePath.lastIndexOf('\\'));
           const apexDirPath = filePath.substring(0, lastSepIndex);
--- a/packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
+++ b/packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
@@ -154,21 +154,11 @@ export const getMethodLocationsFromSymbols = async (
  */
 const convertApiToApexTestMethods = async (classes: ToolingTestClass[]): Promise<ApexTestMethod[]> => {
   // Extract class names from discovery results to drive file lookup
-  const classNames = classes
+  const nonNamespaceClassesWithTestMethods = classes
     .filter(cls => cls.testMethods?.length > 0)
-    .filter(cls => !hasNamespace(cls))
-    .map(cls => cls.name);
-
-  const classNameToUri = await buildClassToUriIndex(classNames);
-  const apiByClassName = new Map<string, ToolingTestClass[]>();
-  for (const cls of classes) {
-    if (cls.testMethods?.length === 0) continue;
-    // Only consider tests in the local (default) namespace; workspace index maps local files only
-    if (hasNamespace(cls)) continue;
-    const list = apiByClassName.get(cls.name) ?? [];
-    list.push(cls);
-    apiByClassName.set(cls.name, list);
-  }
+    .filter(cls => !hasNamespace(cls));
+  const classNameToUri = await buildClassToUriIndex(nonNamespaceClassesWithTestMethods.map(cls => cls.name));
+  const apiByClassName = Map.groupBy(nonNamespaceClassesWithTestMethods, cls => cls.name);
 
   const tests: ApexTestMethod[] = [];
   for (const [className, uri] of classNameToUri) {
--- a/.cursor/commands/analyze-e2e.md
+++ b/.cursor/commands/analyze-e2e.md
@@ -1,0 +1,113 @@
+s# Analyze E2E Tests
+
+Monitor running e2e playwright tests for current branch, download artifacts on failure, provide analysis tools.
+
+## Prerequisites
+
+- `gh` CLI installed and authenticated (`gh auth status` succeeds)
+- Git repo with branch checked out
+
+## Workflow
+
+1. **Get Current Branch**
+   - `git branch --show-current`
+   - Detached HEAD/no branch → error, stop
+
+2. **Find E2E Workflows**
+   - `gh run list -b <branch> --limit 50 --json databaseId,status,conclusion,workflowName,createdAt,headBranch`
+   - Filter `workflowName` containing "(Playwright)"
+   - Examples: "Metadata E2E (Playwright)", "Services E2E (Playwright)", "OrgBrowser E2E (Playwright)"
+   - Multiple found: prioritize `in_progress`/`queued`, then most recent completed (`createdAt` desc)
+   - Multiple running → monitor all; only completed → use most recent
+
+3. **Monitor Workflow Status**
+   - **IMPORTANT**: Monitor until all complete. Never ask to continue—monitor automatically.
+   - `in_progress`/`queued`: `gh run watch <run-id>` or poll `gh run view <run-id> --json status,conclusion` until `status: "completed"`
+   - Multiple running → monitor all (sequentially if needed)
+   - Already completed → skip monitoring
+
+4. **Handle Results**
+   - **NEVER RETURN BEFORE CI COMPLETES (SUCCESS OR FAILURE)**
+
+   **Success:**
+   - Report: "✓ E2E tests passed for workflow `<workflow-name>` on branch `<branch-name>`"
+   - URL: `gh run view <run-id> --web`
+   - Multiple workflows → report each
+
+   **Failure/Cancellation:**
+   - Create: `.e2e-artifacts/<branch-name>/<run-id>-<workflow-name>/` (sanitize: spaces/special → dashes)
+   - Download: `gh run download <run-id> -D .e2e-artifacts/<branch-name>/<run-id>-<workflow-name>`
+   - Extract/unzip if needed
+   - Report failure with artifact location
+   - Multiple failed → download each
+
+5. **Offer Analysis** (if artifacts downloaded)
+   - Search HTML reports: `playwright-report/index.html` or `**/playwright-report/index.html`
+   - Search test results: `test-results/` directory
+   - Reference: `.cursor/rules/coding-playwright-tests.mdc`, `.cursor/rules/iterating-playwright-tests.mdc`
+   - Offer: open HTML report, show test results, open workflow (`gh run view <run-id> --web`)
+
+## Artifact Storage Structure
+
+```
+.e2e-artifacts/
+  <branch-name>/
+    <run-id>-<workflow-name>/
+      playwright-report/
+        index.html
+      playwright-test-results/
+      (other artifacts)
+```
+
+Organized by branch and run ID.
+
+## GitHub CLI Commands
+
+- `gh run list -b <branch> --limit 50 --json databaseId,status,conclusion,workflowName,createdAt,headBranch` - List runs
+- `gh run watch <run-id>` - Monitor until completion
+- `gh run download <run-id> -D <directory>` - Download artifacts
+- `gh run view <run-id> --web` - Open in browser
+
+## Workflow Detection
+
+Filter `gh run list` JSON for `workflowName` containing "(Playwright)". Examples: "Metadata E2E (Playwright)" (`metadataE2E.yml`), "Services E2E (Playwright)" (`servicesE2E.yml`), "OrgBrowser E2E (Playwright)" (`orgBrowserE2E.yml`).
+
+## Error Handling
+
+- **Missing `gh` CLI**: `which gh` or `gh --version`, install: https://cli.github.com/
+- **Unauthenticated**: `gh auth status`, `gh auth login` if needed, stop on failure
+- **No branch checked out**: Detect detached HEAD/no git repo, error, stop
+- **No workflows found**: `git ls-remote --heads origin <branch>`, report branch may not have triggered workflows, suggest pushing
+- **Artifact download failures**: `gh run view <run-id> --json artifacts`, report "No artifacts available" or suggest manual download
+- **Missing branch context**: `git branch --show-current` fails → check git repo, error
+
+## Examples
+
+**Running workflow → failure:**
+
+1. Branch: `feature/my-branch`
+2. Find "Metadata E2E (Playwright)" `in_progress`
+3. Monitor until `failure`
+4. Download to `.e2e-artifacts/feature/my-branch/<run-id>-Metadata-E2E-Playwright/`
+5. Report failure with artifact location, offer HTML report
+
+**Already completed successfully:**
+
+1. Find latest "Services E2E (Playwright)" `success`
+2. Skip monitoring
+3. Report success, provide workflow URL
+
+**No workflows found:**
+
+1. `gh run list` empty
+2. `git ls-remote --heads origin <branch>`
+3. Report branch may not be pushed or doesn't trigger workflows
+
+## Implementation Notes
+
+- `execSync` or `spawn` for git/gh commands
+- Parse JSON from `gh run list` to filter workflows
+- `mkdir -p` before downloading
+- Extract zip if compressed
+- `find` or `glob` for HTML reports
+- Clear, actionable error messages
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ junit-custom.xml
 junit-custom-unitTests.xml
 junit-custom-integrationTests.xml
 junit-custom-vscodeIntegrationTests.xml
-
+.e2e-artifacts/
 # nyc test coverage
 .nyc_output
 
--- a/packages/salesforcedx-vscode-apex-testing/package.json
+++ b/packages/salesforcedx-vscode-apex-testing/package.json
@@ -38,7 +38,7 @@
   ],
   "scripts": {
     "vscode:bundle": "wireit",
-    "run:web": "npm run vscode:bundle && npx vscode-test-web --browserType=chromium --browserOption=--disable-web-security --browserOption=--remote-debugging-port=9222 --extensionDevelopmentPath . --extensionPath ../salesforcedx-vscode-services --open-devtools --port 3001 --quality=stable",
+    "run:web": "npm run vscode:bundle && npx vscode-test-web --browserType=chromium --browserOption=--disable-web-security --browserOption=--remote-debugging-port=9222 --extensionDevelopmentPath . --extensionPath ../salesforcedx-vscode-services --extensionPath ../salesforcedx-vscode-org-browser --open-devtools --port 3001 --quality=stable",
     "vscode:package": "wireit",
     "vscode:sha256": "node ../../scripts/generate-sha256.js >> ../../SHA256",
     "vscode:publish": "node ../../scripts/publish-vsix.js",
--- a/packages/salesforcedx-vscode-apex-testing/src/views/testController.ts
+++ b/packages/salesforcedx-vscode-apex-testing/src/views/testController.ts
@@ -44,12 +44,10 @@ const TEST_RESULT_JSON_FILE = 'test-result.json';
 
 export class ApexTestController {
   private controller: vscode.TestController;
-  private testItems: Map<string, vscode.TestItem> = new Map();
   private suiteItems: Map<string, vscode.TestItem> = new Map();
   private classItems: Map<string, vscode.TestItem> = new Map();
   private methodItems: Map<string, vscode.TestItem> = new Map();
   private suiteParentItem: vscode.TestItem | undefined;
-  private resultFileWatcher: vscode.FileSystemWatcher | undefined;
   private lastProcessedResultFile: string | null = null;
   private connection: Connection | undefined;
   private testService: TestService | undefined;
@@ -171,7 +169,6 @@ export class ApexTestController {
 
   private clearTestItems(): void {
     this.controller.items.replace([]);
-    this.testItems.clear();
     this.suiteItems.clear();
     this.classItems.clear();
     this.methodItems.clear();
@@ -261,7 +258,6 @@ export class ApexTestController {
       }
 
       this.controller.items.add(classItem);
-      this.testItems.set(createClassId(fullClassName), classItem);
     }
   }
 
@@ -292,7 +288,6 @@ export class ApexTestController {
         }
         this.suiteItems.set(suite.TestSuiteName, suiteItem);
         this.suiteParentItem.children.add(suiteItem);
-        this.testItems.set(suiteId, suiteItem);
       }
 
       // Add the parent item first so it appears at the top
@@ -725,7 +720,6 @@ export class ApexTestController {
   }
 
   public dispose(): void {
-    this.resultFileWatcher?.dispose();
     this.controller.dispose();
   }
 }
--- a/packages/salesforcedx-vscode-apex-testing/src/views/testOutlineProvider.ts
+++ b/packages/salesforcedx-vscode-apex-testing/src/views/testOutlineProvider.ts
@@ -62,38 +62,31 @@ export class ApexTestOutlineProvider implements vscode.TreeDataProvider<TestNode
   public getChildren(element: TestNode): TestNode[] {
     if (element) {
       return element.children;
-    } else {
-      if (this.rootNode && this.rootNode.children.length > 0) {
-        return this.rootNode.children;
-      } else {
-        const message = NO_TESTS_MESSAGE;
-        const description = NO_TESTS_DESCRIPTION;
-        const emptyArray = new Array<ApexTestNode>();
-        const testToDisplay = new ApexTestNode(message, null);
-        testToDisplay.description = description;
-        emptyArray.push(testToDisplay);
-        return emptyArray;
-      }
     }
+    if (this.rootNode && this.rootNode.children.length > 0) {
+      return this.rootNode.children;
+    }
+    const message = NO_TESTS_MESSAGE;
+    const description = NO_TESTS_DESCRIPTION;
+    const testToDisplay = new ApexTestNode(message, null);
+    testToDisplay.description = description;
+    return [testToDisplay];
   }
 
   public getTreeItem(element: TestNode): vscode.TreeItem {
     if (element) {
       return element;
-    } else {
-      this.getAllApexTests();
-      const message = NO_TESTS_MESSAGE;
-      const description = NO_TESTS_DESCRIPTION;
-      const emptyArray = new Array<ApexTestNode>();
-      const testToDisplay = new ApexTestNode(message, null);
-      testToDisplay.description = description;
-      emptyArray.push(testToDisplay);
-      if (!(this.rootNode && this.rootNode.children.length > 0)) {
-        this.rootNode = new ApexTestNode(message, null);
-        this.rootNode.children.push(testToDisplay);
-      }
-      return this.rootNode;
     }
+    this.getAllApexTests();
+    const message = NO_TESTS_MESSAGE;
+    const description = NO_TESTS_DESCRIPTION;
+    const testToDisplay = new ApexTestNode(message, null);
+    testToDisplay.description = description;
+    if (!(this.rootNode && this.rootNode.children.length > 0)) {
+      this.rootNode = new ApexTestNode(message, null);
+      this.rootNode.children.push(testToDisplay);
+    }
+    return this.rootNode;
   }
 
   public async refresh(): Promise<void> {
--- a/packages/salesforcedx-vscode-org-browser/test/playwright/web/headlessServer.ts
+++ b/packages/salesforcedx-vscode-org-browser/test/playwright/web/headlessServer.ts
@@ -27,6 +27,7 @@ const startHeadlessServer = async (): Promise<void> => {
       port: 3001,
       printServerLog: true,
       verbose: true,
+      quality: 'stable',
       extensionDevelopmentPath,
       extensionPaths: [servicesExtensionPath],
       browserOptions: [


packages/salesforcedx-vscode-apex-testing/src/channels/index.ts
```diff
@@ -4,9 +4,20 @@
...
 import { nls } from '../messages';
 
 export const OUTPUT_CHANNEL = vscode.window.createOutputChannel(nls.localize('channel_name'));
-export const channelService = new ChannelService(OUTPUT_CHANNEL);
+
+/**
+ * Simple channel service that wraps vscode OutputChannel.
+ * Replaces ChannelService from @salesforce/salesforcedx-utils-vscode
+ */
+export const channelService = {
```
mshanemc, line 45: there's a per-ext channel service inside the services exports.

It's uglier but has the advantage of being able being able to be used when you call things from services (ex: getConnection) from your extension and have the output show up in your extensions's channel.

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,178 +20,129 @@ import {
...
 
   // Telemetry
   await telemetryService.initializeService(context);
   const activationTracker = new ActivationTracker(context, telemetryService);
 
-  const useTestExplorerMode = useTestExplorer();
-  const testOutlineProvider = getTestOutlineProvider();
-  // Only create test controller if Test Explorer mode is enabled
-  const testController = useTestExplorerMode ? getTestController() : undefined;
+  // Set context keys for command visibility
```
mshanemc: this is recently added as part of ProjectService 
https://github.com/forcedotcom/salesforcedx-vscode/blob/e9086dc44affcb4c51f3474223a82eb3f856bac8/packages/salesforcedx-vscode-services/src/core/projectService.ts#L22


as soon as that makes it in to develop this'll be simpler

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,178 +20,129 @@ import {
...
-      });
-      apexClassFileWatcher.onDidChange(() => {
-        debouncedRefresh();
-      });
-      context.subscriptions.push(apexClassFileWatcher);
-
-      // Initialize test discovery for TestController
-      if (testController) {
-        void testController.discoverTests();
+    const apexClassFileWatcher = vscode.workspace.createFileSystemWatcher('**/*.cls');
```
mshanemc: there's a service for that (it's a pubsub anything can subscribe to, then filter stuff it cares about)
https://github.com/forcedotcom/salesforcedx-vscode/blob/sm/deploy-commands/packages/salesforcedx-vscode-services/src/vscode/fileWatcherService.ts#L19C14-L19C32

which works like 
https://github.com/forcedotcom/salesforcedx-vscode/blob/e9086dc44affcb4c51f3474223a82eb3f856bac8/packages/salesforcedx-vscode-metadata/src/statusBar/sourceTrackingStatusBar.ts#L39

packages/salesforcedx-vscode-apex-testing/src/commands/apexTestRun.ts
```diff
@@ -57,7 +47,7 @@ class TestsSelector implements ParametersGatherer<ApexTestQuickPickItem> {
 
     const fileItems = [
       ...testSuites.map(testSuite => ({
-        label: basename(testSuite.toString(), '.testSuite-meta.xml'),
+        label: path.basename(testSuite.toString(), '.testSuite-meta.xml'),
         description: testSuite.fsPath,
```
mshanemc: I've found some bugs around this (where Utils.basename from vscode-uri) is better than converting URIs to strings (and then use uri.path)

Like, once you get to Uri you probably don't want to go back to string.  Might be ok for something like a label but the value should still stay uri

packages/salesforcedx-vscode-apex-testing/src/utils/fileHelpers.ts
```diff
@@ -0,0 +1,51 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as path from 'node:path';
+import type { QuickPickItem } from 'vscode';
+import * as vscode from 'vscode';
+import { URI } from 'vscode-uri';
+import { APEX_CLASS_EXT, IS_TEST_REG_EXP } from '../constants';
+
+export enum TestType {
```
mshanemc: is this an enum we have to use because it's something from the vscode API?  It not, use a string union type

packages/salesforcedx-vscode-apex-testing/src/utils/commandletHelpers.ts
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright (c) 2026, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as vscode from 'vscode';
```
mshanemc, line 8: meta: I would love to get out of this pattern going forward.  ex: you can prompt the user directly using the vscode api for input without all the classes wrapper around gatherers and checkers.

you can have a fn that checks the workspace (use the workspace service) or project (use the project service, etc).

you can decide what to do with your commands output or errors without all this inherited stuff forcing you into showing outputchannels, etc.

packages/salesforcedx-vscode-apex-testing/src/services/extensionProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { SalesforceVSCodeServicesApi } from 'salesforcedx-vscode-services';
+import * as vscode from 'vscode';
+
+let servicesApi: SalesforceVSCodeServicesApi | undefined;
+
+/**
+ * Gets the Salesforce Services extension API, activating if needed.
+ * This provides access to all services from the services extension.
+ */
+export const getServicesApi = async (): Promise<SalesforceVSCodeServicesApi> => {
```
mshanemc: fyi, I have a todo to get this in its own shared package
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002SqadeYAB/view?ws=%2Flightning%2Fr%2FADM_Epic__c%2Fa3QEE0000026sJ72AI%2Fview

packages/salesforcedx-vscode-apex-testing/src/telemetry/telemetry.ts
```diff
@@ -5,6 +5,84 @@
...
-export const telemetryService = TelemetryService.getInstance('salesforcedx-vscode-apex-testing');
+/**
+ * Simple telemetry service that uses vscode's telemetry API.
+ * Replaces TelemetryService from @salesforce/salesforcedx-utils-vscode
+ */
+class TelemetryService {
+  private static instances: Map<string, TelemetryService> = new Map();
+  private reporter: vscode.TelemetryLogger | undefined;
+
+  private constructor(_extensionId: string) {
```
mshanemc: why have unused props?

packages/salesforcedx-vscode-apex-testing/src/telemetry/telemetry.ts
```diff
@@ -5,6 +5,84 @@
...
 
-import { TelemetryService } from '@salesforce/salesforcedx-utils-vscode';
+import * as vscode from 'vscode';
 
-export const telemetryService = TelemetryService.getInstance('salesforcedx-vscode-apex-testing');
+/**
+ * Simple telemetry service that uses vscode's telemetry API.
+ * Replaces TelemetryService from @salesforce/salesforcedx-utils-vscode
+ */
+class TelemetryService {
```
mshanemc: not sure if you saw the telemetry video I did, but I'd like this kind of stuff to go through the Effect/otel path (get o11y, local tracing, AppInsights, etc) plus the org/userId stuff that this'll be missing.

it means wrapping more fns in Effects.fn where you can name the span, and add whatever logs or attributes you want.

[you get stuff like "how long did this fn take" for free, etc]

packages/salesforcedx-vscode-apex-testing/src/utils/activationTracker.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import * as vscode from 'vscode';
+import { telemetryService } from '../telemetry/telemetry';
+
+/**
+ * Simple activation tracker that measures activation time.
+ * Replaces ActivationTracker from @salesforce/salesforcedx-utils-vscode
+ */
+export class ActivationTracker {
```
mshanemc: this is also unnecessary in the Effect/otel stuff (have activate call an activateEffect and you get count/duration automatically)

packages/salesforcedx-vscode-apex-testing/src/utils/pathHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+const STATE_FOLDER = '.sfdx';
+const TOOLS = 'tools';
+const TEST_RESULTS = 'testresults';
+
+/**
+ * Creates a directory if it doesn't exist, using vscode.workspace.fs
+ * In web mode with virtual file systems, directories are created automatically by memfs
+ */
+const createDirectory = async (dirPath: string): Promise<void> => {
+  // Desktop mode - normal directory creation
```
mshanemc: you probably want these:
https://github.com/forcedotcom/salesforcedx-vscode/blob/sm/deploy-commands/packages/salesforcedx-vscode-services/src/vscode/fsService.ts

packages/salesforcedx-vscode-apex-testing/src/utils/pathHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+};
+
+/**
+ * Gets the test results folder path and creates it if it doesn't exist.
+ * Replaces getTestResultsFolder from @salesforce/salesforcedx-utils-vscode
+ */
+export const getTestResultsFolder = async (vscodePath: string, testType: string): Promise<string> => {
+  // In web mode with empty or invalid paths, return a placeholder path
+  if (process.env.ESBUILD_PLATFORM === 'web' && (!vscodePath || vscodePath.trim() === '')) {
```
mshanemc: what's wrong with the web paths?  I think we'd want these to work the same

packages/salesforcedx-vscode-apex-testing/src/utils/testReportGenerator.ts
```diff
@@ -90,42 +93,106 @@ export const writeAndOpenTestReport = async (
...
+  // Convert stream to UTF-8 string for FsService.writeFile
   const uint8Array = await streamToNormalizedUtf8Bytes(transformer);
-  const uri = vscode.Uri.file(reportPath);
-  await vscode.workspace.fs.writeFile(uri, uint8Array);
+  const content = new TextDecoder('utf-8').decode(uint8Array);
+
+  // Use FsService.writeFile from the services extension
+  // It handles toUri conversion and directory creation internally
+  const servicesApi = await getServicesApi();
+  const fsServiceLayer = Layer.mergeAll(
```
mshanemc: yeah, you'll also want this in the 1-Layer so that it doesn't get re-created eah time you `writeAndOpenTestReport`

packages/salesforcedx-vscode-apex-testing/src/utils/testReportGenerator.ts
```diff
@@ -90,42 +93,106 @@ export const writeAndOpenTestReport = async (
...
+    servicesApi.services.FsService.Default,
+    servicesApi.services.ChannelService.Default
```
mshanemc: oh, you did find it!

packages/salesforcedx-vscode-apex-testing/src/utils/testReportGenerator.ts
```diff
@@ -90,42 +93,106 @@ export const writeAndOpenTestReport = async (
...
+  );
+
+  // Construct the URI for the report file
+  // On desktop, use vscode.Uri.file() for proper file:// URI
+  // On web, construct URI relative to workspace root to ensure correct scheme
+  const workspaceRoot = vscode.workspace.workspaceFolders?.[0]?.uri;
+  let uri: vscode.Uri;
+
+  if (process.env.ESBUILD_PLATFORM === 'web' && workspaceRoot) {
```
mshanemc: there's some stuff here for that.

remember to use vscode-uri (and its Utils export for path-math like basename, join, etc on URIs) and not vscode.Uri

packages/salesforcedx-vscode-apex-testing/src/utils/testResultProcessor.ts
```diff
@@ -57,7 +57,6 @@ export const updateTestRunResults = (
   classItems: Map<string, vscode.TestItem>,
   codeCoverage: boolean = false
 ): void => {
```
mshanemc: can that be 1 object param instead of 6 of them sequentially?

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -301,30 +228,95 @@ export const buildClassToUriIndex = async (classNames: string[]): Promise<Map<st
   if (!folder || classNames.length === 0) {
     return new Map<string, vscode.Uri>();
   }
+
+  // In web mode with virtual file systems, findFiles might not work the same way
```
mshanemc: yeah, it seems not to work at all.  They never made `search` a real API (only proposed).

I have a WI for that, but I'm kinda kicking the can down the road.

mshanemc: one thing I think you could do if you need the paths would be to build a ComponentSet (lots of that service in the Services ext these days) out of the packageDirs, filtered by ApexClass, and then that'll have file paths for all the local classes.

It uses SDR to walk the fs, which works fine on web and desktop.

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -301,30 +228,95 @@ export const buildClassToUriIndex = async (classNames: string[]): Promise<Map<st
...
+  result: TestResult,
+  outputDir: string,
+  codeCoverage: boolean,
+  testService: TestService
+): Promise<void> => {
+  // In web mode, use FsService since testService.writeResultFiles uses callback-style fs operations
+  // In desktop mode, use testService.writeResultFiles which works correctly
+  if (process.env.ESBUILD_PLATFORM === 'web') {
+    try {
+      await writeTestResultJson(result, outputDir);
```
mshanemc: so, do we NOT get code coverage on the web?

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -301,30 +228,95 @@ export const buildClassToUriIndex = async (classNames: string[]): Promise<Map<st
...
+
+/** Writes test result JSON file, using FsService in web mode and testService.writeResultFiles in desktop mode */
+export const writeTestResultJsonFile = async (
+  result: TestResult,
+  outputDir: string,
+  codeCoverage: boolean,
+  testService: TestService
+): Promise<void> => {
+  // In web mode, use FsService since testService.writeResultFiles uses callback-style fs operations
+  // In desktop mode, use testService.writeResultFiles which works correctly
```
mshanemc: I'd be interested in what's in `testService.writeResultFiles` that doesn't work correctly on web.

I've fixed a LOT of stuff in SDR/STL/core libraries to make them behave, so maybe it could work.

packages/salesforcedx-vscode-apex-testing/src/views/icons/iconHelpers.ts
```diff
@@ -17,7 +17,8 @@ const VSCODE_APEX_TESTING_EXTENSION_NAME = 'salesforce.salesforcedx-vscode-apex-
  * @returns The Uri to the icon image.
  */
 const getIconPath = (key: keyof typeof ICONS) => {
```
mshanemc: why not just export `getIconPath` and use that instead of wrapping it in an object?

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,178 +20,129 @@ import {
...
-          // The old view will be shown automatically via the when clause in package.json
-        }
-      } catch (error) {
-        console.debug('Failed to refresh after UI mode change:', error);
-      }
-    }
-  });
-  context.subscriptions.push(testDiscoverySettingsWatcher);
+    // Initialize test discovery for TestController
+    testController.discoverTests().catch(error => {
```
mshanemc: you could use https://github.com/forcedotcom/salesforcedx-vscode/blob/sm/deploy-commands/packages/salesforcedx-vscode-services/src/core/defaultOrgService.ts (https://github.com/forcedotcom/salesforcedx-vscode/blob/sm/deploy-commands/packages/salesforcedx-vscode-services/src/core/schemas/defaultOrgInfo.ts for schema) to easily see if there IS an org attached

scripts/bundling/process-polyfill.js
```diff
@@ -110,6 +110,27 @@ const exit = () => {};
 // Command line arguments - return empty array for browser
 const argv = [];
 
+// Versions object - commonly accessed by dependencies
+const versions = {
```
mshanemc, line 114: intersting.  what was using this?

scripts/setup-jest.ts
```diff
@@ -96,8 +99,43 @@ enum TreeItemCollapsibleState {
   Expanded = 2
 }
 
+const createServicesExtensionMock = (vscodeMock: any) => ({
```
mshanemc: I don't think we're going to want to do this globally.

there's a section here on how to `provide` non-default versions of services or layers for testing https://effect.website/docs/requirements-management/layers/#injecting-test-dependencies

scripts/bundling/process-polyfill.js
CristiCanizales: I had to add it because I was getting like couldn't read property .node or something like that, and that was being thrown from `testService.writeResultFiles`

packages/salesforcedx-vscode-apex-testing/src/index.ts
CristiCanizales: Thanks! I realized (I think) we don't actually need it anymore, since we're using discovery api to get the tests and we don't care about local

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
CristiCanizales: well, automagically (after some changes) it works now 🥳

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
CristiCanizales: we do :)

scripts/bundling/process-polyfill.js
CristiCanizales: but looks like it's not needed anymore 🥳

packages/salesforcedx-vscode-apex-testing/src/utils/fileHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const readFile = async (filePath: string): Promise<string> => {
+  try {
+    const uri = URI.file(filePath);
+    const data = await vscode.workspace.fs.readFile(uri);
+    return Buffer.from(data).toString('utf8');
+  } catch (error) {
+    throw new Error(`Failed to read file ${filePath}: ${error instanceof Error ? error.message : String(error)}`);
+  }
+};
```
mshanemc: use fsService.readFile

packages/salesforcedx-vscode-apex-testing/src/testDiscovery/testDiscovery.ts
```diff
@@ -18,30 +17,12 @@ import { telemetryService } from '../telemetry/telemetry';
...
-
-/**
- * Checks if the Test Explorer UI should be used instead of the traditional Test View
- */
-export const useTestExplorer = (): boolean => {
-  const config = vscode.workspace.getConfiguration('salesforcedx-vscode-apex-testing');
-  return config.get<boolean>('testing.useTestExplorer', false);
-};
-
 export const discoverTests = async (
```
mshanemc: complexity: the 2 real consumers both use showAllMethods:true and don't set pageSize.  do you have future plans for those, or would it be simpler of the only option was the namespace?

mshanemc: If you were to wrap this entire thing in Effect, pass in the All Layer, then you could yield the Connection.

That has the advantage of being a typed error, with a nice message already provided for you (and it gives more detail, gets the traces, etc)

It would also have the advantage of simplifying the telemetry stuff (not having to do all the `startTime` stuff manually, etc).  You could still add the attributes for `numClasses` etc to the span.

packages/salesforcedx-vscode-apex-testing/src/utils/pathHelpers.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** Gets the test results folder path and creates it if it doesn't exist */
+export const getTestResultsFolder = async (vscodePath: string, testType: string): Promise<string> => {
+  const pathToTestResultsFolder = path.join(vscodePath, STATE_FOLDER, TOOLS, TEST_RESULTS, testType);
+
+  await Effect.runPromise(
+    Effect.gen(function* () {
+      const api = yield* (yield* ExtensionProviderService).getServicesApi;
+      const fsService = yield* api.services.FsService;
+      yield* fsService.createDirectory(pathToTestResultsFolder);
```
mshanemc: according to the docs, the vscode createDirectory "has mkdir semantics".  So the error you'd get is **probably** not the "already exists" but something else/worse.

catchAll might be ok, but I think you'd at least want to channel the message  

ex
```ts
}).pipe(
      Effect.tapError(error => Effect.logError(error)),
      Effect.catchAll(() => Effect.void), // Ignor
```

or the channelService equivalent

packages/salesforcedx-vscode-apex-testing/src/utils/testReportGenerator.ts
```diff
@@ -90,42 +91,83 @@ export const writeAndOpenTestReport = async (
...
+    await channelService.appendLine(`Failed to write test report: ${String(error)}`);
+    throw error;
+  }
+
+  // Create URI for opening the file
+  // On desktop: use vscode.Uri.file() for proper file:// URI
+  // On web: construct URI relative to workspace root for correct scheme
+  const uri =
+    process.env.ESBUILD_PLATFORM === 'web'
+      ? vscode.Uri.joinPath(vscode.workspace.workspaceFolders![0].uri, reportPath.replace(/^\/[^/]+/, ''))
```
mshanemc: use vscode-uri `Utils`

packages/salesforcedx-vscode-apex-testing/src/utils/pathHelpers.ts
```diff
@@ -0,0 +1,32 @@
...
+import * as path from 'node:path';
+import { AllServicesLayer, ExtensionProviderService } from '../services/extensionProvider';
+
+const STATE_FOLDER = '.sfdx';
+const TOOLS = 'tools';
+const TEST_RESULTS = 'testresults';
+
+/** Gets the test results folder path and creates it if it doesn't exist */
+export const getTestResultsFolder = async (vscodePath: string, testType: string): Promise<string> => {
+  const pathToTestResultsFolder = path.join(vscodePath, STATE_FOLDER, TOOLS, TEST_RESULTS, testType);
```
mshanemc: fyi, I've been wrestling a lot with path/URI stuff too on metadata.  I've been changing a lot of the fsService to accept either (path|URI).  I think I'm leaning toward "stay with uri" as much as possible, only converting to path for things that require it (ex: SDR).

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -295,36 +221,88 @@ const convertApiToApexTestMethods = async (classes: ToolingTestClass[]): Promise
...
+        [api.services.ProjectService, api.services.MetadataRetrieveService],
+        { concurrency: 'unbounded' }
+      );
+
+      // Get package directories from the project
+      const sfProject = yield* projectService.getSfProject;
+      const packageDirs = sfProject.getPackageDirectories().map(dir => dir.fullPath);
+
+      // Build ComponentSet for all ApexClass files in the project
+      const componentSet = yield* retrieveService.buildComponentSetFromSource(
```
mshanemc, line 230: fyi, this is moving soon (componentSetService).  Trying to get things organized better.

I'll fix this ext as part of that merge since this'll go in before mine.

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -295,36 +221,88 @@ const convertApiToApexTestMethods = async (classes: ToolingTestClass[]): Promise
...
+      );
+
+      // Build index from component name to file URI
+      const classNameSet = new Set(classNames);
+      const index = new Map<string, vscode.Uri>();
+
+      for (const component of componentSet.getSourceComponents()) {
+        // component.content is the .cls file path
+        if (component.content && classNameSet.has(component.name)) {
+          // Prefer shorter paths (files closer to workspace root)
```
mshanemc, line 242: for a given component name, why would there be different lengths?

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -295,36 +221,88 @@ const convertApiToApexTestMethods = async (classes: ToolingTestClass[]): Promise
...
+          const existingUri = index.get(component.name);
+          if (!existingUri || component.content.length < getUriPath(existingUri).length) {
+            index.set(component.name, vscode.Uri.file(component.content));
```
mshanemc: use vscode-uri.  I need to make a lint rule about this

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
```diff
@@ -295,36 +221,88 @@ const convertApiToApexTestMethods = async (classes: ToolingTestClass[]): Promise
...
+  const testRunId = result.summary?.testRunId;
+  const jsonFilename = testRunId ? `test-result-${testRunId}.json` : 'test-result.json';
+  const jsonFilePath = path.join(outputDir, jsonFilename);
+  const jsonContent = JSON.stringify(result, null, 2);
+
+  await Effect.runPromise(
+    Effect.gen(function* () {
+      const api = yield* (yield* ExtensionProviderService).getServicesApi;
+      const svc = yield* api.services.FsService;
+      yield* svc.writeFile(jsonFilePath, jsonContent);
```
mshanemc, line 273: here, writeFile does take `path:string|URI` already if you wanted to stay in URI-land

packages/salesforcedx-vscode-apex-testing/src/views/icons/iconHelpers.ts
```diff
@@ -5,23 +5,15 @@
...
-  const iconUri = Utils.joinPath(baseExtensionPath, ...ICONS[key]);
-  return iconUri;
-};
-
-export const iconHelpers = {
-  getIconPath
+/** Get the Uri for an icon located in the resources directory */
+export const getIconPath = (key: keyof typeof ICONS): vscode.Uri => {
+  const extension = vscode.extensions.getExtension(VSCODE_APEX_TESTING_EXTENSION_NAME);
+  const baseExtensionPath = extension?.extensionUri ?? vscode.Uri.file('');
```
mshanemc: when would there not be an extension Uri?  Is this just, "missing icons sem fine?"

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,147 +24,115 @@ import {
...
+  Effect.gen(function* () {
+    yield* Effect.log('Salesforce Apex Testing extension is activating...');
+
+    // Initialize the shared output channel from services API
+    yield* Effect.promise(() => initializeOutputChannel());
+
+    // Check if we're in a Salesforce project and set context (side effect of isSalesforceProject)
+    const api = yield* (yield* ExtensionProviderService).getServicesApi;
+    const projectService = yield* api.services.ProjectService;
+    yield* projectService.isSalesforceProject.pipe(Effect.catchAll(() => Effect.void));
```
mshanemc: if you're not inside a project, would you still want to do the rest of the stuff?

packages/salesforcedx-vscode-apex-testing/src/channels/index.ts
```diff
@@ -4,9 +4,66 @@
...
+  if (_outputChannel) return;
+
+  await Effect.runPromise(
+    Effect.gen(function* () {
+      const api = yield* (yield* ExtensionProviderService).getServicesApi;
+      const svc = yield* api.services.ChannelService;
+      _outputChannel = yield* svc.getChannel;
+    }).pipe(Effect.provide(AllServicesLayer))
+  );
+};
```
mshanemc: since this is only used inside another effect, you don't really want to Effect.runPromise...just return Effect.gen and then the consumer (activateEffect) can `yield*` initializeOutputChannel.

runPromise, runSync, etc. are really for only "run effect code from outside of Effect land"

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,147 +24,115 @@ import {
...
-        refreshTimeout = setTimeout(() => {
-          if (testController) {
-            void testController.refresh();
+/** Check if an org is connected by looking at TargetOrgRef */
+const hasOrgConnected = (orgInfo: { username?: string; orgId?: string }): boolean =>
+  Boolean(orgInfo.username ?? orgInfo.orgId);
+
+/** Initialize test discovery when an org is available, and re-discover on org changes */
+const initializeTestDiscovery = (testController: ReturnType<typeof getTestController>): Promise<void> =>
+  Effect.runPromise(
```
mshanemc: this is another place that could be simply `yield`ed if it weren't wrapped in `runPromise` since it's only used from inside another Effect

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,147 +24,115 @@ import {
...
+    const projectService = yield* api.services.ProjectService;
+    yield* projectService.isSalesforceProject.pipe(Effect.catchAll(() => Effect.void));
+
+    const testOutlineProvider = getTestOutlineProvider();
+    const testController = getTestController();
+    yield* Effect.log('[Apex Testing] Test controller created');
+
+    if (vscode.workspace?.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
+      const workspaceUri = vscode.workspace.workspaceFolders[0].uri;
+      // In web mode, only set up file watchers if we have a file scheme workspace
```
mshanemc: this is definitely not true, those orgChanges are a result of file-watching the config file.

mshanemc: I would suggest using FileWatcher service (subscribe to the pubsub so you get an Effect stream of vscode.FileEvent.  You can can Stream.filter for URI patterns and event types you can care about ().

The other nice feature of it is that you don't have to clean it it up.  It does require a scope (see ExtensionScope for an example) or look in the deploy-from-web for usage examples.

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -177,37 +148,33 @@ export const activate = async (context: vscode.ExtensionContext) => {
       }
     );
```
mshanemc: why not have this with the rest of cmd registration?

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,147 +24,115 @@ import {
...
+      // This handles the startup case where the ref is empty
+      yield* connectionService.getConnection.pipe(
+        Effect.flatMap(() => SubscriptionRef.get(targetOrgRef)),
+        Effect.flatMap(orgInfo => (hasOrgConnected(orgInfo) ? discoverForOrg(orgInfo) : Effect.void)),
+        Effect.catchAll(error => {
+          console.debug('[Apex Testing] Initial connection failed (no org configured?):', error);
+          return Effect.void;
+        })
+      );
+    }).pipe(
```
mshanemc: I gave this a spin (changing orgs and seeing if it'll change the test panel, didn't seem to).

I think what's happening is that `fork` won't outlive the parent Effect that it came from (initializeTestDiscovery) so when it exits, the stuff it "parents" does do.

You can `forkDaemon` so it stays running permanently, but it's probably better to do the ExtensionScope thing and `forkIn` (so that it stays running until the extension scope closes, which you only do during deactivate).

when I tried that, what I saw was that nothing is calling the getConnection again (which would cause the new org details to populate).

Basically 

```ts
// Subscribe to org changes and re-discover tests when org changes
    yield* Effect.forkDaemon(
      targetOrgRef.changes.pipe(
        Stream.tap(org => Effect.promise(() => channelService.appendLine(`Target org changed to ${JSON.stringify(org)}`))),
        // if we don't have an orgId, try to get the connection to cause another evnt to fire with it
        Stream.tap(org => !org.orgId ? connectionService.getConnection : Effect.succeed(undefined)),
        Stream.filter(hasOrgConnected),
        Stream.tap(org => Effect.promise(() => channelService.appendLine(`Discovering tests for org: ${org.username ?? org.orgId}`))),
        Stream.runForEach(discoverForOrg))
    );
```

and then this is all you need to kick off the first one
```ts
 yield* connectionService.getConnection;
    }).pipe(
      Effect.catchAll(error => {
        console.debug('[Apex Testing] Test discovery setup failed:', error);
        return Effect.void;
      }),
      Effect.provide(AllServicesLayer)
    );
```

mshanemc: I removed this "lastDiscovered" stuff and it didn't seem to make a difference.

packages/salesforcedx-vscode-apex-testing/src/utils/orgApexClassProvider.ts
```diff
@@ -71,6 +70,11 @@ class OrgApexClassProvider implements vscode.TextDocumentContentProvider {
     const uri = vscode.Uri.parse(`${SCHEME}:${baseClassName}.cls`);
```
mshanemc: definitely check out Effect.cache
https://effect.website/docs/caching/cache/

there's a video walkthrough https://www.youtube.com/watch?v=e0qYYn4deWs

packages/salesforcedx-vscode-apex-testing/test/jest/utils/orgApexClassProvider.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 jest.mock('../../../src/coreExtensionUtils', () => ({
-  getVscodeCoreExtension: jest.fn()
+  getConnection: jest.fn()
```
mshanemc: part of the fun of Effect dependencies (service/Layer stuff) is that you can inject them into the effect you're calling instead of using the AllServicesLayer

if you want

packages/salesforcedx-vscode-apex-testing/test/jest/utils/testReportGenerator.test.ts
```diff
@@ -5,6 +5,78 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+// Mock vscode.workspace.fs.writeFile - this is used by FsService internally
```
mshanemc, line 8: fsService might be a very nice on to have a test instance of
channelService, too

packages/salesforcedx-vscode-apex-testing/src/telemetry/telemetry.ts
```diff
@@ -5,6 +5,37 @@
...
+  });
+};
+
+/**
+ * Simplified telemetry service for backward compatibility.
+ * Prefer using Effect.withSpan directly in Effect-based code.
+ */
+export const telemetryService = {
+  /** No-op for backward compatibility - initialization happens via AllServicesLayer */
+  initializeService: async (_context: unknown): Promise<void> => {},
```
mshanemc, line 30: i don't think this is used?

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -177,37 +148,33 @@ export const activate = async (context: vscode.ExtensionContext) => {
...
+          return undefined;
+        }
       }
-    }
-  };
-};
+    };
+  }).pipe(Effect.withSpan('apex-testing.activation'), Effect.provide(AllServicesLayer));
+
+export const activate = (context: vscode.ExtensionContext) => Effect.runPromise(activateEffect(context));
```
mshanemc, line 215: awesome!

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -177,37 +148,33 @@ export const activate = async (context: vscode.ExtensionContext) => {
...
-        return provider.getTestClassName(URI.parse(uri.toString()));
-      } catch (error) {
-        console.debug('Failed to get test class name:', error);
-        return undefined;
+    // Commands
+    const commands = registerCommands(context);
+    context.subscriptions.push(commands);
+
+    // Initial refresh of test view to populate tests when extension activates
+    void refreshTestViewOnActivation();
```
mshanemc: void can be risky if anything fails (unhandled promise exception, etc).

if you want it to run in a separate fiber and not wait, you can Effect.forkDaemon(Effect.tryPromise) etc.  (it'll run in the background until it's completed).  and pipe catchAll to it if you really want to swallow errors, or log them via tapError, etc.

packages/salesforcedx-vscode-apex-testing/src/telemetry/telemetry.ts
```diff
@@ -5,6 +5,37 @@
...
+  ).catch(() => {
+    // Best effort - don't fail if telemetry fails
+  });
+};
+
+/**
+ * Simplified telemetry service for backward compatibility.
+ * Prefer using Effect.withSpan directly in Effect-based code.
+ */
+export const telemetryService = {
```
mshanemc, line 28: checked, can see a few spans hitting the docker local telemetry, so those would flow to appinsights

Span attributes
common.cpus	
"Apple M2 Max(12 x 2400)"

common.platformversion	
"25.2.0"

common.systemmemory	
"68719476736.00 GB"

common.vscodemachineid	
"f7d40e71f946d8715f78918f2eebb99c2c6d1c7f6f9e3aa9b1aa4a7cb553e940"

common.vscodesessionid	
"d5f6b965-7ced-47e4-abe1-f72ef6db7fe31769053084611"

common.vscodeuikind	
"Desktop"

common.vscodeversion	
"1.105.1"

devHubOrgId	
"00D6A000001V3UBUA0"

isSandbox	
"false"

isScratch	
"true"

orgId	
"00DEc00000YNX1KMAX"

outputFormat	
"markdown"

reportDurationMs	
27

telemetryTag	
"shane-mclaughlin"

tracksSource	
"true"

trigger	
"testExplorer"

Resource attributes:service.name=salesforcedx-vscode-servicesservice.version=2025-08-15T20:49:30.000Ztelemetry.sdk.language=nodejstelemetry.sdk.name=@effect/opentelemetry
 Events (1)
1.15ms ([Telemetry] apexTest...)
event name	
"[Telemetry] apexTestReportGenerated"

reportDurationMs	
27

outputFormat	
"markdown"

trigger	
"testExplorer"

effect.fiberId	
"#2965"

effect.logLevel

mshanemc: the more of the code is in Effect and the more spans you have, the clearer the picture is of what fn the logs came from, how long the various pieces take, etc.

packages/salesforcedx-vscode-apex-testing/src/index.ts
```diff
@@ -21,147 +24,115 @@ import {
...
+
+          context.subscriptions.push(testResultFileWatcher);
         }
-      } catch (error) {
-        console.debug('Failed to refresh after UI mode change:', error);
       }
+
+      // Initialize test discovery when an org is available, and re-discover on org changes
+      void initializeTestDiscovery(testController);
     }
```
mshanemc: could probably forkDaemon if you want to get activated ASAP

packages/salesforcedx-vscode-apex-testing/src/utils/testUtils.ts
CristiCanizales:
```
force-app/main/default/classes/MyClass.cls (shorter)
packages/my-package/force-app/main/default/classes/MyClass.cls (longer)
```
This should change with [W-20970522](https://gus.lightning.force.com/lightning/r/a07EE00002TDphnYAD/view)

packages/salesforcedx-vscode-apex-testing/src/testDiscovery/testDiscovery.ts
CristiCanizales: not really, I think we'll pretty much just use namespacePrefix so it would indeed be simpler to just keep that one

packages/salesforcedx-vscode-apex-testing/src/utils/testReportGenerator.ts
```diff
@@ -90,42 +92,83 @@ export const writeAndOpenTestReport = async (
...
+    throw error;
+  }
+
+  // Create URI for opening the file
+  // On desktop: use URI.file() for proper file:// URI
+  // On web: construct URI relative to workspace root for correct scheme
+  const uri =
+    process.env.ESBUILD_PLATFORM === 'web'
+      ? Utils.joinPath(vscode.workspace.workspaceFolders![0].uri, reportPath.replace(/^\/[^/]+/, ''))
+      : vscode.Uri.file(reportPath);
```
CristiCanizales, line 121: URI.file does not work here, the button stops working

PR #6780, opened 2026-01-16
[W-20920609] fix: authorizing an org without inputting an alias in the input box uses default alias vscodeOrg

daphne-sfdc:
### What does this PR do?
When you authorize an org, you are asked to input an alias in the input box.  Supposedly if you leave that box blank, the alias should default to `vscodeOrg`.  This PR fixes a regression where an empty string is used as the alias instead of the default value, hence no alias was set.

@W-20920609@

### Functionality Before
Screenshot 2026-01-16 at 2 41 16 PM

### Functionality After
Screenshot 2026-01-16 at 2 43 14 PM

### DIFF
--- a/packages/salesforcedx-vscode-org/src/commands/auth/authParamsGatherer.ts
+++ b/packages/salesforcedx-vscode-org/src/commands/auth/authParamsGatherer.ts
@@ -123,14 +123,15 @@ export class AuthParamsGatherer implements ParametersGatherer<AuthParams> {
           prompt: nls.localize('parameter_gatherer_enter_alias_name'),
           placeHolder: DEFAULT_ALIAS
         });
-    // Hitting enter with no alias will default the alias to 'vscodeOrg'
+    // Closing the dialog (ESC/cancel) will cancel the operation
     if (alias === undefined) {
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-org/src/commands/auth/authParamsGatherer.ts
```diff
@@ -123,14 +123,15 @@ export class AuthParamsGatherer implements ParametersGatherer<AuthParams> {
...
+    // Closing the dialog (ESC/cancel) will cancel the operation
     if (alias === undefined) {
       return { type: 'CANCEL' };
     }
+    // Hitting enter with no alias will default the alias to 'vscodeOrg'
     return {
       type: 'CONTINUE',
       data: {
-        alias: alias ?? DEFAULT_ALIAS,
+        alias: alias || DEFAULT_ALIAS,
```
daphne-sfdc, line 134: `??`: Only treats `null` and `undefined` as "missing" values
`||`: Treats ALL falsy values as "missing" (including `""`, `0`, `false`, `null`, `undefined`, `NaN`)

Issue #6790, opened 2026-01-21
Concise Apex Results Regression


k-capehart:
### Summary

The `Test-run-concise` option is no longer working as expected, and displays both passing and failing test results in the output.

### Steps To Reproduce:

1.  Create an Apex Test Class, with a test that is expected to pass and a test that is expected to fail
2. Run the tests using the VS Code extension, by clicking on "Run All Tests" at the top of the Apex Class

### Expected result

In the output, only the failing result should be displayed.

### Actual result

In the output, both the failing and passing results are displayed.

### Additional information

<img width="1233" height="355" alt="Image" src="https://github.com/user-attachments/assets/e7b7e8f3-5114-449c-9ed8-531498a67211" />

**Salesforce Extension Version in VS Code**: 65.13.1

**Salesforce CLI Version**: @salesforce/cli/2.118.20 darwin-arm64 node-v24.13.0

**OS and version**: MacOS 15.7.3 (24G419)

**VS Code version**: 1.108.1 (and Cursor 2.3.41)

**Most recent version of the extensions where this was working**: I think prior to 1.110 but not entirely sure

PR #6792, opened 2026-01-21
fix: concise configuration not being passed

k-capehart:
### What does this PR do?

Correctl pass the concise option to the formatter.

#6790 

### Functionality Before
Concise option did not work.

### Functionality After
Concise option works.


CristiCanizales:
Hi @k-capehart, thanks for creating this PR! There was something else missing to fix the feature, but I'm taking care of it with PR https://github.com/forcedotcom/salesforcedx-vscode/pull/6794. This should go out in next week's release, thanks again!

Issue #6752, opened 2026-01-06
Change default org and Authorize org disappear from toolbar and command pallet with latest VSCode extension


hungrypipo:
> **Note** 
> Before you submit your issue, make sure that:
> - You're using the latest version of Salesforce CLI.
> - You've searched both open and closed issues for related posts.
> - You've used the `doctor` command to diagnose common issues. 
> - You understand that GitHub Issues don't adhere to any agreement or SLA.
>   - If you require immediate assistance, use official channels such as Salesforce Customer Support.

### Summary
Using
Windows 11
VSCode 1.106.3
NodeJS v20.19.5
@salesforce/cli/2.114.5

### Steps To Reproduce

If I update the salesforce.salesforcedx-vscode-core extension component past v64.17.2 say to v65.9.1 I lose the Open Org and Change Default Org from the lower toolbar and the SFDX:Authorize an Org from the VSCode command pallet.

I don't see any errors anywhere those function are just gone

daphne-sfdc:
Try this:
1. `cd $HOME`
2. `cd .vscode`
3. `cd extensions`

shlebscrx:
Hey, 

Im currently having the same issue: 
Im not able to auth an organization nor upload any code ( my sandbox was authorized before I deleted the repository) 

I already tried the steps suggested in this thread, but without any success.
Problam started  Friday 9th of January 2026

Summary
Vs code version: 
    Version: 1.108.0 (user setup)
    Commit: 94e8ae2b28cb5cc932b86e1070569c4463565c37
    Date: 2026-01-08T13:53:10.781Z
    Electron: 39.2.7
    ElectronBuildId: 12953945
    Chromium: 142.0.7444.235
    Node.js: 22.21.1
    V8: 14.2.231.21-electron.0
    OS: Windows_NT x64 10.0.26200


 sf version --verbose --json
{
  "architecture": "win32-x86",
  "cliVersion": "@salesforce/cli/2.117.7",
  "nodeVersion": "node-v22.21.1",
  "osVersion": "Windows_NT 10.0.26200",
  "rootPath": "C:\\Program Files (x86)\\sf\\client",
  "shell": "C:\\Program Files\\Git\\usr\\bin\\bash.exe",
  "pluginVersions": [
    "@oclif/plugin-autocomplete 3.2.39 (core)",
    "@oclif/plugin-commands 4.1.38 (core)",
    "@oclif/plugin-help 6.2.36 (core)",
[...]

PR #6797, opened 2026-01-22
[W-20824301] fix: CLI Integration extension activates properly when org authorized via session ID is expired

daphne-sfdc:
### What does this PR do?
Previously, multiple users were facing an issue in Agentforce Vibes IDE where their session sat idle for too long and timed out, but they're unable to reauthorize.  There is a popup on their screen telling them that they need to reauthorize, but clicking the **Login** button does nothing because the CLI Integration extension and the Org Management extension were not activated.
Screenshot 2026-01-22 at 3 23 23 PM

@W-20824301@, #6752

### Functionality Before
In order to get the CLI Integration extension to activate, users had to use a workaround by reauthorizing to the org via CLI's `sf org login web --set-default` and reloading the AFV window.

### Functionality After
The CLI Integration extension and Org Management extension properly activate when the org is timed out.  Clicking the **Login** button works as expected to trigger the login screen in the next tab.

### Testing
Prior to creating this PR, I tested the following cases and ensured only 1 popup appears for each case, and the authorization always works as expected.
1. As soon as the popup appears, click Cancel ✅
2. Wait some time before clicking Cancel ✅
3. As soon as the popup appears, click Login, fill out the login form fast ✅
4. Wait some time before clicking Login, fill out the login form fast ✅
5. As soon as the popup appears, click Login, wait some time before filling out the login form ✅
6. Wait some time before clicking Login, wait some time before filling out the login form ✅

### DIFF
--- a/config/jest.base.config.js
+++ b/config/jest.base.config.js
@@ -13,6 +13,8 @@ module.exports = {
   // Ignore .vscode-test directories to prevent Haste module map conflicts
   modulePathIgnorePatterns: ['/.vscode-test/'],
   testPathIgnorePatterns: ['/.vscode-test/']
+  // Note: isolatedModules: true would speed up tests significantly (~4-7x faster)
+  // but breaks dynamic imports (await import(...)) used in orgUtil.ts and elsewhere
   // This collectCoverageFrom will show coverage for all files in a projects, but slows down calculating coverage results.
   // Can be a good tool for measuring coverage of the project as a whole locally, but shouldn't be committed at this time.
   // Off:
--- a/packages/salesforcedx-vscode-apex-testing/jest.config.js
+++ b/packages/salesforcedx-vscode-apex-testing/jest.config.js
@@ -3,5 +3,10 @@ const baseConfig = require('../../config/jest.base.config.js');
 module.exports = {
   ...baseConfig,
   displayName: 'salesforcedx-vscode-apex-testing',
-  roots: ['<rootDir>/test']
+  roots: ['<rootDir>/test'],
+  // Enable isolatedModules for faster test execution
+  // This package doesn't use dynamic imports, so it's safe
+  transform: {
+    '^.+\\.tsx?$': ['ts-jest', { isolatedModules: true }]
+  }
 };
--- a/packages/salesforcedx-vscode-apex/jest.config.js
+++ b/packages/salesforcedx-vscode-apex/jest.config.js
@@ -1,4 +1,10 @@
 // eslint:disable-next-line:no-var-requires
 const baseConfig = require('../../config/jest.base.config');
 
-module.exports = Object.assign({}, baseConfig);
+module.exports = Object.assign({}, baseConfig, {
+  // Enable isolatedModules for faster test execution
+  // This package doesn't use dynamic imports, so it's safe
+  transform: {
+    '^.+\\.tsx?$': ['ts-jest', { isolatedModules: true }]
+  }
+});
--- a/packages/salesforcedx-vscode-i18n/jest.config.js
+++ b/packages/salesforcedx-vscode-i18n/jest.config.js
@@ -2,5 +2,10 @@ const baseConfig = require('../../config/jest.base.config');
 
 module.exports = {
   ...baseConfig,
-  displayName: '@salesforce/vscode-i18n'
+  displayName: '@salesforce/vscode-i18n',
+  // Enable isolatedModules for faster test execution
+  // This package doesn't use dynamic imports, so it's safe
+  transform: {
+    '^.+\\.tsx?$': ['ts-jest', { isolatedModules: true }]
+  }
 };
--- a/packages/salesforcedx-vscode-lightning/jest.config.js
+++ b/packages/salesforcedx-vscode-lightning/jest.config.js
@@ -2,5 +2,11 @@ const baseConfig =  require('../../config/jest.base.config');
 
 module.exports = Object.assign({},
   baseConfig,
-  {}
+  {
+    // Enable isolatedModules for faster test execution
+    // This package doesn't use dynamic imports, so it's safe
+    transform: {
+      '^.+\\.tsx?$': ['ts-jest', { isolatedModules: true }]
+    }
+  }
 );
--- a/packages/salesforcedx-vscode-lwc/jest.config.js
+++ b/packages/salesforcedx-vscode-lwc/jest.config.js
@@ -1,4 +1,10 @@
 // eslint:disable-next-line:no-var-requires
 const baseConfig = require('../../config/jest.base.config');
 
-module.exports = Object.assign({}, baseConfig);
+module.exports = Object.assign({}, baseConfig, {
+  // Enable isolatedModules for faster test execution
+  // This package doesn't use dynamic imports, so it's safe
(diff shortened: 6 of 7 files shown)


packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -305,8 +314,6 @@ export const activate = async (extensionContext: vscode.ExtensionContext): Promi
 };
 
 const initializeProject = async (extensionContext: vscode.ExtensionContext) => {
-  await WorkspaceContext.getInstance().initialize(extensionContext);
```
daphne-sfdc, line 308: The primary fix was to remove this line from **index.ts**.  I did this because the `initializeProject()` function is called by the `activate()` function, which means the initialization is part of the activation of the CLI Integration extension.

If this line is kept, the initialization process tries to get a connection to the default org and gets stuck if the current org is in an invalid state.  This means the CLI Integration extension is stuck in the Activating state.  However, the Org Management doesn't activate unless the CLI Integration extension is activated, hence the **Login** button on the popup doesn't work.

By removing this line, the activation process for the CLI Integration extension bypasses the check for the valid org, allowing both it and the Org Management extension to be activated.  Afterwards, the initialization process happens and detects the invalid org, prompting the popup.  By that time, because the Org Management extension is already activated, the **Login** button works as expected.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -102,22 +109,40 @@ export class WorkspaceContextUtil {
...
-            nls.localize('error_access_token_expired_login_button')
-          );
-          if (selection === 'Login') {
-            await vscode.commands.executeCommand('sf.org.login.web', connectionDetails.connection.instanceUrl);
-          }
-        }
-        this.knownBadConnections.add(this._username);
 
+        // Check if there's already an active login prompt for this user (shared across all extensions)
+        const existingPrompt = await getSharedLoginPrompt(this._username);
```
daphne-sfdc, line 114: Each extension instantiates its own instance of `WorkspaceContextUtil`.  This is a problem because each extension, during its activation process, detects an invalid state in the default org and produces an identical popup asking the user to reauthorize to the org.  If one has the entire Salesforce Extension pack installed, they'll see the re-auth popup appear **6 times**.

We need to check at the **user level**, not at the extension level - each user should see only 1 popup total, which they can choose to Login or Cancel.  Therefore, I created this set of helper functions to do this tracking in the CLI Integration extension.  It's the same logic as the `webUserId` used for telemetry: the same value shared across all the extensions for the same user.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
```diff
@@ -102,22 +109,40 @@ export class WorkspaceContextUtil {
...
+
+        // we only want to display one message per username across ALL extensions, even though many consumers are requesting connections.
+        const isKnownBad = isKnownBadConnection(this._username);
```
mshanemc, line 117: what's the scenario where someone has a prompt open but it's for a **different** username?

packages/salesforcedx-vscode-core/src/index.ts
```diff
@@ -223,6 +226,13 @@ export const activate = async (extensionContext: vscode.ExtensionContext): Promi
     taskViewService,
     telemetryService,
     workspaceContextUtils,
+    getSharedLoginPrompt: (username: string) => sharedAuthState.getLoginPrompt(username),
+    setSharedLoginPrompt: (username: string, promise: Promise<void>) =>
+      sharedAuthState.setLoginPrompt(username, promise),
+    clearSharedLoginPrompt: (username: string) => sharedAuthState.clearLoginPrompt(username),
+    isKnownBadConnection: (username: string) => sharedAuthState.isKnownBad(username),
+    addKnownBadConnection: (username: string) => sharedAuthState.addKnownBad(username),
+    clearKnownBadConnection: (username: string) => sharedAuthState.clearKnownBad(username),
```
mshanemc: for tidiness (API surface) could you just export SharedAuthState (the class instance) instead of every static method from it?  Then the consumer could use the methods they need off of the class instance?

packages/salesforcedx-utils-vscode/src/helpers/authUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Copyright (c) 2026, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { extensions } from 'vscode';
+
+// Type definition for the Core extension API's auth-related methods
+interface SalesforceVSCodeCoreAuthApi {
```
mshanemc: there are other places where we get types-only imports from core like
```
import type { SalesforceVSCodeCoreApi } from 'salesforcedx-vscode-core';

/** makes sure the extension is active and returns the instance */
export const getVscodeCoreExtension = async (): Promise<vscode.Extension<SalesforceVSCodeCoreApi>> => {
  const vscodeCoreExtension = vscode.extensions.getExtension<SalesforceVSCodeCoreApi>(
    'salesforce.salesforcedx-vscode-core'
  );
  if (!vscodeCoreExtension) {
    throw new Error('Could not fetch a SalesforceVSCodeCoreApi instance');
  }

  // vscode could deactivate extensions, so just in case.
  if (!vscodeCoreExtension.isActive) {
    await vscodeCoreExtension.activate();
  }
  return vscodeCoreExtension;
};

```
I get that that might create a circle, haven't tried it to see if type-only is ok.  
If not, it would make MORE sense to export the type from utils-vsocde and use it as part of the the type for `SalesforceVSCodeCoreApi` in vscode-core's type export (and part of that class's definition)
and maybe it gets a lot simpler if the knownBadConnection stuff doesn't stick around.

otherwise there's nothing keeping them aligned `coreExtension.exports`.

packages/salesforcedx-utils-vscode/src/context/workspaceContextUtil.ts
daphne-sfdc: It works exactly as expected!  You can use the prompt to reauthorize to the other org.

mshanemc: asked differently, why is the "knownBadConnection" stuff necessary?  what happens if it doesn't exist?  How would there be 2 existingPrompts for different usernames

daphne-sfdc: I assume you're asking about the case where you have **2 orgs authorized via session ID and expired**?  To answer that question, you **do** get 2 popups!  One for the default org as soon as the CB window is reloaded, and the other when you switch to that org in the org picker.

daphne-sfdc: The `knownBadConnection` stuff is used to prevent multiple popups for the same username.  As soon as you click Login or Cancel on one popup, that org is tracked to prevent another popup appearing for the same username when another extension is activated.

PR #6800, opened 2026-01-23
W-21012167: lls ts config generation

madhur310:
### What does this PR do?
Fix ts config generation by LLS

@W-21012167@

### DIFF
--- a/packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
+++ b/packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
@@ -51,7 +51,6 @@ export class FileSystemDataProvider implements IFileSystemProvider {
     // If connection is available, use LSP workspace/applyEdit to create/write the file
     if (connection) {
       const fileUri = URI.file(normalizedUri).toString();
-      Logger.info(`[FileSystemProvider] Requesting client to create file: ${normalizedUri}`);
 
       const edit: WorkspaceEdit = {
         documentChanges: [
@@ -71,11 +70,9 @@ export class FileSystemDataProvider implements IFileSystemProvider {
 
         if (!result.applied) {
           const errorMsg = result.failureReason ?? 'Unknown error';
-          Logger.error(`[FileSystemProvider] Client rejected file creation: ${normalizedUri}. Reason: ${errorMsg}`);
           throw new Error(`Failed to create file ${normalizedUri}: ${errorMsg}`);
         }
 
-        Logger.info(`[FileSystemProvider] Successfully created file via LSP: ${normalizedUri}`);
       } catch (error) {
         // Handle connection disposal errors gracefully (server might be shutting down)
         const errorMessage = error instanceof Error ? error.message : String(error);
--- a/packages/salesforcedx-lwc-language-server/src/lwcServer.ts
+++ b/packages/salesforcedx-lwc-language-server/src/lwcServer.ts
@@ -684,37 +684,11 @@ export default class Server {
       this.isDelayedInitializationComplete = true;
 
       // Configure TypeScript support now that files are loaded and context is initialized
-      // This ensures sfdx-project.json is available in the FileSystemDataProvider
-      try {
-        const hasTsEnabled = await this.isTsSupportEnabled();
-        if (hasTsEnabled) {
-          // Set connection for file operations (works in both Node.js and web)
-          this.context.setConnection(this.connection);
-          // Make tsconfig generation non-blocking to avoid connection disposal issues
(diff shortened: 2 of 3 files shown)


packages/salesforcedx-aura-language-server/src/auraServer.ts
```diff
@@ -186,7 +186,7 @@ export default class Server {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-argument
         for (const [uri, content] of Object.entries(serializedProvider.fileContents)) {
           if (typeof content === 'string') {
-            this.fileSystemProvider.updateFileContent(uri, content);
+            void this.fileSystemProvider.updateFileContent(uri, content);
```
madhur310, line 189: I see argument to make this await too, but kept it as void since without connection the method behaves synchronously. If reviewer prefers await, I can change it.

packages/salesforcedx-lightning-lsp-common/src/providers/fileSystemDataProvider.ts
```diff
@@ -17,7 +21,14 @@ export interface IFileSystemProvider {
   getFileStat(uri: string): FileStat | undefined;
   fileExists(uri: string): boolean;
   directoryExists(uri: NormalizedPath): boolean;
-  updateFileContent(uri: string, content: string): void;
+  /**
+   * Update file content from client
+   * If connection is provided, uses LSP workspace/applyEdit to create/write the file (works in both Node.js and web)
```
peternhale, line 26: @madhur310 not following the notion of an optional LSP connection.

madhur310: @peternhale yhe method is used to update server's fileSystemProvider object, it may or may not need to be written to client. For e.g. when server is populating fileSystemProvider using client's onDidOpen requests. 

So when server does indeed need to send it to client it will pass in connection object - hence optional.


The other option was to add a helper method that would create server to client applyEdit call, this seemed like the cleaner approach since server would keep its own memory upto date anyways.

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
```diff
@@ -680,14 +671,52 @@ export default class Server {
...
       this.languageService = getLanguageService({
         customDataProviders: [this.lwcDataProvider, this.auraDataProvider],
         useDefaultDataProvider: false
       });
 
       this.isDelayedInitializationComplete = true;
 
+      // Configure TypeScript support now that files are loaded and context is initialized
+      // This ensures sfdx-project.json is available in the FileSystemDataProvider
+      try {
```
peternhale: This is pretty gnarly, maybe find a way to simplify.

peternhale: I believe the same pattern exists in aura server?

packages/salesforcedx-lwc-language-server/src/typingIndexer.ts
```diff
@@ -61,20 +62,21 @@ const createNewMetaTypings = async (indexer: TypingIndexerData): Promise<void> =
...
-    indexer.fileSystemProvider.updateFileContent(filePath, typingFile.content);
-    indexer.fileSystemProvider.updateFileStat(filePath, {
+    // Update file stat first
+    indexer.fileSystemProvider.updateFileStat(typingFile.uri, {
       type: 'file',
       exists: true,
       ctime: Date.now(),
       mtime: Date.now(),
       size: typingFile.content.length
     });
```
peternhale, line 68: These can be combined, eliminating half the total iterations found in both.

packages/salesforcedx-lwc-language-server/src/typingIndexer.ts
```diff
@@ -212,14 +217,25 @@ export default class TypingIndexer {
...
   public static async create(
     attributes: BaseIndexerAttributes,
-    fileSystemProvider: IFileSystemProvider
+    fileSystemProvider: IFileSystemProvider,
+    connection?: Connection
   ): Promise<TypingIndexer> {
     const indexer = new TypingIndexer(attributes, fileSystemProvider);
+    if (connection) {
+      indexer.setConnection(connection);
+    }
```
peternhale: Delegate undefined check to setConnection.

```suggestion
    indexer.setConnection(connection);
```

packages/salesforcedx-lwc-language-server/src/lwcServer.ts
madhur310: aura doesn't do this

PR #6809, opened 2026-01-26
[W-20798320] refactor: move soql-model from soql-tooling to salesforcedx-vscode

daphne-sfdc:
### What does this PR do?
Move soql-model from soql-tooling to salesforcedx-vscode. No functionality change.

@W-20798320@

### Testing
- SOQL Builder works ✅
Screenshot 2026-01-26 at 1 09 57 PM

Screenshot 2026-01-26 at 1 11 53 PM

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/services/model.ts
+++ b/packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/services/model.ts
@@ -5,7 +5,7 @@
  *  For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  *
  */
-import { Soql } from '../../../../soql-model';
+import * as Soql from '../../../../soql-model/model/model';
 import { List, Map } from 'immutable';
 import { JsonMap } from '@salesforce/ts-types';
 import { isLikeStart, isLikeEnds, isLikeContains } from '../services/soqlUtils';
--- a/package-lock.json
+++ b/package-lock.json
@@ -492,7 +492,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.26.9.tgz",
       "integrity": "sha512-lWBYIrF7qK5+GjY5Uy+/hEgp8OJWOD/rpy74GplYRhEauvbHDeFB8t5hPOZxCZ0Oxf4Cc36tK51/l3ymJysrKw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.2",
@@ -3544,7 +3543,6 @@
       "resolved": "https://registry.npmjs.org/@effect/platform/-/platform-0.93.8.tgz",
       "integrity": "sha512-xTEy6fyTy4ijmFC3afKgtvYtn/JyPoIov4ZSUWJZUv3VeOcUPNGrrqG6IJlWkXs3NhvSywKv7wc1kw3epCQVZw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "find-my-way-ts": "^0.1.6",
         "msgpackr": "^1.11.4",
@@ -4262,7 +4260,6 @@
       "resolved": "https://registry.npmjs.org/@eslint/json/-/json-0.3.0.tgz",
       "integrity": "sha512-fdih4NZ6sGMI3YFZUqShMJ/AWvDzDrNepqDlTwRQUfgR434jmajVYIMNc2Vs2M2cHjxZpcHG6FRsrxuiwYvprA==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@humanwhocodes/momoa": "^3.2.0"
       },
@@ -4333,6 +4330,7 @@
       "deprecated": "Use @eslint/config-array instead",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "@humanwhocodes/object-schema": "^1.2.0",
         "debug": "^4.1.1",
@@ -4347,7 +4345,8 @@
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.2.tgz",
       "integrity": "sha512-3oSeUO0TMV67hN1AmbXsK4yaqU7tjiHlbxRDZOpH0KW9+CeX4bRAaX0Anxt0tx2MrpRpWwQaPwIlISEJhYU5Pw==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@humanwhocodes/config-array/node_modules/brace-expansion": {
       "version": "1.1.12",
@@ -4355,6 +4354,7 @@
       "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
@@ -4366,6 +4366,7 @@
       "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
       "dev": true,
       "license": "ISC",
+      "peer": true,
       "dependencies": {
         "brace-expansion": "^1.1.7"
       },
@@ -4401,7 +4402,8 @@
       "integrity": "sha512-ZnQMnLV4e7hDlUvw8H+U8ASL02SS2Gn6+9Ac3wGGLIe7+je2AeAOxPY+izIPJDfFDb7eDjev0Us8MO1iFRN8hA==",
       "deprecated": "Use @eslint/object-schema instead",
       "dev": true,
-      "license": "BSD-3-Clause"
+      "license": "BSD-3-Clause",
+      "peer": true
     },
     "node_modules/@humanwhocodes/retry": {
       "version": "0.4.3",
@@ -5695,7 +5697,6 @@
       "resolved": "https://registry.npmjs.org/@lwc/compiler/-/compiler-8.16.0.tgz",
       "integrity": "sha512-XaAredo9pXS9uCEHuYqB0EeJIbH3owNpT8zToWdG2sGolK091WQ8hFHB2twkBXvsy+YqbqEMbwdKH3WvC9sWEw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/core": "7.26.9",
         "@babel/plugin-transform-async-generator-functions": "7.26.8",
@@ -5729,8 +5730,7 @@
       "version": "8.16.0",
       "resolved": "https://registry.npmjs.org/@lwc/errors/-/errors-8.16.0.tgz",
       "integrity": "sha512-cglxtbp500a2F/WNNDOkwUky2a6KzqPJ4syr6yhuOS5ifgPN2wbuiMJHRzD7Q6EW/e0dHtZxX9Xlyf49JjjzWQ==",
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@lwc/features": {
       "version": "2.37.3",
@@ -5923,7 +5923,6 @@
       "integrity": "sha512-kWc7L0fw1xwvI0zi8OKVBuxRVefwGOrKSQMvrQ3dW+bIIavBY3/NpXmpjMy7bQnLgwgzWQZ8TlM57YHpHNHz4w==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.8.3",
         "@babel/generator": "^7.9.0",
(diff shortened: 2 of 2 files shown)


packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -7,9 +7,36 @@
 import { build } from 'esbuild';
 import copy from 'esbuild-plugin-copy';
 import { nodeConfig } from '../../scripts/bundling/node.mjs';
+import { cpSync, mkdirSync } from 'fs';
+import * as path from 'path';
+
+// Plugin to rewrite soql-common imports to be relative to dist directory
+const rewriteSoqlCommonImports = {
```
mshanemc, line 13: Is this eventually avoidable?  If the various consumers of this all `import` it the normal way, it should get into the bundle, just like any other file, regardless of its location in source, right?

What's keeping that from happening?

packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -7,9 +7,36 @@
 import { build } from 'esbuild';
 import copy from 'esbuild-plugin-copy';
 import { nodeConfig } from '../../scripts/bundling/node.mjs';
+import { cpSync, mkdirSync } from 'fs';
+import * as path from 'path';
```
mshanemc: please set your rules to have unused imports

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/services/model.ts
```diff
@@ -5,7 +5,7 @@
  *  For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  *
  */
-import { Soql } from '@salesforce/soql-model';
+import { Soql } from '../../../../soql-model';
```
mshanemc: suggestion: since we're not treating these as separate modules/packages, let's import them from their real location and not the barrel (then we can get rid of the barrel!)

packages/salesforcedx-vscode-soql/src/soql-model/serialization/deserializer.ts
```diff
@@ -0,0 +1,740 @@
...
+import * as Parser from '../../soql-common/soql-parser.lib/generated/SoqlParser';
+import { SoqlParserListener } from '../../soql-common/soql-parser.lib/generated/SoqlParserListener';
+import { parseHeaderComments } from '../../soql-common/soqlComments';
+import { Messages } from '../messages/messages';
+import * as Impl from '../model/impl';
+import { HeaderCommentsImpl } from '../model/impl/headerCommentsImpl';
+import * as Soql from '../model/model';
+import { SoqlModelUtils } from '../model/util';
+
+export class ModelDeserializer {
```
mshanemc: this is a silly class when all we needed was a deserialize fn that accepts the soqlSyntax.

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* eslint-disable no-param-reassign, @typescript-eslint/consistent-type-assertions */
+import * as Impl from './impl';
+import { AndOr, Condition } from './model';
+
+
+export namespace SoqlModelUtils {
```
mshanemc: most of this looks unused; the recursion hides the unusedness.

you can do a "find references" to be sure on these.

packages/salesforcedx-vscode-soql/src/soql-model/model/impl/index.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
```
mshanemc: empty lines...huh.  I'd get rid of this file and have the consumers import from the original location

packages/salesforcedx-vscode-soql/src/soql-model/model/impl/queryImpl.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.where = where;
+    this.with = soqlwith;
+    this.groupBy = groupBy;
+    this.orderBy = orderBy;
+    this.limit = limit;
+    this.offset = offset;
+    this.bind = bind;
+    this.recordTrackingType = recordTrackingType;
+    this.update = update;
+  }
```
mshanemc, line 33: eww. why not use constructor property declarations to avoid the triplication?

packages/salesforcedx-vscode-soql/src/soql-model/model/impl/orderByExpressionImpl.ts
```diff
@@ -0,0 +1,33 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as Soql from '../model';
+import { SoqlModelObjectImpl } from './soqlModelObjectImpl';
+
+export class OrderByExpressionImpl extends SoqlModelObjectImpl implements Soql.OrderByExpression {
+  public field: Soql.Field;
```
mshanemc: use constructor props

packages/salesforcedx-vscode-soql/src/soql-model/messages/messages.ts
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// prettier-ignore
+/* eslint-disable @typescript-eslint/naming-convention */
+export namespace Messages {
```
mshanemc: probably let's do this like other messages/18n files.

also, we shouldn't be using `namespace`

packages/salesforcedx-vscode-soql/src/soql-builder-ui/package.json
```diff
@@ -8,7 +8,6 @@
     "@lwc/eslint-plugin-lwc": "^0.12.2",
     "@lwc/jest-preset": "^7.0.0",
     "@salesforce/eslint-config-lwc": "^0.11.0",
```
mshanemc, line 10: I'm very confused by the structure of this

if all the packages are going to be in vscode-soql/src why do they have their own package.json?

Is that why that weird path resolution script needed to exist?

I'd prefer we go into one of
- packages in /packages like all the other packages
- src in packages/vscode-soql without subpackages

packages/salesforcedx-vscode-soql/src/soql-builder-ui/webpack.config.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
       fs: false
     },
     alias: {
```
mshanemc, line 11: maybe not for this PR, but I'd like to have all our webpack=>eslint

and if this doesn't stay as a standalone package, then it should probably be in the extensions esbuild (to bundle the soql-builder-ui)

packages/salesforcedx-vscode-soql/src/soql-common/soql-parser.lib/dateFormula.test.js
```diff
@@ -0,0 +1 @@
...90324NakPdd','806831BAFzRP','isDateFormula','1899920YTJkHm','__esModule'];const a3_0x3f2d=function(_0x458e57,_0x19df78){_0x458e57=_0x458e57-0x1e5;let _0x252c69=a3_0x252c[_0x458e57];return _0x252c69;};const a3_0x47d64f=a3_0x3f2d;(function(_0x3953c8,_0x1da90c){const _0x4d2117=a3_0x3f2d;while(!![]){try{const _0x25880a=-parseInt(_0x4d2117(0x1f4))+parseInt(_0x4d2117(0x1ef))+-parseInt(_0x4d2117(0x1f3))+parseInt(_0x4d2117(0x1f1))+-parseInt(_0x4d2117(0x1ec))*-parseInt(_0x4d2117(0x1ea))+parseInt(_0x4d2117(0x1f0))*parseInt(_0x4d2117(0x1eb))+-parseInt(_0x4d2117(0x1e5));if(_0x25880a===_0x1da90c)break;else _0x3953c8['push'](_0x3953c8['shift']());}catch(_0x477b65){_0x3953c8['push'](_0x3953c8['shift']());}}}(a3_0x252c,0xb18ad));Object[a3_0x47d64f(0x1e8)](exports,a3_0x47d64f(0x1e6),{'value':!![]});const dateFormula_1=require('./dateFormula');describe(a3_0x47d64f(0x1ee),()=>{const _0x4e2646=a3_0x47d64f;it('false\x20if\x20input\x20is\x20not\x20a\x20DateFormula',()=>{const _0x254c50=a3_0x3f2d;expect(dateFormula_1[_0x254c50(0x1f5)](_0x254c50(0x1f2)))['toBeFalsy']();}),it(_0x4e2646(0x1ed),()=>{const _0x2d0d31=_0x4e2646;expect(dateFormula_1[_0x2d0d31(0x1f5)](_0x2d0d31(0x1e7)))[_0x2d0d31(0x1e9)]();});});
```
mshanemc: what if we didn't keep the test files?

packages/salesforcedx-vscode-soql/src/soql-model/model/model.ts
```diff
@@ -0,0 +1,285 @@
...
+  Number = 'NUMBER',
+  String = 'STRING',
+}
+
+export interface Literal extends CompareValue {
+  type: LiteralType;
+  value: string;
+}
+
+export interface Condition extends SoqlModelObject {
```
mshanemc, line 198: is this needed?

packages/salesforcedx-vscode-soql/src/soql-model/model/impl/soqlModelObjectImpl.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as Soql from '../model';
+
+export abstract class SoqlModelObjectImpl {
+  public getSyntaxOptions(options?: Soql.SyntaxOptions): Soql.SyntaxOptions {
```
mshanemc, line 11: I see this used in 1 place (queryImpl)
and it passes in the options.  
seems pointless

packages/salesforcedx-vscode-soql/src/soql-model/errorTypes.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/** Re-export ErrorType enum for use in browser contexts without pulling in node dependencies */
```
mshanemc, line 8: I don't understand this comment.

packages/salesforcedx-vscode-soql/src/soql-model/model/model.ts
```diff
@@ -0,0 +1,285 @@
...
+  Phone = 'phone',
+  Picklist = 'picklist',
+  Reference = 'reference',
+  String = 'string',
+  TextArea = 'textarea',
+  Time = 'time',
+  Url = 'url',
+}
+
+export enum UiOperatorValue {
```
mshanemc: some of these enums seem pointless.  should be string unions

packages/salesforcedx-vscode-soql/src/soql-model/validators/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export { splitMultiInputValues } from './inputUtils';
```
mshanemc: remove all the barrels

packages/salesforcedx-vscode-soql/src/soql-model/model/model.ts
daphne-sfdc: Yes it is!  But it wasn't implemented cleanly in the original code that I imported from **soql-tooling,** so the interfaces (now changed to types) were not type-safe: I fixed that in commit [11f3656](https://github.com/forcedotcom/salesforcedx-vscode/pull/6821/commits/11f36562f975b27f282c23a29ee5625a4fdc6417) by adding a `kind` discriminant property.

The purpose of these types is to have each term in the SOQL query be a different type - for example, `SELECT` is a type while `=` is another type.  If you remove the `Condition` type, that means you'd need to replace every usage of `Condition` with `SoqlModelObject`, which is unsafe because every type extends `SoqlModelObject` - therefore you could place *anything* in place of `Condition`.

In the original implementation, `Condition` only serves the purpose of code clarity - you *could* use `SoqlModelObject` (or anything that extends it) in place of `Condition`, and the compiler wouldn't complain.  By adding the discriminant properties, I added type safety and prevented that outcome.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/package.json
daphne-sfdc: This file is needed because **soql-builder-ui** currently uses `lwc-services`.  Once we migrate off `lwc-services` in [W-16009363](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001uVgbwYAC/view), this file can be removed.

packages/salesforcedx-vscode-soql/esbuild.config.mjs
daphne-sfdc: Right now Lines 37-39 have
```
const commonConfig = {
  external: ['vscode', './soql-common/*']
};
```
which bundles `soql-common` separately.  This is done because `soql-common` contains `soql-parser.lib`, which is directly copied from **soql-tooling**, and which is not bundle-able as part of salesforcedx-vscode-soql.

In order to remove this script, we'd need to handle `soql-parser.lib` separately in order to make `soql-common` bundle-able instead of externalized.  I'm going to leave this one for now.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/package.json
daphne-sfdc: Not related to the weird path resolution script - see explanation above for why that exists.

PR #6821, opened 2026-01-30
[W-20798320][W-21078900][W-21179843][W-21179797] refactor: move soql-model from soql-tooling to salesforcedx-vscode + cleanup of all the code migrated from soql-tooling

daphne-sfdc:
### What does this PR do?
Move soql-model from soql-tooling to salesforcedx-vscode. No functionality change.

@W-20798320@, @W-21078900@, @W-21179843@, @W-21179797@

### Testing
- SOQL Builder works ✅
Screenshot 2026-01-26 at 1 09 57 PM

Screenshot 2026-01-26 at 1 11 53 PM


packages/salesforcedx-vscode-soql/src/soql-builder-ui/jestSetup/setupTests.ts
```diff
@@ -71,9 +71,9 @@ export class VsCodeApi {
...
 }
 export const vscodeInstance = new VsCodeApi();
-export function acquireVsCodeApi(): VsCodeApi {
+export const acquireVsCodeApi = (): VsCodeApi => {
   return vscodeInstance;
-}
+};
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 globalThis.acquireVsCodeApi = acquireVsCodeApi;
```
mshanemc, line 79: this test mock raises some interesting questions.  What's the purpose of `standaloneMessageService` and the abstraction of `messageServiceFactory` outside of vscode?

I wonder if someone wrote this in a way to be used outside of vscode and then nobody ever did?

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/services/globals.ts
```diff
@@ -6,28 +6,28 @@
...
+};
 
-export function getBodyClass(): string | null {
+export const getBodyClass = (): string | null => {
   return window.document.body.getAttribute('class');
-}
+};
 
-export function getLocalStorage(): Storage {
+export const getLocalStorage = (): Storage => {
```
mshanemc: if we do decide to drop the isomorhpic stuff `standaloneMessageService` there's probably a lot we can get rid of

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -107,6 +107,15 @@
...
+      "command": "shx mkdir -p out/src/soql-common && shx cp -R src/soql-common/soql-parser.lib out/src/soql-common/soql-parser.lib",
+      "files": [
+        "src/soql-common/soql-parser.lib/**"
+      ],
+      "output": [
+        "out/src/soql-common/soql-parser.lib/**"
+      ]
+    },
     "build:soql-builder-ui": {
       "command": "cd src/soql-builder-ui && npm install && export NODE_OPTIONS=--openssl-legacy-provider && npx lwc-services build -w webpack.config.js -m production",
```
mshanemc, line 116: let's find a way NOT to treat this as a package (ex: move devDeps to this pjson, make the webpack output go to dist for vscode-soql.  As is, it builds to `soql-builder-ui/dist` and then vscode-soql's esbuild has to copy that to where it really needs to go instead of producing in the correct place all the time.  

```
copy({
      assets: {
        from: [`./src/soql-builder-ui/dist/**`],
        to: ['./soql-builder-ui']
      }
    }),
```

then maybe we won't need all the rewriteSoqlCommonImports stuff?

packages/salesforcedx-vscode-soql/src/soql-model/validators/currencyValidator.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { messages } from '../messages/i18n';
+import { FloatValidator } from './floatValidator';
+import { ValidateResult } from './validator';
+
+export class CurrencyValidator extends FloatValidator {
```
mshanemc, line 11: what's the point of extends if it's only got one method and it's being overridden?

packages/salesforcedx-vscode-soql/test/jest/soql-model/validators/integerValidator.test.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { messages } from '../../../../src/soql-model/messages/i18n';
+import { SObjectFieldType } from '../../../../src/soql-model/model/model';
+import { IntegerValidator } from '../../../../src/soql-model/validators/integerValidator';
+
+describe('IntegerValidator should', () => {
+  const validator = new IntegerValidator({ type: SObjectFieldType.Integer });
```
mshanemc, line 13: I really dislike the design pattern stuff here.

validators only have 1 public method (validate).  none of them use `type` option (it's only used as a discriminator by the factory thing)

I think only picklists use the other 2 inputs.

No changes needed now, I'm going to put this on my someday/maybe list of things to rewrite

packages/salesforcedx-vscode-soql/src/soql-model/serialization/serializer.ts
```diff
@@ -0,0 +1,19 @@
...
+import { SoqlModelObject, SyntaxOptions } from '../model/model';
+
+export class ModelSerializer {
+  protected model: SoqlModelObject;
+
+  constructor(model: SoqlModelObject) {
+    this.model = model;
+  }
+  public serialize(options?: SyntaxOptions): string {
+    return this.model.toSoqlSyntax(options);
```
mshanemc, line 17: nothign passes in `options` to serialize  So all of the `syntaxOptions` stuff might be pointless?

packages/salesforcedx-vscode-soql/src/soql-model/model/model.ts
```diff
@@ -0,0 +1,296 @@
...
+  | 'LIKE_END'
+  | 'LIKE_CONTAINS'
+  | 'IN'
+  | 'NOT_IN'
+  | 'INCLUDES'
+  | 'EXCLUDES';
+
+export type SoqlModelObject = {
+  errors?: ModelError[];
+  toSoqlSyntax(options?: SyntaxOptions): string;
```
mshanemc, line 85: the inheritance here is goofy.  Only Query uses the 1 syntaxOptions prop (indent), so then everything else that has a `toSoqlSyntax` on it has to ignore the unused param.

Maybe the `indent` option goes on Query instead? Or maybe `2` is just a constant 

AI says
```
No. Nothing sets the indent property.
No assignments: no .indent = found
Only default instantiation: new SyntaxOptions() in queryImpl.ts line 33 uses the default indent = 2
All calls pass undefined or nothing: serialize() and toSoqlSyntax() calls don't pass a SyntaxOptions parameter, so they use the default
The indent property always remains 2; it's never customized.
```

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,143 @@
...
+    const [left, ...rest] = conditions;
+    return new AndOrConditionImpl(left, andOr as AndOr, arrayToSimpleGroup(rest, andOr));
+  }
+};
+
+const isSimpleGroup = (condition: Condition, andOr?: AndOr): boolean => {
+  // a simple group is a condition that can be expressed as an ANY or ALL group of conditions
+  // ANY: simple conditions all joined by OR
+  // ALL: simple conditions all joined by AND
+  condition = stripNesting(condition);
```
mshanemc, line 111: be really careful with modifying the inputs of a fn (it causes the side effect of mutating whatever passed the reference in).

sometimes, that's intentional (pass to a function and have it perform some operation intentionally)
but it often leads to bugs because something calls `isSimpleGroup` and then tries to do something else with the `condition` but doesn't realized that it's been modified to be whatever `stripNestinged` returned! 

`eslint-disable no-param-reassign` is dangerous.  If you really need it and `isSimpleGroup` is provably correct, then I'd put it at the line level instead of a whole file to prevent growth of this bad pattern.

A better pattern would be `unnestedCondition = stripNesting(condition)` and then use that to get the value you need.

mshanemc: but also rewrite stripNesting to not mutate its param either.

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,143 @@
...
+  let andOr: AndOr | undefined;
+  if (condition instanceof AndOrConditionImpl) {
+    conditions = conditions.concat(simpleGroupToArray(condition.leftCondition).conditions);
+    conditions = conditions.concat(simpleGroupToArray(condition.rightCondition).conditions);
+    andOr = condition.andOr;
+  } else {
+    conditions.push(condition);
+  }
+  return { conditions, andOr };
+};
```
mshanemc, line 89: instead of let, and undefined, and mutating arrays, why not do it immutably

  return condition instanceof AndOrConditionImpl
    ? {
        andOr: condition.andOr,
        conditions: [condition.leftCondition, condition.rightCondition]
          .map(simpleGroupToArray)
          .flatMap(item => item.conditions)
      }
    : { conditions: [condition] };

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const arrayToSimpleGroup = (conditions: Condition[], andOr?: AndOr): Condition => {
```
mshanemc, line 91: I think there's a common type for {conditions: Condition[], andOr?: AndOr}.

It should be a return type for some fns, and an input to others.

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,143 @@
...
+import { NestedConditionImpl } from './impl/nestedConditionImpl';
+import { UnmodeledSyntaxImpl } from './impl/unmodeledSyntaxImpl';
+import { AndOr, Condition } from './model';
+
+const stripNesting = (condition: Condition): Condition => {
+  while (condition instanceof NestedConditionImpl) {
+    condition = condition.condition;
+  }
+  return condition;
+};
```
mshanemc, line 22: the non-mutation version of this would be recursive like

=> condition instanceof NestedConditionImpl ? stripNesting(condition) : condition

packages/salesforcedx-vscode-soql/src/soql-model/model/util.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (conditions.length === 0) {
+    throw Error('no conditions');
+  }
+
+  if (conditions.length === 1) {
+    return conditions[0];
+  } else {
+    const [left, ...rest] = conditions;
+    return new AndOrConditionImpl(left, andOr as AndOr, arrayToSimpleGroup(rest, andOr));
+  }
```
mshanemc, line 104: why is the type assertion necessary?  the fn accepts `undefined` for andOr but the `AndOrConditionImpl` does not.

if I remove the assertion, it seems that the throws have narrowed it to be andOr?

packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -7,9 +7,35 @@
...
+      }
+
+      return { contents, loader: 'js' };
+    });
+  }
+};
 
 const commonConfig = {
-  external: ['vscode']
+  external: ['vscode', './soql-common/*']
```
mshanemc: is it possible to get rid of this, so that it just gets bundled into whatever is using it?  Or is it because of some usage inside the LWC that esbuild can't do it?

packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -41,3 +76,15 @@ await build({
   entryPoints: ['../../node_modules/@salesforce/soql-language-server/lib/server.js'],
   outfile: './dist/server.js'
 });
+
+// Manually copy soql-parser.lib directory since esbuild-plugin-copy doesn't handle nested directories well
```
mshanemc: it works fine other places.  I think an AI did this 

the real problem is that tsconfig `  "include": ["**/*.ts", "../../src/**/*.ts"]`

won't find the stuff in `soql-parser.lib` that's already compiled so it doesn't end up in `out` like other things

---
some options
1. change the tsconfig to do that
2. change compile script to copy that stuff from src => out as part of compile
3. go find wherever the real source TS came from and use that in our src instead of the compiled code so that it follows the normal path

eslint.config.mjs
```diff
@@ -56,6 +58,28 @@ export default [
       '**/playwright-report/**'
     ]
   },
+  {
```
mshanemc, line 62: this is great, definitely cleaned up some script stuff!

packages/salesforcedx-vscode-soql/src/soql-builder-ui/jestSetup/setupTests.ts
jonnyhork: @mshanemc The original plan for SOQL builder was for the app to be portable anywhere we need it, not just VS Code. That never materialized though. The standalone was used as a dev tool for faster iteration.
https://github.com/forcedotcom/soql-tooling-standalone

packages/salesforcedx-vscode-soql/src/soql-model/validators/currencyValidator.ts
daphne-sfdc: Line 13 has `const result = super.validate(input);`.  So we **do** use the validation logic from `FloatValidator` to short-circuit the logic - only if the value is *not* a float do we fall back for additional checking on Line 18.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/jestSetup/setupTests.ts
daphne-sfdc: @jonnyhork Then are we going to keep the standalone, or just get rid of it and keep only support for VSCode?  If the latter, I can follow Shane's suggestion and get rid of this code?

mshanemc: let's get rid of it.  We can also do it later (if we ever wanted to) using polyfills

jonnyhork: Agree, no need to keep anything around that is related to standalone stuff. That was expressed as a possibility by Product many many years ago. It's only ever going to be VS Code now.

packages/salesforcedx-vscode-soql/package.json
daphne-sfdc: I tried to intercept this and send the build output directly to `salesforcedx-vscode-soql/dist`, but that did not work.  It produced these 2 errors when I ran **SFDX: Create Query in SOQL Builder**:
Screenshot 2026-02-04 at 4 12 29 PM
Screenshot 2026-02-04 at 4 29 32 PM

The reason this doesn't work is the same reason **soql-builder-ui** requires its own **package.json**: because it uses `lwc-services` and must be compiled separately during the bundling process rather than compiled with the extension during compile time.

I'll revisit this when we migrate off `lwc-services` in [W-16009363](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001uVgbwYAC/view).

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -6,10 +6,10 @@
  */
 
 import type { Connection } from '@salesforce/core';
+import * as soqlComments from '@salesforce/soql-common/soqlComments';
```
mshanemc, line 9: how does this work?  where is it coming from?
doesn't that just get it from the old npm package?

I thought the idea was to have `/packages/soql-common`

daphne-sfdc: It's local.  In **packages/salesforcedx-vscode-soql/tsconfig.json** there's this mapping of `@salesforce/soql-common` to the local code.
```
    "paths": {
      "@salesforce/soql-common/*": ["./src/soql-common/*"],
      "@salesforce/soql-model/*": ["./src/soql-model/*"]
    }
```

packages/salesforcedx-vscode-soql/src/soql-model/model/model.ts
jonnyhork: https://github.com/forcedotcom/soql-tooling/pull/9#discussion_r476685412

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/where/where.ts
```diff
@@ -25,7 +25,7 @@ export default class Where extends LightningElement {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   @api public sobjectMetadata: any;
   @track public _conditionsStore: JsonMap[] = [];
-  public _andOr = AndOr.AND;
+  public _andOr: Soql.AndOr = 'AND';
```
jonnyhork: I don't think I understand this. Why delete an enum, import another one from another file and then use hard coded strings throughout the file instead of the enum? Where you'd only have to change the value once? Isn't the whole point so that we don't have to type 'AND' everywhere.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/cssCommon/cssCommon.css
```diff
@@ -34,15 +34,6 @@
...
 }
 
-/* TODO: we should change the name of the var
-the width of all the children should be
-default to the width of the parent.
-*/
-.section_selection-container {
-  min-width: var(--soql-input-width);
-  max-width: var(--soql-input-width);
-}
```
jonnyhork, line 45: This is the same selector as above but the rules are different, this would likely be a regression, and I think a comment you deleted even references why we want this. Don't forget there a huge amount of fields to choose from and some of them can be very long, especially custom fields. So unless you manually tested a variety of field lengths, this is probably intended to keep the UI from jumping around too much or something like that.

[chore: remove unneeded TODO comments in whereModifierGroup.css](https://github.com/forcedotcom/salesforcedx-vscode/pull/6821/changes/594600ba8cb21371642e24b9376a4814ae468f81)

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -211,31 +209,31 @@ export default class WhereModifierGroup extends LightningElement {
...
   public isMultipleValueOperator(operatorValue: string): boolean {
     return (
-      operatorValue === Soql.UiOperatorValue.IN ||
-      operatorValue === Soql.UiOperatorValue.NOT_IN ||
-      operatorValue === Soql.UiOperatorValue.INCLUDES ||
-      operatorValue === Soql.UiOperatorValue.EXCLUDES
+      operatorValue === 'IN' ||
+      operatorValue === 'NOT_IN' ||
+      operatorValue === 'INCLUDES' ||
+      operatorValue === 'EXCLUDES'
```
jonnyhork, line 215: I'm curious why you would want to get rid of an enum in favor of hard coding strings in various files?

packages/salesforcedx-vscode-soql/src/soql-model/serialization/serializer.ts
jonnyhork: IIRC there was plans to let the user define some syntax options like indentation, etc.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
mshanemc: TS makes this safe, gives the same autocomplete as the enums.  Enums should be avoided in TS.

mshanemc: the types compile away for runtime, the enums have to be compiled into some weird objects

jonnyhork: Thats fine, I used the term enum too loosely here (even thought this example happens to be one) but the real smell to me was still about using direct string values rather than converting the enum to and object with types.

`export type UiOperatorValue = typeof UiOperatorValue[keyof typeof UiOperatorValue];` or something like that.

mshanemc: given that the string is typed as a string union, what problem does it create that makes you think "smell"?

jonnyhork: The parameter is typed as string, so TypeScript happily accepts 'OUT' and 'NOT_REAL' -- neither of which exist in the UiOperatorValue union or in operatorOptions. No compile error, just a silent runtime bug.

I like the type-safety of pre-defined pointers
`Soql.UiOperatorValue.EXCLUDES`. Currently I can change these to anything and introduce a bug more easily. Additionally, if I need to add or change the values I may have to make that change in multiple places rather than once in the model, or duplicate the values if I want to use it anywhere else in the code base.

[Why not use this? ](https://github.com/forcedotcom/salesforcedx-vscode/pull/6821/changes?mode=single#diff-a9552e2a3100a6e205ef61adaa4fab7464eeee4af371a3eb19efc25dacb54336R66)

Image

mshanemc: ok, I think I understand.  that method (should only accept UiOperatorValue, not string).  Same with _selectedOperator, it should be typed more specifically than `string`.  Have good types, then use them everywhere you can!

if we need to do runtime validation of these (ex: from user input where we can't guarantee that it is a valid member of the type) then it can be helpful to have an exported array of the value like

```
export const operators = ['Eq' ... ]
export type UiOperatorValue = (typeof operators)[number]
```

etc

jonnyhork: > Have good types, then use them everywhere you can!

100%

A lot of this repo reflects the lack of public types for the SOQL lang or something that would be maintained by SOQL people and we could consume and enforce the contract in the form of types. So what you see in soql-tooling is an attempt to model valid SOQL ourselves, at least the parts the builder supports and help developers do the same, also not have to ask "what are all the operator options I need again? Is XYZ valid?" 

It's not an ideal paradigm but it felt like the best option at the time.

[Same kind of thinking here](https://github.com/forcedotcom/salesforcedx-vscode/pull/6821/changes#diff-72975a7e0cc97048927b7444de58b2998849e2d832b04af7787261a356dbb988L315)

mshanemc: kinda surprised there's not an easier way to do that from the antlr grammars...like that should tell us, I think, Is XYZ valid.

jonnyhork: Would be cool to generate @types/soql-lang or something from the grammar

jonnyhork: maybe this [antlr4ts](https://github.com/tunnelvisionlabs/antlr4ts) with

```"scripts": {
  "generate-soql": "antlr4ts -visitor path/to/SOQL.g4"
}
```
🤷

PR #6824, opened 2026-01-30
[W-21112200] ci: remove Slack notifications from all our GHA workflows

daphne-sfdc:
### What does this PR do?
Our GHA workflows currently send notifications to #ide-exp-vscode, #ide-exp-releases, and #e2e-tests-results.  However, we've figured out during our Parking Lot discussion that nobody on the team actually uses these notifications.  So this PR removes the unused notifications to remove a possible moving piece and point of failure.

@W-21112200@


.github/workflows/generateXsdForMetadataTypes.yml
```diff
@@ -85,8 +85,4 @@ jobs:
             --body "This PR updates the \`salesforce_metadata_api_common.xsd\` and \`metadata_types_map_scraped.json\` files with the latest metadata type definitions. Generated by the [Generate XSD for Metadata Types](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) workflow at ${{ steps.timestamp.outputs.value }}." \
             --base develop \
             --head automated/update-xsd-${{ steps.timestamp.outputs.value }}
-
-          # Merge the PR immediately
-          gh pr merge automated/update-xsd-${{ steps.timestamp.outputs.value }} \
-            --squash \
-            --delete-branch
+        continue-on-error: true
```
CristiCanizales, line 88: was this intentional?

daphne-sfdc: Yes that's intentional.
- We originally decided that we wanted the PR to automatically merge.  However, I noticed in the past few runs that sometimes the PR had errors (the workflow ended with fewer metadata types than what we had before), so I'm now playing it safe and manually inspecting each PR to ensure we're not removing support for legitimate metadata types.
- I added `continue-on-error: true` because apparently that PR creation step fails, even though the PR is properly created.  See [this week's run](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/21563537000/job/62131244109).  The workflow should succeed after the PR is created.

PR #6833, opened 2026-02-05
test: launch run:web pre-authed W-20906136

mshanemc:
### What does this PR do?
see docs/QA for explanation of this

@W-20906136@

### DIFF
--- a/packages/salesforcedx-vscode-apex-testing/package.json
+++ b/packages/salesforcedx-vscode-apex-testing/package.json
@@ -87,6 +87,22 @@
         "dist"
       ]
     },
+    "vscode:bundle:local": {
+      "command": "node ./esbuild.config.mjs",
+      "dependencies": [
+        "compile",
+        "../salesforcedx-vscode-services:vscode:bundle:local"
+      ],
+      "files": [
+        "esbuild.config.mjs",
+        "../../scripts/bundling/node.mjs",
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-apex-testing/package.json
```diff
@@ -138,7 +138,7 @@
       "service": true,
       "dependencies": [
         "vscode:bundle",
```
CristiCanizales:
```suggestion
        "vscode:bundle:local",
```
don't we need local here too?

.github/workflows/templatesE2E.yml
CristiCanizales, line 1: why are we getting rid of this one?

packages/salesforcedx-vscode-apex-testing/package.json
mshanemc: I added apex-testing to them.

.github/workflows/templatesE2E.yml
mshanemc: it was never used.  I tried something once, on a branch, and to do it I needed a workflow on `main` and then it didn't work out so I had this left over

CristiCanizales (approved):
Thanks for working on this! Definitely better for development/testing time so we don't need to have workarounds to get an org automatically authed.

:shipit:

PR #6842, opened 2026-02-06
[W-21198602] fix: run all tests now runs fine for large orgs

CristiCanizales:
### What does this PR do?
- Run All Tests from Test Controller works well for large orgs
- Tests results are cleared on org switch
- Better messaging when running empty Apex Test Suites

@W-21198602@


peternhale (changes requested):
See #6845

Issue #6787, opened 2026-01-20
Apex Interactive Debugger Throws Error and Fails to Launch


m-mcmaster:
### Summary

When launching the Apex Interactive Debugger according to this [https://developer.salesforce.com/docs/platform/code-builder/guide/interactive-debugger.html](https://developer.salesforce.com/docs/platform/code-builder/guide/interactive-debugger.html) the following error is thrown:

"An unexpected error occurred while launching the debugger session.  Refer to the Debug Console for details."

The Debug Console has the following output:

  Connected to Streaming API channel /systemTopic/ApexDebuggerSystemEvent.
  Connected to Streaming API channel /systemTopic/ApexDebuggerEvent.
  Command returned the following error: 
  TypeError: e.indexOf is not a function

### Steps To Reproduce:

1.  Create a new SFDX project.

1.  Add a simple Apex class and deploy to your org.

1. Set a breakpoint

1. Add a launch configuration for live debugging to launch.json per [https://developer.salesforce.com/docs/platform/code-builder/guide/interactive-debugger.html](https://developer.salesforce.com/docs/platform/code-builder/guide/interactive-debugger.html)

1. Start the debugger.

### Expected result

No error and the debugger should start.

### Actual result

The following error is thrown:

"An unexpected error occurred while launching the debugger session.  Refer to the Debug Console for details."

The Debug Console has the following output:

  Connected to Streaming API channel /systemTopic/ApexDebuggerSystemEvent.
  Connected to Streaming API channel /systemTopic/ApexDebuggerEvent.
  Command returned the following error:
  TypeError: e.indexOf is not a function

### Additional information

This is failing with multiple users and orgs.  Happens in Agentforce Vibes environment as well.

Salesforce Extension Version in VS Code: 65.15.2
Salesforce CLI Version: @salesforce/cli/2.119.8 win32-x64 node-v22.21.1
OS and version: Windows 11: 22631.6491
VS Code version: 1.1.08.2
Most recent version of the extensions where this was working: Don't know

PR #6846, opened 2026-02-10
[W-21203023] fix: e indexof is not a function error

CristiCanizales:
### What does this PR do?
- Fixes apex debugger so `e.indexOf is not a function` is not thrown and the debugging session can start

@W-21203023@

[User confirmed the fix works](https://github.com/forcedotcom/salesforcedx-vscode/issues/6787#issuecomment-3884959071)

### DIFF
--- a/packages/salesforcedx-apex-debugger/package.json
+++ b/packages/salesforcedx-apex-debugger/package.json
@@ -44,7 +44,6 @@
   },
   "scripts": {
     "bundle:debugger": "wireit",
-    "vscode:bundle": "wireit",
     "compile": "wireit",
     "lint": "wireit",
     "clean": "shx rm -rf node_modules out dist coverage .nyc_output .wireit .eslintcache",
@@ -82,11 +81,6 @@
       "output": [
         "dist/apexDebug.js",
         "dist/apexDebug.js.map"
-      ]
-    },
-    "vscode:bundle": {
-      "dependencies": [
-        "bundle:debugger"
       ]
     },
     "test": {


packages/salesforcedx-apex-debugger/src/extractJsonObject.ts
```diff
@@ -5,7 +5,13 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 export const extractJsonObject = (str: string): any => {
-  const jsonString = str.substring(str.indexOf('{'), str.lastIndexOf('}') + 1);
+let stringValue = str;
```
mshanemc: what was coming in that wasn't a string but could be cast as one that this solves?

mshanemc: indentation looks odd

packages/salesforcedx-apex-debugger/package.json
```diff
@@ -70,15 +71,22 @@
...
+        "out/**",
+        "../../scripts/bundling/bundle-debugger.mjs"
       ],
       "output": [
-        "dist/apexDebug.js"
+        "dist/apexDebug.js",
+        "dist/apexDebug.js.map"
+      ]
+    },
+    "vscode:bundle": {
```
mshanemc: why do we need both `bundle:debugger` and `vscode:bundle` ?

packages/salesforcedx-apex-debugger/src/adapter/apexDebug.ts
```diff
@@ -1369,7 +1369,7 @@ export class ApexDebug extends LoggingDebugSession {
         break;
       }
     }
-    this.myStreamingService.markEventProcessed(message.sobject.Type, message.event.replayId);
+    void this.myStreamingService.markEventProcessed(message.sobject.Type, message.event.replayId);
```
mshanemc: not sure I understand void here.  `markEventProcessed` isn't async, right?

packages/salesforcedx-apex-debugger/package.json
CristiCanizales: so it runs when running general vscode:bundle

mshanemc: why do we need `bundle:debugger` instead of just having `vscode:bundle` do all the stuff that's currently in `bundle:debugger` ?

mshanemc: and then vscode-apex-debugger could have a dependency on the vscode:bundle from apex-debugger?

scripts/bundling/bundle-debugger.mjs
mshanemc: seems fine, maybe belongs in apex-debugger/scripts.  doesn't feel repo-level

packages/salesforcedx-apex-debugger/package.json
CristiCanizales: yep, just did that. Better like that. Thank you!

PR #6847, opened 2026-02-11
fix: Ensure WorkspaceContext.getConnection() waits for WorkspaceContext.initialize() to complete @W-21178020

floralan:
### Background            
This resolves an issue in AFV where the 'no-default-org' view was displaying incorrectly, which is discussed in Slack channel #ide-experience-team

The AFV team uses `WorkspaceContext.getConnection()` at extension startup to retrieve org information. However, if
  `getConnection()` is called before `WorkspaceContext.initialize()` completes, it throws an error because the username has not been set yet. Which causes AFV to display a 'no-default-org' view incorrectly to users when an valid org is connected.

### Testing

  Added comprehensive unit tests covering:
  1. ✅ Initialization promise is created and stored
  2. ✅ Multiple initialize() calls only trigger initialization once (idempotent)
  3. ✅ getConnection() waits for initialization when called concurrently
  4. ✅ getConnection() works when called before initialize()
  5. ✅ getConnection() works when called after initialize() completes
  6. ✅ Multiple concurrent getConnection() calls during initialization all succeed

  All tests pass with 100% coverage of the new logic.

###  Testing Instructions

  1. Verify AFV extension can successfully call getConnection() at startup
  2. Confirm no `No default org is set. Run "SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org" to set one` errors occur during AFV activation

@W-21178020@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/packages/salesforcedx-vscode-core/test/jest/context/workspaceContext.test.ts
+++ b/packages/salesforcedx-vscode-core/test/jest/context/workspaceContext.test.ts
@@ -148,4 +148,92 @@ describe('workspaceContext', () => {
       expect(orgId).not.toBeNull();
     });
   });
+
+  describe('initialization promise logic', () => {
+    let mockWorkspaceContextUtil: any;
+    let mockExtensionContext: any;
+
+    beforeEach(() => {
+      jest.clearAllMocks();
+      mockWorkspaceContextUtil = {
+        onOrgChange: jest.fn(),
+        initialize: jest.fn().mockResolvedValue(undefined),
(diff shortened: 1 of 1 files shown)


daphne-sfdc (approved):
- A4V produces no errors in **Developer: Toggle Developer Tools** during activation.  Same behavior as v65.8.2. ✅
- No regression in Agentforce Vibes IDE. ✅

PR #6848, opened 2026-02-11
feat: pft telemetry @W-21233151

mshanemc:
### What does this PR do?
send telemetry via o11y for PFT
records only commandExecution
sets the endpoint/PFT id for all of our extensions
support for 2PP extensions is them adding the correct props to pjson

@W-21233151@

QA notes: be sure to deactivate dev mode for testing locally


packages/salesforcedx-vscode-services/src/observability/appInsights.ts
```diff
@@ -15,6 +15,9 @@ export const isTelemetryExtensionConfigurationEnabled = (): boolean => {
   // TODO: should we consult the CLI's telemetry preference?
   const enabled =
     workspace.getConfiguration('telemetry').get<string>('telemetryLevel', 'all') !== 'off' &&
-    workspace.getConfiguration('salesforcedx-vscode-core').get<boolean>('telemetry.enabled', false);
+    // on the web, no core extension is every installed so we can't consult the config
+    (process.env.ESBUILD_PLATFORM === 'web' ||
+      workspace.getConfiguration('salesforcedx-vscode-core').get<boolean>('telemetry.enabled', false));
+  console.log('isTelemetryExtensionConfigurationEnabled', enabled);
```
jonnyhork: Do you still want this cosole.log?

packages/salesforcedx-vscode-services/src/observability/appInsights.ts
```diff
@@ -15,6 +15,9 @@ export const isTelemetryExtensionConfigurationEnabled = (): boolean => {
   // TODO: should we consult the CLI's telemetry preference?
   const enabled =
     workspace.getConfiguration('telemetry').get<string>('telemetryLevel', 'all') !== 'off' &&
-    workspace.getConfiguration('salesforcedx-vscode-core').get<boolean>('telemetry.enabled', false);
+    // on the web, no core extension is every installed so we can't consult the config
```
jonnyhork:
```suggestion
    // on the web, no core extension is ever installed so we can't consult the config
```

packages/salesforcedx-utils-vscode/src/telemetry/reporters/o11yReporter.ts
```diff
@@ -69,6 +77,23 @@ export class O11yReporter
...
+    if (!this.productFeatureId) {
+      return;
+    }
+    this.o11yService.logEventWithSchema(
+      {
+        eventName: 'vscodeExtension.executed',
+        productFeatureId: this.productFeatureId,
+        componentId: `${this.extensionId}.${commandId}`,
+        contextName: 'orgId::devhubId',
+        contextValue: `${orgId}::${devHubId}`
```
jonnyhork, line 91: If these orgId and deHubId are not defined, we would still send `undefined:undefined` right? 

Is there a way to omit the property if not defined instead? Not a big deal, but undefined tends to feel like an error to me or and possibly others.

packages/salesforcedx-vscode-services/src/observability/appInsights.ts
mshanemc: nope!

packages/salesforcedx-utils-vscode/src/telemetry/reporters/o11yReporter.ts
mshanemc: no, they show up as empty strings because of line 103/105

mshanemc: you're going to love TS!  because the types on the pft method are string for those props, it won't let you pass in undefined.  You don't have to worry about stuff like that.

[there's also an eslint rule to keep people from using null/undefined stuff in template literals...we don't have it on but we could]

jonnyhork: ah those lines were hidden/collapsed by GitHub...

PR #6854, opened 2026-02-13
fix: web extensions use instanceUrl for isolating by org W-20890801

mshanemc:
### What does this PR do?
use the instanceUrl from settings as part of the dbkey

note: on local extensionHost, you'll see `default` as the domain because FS gets created BEFORE the settings are present.  It's not valuable QA.

to validate, use CBW with the instance populated in the iframe (either CBW on workspace or the CBW repo)

@W-20890801@


packages/salesforcedx-vscode-services/src/virtualFsProvider/indexedDbStorage.ts
```diff
@@ -19,7 +20,24 @@ import {
...
+  '.my.salesforce.mil',
+  '.my-salesforce.com',
+  '.my.sfcrmproducts.cn'
+] as const;
+
+export const parseMyDomain = (instanceUrl: string): string => {
+  const { hostname } = new URL(instanceUrl);
+  const suffix = SALESFORCE_DOMAIN_SUFFIXES.find(s => hostname.endsWith(s));
+  return suffix ? hostname.slice(0, -suffix.length) : hostname;
+};
```
jonnyhork, line 34: I think we can do without this part. It seems to just be a name shortener? Though only for domains on the list? 

I already have an issue where I need to maintain 3 disconnected lists of approved/valid domains, this would be the 4th list. 

If a domain is not in this list we will get inconsistent database naming convention.

The DB isolation should work just fine with the full domain and it's not something users should have to look at in the browser.

I would suggest we use `instanceUrl ? new URL(instanceUrl).hostname : 'default';` for everyone?
1. It guarantees uniqueness (no collision risk).
2. It automatically supports all URLs without code changes.
3. It eliminates the risk of 'split brain' if our list gets out of sync with the SF URLs. Or if SF ever changes a URL format.

jonnyhork: I'm really glad I saw this list though! It made me realize I need to add support for gov/mil and China to all my lists or the app wont load. 
Especially since I just added webide perms to the China editions.

jonnyhork (approved):
Besides the one comment, this looks good and works for me locally.

packages/salesforcedx-vscode-services/src/virtualFsProvider/indexedDbStorage.ts
mshanemc: good idea, will do the whole URL.

PR #6856, opened 2026-02-16
feat: @W-21274286: auth endpoint telemetry - o11yReporter initialize with getConnection and tests

smuduganti-sf:
### What does this PR do?
Enables uploading o11y telemetry to the authenticated org endpoint (e.g. Connect proxy ui-telemetry) when a connection is available, with fallback to the static experience-site endpoint. This reduces reliance on a single shared endpoint and aligns with instrumentation’s move to not count this traffic against customer API limits. O11yReporter now passes a getConnection callback into the o11y service’s initialize() for dynamic endpoint resolution; unit tests were added/updated for this flow.

@W-21274286@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>


tariqrafique (commented):
Looks good to me,
My only concern is how expensive is the getConnection Operation and if it's okay to do that on every telemetry upload

packages/salesforcedx-utils-vscode/src/telemetry/reporters/o11yReporter.ts
```diff
@@ -15,6 +15,10 @@ import { WorkspaceContextUtil } from '../../context/workspaceContextUtil';
 import { isInternalHost } from '../utils/isInternal';
 import { getCommonProperties, getInternalProperties } from './telemetryUtils';
 
+async function getConnection() {
+  return WorkspaceContextUtil.getInstance().getConnection();
```
mshanemc, line 19: does o11yService have handling for situations when this throws?

smuduganti-sf: Yes. In resolveO11yUploadTarget() we wrap the call to getConnection() in a try/catch. If it throws (or we don’t get a valid connection/org), we log the error with console.error and fall back to the static o11yUploadEndpoint (no auth) for that upload. So telemetry still gets sent to the experience-site fallback instead of failing the upload.

packages/salesforcedx-vscode-services/package.json
```diff
@@ -41,7 +41,7 @@
     "@opentelemetry/sdk-trace-web": "2.2.0",
     "@salesforce/core": "^8.23.4",
     "@salesforce/vscode-i18n": "*",
-    "@salesforce/o11y-reporter": "^1.7.4",
+    "@salesforce/o11y-reporter": "1.8.0",
```
gbockus-sf: Was the move from ^ to pin on purpose?

mshanemc: good catch.  We like to avoid pins without reason (ex  reason: someone is consistently bad at semver)

PR #6859, opened 2026-02-16
fix: file presence perf W-21282879

mshanemc:
### What does this PR do?
CompSetService has a new no-dep, no param fn to build a componentSet of the project
compSet is not part of the queue item
the getChildren builds compSet once and all the queue items can use it

much faster
credit to Pete for the idea

@W-21282879@


packages/salesforcedx-vscode-org-browser/src/tree/customField.ts
```diff
@@ -5,42 +5,38 @@
...
 
 export const createCustomFieldNode =
-  (treeProvider: MetadataTypeTreeProvider) => (element: OrgBrowserTreeItem) => (f: CustomObjectField) =>
-    Effect.gen(function* () {
+  (projectComponentSet: ComponentSet) => (treeProvider: MetadataTypeTreeProvider) => (element: OrgBrowserTreeItem) =>
+    Effect.fn('createCustomFieldNode')(function* (field: CustomObjectField) {
       // Create a MetadataListResultItem-like object for the custom field
       const fieldMetadata: MetadataListResultItem = {
-        fullName: `${element.componentName}.${removeNamespacePrefix(element)(f).name}`,
+        fullName: `${element.componentName}.${removeNamespacePrefix(element)(field).name}`,
```
peternhale: Why the need for a curried function? nothing interesting is done in the first fn, just a pass through.

mshanemc: there's never a need for curry.  It's more of a choice between
1: (foo, bar, baz) =>
2: ({foo, bar, baz})
3: (foo)(bar)(baz)

to me, 
- easier to keep the inner "real" function the same when the "outer/dependencies" get refactored or types change
- definitely easier to partially apply (for tests or loops) when the "outer" stuff doesn't change but the result of changes to the inner arguments are of interest.
- so much nicer to bars.map(theFunction(foo)) instead of bars.map((bar) => theFunction(foo, bar))

PR #6866, opened 2026-02-19
[W-19981684] Update git2gus default build to 66

daphne-sfdc:
### What does this PR do?
Update git2gus default build to 66

@W-19981684@


mshanemc (commented):
Make sure that build exists in gus

PR #6886, opened 2026-02-24
feat(apex): read-only Apex for Code Builder Web W-21348518

mshanemc:
### What does this PR do?
Read-only Apex support for Code Builder Web. 

side quests: 
- i18n lint for progress notifications.
- Org browser retrieve refinements (ignore tracking/conflicts when overwrite confirmation exists). 

@W-21348518@


packages/salesforcedx-vscode-services/src/virtualFsProvider/fileSystemProvider.ts
```diff
@@ -28,25 +33,55 @@ const handleFileSystemError = (error: unknown, uri: vscode.Uri): never => {
...
- * This is hooked up to memfs right now, and its watcher handles everything else
- */
+/** Extract SDR suffix from path: "MyClass.cls" -> "cls", "MyClass.cls-meta.xml" -> "cls" */
+const suffixFromPath = (fsPath: string): string | undefined => {
+  const base = path.basename(fsPath);
+  const metaMatch = base.match(/\.([^.]+)-meta\.xml$/);
+  if (metaMatch) return metaMatch[1];
+  const ext = path.extname(fsPath).slice(1);
+  return ext || undefined;
+};
```
jonnyhork, line 43: The tooling API applies the same restrictions on triggers too, we should probably include `.trigger`

mshanemc: this is a utility fn to get the suffix, the comment is an example of what it would do for any type

it does triggers, see
https://github.com/forcedotcom/salesforcedx-vscode/pull/6886/changes#diff-cc204f09177b440c50599b0041f829fb6de71a06504979e0b3c08500256ae6e5R71

mshanemc: but definitely QA it with a trigger if you're not sure

jonnyhork: ok cool, thats the line I was looking for but totally missed it, got distracted by the .cls here. I was confused cause when I tested this, I did see the triggers were read only already.

PR #6888, opened 2026-02-24
[W-21263254] feat: namespaces and packages grouping in apex test explorer

CristiCanizales:
## Summary

Adds **namespace and package grouping** to the Apex Test Explorer. Tests from the org are shown in a hierarchy: **Namespace → Package → Class → Method**, so users can tell which tests are unpackaged, in a 1GP namespace, or in a 2GP package (including unlocked packages).

### Test Explorer tree
- **Namespaces**: "Local Namespace" for classes with no namespace; other namespaces by prefix. Local is ordered first, then others alphabetically.
- **Packages**: Under each namespace:
  - **(Unpackaged Metadata)** for local/unpackaged classes
  - **\<Namespace\> (1GP)** for 1GP namespaced classes
  - **2GP package name** (with "(Unlocked)" when `ContainerOptions === 'Unlocked'`) for classes resolved via Package2Member / Package2
- **Run/debug**: `gatherTests` and test item IDs support the new namespace and package nodes (containers only; execution still uses class/method items).

### Package resolution (`packageResolution.ts` + `schemas.ts`)
- Resolves Apex class IDs to package info (2GP) via Tooling API:
  - Uses **Package2Member** and **Package2** when available (e.g. dev hub / packaging orgs).
  - Falls back to **SubjectId** when **MetadataComponentId** is not available (e.g. some subscriber orgs).
  - Falls back to **InstalledSubscriberPackage** when Package2Member is not supported (e.g. subscriber orgs), matching by namespace and, for no-namespace, using a single no-namespace installed package (e.g. Trigger Actions Framework) with ManageableState filtering where available.
- Results are cached per org; orgs where Package2/Package2Member are unavailable are tracked so we don't keep querying.
- Shared types (`Package2MemberRecord`, `Package2Record`, `InstalledSubscriberPackageRecord`, `ResolvedPackageInfo`) live in `schemas.ts`; `packageResolution` re-exports `ResolvedPackageInfo`.
- Test-only reset for cache and unavailable-org set so tests don't share state.

### Refactor and helpers
- **`orgTestItems.ts`**: Helpers for building the tree (e.g. `buildNamespacePackageStructure`, `sortNamespaceKeys`, `getPackageKeysOrdered`, `getPackageLabelAndId`, `createClassAndMethodsFactory`, `getNamespaceDisplayLabel`). JSDoc added on public types and functions.
- **`testController.ts`**: `populateTestItemsFromOrg` now uses the helpers above, uses a single expression for 2GP/1GP/unpackaged, and batches UI updates (yield every 50 classes globally) so the explorer stays responsive with many tests.

### Constants, i18n, test item utils
- **constants**: `LOCAL_NAMESPACE_KEY`, `UNPACKAGED_PACKAGE_KEY`, `UNPACKAGED_PACKAGE_ID`, and `TEST_ID_PREFIXES.NAMESPACE` / `PACKAGE`.
- **i18n**: `test_explorer_local_namespace_label`, `test_explorer_unpackaged_metadata_label`, `test_explorer_1gp_package_label`.
- **testItemUtils**: `parseTestId` and helpers for namespace/package IDs (`isNamespace`, `isPackage`, `createNamespaceId`, `createPackageId`); `gatherTests` recurses through namespace and package nodes.

## Testing

- **packageResolution.test.ts**: Package resolution (empty IDs, no Package2Member, 2GP resolution, cache, SubjectId fallback, InstalledSubscriberPackage fallback, no-namespace single package, ManageableState filtering, unavailable org with cache). Uses `resetPackageResolutionState()` in `beforeEach` for the `resolvePackage2Members` describe to avoid shared state.
- **testController.test.ts**: Test controller behavior with the new tree (mocks for `resolvePackage2Members` and class URI index).
- **testItemUtils.test.ts**: Parsing and ID helpers for namespace and package.

@W-21263254@

### Functionality before
image

### Functionality after
image


packages/salesforcedx-vscode-apex-testing/src/testDiscovery/schemas.ts
```diff
@@ -45,3 +45,36 @@ export type ToolingTestsPage = {
 export type DiscoverTestsOptions = {
   namespacePrefix?: string;
 };
+
+// Package resolution: Tooling API query result shapes
+export type Package2MemberRecord = {
```
mshanemc: suggestion: use types from wsdl instead of recreating partials of them  

https://github.com/forcedotcom/wsdl/blob/6d655101b39a028294903ff392371b396198ad7b/src/tooling.ts#L55103

packages/salesforcedx-vscode-apex-testing/src/testDiscovery/packageResolution.ts
```diff
@@ -0,0 +1,463 @@
...
+ * Returns a cache key for the current org (caller must pass after connection is available).
+ */
+export const getPackageResolutionCacheKey = (connection: Connection): string => {
+  try {
+    const authFields = connection.getAuthInfoFields();
+    return authFields.orgId ?? authFields.username ?? 'unknown';
+  } catch {
+    return 'unknown';
+  }
+};
```
mshanemc: defaultOrgRef from Services will have an orgId if a connection has happened.  That'll prevent having to do a file read to get the orgId.

packages/salesforcedx-vscode-apex-testing/src/testDiscovery/packageResolution.ts
```diff
@@ -0,0 +1,463 @@
...
+
+/** Thrown when MetadataComponentId is not available so caller should try SubjectId. */
+class TrySubjectIdError extends Error {
+  constructor() {
+    super('TrySubjectId');
+    this.name = 'TrySubjectIdError';
+  }
+}
+
+const getComponentId = (m: Package2MemberRecord): string => m.MetadataComponentId ?? m.SubjectId ?? '';
```
mshanemc: I think `string|undefined` is a lot nicer to work with than the empty string.

packages/salesforcedx-vscode-apex-testing/src/utils/testItemUtils.ts
```diff
@@ -57,6 +57,20 @@ export const parseTestId = (id: string): TestIdInfo => {
     }
   }
 
+  if (id.startsWith(TEST_ID_PREFIXES.NAMESPACE)) {
```
mshanemc: why don't this reuse the "is" stuff from below

packages/salesforcedx-vscode-apex-testing/src/views/orgTestItems.ts
```diff
@@ -0,0 +1,211 @@
...
+      classItem.tags = [inWorkspaceTag];
+    }
+    classItems.set(fullClassName, classItem);
+
+    const methodNames = new Set<string>();
+    for (const entry of classEntries) {
+      for (const testMethod of entry.testMethods ?? []) {
+        methodNames.add(testMethod.name);
+      }
+    }
```
mshanemc:
```suggestion
    const methodNames = new Set<string>();
    const methodNames = new Set(
      classEntries.flatMap(entry => (entry.testMethods ?? []).map(m => m.name))
    );
```

packages/salesforcedx-vscode-apex-testing/src/views/orgTestItems.ts
```diff
@@ -0,0 +1,211 @@
...
+    : packageKeys;
+
+/**
+ * Returns the display label and stable ID for a package node in the Test Explorer.
+ * Handles unpackaged, 1GP (namespaced), and 2GP (including Unlocked suffix when applicable).
+ */
+export const getPackageLabelAndId = (
+  nsKey: string,
+  pkgKey: string,
+  classEntriesList: ClassEntry[],
```
mshanemc: not asking for a change, just making you aware that there are some more advanced types for this kind of thing like https://effect.website/docs/schema/basic-usage/#non-empty-arrays

that would make the `classEntriesList[0].entries[0]` safe.

As is, I checked that the 1 consumer that uses it has done the length checks necesssary before calling, but that feels like something that could bites us in the future

packages/salesforcedx-vscode-apex-testing/src/utils/testItemUtils.ts
CristiCanizales: that was the idea 😆 thx

mshanemc (approved):

PR #6896, opened 2026-02-25
[W-21240055] feat(metadata,services,soql): migrate SObject refresh pipeline to services layer

peternhale:
@W-21240055@

### Description

Collapses `@salesforce/salesforcedx-sobjects-faux-generator` into `salesforcedx-vscode-metadata`, moves the "Refresh SObject Definitions" command from `salesforcedx-vscode-core` to `salesforcedx-vscode-metadata`, and introduces a unified SObject data layer in `salesforcedx-vscode-services` consumed by both the metadata and SOQL extensions.

**`salesforcedx-vscode-services`**
- New `TransmogrifierService`: Effect Schema-typed `SObject`/`SObjectField`/`ChildRelationship` structs with `toMinimalSObject` (raw → typed) and `decodeSObject` (validated JSON decode)
- New `FsService.safeWriteFile`: implicit `createDirectory` before write, eliminating per-file overhead
- `MetadataDescribeService`: added `listSObjects` (with `queryable` flag), `describeCustomObject`, and `describeCustomObjects` (composite batch, 15-concurrency stream); both cached with 30-min TTL via `Effect.Cache`
- `ProjectService`: individual path methods (`getFauxStandardObjectsPath`, `getFauxCustomObjectsPath`, `getSoqlMetadataPath`, `getSoqlStandardObjectsPath`, `getSoqlCustomObjectsPath`, `getTypingsPath`) replacing scattered local constants

**`salesforcedx-vscode-metadata`** (new home for SObject refresh)
- Removed: `salesforcedx-sobjects-faux-generator` package deleted entirely
- `sf.internal.refreshsobjects` command registered here; `salesforcedx-vscode-core` delegates to it
- Streaming refresh pipeline: `listSObjects → describeCustomObjects (batch stream) → toMinimalSObject → parallel file writes` with concurrency 100 — all phases share a single Effect layer/connection
- No dependency on `salesforcedx-utils-vscode`; all paths via `ProjectService`, all I/O via `FsService`, all transforms via `TransmogrifierService`

**`salesforcedx-vscode-soql`**
- `JsforceOrgDataSource` replaced by `ServicesOrgDataSource` using `MetadataDescribeService` + `TransmogrifierService`
- `FileSystemOrgDataSource` fixed: now checks both standard and custom object directories, uses `FsService` for I/O, and `TransmogrifierService.decodeSObject` for validated JSON parsing
- `SObject`/`SObjectField` types imported directly from `salesforcedx-vscode-services`


packages/salesforcedx-vscode-metadata/src/commands/refreshSObjects.ts
```diff
@@ -0,0 +1,111 @@
...
+    Effect.ensuring(Effect.sync(() => { activeState.isActive = false; }))
+  );
+
+/**
+ * Refresh SObject definitions command — Effect pattern, no SfCommandlet framework.
+ * Registered as sf.internal.refreshsobjects in metadata's index.ts.
+ */
+export const refreshSObjectsCommand = (source?: SObjectRefreshSource) =>
+  Effect.gen(function* () {
+    if (activeState.isActive) {
```
mshanemc, line 92: Can it use Effect's semaphore (permits 1) instead of mutating the module-level var?
you still end up with semaphore.make at the module level, or the ext level, or something but it's hopefully cleaner

then you can do ifAvailable if you want to exit early if Option.none.

  /**
   * Runs an effect only if the specified number of permits are immediately
   * available.
   *
   * **Details**
   *
   * This function attempts to acquire the specified number of permits. If they
   * are available, it runs the effect and releases the permits after the effect
   * completes. If permits are not available, the effect does not execute, and
   * the result is `Option.none`.
   */
  withPermitsIfAvailable(permits: number): <A, E, R>(self: Effect<A, E, R>) => Effect<Option.Option<A>, E, R>

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
```diff
@@ -102,28 +131,118 @@ export class MetadataDescribeService extends Effect.Service<MetadataDescribeServ
...
-
-      Effect.log(result.fields.map(f => f.name).join(', '));
-      return result;
+      const orgId = (yield* connectionService.getConnection()).getAuthInfoFields().orgId ?? 'default';
+      return yield* sobjectDescribeCache.get(`${orgId}:${objectName}`);
     });
 
+    const describeCustomObjects = (objectNames: string[]): Stream.Stream<DescribeSObjectResult, MetadataDescribeError> => {
+      if (objectNames.length === 0) return Stream.empty;
+      const batches = Array.from(
```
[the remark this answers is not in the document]
mshanemc: the simpler approach might be (after checking connection) to Stream.fromIterable of your type names so that you get the chunking from the stream

[ai generated, not tried this]
```ts
const describeCustomObjects = (objectNames: string[]): Stream.Stream<DescribeSObjectResult, MetadataDescribeError> =>
  objectNames.length === 0
    ? Stream.empty
    : Stream.fromEffect(
        connectionService.getConnection().pipe(
          Effect.mapError(e => {
            const { cause } = unknownToErrorCause(e);
            return new MetadataDescribeError({ cause, function: 'describeCustomObjects', message: `...` });
          })
        )
      ).pipe(
        Stream.flatMap(conn =>
          Stream.fromIterable(objectNames).pipe(
            Stream.grouped(MAX_SOBJECT_BATCH_SIZE),
            Stream.mapEffect(
              chunk =>
                Effect.tryPromise({ try: () => runSObjectBatch(conn, Chunk.toArray(chunk)), catch: ... }).pipe(
                  Effect.map(res => res?.results?.flatMap(sr => ...) ?? [])
                ),
              { concurrency: BATCH_CONCURRENCY }
            ),
            Stream.flatMap(batchResults => Stream.fromIterable(batchResults))
          )
        )
      );
```

packages/salesforcedx-vscode-services/src/core/transmogrifierService.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+import type { Connection } from '@salesforce/core';
+import * as Effect from 'effect/Effect';
+import * as S from 'effect/Schema';
+
+type RawDescribeSObjectResult = Awaited<ReturnType<Connection['describe']>>;
+
+/** Re-exported raw jsforce describe result for consumer type safety */
+export type DescribeSObjectResult = RawDescribeSObjectResult;
```
mshanemc: if all the services are properly typed, that **should** be unnecessary? hopefully?

packages/salesforcedx-vscode-services/src/core/transmogrifierService.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const mapToSObject = (raw: RawDescribeSObjectResult): SObject => ({
+  name: raw.name,
+  label: raw.label,
+  custom: raw.custom,
+  queryable: raw.queryable,
+  fields: (raw.fields ?? []).map(f => ({
+    aggregatable: f.aggregatable,
+    custom: f.custom,
+    defaultValue: f.defaultValue ?? null,
```
mshanemc: I would love to have `null` not leave the immediate transformation from the server and to just use `undefined` instead to not have to deal with both

packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
```diff
@@ -5,109 +5,148 @@
...
+            const customsDir = yield* api.services.ProjectService.getSoqlCustomObjectsPath().pipe(
+              Effect.provide(projectLayer)
+            );
+            const soqlMetadata = yield* api.services.ProjectService.getSoqlMetadataPath().pipe(
+              Effect.provide(projectLayer)
+            );
+
+            const files: string[] = [];
+
+            const standardEntries = yield* fs.readDirectory(standardsDir).pipe(
```
mshanemc: linked to other comment--this would be much safer using URI than path

packages/salesforcedx-vscode-soql/src/lspClient/orgMetadata.ts
```diff
@@ -5,109 +5,148 @@
...
+              Effect.catchAll(() => Effect.succeed<string[]>([]))
+            );
+            files.push(...standardEntries);
+
+            const customEntries = yield* fs.readDirectory(customsDir).pipe(
+              Effect.provide(fsLayer),
+              Effect.map(entries => entries.map(entry => entry[0])),
+              Effect.catchAll(() => Effect.succeed<string[]>([]))
+            );
+            files.push(...customEntries);
```
mshanemc: these don't depend on each other, so you could do in parallel and not need the array mutation

klewis-sfdc (approved):
From an AFV-testing perspective, this is working and approved.  I built the extensions from this branch, and tested along with a VSIX of the [AFV PR](https://github.com/forcedotcom/salesforcedx-vscode-einstein-gpt/pull/1831) where Pete fixed our SObject Refresh nudger (that reminds customers to use SObject Refresh and gives them a button to call it).  

I saw the nudger working correctly and calling the SObject Refresh command worked perfectly also 👍🏽 

[Slack discussion](https://salesforce-internal.slack.com/archives/G02JDKH7W9K/p1772144026394449?thread_ts=1771969388.746779&cid=G02JDKH7W9K) 

Short video of my QE:

https://github.com/user-attachments/assets/41a6d723-b0ed-4efa-9865-21e566019435

(nm the fact that the upgrade notification mentions Cline - that is just an aritifact of me building the vsix locally)

PR #6904, opened 2026-02-26
[W-21343900] feat: GHA workflow to publish @salesforce/soql-common to NPMJS

daphne-sfdc:
### What does this PR do?
Allows the `@salesforce/soql-common` dependency, which is still used by the SOQL Language Server, to continue being published to NPMJS after its migration from the soql-tooling repo.

@W-21343900@

### Testing

In my fork daphne-sfdc/salesforcedx-vscode-soql-common
1. In packages/soql-common/package.json: Change the "name" to `@elephant32/soql-common`
2. Replace all local usages of `@salesforce/soql-common` with `@elephant32/soql-common` so the monorepo can compile
3. `npm run reinstall` -> `npm run compile`
4. Go to GHA and run the workflow “Publish soql-common Package to npm”

In soql-language-server
1. In package.json: Replace `"@salesforce/soql-common": "0.2.1"` with `"@elephant32/soql-common": "1.2.1"`
2. `yarn install` -> `yarn build`
3. Fix all the compilation errors
4. `yarn link`

Go back to my fork
1. `yarn link @salesforce/soql-language-server`
2. Launch Extensions
3. Confirm the following continue to work as expected:

* SFDX: Execute SOQL Query... ✅
* SFDX: Execute SOQL Query With Currently Selected Text ✅
* SFDX: Create Query in SOQL Builder ✅
* SOQL Builder works ✅

In a .soql file
- Autocompletion ✅
- Correct coloring ✅

### DIFF
--- a/.github/workflows/publishSoqlCommon.yml
+++ b/.github/workflows/publishSoqlCommon.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Create Github Release
         if: steps.changelog.outputs.skipped == 'false'
-        uses: ncipollo/release-action@v1.20.0
+        uses: ncipollo/release-action@b7eabc95ff50cbeeedec83973935c8f306dfcd0b
         with:
           name: ${{ steps.changelog.outputs.tag }}
           tag: ${{ steps.changelog.outputs.tag }}


.github/workflows/publishSoqlCommon.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          token: ${{ secrets.IDEE_GH_TOKEN }}
+          ref: develop
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ vars.NODE_VERSION || 'lts/*' }}
+      - uses: ./.github/actions/gitConfig
+        with:
+          email: ${{ secrets.IDEE_GH_EMAIL }}
+      - name: Bump patch version
```
mshanemc: seems weird to always bump patch.  I think we'd want to follow semver, etc (ex: having the option to do major/minor version bumps)

And is the idea to always do this manually, and hopefully remember when things need publishing?

.github/workflows/publishSoqlCommon.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          git add packages/soql-common/package.json
+          git commit -m "chore: bump @salesforce/soql-common to v${{ steps.bump.outputs.version }}"
+          git push
+      - name: Create and push tag
+        id: tag
+        run: |
+          TAG="soql-common-v${{ steps.bump.outputs.version }}"
+          git tag "$TAG"
+          git push origin "$TAG"
```
mshanemc: maybe make the tag as part of the previous commit, then push the v-change and the tag in one push?

.github/workflows/publishSoqlCommon.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "tag=$TAG" >> $GITHUB_OUTPUT
+
+  publish:
```
mshanemc, line 79: there's an existing "Publish Types Package to npm (shared)" that you might want to look at.  It's got a bunch of nice stuff built in (ex: making sure that version didn't already get published somehow)

.github/workflows/publishSoqlCommon.yml
```diff
@@ -0,0 +1,91 @@
...
+          version-file: './packages/soql-common/package.json'
+          git-path: 'packages/soql-common'
+          release-count: '0'
+          output-file: 'packages/soql-common/CHANGELOG.md'
+          skip-on-empty: ${{ github.event_name == 'push' }}
+          skip-tag: false
+
+      - name: Create Github Release
+        if: steps.changelog.outputs.skipped == 'false'
+        uses: ncipollo/release-action@2c591bcc8ecdcd2db72b97d6147f871fcd833ba5
```
mshanemc: fyi: Security requires us to pin 3PP gha to a SHA.  That sha takes you to 1.1.0

they added a https://github.com/ncipollo/release-action/releases/tag/v1.12.0 "makeLatest" option a few years ago that is false by default...

so I think our best optin would be to
1. check the RN on this action betwen 1.1.0 and 1.12.0 to make sure bumping is not going to cause any other problems.
2. bump the SHA so we don't get `latest` on it at all.

mshanemc: and maybe we set it explicitly `:false` so that's obvious what it's for.  Hopefully that's verifiable on your test repo.

daphne-sfdc: I tried `makeLatest: false` but it actually doesn't work.  When there's no tag set as **Latest**, Github by default selects the one with the most recent date. 😔

PR #6907, opened 2026-02-26
[W-21368115] chore(eslint): improve no-unused-i18n-messages rule 

peternhale:
- Fixes rule to correctly distinguish declarations from references — quoted keys in the messages object were previously counted as usages of themselves
- Replaces boolean used-set with reference count map for accurate detection
- Adds invalidAllowListEntry validation: allowList keys that don't exist in the messages file are reported, closing the gap where stale config entries could go unnoticed
- Refactors noDuplicateI18nValues to use AST-first visitor pattern (VariableDeclarator + Property) instead of fragile node.parent.parent navigation
- Adds test coverage for the quoted-key edge case and allowList validation

### What does this PR do?

- Fixes the `no-unused-i18n-messages` rule to correctly distinguish declarations from references — quoted keys in the messages object were previously counted as usages of themselves
- Replaces a boolean "used" set with a reference count map, enabling more accurate detection
- Adds `invalidAllowListEntry` validation: keys listed in `allowList` that don't exist in the messages file are now reported, closing the gap where stale config entries could go unnoticed
- Refactors `noDuplicateI18nValues` to use an AST-first pattern (`VariableDeclarator` + `Property` visitors) instead of fragile `node.parent.parent` navigation
- Adds test coverage for the quoted-key edge case and the new allowList validation

@W-21368115@

Made-with: Cursor

### DIFF
--- a/packages/eslint-local-rules/src/noDuplicateI18nValues.ts
+++ b/packages/eslint-local-rules/src/noDuplicateI18nValues.ts
@@ -80,24 +80,28 @@ export const noDuplicateI18nValues = RuleCreator.withoutDocs({
     }
 
     const enMessages: MessagesObject = extractMessagesObject(enAst);
+
+    const unwrapAsExpression = (node: TSESTree.Expression): TSESTree.Expression =>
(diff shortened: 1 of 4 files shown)


mshanemc (approved):
works great

PR #6922, opened 2026-03-02
refactor(metadata,soql,org-browser): multi-org metadata cache with shared singleton API W-21240055

peternhale:
@W-21240055@

### Description

Refactors `MetadataDescribeService` to use a per-org cache registry and exposes a shared singleton API (`MetadataDescribeApi`) so all consumer extensions share one cache per org, eliminating duplicate fetches across extensions.

### Key changes

- **Per-org cache registry** (`OrgCacheState` keyed by `orgId`) — replaces a single-org cache that silently served stale data on org switch
- **`listMetadata` caching** — `conn.metadata.list({type})` is now cached alongside `describe`, `listSObjects`, and `describeCustomObject`
- **`MetadataDescribeApi`** — pre-satisfied (`R = never`) Effect API exposed on the services extension index; consumer extensions call it without managing their own `MetadataDescribeService.Default` layer
- **`ManagedRuntime` singletons** in `metadata`, `soql`, and `org-browser` — avoids rebuilding stateful services on every command invocation
- **Cache-probe optimisation** in `describeCustomObjects` — probes cache first, batch-fetches only misses; replaces a double list traversal with a single `reduce`
- **Test fix** — `soqlEditorProvider.test.ts` mock updated to include `getSoqlRuntime`

### DIFF
--- a/packages/salesforcedx-vscode-metadata/src/commands/refreshSObjects.ts
+++ b/packages/salesforcedx-vscode-metadata/src/commands/refreshSObjects.ts
@@ -5,11 +5,10 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import type { SObjectCategory, SObjectRefreshSource } from '../sobjects/types/general';
-import { ExtensionProviderService, getExtensionScope } from '@salesforce/effect-ext-utils';
+import { ExtensionProviderService } from '@salesforce/effect-ext-utils';
 import * as Effect from 'effect/Effect';
+import * as ManagedRuntime from 'effect/ManagedRuntime';
 import * as Option from 'effect/Option';
-import * as Runtime from 'effect/Runtime';
-import * as Scope from 'effect/Scope';
 import * as vscode from 'vscode';
 import { nls } from '../messages';
 import { AllServicesLayer } from '../services/extensionProvider';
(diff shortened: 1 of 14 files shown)


packages/salesforcedx-vscode-metadata/src/commands/refreshSObjects.ts
```diff
@@ -20,6 +19,20 @@ import { streamAndWriteSobjectArtifacts, writeSobjectArtifacts } from './sobject
 /** Command ID for cross-extension refresh completion notification */
 export const SOBJECT_REFRESH_COMPLETE_CMD = 'sf.internal.sobjectrefresh.complete';
 
+/**
+ * Single persistent runtime for artifact writing — built once on first refresh,
+ * reused for all subsequent invocations to avoid rebuilding TransmogrifierService
+ * and other stateful services. SObject describe/list use the services extension's
+ * shared singleton via MetadataDescribeApi (pre-satisfied, R = never).
+ */
+const createArtifactRuntime = () => ManagedRuntime.make(AllServicesLayer);
```
mshanemc, line 28: checking my understanding:

AllServicesLayer gets created early enough in `activate` that it'll be present by the time something that needs this ManagedRuntime is used?

mshanemc: If so, why not just have 1 (in place of AllServicesLayer) instead of having 1 just for refreshSobjects?

packages/salesforcedx-vscode-soql/src/services/org.ts
```diff
@@ -9,21 +9,21 @@ import type { Connection } from '@salesforce/core';
...
+  }));
 
 export const getConnection = (): Promise<Connection> =>
-  Effect.gen(function* () {
+  getSoqlRuntime().runPromise(Effect.gen(function* () {
     const api = yield* getServicesApi;
     const connectionService = yield* api.services.ConnectionService;
     return yield* connectionService.getConnection();
-  }).pipe(Effect.provide(AllServicesLayer), Effect.runPromise);
+  }));
```
mshanemc, line 29: I can't get this to compile.  There's some weird stuff going on the wireit scripts for this pkg that make it think it compiled successfully

mshanemc: well, it compiles (after nuking the cache, etc)  but the editor is fussing about it.  Even stranger.

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
```diff
@@ -63,6 +63,13 @@ export class MetadataDescribeService extends Effect.Service<MetadataDescribeServ
   ],
   effect: Effect.gen(function* () {
     const connectionService = yield* ConnectionService;
+
+    // ---------------------------------------------------------------------------
+    // Performers — execute network calls, used as Cache lookup functions.
+    // orgId is passed explicitly for span annotation; the actual connection is
```
mshanemc, line 69: I don't understand why the orgId is needed at all.  It'll be on the outer span anyway fro the telmetry layer

mshanemc: just to use it as a cache key?

peternhale: Cursor made me do it

mshanemc: I think it's for the cache key

PR #6925, opened 2026-03-03
feat: sobject refresh performance and progress indicator W-21240055

mshanemc:
### What does this PR do?
caching works, but with a lot less boilerplate for services dependency layers
(only provide the non-extensions-specific ones, get the "globals" from buildContext

@W-21240055@


packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
peternhale: I think this should be 1, like listSObjectsCache

peternhale:
```suggestion
const [type, folder, ...theRest] = key.split(':');
return performListMetadata(orgId, type, folder);
```

peternhale (approved):

packages/salesforcedx-vscode-services/src/core/metadataDescribeService.ts
mshanemc: so any orgId change will force a total do-oever?  Seems right

PR #6940, opened 2026-03-05
[W-21453020][W-21453007] refactor: remove node:path and node:os from salesforcedx-vscode-soql + replace imports from @salesforce/salesforcedx-utils-vscode with web-compatible equivalents

daphne-sfdc:
### What does this PR do?
- Removes all imports of `node:path` and `node:os` from **salesforcedx-vscode-soql** so that the SOQL extension can be web-compatible.
- Replaces all imports from `@salesforce/salesforcedx-utils-vscode` that relied on Node with web-compatible equivalents.

@W-21453020@, @W-21453007@

### Testing
- SOQL Builder works ✅
- Can save the SOQL query ✅
Screenshot 2026-03-05 at 12 15 55 AM
Screenshot 2026-03-05 at 12 16 10 AM


packages/salesforcedx-vscode-soql/src/commands/dataQuery.ts
```diff
@@ -64,8 +64,8 @@ class DataQueryExecutor extends LibraryCommandletExecutor<QueryAndApiInputs> {
 
     const timestamp = new Date().toISOString().replaceAll(/[:.]/g, '-');
     const fileName = `soql-query-${timestamp}.csv`;
-    const outputDir = path.join(workspaceUtils.getRootWorkspacePath(), '.sfdx', 'data');
-    const filePath = path.join(outputDir, fileName);
+    const fileUri = Utils.joinPath(URI.file(workspaceUtils.getRootWorkspacePath()), '.sfdx', 'data', fileName);
```
mshanemc: good use of Utils. use Services extension's WorkspaceService to get the root workspace path so we can get off of utils-vscode

packages/salesforcedx-vscode-soql/src/commands/dataQuery.ts
```diff
@@ -77,7 +77,7 @@ class DataQueryExecutor extends LibraryCommandletExecutor<QueryAndApiInputs> {
       )
       .then(selection => {
         if (selection === openFileAction) {
-          vscode.workspace.openTextDocument(vscode.Uri.file(filePath)).then(doc => {
+          vscode.workspace.openTextDocument(fileUri).then(doc => {
```
mshanemc, line 75: fyi, on services' FsService, there's a fn to do both of those in one.

packages/salesforcedx-vscode-soql/src/editor/soqlEditorProvider.ts
```diff
@@ -52,11 +51,9 @@ export class SOQLEditorProvider implements vscode.CustomTextEditorProvider {
   }
 
   private async getWebViewContent(webview: vscode.Webview): Promise<string> {
-    const soqlBuilderWebAssetsModule = getSoqlBuilderLocation(this.extensionContext);
-
-    const pathToHtml = path.join(soqlBuilderWebAssetsModule, HTML_FILE);
-    const htmlContent = await readFile(pathToHtml);
-    return HtmlUtils.transformHtml(htmlContent, soqlBuilderWebAssetsModule, webview);
+    const soqlBuilderUri = getSoqlBuilderLocation(this.extensionContext);
+    const htmlContent = await readFile(Utils.joinPath(soqlBuilderUri, HTML_FILE).fsPath);
```
mshanemc: you'll want to do this readFile with FsService, too.

it takes either URI or string and you'll want to use the URI.  Using `.fsPath` always fails on the web, I'm going to write a rule about it to help.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataHtml.ts
```diff
@@ -18,9 +18,8 @@ export const getHtml = async (
 ): Promise<string> => {
   const { baseStyleUri, tabulatorStyleUri, viewControllerUri, tabulatorUri, saveIconUri } = assets;
 
-  const pathToDataViewDist = path.join(extensionPath, DATA_VIEW_UI_PATH);
-  const pathToHtml = path.join(pathToDataViewDist, HTML_FILE);
-  let html = await readFile(pathToHtml);
+  const dataViewDistUri = Utils.joinPath(URI.file(extensionPath), DATA_VIEW_UI_PATH);
+  let html = await readFile(Utils.joinPath(dataViewDistUri, HTML_FILE).fsPath);
```
mshanemc: same comment as above

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -134,20 +133,21 @@ export class QueryDataViewService {
   }
 
   protected async getWebViewContent(webview: vscode.Webview): Promise<string> {
+    const extensionUri = URI.file(QueryDataViewService.extensionPath);
```
mshanemc: it's probably best to make QueryDataViewService.extensionPath a Uri instead (and rename it accordingly) to avoid all the conversions

fyi, ExtensionContext has both so it's easy to use the Uri from that.

```ts
		/**
		 * The uri of the directory containing the extension.
		 */
		readonly extensionUri: Uri;

		/**
		 * The absolute file path of the directory containing the extension. Shorthand
		 * notation for {@link TextDocument.uri ExtensionContext.extensionUri.fsPath} (independent of the uri scheme).
		 */
		readonly extensionPath: string;
```

packages/salesforcedx-vscode-soql/src/constants.ts
```diff
@@ -13,7 +12,7 @@ export const IMAGES_DIR_NAME = 'images';
 export const BUILDER_VIEW_TYPE = 'soqlCustom.soql';
 export const EDITOR_VIEW_TYPE = 'default';
 export const OPEN_WITH_COMMAND = 'vscode.openWith';
-export const SOQL_BUILDER_UI_PATH = path.join('dist', 'soql-builder-ui');
+export const SOQL_BUILDER_UI_PATH = 'dist/soql-builder-ui';
```
mshanemc: It's hard to tell what you're using these for, and whether assuming forward-slash will always be safe on windows.  and then since they're meant to be shared code, you get into the "who knows what someone else might do with it and expect it to work" issue.


Couple of options: 

`process.env.ESBUILD_PLATFORM === 'web' ? ` and then you can conditionally use the os sep (or just `/` on web)

make these an array like `['dist', 'soql-builder-ui'] and then let the consumer use URI Utils.join() for whatever they're probably going to do with it.

PR #6945, opened 2026-03-05
[W-21471646] feat: Verification code prints when user authorizes org in AFV IDE

hcarnes:
Verification code prints when user authorizes org in AFV IDE

@W-21471646@

### Functionality Before
Verification code did not print when user authorized org in AFV IDE using `SFDX: Authorize an Org` and `SFDX: Authorize a Devhub` commands.

### Functionality After
Verification code prints when user authorizes org in AFV IDE using `SFDX: Authorize an Org` and `SFDX: Authorize a Devhub` commands.


packages/salesforcedx-vscode-org/src/commands/auth/orgLoginWeb.ts
```diff
@@ -14,12 +14,16 @@ import {
   workspaceUtils,
   ContinueResponse
 } from '@salesforce/salesforcedx-utils-vscode';
+import * as crypto from 'node:crypto';
```
peternhale: @mshanemc, the org extension does not apply polyfill for node:crypto. Just checking if pulling in crypto is ok for now.

mshanemc: yes, because this is scoped within the org plugin, `which` isn't going to the web (using CLI, doing all kinds of other things).

mshanemc: you could probably do the same with global.webCrypto but 🤷🏻 won't matter.

packages/salesforcedx-vscode-org/src/commands/auth/orgLoginWeb.ts
```diff
@@ -14,12 +14,16 @@ import {
...
 import * as vscode from 'vscode';
 import { ORG_LOGIN_WEB } from '../../constants';
 import { nls } from '../../messages';
 import { updateConfigAndStateAggregators } from '../../util/orgUtil';
 import { AuthParams, AuthParamsGatherer } from './authParamsGatherer';
 
+export const generateVerificationCode = (token: string): string =>
+  crypto.createHash('sha256').update(token).digest('hex').slice(0, 4);
+
```
peternhale: @hcarnes can all of this, including the logic, be extracted to a different place, a common util within org extension that has all of the bits for creating the verification code?

PR #6953, opened 2026-03-06
[W-21078939] test: migrate soql.e2e.ts from RedHat to Playwright

daphne-sfdc:
### What does this PR do?
- Migrates the SOQL E2E test from RedHat vscode-extension-tester to Playwright.  Currently runs on desktop.
- Removes the caching for desktop tests so that the test always reruns instead of showing "Already fresh" and skipping.

@W-21078939@

### DIFF
--- a/.github/workflows/triggerE2EForCommit.yml
+++ b/.github/workflows/triggerE2EForCommit.yml
@@ -129,14 +129,7 @@ jobs:
       vscodeVersion: ${{ inputs.vscodeVersion || vars.MAX_REDHAT_VSCODE_VERSION || 'latest' }}
       runId: ${{ inputs.runId }}
 
-  # 🔍 SOQL Features
-  soql-tests:
-    uses: ./.github/workflows/runE2ETest.yml
-    secrets: inherit
-    with:
-      testToRun: 'soql.e2e.js'
-      vscodeVersion: ${{ inputs.vscodeVersion || vars.MAX_REDHAT_VSCODE_VERSION || 'latest' }}
-      runId: ${{ inputs.runId }}
+  # 🔍 SOQL Features - migrated to Playwright in soqlE2E.yml
 
   # 📋 Templates
   templates-tests:


packages/salesforcedx-vscode-apex-testing/package.json
```diff
@@ -200,11 +200,6 @@
         "../salesforcedx-vscode-metadata:vscode:bundle",
         "../playwright-vscode-ext:compile"
       ],
-      "files": [
```
mshanemc, line 203: what's the reason for changing all of the other e2e?  my goal with these was to not have them run on every commit, but only if the thing under test, or some depedndency, or the yarn lock changes.

packages/playwright-vscode-ext/src/orgs/minimalScratchOrgSetup.ts
```diff
@@ -44,7 +44,7 @@ export const createMinimalOrg = async (): Promise<OrgAuthResult> => {
   );
 
   const { stdout: createStdout } = await execAsync(
-    `sf org create scratch -d -w 10 -a ${MINIMAL_ORG_ALIAS} --wait 30 --json`,
```
mshanemc, line 47: meh, if the scratch org doesn't happen within 10 min, it's probably not going to and waiting 20 more is unlikely to do anything but slow things down.

Was there some issue you saw, or is this the AI slop?

packages/playwright-vscode-ext/src/utils/helpers.ts
```diff
@@ -280,6 +280,38 @@ export const enableMonacoAutoClosing = async (page: Page): Promise<void> => {
...
+ * "Developer: Show Running Extensions" editor.  More reliable than polling
+ * the command palette, especially on slow CI runners (e.g. Windows).
+ *
+ * While an extension is activating its row contains the text "Activating".
+ * Once done the row shows "Activation: Xms" / "Startup Activation: Xms".
+ * We wait until no rows contain "Activating" any more.
+ *
+ * @param timeout - Maximum ms to wait for all extensions to activate (default 120 000).
+ */
+export const waitForExtensionsActivated = async (page: Page, timeout = 120_000): Promise<void> => {
```
mshanemc, line 294: good idea, could be useful.  There are a few things that I think are happening in background processes where the exension finishes the "activate" but there's still something running (ex: getting the default org's "extended" details like userid, etc) that something depends on.

but probably mostly correct that the test can begin when everything is activated.

packages/salesforcedx-vscode-apex-testing/package.json
daphne-sfdc: When I was running the E2E test locally, it was really bugging me that I was getting the cached result instead of rerunning when I wanted to watch the rerun to help me debug.  I didn't realize this was intentionally done to save time.  Commit reverted.

packages/playwright-vscode-ext/src/orgs/minimalScratchOrgSetup.ts
daphne-sfdc: Then I'll put `-w 10` back in and remove the `--wait 30` flag at the end of the command?  The command wasn't working when I ran E2E locally because it contained the duplicate flag.

packages/playwright-vscode-ext/src/utils/helpers.ts
daphne-sfdc: We had this function in the RedHat E2E tests that I found to be super helpful at that time 🙂

mshanemc (approved):

PR #6955, opened 2026-03-07
feat(apex-testing): in-progress notification and cancellation for test suite create and Run Apex Tests W-21490620

CristiCanizales:
### What does this PR do?

Adds in-progress notification and cancellation for Apex test discovery in two command-palette flows:

- **SFDX: Create Apex Test Suite** – While gathering test classes for the quick pick (after the user enters the suite name), shows a notification “Retrieving tests…” with a Cancel button. If the user cancels, the command exits and the test controller is not refreshed.
- **SFDX: Run Apex Tests** – While gathering test classes and suites for the quick pick, shows the same “Retrieving tests…” notification, cancellable. On cancel, the command exits without opening the quick pick.

**Implementation details:**

- `window.withProgress(ProgressLocation.Notification, …)` with `nls.localize('retrieving_tests_message')` and `cancellable: true` in both flows.
- Cancellation: progress callback checks `token.isCancellationRequested` and throws `CancellationError`; caller catches and returns `{ type: 'CANCEL' }` so the commandlet stops.
- Test controller refresh (Create/Add suite): only runs when the commandlet actually executed (suite created or tests added). `SfCommandlet.run()` now returns `Promise<boolean>`; apexTestSuiteCreate and apexTestSuiteAdd refresh only when `run()` returns `true`.
- i18n: new message key `retrieving_tests_message` (“Retrieving tests…”) in en and ja.

@W-21490620@

### Functionality Before
- No progress feedback while test list was being built; users could think the command was unresponsive.
- No way to cancel once “Retrieving tests…” work had started.
- Test controller refreshed even when the user cancelled Create Apex Test Suite.

### Functionality After
- Notification shows “Retrieving tests…” during discovery; user can cancel.
- On cancel, command exits cleanly; test controller is not refreshed for Create/Add suite.

### DIFF
--- a/packages/salesforcedx-vscode-apex-testing/src/commands/apexTestRunUtils.ts
+++ b/packages/salesforcedx-vscode-apex-testing/src/commands/apexTestRunUtils.ts
@@ -13,9 +13,10 @@ import {
   TestService
 } from '@salesforce/apex-node';
 import { ExtensionProviderService } from '@salesforce/effect-ext-utils';
+import * as Cause from 'effect/Cause';
 import * as Effect from 'effect/Effect';
 import * as Stream from 'effect/Stream';
-import { CancellationToken } from 'vscode';
+import { CancellationToken, CancellationError } from 'vscode';
 import { URI } from 'vscode-uri';
 import * as settings from '../settings';
 import { writeAndOpenTestReport } from '../utils/testReportGenerator';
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-apex-testing/src/utils/commandletHelpers.ts
```diff
@@ -57,15 +57,20 @@ export class SfCommandlet<T> {
...
+    if (!(await this.prechecker.check())) {
+      return false;
+    }
+    const inputs = await this.gatherer.gather();
+    if (inputs.type === 'CONTINUE') {
+      await this.executor.execute(inputs);
+      return true;
+    }
+    if (inputs.msg) {
+      void notificationService.showErrorMessage(inputs.msg);
```
madhur310, line 71: why void and not just `notificationService.showErrorMessage(inputs.msg);` ? seems to be used with direct calls elsewhere.

CristiCanizales: It's so the user clicks to dismiss the notif

PR #6957, opened 2026-03-08
[W-21453020][W-21453007] refactor: remove node:path and node:os from salesforcedx-vscode-soql + replace imports from @salesforce/salesforcedx-utils-vscode with web-compatible equivalents

daphne-sfdc:
### What does this PR do?
- Removes all imports of `node:path` and `node:os` from **salesforcedx-vscode-soql** so that the SOQL extension can be web-compatible.
- Replaces all imports from `@salesforce/salesforcedx-utils-vscode` that relied on Node with web-compatible equivalents.

@W-21453020@, @W-21453007@

### Testing
- SOQL Builder works ✅
- Can save the SOQL query ✅
Screenshot 2026-03-05 at 12 15 55 AM
Screenshot 2026-03-05 at 12 16 10 AM

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
+++ b/packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
@@ -46,7 +46,7 @@ export class QueryDataFileService {
     }
   }
 
-  public async save(): Promise<string> {
+  public async save(): Promise<URI | undefined> {
     const defaultFileName = this.dataProvider.getFileName();
     const docUri = this.document.uri;
     const defaultUri =
@@ -54,9 +54,7 @@ export class QueryDataFileService {
 
     const fileInfo: URI | undefined = await vscode.window.showSaveDialog({ defaultUri });
 
-    if (fileInfo?.fsPath) {
-      // use .fsPath, not .path to account for OS.
-      const selectedFileSavePath = fileInfo.fsPath;
+    if (fileInfo) {
       const fileContentString = this.dataProvider.getFileContent(this.queryText, this.queryData.records);
 
       const workspacePath = await getSoqlRuntime().runPromise(
@@ -69,9 +67,9 @@ export class QueryDataFileService {
       );
       showFileInExplorer(fileInfo, workspacePath);
       showSaveSuccessMessage(Utils.basename(fileInfo));
-      return selectedFileSavePath;
+      return fileInfo;
     }
-    return '';
+    return undefined;
   }
 }
 
--- a/packages/salesforcedx-vscode-soql/test/jest/queryDataView/queryDataFileService.test.ts
+++ b/packages/salesforcedx-vscode-soql/test/jest/queryDataView/queryDataFileService.test.ts
@@ -35,9 +35,9 @@ describe('Query Data File Service', () => {
     (vscode.window.showSaveDialog as any).mockReturnValue(URI.file(savedFilePath));
     mockRunPromise.mockResolvedValue('/test/workspace');
 
-    const selectedFilePath = await queryDataFileService.save();
+    const savedFileUri = await queryDataFileService.save();
 
     expect(mockRunPromise).toHaveBeenCalled();
-    expect(selectedFilePath).toEqual(savedFilePath);
+    expect(savedFileUri?.fsPath).toEqual(URI.file(savedFilePath).fsPath);
   });
 });


packages/salesforcedx-vscode-soql/src/commands/dataQuery.ts
```diff
@@ -154,7 +147,12 @@ export const dataQuery = (): void => {
 const getMaxFetch = async (): Promise<number | undefined> => {
```
mshanemc, line 147: I think this maxFetch stuff and the stuff around 455 is unnecessary if you use https://github.com/forcedotcom/sfdx-core/blob/21f4fda68522edb7d1fe1da5603dcc7b6bdcd5a6/src/org/connection.ts#L358 instead.

have a look in the sfdx-core code and see if you agree.

We've got the lifecyle warnings stuff hooked up too for emitting the warning from
https://github.com/forcedotcom/sfdx-core/blob/21f4fda68522edb7d1fe1da5603dcc7b6bdcd5a6/src/org/connection.ts#L376

packages/salesforcedx-vscode-soql/src/commands/soqlBuilderToggle.ts
```diff
@@ -5,17 +5,12 @@
...
-export const soqlBuilderToggle = (doc: URI): Promise<void> => {
-  telemetryService.sendCommandEvent('soql_builder_toggle', TimingUtils.getCurrentTime());
 
+export const soqlBuilderToggle = Effect.fn('soql_builder_toggle')(function* (doc: URI) {
   const viewType = vscode.window.activeTextEditor ? BUILDER_VIEW_TYPE : EDITOR_VIEW_TYPE;
-
-  void vscode.commands.executeCommand(OPEN_WITH_COMMAND, doc, viewType);
-  return Promise.resolve();
-};
+  yield* Effect.promise(() => vscode.commands.executeCommand(OPEN_WITH_COMMAND, doc, viewType));
```
mshanemc, line 15: `.promise` (instead of `.tryPromise`) is for stuff that can never fail (ie, you don't need to do any error handling).  Is that the cese here?  I think it might be?

packages/salesforcedx-vscode-soql/src/editor/htmlUtils.ts
```diff
@@ -51,11 +50,11 @@ export class HtmlUtils {
    * @param pathToLwcDist
    * @param webview
    */
-  public static transformScriptTags(html: string, pathToLwcDist: string, webview: vscode.Webview): string {
+  public static transformScriptTags(html: string, lwcDistUri: URI, webview: vscode.Webview): string {
```
mshanemc, line 53: If this were not a class, it would be easier to tell that `transformScriptTags` is only used internally to this module and doesn't need to be public.  

Let's not have classes like this with only static members

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -48,39 +48,37 @@ export class QueryDataFileService {
 
   public async save(): Promise<string> {
     const defaultFileName = this.dataProvider.getFileName();
```
mshanemc: AFAICT the return of this ( fsPath:string) is only used for the test assertion.

I'd like it to return a URI (the test can asssert fsPath if that helps) so that if anything laters decides to use the output, it stays in URI land instead of getting into string paths.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataFileService.ts
```diff
@@ -48,39 +48,37 @@ export class QueryDataFileService {
...
+          yield* api.services.FsService.writeFile(fileInfo, fileContentString);
+          const { fsPath } = yield* api.services.WorkspaceService.getWorkspaceInfoOrThrow();
+          return fsPath;
+        })
+      );
+      showFileInExplorer(fileInfo, workspacePath);
+      showSaveSuccessMessage(Utils.basename(fileInfo));
       return selectedFileSavePath;
     }
     return '';
```
mshanemc: let's get rid of this pattern.  If something might not return a value, the return type should be `| undefined`.  empty string to represent "not there" is an anti-pattern.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -90,64 +98,84 @@ export class QueryDataViewService {
...
-      trackErrorWithTelemetry('data_view_save', message);
+      default:
+        return Effect.sync(() => channelService.appendLine(nls.localize('error_unknown_error', type))).pipe(
+          Effect.withSpan('QueryDataView.unknown_message', { attributes: { messageType: type } })
+        );
     }
-  }
+  };
+
+  private handleSaveRecordsEffect = (format: FileFormat): Effect.Effect<void> => {
```
mshanemc, line 151: let's not put return types on Effects (I had an eslint for that, but I don't think it applies to class methods).  This one is actually wrong because fileService.save returns a string.  And the return is weird on the the error.

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -90,64 +98,84 @@ export class QueryDataViewService {
...
+    const self = this;
+    return Effect.gen(function* () {
+      const fileService = new FileService(self.queryText, self.queryData, format, self.document);
+      yield* Effect.promise(() => fileService.save());
+    }).pipe(
+      Effect.catchAllCause(() =>
+        Effect.sync(() => vscode.window.showErrorMessage(nls.localize('error_data_view_save')))
+      )
+    );
+  };
```
mshanemc, line 161: I think this should be an Effect.fn with a 4 param obj instead of a method

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -90,64 +98,84 @@ export class QueryDataViewService {
...
 
   protected async getWebViewContent(webview: vscode.Webview): Promise<string> {
```
mshanemc, line 163: same for this "method"

it really makes to sense for it to be protected if nothing is extending the class, right?

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -38,7 +40,7 @@ type DataViewEvent = {
 export class QueryDataViewService {
   public currentPanel: vscode.WebviewPanel | undefined = undefined;
   public readonly viewType = QUERY_DATA_VIEW_TYPE;
```
mshanemc, line 42: nothing uses this, or modified it.  I think it's a module-level const

packages/salesforcedx-vscode-soql/src/queryDataView/queryDataViewService.ts
```diff
@@ -90,64 +98,84 @@ export class QueryDataViewService {
...
+    const salesforceCloudUri = Utils.joinPath(extensionUri, IMAGES_DIR_NAME, 'Salesforce_Cloud.png');
     this.currentPanel.iconPath = {
       light: salesforceCloudUri,
       dark: salesforceCloudUri
     };
 
     this.currentPanel.webview.html = await this.getWebViewContent(this.currentPanel.webview);
 
-    this.currentPanel.webview.onDidReceiveMessage(this.onDidRecieveMessageHandler, this, this.subscriptions);
+    // Stream-based message handling: each message dispatched as a named OTel span
```
mshanemc, line 109: that's clever.

PR #6959, opened 2026-03-09
fix : @W-21462748 - o11y single copy and core api for o11y service.

smuduganti-sf:
### What does this PR do?

- Ensures a single copy of @salesforce/o11y-reporter is used within the VS Code extension host to prevent duplicate registerInstrumentedApp() errors.
- Introduces getO11yService(extensionId) in salesforcedx-vscode-core so extensions obtain O11yService via the Core API instead of bundling their own copy.
- This centralizes telemetry initialization and ensures shared resources are safely reused across extensions.

@W-21462748@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "wireit": {
       "brace-expansion": "4.0.1"
     },
-    "@salesforce/o11y-reporter": "1.8.2"
+    "@salesforce/o11y-reporter": "^1.8.2"
   },
   "engines": {
     "node": ">=22.15.1"
@@ -17,7 +17,7 @@
   "dependencies": {
     "@actions/core": "^1.11.0",
     "@actions/github": "^6.0.0",
-    "@salesforce/o11y-reporter": "1.8.2",
+    "@salesforce/o11y-reporter": "^1.8.2",
     "@salesforce/vscode-service-provider": "1.5.0",
     "jsforce": "^3.10.10",
     "memfs": "^4.17.2",
--- a/packages/salesforcedx-utils-vscode/package.json
+++ b/packages/salesforcedx-utils-vscode/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@salesforce/core": "^8.23.3",
     "@salesforce/effect-ext-utils": "*",
-    "@salesforce/o11y-reporter": "1.8.2",
+    "@salesforce/o11y-reporter": "^1.8.2",
     "@salesforce/salesforcedx-utils": "*",
     "@salesforce/source-tracking": "^7.8.0",
     "@salesforce/vscode-i18n": "*",
--- a/packages/salesforcedx-vscode-core/package.json
+++ b/packages/salesforcedx-vscode-core/package.json
@@ -28,7 +28,7 @@
   ],
   "dependencies": {
     "@salesforce/core": "^8.24.0",
-    "@salesforce/o11y-reporter": "1.8.2",
+    "@salesforce/o11y-reporter": "^1.8.2",
     "@salesforce/effect-ext-utils": "*",
     "@salesforce/salesforcedx-utils": "*",
     "@salesforce/salesforcedx-utils-vscode": "*",
@@ -67,7 +67,7 @@
     "packageUpdates": {
       "main": "dist/src/index.js",
       "dependencies": {
-        "@salesforce/o11y-reporter": "1.8.2",
+        "@salesforce/o11y-reporter": "^1.8.2",
         "applicationinsights": "1.0.7"
       },
       "devDependencies": {}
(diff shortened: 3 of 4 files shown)


packages/salesforcedx-vscode-services/src/observability/o11ySpanExporter.ts
```diff
@@ -56,6 +73,20 @@ export class O11ySpanExporter implements SpanExporter {
       return this.initPromise;
     }
     this.initPromise = (async () => {
+      const ext = vscode.extensions.getExtension(SALESFORCE_VSCODE_CORE_EXTENSION_ID);
```
mshanemc, line 81: that's not going to work.

there are several conditions
1. on desktop, both extensions will be present and running simultaneously
2. on web, the Core extension is never present.  We still want telemetry

why can't there be 2 o11y services, if they're receiving different data from different extensions?  I understand that o11y throws in that situation, but why does it need to?
---

another possibility is

look for o11y from Core ext, use that if present
if Core extension is not present, create the o11y services from within services ext

packages/salesforcedx-vscode-services/src/observability/o11ySpanExporter.ts
```diff
@@ -56,6 +73,20 @@ export class O11ySpanExporter implements SpanExporter {
...
+      if (!ext) {
+        console.warn(
+          'O11ySpanExporter: salesforcedx-vscode-core not available; span export will be no-op. Ensure the Core extension is installed.'
+        );
+        this.initialized = true;
+        return;
+      }
+      const api = await ext.activate();
```
mshanemc: I don't think you want to call activate on it again blindly in case it's already active.  There is an isActive bool on it that you can check to see if you should

packages/salesforcedx-vscode-services/src/observability/o11ySpanExporter.ts
```diff
@@ -90,15 +125,15 @@ export class O11ySpanExporter implements SpanExporter {
             };
 
             if (success) {
-              this.o11yService.logEvent({
+              this.o11yService!.logEvent({
```
mshanemc: stop fighting the compiler and let it help you.

If it really can be undefined, saying the conditionals (ex: `if (success)`) should be `if (success && this.o11yService) otherwise bad things happen with that `undefined`

packages/salesforcedx-vscode-services/src/observability/o11ySpanExporter.ts
smuduganti-sf: The o11y SDK is designed around a single app registration per process, with shared global resources.

Option to look up the O11y service from the Core extension and reuse that instance when available sounds reasonable. If Core isn’t installed, the Services extension can fall back to creating its own O11y service.

mshanemc: I think it's going to need to go the other way

https://github.com/forcedotcom/salesforcedx-vscode/blob/b998aa24fad6038022261cf82c6c5837442f69ea/packages/salesforcedx-vscode-core/package.json#L51-L52

so core can depend on services, but not other way around (because core won't be there on web).

I'm going to have to think about the least bad way to do this.

smuduganti-sf: My initial thought was to keep utils-vscode as the owner for both AppInsights and o11y-reporter,
since it already acts as the shared utilities layer for telemetry-related concerns. That way, the telemetry libraries and reporter setup live in one place, and other extensions consume them via utils/Core APIs rather than introducing additional direct dependencies.

That said, considering the Core → Services dependency direction and the web scenario where Core isn’t present, it probably makes sense for Services to attempt using Core when available (desktop) and fall back to its own initialization when Core isn’t installed (web).

mshanemc: FYI, we plan to drop the Core dependency on desktop in the medium term in favor of services, utils-vscode will also stop being used by default unless some other extension outside this repo needs it.

I think a bunch of the log noise is coming from the way the services extension is built (the services ext hosts one reporter per extension) because extensions can have different package.json-based configuration (endpoint, pft).  That was done without the knowledge of o11y's "there can only be one" architecture.

gonna have to rethink all of this.

smuduganti-sf: Thanks for the additional context. In the current implementation, each extension can maintain its own O11yService/reporter instance, but the underlying o11y instrumentation app and collectors are shared across the extension host. The SDK effectively allows multiple reporters, but only a single registerInstrumentedApp() per process, which is why the shared initialization is serialized and reused.

Given the direction you mentioned (moving toward Services being the common layer), another option could be to expose getO11yService from the Services extension and make it the single owner of the shared initialization. Extensions could then obtain their instance through Services rather than initializing independently. Happy to adjust the implementation in that direction if it aligns better with the longer-term plan.

PR #6965, opened 2026-03-09
[W-21452992] feat: setup vscode-languageclient/browser

daphne-sfdc:
### What does this PR do?
Setup `vscode-languageclient/browser` so that the SOQL extension can run in the web.

@W-21452992@

### Testing
In soql-language-server (**daphne/W-21495820-setup-vscode-languageserver-browser** branch)
1. `yarn install`
2. `yarn build`
3. `yarn link`

In salesforcedx-vscode (**daphne/W-21452992-setup-vscode-languageclient-browser** branch)
1. `npm run reinstall` -> `npm run compile`
2. `yarn link @salesforce/soql-language-server`
3. `WIREIT_CACHE=none npm run vscode:bundle -w salesforcedx-vscode-services`
4. `WIREIT_CACHE=none npm run vscode:bundle -w salesforcedx-vscode-metadata`
5. `WIREIT_CACHE=none npm run vscode:bundle -w salesforcedx-vscode-soql`
6. `npm run run:web -w salesforcedx-vscode-soql`

Confirm the following work as expected in web:
* Lazy loading works properly (Open command palette immediately after starting up the web worker and be able to see **SFDX: Create Query in SOQL Builder**) ✅
* SFDX: Execute SOQL Query... ✅
* SFDX: Execute SOQL Query With Currently Selected Text ✅
* SFDX: Create Query in SOQL Builder ✅
* SOQL Builder works ✅
* Can save the SOQL query ✅

In a .soql file
- Autocompletion ✅
- Correct coloring ✅

### DIFF
--- a/packages/playwright-vscode-ext/src/utils/helpers.ts
+++ b/packages/playwright-vscode-ext/src/utils/helpers.ts
@@ -281,21 +281,17 @@ export const enableMonacoAutoClosing = async (page: Page): Promise<void> => {
 };
 
 /**
- * Wait for VS Code extensions to finish activating by watching the
- * "Developer: Show Running Extensions" editor. More reliable than polling
+ * Wait for all VS Code extensions to finish activating by watching the
+ * "Developer: Show Running Extensions" editor.  More reliable than polling
  * the command palette, especially on slow CI runners (e.g. Windows).
  *
  * While an extension is activating its row contains the text "Activating".
  * Once done the row shows "Activation: Xms" / "Startup Activation: Xms".
+ * We wait until no rows contain "Activating" any more.
  *
- * @param extensionIds - Substrings of extension names/IDs to wait for (e.g. 'salesforcedx-vscode-services'). If omitted or empty, waits for ALL extensions to finish activating.
- * @param timeout - Maximum ms to wait per extension (default 120 000).
+ * @param timeout - Maximum ms to wait for all extensions to activate (default 120 000).
  */
-export const waitForExtensionsActivated = async (
-  page: Page,
-  extensionIds: string[] = [],
-  timeout = 120_000
-): Promise<void> => {
+export const waitForExtensionsActivated = async (page: Page, timeout = 120_000): Promise<void> => {
   await executeCommandWithCommandPalette(page, 'Developer: Show Running Extensions');
 
   // The editor container gets class "runtime-extensions-editor" via createEditor()
@@ -306,18 +302,9 @@ export const waitForExtensionsActivated = async (
   const rows = editor.locator('.monaco-list-row');
   await expect(rows).not.toHaveCount(0, { timeout: 30_000 });
 
-  if (extensionIds.length === 0) {
-    // Wait until no row still contains "Activating" text
-    const stillActivating = rows.filter({ hasText: 'Activating' });
-    await expect(stillActivating).toHaveCount(0, { timeout });
-  } else {
-    for (const extensionId of extensionIds) {
-      const extensionRow = rows.filter({ hasText: extensionId });
-      await expect(extensionRow).not.toHaveCount(0, { timeout: 30_000 });
-      const stillActivating = extensionRow.filter({ hasText: 'Activating' });
-      await expect(stillActivating).toHaveCount(0, { timeout });
-    }
-  }
+  // Wait until no row still contains "Activating" text
+  const stillActivating = rows.filter({ hasText: 'Activating' });
+  await expect(stillActivating).toHaveCount(0, { timeout });
 
   // Close the Running Extensions tab via command palette (cross-platform, no hover needed)
   const tab = page.getByRole('tab', { name: /Running Extensions/i });
--- a/packages/salesforcedx-vscode-soql/test/playwright/specs/soqlBuilder.desktop.spec.ts
+++ b/packages/salesforcedx-vscode-soql/test/playwright/specs/soqlBuilder.desktop.spec.ts
@@ -25,7 +25,7 @@ test('SOQL Builder: create query and toggle between builder and text editor', as
 
   await test.step('setup workbench', async () => {
     await setupMinimalOrgAndAuth(page);
-    await waitForExtensionsActivated(page, ['salesforcedx-vscode-services', 'salesforcedx-vscode-metadata']);
+    await waitForExtensionsActivated(page);
     await ensureSecondarySideBarHidden(page);
     await saveScreenshot(page, 'setup.complete.png');
   });


packages/salesforcedx-vscode-services/src/index.ts
```diff
@@ -226,9 +226,16 @@ const activationEffect = (context: vscode.ExtensionContext) =>
         concurrency: 'unbounded'
       }
     );
-    // init the connection for all the consumers who might need it
-    // no Connection is a possible state
-    yield* Effect.fork(ConnectionService.getConnection().pipe(Effect.catchAll(() => Effect.void)));
+    // init the connection for all the consumers who might need it; no Connection is a possible state.
+    // In web, getConnection() only reads settings + creates in-memory objects (no network call), so it
```
mshanemc: that comment is false; it's going to make some network calls after it auths the first time to do things like
- get User.Id and User.username, orgId etc for telemetry and other uses

it's kind of async, but it will cause updates to the defaultOrgRef.

it'll return the connection pretty quickly but if you're relying on any of the side effects, it's better to subscribe to the defaultOrgRef for changes

packages/salesforcedx-vscode-services/src/index.ts
```diff
@@ -226,9 +226,16 @@ const activationEffect = (context: vscode.ExtensionContext) =>
...
+    // can be blocking — this ensures defaultOrgRef is populated before any dependent extension activates.
+    // On desktop it remains forked because it resolves the target org from disk/network.
+    const initConnection = ConnectionService.getConnection().pipe(Effect.catchAll(() => Effect.void));
+    yield* process.env.ESBUILD_PLATFORM === 'web' ? initConnection : Effect.fork(initConnection);
+    // set sf:project_opened context before activation resolves so lazy-loaded extensions can show
+    // their commands on startup — must be blocking (not forked) so the context key is set before
+    // VS Code evaluates `when` clauses for command palette visibility
+    yield* ProjectService.isSalesforceProject().pipe(Effect.catchAll(() => Effect.void));
```
mshanemc: is that really necessary?  at least for the web, the project will always load and cause the metdata/log to load, too.

For desktop, was it hiding the commands?

packages/salesforcedx-vscode-soql/esbuild.config.mjs
```diff
@@ -37,10 +39,39 @@ await build({
...
+if (existsSync(serverWorkerPath)) {
+  await build({
+    ...commonConfigBrowser,
+    format: 'iife', // Workers run as plain browser scripts — no module system, no `exports`
+    entryPoints: [serverWorkerPath],
+    outfile: './dist/serverWorker.js'
+  });
+} else {
+  console.warn(
+    '⚠️  Skipping serverWorker.js bundle: @salesforce/soql-language-server does not yet export lib/serverWorker.js. npm link the local repo to build the web LSP worker.'
```
mshanemc: I think that'll never show up in the build scripts, seems dangerous.  Maybe throw an error there?  What was the reason for the warning (other PR isn't merged yet and this should go away when it does?)

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -88,6 +87,7 @@
     "build:soql-builder-ui": "wireit",
     "test": "wireit",
     "test:soql-builder-ui": "wireit",
+    "run:web": "wireit",
```
mshanemc, line 89: ooh, also add soql to the top-level `run:web` that does all the web ext together.

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -173,6 +174,15 @@
         "dist"
       ]
     },
+    "run:web": {
+      "command": "npx vscode-test-web --browserType=chromium --browserOption=--disable-web-security --browserOption=--remote-debugging-port=9222 --extensionDevelopmentPath . --extensionPath ../salesforcedx-vscode-services --open-devtools --port 3002 --quality stable --verbose --printServerLog",
```
mshanemc: ahh, that's probably why the connection/project stuff seemed necessary.  Bringing in metadata ext (which always runs) would solve it.

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
```diff
@@ -4,45 +4,48 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { ExtensionContext, workspace } from 'vscode';
+import { ExtensionContext, Uri, workspace } from 'vscode';
+import type { BaseLanguageClient } from 'vscode-languageclient';
+import type { LanguageClient as BrowserLanguageClient } from 'vscode-languageclient/browser';
 import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from 'vscode-languageclient/node';
```
kylewalke, line 10: @daphne-sfdc Were there no bundling issues you ran into using both imports of node and browser submodules?

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -76,7 +76,6 @@
     "salesforcedx-vscode-services": "*"
   },
   "extensionDependencies": [
-    "salesforce.salesforcedx-vscode-core",
```
mshanemc, line 79: yay!

packages/salesforcedx-vscode-soql/package.json
```diff
@@ -76,7 +76,6 @@
     "salesforcedx-vscode-services": "*"
   },
   "extensionDependencies": [
```
mshanemc, line 77: since you dropped the dep on vscode-core, you can probably
1. remove devDep
2. remove the tsconfig aths
3. remove any wireit deps

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
```diff
@@ -4,45 +4,48 @@
...
-
-  // Create the language client and start the client.
-  client = new LanguageClient('soql-language-server', 'SOQL Language Server', serverOptions, clientOptions);
+export const startLanguageClient = async (extensionContext: ExtensionContext): Promise<void> => {
+  if (process.env.ESBUILD_PLATFORM === 'web') {
+    // In the web bundle, esbuild aliases vscode-languageclient/node -> /browser, so LanguageClient
+    // is the browser version at runtime. Cast to get the correct browser constructor signature.
+    const workerUri = Uri.joinPath(extensionContext.extensionUri, 'dist', 'serverWorker.js');
+    const worker = new Worker(workerUri.toString(true));
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
mshanemc, line 31: yeah, I see that the (not stable release yet) v10 has proper exports so we can get rid of this whenever that's available.

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
```diff
@@ -4,45 +4,48 @@
...
-      configurationSection: 'soql',
-      fileEvents: workspace.createFileSystemWatcher('**/*.soql')
-    },
-    middleware: codeCompletion.middleware
-  };
-
-  // Create the language client and start the client.
-  client = new LanguageClient('soql-language-server', 'SOQL Language Server', serverOptions, clientOptions);
+export const startLanguageClient = async (extensionContext: ExtensionContext): Promise<void> => {
+  if (process.env.ESBUILD_PLATFORM === 'web') {
```
kylewalke, line 26: browser doesn't have access to process unless you polyfilled this elsewhere. Does this work when you've tested it with the test-web package?

kylewalke (commented):
Just a few questions but otherwise looks pretty good 👍

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
mshanemc: this is an esbuild `define` var.  It lets you drop code during bundle time based on environment variables.  Won't appear in the bundle for runtime.

we do this in various ext and in some of the libraries to avoid bundling unnecessary code.

If you haven't used this trick, you're going to love it.

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
mshanemc: that's the `process.env` is there to drop the unnecessary imports.
see below

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
mshanemc: docs https://esbuild.github.io/api/#define

bundle-time `define`
https://github.com/forcedotcom/salesforcedx-vscode/blob/b998aa24fad6038022261cf82c6c5837442f69ea/scripts/bundling/web.mjs#L58

packages/salesforcedx-vscode-soql/package.json
daphne-sfdc: Yes, just did that and you're right!  Workaround in index.ts reverted.

packages/salesforcedx-vscode-soql/esbuild.config.mjs
daphne-sfdc: Good catch!  That was only for the linking - and actually not even needed for the linking!  Fixed to be the same format as desktop and apex-testing.

packages/salesforcedx-vscode-soql/src/lspClient/client.ts
daphne-sfdc: Actually it looks like ESLint v10 became stable on 2/6/2026!
https://eslint.org/blog/2026/02/eslint-v10.0.0-released
Screenshot 2026-03-10 at 12 47 09 PM

daphne-sfdc: But it contains some breaking changes, so we'll need to do the migration in a separate GUS WI, not here 🙂

daphne-sfdc: New GUS WI here: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002W3FHGYA3/view

mshanemc: I was kinda vague.  v10 of vscode-languageclient which has better exports/types

packages/salesforcedx-vscode-services/src/index.ts
daphne-sfdc: Apparently yes this is necessary for web.  The change for connection was truly unnecessary after adding the dependency on the Metadata extension, but removing this line causes the lazy loading to not work properly - the **SFDX: Create Query in SOQL Builder** command does not show up unless you force the SOQL extension to load via opening the SOQL Builder via the walkthrough first.

`sf:project_opened` is only set as a side effect of calling `ProjectService.isSalesforceProject()`, and Metadata's activation never calls it.

PR #6966, opened 2026-03-09
refactor(services): use new findFiles fsService W-21220952

mshanemc:
### What does this PR do?
- Bump @salesforce/core, use new findFiles from fsService
- Replace let with const

@W-21220952@

### DIFF
--- a/packages/salesforcedx-vscode-services/src/virtualFsProvider/fileSystemProvider.ts
+++ b/packages/salesforcedx-vscode-services/src/virtualFsProvider/fileSystemProvider.ts
@@ -188,7 +188,13 @@ export class FsProvider implements vscode.FileSystemProvider {
       vscode.workspace.workspaceFolders?.[0]?.uri;
     if (!baseUri) return [];
     const excludeArr = exclude == null ? undefined : typeof exclude === 'string' ? [exclude] : [exclude.pattern];
-    const matches = await fs.promises.glob(pattern, { cwd: baseUri.path, exclude: excludeArr });
+    // Only runs on web (memfs); node:fs.glob returns AsyncIterator but we never hit that path
+    // this is fixed in higher memfs versions, but there's other conflicts there (would break sfdx-core support for node 20)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- web-only, memfs returns Promise<string[]>
+    const matches = await (fs.promises.glob(pattern, {
+      cwd: baseUri.path,
+      exclude: excludeArr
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-services/src/virtualFsProvider/fileSystemProvider.ts
```diff
@@ -17,35 +17,36 @@ import * as Layer from 'effect/Layer';
...
   if (metaMatch) return metaMatch[1];
-  const ext = path.extname(fsPath).slice(1);
+  const ext = Utils.extname(uri).slice(1);
   return ext || undefined;
 };
 
 /** Effect that uses MetadataRegistryService (cached) to resolve metadata type from URI */
-export const isItReadOnlyEffect = Effect.fn('isItReadOnly')(function* (readOnlyTypes: MetadataType[], uri: vscode.Uri) {
+export const isItReadOnlyEffect = Effect.fn('isItReadOnly')(function* (readOnlyTypes: MetadataType[], uri: URI) {
   if (readOnlyTypes.length === 0) return false;
```
peternhale: @mshanemc 
```suggestion
export const isItReadOnly = Effect.fn('isItReadOnly')(function* (readOnlyTypes: MetadataType[], uri: URI) {
  if (readOnlyTypes.length === 0) return false;
```

packages/salesforcedx-vscode-services/src/virtualFsProvider/fileSystemProvider.ts
```diff
@@ -168,7 +169,29 @@ export class FsProvider implements vscode.FileSystemProvider {
     ]);
   }
 
-  public watch(_uri: vscode.Uri, _options: { recursive: boolean; excludes: string[] }): vscode.Disposable {
+  public watch(_uri: URI, _options: { recursive: boolean; excludes: string[] }): vscode.Disposable {
     return new vscode.Disposable(() => {});
   }
+
+  /** Find files by glob. baseUri optional: when provided (via RelativePattern), search under that URI only; otherwise use workspace folder. */
+  public async findFiles(
```
peternhale, line 180: The findFiles signature also accepts a cancellation token. Given the implementation, I can see that cancelling an inflight call is not possible.

Maybe add a note?

mshanemc: yeah, I had considered keeping it for consistency and then just making it do nothing but didn't like that either.

jsdoc comment is a good idea.

PR #6981, opened 2026-03-11
[W-21532592] feat: SFDX: Create Query in SOQL Builder asks for filename before creating

daphne-sfdc:
### Before
**SFDX: Create Query in SOQL Builder** produces an unsaved file **untitled.soql**.

Screenshot 2026-03-11 at 4 03 11 PM

### After
**SFDX: Create Query in SOQL Builder** asks for the filename and directory to save the query in, like all the other create commands.

Screenshot 2026-03-11 at 4 05 11 PM

Screenshot 2026-03-11 at 4 06 22 PM<br />

If the filename already exists in the selected directory, you'll see this popup:

Screenshot 2026-03-11 at 4 07 24 PM

## What issues does this PR fix or reference?
@W-21532592@

## Testing
- File name does not exist ✅
- File name already exists -> overwrite ✅
- File name already exists -> cancel ✅
- Choose custom directory ✅

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
+++ b/packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
@@ -36,16 +36,19 @@ const promptForOutputDir = Effect.fn('soqlFileCreate.promptForOutputDir')(functi
   const api = yield* (yield* ExtensionProviderService).getServicesApi;
   const workspaceInfo = yield* api.services.WorkspaceService.getWorkspaceInfoOrThrow();
 
-  type DirItem = { label: string; description: string | undefined; uri: URI | undefined };
-
-  const rootItem: DirItem = { label: workspaceInfo.uri.fsPath, description: '(default)', uri: workspaceInfo.uri };
-  const customItem: DirItem = { label: CUSTOM_DIR_LABEL, description: undefined, uri: undefined };
+  const defaultUri = Utils.joinPath(workspaceInfo.uri, 'scripts', 'soql');
 
   const selected = yield* Effect.promise(() =>
-    vscode.window.showQuickPick([rootItem, customItem], {
-      placeHolder: nls.localize('soql_output_dir_prompt'),
-      matchOnDescription: true
-    })
+    vscode.window.showQuickPick(
+      [
+        { label: defaultUri.fsPath, description: '(default)', uri: defaultUri },
+        { label: CUSTOM_DIR_LABEL, description: undefined, uri: undefined }
+      ],
+      {
+        placeHolder: nls.localize('soql_output_dir_prompt'),
+        matchOnDescription: true
+      }
+    )
   );
 
   if (!selected) return undefined;
@@ -83,10 +86,7 @@ const confirmOverwrite = Effect.fn('soqlFileCreate.confirmOverwrite')(function* 
   return choice === nls.localize('soql_overwrite_button');
 });
 
-const createAndOpenFile = Effect.fn('soqlFileCreate.createAndOpenFile')(function* (
-  fileName: string,
-  outputDir: URI
-) {
+const createAndOpenFile = Effect.fn('soqlFileCreate.createAndOpenFile')(function* (fileName: string, outputDir: URI) {
   const api = yield* (yield* ExtensionProviderService).getServicesApi;
   const fileUri = Utils.joinPath(outputDir, `${fileName}.soql`);
 


packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -5,18 +5,113 @@
...
+
+const CUSTOM_DIR_LABEL = `$(file-directory) ${nls.localize('soql_custom_output_directory')}`;
+
+const promptForOutputDir = Effect.fn('soqlFileCreate.promptForOutputDir')(function* (project: SfProject) {
+  const api = yield* (yield* ExtensionProviderService).getServicesApi;
+  const workspaceInfo = yield* api.services.WorkspaceService.getWorkspaceInfoOrThrow();
+
+  type DirItem = { label: string; description: string | undefined; uri: URI | undefined };
+
+  const pkgItems: DirItem[] = project.getPackageDirectories().map(pkg => ({
```
mshanemc: this was probably based on cmds that create metadata, but soql should live outside the project's package directories since you don't deploy them

- It kinda works ok becuase I think SDR chooses to ignore them
`/scripts/soql` (at project root) would be a good default place (and hardcoding it will make this a bit simpler).  This aligns with our docs https://developer.salesforce.com/docs/platform/sfvscode-extensions/guide/soql-writing.html and how the templates directory behaves by default for project templates.

other note: you might need to create the dir if not already present.

For web, you **could** also add it to fileSystemSetup's code so that it's always created

mshanemc: otherwise, you also get the weird UI where right clicking in the file explorer gives you deploy options becuase inPackageDirectories is true, deployOnSave will try to, etc.

daphne-sfdc: I moved the default directory for **SFDX: Create Query in SOQL Builder** to the root directory of the project.  However, that doesn't make those deploy, diff, etc. commands disappear from the context menu.  I remember we had a work item to fix that, somewhere in our backlog?

As for SOQL queries created inside `force-app` - I did try deploying a folder containing SOQL queries, and the queries are just ignored.  It's the exact same idea as Anonymous Apex files located inside the `classes` folder.  So it's already handled if the user still chooses a custom directory and selects something inside `force-app`.

mshanemc: that contextMenu stuff **should** be fixed in the metadata ext, so it should work correctly with `run:web` and on desktop if you have the "use new metadata commands" setting enabled.

because of `sf:in_package_directories`

mshanemc: yep, those deploy hides are working ok outside pkgDirs. (I didn't restrict the manifest command because that wasn't specific to pkgDirs before and it'll probably mess up someone who's doing mdapi-based stuff).

Screenshot 2026-03-12 at 7 38 52 AM


I'm going to do a commit for this part
>/scripts/soql (at project root) would be a good default place (and hardcoding it will make this a bit simpler). This aligns with our docs https://developer.salesforce.com/docs/platform/sfvscode-extensions/guide/soql-writing.html

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
```diff
@@ -5,18 +5,104 @@
...
+
+  const rootItem: DirItem = { label: workspaceInfo.uri.fsPath, description: '(default)', uri: workspaceInfo.uri };
+  const customItem: DirItem = { label: CUSTOM_DIR_LABEL, description: undefined, uri: undefined };
+
+  const selected = yield* Effect.promise(() =>
+    vscode.window.showQuickPick([rootItem, customItem], {
+      placeHolder: nls.localize('soql_output_dir_prompt'),
+      matchOnDescription: true
+    })
+  );
```
mshanemc, line 52: let's avoid defining types for the stuff vscode owns.  If you really need the type, import it as a type from `vscode`.

It's fine to omit them here becuase when yoou pass them to showQuickPick the compiler will validate that you're passing correctly typed inputs.

packages/salesforcedx-vscode-soql/src/commands/soqlFileCreate.ts
daphne-sfdc: Ah, I didn't have the "Use Metadata Extension Commands" setting checked!  Yes, after checking that box, those commands disappeared from the context menu. 👍

PR #6992, opened 2026-03-13
fix(org): fix org logout, alias management, and org picker UX W-21567195

peternhale:
### What does this PR do?

Fixes several stale-cache bugs in org logout and auth flows, and improves the org picker UX for Display, Delete, and Logout commands.

**Bug fixes:**

- `orgLogout` now unsets `target-org` config and removes all disk aliases (including aliases added after the default org was set) for the logged-out username
- `orgLogout` (selected orgs) now correctly detects and unsets alias-based `target-org` and `target-dev-hub` configs — previously an empty alias array was passed to the config checks, causing both to be silently skipped
- Org picker after scratch org create now shows the correct icon and alias — was broken by a stale `StateAggregator` cache; fixed by reading aliases directly from disk and calling `updateConfigAndStateAggregators()` post-create
- `target-dev-hub` is now recognized after a logout/re-auth cycle — `DevUsernameChecker` now uses `ConfigService.getTargetDevHub()` (services extension) which always reloads from disk, bypassing the stale `@salesforce/core` `ConfigAggregator` static singleton
- Adds an `AliasFileWatcherService` reactor so that the status bar always shows the original primary alias when multiple aliases point to the same org

**UX improvements:**

- **SFDX: Display Org Details…** — replaced free-text input with an org QuickPick
- **SFDX: Delete Org** — org picker is now a multi-select filtered to scratch orgs and sandboxes only (the only deletable types); confirmation modal before delete; each deleted org has its auth, aliases, `target-org`, and `target-dev-hub` cleaned up
- **SFDX: Log Out from Authorized Orgs** — replaced `sf org logout --all` CLI call with a multi-select org QuickPick, per-org cleanup (auth + aliases + config), and a context-aware confirmation dialog (scratch/sandbox language when applicable); command renamed from "Log Out from All Authorized Orgs"
- **SFDX: Display Org Information** now shows all aliases for an org instead of just the first one

**Services extension:**

- `AliasService`: `getAliasesFromUsername`, `unsetAliases` — disk-based alias operations bypassing stale `StateAggregator`
- `ConfigService`: `getTargetDevHub`, `isCurrentTargetDevHub`, `unsetTargetDevHub` — fresh dev hub config operations
- `AliasFileWatcherService`: general-purpose watcher for `~/.sf/alias.json` changes

@W-21567195@

### DIFF
--- a/packages/salesforcedx-vscode-org/src/util/orgUtil.ts
+++ b/packages/salesforcedx-vscode-org/src/util/orgUtil.ts
@@ -102,8 +102,9 @@ export const updateConfigAndStateAggregators = async (): Promise<void> => {
   const configAggregatorProvider = ConfigAggregatorProvider.getInstance();
   await configAggregatorProvider.reloadConfigAggregators();
   // Also force the StateAggregator to reload to have the latest
-  // authorization info.
-  StateAggregator.clearInstance(workspaceUtils.getRootWorkspacePath());
+  // authorization info. Called without args to clear ALL cached instances,
+  // including the default one used by AuthInfo.listAllAuthorizations().
+  await StateAggregator.clearInstanceAsync();
 
   // Trigger Apex Test Controller to discover tests after org auth/set-default. Delay so config
   // and TargetOrgRef can propagate before refresh runs.
--- a/packages/salesforcedx-vscode-org/test/jest/util/orgUtil.test.ts
+++ b/packages/salesforcedx-vscode-org/test/jest/util/orgUtil.test.ts
@@ -348,7 +348,7 @@ describe('testing setTargetOrgOrAlias', () => {
     mockConfigAggregatorProvider = jest
       .spyOn(ConfigAggregatorProvider, 'getInstance')
       .mockReturnValue(mockConfigAggregatorProviderInstance as any);
-    stateAggregatorClearInstanceMock = jest.spyOn(StateAggregator, 'clearInstance');
+    stateAggregatorClearInstanceMock = jest.spyOn(StateAggregator, 'clearInstanceAsync').mockResolvedValue();
   });
 
   it('should set provided username or alias as default configs', async () => {
@@ -362,7 +362,7 @@ describe('testing setTargetOrgOrAlias', () => {
   it('should change the current working directory to the original working directory', async () => {
     const username = 'vscodeO';
     await setTargetOrgOrAlias(username);
-    expect(workspacePathStub).toHaveBeenCalledTimes(2);
+    expect(workspacePathStub).toHaveBeenCalledTimes(1);
     expect(chdirStub).toHaveBeenCalledTimes(2);
     expect(chdirStub).toHaveBeenNthCalledWith(1, fakeWorkspace);
     expect(chdirStub).toHaveBeenNthCalledWith(2, fakeOriginalDirectory);


packages/salesforcedx-vscode-org/src/util/orgUtil.ts
```diff
@@ -278,6 +279,44 @@ type DefaultOrgConfig = {
   defaultOrgUsername: string | undefined;
 };
 
+/** Read alias.json from disk, bypassing StateAggregator cache. Returns alias → username map. */
```
mshanemc: see Alias in services ext

packages/salesforcedx-vscode-services/src/core/alias.ts
```diff
@@ -59,6 +59,14 @@ export class AliasService extends Effect.Service<AliasService>()('AliasService',
       );
     });
 
-    return { getAllAliases, getAliasesFromUsername, getUsernameFromAlias };
+    /** Remove aliases by name. Reads from disk before writing to handle same-session aliases. */
```
mshanemc: I don't think the comment is true

packages/salesforcedx-vscode-services/test/jest/core/aliasFileWatcher.test.ts
```diff
@@ -0,0 +1,180 @@
...
+// ---------------------------------------------------------------------------
+// AliasFileWatcherService – path-filtering integration tests
+// ---------------------------------------------------------------------------
+
+describe('AliasFileWatcherService', () => {
+  const runWithMockFileWatcher = (
+    fileWatcherPubSub: PubSub.PubSub<FileChangeEvent>,
+    test: (aliasWatcher: AliasFileWatcherService) => Effect.Effect<void, never, Scope.Scope>
+  ) => {
+    // AliasFileWatcherService.Default requires FileWatcherService externally (no baked-in Default).
```
mshanemc: why wouldn't you want to bake in the default?  What would a consumer pass in that would be unique?

packages/salesforcedx-vscode-org/src/util/orgUtil.ts
```diff
@@ -308,16 +347,18 @@ export const determineOrgMarkers = (orgAuth: OrgAuthorization, defaultConfig: De
 
   // Check if this org is the default DevHub (by property value or resolved username)
   const matchesDevHubProperty =
-    defaultConfig.defaultDevHubProperty && possibleDefaults.has(String(defaultConfig.defaultDevHubProperty));
+    defaultConfig.defaultDevHubProperty != null && possibleDefaults.has(String(defaultConfig.defaultDevHubProperty));
```
mshanemc, line 348: it shouldn't ever be `null`.  Why the change?

packages/salesforcedx-vscode-org/src/messages/i18n.ts
```diff
@@ -52,12 +52,19 @@ export const messages = {
   org_login_web_authorize_org_text: 'SFDX: Authorize an Org',
   org_login_web_verification_code_message: 'Verification Code: %s — If prompted, enter this code in your browser window.',
   org_login_web_verification_code_suffix: '(Verification Code: %s)',
-  org_logout_all_text: 'SFDX: Log Out from All Authorized Orgs',
+  org_logout_all_text: 'SFDX: Log Out from Authorized Orgs',
   org_logout_default_text: 'SFDX: Log Out from Default Org',
   org_logout_no_default_org: 'No default org to logout from',
+  org_delete_select_orgs_placeholder: 'Select scratch orgs and sandboxes to delete',
+  org_delete_confirm_prompt: 'Permanently delete %s org(s)? This cannot be undone.',
```
mshanemc: if you make this a %d instead of a %s then you don't have to convert number to string before passing do it

packages/salesforcedx-vscode-org/src/preconditionCheckers/devUsernameChecker.ts
```diff
@@ -5,13 +5,26 @@
...
+  const api = yield* (yield* ExtensionProviderService).getServicesApi;
+  return yield* api.services.ConfigService.getTargetDevHub();
+});
 
 /** Checks if a Dev Hub is configured */
 export class DevUsernameChecker implements PreconditionChecker {
   public async check(): Promise<boolean> {
-    const targetDevHubOrAlias = await getTargetDevHubOrAlias(true);
+    const targetDevHubOrAlias = await getTargetDevHub().pipe(
+      Effect.provide(AllServicesLayer),
```
mshanemc: I think you should construct another ManagedRuntime like that previous sobjects PR and use it through org extension

packages/salesforcedx-vscode-org/src/preconditionCheckers/devUsernameChecker.ts
```diff
@@ -5,13 +5,26 @@
...
+import * as Effect from 'effect/Effect';
+import { AllServicesLayer } from '../extensionProvider';
+
+const getTargetDevHub = Effect.fn('DevUsernameChecker.getTargetDevHub')(function* () {
+  const api = yield* (yield* ExtensionProviderService).getServicesApi;
+  return yield* api.services.ConfigService.getTargetDevHub();
+});
 
 /** Checks if a Dev Hub is configured */
 export class DevUsernameChecker implements PreconditionChecker {
```
mshanemc: I think, rather than propagating CompositePreconditionChecker pattern, you could do the check at the beginning of orgCreate fn

packages/salesforcedx-vscode-services/src/index.ts
```diff
@@ -83,6 +85,7 @@ export type SalesforceVSCodeServicesApi = {
       | TransmogrifierService
       | WorkspaceService
     >;
+    AliasFileWatcherService: typeof AliasFileWatcherService;
```
mshanemc: let's not export it in the API unless someone else needs it.  Right now, it's only consumed within services

packages/salesforcedx-vscode-org/src/orgPicker/orgList.ts
```diff
@@ -158,21 +163,40 @@ export const authorizationsToQuickPickItems = (
...
     .map(orgAuthToQuickPickItem(defaultConfig));
 
+/** Type guard: true when a QuickPickItem has an orgUsername property (i.e. is an OrgQuickPickItem). */
+export const isOrgItem = (item: vscode.QuickPickItem): item is OrgQuickPickItem => 'orgUsername' in item;
+
+/** Build grouped QuickPick items with type separators, with optional org-type filter. */
+export const buildOrgQuickPickItems = (
+  authorizations: OrgAuthorization[],
+  defaultConfig: DefaultOrgConfig,
+  filter?: (org: OrgAuthorization) => boolean
```
mshanemc, line 173: the 1 consumer of this that has a filter (delete) should do the filter first rather the making this more complex

packages/salesforcedx-vscode-org/src/commands/orgDelete.ts
```diff
@@ -5,49 +5,127 @@
...
-    const builder = new SfCommandBuilder()
+  public async run(response: ContinueResponse<{ orgs: OrgToDelete[] }>): Promise<boolean> {
+    const { orgs } = response.data;
+    try {
+      const authRemover = await AuthRemover.create();
+      let allSucceeded = true;
+      for (const { username, orgType } of orgs) {
+        // Fetch aliases before cleanup so they can be used for config matching
+        const aliases = await getAliasesForUsername(username).pipe(Effect.provide(AllServicesLayer), Effect.runPromise);
+        const success = await runDeleteCli(username, orgType);
```
mshanemc, line 109: is there something wrong with the org:delete command that it's not doing those things? (unsetting the configs and alias) that we also have to do them afterward?  I thought it did that via authRemover.  That's why we got rid of all this before.

I've deleted orgs before and see the UI (ex: org picker) react to it properly.

packages/salesforcedx-vscode-org/src/commands/auth/orgLogout.ts
```diff
@@ -40,70 +34,116 @@ class SimpleGatherer<T> implements ParametersGatherer<T> {
...
 
 export const orgLogoutAll = async () => {
-  const commandlet = new SfCommandlet(sfProjectPreconditionChecker, new EmptyParametersGatherer(), new OrgLogoutAll());
+  const commandlet = new SfCommandlet(sfProjectPreconditionChecker, new SelectOrgsForLogout(), new OrgLogoutSelected());
   await commandlet.run();
 };
 
-class OrgLogoutDefault extends LibraryCommandletExecutor<string> {
-  constructor() {
+const getAliasesForUsername = Effect.fn('OrgLogout.getAliasesForUsername')(function* (username: string) {
```
mshanemc, line 75: there seem to be a lot of copies of these.  If more of the `run` method is inside an effect, you could avoid creating all of these for each command.

packages/salesforcedx-vscode-services/src/index.ts
peternhale: Ok

packages/salesforcedx-vscode-org/src/commands/orgDelete.ts
peternhale: I have aliases
```
dreamhouselwcclean │ test-v9d2qslqm5xw@example.com 
b                  │ test-bvjl3aariyyc@example.com
a                  │ test-bvjl3aariyyc@example.com
```
With "SFDX: Delete Default Org" all aliases are removed properly

But not the case with "SFDX: Delete Org...", aliases "a" and "b" are not removed.

Just applying the same logic to ensure cleanup is complete.

mshanemc: that sounds like a core library bug we should fix if the multi-alias thing isn't cleaning up the aliases.  I'd rather approve that PR in sfdx-core than have extra "defensive" in multiple places in the extensions

packages/salesforcedx-vscode-org/src/util/orgUtil.ts
peternhale: Right but with 
```
const matchesDevHubProperty =
    defaultConfig.defaultDevHubProperty && possibleDefaults.has(String(defaultConfig.defaultDevHubProperty));
```
matchesDevHubProperty infers to `const matchesDevHubProperty: boolean | "" | undefined`

with `!= null`

matchesDevHubProperty infers to `const matchesDevHubProperty: boolean`
Seems safer to me.

packages/salesforcedx-vscode-org/src/commands/orgDelete.ts
peternhale: I peeked at that when I first found that all aliases for the same username were not being removed (org delete and logout). Not sure it will be a simple fix. Just feels like between auth info, config aggregators, alias management that there needs to be a more cohesive way to manage these states. I just don't see a single place where org related state is managed. Something the CLI team should consider, not us.

PR #7011, opened 2026-03-18
W-21616134: trigger CBW release after publishing extensions to marketplace

jonnyhork:
## Summary

- Add `trigger-cbw-release` job to `publishVSCode.yml` that fires a `repository_dispatch` event to `forcedotcom/code-builder-web` after extensions are successfully published to the MS Marketplace.
- Fire-and-forget — sends a single API call with the published version in the payload and does not block CTC close jobs.
- Uses existing `IDEE_GH_TOKEN` (scope updated to include `code-builder-web` repo).

Work item: @W-21616134@

**Companion PR**: In `forcedotcom/code-builder-web` — adds the `promote-on-publish.yml` receiver workflow and `auto-promote` support in `release.yml`.

## Test plan

- [ ] Verify `trigger-cbw-release` job runs after `publish` succeeds and sends the dispatch
- [ ] Verify CTC close jobs are unaffected (no new dependency)
- [x] Verify `IDEE_GH_TOKEN` has `repo` scope on `forcedotcom/code-builder-web`

### DIFF
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -84,9 +84,11 @@ jobs:
       changeCaseId: ${{ needs.ctc-open.outputs.changeCaseId }}
       status: Not Implemented
 
+  # Set repo variable CBW_TRIGGER_ENABLED=false to disable dispatch to code-builder-web.
+  # Default (unset): dispatch after publish. Use when publishing without CBW promotion.
   trigger-cbw-release:
     needs: [publish, prepare-environment-from-main]
-    if: needs.publish.result == 'success'
+    if: needs.publish.result == 'success' && vars.CBW_TRIGGER_ENABLED != 'false'
     continue-on-error: true
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/publishVSCode.yml
```diff
@@ -83,3 +83,24 @@ jobs:
     with:
       changeCaseId: ${{ needs.ctc-open.outputs.changeCaseId }}
       status: Not Implemented
+
+  # Set repo variable CBW_TRIGGER_ENABLED=false to disable dispatch to code-builder-web.
+  # Default (unset): dispatch after publish. Use when publishing without CBW promotion.
+  trigger-cbw-release:
+    needs: [publish, prepare-environment-from-main]
+    if: needs.publish.result == 'success' && vars.CBW_TRIGGER_ENABLED != 'false'
```
jonnyhork, line 91: If we ever want to publish to MS Marketplace without promoting CBW we can turn off the functionality by adding `CBW_TRIGGER_ENABLED=false` as a repo variable

PR #7030, opened 2026-03-20
test(e2e): fix flaky authentication and reduce e2e noise W-21680928

CristiCanizales:
### What does this PR do?
Fixes broken e2e tests and reduces CI noise from consistently failing test suites.

@W-21680928@

### Changes

**authentication.e2e.ts**
- Org picker label format changed from `alias - username` to codicon-prefixed labels with separate descriptions; `selectQuickPick` exact match no longer works
- Replaced with `findQuickPickItem` partial matching by alias
- Fixed expected SFDX commands check: use `includes` on aria-label instead of brittle `slice(5)`
- Fixed status bar assertion: use `getStatusBarItemWhichIncludes` instead of exact `getItem` match
- Added `reloadWindow` before setting scratch org as default so it appears in the org list

**snippets.e2e.ts**
- Added `reloadWindow` to let LWC Language Server index before triggering snippets
- Increased pause for autocompletion to appear
- Removed `tryToHideCopilot` (no longer needed)

**auraLsp.e2e.ts / lwcLsp.e2e.ts**
- Added explicit `openFile` before Go to Definition tests to ensure the file is active

**templates.e2e.ts**
- Collapse folders in Explorer before creating Sample Analytics Template to avoid UI clutter

**CI workflows**
- Commented out `createOasDoc` e2e test from all workflows (consistently broken)

**salesforcedx-vscode-metadata/package.json**
- Added `sf.metadata.lightning.generate.lwc` command to Explorer context menu

### DIFF
--- a/packages/salesforcedx-vscode-automation-tests/test/specs/auraLsp.e2e.ts
+++ b/packages/salesforcedx-vscode-automation-tests/test/specs/auraLsp.e2e.ts
@@ -20,7 +20,7 @@ import {
   reloadWindow
 } from '@salesforce/salesforcedx-vscode-test-tools/lib/src/ui-interaction';
 import { expect } from 'chai';
-import path from 'node:path';
+import * as path from 'node:path';
 import { By, after } from 'vscode-extension-tester';
 import { defaultExtensionConfigs } from '../testData/constants';
 import { getFolderPath } from '../utils/buildFilePathHelper';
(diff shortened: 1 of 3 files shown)


.github/workflows/apexE2E.yml
CristiCanizales, line 1: This has been broken forever, so just commenting it out while we decide what to do with it.

PR #7032, opened 2026-03-20
ci(e2e): org/core command presence desktop tests and orgE2E workflow W-21683506

mshanemc:
### What does this PR do?
- Remove RedHat `anInitialSuite.e2e.ts`
- Desktop Playwright: org commands + core empty-workspace SF commands
- New `orgE2E.yml`; `triggerE2EForCommit` wires org job
- playwright-vscode-ext fixture/helper + eslint hookup; docs touch

note: the apexClassCreate=>metadata PR has e2e to validate the presence/absence of those commands based on project.

@W-21683506@

### Functionality After

### DIFF
--- a/.github/workflows/orgE2E.yml
+++ b/.github/workflows/orgE2E.yml
@@ -1,0 +1,101 @@
+name: Org E2E (Playwright)
+
+on:
+  workflow_dispatch:
+  push:
+    branches-ignore: [main, develop]
+    paths-ignore:
+      - '.claude/**'
+      - '.cursor/**'
+      - 'contributing/**'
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-core/test/playwright/specs/emptyWorkspaceSfdxCommands.desktop.spec.ts
CristiCanizales, line 1: Here we should be testing that *only* the create project commands show up, and not any other, because these can also show up in a project, so it should be only these 2 + create project for isv, with a total of only 3 commands starting with "SFDX:"

PR #7035, opened 2026-03-21
fix(org): clear global StateAggregator cache so new orgs appear in picker W-21567195

CristiCanizales:
What does this PR do?
Fixes StateAggregator.clearInstance() in updateConfigAndStateAggregators to clear the correct cache instance. Previously it passed workspaceUtils.getRootWorkspacePath() as the cache key, which only cleared a workspace-keyed entry. AuthInfo.listAllAuthorizations() retrieves the StateAggregator via getInstance() (no args), which uses Global.DIR as the key — a different entry that remained stale. Calling clearInstance() without args matches Global.DIR and clears the instance that listAllAuthorizations depends on.

@W-21567195@

### DIFF
--- a/packages/salesforcedx-vscode-org/src/util/orgUtil.ts
+++ b/packages/salesforcedx-vscode-org/src/util/orgUtil.ts
@@ -104,7 +104,7 @@ export const updateConfigAndStateAggregators = async (): Promise<void> => {
   // Also force the StateAggregator to reload to have the latest
   // authorization info. Called without args to clear ALL cached instances,
   // including the default one used by AuthInfo.listAllAuthorizations().
-  StateAggregator.clearInstance();
+  await StateAggregator.clearInstanceAsync();
 
   // Trigger Apex Test Controller to discover tests after org auth/set-default. Delay so config
   // and TargetOrgRef can propagate before refresh runs.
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-org/src/util/orgUtil.ts
```diff
@@ -102,8 +102,9 @@ export const updateConfigAndStateAggregators = async (): Promise<void> => {
   const configAggregatorProvider = ConfigAggregatorProvider.getInstance();
   await configAggregatorProvider.reloadConfigAggregators();
   // Also force the StateAggregator to reload to have the latest
-  // authorization info.
-  StateAggregator.clearInstance(workspaceUtils.getRootWorkspacePath());
+  // authorization info. Called without args to clear ALL cached instances,
+  // including the default one used by AuthInfo.listAllAuthorizations().
+  StateAggregator.clearInstance();
```
mshanemc: // there's a thread-safe async version so that if somethig else is calling getInstance it won't crash that (there's mutex on the async ones, but not the sync)

```suggestion
  await StateAggregator.clearInstanceAsync();
```

I really need to create a non-cached version of this in sfdx-core.  It's caused so many problems.

PR #7051, opened 2026-03-24
W-21761767 feat: Add url to lwc error

yashank676:
[@W-21761767@](https://gus.lightning.force.com/a07EE00002Wze6xYAB)
### What does this PR do?

- Adds clickable URL links to LWC compiler and template diagnostic errors in the Problems panel
- When an LWC error has an associated documentation URL (e.g. `https://lwc.dev/guide/error_codes#lwcXXXX`), it is now:
  - Appended to the diagnostic message text for visibility on hover
  - Set as a `codeDescription.href` so VS Code renders it as a clickable link on the error code
- Upgrades `@lwc/compiler`, `@lwc/errors`, and `@lwc/template-compiler` from v8.x to v9.0.3 (ESM-only packages)
- Adds CJS compatibility type declarations (`typings/lwc.d.ts`) for the ESM-only `@lwc/*` v9.x packages
- Fixes Jest config to transform `@lwc/*` ESM packages via `transformIgnorePatterns`

Improves developer experience by surfacing LWC error documentation links directly in the editor diagnostics.

### Functionality Before

LWC compiler/template errors displayed only the error message text with no link to documentation.
Screenshot 2026-03-24 at 8 42 20 PM

### Functionality After

LWC errors that have a documentation URL now show:
- The URL appended to the message (e.g. "More Details: https://lwc.dev/guide/error_codes#lwc1001")
- A clickable error code in the Problems panel that links to the relevant documentation page
Screenshot 2026-03-24 at 8 15 30 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2241,7 +2241,7 @@
     },
     "node_modules/@esbuild/aix-ppc64": {
       "version": "0.27.4",
-      "resolved": "https://nexus-proxy.repo.local.sfdc.net/nexus/content/groups/npm-all/@esbuild/aix-ppc64/-/aix-ppc64-0.27.4.tgz",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.27.4.tgz",
       "integrity": "sha512-cQPwL2mp2nSmHHJlCyoXgHGhbEPMrEEU5xhkcy3Hs/O7nGZqEpZ2sUtLaL9MORLtDfRvVl2/3PAuEkYZH0Ty8Q==",
       "cpu": [
         "ppc64"
@@ -2258,7 +2258,7 @@
     },
     "node_modules/@esbuild/android-arm": {
       "version": "0.27.4",
-      "resolved": "https://nexus-proxy.repo.local.sfdc.net/nexus/content/groups/npm-all/@esbuild/android-arm/-/android-arm-0.27.4.tgz",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.27.4.tgz",
       "integrity": "sha512-X9bUgvxiC8CHAGKYufLIHGXPJWnr0OCdR0anD2e21vdvgCI8lIfqFbnoeOz7lBjdrAGUhqLZLcQo6MLhTO2DKQ==",
       "cpu": [
         "arm"
@@ -2275,7 +2275,7 @@
     },
     "node_modules/@esbuild/android-arm64": {
       "version": "0.27.4",
-      "resolved": "https://nexus-proxy.repo.local.sfdc.net/nexus/content/groups/npm-all/@esbuild/android-arm64/-/android-arm64-0.27.4.tgz",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.27.4.tgz",
       "integrity": "sha512-gdLscB7v75wRfu7QSm/zg6Rx29VLdy9eTr2t44sfTW7CxwAtQghZ4ZnqHk3/ogz7xao0QAgrkradbBzcqFPasw==",
       "cpu": [
         "arm64"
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-lwc-language-server/src/javascript/compiler.ts
```diff
@@ -81,13 +81,21 @@ const toDiagnostic = (err: any): Diagnostic => {
...
     range,
     severity: DiagnosticSeverity.Error,
     source: DIAGNOSTIC_SOURCE,
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-    message: extractMessageFromBabelError(message)
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
+    message: err.url ? `${extractMessageFromBabelError(message)}\nMore Details: ${err.url}` : extractMessageFromBabelError(message)
   };
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  if (err.url && err.code != null) {
```
madhur310: We need unit tests for the new code, don't want silent failures down the line

packages/salesforcedx-lwc-language-server/src/template/linter.ts
```diff
@@ -58,13 +58,17 @@ const lintLwcMarkup = (document: TextDocument): Diagnostic[] => {
 
     let warningsLwc: Diagnostic[] = warnings.map((warning) => {
         const { start = 0, length = 0 } = warning.location ?? { start: 0, length: 0 };
-
-        return {
+        const diagnostic: Diagnostic = {
             range: toRange(document, start, length),
-            message: warning.message,
+            message: warning.url ? `${warning.message}\nRef: ${warning.url}` : warning.message,
```
madhur310: We should have linter and compiler throw similar messages.

Lets rename both to `Ref: ...` or `More Details: ...`

packages/salesforcedx-lwc-language-server/src/javascript/compiler.ts
madhur310: Since you have added .d.ts, lets update this to use it. `toDiagnostic(err: CompilerError)  ` would eliminate all the ts lint disables too ?

packages/salesforcedx-lwc-language-server/jest.config.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
   displayName: 'unit',
   transform: {
-    '.ts': 'ts-jest'
+    '.ts': 'ts-jest',
+    'node_modules/@lwc/.+\\.js$': 'ts-jest'
```
madhur310: Seems odd, Is this the right approach? 

Why not something like:

```
'node_modules/@lwc/.+\\.js$': ['babel-jest', {
    plugins: ['@babel/plugin-transform-modules-commonjs']
  }]

```

yashank676: thanks for pointing it out, changed

packages/salesforcedx-lwc-language-server/src/javascript/compiler.ts
```diff
@@ -101,7 +98,8 @@ export const compileSource = (source: string, fileName = 'foo.js'): CompilerResu
     transformSync(source, fileName, transformOptions);
   } catch (err) {
     return {
-      diagnostics: [toDiagnostic(err)]
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      diagnostics: [toDiagnostic(err as CompilerDiagnostic)]
```
madhur310: Why type assertion? I think there are lint rules against that.
Cleaner would be something like:

```
} catch (err) {
  if (err instanceof CompilerError) {
    return { diagnostics: [toDiagnostic(err)] };
  }
  throw err;
}
```

packages/salesforcedx-lwc-language-server/src/javascript/compiler.ts
```diff
@@ -65,29 +66,25 @@ const extractMessageFromBabelError = (message: string): string => {
...
-  return {
+  const diagnostic: Diagnostic = {
     range,
     severity: DiagnosticSeverity.Error,
     source: DIAGNOSTIC_SOURCE,
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-    message: extractMessageFromBabelError(message)
+    message: err.url ? `${extractMessageFromBabelError(message)}\nMore Details: ${err.url}` : extractMessageFromBabelError(message)
   };
+  if (err.url && err.code != null) {
```
madhur310: Could err.code ever be null given the types you defined? I think `!=` check is redundant. May be we should make type as nullable?

yashank676: removed the redundant check since code will always be present

PR #7063, opened 2026-03-25
chore(telemetry): improve span filtering for production exporters W-21746198

mshanemc:
### Summary
Introduces `telemetryIgnore` attribute to selectively exclude spans from production telemetry (App Insights/O11y) while keeping them in local debug logs. This reduces noise in production while maintaining visibility during development.

- Added `isTelemetryIgnored` and `isSpanValidForProductionTelemetry` helpers
- Updated Web and Node exporters to filter ignored spans
- Refactored `ExecuteAnonymousService` for consistent filtering (the diagnostic clear event runs A LOT and we don't need spans for all those)
- Updated App Insights query documentation for OTEL/v3 patterns

@W-21746198@

### DIFF
--- a/.claude/skills/playwright-e2e/SKILL.md
+++ b/.claude/skills/playwright-e2e/SKILL.md
@@ -28,10 +28,11 @@ Shared code (helpers, locators, configuration) for tests.
 
 ## Span files (when debugging traces)
 
-Local only — span export disabled in CI/GHA.
+Available local + CI/GHA.
 
 - Output: `~/.sf/vscode-spans/` — `web-*.jsonl` (test:web), `node-*.jsonl` (test:desktop)
-- Auto-enabled when !CI (no manual enable needed)
+- Auto-enabled (no manual enable needed)
+- CI runs: copied into package `test-results/spans/` artifacts (see workflow upload/download in `references/analyze-e2e.md`)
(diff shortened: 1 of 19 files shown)


packages/salesforcedx-vscode-services/src/observability/applicationInsightsWebExporter.ts
```diff
@@ -55,7 +61,9 @@ export class ApplicationInsightsWebExporter implements SpanExporter {
   // eslint-disable-next-line class-methods-use-this
   public export(spans: ReadableSpan[], resultCallback: (result: ExportResult) => void): void {
     void Effect.runPromise(
-      Effect.all(spans.filter(isTopLevelSpan).map(exportSpan), { concurrency: 'unbounded' }).pipe(
+      Effect.all(spans.filter(span => isTopLevelSpan(span) && !isTelemetryIgnored(span)).map(exportSpan), {
+        concurrency: 'unbounded'
```
peternhale: @mshanemc, the use of unbounded seems risky, especially in the services extension.

mshanemc: agree, I made it a stream without the concurrency

PR #7064, opened 2026-03-25
[W-21655124][W-21655142] chore(soql): render nested subquery columns with grouped headers

peternhale:
change soql render to use group headers for dotted columns
@W-21655124@, @W-21655142@


daphne-sfdc (approved):
SOQL Builder UI correctly nests both relationship fields and subqueries ✅
Image

Output Tab and CSV have their column ordering unchanged ✅

PR #7067, opened 2026-03-25
refactor(soql): remove legacy utils dependencies W-21750066

mshanemc:
### Summary
- Removed dependencies on `salesforcedx-utils` and `salesforcedx-utils-vscode` from SOQL extension.
- Refactored `dataQuery.ts` and `queryUtils.ts` to use `Effect-TS` and `ExtensionProviderService`.
- Enabled `strict` and `noImplicitAny` in `tsconfig.json`.

@W-21750066@

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/commands/queryPlan.ts
+++ b/packages/salesforcedx-vscode-soql/src/commands/queryPlan.ts
@@ -5,6 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { Column, createTable, getServicesApi, Row } from '@salesforce/effect-ext-utils';
+import * as Chunk from 'effect/Chunk';
 import * as Effect from 'effect/Effect';
 import * as HashSet from 'effect/HashSet';
 import * as Schema from 'effect/Schema';
@@ -13,11 +14,13 @@ import { nls } from '../messages';
 import { channelService } from '../services/channel';
 import { formatErrorMessage, getDocumentQueryInputsForPlan, getQueryInputsForPlan } from './queryUtils';
 
-const QueryPlanNote = Schema.Struct({
-  description: Schema.String,
-  fields: Schema.Array(Schema.String),
-  tableEnumOrId: Schema.String
-});
+const QueryPlanNote = Schema.Data(
+  Schema.Struct({
+    description: Schema.String,
+    fields: Schema.Chunk(Schema.String),
+    tableEnumOrId: Schema.String
+  })
+);
 
 const QueryPlanEntry = Schema.Struct({
   cardinality: Schema.Number,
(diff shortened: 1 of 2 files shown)


packages/salesforcedx-vscode-soql/package.json
```diff
@@ -117,9 +114,7 @@
         "package.json",
         "package.nls*.json",
         "../soql-common/out/**",
-        "../salesforcedx-utils-vscode/out/**",
         "../salesforcedx-vscode-i18n/out/**",
```
mshanemc, line 116: note: you shouldn't have to do this...

```
        "../salesforcedx-vscode-i18n:compile",
```
is a dependency, which should have to run (and therefor produce its `out` changes) before this compiles.

packages/salesforcedx-vscode-soql/src/commands/queryPlan.ts
```diff
@@ -65,69 +60,57 @@ const formatQueryPlanResults = (response: QueryPlanResponse): string => {
...
-    .filter(note => {
-      const key = `${note.description}|${note.tableEnumOrId}|${note.fields.join(',')}`;
-      if (seenNotes.has(key)) {
-        return false;
-      }
-      seenNotes.add(key);
-      return true;
-    });
-  if (allNotes.length === 0) {
+  const allNotes = HashSet.fromIterable(plans.flatMap(plan => plan.notes ?? []));
```
mshanemc, line 66: hashSet does "values-based-dedupe" of objects so you don't have to do seen/key/set stuff

[you can provide schema help to it so it can take shortcuts (ex: you have a key-like field or field combo if perf becomes important]

packages/salesforcedx-vscode-soql/src/commands/dataQuery.ts
```diff
@@ -275,43 +284,19 @@ export const formatFieldValueForDisplay = (value: unknown): string => {
...
- * @param query - SOQL query string to execute
- * @param useTooling - Whether to use the Tooling API instead of REST
- */
-const runSoqlQuery = async (connection: Connection, query: string, useTooling = false): Promise<QueryResult> => {
-  channelService.appendLine(
-    nls.localize('data_query_running_query', useTooling ? nls.localize('tooling_API') : nls.localize('REST_API'))
-  );
-  const result = await connection.autoFetchQuery(query, { tooling: useTooling });
-  if (result.records.length > 0 && result.totalSize > result.records.length) {
-    const missingRecords = result.totalSize - result.records.length;
```
mshanemc, line 311: services ext put the warnings from libraries in the channel, and the library here is already doing the warnings.

see subscribeLifecycleWarnings

packages/salesforcedx-vscode-soql/src/commands/queryPlan.ts
```diff
@@ -65,69 +60,57 @@ const formatQueryPlanResults = (response: QueryPlanResponse): string => {
...
+    channelService.clear();
   }
-}
+
+  try {
+    const connection = yield* servicesApi.services.ConnectionService.getConnection();
+    channelService.appendLine(nls.localize('query_plan_running', nls.localize('REST_API')));
+
+    const encodedQuery = encodeURIComponent(query);
+    const path = `/query?explain=${encodedQuery}`;
```
mshanemc, line 89: jsforce lets you start with the stuff after `/services/data/v${apiVersion}` since the connection already has a version.

you really only need to do that if you give the user the option to change the API version OR need to manually set it (like testDiscovery does)

PR #7077, opened 2026-03-26
fix: remove override of vscode settings - W-21220961

madhur310:
@W-21220961@

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>


packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -292,18 +275,11 @@ export abstract class BaseWorkspaceContext {
...
-    await this.writeSettingsJson();
     await this.writeCodeWorkspace();
     await this.writeJsconfigJson();
     await this.writeTypings();
   }
 
-  private async writeSettingsJson(): Promise<void> {
-    const settingsPath = path.join(this.workspaceRoots[0], '.vscode', 'settings.json');
-    const settings = getCoreSettings(this.workspaceRoots);
-    await this.fileSystemAccessor.updateFileContent(settingsPath, JSON.stringify(settings, null, 2));
```
madhur310, line 304: Server shouldn't be doing this at all, this was missed before because of the async logic that prevented this persist from going back to client.

madhur310: For context old LWC/Aura equivalents didn't have this:

https://github.com/forcedotcom/lightning-language-server/blob/main/packages/aura-language-server/src/context/aura-context.ts
https://github.com/forcedotcom/lightning-language-server/blob/main/packages/lwc-language-server/src/context/lwc-context.ts

mshanemc: do you think this was/is useful for Core developers?  I wonder if it some point this was behind some kind of workspaceType check?

or they would be equally miffed to have their settings get squashed?

packages/salesforcedx-lightning-lsp-common/src/baseContext.ts
```diff
@@ -292,19 +292,31 @@ export abstract class BaseWorkspaceContext {
...
     await this.writeJsconfigJson();
     await this.writeTypings();
   }
 
-  private async writeSettingsJson(): Promise<void> {
+  private async writeSettings(): Promise<void> {
+    switch (this.type) {
+      case 'CORE_ALL':
+        await this.updateCoreCodeWorkspace();
+      case 'CORE_PARTIAL':
```
madhur310, line 304: note the fall through, so updateCoreSettings is called for both core WS.

packages/salesforcedx-vscode-lwc/esbuild.config.mjs
```diff
@@ -8,22 +8,12 @@ import { nodeConfig } from '../../scripts/bundling/node.mjs';
 import { commonConfigBrowser } from '../../scripts/bundling/web.mjs';
 import { build } from 'esbuild';
 import { writeFile } from 'fs/promises';
-import { resolve } from 'path';
```
madhur310, line 11: this file

PR #7081, opened 2026-03-27
refactor: integrate improved directory selection for apex test class creation

### LINKED ISSUE #7057: "Create Apex Unit Test Class" command no longer prompts for directory
### Summary

"Create Apex Unit Test Class" command no longer prompts for the directory to save the newly created files to, which it previously did and which it's counterpart command "Create Apex Class" does.

### Steps To Reproduce:

1.  Run "Create Apex Unit Test Class" command
1.  Enter Apex Class name
1.  Select Template

### Expected result

Command prompts for directory to place new files in

### Actual result

Command does not prompt for directory, and instead always creates files in default folder

### Additional information

This is a recent change within the past week or so, not entirely sure if it was made in the last release or the previous one.

Also, with the previous (desired) behavior the pre-populated suggested directories seemed to have a specific directory pattern that it looked for, which unfortunately did not capture the directory structure that we use (`force-app/main/tests/classes`). Not sure what the pattern that it matches was, but it seemed strange that it didn't match on ours despite it being IMO a pretty sensible naming structure. Would be nice if this could be tweaked to be a little more broad-searching for test folders. Not a huge deal, worst case we could update our naming/structure to be more "standard".

**Salesforce Extension Version in VS Code**: 66.2.3

**Salesforce CLI Version**: 2.127.2

**OS and version**: macOS Tahoe 26.3.1

**VS Code version**: 1.112.0

**Most recent version of the extensions where this was working**: Unsure, no more than a couple versions back

johnny-official:
### Problem

This file is the entry point for the "Create Apex Unit Test Class" command. It needs to be updated to leverage the new utility function for generating suggested Apex test class directories and the refined shared UI utility for directory selection. This will restore the prompt for a directory and provide a better user experience with relevant suggestions.

### Changes

This file is the entry point for the "Create Apex Unit Test Class" command. It needs to be updated to leverage the new utility function for generating suggested Apex test class directories and the refined shared UI utility for directory selection. This will restore the prompt for a directory and provide a better user experience with relevant suggestions.

**Modified files:**
- `packages/salesforcedx-vscode-apex/src/commands/forceApexTestClassCreate.ts` (modified)

### Functionality Before
<insert gif and/or summary>

### Functionality After
<insert gif and/or summary>

Closes #7057

### DIFF
--- a/packages/salesforcedx-vscode-apex/src/commands/forceApexTestClassCreate.ts
+++ b/packages/salesforcedx-vscode-apex/src/commands/forceApexTestClassCreate.ts
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { 
+  SfdxCommandlet,
+  SfdxCommandletExecutor,
+  SfdxWorkspaceChecker,
+  ParameterGatherer,
+  ContinueResponse,
+  CancelResponse,
+  PostconditionChecker
+} from '@salesforce/salesforcedx-vscode-core/src/commands/base';
+import { getRootWorkspacePath } from '@salesforce/salesforcedx-vscode-core/src/utils/rootWorkspace';
+import {
+  getApexTestClassDirectorySuggestions,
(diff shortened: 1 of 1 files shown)


mshanemc:
thanks for taking the time to make this.  That change was done as part of a major command migration effort, so we can't merge this to where the old command was.

Try out the new one, and if you think it still has the an issue around directory selection, please give us a new PR

PR #7094, opened 2026-03-28
[W-21651058] feat(soql): add "Get Query Plan" button to SOQL Builder UI

daphne-sfdc:
### What does this PR do?
Adds a new button "Get Query Plan" to the SOQL Builder UI.  When clicked, the query plan appears in the Output Tab.
Screenshot 2026-03-28 at 12 30 17 AM

@W-21651058@

### DIFF
--- a/packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/header/header.css
+++ b/packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/header/header.css
@@ -27,9 +27,38 @@ button {
 }
 
 button:disabled {
+  position: relative;
   background-color: var(--vscode-list-inactiveSelectionBackground, var(--soql-color-medium-grey));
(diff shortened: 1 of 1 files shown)


packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/header/header.css
```diff
@@ -17,6 +17,15 @@ button {
...
 }
 
+.run-button {
+  width: 95px;
+}
+
+.query-plan-button {
+  margin-left: 8px;
+  width: 125px;
+}
```
daphne-sfdc, line 27: I added this so that the "Run Query" and "Get Query Plan" buttons don't change size when clicked.  Before, the button's size was based on the width of the text inside, so the "Run Query" button didn't change size only because "Running..." happened to be the same width.

PR #7150, opened 2026-04-08
chore: enforce knip dead code checks via eslint + CI W-21801858

mshanemc:
### What does this PR do?

Enforces knip dead code detection in CI and local tooling:
- Adds `knip.json` config
- Adds `noExportTaggedErrorInServices` eslint rule
- Removes dead code flagged by knip across multiple packages
- Cleans up skills and verifier docs

@W-21801858@


.claude/skills/feature-branch/SKILL.md
```diff
@@ -5,34 +5,17 @@ description: Create feature branches for all work. Use when creating branches, c
...
 # Feature Branch
 
-All work must be on feature branches. Never commit directly to develop or main.
+Never commit to `develop` or `main`.
 
 ## Do
 
 ```bash
 git fetch origin develop
-git checkout develop
```
kylewalke, line 14: Why do we want to not track the feature branch by default?

kylewalke (approved):
Since it's mostly removal of dead code and such with the changes to the skills and rules, I think it looks all good. I'd prefer to wait until after tdx though as the changes to remove the exports from so many packages means risk I'd want to avoid before code freeze if something uncaught by testing (e2e/unit/manual) slips through the cracks. I'll approve though and start some manual testing.

.claude/skills/feature-branch/SKILL.md
mshanemc: this command means "don't create a feat branch that tracks develop"

[learned the hard way when the AI did that once.  Pushing to the feature branch actually pushed to develop]

PR #7161, opened 2026-04-10
W-21981522: fix(metadata): suppress status bar polling during conflict detection

### LINKED ISSUE #7160: v66.5.2 - Salesforce Services extension is unresponsive
### Summary

After the extension updated to v66.5.2, it is showing as unresponsive - the org browser doesn't load, sfdx deploy/retrieve displays 'activating extensions' and then nothing happens. 

This means we can't deploy/retrieve through the right click context menu, and instead need to resort to using the cli.

### Steps To Reproduce:

1.  Upgrade to v66.5.2
2. Attempt to retrieve/deploy a file

### Expected result

The file retrieves.

### Actual result

The file doesn't retrieve. 'Activating Extensions' is displayed in the bottom corner, and then nothing else visibly happens.

**Salesforce Extension Version in VS Code**: 66.5.2

**Salesforce CLI Version**: 2.130.9

**OS and version**: Mac Tahoe 26.3.1

**VS Code version**: 1.115

**Most recent version of the extensions where this was working**: v.66.4.4 (last known version)

jonnyhork:
## Summary
@W-21981522@

Fixes #7160

### The Issue (#7160)

Users on v66.5.2 reported the extension becoming **completely unresponsive** — deploy/retrieve commands freeze at "Activating Extensions," the Org Browser doesn't load, and the only workaround is dropping to the CLI.

### Root Cause

The source tracking status bar has a polling mechanism that periodically calls `getStatus()` to show remote/local change counts. It uses a shared `ActiveMetadataOperationRef` (an Effect `SubscriptionRef<number>` acting as a semaphore counter) to know when to suppress polling. The `suppressDuringOperation` filter in `sourceTrackingStatusBar.ts` checks this counter:

```ts
const suppressDuringOperation = Stream.filterEffect(() =>
  SubscriptionRef.get(activeOpRef).pipe(Effect.andThen(count => count === 0))
);
```

Deploy and retrieve operations **already** increment this counter via `withActiveMetadataOperationPipeline` in `metadataDeployService.ts` and `metadataRetrieveService.ts`. So during deploys/retrieves, polling correctly pauses.

**The gap:** Conflict detection — which runs `getConflicts()` and `retrieveToCacheDirectory()` under the hood — makes expensive network calls and holds the same source tracking semaphores, but **never incremented the counter**. So the status bar kept firing `getStatus()` polls that queued up behind the conflict detection semaphores, creating a serial chain of network operations. In orgs with many conflicts, this cascade made everything appear frozen.

### The Fix

New utility `withActiveMetadataOperation` in the metadata package wraps an Effect so the active-metadata-operation counter is incremented while it runs, signaling the status bar to suppress polling.

**Why a new utility vs reusing `withActiveMetadataOperationPipeline`?** The existing pipeline utility is internal to `salesforcedx-vscode-services` — it's not part of the public API (not re-exported from `index.ts`). The metadata package follows a strict convention: only `import type` from `salesforcedx-vscode-services`, all runtime access through `ExtensionProviderService` → `getServicesApi` → `api.services.*`. The new utility follows this pattern, accessing the ref via `api.services.ActiveMetadataOperationRef()` rather than reaching into the services package internals.

**Four call sites wrapped:**

| File | Call site | Trigger |
|---|---|---|
| `conflictFlow.ts` | `detectConflicts` | All deploy/retrieve/delete commands |
| `conflictView.ts` | `openConflictViewCommand` | Status bar click → view conflicts |
| `deployOnSaveService.ts` | `deployQueuedFiles` | Deploy-on-save conflict check |
| `deployOnSaveService.ts` | `handleDeployConflict` | Deploy-on-save conflict handling |

### Behavior Change

**Before:**
1. `detectConflicts()` runs — network calls to check for conflicts (counter stays at 0)
2. Status bar sees counter = 0, keeps polling `getStatus()` on every interval tick
3. `getStatus()` tries to acquire the same source tracking semaphores that conflict detection holds
4. Polls queue up behind the locked semaphores, creating a serial chain of network calls
5. The actual deploy/retrieve can't start until the conflict detection + all queued polls drain
6. Result: frozen UI, especially in orgs with many tracked components

**After:**
1. `withActiveMetadataOperation` increments counter to 1
2. `detectConflicts()` runs — network calls proceed normally
3. Status bar sees counter > 0, **skips polling** (both file-change and interval-based triggers are suppressed)
4. Status bar shows spinner instead of stale data
5. Conflict detection finishes, counter decrements to 0
6. Status bar receives `operationComplete` event, fires a single `refresh()` to get fresh data
7. Deploy/retrieve proceeds without competing poll traffic

### Follow-ups (not in this PR)
- Add a timeout to the status bar `refresh` function so it doesn't hang indefinitely if `getStatus()` blocks on semaphores or network
- Cache or debounce `rereadLocal`/`rereadRemote` in `getStatus()` to reduce I/O on every poll
- Filter file change events in the status bar stream to exclude `.sf/` and `.sfdx/` directories (avoids re-triggering polls when conflict detection writes cache files)

## Test plan
- [ ] Verify TypeScript compilation passes (`npx tsc --noEmit` — confirmed locally)
- [ ] Verify existing jest tests pass (`jest --testPathPattern='conflict|deployOnSave'` — 20/20 pass)
- [ ] Manual test: Connect to an org with many conflicts, attempt deploy/retrieve — extension should remain responsive
- [ ] Manual test: Status bar should show spinner during conflict detection instead of continuing to poll
- [ ] Manual test: Deploy-on-save with conflicts should not freeze the extension


packages/salesforcedx-vscode-metadata/src/utils/withActiveMetadataOperation.ts
```diff
@@ -0,0 +1,23 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { ExtensionProviderService } from '@salesforce/effect-ext-utils';
+import * as Effect from 'effect/Effect';
+import * as SubscriptionRef from 'effect/SubscriptionRef';
+
+/** Wrap an effect so the active-metadata-operation counter is incremented while it runs.
+ * This signals the status bar to suppress polling during the wrapped operation. */
+export const withActiveMetadataOperation = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
```
madhur310: why not reuse https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-services/src/core/activeMetadataOperationRef.ts#L20

jonnyhork: The answer is in the description of the PR

jonnyhork: but I am not very familiar with the architecture of the extensions, so still wrapping my head around the right patterns

mshanemc:
I'd like some time to review this before it goes in.  Has it actually been tested on a high-conflict org?

PR #7164, opened 2026-04-10
fix: CBW esbuild web config returns undefined when no org alias set W-21985385

mshanemc:
### What does this PR do?

Scopes `buildWebConfig()` config-map creation and all related logic (file traces, extra settings) inside the `ESBUILD_WEB_ORG_ALIAS` guard so CI builds that lack the env var get `undefined` instead of a populated config object that triggers workspace-config setup.

@W-21985385@

### Before/After

**Before:** `buildWebConfig()` always returned a config string (at minimum `enableFileTraces`), even without `ESBUILD_WEB_ORG_ALIAS`, causing web auth workspace config to execute in CI.

**After:** Returns `undefined` when `ESBUILD_WEB_ORG_ALIAS` is unset; all config logic only runs when the env var is present.


daphne-sfdc (approved):
Tested in Web Console orgfarm.  The Services extension now completes activation with no errors. ✅

Before:
Image

After:
Image

PR #7167, opened 2026-04-11
fix(perf): source tracking contention fixes and STL bumps W-21994811

mshanemc:
### What does this PR do?
- Fix STL semaphore ordering: acquire permits before creating tracking instance to reduce contention
- Remove dead code: `SourceTrackingProvider`, `componentSetUtils` (moved to services)
- Replace multi-pass filter with `Object.groupBy` in `separateChangesByOriginAndConflict`
- Add perf tests for `dedupeStatus`, `calculateCounts`, `separateChanges` (25k rows)
- Truncate conflict channel output at 30 entries
- Annotate STL spans with `statusRows` count
- Bump `@salesforce/source-tracking` across packages

### What issues does this PR fix or reference?
@W-21994811@

Made with [Cursor](https://cursor.com)

PR #7168, opened 2026-04-11
fix(metadata): correct success toasts around conflict override and cancel - W-21994945

CristiCanizales:
### What does this PR do?

Fixes two related issues in the Salesforce Metadata extension around **optional success notifications** (`salesforcedx-vscode-metadata.showSuccessNotification`) and the **deploy/retrieve conflict modal**:

1. **Success toast missing after Override** — `Effect.fn` applies pipeline steps in order. `withConfigurableSuccessNotification` uses `Effect.tap` on the generator effect only. When the generator failed first with `ConflictsDetectedError` and the `catchTag` handler ran the retry successfully, that success never passed through the tap, so the configured success message did not appear after override.

2. **False success toast after View conflicts** — `handleConflictWithRetry` returned `Effect.void` when the user chose “View conflicts” instead of Override, so the command still completed successfully and the success tap could run (notably on deploy paths that wrap an inner effect with the notification).

**Changes**

- **`retrieveManifestCommand`, `retrieveSourcePathsCommand`, `deployManifestCommand`** — Move `withConfigurableSuccessNotification` to run **after** the `ConflictsDetectedError` handler (and after `NoActiveEditor`/manifest errors where applicable), so the tap runs on the final success path including post-conflict retry.
- **`projectDeployStartCommand`** — Apply conflict handling first, then the success notification, then `EmptyComponentSetError`. Order avoids both missing the toast after conflict resolution and showing the deploy-success toast when the only outcome is “no local changes.”
- **`handleConflictWithRetry`** — When the user does not choose Override (`View conflicts`), yield `UserCancellationError` (same pattern as dismissing the modal). `registerCommand` already catches that tag and completes silently, so there is no error toast and no false success toast.

@W-21994945@

**GUS:** [W-21994945](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002XwwBIYAZ/view)

### Functionality Before

- With success notifications enabled, Override after a conflict on retrieve (manifest / source path) or deploy (manifest) did not show the success information message.
- Deploy with conflicts → View conflicts → cancel could still show a success notification.

### Functionality After

- Success notification appears when the user overrides conflicts and the operation completes, if the setting is on.
- Choosing View conflicts (non-override path) does not show a success notification; behavior matches user cancellation.

### DIFF
--- a/packages/salesforcedx-vscode-metadata/src/commands/deployManifest.ts
+++ b/packages/salesforcedx-vscode-metadata/src/commands/deployManifest.ts
@@ -27,7 +27,6 @@ export const deployManifestCommand = Effect.fn('deployManifestCommand')(
       Effect.flatMap(cs => deployComponentSet({ componentSet: cs }))
     );
   },
-  withConfigurableSuccessNotification(nls.localize('command_succeeded_text', nls.localize('deploy_in_manifest_text'))),
   Effect.catchTag(
     'NoActiveEditorError',
     () => new ManifestSelectionRequiredError({ message: nls.localize('deploy_select_manifest') })
@@ -38,5 +37,6 @@ export const deployManifestCommand = Effect.fn('deployManifestCommand')(
       operationType: err.operationType,
       retryOperation: deployComponentSet({ componentSet: err.componentSet })
     })
-  )
+  ),
+  withConfigurableSuccessNotification(nls.localize('command_succeeded_text', nls.localize('deploy_in_manifest_text')))
 );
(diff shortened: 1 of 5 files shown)


daphne-sfdc (approved):
- Deploy with conflicts ✅
  - View Conflicts and Cancel Deploy -> sent to Org Differences view + no notification ✅
  - Override Conflicts and Deploy -> success + notification present ✅
  - Cancel -> nothing happens + no notification ✅
***
- Retrieve with conflicts ✅
  - View Conflicts and Cancel Retrieve -> sent to Org Differences view + no notification ✅
  - Override Conflicts and Retrieve -> success + notification present ✅
  - Cancel -> nothing happens + no notification ✅

mshanemc:
merged into #7167

PR #7170, opened 2026-04-11
fix(metadata): align deploy output with org-applied status and clarify failed deploys W-21995157

CristiCanizales:
### Summary

Deploy results in the **Salesforce Metadata** output channel could list components under **Deployed Source** with API states such as **Created** even when the **overall deploy did not succeed**, so **nothing was applied** to the org. This change keys the “deployed” sections off **`Succeeded` / `SucceededPartial`**, uses separate copy when the deploy did not apply, and replaces misleading per-component API wording with **hypothetical** labels. **Unit tests** cover success, partial success, failure, cancel, missing status, deletes, merged API errors, and multi-component cases.

### Problem

- File-level responses can still show **non-failed** rows (e.g. `Created`, `Unchanged`) while **`response.status`** is **Failed** (or otherwise not a full/partial success), after which **no metadata changes are applied**.
- Listing those under **Deployed Source** and echoing **Created** suggests the org was updated when it was not.

### Solution

- **`deployAppliedToOrg`**: Treat the deploy as applied to the org only when status is **`Succeeded`** or **`SucceededPartial`** (string-literal comparison with a targeted **`eslint-disable`** for `no-unsafe-enum-comparison` so we don’t need a runtime import of `RequestStatus` in this file).
- **When not applied**: Use sections such as **Components without file-level errors — not deployed** (and the analogous **deletes** section), plus a short explanation that **no metadata changes were applied**.
- **Per-component lines when not applied**: Map component change kind to neutral copy (e.g. **Would have been created / updated / had no changes / been deleted**) instead of raw **`Created`** / **`Unchanged`**-style API labels.
- **When applied**: Keep existing **Deployed Source** / **Deleted Source** formatting with raw API state, consistent with real org updates.

### Testing

- **`formatDeployOutput.test.ts`**: Expanded coverage (failed vs succeeded vs partial, canceled, empty status, unchanged/changed/created/deletes, errors-only, merged `componentFailures`, multiple deployed rows, path building with **`path.join`**).

### Files (main)

- `packages/salesforcedx-vscode-metadata/src/shared/deploy/formatDeployOutput.ts`
- `packages/salesforcedx-vscode-metadata/test/jest/shared/deploy/formatDeployOutput.test.ts`

@W-21995157@


kylewalke (approved):

kylewalke (changes requested):
can we rebase/cherry pick the original commit on develop so it's limited in changes to only the 2 files like before?

kylewalke (approved):

PR #7183, opened 2026-04-14
feat(soql): add a 'Set a Default Org' button in the SOQL Builder UI under the blue banner - W-22015596

daphne-sfdc:
### What does this PR do?
Currently when there is no default org set, the SOQL Builder UI hides the dropdowns using a blue banner informing you that you need to authorize to an org.  This PR adds a "Set a Default Org" button underneath that banner so users can directly access the org picker from there instead of having to go through the status bar or command palette.

@W-22015596@

### Functionality Before
Screenshot 2026-04-14 at 2 38 22 PM

### Functionality After
Screenshot 2026-04-14 at 2 35 14 PM


kylewalke:
Screenshot 2026-04-14 at 4 22 48 PM
verified it works as expected!

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/app/app.html
```diff
@@ -29,6 +29,15 @@
               "SFDX: Create a Default Scratch Org" or "SFDX: Authorize an Org"
               to set one.
             </div>
+            <div class="info-notification__dismiss">
+              <button
+                class="text-center"
+                onclick={handleSetDefaultOrg}
+              >
+                <!-- TODO: i18n -->
+                Set a Default Org
```
kylewalke, line 36: Any reason for not including the label here and leaving as a todo instead?

kylewalke (approved):
Just one q about the label otherwise looks good!

jonnyhork:
Nice @daphne-sfdc! Allowing them to connect to fix the issue with one button click will be nice, thanks for doing that.

Not a blocker on this PR but my first impressions were that the UI and button feel disjointed some how like the button is off on it's own or just tacked on there. Have you played with different layouts, for example having the button in the info box, etc. Without iterating on it myself, I'm not sure exactly what I have in mind but I'm just curious if you know there was like option ABC we could explore that might feel more cohesive. Maybe it needs like an outline around the information and the button itself so it's like in one card so to speak.

But the very least maybe we should update the text to reference the button now that we have it they basically still read instructions to say hey run these commands without any mention of "just click here" you know? Maybe Sonal has some thoughts on the text if you feel like you want help with that.

packages/salesforcedx-vscode-soql/src/soql-builder-ui/modules/querybuilder/app/app.html
daphne-sfdc: Yes this is a common question - I answered it for Madhur in another PR 🙂

The SOQL Builder UI is actually a separate LWC app running inside a VSCode webview, different from the SOQL extension itself. Yes we do have i18n setup for the extension itself, but it is not accessible from the SOQL Builder UI. To add i18n for the SOQL Builder UI, I'd need to build out a whole new i18n system for it. I have a GUS WI to add the i18n.

PR #7184, opened 2026-04-14
fix(metadata): match diff pairs by component identity, not path - W-22026667

### LINKED ISSUE #7177: Diff Source Against Org no longer works for classes not located in the default directoryName ("classes")
### Summary

Until recently, the **SFDX: Diff Source Against Org** command worked well with our repository directory structure, but now it returns "No matching files found to diff" for Apex classes not located in a folder named "classes"

### Steps To Reproduce:

1.  In VS Code, either right-click on an Apex class source file or on the containing folder to target all classes.
1.  Select "SFDX: Diff Source Against Org"

### Expected result

Files retrieved from the Org would be compared to the local file(s).

### Actual result

An error message is displayed: "No matching files found to diff"

### Additional information

I can see from the logs that the files are retrieved from the org and stored locally under `/.sf`. The error message appears after that in the logs.

*Background:* Our project has multiple packages and we sort classes into less generically named directories within each package to remove the clutter. We use names like "controllers", "selectors", "models", and "services". This greatly improves organization and discoverability.  This hasn't been a problem until now and it works fine with source deploy and retrieve. According to the [metadataRegistry](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/registry/metadataRegistry.json), Apex classes do not require a `strictDirectoryName`.

To confirm the issue, I renamed "controllers" to "classes" and the command worked as expected. **_This is not a solution!_**

**Salesforce Extension Version in VS Code**:  `66.5.4`

**Salesforce CLI Version**: `@salesforce/cli/2.130.9 linux-x64 node-v22.15.0`

**OS and version**: `Ubuntu 24.04.4`

**VS Code version**: `1.115.0`

**Most recent version of the extensions where this was working**: Not certain, but recent. `66.4.4` maybe?

mshanemc:
@W-22026667@,  https://github.com/forcedotcom/salesforcedx-vscode/issues/7177

### Description
Fixes "Diff Source Against Org" for components not in the default `directoryName`
(e.g. Apex classes stored in a folder other than `classes/`).

Previously `matchUrisToComponents` matched by file path against retrieved
`SourceComponent[]`; path-based matching fails when local and remote directory
names differ. Now matches by component type + fullName using the local
`ComponentSet` and the retrieved `ComponentSet`.

- `matchUrisToComponents` now takes local + retrieved `ComponentSet` instead of `SourceComponent[]`
- `localUriFilter` (was `initialUris`) scopes results to originally-requested files


CristiCanizales (approved):
Thanks for fixing this so quickly! I confirmed the fix works in Mac, CB and web 🙌

CristiCanizales (changes requested):
Please make sure this is working on windows, as the e2e failure seems legit

mshanemc:
got the windows bug fixed; working separately on making soql e2e more stable.

PR #7193, opened 2026-04-15
fix(playwright): stabilize E2E for VS Code 1.116.0 - W-22149130

CristiCanizales:
### What does this PR do?

Fixes Playwright e2e tests broken by VS Code 1.116+ quick-input rendering and Monaco selection handling, and adds shared helpers so other packages can follow the same pattern.

**Shared helpers (`@salesforce/playwright-vscode-ext`)**

- New `activeQuickInputWidget` / `activeQuickInputTextField` — VS Code may retain prior quick-input instances in the DOM; callers must target the active (last) one. From 1.116+ the widget and `input.input` often fail `toBeVisible()` while still interactive — use `waitFor({ state: 'attached' })` and `click`/`fill` with `{ force: true }`.
- New `OpenCommandPaletteOptions.preserveSelection` on `openCommandPalette` / `executeCommandWithCommandPalette`. When set, skips:
  - the focus-click on `.monaco-workbench` before F1 (blurs Monaco on web, clears selection),
  - Escape-based dismiss of existing quick inputs (`cancelSelection` binding wipes the editor selection),
  - the palette-input click after F1 (blur-driven selection reset on web).
  Required for any palette command gated by the `editorHasSelection` context.
- `closeSettingsTab`: also handle the centered/overlay Settings editor (not only the tabbed form); focus + `Ctrl/Cmd+W` as fallback.
- `closeWelcomeTabs` / `ensureSecondarySideBarHidden`: dismiss welcome onboarding overlay, use `force: true` clicks that survive overlay interception.
- `createDesktopTest`: guard `afterEach` when `page` is null (hook timeouts tear the fixture down early); drop `chat.disableAIFeatures: true` user setting.

**Spec fixes**

- SOQL `soql-query-plan` / `soql-run-query` / Apex Log `executeAnonymous`: replace the palette `Select All` step with a single triple-click on `.view-line`. Triple-click is a native Monaco gesture that focuses the editor and creates a real (non-DOM) selection; `click + Ctrl/Cmd+A` on web frequently lands the shortcut at the browser level. Sanity-check with the status-bar `(N selected)` indicator. Pass `{ preserveSelection: true }` to the palette helper so the selection survives palette open.
- SOQL `soql-run-query`: click the "REST API" option instead of `page.keyboard.press('Enter')` — the Enter keystroke sometimes doesn't register on the active quick pick on desktop-electron, leaving the picker open.
- Apex Log `autoCollection`: clean leftover trace flag from a previous run before asserting the "No Tracing" starting state (test-isolation leakage was the root cause); re-confirm status bar after the `upsertSettings` round-trip; bump `verifyCommandExists` timeout.
- Apex Log tests: use `test.describe.configure({ timeout: 180_000 })` instead of inner `test.setTimeout` — the inner call runs too late for fixture (VS Code launch) setup.
- SOQL `soql-builder`: use `activeQuickInputWidget` over `QUICK_INPUT_WIDGET`.
- Command palette and context menu headless specs: migrate to attached-state assertions.

**Misc**

- `salesforcedx-vscode-soql/package.json`: add `effect-ext-utils:compile` to `compile.dependencies` and its `out/**` to files.
- `knip.json`: add entries for `salesforcedx-lightning-lsp-common`.
- `.github/workflows/playwrightVscodeExtE2E.yml` / `orgBrowserE2E.yml`: ignore non-critical errors in PR-scoped steps; YAML formatting.
- `docs/Testing.md`: document the new helpers and `editorHasSelection` gotcha.

@W-22149130@

Test run: [Playwright E2E Full Suite](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/24702001689) ✅


packages/playwright-vscode-ext/src/utils/fileHelpers.ts
```diff
@@ -78,26 +72,50 @@ export const createApexClass = async (page: Page, className: string, content?: s
...
+  }
   await page.keyboard.type(className);
   await page.keyboard.press('Enter');
 
-  // Third prompt: Quick Pick to select output directory - just press Enter to accept default
+  // Third prompt: Quick Pick to select output directory - click first option (default dir).
+  // Same click + Enter fallback pattern as the template prompt.
   await waitForQuickInputFirstOption(page);
-  await page.keyboard.press('Enter');
+  const firstDirOption = activeQuickInputWidget(page).getByRole('option').first();
```
mshanemc: do you think we should have a reuseable "click the first one" to avoid having to work at such a low level?

there might be a few places where you want to choose an alternative but we might have a lot of this default case.

packages/salesforcedx-vscode-soql/test/playwright/specs/soql-run-query.spec.ts
```diff
@@ -35,6 +35,17 @@ const SOQL_QUERY = 'SELECT Id, Name FROM Account LIMIT 10';
 // Waiting for it directly avoids racing with ensureOutputPanelOpen.
 const OUTPUT_PANEL = '[id="workbench.panel.output"]';
 
+// Select "REST API" in the API-type quick pick. Clicking the option is more reliable than
+// `page.keyboard.press('Enter')` — on desktop-electron the Enter keystroke sometimes does
+// not register on the active quick pick, leaving the picker open and the command never executed.
+const selectRestApiOption = async (page: Page): Promise<void> => {
```
mshanemc: maybe there should be a helper for "selectOption" and you pass it some regex like `/^REST API/` etc.

mshanemc (approved):

PR #7201, opened 2026-04-16
ci: fix test tsconfig and compilation across extension packages - W-22063325

mshanemc:
@W-22063325@

### What changes were made?
- Separate test `tsconfig.json` files per package (jest + playwright) so each compiles independently to avoid tests in `out` dir
- Move playwright configs under `test/playwright/` for consistent structure
- Remove composite project references from main `tsconfig.json` where tests don't belong
- Update mock patterns in soql `dataQuery` tests to align with current module shapes
- Remove the `--build` on tsc so that we rely on wireit to have the dep types built instead of crawling through deps at compile time (double compile)
- Enable `isolatedModules: true` in `jest.base.config.js` (~4-7x faster ts-jest); packages using dynamic imports (`salesforcedx-vscode-org`, `salesforcedx-utils-vscode`) override with `false`
- Set `WIREIT_PARALLEL: 1` on Linux and Windows CI — jest already parallelizes internally; competing wireit parallelism was causing slowness/flakiness
- Increase Linux unit test timeout to reduce flakiness
- Fix apex wireit compile output glob (`out/**` → `out/src/**`) to prevent overwriting the language server jar
- Add `isolatedModules: true` globally to `tsconfig.common.json`; fix type-only re-exports across packages (`export type`) to comply
- Replace `const enum`s in `salesforcedx-vscode-lwc` (`TestResultStatus`, `TestInfoKind`, `TestRunType`) with string union types; delete unused `TestType` and `testType` field entirely
- Add `check:knip` to the stop hook so dead exports are caught automatically on every agent stop
- move the LLS tests from `/src/**/__test__` to `/test` to match the other packages
 
I **think** this also fixes the compilation issues with Apex ext


.cursor/plans/lwc_playwright_e2e_tests_30f6a5c2.plan.md
```diff
@@ -0,0 +1,125 @@
+---
```
madhur310: Why part of this PR? Just as forward looking thing, but the test names won't match?

mshanemc: wanted to peek at the complexity of it, decided not to

mshanemc: will delete

madhur310 (approved):

PR #7206, opened 2026-04-16
feat(soql): lint query builder templates against messages catalog - W-21873039

peternhale:
- Move querybuilder i18n catalog to messages/i18n.ts; update imports and tooling
- Extend no-unused-i18n-messages with HTML disk scan; add querybuilder-html-i18n-keys rule
- Integrate @html-eslint for SOQL templates; narrow eslint ignores; VS Code eslint for html
- SOQL package lint: wireit inputs include html; drop eslint file cache for cross-file rules

@W-21873039@

Made-with: Cursor


daphne-sfdc (approved):
Key only present in **i18n.ts** ✅
Image

Key only present in **app.html** ✅
Image

PR #7225, opened 2026-04-20
feat(prompts): filter packageDirs to those containing target folder - W-22049669

mshanemc:
### What does this PR do?

When creating metadata (Apex class, trigger, LWC), prompt now filters `packageDirs` to only those that contain a directory matching the target folder name (e.g. `classes`, `triggers`, `lwc`). Previously all package directories were offered regardless of whether the folder existed.

- Added `readDirectoryWithTypes` to `FsService` — like `readDirectory` but preserves `FileType` for BFS traversal without extra stat calls
- Added `findFoldersByName` BFS helper in `PromptService` to locate matching subdirs under each package root
- `PromptService.chooseDirectory` now deduplicates and filters candidates to dirs that contain the named folder; falls back to all packageDirs if none match
- Added Playwright fixture `desktopFixtures` + new multi-package-dir headless spec to cover the filtering behaviour

@W-22049669@

### Before / After

Before: all `packageDirs` shown regardless of content.
After: only `packageDirs` that contain the target subfolder are offered; if none qualify, all are shown as fallback.


kylewalke:
Screenshot 2026-04-20 at 2 55 56 PM
Screenshot 2026-04-20 at 3 02 04 PM
Screenshot 2026-04-20 at 3 04 07 PM
Screenshot 2026-04-20 at 3 04 27 PM


I was able to see for apex classes, triggers and lwc's the new location showing up for both top level and beyond top level locations when creating new assets via command palette. ✅ 

Approved

kylewalke (approved):

PR #7236, opened 2026-04-22
ci: make coverage opt-in, not default - W-22187169

mshanemc:
### What does this PR do?

Removes coverage collection from default test runs. Coverage must now be explicitly enabled via `COVERAGE=true` env var or a dedicated script. Reduces CI time for the common case.

- Strips `--coverage` from default `test` scripts across all packages
- Removes coverage flags from `unitTestsLinux.yml` workflow
- Updates `Testing.md` with opt-in instructions

@W-22187169@

### Functionality Before
Coverage collected on every `npm test` run and in CI by default.

### Functionality After
Coverage only collected when explicitly requested.


docs/Testing.md
```diff
@@ -29,11 +29,7 @@ This repo uses `jest` for testing and it's probably the best place to start. [Do
 
 ### coverage
 
-Coverage is not collected by default. Pass `--coverage` to jest manually when needed; no requirement to do so.
-
-```bash
-npm run test -w packages/salesforcedx-vscode-core -- --coverage
-```
+We compute test coverage locally but don't send it anywhere, nor is there a requirement to do so
```
madhur310: We don't test coverage by default, should this reflect that?

madhur310 (approved):

PR #7254, opened 2026-04-24
ci(e2e): desktop E2E uses VSIX via E2E_FROM_VSIX - W-21432332

mshanemc:
### What does this PR do?

Adds an opt-in VSIX-based mode for desktop E2E tests. Previously, desktop tests launched VS Code via `--extensionDevelopmentPath`, which loads the raw `tsc` output (`out/src`) — not the bundled `dist/` artifact that ships to customers. With `E2E_FROM_VSIX=1`, the test fixture builds a real VSIX with `vscode:package`, installs it into a hash-keyed cache dir under `.vscode-test/ext-<hash>/`, and launches VS Code with `--extensions-dir` instead of `--extensionDevelopmentPath`. Dev-path remains the default for fast local iteration.

Key implementation details:
- `createDesktopTest.ts`: new `useVsix` option (default: `process.env.E2E_FROM_VSIX === '1'`); new `installedExtensionsDir` worker fixture with per-pid temp dirs to prevent concurrent worker install conflicts
- `salesforcedx-vscode-org-browser`: `test:desktop` wireit deps switch from `vscode:bundle` → `vscode:package` for self and services
- `orgBrowserE2E.yml`: `E2E_FROM_VSIX: 1` added to `e2e-desktop` job env
- All 11 `test:desktop` wireit scripts have `E2E_FROM_VSIX` marked `external` so cache keys are mode-aware

@W-21432332@

### Functionality Before

Desktop E2E loads unbundled `out/src` tsc output via `--extensionDevelopmentPath`. Packaging bugs (wrong `dist/` paths, `.vscodeignore` gaps, `packageUpdates` not applied) are invisible to tests.

### Functionality After

CI runs desktop E2E against the real VSIX (`dist/` as `main`, all packaging transforms applied). Local dev unchanged by default; `E2E_FROM_VSIX=1` opts in locally.


.github/workflows/orgBrowserE2E.yml
```diff
@@ -267,6 +268,7 @@ jobs:
           DREAMHOUSE_ORG_ALIAS: orgBrowserDreamhouseTestOrg
           CI: 1
           VSCODE_DESKTOP: 1
+          E2E_FROM_VSIX: 1
```
daphne-sfdc: I see the new environment variable added here in this step only in this test.  Should be for all E2E test YML files.

.github/workflows/orgBrowserE2E.yml
```diff
@@ -276,6 +278,7 @@ jobs:
...
```
daphne-sfdc: And in this step too: only present in this file.

daphne-sfdc (approved):
Playwright E2E tests run on this commit use VSIXs and are all passing ✅

PR #7258, opened 2026-04-27
refactor(lwc): move internal LWC template command to vscode-metadata - W-22244275

mshanemc:
### What does this PR do?

- Remove `internalLightningGenerateLwc` command from vscode-core
- Reuse `createLwcCommand` in vscode-metadata with `{ internal: true }` option
- Register `sf.internal.lightning.generate.lwc` in vscode-metadata alongside the existing LWC create command
- Add `SettingsService.getInternalDev()` to read the `internal-development` setting
- Set `sf:internal_dev` context key from vscode-services activation (was previously set from vscode-core)
- Remove now-unused `LWC_TYPE` constant, `SelectFileNameAndLanguage` gatherer, and related i18n keys
- Narrow exported type visibility where knip flagged unused exports

@W-22244275@


packages/salesforcedx-vscode-lwc/src/commands/createLwc.ts
```diff
@@ -62,7 +62,13 @@ const determineComponentTemplate = Effect.fn('determineComponentTemplate')(funct
   );
 });
 
-export const createLwcCommand = Effect.fn('createLwcCommand')(function* (outputDirParam?: URI) {
+/** Create LWC via TemplateService from services extension.
+ * outputDir: when invoked from explorer context (right-click lwc folder), VS Code passes the folder URI
+ * options.internal: when true, skips lwc/ dir validation and .js-meta.xml generation (internal dev mode) */
```
madhur310: This comment is dependent on what TemplateService does, so may be we should state that? What if TemplateService changes behavior then this would stale quick?

packages/salesforcedx-vscode-lwc/src/index.ts
```diff
@@ -161,6 +161,9 @@ export const activate = async (extensionContext: ExtensionContext) => {
       yield* registerCommand('sf.metadata.lightning.generate.lwc', (outputDirParam?: URI) =>
         createLwcCommand(outputDirParam)
       );
+      yield* registerCommand('sf.internal.lightning.generate.lwc', (sourceUri?: URI) =>
+        createLwcCommand(sourceUri, { internal: true })
```
madhur310, line 165: Do we need a InternalDevWorkspaceChecker equivalent check here?

mshanemc: I don't think it hurts to register it. It's still gated by the `when` clause (won't show up unless internal)

packages/salesforcedx-vscode-lwc/src/commands/createLwc.ts
mshanemc: there's a subagent for maintaining that kind of thing these days.  😄 

but is your idea just have options.internal exist and not say what that's going to do, since it's in templates library?

madhur310: yeah I was thinking along the lines of the comment should exist on templates library since thats where the functionality exists - and this comment can go stale if someone updates that. But I don't think this is a blocker, so approving.

madhur310 (approved):

packages/salesforcedx-vscode-lwc/src/commands/createLwc.ts
mshanemc: done!

PR #7260, opened 2026-04-27
refactor(aura): move template commands to services-based Effect pattern - W-22247452

mshanemc:
### What does this PR do?

Rebuilds 4 Aura template commands (App, Component, Event, Interface) using the Effect services pattern in `salesforcedx-vscode-lightning`, removing them from `salesforcedx-vscode-core`.

- Add services boilerplate (allServicesLayerRef, runtime, extensionProvider) to Aura extension
- 4 Effect.fn commands with PromptService/TemplateService/FsService
- 8 command registrations (4 public + 4 internal) via `registerCommandWithRuntime`
- Bundle Aura templates (lightningapp, lightningcomponent/aura, lightningevent, lightninginterface) in services extension
- Drop cross-framework LWC/Aura duplicate check at create time (matches LWC create behavior; conflicts surface at deploy)
- Move Playwright desktop E2E tests from core to Aura extension
- Add `auraE2E.yml` CI workflow (desktop-only, no-org, 3 OS matrix)
- Clean up unused exports cascading from removed commands (MetadataTypeGatherer, SimpleGatherer, CompositePostconditionChecker, LwcAuraDuplicateComponentCheckerForCreate, etc.)

@W-22247452@

### Before screenshots

N/A

### After screenshots

N/A


packages/salesforcedx-vscode-lightning/src/commands/createAuraComponent.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!value || value.trim().length === 0) return nls.localize('aura_component_name_empty_error');
+        if (!/^[A-Za-z][A-Za-z0-9_]*$/.test(value)) return nls.localize('aura_component_name_format_error');
+        return undefined;
+      }
+    })
+  ).pipe(
+    Effect.map(raw => raw?.trim()),
+    Effect.flatMap(promptService.considerUndefinedAsCancellation)
+  );
+});
```
madhur310: identical across all, should reuse?

.cursor/plans/move_aura_template_commands_bfe17cec.plan.md
```diff
@@ -0,0 +1,167 @@
+---
+name: Move Aura Template Commands
```
madhur310, line 2: Should we add this to the repo?

packages/salesforcedx-vscode-lightning/src/index.ts
```diff
@@ -23,17 +26,54 @@ import {
...
+  );
+});
+
 export const activate = async (extensionContext: ExtensionContext) => {
   const extensionStartTime = TimingUtils.getCurrentTime();
 
+  // Initialize services layer and register commands
+  const extensionScope = Effect.runSync(getExtensionScope());
+  setAllServicesLayer(buildAllServicesLayer(extensionContext));
+  await getRuntime().runPromise(activateCommands().pipe(Scope.extend(extensionScope)));
```
madhur310: shouldn't this be after activation mode check at line 79? So commands are not callable if language server never starts?

.cursor/plans/move_aura_template_commands_bfe17cec.plan.md
```diff
@@ -0,0 +1,167 @@
...
+
+Template options per command:
+
+- **App**: `LightningAppOptions` -- `{ appname, template: 'DefaultLightningApp', internal }`
+- **Component**: `LightningComponentOptions` -- `{ componentname, template: 'default', type: 'aura', internal }`
+- **Event**: `LightningEventOptions` -- `{ eventname, template: 'DefaultLightningEvt', internal }`
+- **Interface**: `LightningInterfaceOptions` -- `{ interfacename, template: 'DefaultLightningIntf', internal }`
+
+**Note on cross-framework duplicate check:** The old `lightningGenerateAuraComponent` had a `LwcAuraDuplicateComponentCheckerForCreate` that checked the sibling `lwc/` folder before creating an Aura component, enforcing the [platform restriction](https://developer.salesforce.com/docs/platform/lwc/guide/create-components-namespace.html) that LWC and Aura components in the same namespace can't share a name. The new implementation drops this check to match the LWC create command behavior (which also doesn't cross-check). Users will discover conflicts at deploy time.
```
madhur310, line 87: Why the removal of this check, seems like a nice to have QOL thing. What error would the user see at deploy time?

mshanemc: I thought about this one for a bit.  The problem is that it only considers local stuff and not components in the org.  So it's a half-solution.

Probably has some value, but reducing complexity does, too.

LWC Rename command uses it, too, for the same reason.

packages/salesforcedx-vscode-lightning/src/index.ts
mshanemc: they still show up in palette whether the ext activates or not.  But I still agree with the comment

.cursor/plans/move_aura_template_commands_bfe17cec.plan.md
mshanemc: 🤷🏻 we discussed it on a standup a while back.  I think I'm the only one who does it, and even then I only keep the interesting ones.  Seems harmless.

.cursor/plans/move_aura_template_commands_bfe17cec.plan.md
mshanemc: LWC create wasn't doing the inverse check.  And I had concerns about the NS and internal dev cases, and also would miss ones in another package directory.

Thoughts?  Worth trying to plug all the holes to bring it back?

madhur310 (approved):

PR #7261, opened 2026-04-28
feat(soql): add a Max Rows input to SOQL Builder UI to allow max rows retrieved to be defined by the user - W-22199672

daphne-sfdc:
### What does this PR do?
- Previously, the SOQL Builder UI retrieves only the first 2000 rows if there are more than 2000 rows in the result, with no way to change that value.  This PR adds a new `Max Query Limit` setting for users to change that value if they'd like.  This is the equivalent of the `org-max-query-limit` config variable that we use when running SOQL queries via CLI.  The setting's default value is 50000 rows.
- If the result set is truncated due to the number of rows being greater than `Max Query Limit`, there is an info icon placed next to the "Returned X of Y total records" message that informs users about the setting when hovered.
- The Text Editor View's code lens and command palette commands now use the setting instead of `org-max-query-limit`.  This change was made because of the following reasons: (1) Consistency between the Builder and Text Editor views: the setting changes both. (2) Web Console cannot have CLI installed, so users cannot change `org-max-query-limit` there, and the value is always fixed to 10000.  Now the setting provides a way for the value to be changed in Text Editor View.

#5373, @W-22199672@

### Functionality Before
Always 2000 rows
Screenshot 2026-04-27 at 8 21 55 PM

### Functionality After
`Max Rows` set to 1111
Screenshot 2026-04-29 at 4 33 44 PM

`Max Rows` unset - default changed to 50000 rows
Screenshot 2026-04-29 at 4 33 08 PM

Hovering over the info icon:
Screenshot 2026-04-29 at 5 07 47 PM

Text Editor View now uses the `Max Query Limit` setting.  When there are truncated results:
Screenshot 2026-04-29 at 5 12 12 PM


packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
```diff
@@ -17,16 +17,16 @@ const hasMessage = (obj: unknown): obj is { message: unknown } =>
 const getErrorMessage = (error: unknown): string =>
   error instanceof Error ? error.message : hasMessage(error) ? String(error.message) : String(error);
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-type QueryResult<T> = Awaited<ReturnType<Connection['query']>>;
+type QueryResult = Awaited<ReturnType<Connection['query']>>;
 
 export const runQuery =
   (conn: Connection) =>
-  async (queryText: string, options = { showErrors: true }): Promise<QueryResult<JsonMap>> => {
```
jonnyhork, line 25: We were passing in `showErrors = true` before, is it intentional to pass in `undefined` now? The error might be caught by some other handler but just want to make sure this doesn't result in some behavior change?

packages/salesforcedx-vscode-soql/src/soql-data-view/queryDataViewController.js
```diff
@@ -78,6 +78,27 @@
...
+          tooltipText.style.display = 'block';
+        });
+        hintEl.addEventListener('mouseleave', () => {
+          tooltipText.style.display = 'none';
+        });
+        hintEl.dataset.listenersAttached = 'true';
+      }
+    } else {
+      hintEl.setAttribute('hidden', '');
+    }
```
jonnyhork, line 88: While this works, it stands out as a bit of a code smell to me because we are mixing CSS/styling in to the JS and setting inline styles. I would recommend to just use `:hover` on the hint icon container and let CSS
handle the tooltip display entirely? Something like `.info-tooltip:hover .info-tooltip__text { display: block }` with `position: absolute` relative to the icon. That way you don't need the mouseenter/mouseleave listeners or the getBoundingClientRect math at all, and the JS just manages showing/hiding the hint element via the hidden attribute.

jonnyhork (approved):
No blockers but I recommend refactoring the tooltip a bit =)

packages/salesforcedx-vscode-soql/src/editor/queryRunner.ts
daphne-sfdc: Good catch!  I put back `showErrors = true`.

PR #7273, opened 2026-04-29
W-21981771: Use workflow_call to trigger CBW release after publish

jonnyhork:
## Summary
@W-21981771@

- Replace `repository_dispatch` API call with `uses:` syntax to call `release-on-extension-publish.yml` in `code-builder-web`
- Add `scripts/parse-extension-names.js` to extract published extension names from VSIX filenames and pass them as `workflow_call` inputs — CBW filters to only the extensions it bundles
- Determine CBW release type (patch vs minor) by comparing the current version tag against the previous one

### Dependency
- Depends on forcedotcom/code-builder-web#76 (must be merged first — provides the `release-on-extension-publish.yml` workflow)

## Test plan
- [ ] Verify script extracts correct names: `echo "salesforcedx-vscode-apex-log-66.5.2.vsix" > /tmp/test && node scripts/parse-extension-names.js /tmp/test`
- [ ] End-to-end: after both PRs merge, next extension publish triggers CBW release via `uses:` and only polls bundled extensions

### DIFF
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -126,4 +126,3 @@ jobs:
       extensions: ${{ needs.publish.outputs.extensions }}
       version: ${{ needs.prepare-environment-from-main.outputs.RELEASE_VERSION }}
       release-type: ${{ needs.prepare-environment-from-main.outputs.CBW_RELEASE_TYPE }}
-    secrets: inherit
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -87,7 +87,7 @@ jobs:
       - name: Collect published extension names
         id: collect-extensions
         run: |
-          NAMES=$(node scripts/collect-extension-names.js ./extensions)
+          NAMES=$(node scripts/parse-extension-names.js ./extensions)
           echo "names=$NAMES" >> "$GITHUB_OUTPUT"
           echo "Published extensions: $NAMES"
       - uses: salesforcecli/github-workflows/.github/actions/npmInstallWithRetries@main
--- a/.github/workflows/publishVSCode.yml
+++ b/.github/workflows/publishVSCode.yml
@@ -87,16 +87,7 @@ jobs:
       - name: Collect published extension names
         id: collect-extensions
         run: |
-          NAMES=$(node -e "
-            const fs = require('fs');
-            const names = fs.readdirSync('./extensions')
-              .filter(f => f.endsWith('.vsix'))
-              .map(f => f.replace(/-\d+\.\d+\.\d+\.vsix$/, ''))
-              .filter((v, i, a) => a.indexOf(v) === i)
-              .sort()
-              .join(',');
-            process.stdout.write(names);
-          ")
+          NAMES=$(node scripts/collect-extension-names.js ./extensions)
           echo "names=$NAMES" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


.github/workflows/publishVSCode.yml
```diff
@@ -108,24 +116,13 @@ jobs:
       status: Not Implemented
 
   # Set repo variable CBW_TRIGGER_ENABLED=false to disable dispatch to code-builder-web.
-  # Default (unset): dispatch after publish. Use when publishing without CBW promotion.
+  # Default (unset): trigger CBW release after publish. Use when publishing without CBW promotion.
   trigger-cbw-release:
     needs: [publish, prepare-environment-from-main]
     if: needs.publish.result == 'success' && vars.CBW_TRIGGER_ENABLED != 'false'
     continue-on-error: true
```
mshanemc: this fails schema validation with 

Image

you should install the gha ext 

fyi:
continue-on-error is not valid on a job that uses uses: (reusable workflow call). actionlint flags this — schema reference: [reusable-workflow job keys](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow). Allowed keys are name, uses, with, secrets, needs, if, permissions, strategy, concurrency. To get retry/ignore-failure behavior, wrap the call or move continue-on-error to a step in a non-reusable job.

mshanemc: you can also temporarily change the `@main` to the branch on your repo so that it checks your work for you

jonnyhork: I have the GHA ext installed. continue-on-error is the only real bug. 
image

jonnyhork:
@mshanemc I made some updates:
- Removed `continue-on-error` — you're right, it's not valid on `uses:` jobs. Since nothing `needs` this job, a failure won't block the workflow. And honestly we'd want to know if the trigger fails.
- Added `secrets: inherit` — the CBW reusable workflow now generates a short-lived GitHub App token to dispatch `release.yml`. The app credentials live as org-level secrets (`CBW_DISPATCH_APP_ID` variable + `CBW_DISPATCH_APP_PRIVATE_KEY` secret), so `secrets: inherit` passes them through.

**Upodated the way we auth:**
The previous `repository_dispatch` approach used `IDEE_GH_TOKEN` to POST directly to code-builder-web. When I switched to `workflow_call`, I didn't account for the context flip — `github.token` and `github.repository` resolve to the *caller* repo (salesforcedx-vscode), not code-builder-web. So the `gh workflow run` was targeting the wrong repo with an unprivileged token.

The fix on the CBW side: generate a GitHub App token scoped to `forcedotcom/code-builder-web` and hardcode the `--repo` flag. No more reliance on `github.token` or `github.repository`.

**Re: cross-workflow invocation (your comment on the CBW PR):**
I can't nest `workflow_call` here because secrets don't cross the repo boundary — the reusable workflow runs in the caller's context, so code-builder-web's AWS/release secrets aren't available. `gh workflow run` is the right pattern for this; it dispatches a fresh run in CBW's own context. The app token + hardcoded repo addresses the drift risk you flagged.

**Onboarding new repos:** add the repo to the org secret's access list + add the `workflow_call` job. That's it — documented in the CBW README.

.github/workflows/publishVSCode.yml
mshanemc: yes, that's correct.

mshanemc (approved):

PR #7279, opened 2026-04-30
ci: validate github actions workflows on PR - W-22294861

mshanemc:
### What does this PR do?

Adds a `check:actions` script that runs [`@action-validator/core`](https://www.npmjs.com/package/@action-validator/core) over `.github/workflows/*.yml` and `.github/actions/*/action.yml`. Fails on schema errors so broken YAML/expression typos are caught at PR time instead of post-merge.

- `scripts/validateActions.ts` — Effect Stream over both glob sets; logs failures, exits non-zero on any errors
- wireit `check:actions` task; included in `check:everything`
- new `validate-actions` job in `validatePR.yml`
- docs in `contributing/developing.md` and verification skill/cursor rule
- eslint override to allow `node:fs` in this script

Sample run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/25175960233/job/73807993686?pr=7279

@W-22294861@

### DIFF
--- a/scripts/validateActions.ts
+++ b/scripts/validateActions.ts
@@ -5,22 +5,33 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { validateAction, validateWorkflow } from '@action-validator/core';
+import { type ValidationError, type ValidationState, validateAction, validateWorkflow } from '@action-validator/core';
 import { Console, Effect, Stream } from 'effect';
 import { globSync, readFileSync } from 'node:fs';
 
 const program = Stream.concat(
-  Stream.fromIterable(globSync('.github/workflows/*.yml')).pipe(
+  Stream.fromIterable(globSync('.github/workflows/*.{yml,yaml}')).pipe(
     Stream.map(file => ({ file, result: validateWorkflow(readFileSync(file, 'utf8')) }))
   ),
-  Stream.fromIterable(globSync('.github/actions/*/action.yml')).pipe(
+  Stream.fromIterable(globSync('.github/actions/*/action.{yml,yaml}')).pipe(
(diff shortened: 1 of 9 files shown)


jonnyhork:
The validation output could use a bit of formatting? If there are a lot of errors the raw JSON output would be hard to read, especially with the redundant properties.
image

Could we clean simplify the report as
```
.github/workflows/publishVSCode.yml:
  /jobs/trigger-cbw-release: Additional property 'secrets' is not allowed
  /jobs/trigger-cbw-release: Additional property 'uses' is not allowed
  /jobs/trigger-cbw-release: Additional property 'with' is not allowed
  /jobs/trigger-cbw-release/runs-on: This property is required
  /jobs/trigger-cbw-release: Additional property 'continue-on-error' is not allowed
  ```
  
I am not sure this can be helped but in the example above only the `continue-on-error` is a real validation error, but you still get misleading errors in the report e.g. everything else is not a real failure. If I remove the `continue-on-error` the validation passes.

This could still be valuable to have but the devs on the receiving end will be misled for a bit.

File: scripts/validateActions.ts [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { validateAction, validateWorkflow } from '@action-validator/core';
+import { Console, Effect, Stream } from 'effect';
+import { globSync, readFileSync } from 'node:fs';
+
+const program = Stream.concat(
+  Stream.fromIterable(globSync('.github/workflows/*.yml')).pipe(
```
jonnyhork: Do we want to cover `.yaml` to just incase?

jonnyhork (approved):
Some UX improvements to be considered, but not blocking

PR #7267, opened 2026-04-29
refactor(extensions): use more services, remove activation tracker - W-22267536

mshanemc:
### What does this PR do?
- Remove `ActivationTracker` and related utils from `salesforcedx-utils-vscode`; extensions use services telemetry instead
- Move Aura template commands (create app/component/event/interface) from `vscode-core` to `vscode-lightning`
- Move analytics template generation command from `vscode-core` to `vscode-metadata`; add headless Playwright web E2E spec
- Add LWC LSP Playwright E2E coverage for web and desktop (indexing, SFDX typings, autocomplete, go-to-definition, custom component index), including CI and shared workspace/auth utilities
- Convert multiple extensions (`apex`, `apex-debugger`, `apex-oas`, `apex-replay-debugger`, `lwc`, `lightning`, `org`) to consume `salesforcedx-vscode-services` via `extensionProvider.ts` + `runtime.ts` pattern
- Reduce `vscode-lwc` dependencies on `salesforcedx-utils-vscode` (file watchers, metaSupport, path utils)
- Record extension pack telemetry through services instead of custom activation tracker
- Move internal LWC template command to `vscode-metadata`
- Remove core's tracking-enabled/disabled management; telemetry state handled entirely by services
- Get connection/userId from services instead of utils
- Disassemble remaining utils helpers (`validations`, `internalCommandUtils`, `overwriteComponentPrompt`, `flagParameter`, `compositePostconditionChecker`) - delete dead code, inline survivors
- Remove dead public API surface from vscode-core (`activate()` return, exports) informed by new skills documenting actual external consumers
- Drop dead template-command path-strategy machinery in core (`MetadataDictionary`, `PathStrategyFactory`/`SourcePathStrategy`, `SelectOutputDir`) - `LibraryProjectGenerateExecutor` is the only remaining subclass and overrides `openCreatedTemplateInVSCode`
- `config list` command uses `@salesforce/core` library directly instead of shelling out to CLI; headless Playwright e2e spec added
- Update verification guidance to rerun transient Wireit internal errors before treating them as failures

### What issues does this PR fix or reference?
@W-22267536@

### Functionality Before
- Activation tracker in utils-vscode measured extension activation times with custom timer logic
- Aura and analytics template commands lived in vscode-core
- LWC LSP coverage did not include the new headless web/desktop Playwright suite for indexing, typings, completion, definitions, and component index behavior
- Extensions used ad-hoc patterns for service initialization
- `config list` shelled out to CLI
[...]

PR #7282, opened 2026-05-01
refactor(services): share buildAllServicesLayer helper; normalize apex-replay-debugger - W-22267536

kylewalke:
### What does this PR do?

Two review-suggestion refactors for #7267:

**1. Share `buildAllServicesLayer` helper.** Extracted the duplicated `ExtensionProviderServiceLive` definition + `Layer.unwrapEffect(...)` body into `@salesforce/effect-ext-utils`. Each consumer's `services/extensionProvider.ts` collapses to a thin wrapper that supplies its channel-name fallback. Applied to apex, apex-debugger, apex-oas, core, lwc, and apex-replay-debugger. `jscpd` over those `services/` dirs now reports 0 clones.

**2. Normalize `apex-replay-debugger` to the typed pattern.** Deleted `services/allServicesLayerRef.ts` (the `Layer<any, never>` opaque cast) and switched `runtime.ts` to import `AllServicesLayer` directly from `./extensionProvider`, matching apex / apex-debugger / apex-oas / core / lwc.

The opaque cast was masking a real type gap: `prebuiltServicesDependencies`' runtime context (built from `globalLayers` in the services extension) provides `TraceFlagService.Default` and `ExecuteAnonymousService.Default`, but the static `Context.Context<…>` union under-reported them. apex-replay-debugger consumes both. Added them to the `SalesforceVSCodeServicesApi` union — additive, non-breaking — so the typed pattern compiles.

Net diff: +69 / −220.

`packages/salesforcedx-vscode-lightning/src/services/allServicesLayerRef.ts` uses the same opaque indirection but landed via #7260 and is out of scope here; can be normalized in a follow-up.

@W-22267536@


packages/salesforcedx-vscode-lwc/src/services/extensionProvider.ts
```diff
@@ -5,43 +5,11 @@
...
-import {
-  ExtensionPackageJsonSchema,
-  ExtensionProviderService,
-  type ExtensionPackageJson,
-  getServicesApi
-} from '@salesforce/effect-ext-utils';
-import * as Effect from 'effect/Effect';
-import * as Layer from 'effect/Layer';
-import * as Schema from 'effect/Schema';
+import { buildAllServicesLayer as buildAllServicesLayerCore } from '@salesforce/effect-ext-utils';
```
madhur310: why not just update the import using packages/salesforcedx-vscode-lwc/src/services/extensionProvider.ts 's buildAllServicesLayer method instead of this?

packages/salesforcedx-vscode-lwc/src/services/extensionProvider.ts
```diff
@@ -5,43 +5,11 @@
...
-        Layer.succeedContext(api.services.prebuiltServicesDependencies),
-        ExtensionProviderServiceLive,
-        errorHandlerWithChannel,
-        api.services.ExtensionContextServiceLayer(context),
-        api.services.SdkLayerFor(context),
-        channelLayer
-      );
-    }).pipe(Effect.provide(ExtensionProviderServiceLive))
-  );
+  buildAllServicesLayerCore(context, 'Lightning Web Components');
```
madhur310: i18n?

madhur310: same for other files too

madhur310 (approved):
Approving with suggestions (but don't think they are blockers)

madhur310 (approved):
Approving with suggestions (but don't think they are blockers)

packages/salesforcedx-vscode-lwc/src/services/extensionProvider.ts
kylewalke: Good catch, I can add those in.

packages/salesforcedx-vscode-lwc/src/services/extensionProvider.ts
kylewalke: My change was to dedupe the repeated effect logic in the extensionProviders in all the extensions so that the export buildAllServicesLayer from the extensionProviders services all pull from a common source which is the effect-ext-utils file defined above. I could leave it in the core ext but the vibe I got was that we don't want to be adding more into core-ext and actively want to be moving things out of it.

kylewalke: Cue the whole salesforce-services-ext vs salesforce-core-ext conversation that's been brewing

PR #7303, opened 2026-05-08
fix(metadata): add success notification for Deploy on Save - W-22399776

daphne-sfdc:
### What does this PR do?
When users have the `salesforcedx-vscode-metadata.showSuccessNotification` setting turned on, they will get success notifications when deploys complete successfully.  However, that functionality did not translate to Deploy on Save.  This PR adds success notifications for Deploy on Save when the setting is checked.

Screenshot 2026-05-08 at 12 19 24 AM

@W-22399776@

### DIFF
--- a/packages/salesforcedx-vscode-metadata/src/services/deployOnSaveService.ts
+++ b/packages/salesforcedx-vscode-metadata/src/services/deployOnSaveService.ts
@@ -21,6 +21,7 @@ import { nls } from '../messages';
 import { getDeployOnSaveEnabled, getIgnoreConflicts } from '../settings/deployOnSaveSettings';
 import { deployComponentSet } from '../shared/deploy/deployComponentSet';
 import { DeployCompletedWithErrorsError } from '../shared/deploy/deployErrors';
+import { withConfigurableSuccessNotification } from '../utils/withConfigurableSuccessNotification';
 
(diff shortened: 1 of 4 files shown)


packages/salesforcedx-vscode-metadata/src/services/deployOnSaveService.ts
```diff
@@ -80,7 +81,9 @@ const deployQueuedFiles = Effect.fn('deployOnSave:deployQueuedFiles')(function*
     }
   }
 
-  return yield* deployComponentSet({ componentSet });
+  return yield* deployComponentSet({ componentSet }).pipe(
```
peternhale: @daphne-sfdc I see deploy on save as something that goes unmentioned. As user enables it expecting silence.

Also, I typically enable vscode "autosave", with an idle delay of 1 second. Imagine the notification storm with autosave and deploy on save enabled.

peternhale (changes requested):

packages/salesforcedx-vscode-metadata/src/services/deployOnSaveService.ts
daphne-sfdc: Thanks for the feedback.  I have moved Deploy on Save success notifications to a separate setting.
Screenshot 2026-05-08 at 9 36 39 AM

